PR #53, opened 2022-04-01
TorchOpt compatibility fixes

vmoens:


vmoens:
The PR is good to be merged IMO but some updated docstrings could help.

PR #62, opened 2022-04-13
Add methods _TensorDict.values() and _TensorDict.values_meta

DjImene:


vmoens:
> 

Awesome thanks for this!
Could you write some tests in `test/test_tensordict.py`?

DjImene:
Thank you, I am still trying to figure the test part, I am going through the test file. Do you have any guidelines?

vmoens:
> Thank you, I am still trying to figure the test part, I am going through the test file. Do you have any guidelines?

actually I think it'd be best to write a group of tests for keys(), values() and items(), but this is beyond the scope of this PR.
Thanks for the work!

PR #91, opened 2022-04-25
Add tensorboard dependency to setup.py

shagunsodhani:


vmoens (approved):
Well spotted @shagunsodhani! Thanks for this

Issue #97, opened 2022-04-26
Batched environment is not closed 


shagunsodhani:
Error Log:

```
Traceback (most recent call last):
  File "/private/home/sodhani/projects/rl/torchrl/envs/vec_env.py", line 256, in __del__
RuntimeError: Batched environment must be explicitely closed before it turns out of scope.
```
Full Trace:

```
(torch_rl) ➜  rl git:(main) ✗ python examples/ppo/ppo.py --config=examples/ppo/configs/humanoid.txt --total_frames=10

/private/home/sodhani/.conda/envs/torch_rl/lib/python3.9/site-packages/torchvision/transforms/functional_pil.py:228: DeprecationWarning: BILINEAR is deprecated and will be removed in Pillow 10 (2023-07-01). Use Resampling.BILINEAR instead.
  interpolation: int = Image.BILINEAR,
...
  interpolation: int = Image.NEAREST,
/private/home/sodhani/.conda/envs/torch_rl/lib/python3.9/site-packages/torchvision/transforms/functional_pil.py:328: DeprecationWarning: BICUBIC is deprecated and will be removed in Pillow 10 (2023-07-01). Use Resampling.BICUBIC instead.
  interpolation: int = Image.BICUBIC,
collector = MultiSyncDataCollector();
loss_module = ClipPPOLoss(
  (actor): ProbabilisticActor(module=NormalParamWrapper(
    (operator): MLP(
      (0): Linear(in_features=67, out_features=400, bias=True)
      (1): Tanh()
      (2): Linear(in_features=400, out_features=300, bias=True)
      (3): Tanh()
      (4): Linear(in_features=300, out_features=42, bias=True)
    )
  ), distribution_class=<class 'torchrl.modules.distributions.continuous.TanhNormal'>, device=cuda:0)
  (critic): ValueOperator(
      module=MLP(
        (0): Linear(in_features=67, out_features=400, bias=True)
        (1): Tanh()
        (2): Linear(in_features=400, out_features=300, bias=True)
        (3): Tanh()
        (4): Linear(in_features=300, out_features=1, bias=True)
      ),
      device=cuda:0,
      in_keys=['observation_vector'],
      out_keys=['state_value'])
);
recorder = TransformedEnv(env=DMControlEnv(env=humanoid, task=walk, batch_size=torch.Size([])), transform=Compose(
	VideoRecorder(keys=['next_pixels']),
	RewardScaling(loc=0.0000, scale=1.0000, keys=['reward']),
[...]

PR #98, opened 2022-04-26
Fix PPO open env

vmoens:
closes the env as per #97 
Long-term fix should be to close all collectors and envs upon deletion, but we should be careful with that.

### DIFF
--- a/examples/ppo/ppo.py
+++ b/examples/ppo/ppo.py
@@ -119,6 +119,8 @@ if __name__ == "__main__":
         recorder_rm = recorder
 
     recorder_rm.load_state_dict(create_env_fn.state_dict()["worker0"])
+    create_env_fn.close()
+
     # reset reward scaling
     for t in recorder.transform:
         if isinstance(t, RewardScaling):
--- a/examples/ddpg/ddpg.py
+++ b/examples/ddpg/ddpg.py
@@ -146,6 +146,7 @@ if __name__ == "__main__":
     else:
         recorder_rm = recorder
     recorder_rm.load_state_dict(create_env_fn.state_dict()["worker0"])
+    create_env_fn.close()
     # reset reward scaling
     for t in recorder.transform:
         if isinstance(t, RewardScaling):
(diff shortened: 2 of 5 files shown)


examples/ppo/ppo.py
```diff
@@ -119,6 +119,8 @@ def make_args():
         recorder_rm = recorder
 
     recorder_rm.load_state_dict(create_env_fn.state_dict()["worker0"])
+    create_env_fn.close()
+
```
shagunsodhani, line 123: Not sure if I follow but shouldnt we close the env after training?

vmoens: not really
here what happens is that we might be normalizing observations on the fly using `VecNorm`
`VecNorm` uses shared memory to place the normalization stats (location and scale) such that they're available for every process.
If you have a recording environment, you must sync it with the others (i.e. give it access to the shared tensors). This is what `load_state_dict` does.
Now our `create_env_fn` is actually a `ParallelEnv` instance, and when we query its state dict, it must (1) run every process (which is not done upon creation of the object but only when needed, for instance when calling `reset` or `state_dict`) and (2) read the constant from one process.
Thing is that once we have queries `state_dict` we have also run the processes. Then they should be closed.


As a side note, I mentioned that `create_env_fn` is a parallel env. If you look at its `__init__`, it stores a (list of) picklable environment creation functions and stays idle until it is queried. When we create a data collector, we pass it a list of functions/objects to run on each subprocess. If it's a function, then each subprocess will create an environement from that function and use it to collect data. If it's an object (e.g. ParallelEnv) then each process will play with it to collect data. The ParallelEnv can be sent to the workers as it does not contain an instantiated env but only a pickable function that creates envs.
I made this very ugly sketch to show how creating envs in the collector works in the single and multi process case, hope you'll like it :p 

env dispatch 001

shagunsodhani: Thanks for the detailed response! Appreciate it :)

PR #103, opened 2022-05-02
Issue 85: Add tests for catframe

JohnnyC08:
For issue: https://github.com/facebookresearch/rl/issues/85

Fixed a typo of `franes` -> `frames`

Added three tests:

```
test_catframes_transform_observation_spec
test_catframes_buffer_check_latest_frame
test_catframes_reset
```

The test `test_catframes_transform_observation_spec` tests the transform_observation_spec to ensure the spaces of the frames are updated.

The test `test_catframes_buffer_check_latest_frame` checks the contents of the transformer and ensures that the frames are concatenated and that if the buffer isn't full, then the current frame is propagated forward.

The test `test_catframes_reset` checks that the buffer gets emptied and that we pass the tensor dict back correctly.


test/test_transforms.py
```diff
@@ -525,14 +525,80 @@ def test_observationnorm(
...
+
+        result = cat_frames.transform_observation_spec(observation_spec)
+        observation_spec = CompositeSpec(
+            **{
+                key: NdBoundedTensorSpec(space_min, space_max, (1, 32, 32), dtype=torch.double)
+                for key, space_min, space_max in zip(keys, mins, maxes)
+            }
+        )
+
+        print()
```
vmoens, line 551: do we have to keep it?

test/test_transforms.py
```diff
@@ -525,14 +525,80 @@ def test_observationnorm(
...
+            dict(zip(keys, key_tensors)), [1]
+        )
+        cat_frames = CatFrames(N=N, keys=keys)
+
+        cat_frames(td)
+        buffer_length1 = len(cat_frames.buffer)
+        passed_back_td = cat_frames.reset(td)
+
+        assert 2 == buffer_length1
+        assert td == passed_back_td
```
vmoens, line 600: `is` instead of `==`?

vmoens (approved):
Have a look at the two minor comments, otherwise LGTM! Thanks for this!

test/test_transforms.py
JohnnyC08: Wow my mistake from debugging. Sorry for that I will fix.

PR #107, opened 2022-05-07
Sync latest RLMeta SegmentTree implementation to TorchRL

xiaomengy:


vmoens:
@xiaomengy thanks for this brilliant contribution!

Here's the [error message](https://github.com/facebookresearch/rl/files/8645726/error.txt) from the built, any tips on how to solve this?

vmoens:
@xiaomengy I fixed the issues by adopting a code structure more similar to what you have in rlmeta.
Does it make sense to you?

xiaomengy:
> @xiaomengy I fixed the issues by adopting a code structure more similar to what you have in rlmeta. Does it make sense to you?

Thanks for the update. Actually I was planning to do the same thing in the next PR

PR #114, opened 2022-05-09
Extend tensor-like behaviour of TensorDict classes

xavoliva:
Issue: https://github.com/facebookresearch/rl/issues/58

### DIFF
--- a/test/test_tensordict.py
+++ b/test/test_tensordict.py
@@ -231,22 +231,22 @@ def test_permute(device):
         "c": torch.randn(4, 5, 6, device=device),
     }
     td1 = TensorDict(batch_size=(4, 5, 6), source=d)
-    td2 = td1.permute(2, 1, 0)
+    td2 = torch.permute(td1, dims=(2, 1, 0))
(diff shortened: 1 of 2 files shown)


vmoens (commented):
LGTM! Thanks for this it's terrific! Can we add clone and permute too, to tick all the boxes of the issue?

vmoens (approved):
Neat!

vmoens:
> I added support for the permute operation now. I took into account that the APIs for torch.permute(input, **dims**) and torch.Tensor.permute(***dims**) are different (see https://pytorch.org/docs/stable/generated/torch.permute.html and https://pytorch.org/docs/stable/generated/torch.Tensor.permute.html#torch.Tensor.permute). TensorDict shows the same behavior now.
> 
> I had already added the clone functionality to the previous commit. Note that torch.clone(tensor_dict) always takes the default recursive=True (and thus it is not possible to call torch.clone(td, recursive=False)).

Perfect
Sorry I missed the `clone()`. Indeed the default behaviour should be kept, as for `torch.stack`.

PR #120, opened 2022-05-12
TD Lambda advantage function

vmoens:


torchrl/objectives/returns/functional.py
```diff
@@ -88,3 +93,47 @@ def td_advantage_estimate(
     not_done = 1 - done.to(next_state_value.dtype)
     advantage = reward + gamma * not_done * next_state_value - state_value
     return advantage
+
+
+def td_lambda_advantage_estimate(
+    gamma: float,
+    lamda: float,
```
Benjamin-eecs, line 100: lambda, missing the 'b' here

torchrl/objectives/returns/advantages.py
```diff
@@ -20,9 +20,14 @@
 from torchrl.data.tensordict.tensordict import _TensorDict
 from torchrl.envs.utils import step_tensordict
 from torchrl.modules import TDModule
```
Benjamin-eecs, line 22: TDModule(TensorDict) here actually may be confused as temporal difference module, as someone mentioned before I think in the discusstion

torchrl/objectives/returns/functional.py
vmoens: I know but python doesn't like lambda :)

torchrl/objectives/returns/functional.py
```diff
@@ -88,3 +93,47 @@ def td_advantage_estimate(
...
+    reward: torch.Tensor,
+    done: torch.Tensor,
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    """
+    Get generalized advantage estimate of a trajectory
+    Refer to "HIGH-DIMENSIONAL CONTINUOUS CONTROL USING GENERALIZED ADVANTAGE ESTIMATION"
+    https://arxiv.org/pdf/1506.02438.pdf for more context.
+
+    Args:
+        gamma (scalar): exponential mean discount.
```
Benjamin-eecs, line 112: About the gamma and lambda, it can be different in different timesteps and batchs, should we consider that? It is considered in rlax, not common choice though.

vmoens: Yes good point!
Again I was planning to do that in a separate PR as it could be the case for GAE too.

Benjamin-eecs (commented):
lamda return implementation is correct. Leave naming as future todo.

torchrl/objectives/returns/functional.py
Benjamin-eecs: maybe 'lambda_'?

torchrl/objectives/returns/advantages.py
Benjamin-eecs: Yeah, for sure.

torchrl/objectives/returns/functional.py
Benjamin-eecs: Great

torchrl/objectives/returns/functional.py
vmoens: I've raised the issue with pytorch core and domain folks, i'll keep you posted
Seems like many ppl use `lmbda`

Benjamin-eecs: ok, cool.

torchrl/objectives/returns/functional.py
```diff
@@ -88,3 +93,109 @@ def td_advantage_estimate(
...
+
+    shape = val.shape
+    w = w.squeeze(-1).unsqueeze(0).unsqueeze(0)
+    val_pad = val_pad.squeeze(-1).unsqueeze(0).unsqueeze(0)
+    out = torch.conv1d(val_pad, w)
+    out = out.view(shape)
+    return out
+
+
+def vec_td_lambda_advantage_estimate(gamma, lamda, state_value,
```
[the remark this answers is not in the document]
Benjamin-eecs: I think the speedup implementation looks great, maybe extend to gamma and lambda tensors later.

torchrl/objectives/returns/functional.py
```diff
@@ -88,3 +93,109 @@ def td_advantage_estimate(
...
+
+    gammas = torch.ones(T + 1, 1, device=device)
+    gammas[1:] = gamma
+    gammas = torch.cumprod(gammas, -2)
+
+    lambdas = torch.ones(T + 1, 1, device=device)
+    lambdas[1:] = lamda
+    lambdas = torch.cumprod(lambdas, -2)
+
+    first_below_thr = gammas < 1e-7
```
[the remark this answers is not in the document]
Benjamin-eecs: If it does, wondering if it aligns with non-vectorized version

torchrl/objectives/returns/functional.py
shagunsodhani: I did not follow the point about "it can be different in different timesteps and batchs, should we consider that". Could you please share some pointers

torchrl/objectives/returns/functional.py
```diff
@@ -88,3 +93,130 @@ def td_advantage_estimate(
...
+    for i in reversed(range(T)):
+        g = returns[..., i, :] = reward[..., i, :] + gamma * (
+            (1 - lamda) * next_state_value[..., i, :] + lamda * g
+        )
+    advantage = returns - state_value
+    return advantage
+
+
+def _custom_conv1d(val, w):
+    val_pad = torch.cat(
```
shagunsodhani, line 144: could you add a one-line comment about why is this custom conv needed?

PR #130, opened 2022-05-16
Bugfix: discount computation with multi-step

vmoens:


vmoens:
This was a phantom bug, discarding

PR #131, opened 2022-05-16
looks like typo, we should check is_shared when converting to memmap

kovalevfm:


vmoens (approved):
Indeed, I thought I corrected it in #124 but i must have missed it for some reason
Good catch!

Issue #96, opened 2022-04-26
`requires_grad` attribute for meta tensors


vmoens:
MetaTensors is a special TorchRL class that tracks the metadata of tensors stored in the TensorDict.
Its main reason of existence is to avoid the overhead of loading tensors just to read an attribute. We already have a SavedTensordict class, for instance, that represents tensors saved on disk. If we had to load the whole tensordict from disk every time an attribute (shape, etc.) is gathered this could create an unecessary overhead.
We carefully choose which attribute should be part of MetaTensors. Some may not be easy to infer and are therefore not part of the metatensor. 
It may be useful to implement a `requires_grad` attribute in MetaTensor. Its purpose would simply be to tell if a tensor is part of a graph or not.

In the tests, one should make sure that these aspects are accounted for:
- [ ] For LazyStackTensorDict objects, the resulting tensors will require a gradient if a single tensor of the list requires a gradient. The respective MetaTensor should follow the same logic.
- [ ] In theory, MemmapTensors and SavedTensorDicts are not compatible with gradients that are part of computational graphs. This means that when writing a tensor that has requires_grad in a SavedTensorDict, an error should be raised. Same if a regular TensorDict is cast using `tensordict.to(SavedTensorDict)` or `tensordict.memmap_()`.

PR #135, opened 2022-05-16
requires_grad attribute for meta tensors #96

kovalevfm:
[ #96]

### DIFF
--- a/test/test_tensordict.py
+++ b/test/test_tensordict.py
@@ -1458,7 +1458,7 @@ def test_requires_grad(device):
     ]
     sub_td = LazyStackedTensorDict(*tensordicts, stack_dim=0)
     # First stacked tensor has requires_grad == True
-    assert list(sub_td.values_meta())[0].requires_grad == True
+    assert list(sub_td.values_meta())[0].requires_grad is True
     td0 = SavedTensorDict(tensordicts[0])
     with pytest.raises(
         Exception,
--- a/test/test_tensordict.py
+++ b/test/test_tensordict.py
@@ -12,8 +12,8 @@ import pytest
 import torch
 from _utils_internal import get_available_devices
 from torch import multiprocessing as mp
-from torchrl.data import TensorDict, SavedTensorDict
-from torchrl.data.tensordict.tensordict import LazyStackedTensorDict, assert_allclose_td
+from torchrl.data import SavedTensorDict, TensorDict
+from torchrl.data.tensordict.tensordict import assert_allclose_td, LazyStackedTensorDict
 from torchrl.data.tensordict.utils import _getitem_batch_size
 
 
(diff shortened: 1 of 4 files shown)


test/test_tensordict.py
```diff
@@ -1438,6 +1438,37 @@ def test_getitem_batch_size():
...
+                    11, 12, 5, device=device, requires_grad=True if i == 5 else False
+                ),
+                "key2": torch.zeros(
+                    11, 12, 50, device=device, dtype=torch.bool
+                ).bernoulli_(),
+            },
+        )
+        for i in range(10)
+    ]
+    sub_td = LazyStackedTensorDict(*tensordicts, stack_dim=0)
```
vmoens, line 1459: maybe `stacked_td` instead of `sub_td`?
You could try to index the tensordict too by the way:
```python
sub_td1 = td[0]
assert sub_td1.get("key1").requires_grad
assert sub_td1.get_meta("key1").requires_grad

sub_td2 = td[torch.tensor([1, 2])]
assert sub_td2.get("key1").requires_grad
assert sub_td2.get_meta("key1").requires_grad
```

We could also do the same for unsqueeze and/or view?

test/test_tensordict.py
```diff
@@ -1438,6 +1438,37 @@ def test_getitem_batch_size():
         assert expected_shape == resulting_shape, idx
 
 
+@pytest.mark.parametrize("device", get_available_devices())
+def test_requires_grad(device):
```
vmoens, line 1442: This looks great to test that tensordict support gradients when created, but what about modification of existing tensordicts?


Do you think it would be possible to also test this with `TestTensorDicts`? 
You can go through various examples of TensorDicts with 
```python
    @pytest.mark.parametrize(
        "td_name",
        [
            "td",
            "stacked_td",
            "sub_td",
            "idx_td",
            "saved_td",
            "unsqueezed_td",
            "td_reset_bs",
        ],
    )
    def test_blabla(td_name):
        exec()
```

What you would need to do is replace one tensor of the tensordict with a tensor that requires a gradient and check that this is supported. Something like
```python
td.set("a", torch.randn_like(td.get("a")).requires_grad_())
```
You'd need to check that the MetaTensor associated with that tensor is updated too.
```python
assert td.get_meta("a").requires_grad
```

vmoens (changes requested):

test/test_tensordict.py
kovalevfm: I've added test class with different TensorDicts with requires_grad and some operations like view and squeeze.

test/test_tensordict.py
```diff
@@ -1115,6 +1115,159 @@ def test_items_values_keys(self, td_name):
...
+            "idx_td",
+            "saved_td",
+            "unsqueezed_td",
+            "td_reset_bs",
+        ],
+    )
+    def test_set_requires_grad(self, td_name):
+        td = getattr(self, td_name)
+        assert not td._get_meta("a").requires_grad
+        td.set("a", torch.randn_like(td.get("a")).requires_grad_())
```
vmoens, line 1133: we probably want to update that given the new API of `sub_td.set`

test/test_tensordict.py
```diff
@@ -1115,6 +1115,159 @@ def test_items_values_keys(self, td_name):
...
+        assert td._get_meta("a").requires_grad
+
+
+class TestTensorDictsRequiresGrad:
```
vmoens, line 1137: oh wow that's a whole new class!
I'm ok with it, great work

vmoens (approved):
Let's just make sure it runs with the new API but then we're good to go

test/test_tensordict.py
kovalevfm: It helped to fund couple bugs, totally worth it

vmoens: indeed!

vmoens (approved):

PR #148, opened 2022-05-18
Make TDSequence index-able by adding `__getitem__` dunder method.

### LINKED ISSUE #128: Make `TDSequence` support indexing
A `TDSequence` is the equivalent of `nn.Sequence` for TDModules. 
As of now, it does not support indexing. A short method mapping `TDSequence.__getitem__(index)` to `TDSequence.module.__getitem__(index)` should be implemented.

Tests should be written in `test/test_tdmodules.py` to check that it behaves as expected.

karthikrangasai:
Fixes #128 

Add `__getitem__` method to `TDSequence` to make it index-able, equivalent to `torch.nn.Sequential`

### DIFF
--- a/test/test_tdmodules.py
+++ b/test/test_tdmodules.py
@@ -355,6 +355,10 @@ class TestTDSequence:
             )
             tdmodule = TDSequence(tdmodule1, tdmodule2)
 
+        assert hasattr(tdmodule, "__getitem__")
+        assert tdmodule[0] == tdmodule1
+        assert tdmodule[1] == tdmodule2
+
         td = TensorDict({"in": torch.randn(3, 3)}, [3])
         tdmodule(td)
         assert td.shape == torch.Size([3])
@@ -420,6 +424,10 @@ class TestTDSequence:
             )
             tdmodule = TDSequence(tdmodule1, tdmodule2)
 
+        assert hasattr(tdmodule, "__getitem__")
+        assert tdmodule[0] == tdmodule1
+        assert tdmodule[1] == tdmodule2
+
         td = TensorDict({"in": torch.randn(3, 3)}, [3])
         tdmodule(td, params=params)
         assert td.shape == torch.Size([3])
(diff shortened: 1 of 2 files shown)


karthikrangasai:
Hello @vmoens ,

should the `TDSequence` class also implement other sequence container type methods like `__setitem__` and `__delitem__` ?

vmoens:
> Hello @vmoens ,
> 
> 
> 
> should the `TDSequence` class also implement other sequence container type methods like `__setitem__` and `__delitem__` ?

That would be delightful if you want to take care of that!

test/test_tdmodules.py
```diff
@@ -355,6 +355,10 @@ def test_stateful(self, safe, spec_type, probabilistic, lazy):
             )
             tdmodule = TDSequence(tdmodule1, tdmodule2)
 
+        assert hasattr(tdmodule, "__getitem__")
+        assert tdmodule[0] == tdmodule1
```
vmoens, line 359: perhaps `is` instead of `==`? Not sure such behaviour will be robust to any single use case

test/test_tdmodules.py
```diff
@@ -420,6 +424,10 @@ def test_functional(self, safe, spec_type, probabilistic):
...
```
vmoens, line 428: same

test/test_tdmodules.py
```diff
@@ -478,17 +486,17 @@ def test_functional_with_buffer(self, safe, spec_type, probabilistic):
...
-                safe=False,
+                fnet1, None, in_keys=["in"], out_keys=["hidden"], safe=False
             )
             tdmodule2 = tdclass(
                 fnet2, spec, in_keys=["hidden"], out_keys=["out"], safe=safe, **kwargs
             )
             tdmodule = TDSequence(tdmodule1, tdmodule2)
 
+        assert hasattr(tdmodule, "__getitem__")
+        assert tdmodule[0] == tdmodule1
```
vmoens, line 497: same

test/test_tdmodules.py
```diff
@@ -555,6 +563,10 @@ def test_vmap(self, safe, spec_type, probabilistic):
...
```
vmoens, line 567: same :)

vmoens (commented):
Have a look at those minor things but otherwise LGTM!

vmoens (approved):
Good to go! Thanks for contributing!

Issue #126, opened 2022-05-14
`tensordict.__getitem__(string)` with `_TensorDict` subclasses.


vmoens:
`TensorDict` can be indexed in two ways: 
- If a tensor, integer, etc. is provided, then the `__getitem__` method will return a `TensorDict` tensors indexed at the right location.
- If a string is given, then the relevant key will be returned. This is exactly equivalent to `tensordict.get(string)` except that no default value can be provided. 

Currently, the second behaviour is only supported by `TensorDict` (not by other `_TensorDict` subclasses, e.g. not by `LazyStackedTensorDict` etc). This should be implemented.
 
Additionally, a test should be written in `test/test_tensordict.py` in the form of 
```python
class TestTensorDicts:
    ....
    @pytest.mark.parametrize(
        "td_name",
        [
            "td",
            "stacked_td",
            "sub_td",
            "idx_td",
            "saved_td",
            "unsqueezed_td",
            "td_reset_bs",
        ],
    )
    def test_getitem_string(self, td_name):
         ....
```
such that we make sure that all tensordict subclasses are covered. The test should demonstrate that the output of `get` and `__getitem__` are equivalent when a string is provided.

karthikrangasai:
Can I take this up ?

Issue #127, opened 2022-05-14
`tensordict.__setitem__(string)` with `_TensorDict` subclasses


vmoens:
As in #126, `_TensorDict` subclasses should support a `__setitem__(string)` method. This should work as `_TensorDict.set(key, value, inplace=False)` (i.e. `set` default behaviour).

Tests should be cover all tensordict subclasses.

vmoens:
Sure! While you're at it an alias from `tensordict.__delitem__` to `tensordict.del_` would be nice to have 😁

PR #149, opened 2022-05-18
`tensordict.__setitem__(string)`, `tensordict.__getitem__(string)`, and `tensordict.__delitem__(string)` for subclasses

karthikrangasai:
Make `_TensorDict` subclasses accept string as key for `__getitem__`, `__settitem__`, and `__delitem__` methods.

Fixes: #126 #127 

NOTE:
Refactored `test/tensordict.py` file to pytest parametrize the Class directly instead of individual methods.

### DIFF
--- a/test/test_tensordict.py
+++ b/test/test_tensordict.py
@@ -431,6 +431,18 @@ def test_savedtensordict(device):
     assert ss.get("a").device == device
 
 
+@pytest.mark.parametrize(
+    "td_name",
+    [
+        "td",
+        "stacked_td",
+        "sub_td",
+        "idx_td",
+        "saved_td",
+        "unsqueezed_td",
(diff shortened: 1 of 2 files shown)


test/test_tensordict.py
```diff
@@ -842,28 +662,14 @@ def test_cpu_cuda(self, td_name):
...
-            "td",
-            "stacked_td",
-            "sub_td",
-            "idx_td",
-            "saved_td",
-            "unsqueezed_td",
-            "td_reset_bs",
-        ],
-    )
+        if td_name not in ["sub_td", "idx_td", "td_reset_bs"]:
```
vmoens, line 666: clever move :)

vmoens (approved):
Great work thanks a lot for this

PR #159, opened 2022-05-23
Make `NormalParamWrapper` pass other inputs 

vmoens:
`NormalParamWrapper` should leave other input untouched, in case they're needed for some other purpose (e.g. intermediate hidden value used for value computation etc.)


vmoens:
broken tests due to `share_memory_` cuda error in nightly

PR #164, opened 2022-05-24
Removed default n_steps ; refactored n_steps to max_steps in envs/

rl-enjoyer:
[ Work in progress, just creating MR ]

### DIFF
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -453,7 +453,7 @@ class _EnvClass:
                     tensordict = tensordict.to(env_device)
                 tensordict = self.step(tensordict)
                 tensordicts.append(tensordict.clone())
-                if tensordict.get("done").all() or i == max_steps - 1:
+                if tensordict.get("done").any() or i == max_steps - 1:
                     break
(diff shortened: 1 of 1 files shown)


torchrl/envs/common.py
```diff
@@ -447,15 +445,15 @@ def policy(td):
...
-            for i in range(n_steps):
+            for i in range(max_steps):
                 if auto_cast_to_device:
                     tensordict = tensordict.to(policy_device)
                 tensordict = policy(tensordict)
                 if auto_cast_to_device:
                     tensordict = tensordict.to(env_device)
                 tensordict = self.step(tensordict)
                 tensordicts.append(tensordict.clone())
-                if tensordict.get("done").any() or i == n_steps - 1:
```
vmoens, line 458: I would leave it to any (just changed that yesterday) otherwise it'll try to do steps in envs that are done

vmoens (commented):
LGTM!
Can you make the small change all -> any? Then we're done! Good work!!

torchrl/envs/common.py
NolanC33: Sounds good, done.

PR #179, opened 2022-06-01
Add ellipsis indexing for TensorDict

BoboBananas:
Implemented ellipsis indexing by creating function convert_ellipsis_to_idx to convert ellipsis into slice(None), depending on batch size of TensorDict. 
For example, for td = TensorDict({'a': torch.zeros(3, 4, 5), 'b': torch.ones(3, 4, 5, 6)}, batch_size=[3, 4, 5])
tensordict[..., 0] becomes tensordict[(slice(None), slice(None), 0)]
Added tests for __getitem__ and __setitem__ into tests/test_tensordict.py.

### DIFF
--- a/test/test_tensordict.py
+++ b/test/test_tensordict.py
@@ -14,7 +14,7 @@ from _utils_internal import get_available_devices
 from torch import multiprocessing as mp
 from torchrl.data import SavedTensorDict, TensorDict
 from torchrl.data.tensordict.tensordict import assert_allclose_td, LazyStackedTensorDict
-from torchrl.data.tensordict.utils import _getitem_batch_size, convert_ellipsis_to_idx
+from torchrl.data.tensordict.utils import _getitem_batch_size
 
 
 @pytest.mark.parametrize("device", get_available_devices())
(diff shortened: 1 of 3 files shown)


torchrl/data/tensordict/utils.py
```diff
@@ -82,3 +82,52 @@ def _getitem_batch_size(
...
     return torch.Size(bs)
+
+
+def convert_ellipsis_to_idx(idx: tuple, batch_size: list):
+    """
+    Given an index containing an ellipsis or just an ellipsis, converts any ellipsis to slice(None)
+    Example: idx = (..., 0), batch_size = [1,2,3] -> new_index = (slice(None), slice(None), 0)
+
+    Args:
+        idx: Input index
```
vmoens, line 93: perhaps add dtype? `idx (tuple): ...` `batch_size (list): ...`

test/test_tensordict.py
```diff
@@ -764,6 +764,56 @@ def test_set_nontensor(self, td_name):
...
+
+        for i in range(len(actual_idx)):
+            idx = actual_idx[i]
+            td_clone = td.clone()
+            actual_td = td[idx].clone().zero_()
+            td_clone[idx] = actual_td
+
+            for key in td.keys():
+                assert (td[idx].get(key) == 0).all()
+
```
vmoens, line 816: Do you think we could write unit tests for `convert_ellipsis_to_idx`?
Those would include: `...`, `(..., 0, ...)`, `(0, ..., 0)`, `(..., 0)`, `(0, ...)`, `(1:2, ...)` `(0, ..., 0, ...)` and maybe no ellipsis at all `(1, 2)` or anything you could think of :)

vmoens (changes requested):
Great work! Just a couple of things and we're good to go

torchrl/data/tensordict/utils.py
BoboBananas: do you mean add typing to the variables under "Args"?

vmoens: Yes 
The doc should say what type you expect for the inputs.
IMO that is more important than type hinting

torchrl/data/tensordict/utils.py
```diff
@@ -82,3 +82,52 @@ def _getitem_batch_size(
     list_iter_bs = list(iter_bs)
     bs += list_iter_bs
     return torch.Size(bs)
+
+
+def convert_ellipsis_to_idx(idx: tuple, batch_size: list):
```
shagunsodhani, line 87: what about using `tuple` vs `typing.Tuple` ? I think the tests check for Python 3.8 as well so we should be compatible to that

vmoens: Yes we should stick to `typing` whenever possible. For now we don't check typing so those constraints are quite loose -- I'm sure there are a few `x: dict` or untyped functions here and there in the repo 😅

test/test_tensordict.py
BoboBananas: Added units tests

torchrl/data/tensordict/utils.py
BoboBananas: Changed to Tuple and List from typing library

torchrl/data/tensordict/utils.py
BoboBananas: Changed to Tuple and List from typing library

test/test_tensordict.py
```diff
@@ -764,6 +764,81 @@ def test_set_nontensor(self, td_name):
...
+        ]
+
+        for i in range(len(ellipsis_idx)):
+            assert (
+                convert_ellipsis_to_idx(ellipsis_idx[i], batch_size) == expected_idx[i]
+            )
+
+        for i in range(len(error_idx)):
+            with pytest.raises(RuntimeError):
+                _ = convert_ellipsis_to_idx(error_idx[i], batch_size)
```
vmoens, line 790: Let's move it out of the class, no need to repeat it for every td_class

test/test_tensordict.py
```diff
@@ -764,6 +764,81 @@ def test_set_nontensor(self, td_name):
...
+
+    def test_getitem_ellipsis(self, td_name):
+        torch.manual_seed(1)
+        td = TensorDict(
+            {"a": torch.randn(1, 3, 4, 5, 6), "b": torch.randn(1, 3, 4, 5)},
+            batch_size=[1, 3, 4],
+        )
```
vmoens, line 797: The purpose of td_name is to check that the indexing works for every tensordict class. Here we only test TensorDict

test/test_tensordict.py
```diff
@@ -764,6 +764,81 @@ def test_set_nontensor(self, td_name):
...
+
+                assert actual_tensor.shape == expected_tensor.shape
+                torch.testing.assert_allclose(actual_tensor, expected_tensor)
+
+    def test_setitem_ellipsis(self, td_name):
+        torch.manual_seed(1)
+        td = TensorDict(
+            {"a": torch.randn(1, 3, 4, 5, 6), "b": torch.randn(1, 3, 4, 5)},
+            batch_size=[1, 3, 4],
+        )
```
vmoens, line 823: Same here

torchrl/data/tensordict/utils.py
```diff
@@ -82,3 +83,52 @@ def _getitem_batch_size(
...
+
+
+def convert_ellipsis_to_idx(idx: Union[Tuple, Ellipsis], batch_size: List[int]):
+    """
+    Given an index containing an ellipsis or just an ellipsis, converts any ellipsis to slice(None)
+    Example: idx = (..., 0), batch_size = [1,2,3] -> new_index = (slice(None), slice(None), 0)
+
+    Args:
+        idx (Tuple, Ellipsis): Input index
+        batch_size (List): Shape of tensor to be indexed
```
vmoens, line 95: Here you can use `list` and `tuple`, it's not type hinting but documentation

vmoens (changes requested):
You should probably refactor lines 449-456 too to make a tensordict of 4 dimensions
Something like `td = TensorDict({...}, batch_size=[3, 1, 2, 3])`.

You should also change line 482 in that case.

If that breaks things I'll help to fix them! 

@shagunsodhani hope it's not too much work. Otherwise I can take it over, I think we're almost there but here we touch at the internal machinery of TensorDict tests...

test/test_tensordict.py
```diff
@@ -814,6 +841,49 @@ def test_set_nontensor(self, td_name):
...
+        td = getattr(self, td_name)
+        batch_size = len(td.batch_size)
+
+        actual_idx = [..., (..., 0), (0, ...), (0, ..., 0)]
+        expected_idx = [
+            (slice(None),) * batch_size,
+            (slice(None),) * (batch_size - 1) + (0,),
+            (0,) + (slice(None),) * (batch_size - 1),
+            (0,) + (slice(None),) * (batch_size - 2) + (0,),
+        ]
```
vmoens, line 856: Could we use `pytest.mark.parameterize` for that?

```
@pytest.mark.parameterize("actual_index,expected_index", [((..., 0), (:, :, :, 0)), blabla])
def test_getitem_ellipsis(actual_index, expected_index):
     Blabla
```

vmoens (changes requested):

test/test_tensordict.py
```diff
@@ -814,6 +841,49 @@ def test_set_nontensor(self, td_name):
...
+                assert actual_tensor.shape == expected_tensor.shape
+                torch.testing.assert_allclose(actual_tensor, expected_tensor)
+
+    def test_setitem_ellipsis(self, td_name):
+        torch.manual_seed(1)
+        td = getattr(self, td_name)
+
+        actual_idx = [..., (..., 0), (0, ...), (0, ..., 0)]
+
+        for i in range(len(actual_idx)):
```
vmoens, line 878: Same here let's avoid looping inside the test function

vmoens (changes requested):
We're almost there!
It's better to have unit tests that don't contain loops, it'll make the debugging and errors clearer

test/test_tensordict.py
vmoens: Yes but we have a `assert_allclose_td` method that does that loop for you :)

vmoens:
Fantastic! Thanks for this contribution!

Issue #129, opened 2022-05-15
Refactor multi-env / multiprocessing seeding logic to make it uncorrelated across runs


vmoens:
### Problem description

Currently, when multiple environments are created, seeding of each of them is accomplished in a sequential manner. For instance, for [`ParallelEnv`](https://github.com/facebookresearch/rl/blob/4f49f9848aa94874db8415b8996ea95fbe64295a/torchrl/envs/vec_env.py#L556-L566), we iterate through the processes in the following manner:
```python
for i in range(self.num_workers):
    send_seed_to_process(seed)
    seed = seed + 1
```
The same logic is applied to [`DataCollectors`](https://github.com/facebookresearch/rl/blob/4f49f9848aa94874db8415b8996ea95fbe64295a/torchrl/collectors/collectors.py#L813-L842), with the difference that we also receive the seed from each environment back as one data collector can contain multiple `ParallelEnv` (and hence we want to make sure that the seeds don't overlap across collectors).
```python
for i in range(self.num_workers):
    send_seed_to_process(seed)
    final_seed = get_final_seed()  # e.g. seed=0 -> final_seed = 3 if there are 4 envs in the worker
    seed = final_seed + 1
```

In a way, this logic is broken: say we run multiple times the same algorithm with seeds `0`, `1` and `2`, with, say, 3 envs per run. Then the first run will have ens with seeds `[0, 1, 2]`, the second `[1, 2, 3]` and the third `[2, 3, 4]`. Those overlapping seeds are not ideal: experiments will be correlated with each other, at least at the level of the environments.

### Proposed solution
We should create deterministic chains of seeds that are independent (non-overlapping) across successive original seeds.
One way to do this would be to use the `rng` from 

```python
rng = np.random.default_rng(seed)
for i in range(num_envs):
    envs[i].set_seed(seed)
    seed = int.from_bytes(rng.bytes(8), 'big')
    rng = np.random.default_rng(seed)
```

IMO it is not strictly necessary to reset the `rng` at each seed but it may be useful for reproductibility further down the road (i.e. say we track the seeds, we can pick up the chain at a certain point). 

[...]

PR #182, opened 2022-06-02
129_enhance_multiprocessing_seeding_logic

hpark90:
Resolves issue #129

### DIFF
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -28,6 +28,7 @@ from torchrl.envs import ParallelEnv
 from torchrl.envs.libs.gym import _has_gym
 from torchrl.envs.transforms import TransformedEnv, VecNorm
 from torchrl.modules import OrnsteinUhlenbeckProcessWrapper, Actor
+from torchrl import seed_generator
 
 # torch.set_default_dtype(torch.double)
 
@@ -236,8 +237,10 @@ def test_concurrent_collector_seed(num_env, env_name, seed=100):
     ccollector.shutdown()
 
 
-@pytest.mark.parametrize("num_env", [3, 1])
-@pytest.mark.parametrize("env_name", ["conv", "vec"])
+# @pytest.mark.parametrize("num_env", [3, 1])
+# @pytest.mark.parametrize("env_name", ["conv", "vec"])
+@pytest.mark.parametrize("num_env", [3])
+@pytest.mark.parametrize("env_name", ["vec"])
 def test_collector_consistency(num_env, env_name, seed=100):
     if num_env == 1:
 
@@ -247,12 +250,13 @@ def test_collector_consistency(num_env, env_name, seed=100):
             return env
 
     else:
-
+        seed_gen = seed_generator(seed, num_env)
+        seed_args = [{"seed": next(seed_gen)} for i in range(num_env)]
         def env_fn(seed):
             env = ParallelEnv(
                 num_workers=num_env,
                 create_env_fn=make_make_env(env_name),
-                create_env_kwargs=[{"seed": i} for i in range(seed, seed + num_env)],
+                create_env_kwargs=seed_args,
             )
             return env
 
@@ -291,6 +295,7 @@ def test_collector_consistency(num_env, env_name, seed=100):
     if num_env == 1:
         # rollouts collected through DataCollector are padded using pad_sequence, which introduces a first dimension
         rollout1a = rollout1a.unsqueeze(0)
+
     assert (
         rollout1a.batch_size == b1.batch_size
     ), f"got batch_size {rollout1a.batch_size} and {b1.batch_size}"
@@ -420,15 +425,20 @@ def test_collector_vecnorm_envcreator():
     """
     from torchrl.envs import GymEnv
 
+    num_envs = 4
     env_make = EnvCreator(lambda: TransformedEnv(GymEnv("Pendulum-v1"), VecNorm()))
-    env_make = ParallelEnv(4, env_make)
+    env_make = ParallelEnv(num_envs, env_make)
 
     policy = RandomPolicy(env_make.action_spec)
+    num_data_collectors = 2
     c = MultiSyncDataCollector(
-        [env_make, env_make], policy=policy, total_frames=int(1e6)
+        [env_make] * num_data_collectors, policy=policy, total_frames=int(1e6)
     )
-    final_seed = c.set_seed(0)
-    assert final_seed == 7
+
+    init_seed = 0
+    final_seed = c.set_seed(init_seed)
+    *_, last_seed = seed_generator(init_seed, num_data_collectors)
+    assert final_seed == last_seed
 
     c_iter = iter(c)
     next(c_iter)
--- a/test/test_helpers.py
+++ b/test/test_helpers.py
@@ -27,6 +27,7 @@ from torchrl.trainers.helpers.models import (
     make_redq_model,
 )
 
+from torchrl import seed_generator
 
 ## these tests aren't truly unitary but setting up a fake env for the
 # purpose of building a model with args is a lot of unstable scaffoldings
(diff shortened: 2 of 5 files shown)


test/test_collector.py
```diff
@@ -236,8 +237,10 @@ def env_fn():
     ccollector.shutdown()
 
 
-@pytest.mark.parametrize("num_env", [3, 1])
-@pytest.mark.parametrize("env_name", ["conv", "vec"])
+# @pytest.mark.parametrize("num_env", [3, 1])
```
vmoens, line 240: Any specific reason why you dropped those tests?

torchrl/__init__.py
```diff
@@ -88,3 +91,8 @@ def _check_for_faulty_process(processes):
         raise RuntimeError(
             "At least on process failed. Check for more infos in the log."
         )
+def seed_generator(seed, repeat):
+    MAX_SEED_VAL =  2**32 - 1
```
vmoens, line 95: if it's a global variable you can move it out of the function, otherwise I would not use caps

torchrl/__init__.py
```diff
@@ -88,3 +91,8 @@ def _check_for_faulty_process(processes):
...
+    rng = np.random.default_rng(seed)
+    for i in range(repeat):
+        yield int.from_bytes(rng.bytes(8), "big") % MAX_SEED_VAL
```
vmoens, line 98: There should be a new line (linter)
Refer to CONTRIBUTING.md on the main branch: you can use `pre-commit` which will run the lint for you and make it all work!

vmoens (approved):
Just a couple of comments but looks like we're on the right track!

vmoens (changes requested):
Sorry meant to ask for changes :)

test/test_collector.py
hpark90: This was while I was debugging. Sorry about confusion.

hpark90:
@vmoens 

I made some logic change and most of unit tests pass.
In test_collector unit test, the following two test sets fail.

> FAILED test/test_collector.py::test_collector_consistency[conv-3] - AssertionError: key done does not match, got mse = 0.5774   key action does not match, got mse = 0.0000
> FAILED test/test_collector.py::test_collector_consistency[vec-3] - AssertionError: key done does not match, got mse = 0.5774    key action does not match, got mse = 0.0000
> FAILED test/test_collector.py::test_traj_len_consistency[conv-SyncDataCollector-3] - RuntimeError: The size of tensor a (20) must match the size of tensor b (2) at non-singleton dimension 1
> FAILED test/test_collector.py::test_traj_len_consistency[conv-aSyncDataCollector-3] - RuntimeError: At least on process failed. Check for more infos in the log.
> FAILED test/test_collector.py::test_traj_len_consistency[vec-SyncDataCollector-3] - RuntimeError: The size of tensor a (20) must match the size of tensor b (2) at non-singleton dimension 1
> FAILED test/test_collector.py::test_traj_len_consistency[vec-aSyncDataCollector-3] - RuntimeError: At least on process failed. Check for more infos in the log.

I think it fails because of this line in mock class. 
https://github.com/facebookresearch/rl/blob/fc7876666fa0c6733a4b272aa7cfc5b997bc2e04/test/mocking_classes.py#L72
For instance, I change this line to `self.count = 100` test passes.
Can you please explain what this line does and suggest best way to update this line?

Thank you,
Hojin

test/test_helpers.py
```diff
@@ -432,6 +432,34 @@ def test_redq_make(device, from_pixels, gsde, exploration):
     del proof_environment
 
 
+def test_seed_generator():
```
vmoens, line 435: great test function

Perhaps let's move the loop out of it, and do something like

```
@pytest.mark.parameterize("initial_seed", range(5))
def test_seed_generator(initial_seed):
    ...
```

test/test_helpers.py
```diff
@@ -432,6 +432,34 @@ def test_redq_make(device, from_pixels, gsde, exploration):
...
+            seeds.append(seed)
+        return seeds
+
+    # Check unique seed generation
+    for initial_seed in range(num_tests):
+        seeds = generate_seeds(initial_seed, num_seeds)
+        assert len(seeds) == num_seeds
+        assert len(seeds) == len(set(seeds))
+        if prev_seeds:
+            assert set(prev_seeds) != set(seeds)
```
vmoens, line 453: great! but not super readable
I guess what we want is to test that the union is empty

what about
```
assert len(set(prev_seeds).union(seeds)) == 0
```

torchrl/collectors/collectors.py
```diff
@@ -841,12 +841,9 @@ def set_seed(self, seed: int) -> int:
         _check_for_faulty_process(self.procs)
         for idx in range(self.num_workers):
             self.pipes[idx].send((seed, "seed"))
-            new_seed, msg = self.pipes[idx].recv()
+            seed, msg = self.pipes[idx].recv()
             if msg != "seeded":
                 raise RuntimeError(f"Expected msg='seeded', got {msg}")
-            seed = new_seed
```
vmoens, line 847: here we increment the seed by 1, such that all sub-collectors have a different initial seed

Do we still do that?
Don't we need to generate a new seed based on the previous one?

torchrl/envs/vec_env.py
```diff
@@ -714,10 +714,10 @@ def _shutdown_workers(self) -> None:
     def set_seed(self, seed: int) -> int:
         self._seeds = []
         for i, channel in enumerate(self.parent_channels):
+            seed = seed_generator(seed)
             channel.send(("seed", seed))
             self._seeds.append(seed)
-            if i < self.num_workers - 1:
```
vmoens, line 719: same here

torchrl/collectors/collectors.py
hpark90: This Design
1. Collection sends a root seed to ParallelEnv.
2. ParallelEnv generates seed sequence excl. the root seed and seed to Env.
3. ParallelEnv returns the last used seed.
4. Collection receives a last used seed from ParallelEnv. This value is now a new root seed and feeded into the next ParallelEnv.

Pros
1. Collection does not know about the seeding algorithm. 
Cons
1. The root seed is not used. But do we need to know about it? We can always restore the exact sequence from default_rng function.

Alternative(Old) Design
1. Collection sends a root seed to ParallelEnv.
2. ParallelEnv generates seed sequence incl. the root seed and seed to Env.
3. ParallelEnv returns the last used seed.
4. Collection receives a last used seed from ParallelEnv.
5. Collection gets the next sequence from the last used seed it received from ParallelEnv. This value is now a new root seed and feeded into the next ParallelEnv.

Pros
1. Root seed is used. 
Cons
1. Collection does not need to know about the seeding algorithm.

I liked the first approach, but please let me know if you prefer the second approach!

torchrl/envs/vec_env.py
```diff
@@ -514,9 +515,8 @@ def _shutdown_workers(self) -> None:
     @_check_start
     def set_seed(self, seed: int) -> int:
         for i, env in enumerate(self._envs):
+            seed = seed_generator(seed)
```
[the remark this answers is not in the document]
hpark90: Please see the above and let me know!

torchrl/envs/vec_env.py
hpark90: Please see the above and let me know!

test/test_helpers.py
```diff
@@ -432,6 +431,29 @@ def test_redq_make(device, from_pixels, gsde, exploration):
...
+    if initial_seed == 0:
+        with pytest.raises(ValueError) as e:
+            seeds0 = generate_seeds(initial_seed - 1, num_seeds)
+        return
+    else:
+        seeds0 = generate_seeds(initial_seed - 1, num_seeds)
+    seeds1 = generate_seeds(initial_seed, num_seeds)
+    assert len(seeds1) == num_seeds
+    assert len(seeds1) == len(set(seeds1))
+    assert not set(seeds0).intersection(set(seeds1))
```
vmoens, line 449: I find it clearer to assert that the len is 0 if that's ok

torchrl/envs/vec_env.py
```diff
@@ -514,10 +514,9 @@ def _shutdown_workers(self) -> None:
     @_check_start
     def set_seed(self, seed: int) -> int:
         for i, env in enumerate(self._envs):
-            env.set_seed(seed)
-            if i < self.num_workers - 1:
-                seed = seed + 1
-        return seed
+            new_seed = env.set_seed(seed)
+            seed = new_seed
+        return new_seed
```
vmoens, line 519: Let's return seed instead (see below)

vmoens (approved):
Amazing work!
Let's fix the 2 things I pointed and we're good to go!!

PR #198, opened 2022-06-15
Integrating hydra with SAC

BoboBananas:
Integrated hydra with SAC using structured configs.
Checked by validating hydra version of SAC gives similar results as the original version.


vmoens (approved):
LGTM thanks

PR #199, opened 2022-06-15
Integrating hydra with DDPG

BoboBananas:
Integrated hydra with DDPG using structured configs.
Checked by validating hydra version of DDPG gives similar results as the original version.


vmoens (approved):
LGTM thanks!

PR #201, opened 2022-06-15
Integrating hydra with DQN

BoboBananas:
Integrated hydra with DQN using structured configs.


examples/ddpg/ddpg.py
```diff
@@ -96,50 +84,50 @@ def main(args):
...
+        cfg=cfg,
         device=device,
     )
-    loss_module, target_net_updater = make_ddpg_loss(model, args)
+    loss_module, target_net_updater = make_ddpg_loss(model, cfg)
 
     actor_model_explore = model[0]
-    if args.ou_exploration:
-        if args.gSDE:
+    if cfg.ou_exploration:
```
vmoens, line 117: can we delete `ou_exploration` and `gSDE` from DQN? they make no sense here. Ideally they shouldn't even be in the config

examples/ppo/ppo.py
```diff
@@ -10,7 +10,6 @@
 import hydra
```
vmoens, line 10: I'm not sure where do those changes in PPO come from in this PR

torchrl/trainers/helpers/models.py
```diff
@@ -1351,3 +1351,5 @@ class DiscreteModelConfig:
     # whether a distributional loss should be used.
     atoms: int = 51
     # number of atoms used for the distributional loss
+    gSDE: bool = False
```
vmoens, line 1354: Yeah let's not do that :) No gSDE for discrete actions

vmoens (commented):
Let's see what we can do about gSDE

torchrl/trainers/helpers/models.py
BoboBananas: https://github.com/facebookresearch/rl/pull/201/commits/03b8c78b7c39cfd40d4d927d08339ce6fe70fc2c

BoboBananas: I removed epsilon greedy wrapper which removed the error with action_value being None

examples/dqn/dqn.py
```diff
@@ -103,17 +101,8 @@ def main(cfg: "DictConfig"):
     )
 
     loss_module, target_net_updater = make_dqn_loss(model, cfg)
-    model_explore = EGreedyWrapper(model, annealing_num_steps=cfg.annealing_frames).to(
```
vmoens: I think we still want that, that's the typical exploration technique for dqn

examples/dqn/dqn.py
```diff
@@ -124,7 +113,7 @@ def main(cfg: "DictConfig"):
 
     collector = make_collector_offpolicy(
         make_env=create_env_fn,
-        actor_model_explore=model_explore,
```
vmoens, line 129: let's keep model_explore

vmoens (commented):
EGreedyWrapper should still be used

PR #204, opened 2022-06-16
Gym seeding logic and robustness

vmoens:
Makes seeding robust to new gym API where seeding has to be achieved through `.reset(seed=seed)`, but registered envs may not support that.

What we want to avoid is calling the part under `except` too often, as it has a significant overhead. Hence we use an attribute that tracks if `reset` can be used for seeding or not, and not we always use `seed`.

cc @yidingjiang

### DIFF
--- a/torchrl/envs/libs/gym.py
+++ b/torchrl/envs/libs/gym.py
@@ -191,7 +191,7 @@ class GymWrapper(GymLikeEnv):
                     self._seed_calls_reset = False
         if self._seed_calls_reset and not skip:
             self.reset(seed=seed)
-        else:
+        elif not self._seed_calls_reset:
             self._env.seed(seed=seed)
         return seed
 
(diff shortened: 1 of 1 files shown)


yidingjiang:
lgtm

torchrl/envs/libs/gym.py
```diff
@@ -172,10 +173,26 @@ def lib(self) -> ModuleType:
...
+                except TypeError as err:
+                    warnings.warn(
+                        f"reset with seed kwarg returned an exception: {err}.\n"
+                        f"Calling env.seed from now on."
+                    )
+                    self._seed_calls_reset = False
+        if self._seed_calls_reset and not skip:
             self.reset(seed=seed)
+        else:
+            self._env.seed(seed=seed)
```
yidingjiang, line 195: If `self._seed_calls_reset`  is `False`, wouldn't this line still error out as the `_env` does not accept a seed? i.e., `TypeError: reset() got an unexpected keyword argument 'seed'`

PR #208, opened 2022-06-20
Memmap Efficiency

vmoens:


vmoens:
Putting on hold since many changes have been done since then

PR #209, opened 2022-06-21
T123527109: TorchRL: Make __del__ call shutdown and close in DataCollectors and Envs

OdedKrams:
Rearrange the code and remove unneeded error handling so shutdown and close functions will be called from the __del__ function.
From the unit tests it seems that:
* Envs don't need to be close nor deleted before getting out of scope. The __del__ and the GC are doing the job well.
* In DataCollectors it seems there is some race condition in the GC so that the internal Env may be deleted before the DataCollector. So the DataCollector must be deleted before getting out of scope. Its __del__ function delete and remove all objects in the proper way. Without calling `del collector` in the end to the unit test, the test hangs in the end.

### DIFF
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -107,7 +107,7 @@ def test_concurrent_collector_consistency(num_env, env_name, seed=40):
             break
     with pytest.raises(AssertionError):
         assert_allclose_td(b1, b2)
-    collector.shutdown()
+    del collector
 
     ccollector = aSyncDataCollector(
         create_env_fn=env_fn,
@@ -131,7 +131,7 @@ def test_concurrent_collector_consistency(num_env, env_name, seed=40):
     assert_allclose_td(b1c, b1)
     assert_allclose_td(b2c, b2)
 
-    ccollector.shutdown()
+    del ccollector
 
 
 @pytest.mark.parametrize("num_env", [1, 3])
@@ -170,7 +170,7 @@ def test_collector_batch_size(num_env, env_name, seed=100):
         assert b.numel() == -(-frames_per_batch // num_env) * num_env
         if i == 5:
             break
-    ccollector.shutdown()
+    del ccollector
 
     ccollector = MultiSyncDataCollector(
         create_env_fn=[env_fn for _ in range(num_workers)],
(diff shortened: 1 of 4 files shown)


vmoens:
> In DataCollectors it seems there is some race condition in the GC so that the internal Env may be deleted before the DataCollector. So the DataCollector must be deleted before getting out of scope. Its del function delete and remove all objects in the proper way. Without calling del collector in the end to the unit test, the test hangs in the end.

I noticed that too, thanks for identifying the issue! 
So what's your take on this?
It might be the reason why we decided to explicitly ask the user to close data collectors and envs, otherwise there may be a silent problem with their script (ie the script never terminates). This may have really bad consequences (e.g. a training script on a cluster that borrows nodes for ages).

I see basically 2 options: either we make sure that this phenomenon never happens or we revert and ask users to close the collector / env when it's not used, otherwise an error is thrown.

test/test_collector.py
```diff
@@ -134,6 +134,41 @@ def env_fn(seed):
...
+        create_env_fn=env_fn,
+        create_env_kwargs={"seed": seed},
+        policy=policy,
+        frames_per_batch=20,
+        max_frames_per_traj=2000,
+        total_frames=20000,
+        pin_memory=False,
+    )
+    for i, d in enumerate(ccollector):
+        if i == 0:
```
vmoens, line 159: we can just break when i == 2 or something

test/test_collector.py
```diff
@@ -107,7 +107,7 @@ def env_fn(seed):
             break
     with pytest.raises(AssertionError):
         assert_allclose_td(b1, b2)
-    collector.shutdown()
```
vmoens, line 110: can we keep these, as we test it somewhere else?
Like this if for some reason the new feature breaks only that test will fail

vmoens (commented):
Let's think about what would be the behaviour of the test without the new feature.

test/test_collector.py
OdedKrams: Sounds good

test/test_env.py
```diff
@@ -129,7 +129,6 @@ def test_env_seed(env_name, frame_skip, seed=0):
         assert_allclose_td(td0a, td0c.select(*td0a.keys()))
     with pytest.raises(AssertionError):
         assert_allclose_td(td1a, td1c)
-    env.close()
```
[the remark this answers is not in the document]
OdedKrams: 👍

PR #230, opened 2022-06-28
Refactor CompositeSpec tests

khmigor:
https://github.com/facebookresearch/rl/issues/220

### DIFF
--- a/test/test_tensor_spec.py
+++ b/test/test_tensor_spec.py
@@ -6,6 +6,7 @@
 import numpy as np
 import pytest
 import torch
+from _utils_internal import get_available_devices
 from torchrl.data.tensor_specs import (
     NdUnboundedContinuousTensorSpec,
     NdBoundedTensorSpec,
@@ -16,7 +17,8 @@ from torchrl.data.tensor_specs import (
     UnboundedContinuousTensorSpec,
     OneHotDiscreteTensorSpec,
 )
-
+from torchrl.data.tensordict.tensordict import TensorDict
+from unittest.mock import Mock
 
 @pytest.mark.parametrize("dtype", [torch.float32, torch.float16, torch.float64, None])
 def test_bounded(dtype):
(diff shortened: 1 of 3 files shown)


vmoens:
Thanks for this!

Let's have a look at the failed tests.
For the linting issue, please refer to CONTRIBUTING.md on the home page of the repo

khmigor:
Yes, looking at this right now. Unfortunately, I wasn't able to make suggested pre-commit linter work, but I ran some linters separately, and that should fix most issues.
As for the tests, I have lots of failed tests (locally) even when running on the main branch, which seems weird. Probably has something to do with mac architecture, so I'm reinstalling the dependencies just to double check.

test/test_tensor_spec.py
```diff
@@ -226,37 +230,150 @@ def test_mult_onehot(shape, ns):
...
+    def test_setitem_forbidden_keys(self, is_complete, device, dtype):
+        ts = self._composite_spec(is_complete, device, dtype)
+        for key in {"shape", "device", "dtype", "space"}:
+            with pytest.raises(AttributeError, match="cannot be set"):
+                ts[key] = 42
+
+    def test_setitem_matches_device(self, is_complete, device, dtype):
+        ts = self._composite_spec(is_complete, device, dtype)
+
+        bad_value, good_value = Mock(), Mock()
```
vmoens, line 273: Great idea of using Mock though I see two caveats:
At one point we might check that specs are or type TensorSpec, does mock works in that case? (Ie does isinstance(mock, type) always true for all types?)
Second we rely on pytest for our tests and I'm not sure if we want to mix unittest and pytest.

test/test_tensor_spec.py
```diff
@@ -226,37 +230,150 @@ def test_mult_onehot(shape, ns):
...
+            ts["bad"] = bad_value
+
+        ts["good"] = good_value
+        assert ts["good"] == good_value
+
+    def test_del(self, is_complete, device, dtype):
+        ts = self._composite_spec(is_complete, device, dtype)
+        assert "obs" in ts.keys()
+        assert "act" in ts.keys()
+        ts.del_("obs")
```
vmoens, line 289: Can we try with 
`del ts["obs"]` 
Too?

test/test_tensor_spec.py
```diff
@@ -226,37 +230,150 @@ def test_mult_onehot(shape, ns):
...
+        rand_td = ts.rand(shape)
+        assert rand_td.shape == torch.Size(shape)
+        assert rand_td.get("obs").shape == torch.Size([*shape, 3, 32, 32])
+        assert rand_td.get("obs").dtype == dtype
+        if is_complete:
+            assert rand_td.get("act").shape == torch.Size([*shape, 7])
+            assert rand_td.get("act").dtype == dtype
+
+    def test_repr(self, is_complete, device, dtype):
+        ts = self._composite_spec(is_complete, device, dtype)
```
vmoens, line 346: Brilliant! Repr are almost never tested in the lib, they should definitely be!

vmoens (commented):
Almost good to go!
Can you add a __delitem__ method that points to del_, I don't think it exists yet

vmoens (approved):
Staggering! Thanks a lot

PR #231, opened 2022-06-28
Implement unittests for MemmapTensor.

### LINKED ISSUE #221: Test `MemmapTensor` creation with shape, dtype and device
`MemmapTensor`s can be created via `MemmapTensor(tensor)`, `MemmapTensor(memmap_tensor)`, `MemmapTensor(array)` and `MemmapTensor(*size, device=device, dtype=dtype)`.
We don't explicitely test those four methods of creating a memmap tensor.

The test should include:
- [ ] tests regarding device: is the device the one expected (using `get_available_devices()` and `pytest.mark.parameterize`). Is the device accurate after calling `to(device)`?
- [ ] similar tests for dtype
- [ ] testing that inputting a zero-tensor / zero-array / zero-memmap to `MemmapTensor` results in the same tensor (ie has the same value) than creating a `MemmapTensor` via its shape

fredfung007:
- Added a test if the created MemmapTensor is on the same device as the input tensor. 
- Added a test if device is correct when .to(device) is called.
- Added a test if the device arg for MemmapTensor init is respected.
- Added a test if all entries are zeros when MemmapTensor is created with size.
- Added a test if MemmapTensor can be created with a given data type and shape.

This PR closes #221.

### DIFF
--- a/test/test_memmap.py
+++ b/test/test_memmap.py
@@ -160,6 +160,13 @@ def test_memmap_same_device_as_tensor(device):
     m = MemmapTensor(t)
     assert m.device == torch.device(device)
     for other_device in get_available_devices():
+        if other_device != device:
+            with pytest.raises(
+                RuntimeError,
+                match="Expected all tensors to be on the same device, "
+                + "but found at least two devices .*",
+            ):
+                assert torch.all(m + torch.ones([3, 4], device=other_device) == 1)
         m.to(other_device)
         assert m.device == torch.device(other_device)
 
@@ -171,17 +178,24 @@ def test_memmap_create_on_same_device(device):
     assert m.device == torch.device(device)
 
 
+@pytest.mark.parametrize("device", get_available_devices())
 @pytest.mark.parametrize(
     "value", [torch.zeros([3, 4]), MemmapTensor(torch.zeros([3, 4]))]
 )
-def test_memmap_zero_value(value):
+def test_memmap_zero_value(device, value):
     """
     Test if all entries are zeros when MemmapTensor is created with size.
     """
-    m = MemmapTensor(value)
-    expected_memmap_tensor = MemmapTensor([3, 4])
-    assert m.shape == (3, 4)
-    assert torch.all(m == expected_memmap_tensor)
+    value.to(device)
+    expected_memmap_tensor = MemmapTensor(value)
+    m1 = MemmapTensor([3, 4])
+    assert m1.shape == (3, 4)
+    assert torch.all(m1 == expected_memmap_tensor)
+    assert torch.all(m1 + torch.ones([3, 4]) == 1)
+    m2 = MemmapTensor(3, 4)
+    assert m2.shape == (3, 4)
+    assert torch.all(m2 == expected_memmap_tensor)
+    assert torch.all(m2 + torch.ones([3, 4]) == 1)
 
 
 if __name__ == "__main__":
--- a/test/test_memmap.py
+++ b/test/test_memmap.py
@@ -10,9 +10,8 @@ import tempfile
 import numpy as np
 import pytest
 import torch
-from torchrl.data.tensordict.memmap import MemmapTensor
-
 from _utils_internal import get_available_devices
+from torchrl.data.tensordict.memmap import MemmapTensor
 
 
 def test_memmap_type():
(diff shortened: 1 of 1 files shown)


test/test_memmap.py
```diff
@@ -137,9 +150,38 @@ def test_memmap_clone():
...
+def test_memmap_create_on_same_device(device):
+    """Test if the device arg for MemmapTensor init is respected."""
+    m = MemmapTensor([3, 4], device=device)
+    assert m.device == torch.device(device)
+
+
+@pytest.mark.parametrize(
+    "value", [torch.zeros([3, 4]), MemmapTensor(torch.zeros([3, 4]))]
+)
+def test_memmap_zero_value(value):
```
vmoens, line 177: Can we run this over all available devices?

test/test_memmap.py
```diff
@@ -137,9 +150,38 @@ def test_memmap_clone():
...
+    """
+    Test if all entries are zeros when MemmapTensor is created with size.
+    """
+    m = MemmapTensor(value)
+    expected_memmap_tensor = MemmapTensor([3, 4])
```
vmoens, line 182: Can we test `MemmapTensor([3, 4])` and `MemmapTensor(3, 4)`?

test/test_memmap.py
```diff
@@ -137,9 +150,38 @@ def test_memmap_clone():
...
+    for other_device in get_available_devices():
+        m.to(other_device)
+        assert m.device == torch.device(other_device)
+
+
+@pytest.mark.parametrize("device", get_available_devices())
+def test_memmap_create_on_same_device(device):
+    """Test if the device arg for MemmapTensor init is respected."""
+    m = MemmapTensor([3, 4], device=device)
+    assert m.device == torch.device(device)
```
vmoens, line 171: Can we test a simple operation, like `assert (m + torch.ones(3, 4, device=device) == 1).all()`?

test/test_memmap.py
```diff
@@ -137,9 +150,38 @@ def test_memmap_clone():
...
+    """
+    Created MemmapTensor should be on the same device as the input tensor.
+    Check if device is correct when .to(device) is called.
+    """
+    t = torch.tensor([1], device=device)
+    m = MemmapTensor(t)
+    assert m.device == torch.device(device)
+    for other_device in get_available_devices():
+        m.to(other_device)
+        assert m.device == torch.device(other_device)
```
vmoens, line 164: Can we test a simple operation, like `assert (m + torch.ones(3, 4, device= other_device) == 1).all()`?
It should return an error if something is not on the same device

vmoens (commented):
Fantastic work, thanks so much for this! Almost good to go!

test/test_memmap.py
fredfung007: Added test cases.

test/test_memmap.py
```diff
@@ -137,9 +150,52 @@ def test_memmap_clone():
...
+@pytest.mark.parametrize(
+    "value", [torch.zeros([3, 4]), MemmapTensor(torch.zeros([3, 4]))]
+)
+def test_memmap_zero_value(device, value):
+    """
+    Test if all entries are zeros when MemmapTensor is created with size.
+    """
+    value.to(device)
+    expected_memmap_tensor = MemmapTensor(value)
+    m1 = MemmapTensor([3, 4])
```
vmoens, line 191: I guss this guy should be created on device too
```python
    m1 = MemmapTensor([3, 4], device=device)
```

test/test_memmap.py
```diff
@@ -137,9 +150,52 @@ def test_memmap_clone():
...
+    assert m1.shape == (3, 4)
+    assert torch.all(m1 == expected_memmap_tensor)
+    assert torch.all(m1 + torch.ones([3, 4], device=device) == 1)
+    m2 = MemmapTensor(3, 4)
```
vmoens, line 195: same here
by the way, we could perhaps nest those two tests using parameters

```python
@pytest.mark.parametrize("shape", [ [[3, 4]], [3, 4] ])
def test_foo(shape, device):
    m1 = MemmapTensor(*shape, device=device)  # first cast is a list, second case is a sequence of integers
```

fredfung007: Thanks @vmoens!
I have made corresponding changes. I was setting up the WSL environment locally on a windows machine to test cuda/cpu. Sorry for delayed responses.

vmoens (approved):
Splendid! Thanks so much for this!

PR #234, opened 2022-06-29
[Feature] custom info_dict reader methods

vmoens:
Proposes a `set_info_dict_reader` method for gym-like environments that'll read the info dictionary and write down the values on a `TensorDict` in a way defined by the user.
This will allow users to use `GymWrapper` and `GymEnv` without effort in cases where the info dictionary is weird and unpredictable.

cc @BoboBananas

### DIFF
--- a/torchrl/envs/gym_like.py
+++ b/torchrl/envs/gym_like.py
@@ -10,9 +10,43 @@ from torchrl.data.tensordict.tensordict import _TensorDict
 from torchrl.envs.common import _EnvWrapper
 from torchrl.envs.utils import step_tensordict
 
+__all__ = ["GymLikeEnv", "default_info_dict_reader"]
+
+
+class default_info_dict_reader:
+    """
+    Default info-key reader.
+
+    In cases where keys can be directly written to a tensordict (mostly if they abide to the
+    tensordict shape), one simply needs to indicate the keys to be registered during
+    instantiation.
+
+    Examples:
+        >>> from torchrl.envs import GymWrapper, default_info_dict_reader
+        >>> reader = default_info_dict_reader(["my_info_key"])
+        >>> # assuming "some_env-v0" returns a dict with a key "my_info_key"
+        >>> env = GymWrapper(gym.make("some_env-v0"))
+        >>> env.set_info_dict_reader(info_dict_reader=reader)
+        >>> tensordict = env.reset()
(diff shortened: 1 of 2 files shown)


torchrl/envs/gym_like.py
```diff
@@ -10,9 +10,43 @@
...
+        >>> env = GymWrapper(gym.make("some_env-v0"))
+        >>> env.set_info_dict_reader(info_dict_reader=reader)
+        >>> tensordict = env.reset()
+        >>> tensordict = env.rand_step(tensordict)
+        >>> assert "my_info_key" in tensordict.keys()
+
+    """
+
+    def __init__(self, keys=None):
+        if keys is None:
```
shagunsodhani, line 37: do we want to ensure that keys like 'observation', 'next_observation', 'action', 'reward', 'done' are not passed so that they arent read from info_dict

torchrl/envs/gym_like.py
```diff
@@ -100,6 +132,42 @@ def _output_transform(self, step_outputs_tuple: Tuple) -> Tuple:
...
+            >>> tensordict = env.rand_step(tensordict)
+            >>> assert "my_info_key" in tensordict.keys()
+
+        """
+        self.info_dict_reader = info_dict_reader
+        return self
+
+    @property
+    def info_dict_reader(self):
+        if "_info_dict_reader" not in self.__dir__():
```
shagunsodhani, line 163: I assume that this is checking if `_info_dict_reader` attribute exists or not. I am not sure if checking the `self.__dir__()` is the recommended method. I have seen use of `hasattr` method lot more. So just flagging this, in case it turns out to be relevant.

shagunsodhani (approved):
lgtm - left some minor comments

torchrl/envs/gym_like.py
shagunsodhani: Interesting. I didnt realise hashattr would do that but it makes sense!

vmoens: Yeah it's not a clever function
It just runs getattr and tells you ok if it did not return an error...

torchrl/envs/gym_like.py
shagunsodhani: Makes sense - good to go from my side!

PR #236, opened 2022-06-30
[Feature] TypeCheck refactoring

vmoens:


torchrl/data/tensor_specs.py
```diff
@@ -1118,7 +1122,7 @@ def __repr__(self) -> str:
     def type_check(
         self,
         value: Union[torch.Tensor, TensorDictBase],
-        selected_keys: Union[str, Optional[Sequence[str]]] = None,
+        selected_keys: Optional[Union[str, Optional[Sequence[str]]]] = None,
```
shagunsodhani, line 1125: Maybe `Optional[Union[str, Optional[Sequence[str]]]]` can be shortened to `Optional[Union[str, Sequence[str]]]`

PR #240, opened 2022-06-30
Differentiable PPOLoss for IRL

vmoens:
Usage:

```
>>> floss_fn, params, buffers = functorch.make_functional_with_buffers(loss_fn)
>>> loss = floss_fn(params, buffers, td)
```

See [tests](https://github.com/facebookresearch/rl/blob/e18a0d2192668b83015d0224a48e3ffa49e0cc50/test/test_cost.py#L1400) for more context.

( I made _LossModule public -- this can be moved to another PR )


torchrl/objectives/costs/ppo.py
```diff
@@ -81,6 +81,17 @@ def __init__(
...
+    def make_functional_with_buffers(self):
+        self.actor, (
+            actor_params,
+            actor_buffers,
+        ) = self.actor.make_functional_with_buffers()
+        self.critic, (
+            critic_params,
+            critic_buffers,
+        ) = self.critic.make_functional_with_buffers()
+        return (actor_params, actor_buffers, critic_params, critic_buffers)
```
shagunsodhani, line 93: maybe we can return a class here - that wraps over this tuple (though I am not sure how much extra value that adds).

vmoens: see updated PR :)

shagunsodhani (approved):
lgtm!

PR #245, opened 2022-07-05
Refactor the `torch.stack` with destination

khmigor:
**Not** implemented:
1. `_stack_onto_` for `_CustomOpTensorDict`
2. `_stack_onto_at_` for `SubTensorDict`
3. Tests for stacking on subclasses with: 
* `memmap_td` (such a test fails - probably it requires changes in `stack` for `tensordict/memmap.py` ?), 
* `unsqueezed_td` (because of 1. above).
4. More specific tests for `_stack_onto_at_`.


torchrl/data/tensordict/tensordict.py
```diff
@@ -2097,11 +2155,10 @@ def stack(
                 )
 
         for key in keys:
-            out.set(
-                key,
-                torch.stack([td.get(key) for td in list_of_tensordicts], dim),
-                inplace=True,
+            out._stack_onto_(
```
vmoens, line 2158: we have errors because sometimes the key is missing from the destination.
IMO the expected outcome would be to find those keys in the `out` tensordict.
It should look like this:

```
        out_keys = set(out.keys())
        if strict:
            in_keys = set(keys)
            if len(out_keys - in_keys) > 0:
                raise RuntimeError(
                    "The output tensordict has keys that are missing in the "
                    "tensordict that has to be written: {out_keys - in_keys}. "
                    "As per the call to `stack(..., strict=True)`, this "
                    "is not permitted."
                )
            elif len(in_keys - out_keys) > 0:
                raise RuntimeError(
                    "The resulting tensordict has keys that are missing in "
                    f"its destination: {in_keys - out_keys}. As per the call "
                    "to `stack(..., strict=True)`, this is not permitted."
                )

        for key in keys:
            if key in out_keys:
                out._stack_onto_(
                    key, [_tensordict[key] for _tensordict in list_of_tensordicts], dim
                )
            else:
               out.set(key, torch.stack([_tensordict[key] for _tensordict in list_of_tensordicts], dim))
```

vmoens (approved):
LGTM thanks!

PR #251, opened 2022-07-07
Fix a bug when segment_tree size is exactly 2^N

xiaomengy:


vmoens (approved):
LGTM thanks for spotting this!

PR #255, opened 2022-07-07
[Doc] Added TensorDict tutorial

nicolas-dufour:
Tutorial on Tensor Dict and TensorDictModule.
Contains simple examples of TensorDict and TensorDictModule operations.
Also contains implementation of a Transformer model using Tensor Dict and TensorDictModule to showcase how this modules work


vmoens:
Few comments:

1. On a high level, I think it would make sense to split the file in tuto for TensorDict and tuto for TensorDictModule
2. The structure of the tuto for TensorDict should be like this IMO:
- What's a TensorDict, what's the 10000 feet view of it, what it can do (broadly speaking): e.g. a TensorDict is a **horse** that can carry **Teletubbies** that must be **high on eucalyptus**. It can crawl and jump but not walk. It is aimed at being nice to watch.
- Some exciting stuff about it first: don't start with "it shares a lot with dict" otherwise people may be bored already. Find something you feel excited about (that is not too long) and that people will be like "oh gosh I wish I knew about it when i was doing my latest project!"
- We can ditch the initial part about PPO and DQN.
3. `td.update` is cool, but you can also talk about __setindex__ which is cool too 
```python
td[1:] = td2
```
4. a crucial method is `to_tensordict`. Unlike clone (which will return the same tensordict subclass) `to_tensordict` will return a regular tensordict. This is super useful to make stuff contiguous in memory.
5. let's try to get the linting right: `{"a":torch.randn(3)}` is `{"a": torch.randn(3)}`, `[3,4,5]` is `[3, 4, 5]` etc.
6. Let's tell people that their TensorDictModules should inherit from the official TensorDictModule class
(1)
```python
class TransformerBlockTensorDict(nn.Module):  # don't do that
```
should be (2)
```python
class TransformerBlockTensorDict(TensorDictModule):
```
I would advise to explicitly tell people not to do (1) but (2).
7. `nn.Sequential(TensorDictModule(...), ...)` should be `SequentialTensorDict(...)`

torchrl/modules/tensordict_module/sequence.py
```diff
@@ -118,27 +118,18 @@ def __init__(
         in_keys = []
         out_keys = []
         for module in modules:
-            in_keys += module.in_keys
+            # we sometimes use in_keys to select keys of a tensordict that are
+            # necessary to run a TensorDictModule. If a key is an intermediary in
+            # the chain, there is no reason why it should belong to the input
+            # TensorDict.
+            in_keys += [key for key in module.in_keys if key not in out_keys]
```
vmoens, line 125: these changes should not be part of this PR

vmoens: Once you update your branch from main after #261 is merged this should disappear

vmoens:
21. Also let's talk about this

```python
tensordict = TensorDict({}, [10])
for i in range(10):
    tensordict[i] = TensorDict({"a": torch.randn(3, 4)}, [])
print(tensordict)
```

22. Let's make sure that the notebook runs without exception. If you want to raise one, place it in a try / except.

vmoens:
23. Let's talk about the fact that tensordicts do not support algebraic operations

vmoens:
There is some more magic to squeeze and unsqueeze which is this
```
tensordict = TensorDict({}, [10])
assert tensordict.unsqueeze(0).squeeze(0) is tensordict
```

Same for `LazyStackedTensorDict`

vmoens:
Rather than printing stuff, perhaps you could put `assert`s with a small comment, e.g.
```python
# Converting a LazyStackedTensorDict to a regular TensorDict is easy:
assert isinstance(staked_tensordict.contiguous(), TensorDict)
```
rather than
```python
print(staked_tensordict.contiguous())
```
like that the users can understand what you're printing. Otherwise they need to read the log, and figure out what there is to see.

vmoens:
Can we remove the png from here?

PR #256, opened 2022-07-07
[Feature] Nested tensordicts

vmoens:
- [X] tests
- [x] flatten (do we need that?)  => #264 
- [x] nested composite spec -> test rand and encode => #265 
- [x] indexing using tuples `td[key1, key2]" => #262

```python
from torchrl.data import TensorDict
import torch

td1 = TensorDict({"a": torch.randn(4, 1)}, [4])
td2 = TensorDict({"b": torch.randn(4, 1), "td": td1}, [4])

print(td2)

td1 = TensorDict({"a": torch.randn(4, 1)}, [])
td2 = TensorDict({"b": torch.randn(4, 1), "td": td1}, [4])

print(td2)

td1 = TensorDict({"a": torch.randn(4, 1)}, [])
td2 = TensorDict({"b": torch.randn(4, 1)}, [4])
td2["td"] = td1

print(td2)

td1 = TensorDict({"a": torch.randn(4, 1)}, [])
td2b = TensorDict({"b": torch.randn(4, 1)}, [4])
td2b["td"] = td1

td2b[:] = td2

print(td2)

```

### DIFF
--- a/torchrl/data/tensordict/metatensor.py
+++ b/torchrl/data/tensordict/metatensor.py
@@ -71,7 +71,7 @@ class MetaTensor:
         _is_memmap: Optional[bool] = None,
         _is_tensordict: Optional[bool] = None,
     ):
-        _repr = None
+        _repr_tensordict = None
         if len(shape) == 1 and not isinstance(shape[0], (Number,)):
             tensor = shape[0]
             shape = tensor.shape
@@ -86,7 +86,7 @@ class MetaTensor:
                 dtype = tensor.dtype
             else:
                 dtype = None
-                _repr = str(tensor)
+                _repr_tensordict = str(tensor)
 
             requires_grad = (
                 tensor.requires_grad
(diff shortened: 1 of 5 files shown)


torchrl/data/tensordict/tensordict.py
```diff
@@ -67,6 +66,7 @@ class _TensorDict(Mapping, metaclass=abc.ABCMeta):
 
     _safe = False
     _lazy = False
+    is_meta = False
```
shagunsodhani, line 69: not a big deal but you want to name this as `_meta` to be consistent with the rest of the code?

torchrl/data/tensordict/metatensor.py
```diff
@@ -70,7 +70,8 @@ def __init__(
         _is_shared: Optional[bool] = None,
         _is_memmap: Optional[bool] = None,
     ):
-
+        _is_tensordict = False
+        _repr = None
```
shagunsodhani, line 74: If we want `repr` to be non-None only for tensordict, we may rename this variable to be something else?

torchrl/data/tensordict/tensordict.py
vmoens: it comes from the fact that we have checks whether a tensor has [an attribute "is_meta"](https://pytorch.org/docs/stable/generated/torch.Tensor.is_meta.html) (i.e. is it on "meta" device)
TensorDict does not support meta device...

torchrl/data/tensordict/metatensor.py
vmoens: refactored! Thanks for the suggestion

PR #257, opened 2022-07-07
[Feature] Implement padding for tensordicts

ajhinsvark:
This is for implementing https://github.com/facebookresearch/rl/issues/246

### DIFF
--- a/torchrl/data/tensordict/tensordict.py
+++ b/torchrl/data/tensordict/tensordict.py
@@ -2129,11 +2129,11 @@ def pad(tensordict: _TensorDict, pad_size: Sequence[int], value: float = 0.0):
         ...     'b': torch.ones(3, 4, 1, 1)}, batch_size=[3, 4])
         >>> dim0_left, dim0_right, dim1_left, dim1_right = [0, 1, 0, 2]
         >>> padded_td = pad(td, [dim0_left, dim0_right, dim1_left, dim1_right], value=0.0)
-        >>> print(td.batch_size)
-        torch.Size([4, 5])
-        >>> print(td.get("a").shape)
+        >>> print(padded_td.batch_size)
+        torch.Size([4, 6])
+        >>> print(padded_td.get("a").shape)
         torch.Size([4, 6, 1])
-        >>> print(td.get("b").shape)
+        >>> print(padded_td.get("b").shape)
         torch.Size([4, 6, 1, 1])
     """
 


test/test_tensordict.py
```diff
@@ -80,6 +81,23 @@ def test_stack(device):
     assert (td_reconstruct == td).all()
 
 
+@pytest.mark.parametrize("device", get_available_devices())
+def test_pad(device):
```
vmoens, line 85: Let's use `TestTensorDict`

Have a look at the tests from that class: you can test your function for each tensordict subclass.

```python
    def test_pad(self, td_name):
        torch.manual_seed(1)
        td = getattr(self, td_name)
        etc.
```

test/test_tensordict.py
```diff
@@ -80,6 +81,23 @@ def test_stack(device):
...
+    torch.manual_seed(1)
+    dim0_left, dim0_right, dim1_left, dim1_right = [0, 1, 0, 2]
```
vmoens, line 87: Can we test with other paddings too? Like `[0, 1, 0, 2]`, `[1, 0, 0, 2]`, `[1, 0, 2, 1]` or similar?
Just to make sure that nothing goes unnoticed.

test/test_tensordict.py
```diff
@@ -80,6 +81,23 @@ def test_stack(device):
     assert (td_reconstruct == td).all()
 
 
+@pytest.mark.parametrize("device", get_available_devices())
+def test_pad(device):
+    torch.manual_seed(1)
```
vmoens, line 86: if nothing is random we can ditch this

vmoens (commented):
Fantastic work! See my comments

torchrl/data/tensordict/tensordict.py
```diff
@@ -2104,6 +2104,25 @@ def stack(
     return out
 
 
+def pad(tensordict: _TensorDict, pad_size: Sequence[int], value: float = 0.0):
```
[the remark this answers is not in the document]
ajhinsvark: Do we have docstrings here, I could add explanation for the args.  I think allowing fewer dimensions makes sense since it would match the behavior of https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html (though adding left to right in dimensions rather than right to left).  A greater number returning an error makes sense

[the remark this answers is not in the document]
vmoens: Yeah if you want to sketch docstrings for pad it'd be awesome!
I would prefer if padding was going from left to right indeed, it's kind of the pattern of tensordict. If the tensordict is transparent about it I guess ppl will be fine with it (doc is everything when making weird decisions 🙃 )

test/test_tensordict.py
```diff
@@ -64,6 +65,27 @@ def test_tensordict_set(device):
     assert td._tensordict_meta["key1"].shape == td._tensordict["key1"].shape
 
 
+def test_pad():
```
ajhinsvark, line 68: I left this in because I wanted to test that pre and post padding for each dimension was working properly. I could move this to the TestTensorDicts function but it seemed repetitive

vmoens: oh i see what you mean
ok that's perfect

vmoens (approved):
LGTM! thanks for contributing!

PR #261, opened 2022-07-08
[BugFix]: avoid deleting true in_keys in TensorDictSequence

vmoens:
cc @nicolas-dufour


nicolas-dufour:
Now there are duplicates in in_keys since 2 modules use it as an in_key before it is used as an out key

vmoens:
Makes sense
Let's put one more condition before adding them

PR #264, opened 2022-07-11
[Feature]: flatten nested tensordicts

vmoens:
## Description

Introduces new `flatten_key` function and method.

## Motivation and Context

Allows to get a single, non-nested tensordict from a nested one.

## Types of changes

New function in `tensordict.py` and method in `_TensorDict` parent class.
Tests the function over all tensordict subtypes.

### DIFF
--- a/torchrl/data/tensordict/tensordict.py
+++ b/torchrl/data/tensordict/tensordict.py
@@ -3769,14 +3769,7 @@ def _td_fields(td: _TensorDict) -> str:
     return indent(
         "\n"
         + ",\n".join(
-            sorted(
-                [
-                    f"{key}: {item.class_name}({item.shape}, dtype={item.dtype})"
-                    if not item.is_tensordict()
-                    else f"{key}: {item._repr}"
-                    for key, item in td.items_meta()
-                ]
-            )
+            sorted([f"{key}: {item.get_repr()}" for key, item in td.items_meta()])
         ),
         4 * " ",
     )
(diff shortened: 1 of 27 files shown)


torchrl/data/tensordict/tensordict.py
```diff
@@ -3812,3 +3971,56 @@ def _check_keys(
...
+        # tensordict
+        out = TensorDict(
+            {},
+            [*parent_batch_size, *tensor.shape[self_batch_dims:]],
+            device=self_device,
+        )
+        return out
+
+
+def flatten_keys(
```
shagunsodhani, line 3997: we may eventually want to have a `unflatten_key` method as well

torchrl/data/tensordict/tensordict.py
```diff
@@ -1231,6 +1302,9 @@ def __iter__(self) -> Generator:
         for i in range(length):
             yield self[i]
 
+    def flatten_keys(self, separator=",", inplace: bool = True) -> _TensorDict:
+        return flatten_keys(self, separator=separator, inplace=inplace)
```
shagunsodhani, line 1306: Maybe its better to mark `flatten_keys` as a static method. The current implementation (while correct) looks a bit confusing

shagunsodhani (approved):

torchrl/data/tensordict/tensordict.py
vmoens: ok i removed the function and made it a method. 
not sure we can make it static as we need the `self` arg

torchrl/data/tensordict/tensordict.py
vmoens: makes sense I can do that

vmoens: See latest commit :)

PR #265, opened 2022-07-11
[Test]: test nested CompositeSpec

vmoens:
## Description

Tests nested CompositeSpec structures.

## Motivation and Context

We can build nested TensorDicts. Similarily, we should be able to build nested TDs just by querying `nested_composite_spec.rand()`

## Types of changes

Writes tests for nested composite specs.

### DIFF
--- a/torchrl/data/tensordict/tensordict.py
+++ b/torchrl/data/tensordict/tensordict.py
@@ -3769,14 +3769,7 @@ def _td_fields(td: _TensorDict) -> str:
     return indent(
         "\n"
         + ",\n".join(
-            sorted(
-                [
-                    f"{key}: {item.class_name}({item.shape}, dtype={item.dtype})"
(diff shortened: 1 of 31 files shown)


torchrl/data/tensordict/tensordict.py
```diff
@@ -1,4 +1,4 @@
-#pad_size, value Copyright (c) Meta Platforms, Inc. and affiliates.
+# pad_size, value Copyright (c) Meta Platforms, Inc. and affiliates.
```
shagunsodhani, line 1: `pad_size, value` string should be deleted maybe ?

shagunsodhani (approved):

torchrl/data/tensordict/tensordict.py
vmoens: 😅 yes they are now

PR #274, opened 2022-07-15
[Feature] Wandb logger

nicolas-dufour:
## Description

Adds compatibility for Weights and Bias Logger

## Motivation and Context
Now that we support multiple logger, it's time to add another.

## Types of changes

What types of changes does your code introduce? Put an `x` in all the boxes that apply:

- [] Breaking change (fix or feature that would cause existing functionality to change)

## Implemented Tasks

- Implemented wandb logger

## Checklist

- [x ] I have read the [CONTRIBUTION](CONTRIBUTING.md) guide (**required**)


torchrl/trainers/loggers.py
```diff
@@ -4,9 +4,21 @@
 # LICENSE file in the root directory of this source tree.
 
 import abc
+import os
 
+from omegaconf import OmegaConf
```
vmoens, line 9: OmegaConf is an optional dependency, as such let's avoid importing it like this.
You can do as you did below when importing wandb if importing the library has low overhead.

torchrl/trainers/loggers.py
```diff
@@ -4,9 +4,21 @@
...
 from torch import Tensor
 
+try:
+    import wandb
```
vmoens, line 13: To import tensorboard, we used to do it only in 
```python
def main()
    import tensorboard
```
The reason being that when we launch our parallel / distributed envs we want to have as few imports as possible (and `main` is called only in the master node/process).

Can we find a way of doing this here too, ie avoiding that each and every workers loads wandb / tensorboard when it is not needed?

torchrl/trainers/loggers.py
```diff
@@ -120,3 +132,116 @@ def log_hparams(self, cfg: "DictConfig") -> None:
...
+
+        self._has_imported_moviepy = False
+
+        self.video_log_counter = 0
+
+    def _create_experiment(self) -> "WandbLogger":
+        """
+        Creates a wandb experiment.
+
+        Args:
```
vmoens, line 182: can we add `Returns: blabla`?

vmoens (commented):
We need to figure out how to test all this.
I guess we'd want to change the optional dependencies in setup.py too?

vmoens (approved):
LGTM, great work!

PR #290, opened 2022-07-18
[BugFix]: Fix examples

vmoens:
## Description

Following Hydra integration, the examples are broken. We fix these and provide a short instruction markdown to explain how to run them.

## Types of changes

This is a bugfix.
It provides one yaml file per training script.

## Next changes (for another PR)

- [ ] In each directory, explain what are the hyperparams one can play with
- [ ] Implement rough tests to check that all scripts run. Put this test in circleci (independently of other tests such that they all run in parallel as we expect the script tests to be time consuming)


shagunsodhani (approved):
This looks great! I am curious as to why didnt we catch this when @BoboBananas pushed their changes for integrating Hydra ?

vmoens:
I should have run the examples at the time... I completely missed that

PR #295, opened 2022-07-19
[Feature] Truly invertible tensordict permutation of dimensions

### LINKED ISSUE #278: [Feature Request] Make `tensordict.permute(2, 1, 0).permute(0, 1, 2)` return the original tensordict
## Motivation

Inverse operations on `Tensordict`s can return the original data.
This is currently not the case for `tensordict.permute`.

## Solution

When calling `tensordict.permute(a, b, c)` we get and instance of a `PermutedTensorDict`.
If the inverse permutation matched `np.argsort([a, b, c])` then the original tensordict should be returned.

One can take inspiration from `Unsqueezed` and `SqueezedTensorDict` classes for this.

We should also cover these in the tests.

ramonmedel:
## Description

Inverse operations on Tensordicts can return the original data. This is currently not the case for tensordict.permute.

We need to return the original tensordict when calling tensordict.permute(a, b, c) we get an instance of a PermutedTensorDict and if we call again the same method with the same parameters in the resulting PermutedTensorDict 

`assert tensordict.permute(dims_list, dim).permute(dims_list, dim) is tensordict`

## Motivation and Context

`close #278`

- [ ] I have raised an issue to propose this change ([required](https://github.com/facebookresearch/rl/blob/main/CONTRIBUTING.md) for new features and bug fixes)

## Types of changes

What types of changes does your code introduce? Put an `x` in all the boxes that apply:

## Checklist

### DIFF
--- a/test/test_tensordict.py
+++ b/test/test_tensordict.py
@@ -297,6 +297,24 @@ def test_permute(device):
 
 
 @pytest.mark.parametrize("device", get_available_devices())
+def test_permute_applied_twice(device):
+    torch.manual_seed(1)
+    d = {
+        "a": torch.randn(4, 5, 6, 9, device=device),
+        "b": torch.randn(4, 5, 6, 7, device=device),
+        "c": torch.randn(4, 5, 6, device=device),
+    }
+    td1 = TensorDict(batch_size=(4, 5, 6), source=d)
+    td2 = torch.permute(td1, dims=(2, 1, 0))
+    td3 = torch.permute(td2, dims=(2, 1, 0))
+    assert td3 is td1
+    td1 = TensorDict(batch_size=(4, 5, 6), source=d)
+    td2 = torch.permute(td1, dims=(2, 1, 0))
+    td3 = torch.permute(td2, dims=(0, 1, 2))
+    assert td3 is not td1
+
+
+@pytest.mark.parametrize("device", get_available_devices())
 def test_permute_exceptions(device):
     torch.manual_seed(1)
     d = {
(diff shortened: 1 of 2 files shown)


test/test_tensordict.py
```diff
@@ -296,6 +296,24 @@ def test_permute(device):
     assert torch.sum(t["a"]) == torch.Tensor([0])
 
 
+@pytest.mark.parametrize("device", get_available_devices())
+def test_permute_applied_twice(device):
```
vmoens, line 300: It would be cool to move this to `TestTensorDict`
Here's what it will look like

```python
...
class TestTensorDict:
    ...
    def test_permute_applied_twice(self, td_name, device):
        torch.manual_seed(0)
        tensordict = getattr(self, td_name)(device)
        for _ in range(10):
             p =  torch.randperm(4)
             inv_p = p.argsort()
             other_p = inv_p
             while (other_p == inv_p).all():
                 other_p = torch.randperm(4)
             assert tensordict.permute(*p).permute(*inv_p) is tensordict
             assert tensordict.permute(*p).permute(*other_p) is not tensordict
             
```

vmoens (approved):
Can you use something like the test function i put in the comment instead?
Otherwise it looks great thanks for this!

test/test_tensordict.py
ramonmedel: Hi @vmoens , there is no class in the test files, we are using `pytest` and it doesn't work like that

vmoens: Actually it does :-)
Have a look at [this](https://github.com/facebookresearch/rl/blob/850381115c074bba89da23a6a66d7485bbc366af/test/test_tensordict.py#L519)
It would be great if we could fit this test in that. You can pretty much copy paste what i have written above

ramonmedel: Sorry for the confusion @vmoens, I have added your method as you requested

vmoens (approved):
LGTM thanks for contributing!

PR #302, opened 2022-07-20
Added test for BinarizeReward

srikanthmg85:
## Description

Added unit test for Binarize reward

## Motivation and Context

Address issue [https://github.com/facebookresearch/rl/issues/87]

## Types of changes

Unit test for existing functionality

## Checklist

- [ x] I have read the [CONTRIBUTION](https://github.com/facebookresearch/rl/blob/main/CONTRIBUTING.md) guide (**required**)

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -31,6 +31,7 @@ from torchrl.envs import (
     CatTensors,
     FlattenObservation,
     RewardScaling,
+    BinarizeReward
 )
 from torchrl.envs.libs.gym import _has_gym
 from torchrl.envs.transforms import VecNorm, TransformedEnv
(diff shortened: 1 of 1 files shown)


test/test_transforms.py
```diff
@@ -874,8 +875,26 @@ def test_noop_reset_env(self, random, device, compose):
...
+        td = TensorDict(
+            {
+                "misc": misc,
+                "reward": reward,
+            },
+            batch,
+        )
+        br(td)
+        assert td["reward"] is reward
+        assert (td["reward"] != reward_copy).all()
```
vmoens, line 896: Don't we want to test that transform does what it should too?

vmoens (commented):
Thanks for this,
Almost there!

PR #309, opened 2022-07-20
[WIP] try using mesalib instead of osmesa to fix 403s

nairbv:
Curious if this would make CI work, don't merge yet.


vmoens:
closed by #310

PR #319, opened 2022-07-25
Adding additional checks to `TensorDict.view` to minimise new object creation

bamaxw:
## Description

Currently running the following code results in a double-nested ViewedTensorDict, even though the final size - `[-1, 4]` is equivalent to the original size of `[3, 4]`. 

```python
>>> td = TensorDict({"a": torch.rand(3, 4)}, [3, 4])

>>> td.view(-1).view(-1, 4)
ViewedTensorDict(
    source=ViewedTensorDict(
        source=TensorDict({"a": ...}, batch_size=torch.Size([3, 4]), ...),
        op=view(size=torch.Size([-1]))
    ),
    op=view(size=torch.Size([-1, 4]))
)
```

This PR adds additional checks so that we can detect these cases, and handle them appropriately - i.e. return the original TensorDict.

In addition to that - running `TensorDict.view` with the shape of the tensordict being viewed - no matter whether it includes negative numbers or not (e.g. `[3, 4]`, `[3, -1]`, etc) - will now return `self` instead of a `ViewedTensorDict` with original shape as well.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/torchrl/data/tensordict/memmap.py
+++ b/torchrl/data/tensordict/memmap.py
@@ -52,7 +52,7 @@ class MemmapTensor(object):
     and remote workers that have access to
     a common storage, and as such it supports serialization and
     deserialization. It is possible to choose if the ownership is
-    transferred upon serialization / deserialization: If owenership is not
+    transferred upon serialization / deserialization: If ownership is not
     transferred (transfer_ownership=False, default), then the process where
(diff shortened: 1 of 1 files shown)


vmoens (approved):
LGTM thanks for this!

PR #330, opened 2022-07-27
[Feature] Replay Buffers refactor

### LINKED ISSUE #320: [Feature Request] Refactor replay buffer to a single class
## Motivation

We currently have two distinct replay buffer classes (+ subclasses for TensorDict custom support).
On the long term, this will be hard to maintain, as the number of classes will keep on growing following every single use case. It'll be hard to have replay buffers that inherit inherit some behaviour from multiple classes.

We should have a generic replay buffer class that takes a storage object (it already does since #223), samplers (not yet implemented) and value update strategy (first-in first out or others). This would be somewhat similar to what Reverb is doing.

## Solution

A ReplayBuffer class built in the following way
```python
rb = ReplayBuffer(
    storage=storage,
    sampler=sampler,
    writer=writer
)
```
The sampler should be able to return indices with some heuristic. The parent class should have a `sample` method and an `update` method (no ops for circular replay buffer, similar to `update_priority` for prioritized replay buffers).
The writer should have an add and extend method, and be linked to the storage.
The ReplayBuffer class should do the wiring between those three objects.

## Additional context

This feature is likely going to be bc-breaking. As such, it'd be good to implement it as a prototype (for instance in a new `data/replay_buffers/prototype.py` file?)

## Checklist
- [x] storage refactoring
- [ ] sampler refactoring
- [ ] writer refactoring

cc @thibault-lahire

bamaxw:
## Description

Refactors multiple implementations of `ReplayBuffer` from `torchrl/data/replay_buffers/replay_buffers.py` into a single modular `ReplayBuffer` that takes `Storage`, `Sampler`, and `Writer` as arguments.

* `Storage` determines how ReplayBuffer data is stored e.g. ListStorage (using simple python lists) or TensorStorage (using torch tensors)
* `Sampler` determines how data is sampled from the ReplayBuffer e.g. RandomSampler (sampling data at random) or PrioritizedSampler (sampling data according to priorities)
* `Writer` determines how data is written in the storage e.g. RoundRobinWriter (writes data in a circular fashion from index 0 to `Storage.size` and then starts at 0 again)

## Motivation and Context

Addresses issue: https://github.com/facebookresearch/rl/issues/320

close #320 

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist


vmoens (commented):
That's really cool @bamaxw! Thanks for this

I would leave it open for a little while, while we're experimenting it.

Cc @luisenp and @thibault-lahire do you think this API would suit your needs?

@xiaomengy what would you think of this refactoring of the RB? Any thought?

torchrl/data/replay_buffers/storages.py
```diff
@@ -46,7 +49,8 @@ def __len__(self):
 
 
 class ListStorage(Storage):
-    def __init__(self):
+    def __init__(self, size: int):
```
shagunsodhani, line 52: I am not sure if it is a good idea to have `size` as a parameter for this class as it does not correspond to say the max size or the current size.

test/test_rb.py
```diff
@@ -32,6 +33,247 @@
 }
 
 
+@pytest.mark.parametrize(
```
shagunsodhani, line 36: note to self: to review this file later

shagunsodhani (commented):
Looks pretty good! Thanks for working in it @bamaxw Could you please also update the description of the PR ?

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -0,0 +1,177 @@
...
+
+from ..tensordict.tensordict import TensorDictBase, LazyStackedTensorDict
+from .replay_buffers import pin_memory_output, stack_tensors, stack_td
+from .samplers import Sampler, RandomSampler
+from .storages import Storage, ListStorage
+from .utils import INT_CLASSES, to_numpy
+from .writers import Writer, RoundRobinWriter
+
+
+class ReplayBuffer:
```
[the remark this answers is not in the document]
bamaxw: Yes, I think so. @vmoens - can you confirm?

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -0,0 +1,177 @@
...
+    def __init__(
+        self,
+        storage: Storage = None,
+        sampler: Sampler = None,
+        writer: Writer = None,
+        collate_fn: Callable = None,
+        pin_memory: bool = False,
+        prefetch: int = None,
+    ) -> None:
+        self._storage = storage or ListStorage(size=1_000)
```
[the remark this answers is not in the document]
bamaxw: It isn't enforced in the `ListStorage`, but it's enforced in the `Writer`. That should be enough - as all the writes are performed by the writer anyways - so the `size` will be respected, but I'm happy to add the enforcement logic to `ListStorage` itself as well, if you think there's a reason to do that.

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -0,0 +1,177 @@
...
+        self._pin_memory = pin_memory
+
+        self._prefetch = bool(prefetch)
+        self._prefetch_cap = prefetch or 0
+        self._prefetch_queue = collections.deque()
+        if self._prefetch_cap:
+            self._prefetch_executor = ThreadPoolExecutor(max_workers=self._prefetch_cap)
+
+    def __len__(self) -> int:
+        return len(self._storage)
```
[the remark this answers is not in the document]
bamaxw: Ah yes - I removed the locks when I was writing the prototype. I just added the locks back in only now.

But I'd question the need for the lock here anyways?

If we go through a scenario in which it's important for the `len(rb)` to return **exactly** the right length, if we do something like:

```python
if len(rb) == x:
    # Already here - there's no guarantee that len(rb) is the same as on the line above
    # regardless of whether we stick a lock in `ReplayBuffer.__len__` or not, right?
    do_something()
```

and so if you want to use `len()` for something important - you'll need to lock it from the outside anyways, or if you just use it for logging - an approximation (without lock) is probably fine?

In cases where we actually do want to make sure we check length of the RB, and then act on that atomically - do we want to expose the internal `ReplayBuffer._replay_lock` to the RB users? So they can do stuff like:

```python
with rb.replay_lock:
    if len(rb) == 10:
        do_something_that_is_possible_only_when_rb_is_of_size_10_exactly()
```

Let me know if you think I'm missing something here - I'll be happy to add the lock there if it's actually needed.

torchrl/data/replay_buffers/samplers.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class RandomSampler(Sampler):
+    def sample(self, storage: Storage, batch_size: int) -> Tuple[np.array, dict]:
+        index = np.random.randint(0, len(storage), size=batch_size)
+        return index, {}
+
+
+class PrioritizedSampler(Sampler):
+    """
```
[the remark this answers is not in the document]
bamaxw: Yeah - I mostly copy-pasted code from the `PrioritizedReplayBuffer` to the `PrioritizedSampler` as and where appropriate. It's the scaffolding that's changed, not the actual implementation.

torchrl/data/replay_buffers/storages.py
bamaxw: `Writer` should take the `size` here under consideration when determining whether to add more items to the end of the storage. (see: https://github.com/facebookresearch/rl/pull/330#discussion_r940102113)

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -0,0 +1,140 @@
...
+            ret = self._prefetch_queue.popleft().result()
+
+        while len(self._prefetch_queue) < self._prefetch_cap:
+            fut = self._prefetch_executor.submit(self._sample, batch_size)
+            self._prefetch_queue.append(fut)
+
+        return ret
+
+
+class TensorDictReplayBuffer(ReplayBuffer):
```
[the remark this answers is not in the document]
bamaxw: Oh - I already made it happen 😅 . Let me know if you want me to remove it for the time being, but I've ran your tests on both ReplayBuffer and TensorDictReplayBuffer and they seem to run fine!

test/test_rb.py
```diff
@@ -16,6 +16,7 @@
     TensorDictReplayBuffer,
 )
 from torchrl.data.replay_buffers import TensorDictPrioritizedReplayBuffer
+from torchrl.data.replay_buffers import rb_prototype as prototype_rb, samplers, writers
```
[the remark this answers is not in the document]
bamaxw: I've fixed this now - I missed it before - thanks for flagging!

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -0,0 +1,177 @@
...
+from .samplers import Sampler, RandomSampler
+from .storages import Storage, ListStorage
+from .utils import INT_CLASSES, to_numpy
+from .writers import Writer, RoundRobinWriter
+
+
+class ReplayBuffer:
+    def __init__(
+        self,
+        storage: Storage = None,
```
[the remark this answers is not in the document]
bamaxw: All fixed now, thanks!

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -0,0 +1,177 @@
...
+            ret = self._prefetch_queue.popleft().result()
+
+        while len(self._prefetch_queue) < self._prefetch_cap:
+            fut = self._prefetch_executor.submit(self._sample, batch_size)
+            self._prefetch_queue.append(fut)
+
+        return ret
+
+
+class TensorDictReplayBuffer(ReplayBuffer):
```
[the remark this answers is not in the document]
bamaxw: Done! (copied from the original `torchrl.data.replay_buffers.TensorDictReplayBuffer`)

torchrl/data/replay_buffers/rb_prototype.py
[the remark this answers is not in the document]
bamaxw: I added the documentation for that now.

torchrl/data/replay_buffers/rb_prototype.py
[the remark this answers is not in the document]
vmoens: Yes it's the idea!

torchrl/data/replay_buffers/rb_prototype.py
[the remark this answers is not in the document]
vmoens: I'm a bit torn, but more in favour of enforcing.

I see why we would not enforce it (a list can be as long as we want, unlike a pre allocated array), but it happens that it gets longer than expected we should get a warning or an error. It would make sense to enforce it IMO

torchrl/data/replay_buffers/storages.py
vmoens: Could we name it max_size perhaps? Would that be if for you @shagunsodhani ?

torchrl/data/replay_buffers/rb_prototype.py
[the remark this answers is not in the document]
shagunsodhani: +1 for enforcing it

torchrl/data/replay_buffers/rb_prototype.py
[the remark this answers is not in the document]
shagunsodhani: Great points. I wanted to have a lock from the pov of correctness (since we have multiple writers here). Let me think more about if a lock is really needed. For the second part (i.e.  expose the internal ReplayBuffer._replay_lock), that is just a regular lock correct? Is there any difference between using this lock vs creating our own, high-level lock ?

torchrl/data/replay_buffers/storages.py
shagunsodhani: yeah I think `max_size` makes sense, esp given that we want to enforce the size now (https://github.com/facebookresearch/rl/pull/330#discussion_r941767100)

test/test_rb.py
```diff
@@ -16,7 +16,7 @@
     TensorDictReplayBuffer,
 )
 from torchrl.data.replay_buffers import TensorDictPrioritizedReplayBuffer
-from torchrl.data.replay_buffers import rb_prototype as prototype_rb, samplers, writers
+from torchrl.data.replay_buffers import rb_prototype, samplers, writers
```
shagunsodhani, line 19: 18 and 19 can be in the same line

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -14,6 +14,23 @@
 
 
 class ReplayBuffer:
+    """
+    #TODO: Description of the ReplayBuffer class needed.
```
shagunsodhani, line 18: Is the `TODO` part needed?

torchrl/data/replay_buffers/storages.py
bamaxw: Perfect, I changed it to `max_size` now.

torchrl/data/replay_buffers/rb_prototype.py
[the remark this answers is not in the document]
bamaxw: Ok - it's enforced now as part of `ListStorage.set` - I also renamed the `size` attribute to `max_size` as per https://github.com/facebookresearch/rl/pull/330#discussion_r941973165

test/test_rb.py
vmoens: the linting should take care of that eventually (the tests do not pass, I would guess that this is one of the issues)

torchrl/data/replay_buffers/rb_prototype.py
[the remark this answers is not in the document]
bamaxw: I think the only difference would be performance - using an internal `ReplayBuffer._replay_lock` means that it is used only for internal `ReplayBuffer`'s operations that need to be under lock, while if we use our own lock instead - we are forced to lock for the entire duration for any `ReplayBuffer` method. One example of that would be the `ReplayBuffer.__getitem__()` which in very very simple terms does two things:

```python
class ReplayBuffer:
    def __getitem__(self, index):
        with self.replay_lock:
            access_storage()  # 1. accesses Storage
        collate_data()  # 2. collates data if necessary
```

now if from the outside you want to do something like:

```python
with rb.replay_lock():
    do_something()
```

the `do_something` will run as soon as the storage has been accessed in `__getitem__`, but it won't have to wait for the `collate_data()` to execute.

While if you lock everything from the outside with your own lock - you're forced to basically run everything sequentially:

```python
# thread 1
with lock():
    rb[10]
    
# thread 2
with lock():
    do_something()
```

In this case `lock` in thread 2 will only stop blocking after the entire `rb.__getitem__()` has executed, even though technically we don't need to wait for the `collate_data()` to `do_something()`.

This is a kind of dummy example, but hopefully it conveys what I mean. Let me know what do you think.

torchrl/data/replay_buffers/rb_prototype.py
bamaxw: I can take it or leave it. Do you mean to replace it with just `Description of the ReplayBuffer class needed.` ?

test/test_rb.py
bamaxw: FYI - these tests are pretty much identical to the other tests in `test_rb.py` that were there before my changes. The only difference is that the ReplayBuffers are from `torchrl.data.replay_buffers.rb_prototype` instead of `torchrl.data.replay_buffers.replay_buffers` and there are some new methods that didn't use to exist like `update_tensordict_priority`.

torchrl/data/replay_buffers/samplers.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # Importance sampling weight formula:
+        #   w_i = (p_i / sum(p) * N) ^ (-beta)
+        #   weight_i = w_i / max(w)
+        #   weight_i = (p_i / sum(p) * N) ^ (-beta) /
+        #       ((min(p) / sum(p) * N) ^ (-beta))
+        #   weight_i = ((p_i / sum(p) * N) / (min(p) / sum(p) * N)) ^ (-beta)
+        #   weight_i = (p_i / min(p)) ^ (-beta)
+        # weight = np.power(weight / (p_min + self._eps), -self._beta)
+        weight = np.power(weight / p_min, -self._beta)
```
xiaomengy, line 117: For the weight part, we can consider to use the batch min which is weight.min() to replace the global min from the ReplayBuffer. Actually it seems the original paper didn't clarify it clearly and there are two implementations.
1. OpenAI baselines uses the global min which is same as we have now. https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/deepq/replay_buffer.py#L158
2. DeepMind Acme uses the batch min https://github.com/deepmind/acme/blob/5fac34092330fbe7d758adc487c4b97c9f58f777/acme/agents/jax/dqn/losses.py#L71

Since DeepMind uses the batch min/max, we can consider to use that which can remove the _min_tree to slightly improve the performance.

I already made this change in RLMeta https://github.com/facebookresearch/rlmeta/blob/75029c8c1b08f57f19672749efb54f571d9acab8/rlmeta/core/replay_buffer.py#L242

I also have some improvement in the implementation of SegmentTree and will create another PR to merge them in torchrl.

torchrl/data/replay_buffers/samplers.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def add(self, index: int) -> None:
+        super().add(index)
+        self._add_or_extend(index)
+
+    def extend(self, index: torch.Tensor) -> None:
+        super().extend(index)
+        self._add_or_extend(index)
+
+    def update_priority(
```
xiaomengy, line 144: I'd suggest to pay attention to one thing here for update_priority.
For distributed cases, if we do extend and update_priority from different clients, one possible case is the index we want to update is already replaced by a newly appended item, which makes the update incorrect. If the ReplayBuffer is large, the probability can be small so the overall impact should be limited.
In the current version of RLMeta, we added a timestamp to deal with this. https://github.com/facebookresearch/rlmeta/blob/75029c8c1b08f57f19672749efb54f571d9acab8/rlmeta/core/replay_buffer.py#L204
We also have an on-going PR to use global keys instead of buffer indices like reverb in https://github.com/facebookresearch/rlmeta/pull/70. We have more discussion for this design if you like.

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -0,0 +1,248 @@
...
+        Args:
+            batch_size (int): size of data to be collected.
+
+        Returns:
+            A batch of data selected in the replay buffer.
+        """
+        if not self._prefetch:
+            return self._sample(batch_size)
+
+        if len(self._prefetch_fut) == 0:
```
nicolas-dufour, line 146: self._prefect_fut do not exist. Maybe it should be self._prefetch_queue?

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -0,0 +1,248 @@
...
+                tensordicts.batch_size = tensordicts.batch_size[:1]
+            tensordicts.set(
+                "index",
+                torch.zeros(
+                    tensordicts.shape, device=tensordicts.device, dtype=torch.int
+                ),
+            )
+
+        if not isinstance(tensordicts, TensorDictBase):
+            stacked_td = torch.stack(data, 0)
```
nicolas-dufour, line 222: data is not set. Maybe ```stacked_td = torch.stack(tensordicts, 0)```?

nicolas-dufour (changes requested):

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -0,0 +1,248 @@
...
+    def __init__(
+        self,
+        storage: Optional[Storage] = None,
+        sampler: Optional[Sampler] = None,
+        writer: Optional[Writer] = None,
+        collate_fn: Optional[Callable] = None,
+        pin_memory: bool = False,
+        prefetch: Optional[int] = None,
+    ) -> None:
+        self._storage = storage or ListStorage(max_size=1_000)
```
nicolas-dufour, line 44: This do not work. I tried to set storage = LazyMemmapStorage but it got reseted to ListStorage

nicolas-dufour (changes requested):

torchrl/data/replay_buffers/samplers.py
vmoens: @xiaomengy 

> Since DeepMind uses the batch min/max, we can consider to use that which can remove the _min_tree to slightly improve the performance.

got it! Improve performance = runtime, not training performance right?

vmoens: > I also have some improvement in the implementation of SegmentTree and will create another PR to merge them in torchrl

lovely can't wait to see this

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -41,9 +41,9 @@ def __init__(
         pin_memory: bool = False,
         prefetch: Optional[int] = None,
     ) -> None:
-        self._storage = storage or ListStorage(max_size=1_000)
-        self._sampler = sampler or RandomSampler()
-        self._writer = writer or RoundRobinWriter()
+        self._storage = storage if storage is not None else ListStorage(max_size=1_000)
```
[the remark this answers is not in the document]
vmoens: Yeah for some reason it is not working. I guess it does `A or B`, which is a Boolean (not a storage or anything). 
in general I try to avoid using a non-Boolean as a Boolean, it brings surprising results way too often! Also it harms code readability: `if stuff else []` is less clear than `if stuff is not None else []`.

torchrl/data/replay_buffers/samplers.py
vmoens: Thanks for pointing this out.
After I merge the PR I will reference this as an issue. We should probably port TimeStamps to torchrl then

vmoens (approved):
LGTM thanks @bamaxw! This is truly awesome!

PR #331, opened 2022-07-27
TransformedEnv sets added Transforms into eval mode

### LINKED ISSUE #318: [BUG] `TransformedEnv` should call `transform.eval()` when a transform is added
Currently, `TransformedEnv` adds transforms "as-is", and only casts them to device. 
We should ensure that transforms that are added once the module is instantiated are being set to eval mode (if needed).

## To reproduce
```python
base_env = GymEnv("Pendulum-v1")
r = RewardScaling(0, 1)
t = TransformedEnv(base_env, r)
assert not t.transform.training  # passes
t.append(RewardScaling)
assert not t.transform[1].training  # breaks
```

## Solution
Since `append_transform` and `insert_transform` untimately call `Compose.append` and `Compose.insert`, modifying those methods in `Compose` should be sufficient.
We'd need to make sure that (1) When `Compose` is instantiated, it is automatically set to eval mode (and add that to the docstring!), (2) that `append` and `insert` both turn the transform to eval mode _if needed_

## Testing
New tests should be written in [test/test_transform.py](test/test_transform.py) to make sure that this behaviour is appropriate. We should test that all transforms are in eval if `train()` is not called, that they're all in training mode if `train` is called, that this is the case if the transformed env is created like this:
```python
base_env = GymEnv("Pendulum-v1")
t = TransformedEnv(base_env)
```
or this
```python
base_env = GymEnv("Pendulum-v1")
r = RewardScaling(0, 1)
t = TransformedEnv(base_env, r)
```

alexanderlobov:
## Description

`TransformedEnv` sets by default all transforms into eval mode.

close #318.

### DIFF
--- a/torchrl/version.py
+++ b/torchrl/version.py
@@ -1,0 +1,2 @@
+__version__ = '0.0.1c0+00ddc39'
+git_version = '00ddc39f2b2c5fcbcea6f1fc310a3a8e98069213'
--- a/torchrl/version.py
+++ b/torchrl/version.py
@@ -1,2 +1,2 @@
-__version__ = '0.0.1c0+00ddc39'
-git_version = '00ddc39f2b2c5fcbcea6f1fc310a3a8e98069213'
+__version__ = "0.0.1c0+00ddc39"
+git_version = "00ddc39f2b2c5fcbcea6f1fc310a3a8e98069213"
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -189,11 +189,13 @@ def test_vecnorm_parallel_auto(nprc):
     prcs = []
     if _has_gym:
         make_env = EnvCreator(
-            lambda: TransformedEnv(GymEnv("Pendulum-v1"), VecNorm(decay=1.0))
+            lambda: TransformedEnv(GymEnv("Pendulum-v1"), VecNorm(decay=1.0)).train()
         )
     else:
         make_env = EnvCreator(
-            lambda: TransformedEnv(ContinuousActionVecMockEnv(), VecNorm(decay=1.0))
+            lambda: TransformedEnv(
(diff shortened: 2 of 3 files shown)


alexanderlobov:
Moving `_erase_metadata` from the end to the beginning of `insert_transform`
fails the test:
FAILED test_transforms.py::TestTransforms::test_insert - AssertionError: assert None is not None

I reverted this change in the `insert_transform`.

alexanderlobov:
One of the tests that fails for me locally also fails in the main branch:

FAILED test_transforms.py::test_vecnorm_parallel_auto[5] - assert tensor([56.]) == ((5 * 11) + 2)

There are two other failing tests:

FAILED test_transforms.py::test_vecnorm_parallel_auto[2] - assert tensor([0.]) == ((2 * 11) + 2)
FAILED test_transforms.py::test_parallelenv_vecnorm - AssertionError: next_observation_sum

I'm looking into it, but any hint is appreciated

alexanderlobov:
@vmoens

torchrl/version.py
```diff
@@ -0,0 +1,2 @@
+__version__ = "0.0.1c0+00ddc39"
```
vmoens, line 1: is this intended?

vmoens (approved):
LGTM, see my minor comment

torchrl/version.py
alexanderlobov: It's a change by microfmt. I skipped it at the beginning, but then github's CI check failed, so I added it. I added it in a separate commit to not mix it up with the main changes. I can (and I would) remove the commit if needed.

vmoens: I don't think version.py is on GitHub so let's keep it that way

vmoens (approved):
LGTM Thanks for this contribution

PR #332, opened 2022-07-28
[Feature]: Added support for TensorDictSequence module subsampling

nicolas-dufour:
## Description

This PR adds the possibility of retrieving a consistent sub-network from a TensorDictSequence that is restricted to a given set of inputs and outputs. Only the blocks that are involved in the computation between input and output will be called.

## Motivation and Context
The tracing capabilities of TensorDictSequence make it possible to look at restricted sub-networks with little effort. This allows for faster code execution (compared to running full modules where only part of the output is needed) and greater readability.

## Types of changes

## Checklist


vmoens (commented):
to make this scriptable, I would rather have a seprate method that would return a sub-TensorDictSequence:

```
seq = TensorDictSequence(...)
sub_seq = seq.select(in_keys_filter=..., out_keys_filter=...)
```
Would that be doable?

nicolas-dufour:
@vmoens won't this be a problem to create new sub_seq to frequently. Typically for Dreamer, this would require to access the subsequence after every training step. So if we create a subsequence each time this will be bad for performance no?

vmoens:
Can't you reuse it? What would change between two iterations?

nicolas-dufour:
The weights of the main sequence will have been updated so we need to update the weights of the sub sequence

vmoens:
Not if they're the same weights
Updates are done in place
If that's an issue you can always work in functional mode 😊

nicolas-dufour:
I made the proposed changed and works as expected thanks for the suggestion

torchrl/modules/tensordict_module/sequence.py
```diff
@@ -191,9 +195,57 @@ def _split_param(
...
+            out_keys: output keys of the subsequence we want to select
+
+        Returns:
+            A new TensorDictSequence with only the modules that are necessary acording to the given input and output keys.
+        """
+        if in_keys is None:
+            in_keys = deepcopy(self.in_keys)
+        if out_keys is None:
+            out_keys = deepcopy(self.out_keys)
+        id_to_keep = set([i for i in range(len(self.module))])
```
vmoens, line 216: You can remove the []
You can also use {} instead of set()

torchrl/modules/tensordict_module/sequence.py
```diff
@@ -191,9 +195,57 @@ def _split_param(
...
+        for i, module in enumerate(self.module):
+            if all(key in in_keys for key in module.in_keys):
```
vmoens, line 218: What's the usage of selecting in_keys? I can understand why we want to restrict the outputs, but I don't really see when we want to restrict inputs.

vmoens: Also what happens if you say you want some out_keys but they conflict with the in_keys? Is the sequence going to be empty?

torchrl/modules/tensordict_module/sequence.py
```diff
@@ -228,7 +280,7 @@ def forward(
                 tensordict = module(tensordict)
         else:
             raise RuntimeError(
-                "TensorDictSequence does not support keyword arguments other than 'tensordict_out', 'params', 'buffers' and 'vmap'"
+                "TensorDictSequence does not support keyword arguments other than 'tensordict_out', 'in_keys', 'out_keys' 'params', 'buffers' and 'vmap'"
```
vmoens, line 283: I don't think this is still valid is it?

vmoens (commented):
Are we sure we want to select the in_keys?

torchrl/modules/tensordict_module/sequence.py
nicolas-dufour: We want to be able to select the in_keys to be able to directly input an intermediarry block. For example imagine you have a hidden layer that you want to inject from a precomputed tensordict, this allows to do so.

nicolas-dufour: Yes it would be empty if your out keys are before the in_keys

vmoens: Got it
We're doing something similar in #352: you can input an incomplete tensordict and only the relevant ops will be executed. I wonder if we need both ways of doing the same thing. The advantage of your implementation is that it is self-consistent though.

vmoens:
I see that the tests are failing, have you tried merging main into this branch?
Then I think we're ready to go if the tests pass

torchrl/modules/tensordict_module/sequence.py
vmoens: Where are the 'in_keys', 'out_keys' kwargs? I don't see them

PR #333, opened 2022-08-01
[Feature]: ModelBased Envs

nicolas-dufour:
## Description

This PR implements a Model Based Env. The goal is to be able to train a model for MBRL inside this env and then use it as any other env.

MBEnv inherits from both EnvBase and TensorDictSequence. In particular this allows for specifying different input and output keys for training and testing steps

## How to use it

This ModelBased Env is meant to be an abstract class from which other classes will be created.

```python
class DummyModelBasedEnv(ModelBasedEnv):

    def __init__(
        self,
        world_model
        device="cpu",
        dtype=None,
        batch_size=None,
    ):
        super().__init__(
            world_model,
            device,
            dtype,
            batch_size,
        )
        self.action_spec = NdUnboundedContinuousTensorSpec((1,))
        self.observation_spec = NdUnboundedContinuousTensorSpec((4,))
        self.reward_spec = NdUnboundedContinuousTensorSpec((1,))

    def _reset(self, tensordict: TensorDict, **kwargs) -> TensorDict:
        td = TensorDict(
            {
                "hidden_state": torch.randn(*self.batch_size, 4),
                "next_hidden_state": torch.randn(*self.batch_size, 4),
            },
            batch_size=self.batch_size,
        )
        return td

    def _set_seed(self, seed: int) -> int:
        return seed + 1
 
```
We can define the world model:
```python
import torch
import torch.nn as nn

from torchrl.data.tensordict.tensordict import TensorDictBase, TensorDict
from torchrl.envs.common import EnvBase
from torchrl.envs.model_based import ModelBasedEnv
from torchrl.modules import TensorDictModule
from torchrl.modules.tensordict_module.world_models import WorldModelWrapper

class ActionStateMergeLinear(nn.Module):
    def __init__(self, in_size, out_size):
        super().__init__()
        self.linear = nn.Linear(in_size, out_size)

    def forward(self, state, action):
        return self.linear(torch.cat([state, action], dim=-1))

world_modeler = TensorDictModule(
                    ActionStateMergeLinear(5, 4),
                    in_keys=["hidden_state", "action"],
                    out_keys=["next_hidden_state"],
                )
reward_model = TensorDictModule(
                   ActionStateMergeLinear(5, 1),
                    in_keys=["hidden_state", "action"]
                    out_keys=["reward"],
                )
world_model = WorldModelWrapper(
                world_modeler,
                reward_model,
            ),
```
Then, you can use functionalities of BaseEnv with your model based env
 ```python
world_model_env = DummyModelBasedEnv(world_model)
td = world_model_env.rollout(
      max_steps=100, 
      tensordict=world_model_env.observation_spec.rand(shape=[10])
)
```

## Motivation and Context

This change is needed to support MBRL in the library
## Types of changes

## Checklist

### CODE REVIEW

File: test/mocking_classes.py [hunk 1]
```diff
@@ -344,3 +348,60 @@ class DiscreteActionConvPolicy(DiscreteActionVecPolicy):
...
+            {
+                "hidden_observation": torch.randn(*self.batch_size, 4),
+                "next_hidden_observation": torch.randn(*self.batch_size, 4),
+                "action": torch.randn(*self.batch_size, 1),
+            },
+            batch_size=self.batch_size,
+        )
+        return td
+
+    def _set_seed(self, seed: int) -> int:
```
vmoens, line 397: Shouldn't set seed return another, next seed?

torchrl/envs/common.py
```diff
@@ -541,6 +541,7 @@ def to(self, device: DEVICE_TYPING) -> EnvBase:
 
         self.is_done = self.is_done.to(device)
         self.device = device
+        super().to(device)
         return self
```
vmoens, line 545: What does this do? Why do we need to keep self.device = device just above?

torchrl/envs/model_based.py
```diff
@@ -0,0 +1,105 @@
...
+import torch
+import torch.nn as nn
+
+from torchrl.data import TensorDict
+from torchrl.modules.tensordict_module.world_models import WorldModelWrapper
+from ..data.utils import DEVICE_TYPING
+from ..modules.tensordict_module import TensorDictModule, TensorDictSequence
+from .common import EnvBase
+
+dtype_map = {
```
vmoens, line 21: We already have that can't we reuse it?

torchrl/envs/model_based.py
```diff
@@ -0,0 +1,105 @@
...
+
+    @abc.abstractmethod
+    def _reset(self, tensordict: TensorDict, **kwargs) -> TensorDict:
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def _set_seed(self, seed: Optional[int]) -> int:
+        raise NotImplementedError
+
+    def to(self, device: DEVICE_TYPING) -> ModelBasedEnv:
```
vmoens, line 102: Don't we want to raise an issue if someone inputs something that is not a device (eg a dtype)? Actually it could make sense that we use a dtype no?

torchrl/envs/model_based.py
```diff
@@ -0,0 +1,105 @@
...
+        super().to(device)
+        self.world_model.to(device)
```
vmoens, line 104: If it's a nn.Module, super should do this no?

torchrl/modules/tensordict_module/common.py
```diff
@@ -166,7 +166,7 @@ def __init__(
         safe: bool = False,
     ):
 
-        super().__init__()
+        super(TensorDictModule, self).__init__()
```
vmoens, line 169: Why do we need this change?

torchrl/modules/tensordict_module/world_models.py
```diff
@@ -0,0 +1,56 @@
...
+import torch.distributions as d
+import torch.nn as nn
+import torch.nn.functional as F
+
+from torchrl.modules.tensordict_module import TensorDictModule, TensorDictSequence
+
+__all__ = ["WorldModelWrapper"]
+
+
+class WorldModelWrapper(TensorDictSequence):
```
vmoens, line 19: Does this ever happen that we have a common backbone between the two? Should we consider this option?

torchrl/trainers/helpers/envs.py
```diff
@@ -125,7 +125,7 @@ def make_env_transforms(
         env.append_transform(ToTensorImage())
         if cfg.center_crop:
             env.append_transform(CenterCrop(*cfg.center_crop))
-        env.append_transform(Resize(84, 84))
+        env.append_transform(Resize(cfg.image_size, cfg.image_size))
```
vmoens, line 128: Good catch :)

vmoens (commented):
Very good! 
Can you have a look at these?

torchrl/modules/tensordict_module/world_models.py
nicolas-dufour: What do you mean? The reward modeler will mostly be used on top of the world Model.  The goal of WorldModelWrapper is to concatenate world model and reward model but keeping an easy access to both of them.

torchrl/modules/tensordict_module/common.py
nicolas-dufour: I will revert it, it think it got changed by Copilot and i didn't pay attention when saving

torchrl/modules/tensordict_module/world_models.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Args:
+        world_model (TensorDictModule): a world model that generates a world state.
+        reward_model (TensorDictModule): a reward model, that reads the world state and returns a reward
+
+    """
+
+    def __init__(
+        self,
+        world_modeler_operator: TensorDictModule,
```
vmoens, line 34: is modeleR intended?

torchrl/modules/tensordict_module/world_models.py
vmoens: In ActorCritic and things like that the `get_policy_operator` and similar functions isolate the components that you need to run it on a tensordict collected from the replay buffer.
If WorldModelWrapper is a sequence where the reward model depends upon the `world_modeler_operator` then the fact that `get_reward_operator` returns just the reward model is counter intuitive.
I'm not super in favour of methods like
```python
def get_module(self):
    return self.module
```
Since you did not write `self._module` but `self.module`, module is considered public. Hence there is no advantage in having a `get_module` function.

We could think about the following settings:
1. reward model is based on world model: then `get_reward_model` should return what is needed to compute the reward (this should be documented in the docstring if it's the general setting)
```python
def get_reward_model(self):
    if isinstance(self.wold_model_operator, TensorDictSequence):
        sub_world_model = self.wold_model_operator.select_subsequence(out_keys=self.reward_model.in_keys)
   else:
        sub_world_model = self.wold_model_operator
    return TensorDictSequence(sub_world_model, selef.reward_model)
```
2. (*see below) reward model is a callable: this will require us to adapt TensorDictModule slightly, by automatically doing something like
```python
class callable_module(nn.Module):
    def __init__(self, callable):
        super().__init__()
        self.callable = callable

    def forward(self, *args):
        return self.callable(*args)


class TensorDictModule(...):
    def __init___(self, module, ...):
        if not isinstance(module, nn.Module):
           module = callable_module(module)
```
3. reward and world model are independent: then we don't need the get_funcitons IMO
4. there is a backbone
```python
class WorldModel(...):
    def __init__(self, backbone, obsmodel, rewardmodel):
        ...

    def get_obs_model(self):
        return TensorDictSequence(self.backbone, self.obsmodel)

    def get_reward_model(self):
        return TensorDictSequence(self.backbone, self.rewardmodel)

```

* Regarding (2.) let's consider cases where the reward model is actually known: in some instances the reward function is a differentiable function given a priori.

torchrl/modules/tensordict_module/world_models.py
vmoens: Also we have a wrapper which is a sequence,
but has the name of one of the components (`WorldModelWrapper` matches `world_model`), so the names are a bit confusing

torchrl/modules/tensordict_module/world_models.py
```diff
@@ -0,0 +1,56 @@
...
+
+class WorldModelWrapper(TensorDictSequence):
+    """
+    World model wrapper.
+    This module wraps together a world model and a reward model.
+    The world model is used to predict an imaginary world state.
+    The reward model is used to predict the reward of the imaginary world state.
+
+    Args:
+        world_model (TensorDictModule): a world model that generates a world state.
```
vmoens, line 27: I don't see a `world_model` arg in your `__init__`

torchrl/modules/tensordict_module/world_models.py
```diff
@@ -0,0 +1,56 @@
...
+        reward_model (TensorDictModule): a reward model, that reads the world state and returns a reward
```
vmoens, line 28: same, the name does not match

vmoens (commented):
some more comments

torchrl/envs/model_based.py
nicolas-dufour: Maybe let's do the check in EnvBase then? I'm not sure how changing the dtype should be handled

nicolas-dufour: Why can device can be of type int in DEVICE_TYPING?

torchrl/envs/model_based.py
nicolas-dufour: Yes you are right

torchrl/modules/tensordict_module/world_models.py
nicolas-dufour: For me, the idea in being able to retrieve the reward model is that we might want to compute the reward on some precomputed states. The thing with having this standalone reward model is that we loose this functionality.

torchrl/modules/tensordict_module/world_models.py
nicolas-dufour: yes world modeler is intended to make the distinction with world model which comports the reward modeling. I however think that it could be good to find a better name but i haven't found any atm.

torchrl/modules/tensordict_module/world_models.py
```diff
@@ -0,0 +1,56 @@
...
+from torchrl.modules.tensordict_module import TensorDictModule, TensorDictSequence
+
+__all__ = ["WorldModelWrapper"]
+
+
+class WorldModelWrapper(TensorDictSequence):
+    """
+    World model wrapper.
+    This module wraps together a world model and a reward model.
+    The world model is used to predict an imaginary world state.
```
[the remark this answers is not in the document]
nicolas-dufour: No, we also generate latent world states. The world model isn't necessarly generating observation. For example dreamer even if its generate obs as a proxy task, focus on generating latent world states and learn on those

torchrl/modules/tensordict_module/world_models.py
```diff
@@ -0,0 +1,56 @@
...
+    ):
+        super().__init__(
+            world_model,
+            reward_model,
+        )
+
+    def get_world_modeler_operator(self) -> TensorDictSequence:
+        """
+
+        Returns a stand-alone policy operator that maps an observation to an action.
```
luisenp, line 45: Is this a policy operator mapping observations to actions? Shouldn't this be something like mapping a model input (which could include observation, latent state, action, etc.) to a new world model state?

luisenp: Likewise for the reward operator. Maybe this was some actor-critic copy paste? :P

torchrl/envs/model_based.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def _step(
+        self,
+        tensordict: TensorDict,
+    ) -> TensorDict:
+        # step method requires to be immutable
+        tensordict_out = tensordict.clone()
+        # Compute world state
+        tensordict_out = self.world_model(tensordict_out)
+        # Step requires a done flag. No sense for MBRL so we set it to False
```
luisenp, line 84: I'm not sure this statement is generally true. There are some MBRL methods where the model also predicts termination conditions which are used by the policies.

torchrl/modules/tensordict_module/world_models.py
nicolas-dufour: No, this is not a mapping of obs to actions its a mapping from world latent state to next world latent state. The action part is provided separatly:
```
world_modeler = TensorDictModule(
                    ActionStateMergeLinear(5, 4),
                    in_keys=["hidden_state", "action"],
                    out_keys=["next_hidden_state"],
                )
reward_model = TensorDictModule(
                   ActionStateMergeLinear(5, 1),
                    in_keys=["hidden_state", "action"]
                    out_keys=["reward"],
                )
world_model = WorldModelWrapper(
                world_modeler,
                reward_model,
            ),
  ```
The description is indeed wrong i will fix it thanks for pointing it out

File: torchrl/envs/model_based.py (hunk 14 above)
Comment (nicolas-dufour) [reply torchrl/envs/model_based.py:84]: Ok so you would say that it's better to let the world_model to handle it? Or having a method that  from tensordict out creates the done tensor? Do you have a paper in mind so i can read it to understand the logic behind this? Thank you for the help!

Comment (luisenp) [reply torchrl/envs/model_based.py:84]: Sometimes the world model itself would handle a learned termination condition. [Here](http://proceedings.mlr.press/v144/amos21a/amos21a.pdf) is a recent example (see Eq. 12). For example, one could establish some convention that if the tensordict coming from the model has a "done" tag you would use it, otherwise, you set it to a default value.

BTW, in other methods sometimes not everything is learned. For example, in PETS the authors assume that the reward function is known, and in MBPO the termination function is also known. In MBRL-Lib we [allow](https://github.com/facebookresearch/mbrl-lib/blob/main/mbrl/models/model_env.py#L42) users to optionally pass arbitrary tensors -> tensor functions that can be used to calculate rewards/termination for a given observation and action values. 

Here this could maybe be taken care of by each `ModelBasedEnv` subclass, with some abstractions defined by the base class?

Comment (nicolas-dufour) [reply torchrl/envs/model_based.py:84]: I changed the "done" part such as world model can predict it. Thanks for the help!

Comment (nicolas-dufour) [reply torchrl/envs/model_based.py:84]: For the reward part i don 't think the current MBEnv is incompatible with what you are describing, indeed, this can be handled by the definition of the world model. You just need to define your reward function as a TDModule that have require_grad = False and then wrap it with the rest of the model part. If you think it's better that MBEnv takes world modeler and reward as an input instead i can make the change. Tell me what you think!

Comment (luisenp) [reply torchrl/envs/model_based.py:84]: I guess it depends a bit on the philosophy of the library. For MBRL-Lib we wanted to make things very modular, and in fact rely a lot on Hydra to let users swap things around easily. That's why it made sense to me to provide these as separate arguments and let the user just change which option to use via config files, rather than having them modify their model code. 

It seems to me that TorchRL is more lower level oriented (as in, here are well-tested and efficient structures and utilities that should make life easier for users), so I think current design is OK as long as this functionality can be achieved, which I agree should be the case. But I'm also curious to hear what @vmoens thinks.

File: torchrl/envs/model_based.py [hunk 15]
```diff
@@ -0,0 +1,106 @@
...
+from ..modules.tensordict_module import TensorDictModule, TensorDictSequence
+from .common import EnvBase
+
+
+class ModelBasedEnv(EnvBase, metaclass=abc.ABCMeta):
+    """
+    Basic environnement for Model Based RL algorithms.
+
+    This class is a wrapper around the model of the MBRL algorithm.
+    This class is meant to give an env framework to a world model and a reward model.
```
vmoens, line 27: do we want to cover done too?
also we could cover other things (safety constraints etc).
Perhaps something like:
`This class is meant to give an env framework to a world model (including but not limited to observations, reward, done state and safety constraints models).`

torchrl/envs/model_based.py
```diff
@@ -0,0 +1,106 @@
...
+        batch_size: Optional[torch.Size] = None,
+    ):
+        super(ModelBasedEnv, self).__init__(
+            device=device, dtype=dtype, batch_size=batch_size
+        )
+        self.world_model = world_model
+        self.world_model_params = params
+        self.world_model_buffers = buffers
+
+    def set_specs_from_env(self, env: EnvBase):
```
vmoens, line 72: let's add input_spec too

torchrl/envs/model_based.py
vmoens: A generic MBRL env class will have to predict a few things: observations, reward, done state, safety measure, possibly other things too.
I understand your point @luisenp that asking users to change their model can reduce modularity in some instances, but on the high level i think having just one model that is responsible of executing the simulation step makes it more easy to generalize the class over algorithms.
Say we build the class like this
```python
class ModelBasedEnv(EnvBase, metaclass=abc.ABCMeta):
    def __init__(
        self,
        obs_mode,
        reward_model,
        done_model,
        safety_model):
        foo()
```
I'm concerned that tomorrow someone comes and says: hey guys, how can I fit a new model that predicts something else. Then we'll either have to tell this person how to code her own env from scratch or refactor our implementation for a single use case.
I think that here we can get out of this by carefully thinking about the general use case and precisely documenting how to build a model-based environment that has or has not a model for reward, done, safety etc.
I think we can do all of this with just a few lines of code and the `TensorDictModule` will allow you to recycle most of the code across experiments.
The way I see it, it should look like this

```
# modules
obs_model = TensorDictModule(obs_net, in_keys=["observation", "action"], out_keys=["next_observation"])
reward_model = TensorDictModule(reward_net, in_keys=["observation", "action"], out_keys=["reward"])
done_model = TensorDictModule(done_net, in_keys=["observation", "action"], out_keys=["done"])
safety_model = TensorDictModule(safety_net, in_keys=["observation", "action"], out_keys=["safety_measure"])

# we can build an env with or without done
env1 = ModelBasedEnv(TensorDictSequence(obs_model, reward_model, safety_model))  # will assume done is False
env2 = ModelBasedEnv(TensorDictSequence(obs_model, reward_model, done_model, safety_model))  # will compute done from obs

# we can build an env with or without reward
env3 = ModelBasedEnv(TensorDictSequence(obs_model, safety_model))  # in this case the output will miss the reward key, has it does not have a default value

```
(by the way, safety net is a great name haha)

Does that make sense?
I feel this would work with a hydra config easily but IMO it's still highly modular (perhaps I misunderstand what you mean by modular, in which case I'm interested to know what it would take for it to satisfy that).

On the "con" side it's true that this class is not super explicit but, again, nothing can't be done with a good doc.

@nicolas-dufour 
we could move model_based.py to a model_based directory, and put a `readme.md` where we give a simple example of how that works, what do you think?

torchrl/envs/model_based.py
```diff
@@ -0,0 +1,106 @@
...
+        buffers: Optional[List[torch.Tensor]] = None,
+        device: DEVICE_TYPING = "cpu",
+        dtype: Optional[Union[torch.dtype, np.dtype]] = None,
+        batch_size: Optional[torch.Size] = None,
+    ):
+        super(ModelBasedEnv, self).__init__(
+            device=device, dtype=dtype, batch_size=batch_size
+        )
+        self.world_model = world_model
+        self.world_model_params = params
```
vmoens, line 69: here we're doing some weird stuff: we have a functional module that is not functional
it's true that we do the same for the losses though
I wonder if we could pass params and buffers in the step and reset? Maybe not super easy to do as `step` calls `_step`, so we'd need to refactor `step` too to allow it to pass kwargs to `_step`...

vmoens (commented):
Some comments
@luisenp thanks for the comments, let me know if you want to bounce back on mine

torchrl/envs/model_based.py
nicolas-dufour: The thing is that input_spec differs from model to model. That's why in dreamer i set latent_spec but maybe it should better be input_spec?

torchrl/envs/model_based.py
nicolas-dufour: Yep this sound good!

torchrl/envs/model_based.py
vmoens: Not sure I get it
The purpose of specs is to change following the experimental context no?
What does latent spec bring that input spec does not have?
Can't we fit it all in 
Action, input, observation, and reward spec?

torchrl/envs/model_based.py
luisenp: Hi @vmoens! I wonder if we were talking about different things? I pretty much agree with your vision of having a general model design and let the user decide what to provide or not for their use case via `TensorDictModule`. This completely makes sense to me. 

In my last comment I was referring specifically to letting the user have an easy way to pass custom reward (or termination) functions that are not necessarily considered part of the model (e.g. pass the ground truth reward, or no reward, or whatever), and swap ones for others. But this is not the same as suggesting that separate inputs should be passed for every possible submodel, which is maybe what my comment was interpreted as? The model can still provide arbitrary outputs, but, for example, if a hard-coded reward is given, then any (optional) reward predictions returned by the model are just ignored.

Moreover, since in `mbrl` we already provide a bunch of utilities building from pre-defined Hydra configs, a user can set up something like "use PlaNet state-based model to solve gym-HalfCheetah but actually run it with the ground truth reward function available here", and they can do this via command line with no changes to the model code. If they want to try a new reward function, they can just add another tensor->tensor function to a module that can be imported, and point to it with a simple command line argument. This is what I meant by modularity. I do understand that this can already be achieved in current `torch-RL` by putting these custom hard-coded functions inside the world model, but it seems to me that this would require some extra boilerplate?

But, in any case, I'm not arguing that what we did is the right way to do things, just that it made sense for what we wanted to do with `mbrl` (which was to have plug-and-play components with Hydra as a glue for everything).

vmoens: Thanks for the clarification!
Indeed I think I lack a bit of context there :) . Would you have something to point me to for me to get a better understanding of what it would take for us to make this possible?

torchrl/envs/model_based.py
nicolas-dufour: yeah i'll rewrite latent spec as observation spec and that should do the trick

torchrl/envs/model_based/common.py
```diff
@@ -0,0 +1,150 @@
...
+    Basic environnement for Model Based RL algorithms.
+
+    This class is a wrapper around the model of the MBRL algorithm.
+    This class is meant to give an env framework to a world model (including but not limited to observations, reward, done state and safety constraints models).
+    It is meant to behave as a classical environment.
+
+    This class is meant to be used as a base class for other environments. It is not meant to be used directly.
+
+    An example of use is the following:
+    ```python
```
vmoens, line 32: let's write examples using the proper syntax (unless there's a good reason?)
```
"""
Example:
    >>> etc
"""
```

Like this we can test these scripts easily (rather than parsing quoted code snippets)

torchrl/envs/model_based/common.py
```diff
@@ -0,0 +1,150 @@
...
+        rand_step (TensorDict, optional -> TensorDict): random step given the action spec
+        rollout (Callable, ... -> TensorDict): executes a rollout in the environment with the given policy (or random
+            steps if no policy is provided)
+
+    """
+
+    def __init__(
+        self,
+        world_model: TensorDictModule,
+        params: Optional[List[torch.Tensor]] = None,
```
vmoens, line 101: if we want to make things functional, could we do like we do for the losses?

```
def __init__(self, model):
     if not isinstance(model, functorch.FunctionalModuleWithBuffers):
         self.model, self.params, self.buffers = model.make_functional_with_buffers()
     else:
        # no idea... perhaps that will never happen
```

or something similar. 
Here I think it's confusing because it's asking to the user to give a functional module and then not to care about the fact that it is functional...

But maybe you have a more specific use case in mind

torchrl/envs/model_based/common.py
```diff
@@ -0,0 +1,150 @@
...
+        dtype: Optional[Union[torch.dtype, np.dtype]] = None,
+        batch_size: Optional[torch.Size] = None,
+    ):
+        super(ModelBasedEnv, self).__init__(
+            device=device, dtype=dtype, batch_size=batch_size
+        )
+        self.world_model = world_model
+        self.world_model_params = params
+        self.world_model_buffers = buffers
+        self._inplace_update = False
```
vmoens, line 113: shouldn't this go in `__new__`?

torchrl/envs/model_based/common.py
```diff
@@ -0,0 +1,150 @@
...
+        self.action_spec = deepcopy(env.action_spec)
+        self.reward_spec = deepcopy(env.reward_spec)
+        self.input_spec = deepcopy(env.input_spec)
+
+    def _step(
+        self,
+        tensordict: TensorDict,
+    ) -> TensorDict:
+        # step method requires to be immutable
+        tensordict_out = tensordict.clone()
```
vmoens, line 129: we can even clone with recursive=False, I think. It will be much faster and less mem-consuming

torchrl/envs/model_based/common.py
```diff
@@ -0,0 +1,150 @@
...
+            tensordict_out = self.world_model(
+                tensordict_out,
+                params=self.world_model_params,
+                buffers=self.world_model_buffers,
+            )
+        else:
+            tensordict_out = self.world_model(tensordict_out)
+        # Step requires a done flag. No sense for MBRL so we set it to False
+        if "done" not in self.world_model.out_keys:
+            tensordict_out["done"] = torch.zeros(tensordict_out.shape, dtype=torch.bool)
```
vmoens, line 141: let's create the done state immediately on the appropriate device

vmoens (commented):
Still a bit of work but we're almost done

torchrl/envs/model_based.py
luisenp: [This](https://github.com/facebookresearch/mbrl-lib/blob/main/mbrl/models/model_env.py#L42) is what we did in `mbrl`, and here is an example [config](https://github.com/facebookresearch/mbrl-lib/blob/main/mbrl/examples/conf/overrides/pets_halfcheetah.yaml#L3) that specifies the termination function to use. The rewards_fn is not specified because we also provide some defaults in `utils` package to use when `learned_rewards: false`. Also note that, in principle, `termination_fn` should also be an optional kwarg, but when I first wrote `ModelEnv` I hand't consider that termination functions could be learned. 

I guess the question for `torchRL` is how do you want to support non-learned components of the model. If the philosophy of the library is to provide a flexible API but let users figure out how to configure things, then it seems fine to me that they just code these into the world model---and I think this is already how it works, if I understand correctly.

torchrl/modules/tensordict_module/world_models.py
nicolas-dufour: Wouldn't it be possible to compute directly the reward from the obervation and the action. The goal of this module is to easily be able to separate the state generator and the reward predictor. We could of course use a TDSequence and then do tdseq[0] and tedseq[1] but this fills less intuitive for the user. The thing is i dont see why this would be less relevant than something like ActorCriticWrapper?

torchrl/envs/model_based/common.py
nicolas-dufour: um, the reason i went into funtionalizing it was to be able to pass ensemble weights for the world model. So what we could do would be to functionalize if we don't pass params and buffers?

vmoens: I don't mind that we store the params and buffers independently, but I think this should happen when instantiating the world model, to hide that away from the user. Not sure it makes sense though.
Basically:
```python
self.world_model, self.world_model_params, self.world_model_buffers = make_functional_with_buffers(world_model)
```

torchrl/modules/tensordict_module/world_models.py
vmoens: I agree that ActorCriticWrapper does the same thing :) But for context it has corresponding classes that do not do the same thing.
We can keep this, I'm happy with it for now. 
We will always be able to remove those classes if we feel they bring little advantage (usually we should try to avoid classes that just rename another class in a specific context like those two).

vmoens (commented):
Let's try to make the tests pass and wait until #399 is landed

vmoens:
@nicolas-dufour curious: what's stopping us from landing this?

nicolas-dufour:
We just need to land batch_locked first

torchrl/modules/tensordict_module/world_models.py
```diff
@@ -0,0 +1,49 @@
...
+
+__all__ = ["WorldModelWrapper"]
+
+
+class WorldModelWrapper(TensorDictSequential):
+    """
+    World model wrapper.
+    This module wraps together a world model and a reward model.
+    The world state model is used to predict an imaginary world state.
+    The reward model is used to predict the reward of the imaginary world state.
```
vmoens, line 17: `The reward model is used to predict the reward of the imagined transition.`

torchrl/modules/tensordict_module/world_models.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Args:
+        world_state_model (TensorDictModule): a world state model that generates a new world state.
+        reward_model (TensorDictModule): a reward model, that reads the world state and returns a reward
```
vmoens, line 21: missing period in the end

torchrl/envs/model_based/common.py
```diff
@@ -0,0 +1,158 @@
...
+        - action_spec (TensorSpec): sampling spec of the actions;
+        - reward_spec (TensorSpec): sampling spec of the rewards;
+        - batch_size (torch.Size): number of environments contained in the instance;
+        - device (torch.device): device where the env input and output are expected to live
+        - is_done (torch.Tensor): boolean value(s) indicating if the environment has reached a done state since the
+            last reset
+
+    Args:
+        world_model (nn.Module): model that generates world states and its corresponding rewards;
+        device (torch.device): device where the env input and output are expected to live
```
vmoens, line 83: please indicate whether these args are optional or not

vmoens (commented):
Mostly comments about docstrings and such

torchrl/envs/model_based/common.py
```diff
@@ -0,0 +1,158 @@
...
+
+class ModelBasedEnv(EnvBase, metaclass=abc.ABCMeta):
+    """
+    Basic environnement for Model Based RL algorithms.
+
+    This class is a wrapper around the model of the MBRL algorithm.
+    This class is meant to give an env framework to a world model (including but not limited to observations, reward, done state and safety constraints models).
+    It is meant to behave as a classical environment.
+
+    This class is meant to be used as a base class for other environments. It is not meant to be used directly.
```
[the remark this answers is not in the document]
nicolas-dufour: Yes since it does not make sense to use it as is

torchrl/envs/model_based/common.py
```diff
@@ -0,0 +1,158 @@
...
+    >>> td = env.reset()
+    >>> env.rollout(td, max_steps=10)
+    ```
+
+
+    Properties:
+        - observation_spec (CompositeSpec): sampling spec of the observations;
+        - action_spec (TensorSpec): sampling spec of the actions;
+        - reward_spec (TensorSpec): sampling spec of the rewards;
+        - batch_size (torch.Size): number of environments contained in the instance;
```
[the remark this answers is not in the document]
nicolas-dufour: batch_size make sense if for some reason we want to constrain the batch_size to a specific batch_size

torchrl/modules/tensordict_module/world_models.py
```diff
@@ -0,0 +1,49 @@
...
+from torchrl.modules.tensordict_module import TensorDictModule, TensorDictSequential
+
+__all__ = ["WorldModelWrapper"]
+
+
+class WorldModelWrapper(TensorDictSequential):
+    """
+    World model wrapper.
+    This module wraps together a world model and a reward model.
+    The world state model is used to predict an imaginary world state.
```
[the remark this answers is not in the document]
nicolas-dufour: For me the world model includes the reward model prediction. This explicit the component to be focused on the modeling of the states

torchrl/envs/common.py
```diff
@@ -627,9 +627,9 @@ def to(self, device: DEVICE_TYPING) -> EnvBase:
         self.reward_spec = self.reward_spec.to(device)
         self.observation_spec = self.observation_spec.to(device)
         self.input_spec = self.input_spec.to(device)
-
         self.is_done = self.is_done.to(device)
         self.device = device
+        super().to(device)
```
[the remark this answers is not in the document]
nicolas-dufour: reverted since removing self.device = device breaks the tests

test/test_env.py
```diff
@@ -412,7 +412,7 @@ def test_batch_lock_mb_env(self, device, seed=0):
         )
         mb_env = DummyModelBasedEnvBase(
             world_model, device=device, batch_size=torch.Size([10])
-        )
+        ).to(device)
```
vmoens, line 415: same

test/test_env.py
```diff
@@ -435,7 +435,7 @@ def test_batch_lock_mb_env(self, device, seed=0):
 
         mb_env = DummyModelBasedEnvBase(
             world_model, device=device, batch_size=torch.Size([])
-        )
+        ).to(device)
```
vmoens, line 438: same

torchrl/modules/tensordict_module/world_models.py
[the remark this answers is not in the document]
vmoens: but the sentence above says "    This module wraps together a world model and a reward model."

torchrl/envs/model_based/common.py
```diff
@@ -0,0 +1,162 @@
...
+            )
+        return tensordict_out
+
+    @abc.abstractmethod
+    def _reset(self, tensordict: TensorDict, **kwargs) -> TensorDict:
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def _set_seed(self, seed: Optional[int]) -> int:
+        raise NotImplementedError
```
vmoens, line 162: put a warning to tell that there's no need to call set seed

torchrl/envs/transforms/transforms.py
```diff
@@ -312,6 +312,16 @@ def __init__(
         self._observation_spec = None
         self.batch_size = self.base_env.batch_size
 
+    def __new__(cls, env, *args, **kwargs):
```
vmoens, line 315: ditch `__new__`

torchrl/modules/tensordict_module/world_models.py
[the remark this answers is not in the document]
vmoens: let's name it transition model

test/test_env.py
```diff
@@ -190,6 +193,71 @@ def test_rollout_predictability(device):
     ).all()
 
 
+def _make_model_based_envs(
```
vmoens, line 196: where is this method used?

test/mocking_classes.py
```diff
@@ -652,3 +654,59 @@ class DiscreteActionConvPolicy(DiscreteActionVecPolicy):
...
+        dtype=None,
+        batch_size=None,
+    ):
+        super(DummyModelBasedEnvBase, self).__init__(
+            world_model,
+            device=device,
+            dtype=dtype,
+            batch_size=batch_size,
+        )
+        self.action_spec = NdUnboundedContinuousTensorSpec((1,))
```
vmoens, line 684: we can delete that as it's set below in input_spec (`env._action_spec` does not exist anymore)

vmoens (approved):
Fantastic work thanks @nicolas-dufour

PR #334, opened 2022-08-01
[Refactor] Change to_tensordict to clone when called

nicolas-dufour:
## Description

This PR aims to change tensordict.to_tensordict behaviour. The idea is to make a clone each time the method is called.
## Motivation and Context

The motivation behind it is clone the tensordict when performing .to_tensordict. 

## Types of changes

## Checklist

### DIFF
--- a/torchrl/data/tensordict/tensordict.py
+++ b/torchrl/data/tensordict/tensordict.py
@@ -861,7 +861,7 @@ dtype=torch.float32)},
             a new TensorDict object containing the same values.
 
         """
-        return self.to(
+        return self.clone().to(
             TensorDict,
             batch_size=self.batch_size,
             device=self.device,
@@ -2788,7 +2788,7 @@ torch.Size([3, 2])
             if isinstance(self, dest):
                 return self
             return dest(
-                source=self.clone(),
+                source=self,
             )
         elif isinstance(dest, (torch.device, str, int)):
             dest = torch.device(dest)


vmoens:
Thanks @nicolas-dufour for this.
Here's how I would see this instead:

For TensorDictBase, return a tensordict like this
```python
TensorDict({key: value.clone() if not isinstance(value, TensordictBase) else value.to_tensordict() for key, value in self.items()}, **metadata)
```
And metadata are all the values precomputed elsewhere EXCEPT memmap (it won't be memmap because we clone) shared mem (idem). We can leave the metatensor empty (since the metadata will change).

For savedtensordict we would need to do the same but load the tensordict first.

Hope that makes sense!

torchrl/data/tensordict/tensordict.py
```diff
@@ -2786,9 +2789,9 @@ def to(
     ) -> TensorDictBase:
         if isinstance(dest, type) and issubclass(dest, TensorDictBase):
             if isinstance(self, dest):
-                return self
+                return self.clone()
```
vmoens, line 2792: Let's keep self here

torchrl/data/tensordict/tensordict.py
```diff
@@ -2786,9 +2789,9 @@ def to(
...
             return dest(
-                source=self.clone(),
```
vmoens, line 2791: We can clone, as we do t want the two tensordict to share data

torchrl/data/tensordict/tensordict.py
```diff
@@ -3619,6 +3622,9 @@ def batch_size(self) -> torch.Size:
 
     @batch_size.setter
     def batch_size(self, new_size: torch.Size):
+        td = self._load()
```
vmoens, line 3625: Let's move this to_batch_size_setter

vmoens (approved):
LGTM! Let's address these two comments and make the tests pass, then we're good to go!

vmoens:
> LGTM! Let's address these two comments and make the tests pass, then we're good to go!

@nicolas-dufour also we should rename this PR
why not: "make `to_tensordict()` create a copy of the content"?

Issue #329, opened 2022-07-27
[BUG] rendering of dm_control fails on circleci


vmoens:
## Description

Rendering must be tested in headless mode in the CI. dm_control relies on mujoco for physic simulation and rendering. In turn, mujoco rendering can use one of the following backends (indicated by the env variable MUJOCO_GL=$backend)
backend=egl: supports headless rendering. Requires OpenGL /pyopengl.
backend=osmesa: supports headless rendering but slow 
backend=glfw: does not support headless rendering

## Problem
We used to rely on osmesa for rendering on gpu machines in circleci. Osmesa was installed with conda from menpo 
conda install -c menpo osmesa
This command now fails due to an http error. Installing libmesa from conda-forge should work but the rendering stays idle in the tests, hence all gpu tests are failing.

We could try to use egl but this fails too because of an error described here: https://github.com/deepmind/mujoco/issues/398
Strangely this error disappears when logging through ssh on the machine!

The solution should make it possible to run test_env.py without error.

## Pointers
https://github.com/deepmind/dm_control/issues/136
https://github.com/deepmind/mujoco/issues/398

vmoens:
Closed as of #339

PR #335, opened 2022-08-03
[BugFix]: update linux setup_env to include libosmesa6 and libgl1-mesa-glx

tongbaojia:
## Description

Trying to fix the bug in #329.

## Motivation and Context

The PR should fix the rendering of dm_control.

## Types of changes

## Checklist


.circleci/unittest/linux/scripts/setup_env.sh
```diff
@@ -56,6 +56,9 @@ cat "${this_dir}/environment.yml"
 if [[ $OSTYPE == 'darwin'* ]]; then
   PRIVATE_MUJOCO_GL=glfw
 else
+  # Software rendering requires GLX and OSMesa.
+  apt update
+  apt install -y libgl1-mesa-glx libosmesa6
```
vmoens, line 61: You should be using yum and not apt IIRC

vmoens (commented):
I think you should be using yum instead

setup.py
```diff
@@ -168,7 +168,12 @@ def _main():
             "build_ext": BuildExtension.with_options(no_python_abi_suffix=True),
             "clean": clean,
         },
-        install_requires=[pytorch_package_dep, "numpy", "packaging"],
+        install_requires=[
+            pytorch_package_dep,
+            "numpy",
+            "packaging",
+            "functorch @ git+https://github.com/pytorch/functorch@v0.2.0",
```
vmoens, line 175: We're installing functorch in install.sh but it's broken for some reason. I will fix this tonight if that's ok. In the meantime can you remove it from there?

vmoens (approved):
Let's keep functorch installation in install.sh

setup.py
tongbaojia: OFC. The tests won't pass for now though -- and I am not sure the `setup_evn.sh` update resolves all the `OSMesa` dependency cause I can't test it locally.

tongbaojia: BTW I think the reason is listed here:
https://pytorch.org/functorch/stable/install.html

```
functorch’s Linux binaries are compatible with all PyTorch 1.12.0 binaries aside from the PyTorch 1.12.0 cu102 binary; functorch will raise an error if it is used with an incompatible PyTorch binary. This is due to a bug in PyTorch ([pytorch/pytorch#80489](https://github.com/pytorch/pytorch/issues/80489)) and will be fixed in the next PyTorch minor release.

As a workaround, please install functorch from source: pip install --user git+https://github.com/pytorch/functorch@v0.2.0
```

vmoens: In theory we're using cuda 11.3
Bear with me I'll try to fix it later today!

tongbaojia: Take ur time :)

vmoens: It should be ok now!

.circleci/unittest/linux/scripts/setup_env.sh
```diff
@@ -56,6 +56,8 @@ cat "${this_dir}/environment.yml"
 if [[ $OSTYPE == 'darwin'* ]]; then
   PRIVATE_MUJOCO_GL=glfw
 else
+  # Software rendering requires GLX and OSMesa.
+  yum install -y mesa-libOSMesa-devel.x86_64 mesa-libGL-devel.x86_64 mesa-libGLU-devel.x86_64
```
vmoens, line 60: Do we still need to install osmesa via conda after that?
Also, do you think we could make it work with `MUJOCO_GL=egl`? It should speed up some tests by 2 orders of magnitude or so...

setup.py
```diff
@@ -168,7 +168,11 @@ def _main():
             "build_ext": BuildExtension.with_options(no_python_abi_suffix=True),
             "clean": clean,
         },
-        install_requires=[pytorch_package_dep, "numpy", "packaging"],
+        install_requires=[
```
vmoens, line 171: Can we revert this? To avoid noise in the repo history :)

.circleci/unittest/linux/scripts/setup_env.sh
tongbaojia: I can def update and see what happens :)

vmoens: It seems that now egl is failing. 
Sometimes it works when you ssh on the CircleCI instance (see my opened issue on Mujoco, link in the issue description).
They suggested that if it is the case, we should look at the env variable with and without ssh.
I don't know if you have the credentials to ssh on those machines though...
That being said, if tests pass with osmesa now I'm already more than happy! Solving it for egl is really more like going the extra mile

tongbaojia: haha okay I will revert --> I probably don't have ssh access since this is my first week...

setup.py
```diff
@@ -168,7 +168,7 @@ def _main():
             "build_ext": BuildExtension.with_options(no_python_abi_suffix=True),
             "clean": clean,
         },
-        install_requires=[pytorch_package_dep, "numpy", "packaging"],
+        install_requires=[pytorch_package_dep, "numpy", "packaging", "PyOpenGL==3.1.0"],
```
tongbaojia, line 171: I am not happy with this but it seems version `3.1.6` still has issues that couldn't be resolved. More on: https://github.com/NixOS/nixpkgs/issues/76822.

.circleci/unittest/linux/scripts/setup_env.sh
tongbaojia: The cpu tests are passing just fine -- but none of the gpu tests passed. Any suggestion?

vmoens: That's the core of this issue:
rendering for dm_control is tested with GPUs. It used to work, but now that the osmesa we installed via conda is not accessible, we must find a workaround.
I would suggest to try to ssh onto the machine: click on the tests that fails and then ssh with this
image

It's very likely that you won't have the credentials to do that tho...

tongbaojia: I don't have the write permission to rerun the job with SSH...
I remember in the bug report you mentioned that when logging through SSH the bug disappears. Is it still true in the current PR? Could you help trigger it and test it? TY!

vmoens: @tongbaojia 
I have done a bit of debugging.
With 93fb291468f2a731ae1456afac42ed191f3dd44d, running the `run_tests.sh` as is did not succeed. After a few attemps, I realized that installing 

```
pip3 install pyrender
pip3 install pyopengl --upgrade
```

_in that order_ worked! You must make sure that 
```
export MUJOCO_GL=egl
export PYOPENGL_PLATFORM=egl
```
(they should be properly set in the conda env)

The tests seem to run fine with that, but it's in ssh (might break in pure headless).

Hope that helps!

vmoens: Here's the script of run_tests.sh that i managed to run:
```
#!/usr/bin/env bash

set -e

eval "$(./conda/bin/conda shell.bash hook)"
conda activate ./env

pip3 install pyrender
pip3 install pyopengl --upgrade

export MUJOCO_GL=egl
export PYOPENGL_PLATFORM=egl

export PYTORCH_TEST_WITH_SLOW='1'
python -m torch.utils.collect_env
# Avoid error: "fatal: unsafe repository"
git config --global --add safe.directory '*'

root_dir="$(git rev-parse --show-toplevel)"
env_dir="${root_dir}/env"
lib_dir="${env_dir}/lib"

# solves ImportError: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$lib_dir

pytest test/smoke_test.py -v --durations 20
pytest test/smoke_test_deps.py -v --durations 20
python3 test/test_libs.py  # those are the tests that should break
pytest --instafail -v --durations 20
```

I have added 5 lines of code, for debugging. The packages should be installed elsewhere (in setup_env.sh) and the env variable should be set in the conda env. The test/test_lib.py is already run by pytest so we can remove it.

tongbaojia: @vmoens I tired to just use the run_tests.sh in [a8e52ae](https://github.com/facebookresearch/rl/pull/335/commits/a8e52ae07e2d6adbdb6abe1beecb0e01dd57c839). Unfortunately it doesn't seem to work in headless mode. 

Could it be the ssh process changed some display env variables?

.circleci/unittest/linux/scripts/setup_env.sh
```diff
@@ -56,17 +56,25 @@ cat "${this_dir}/environment.yml"
...
 export MUJOCO_GL=$PRIVATE_MUJOCO_GL
 conda env config vars set MUJOCO_PY_MUJOCO_PATH=$root_dir/.mujoco/mujoco210 \
   DISPLAY=unix:0.0 \
   MJLIB_PATH=$root_dir/.mujoco/mujoco-2.1.1/lib/libmujoco.so.2.1.1 \
-  LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$root_dir/.mujoco/mujoco210/bin \
+  LD_LIBRARY_PATH=$root_dir/.mujoco/mujoco210/bin \
   SDL_VIDEODRIVER=dummy \
-  MUJOCO_GL=$PRIVATE_MUJOCO_GL
+  MUJOCO_GL=$PRIVATE_MUJOCO_GL\
+  PYOPENGL_PLATFORM=osmesa
```
vmoens, line 69: In my PR I used egl, can we stick to that?

vmoens (approved):
Let's use egl not osmesa

.circleci/unittest/linux/scripts/setup_env.sh
tongbaojia: sure let's do that.

vmoens:
Closing as of #339

PR #340, opened 2022-08-05
[BugFix] Clone memmap tensors on regular tensors and other replay buffer improvements

vmoens:
## Description

This PR introduces several changes aimed at making replay buffer sampling more efficient:
- MemmapTensor.clone() now returns a regular tensor. This is a more intuitive behaviour than the previous one (where another MemmapTensor was returned). Also, to get another MemmapTensor, one can simply call `MemmapTensor(memmap.clone())`.
- tensordict indexing now never returns a `SubTensorDict`. To get a `SubTensorDict`, one must built it explicitely by calling the appropriate dedicated method. This is to make sure that no overhead is introduced by working with this class.
- Replay buffer storage is now cast to the appropriate device by the helper function: this makes data sampling faster and avoids the creation of tensors on cpu that are then cast on gpu (tensors are now created on gpu immediately).


torchrl/trainers/helpers/losses.py
```diff
@@ -29,9 +29,10 @@
     SoftUpdate,
 )
 from torchrl.objectives.costs.common import LossModule
-from torchrl.objectives.costs.redq import REDQLoss
+from torchrl.objectives.costs.deprecated import REDQLoss_deprecated
```
vmoens, line 32: `REDQLoss` is memory intensive when used with large models. 
Some work is needed to make it less memory intensive.

PR #341, opened 2022-08-05
[Feature]: Dreamer support

nicolas-dufour:
## Description

In this PR we add Dreamer a model based environnement.

#### Implemented objects
To retrieve the dreamers objects, call make_dreamer

```python
from torchrl.trainers.helpers.models import make_dreamer
world_model, model_based_env, actor_model, value_model, policy = make_dreamer(
        proof_environment=proof_env, cfg=cfg, device=device, use_decoder_in_env=True,action_key="action",
        value_key="predicted_value",
    )
```
proof_env the env on which we are going to train afterward, cfg the config file (see DreamerConfig in torchrl.trainers.helpers.models)

world_model is the world model of dreamer. From a given obs, it predicts a reward, a latent world state and a reconstruction of the obs.
model_based_env is an env that works on the latent world state (no obs). It can generate new latent world state from initial latent world state. If use_decoder_in_env, the env can decode the generated states with decode_obs method
actor_model is the associated dreamer actor model
value_model predicts a value
policy combines world_model and actor, such as from a given observation, we predict the action to take

We also provide 3 loss models:

```python
world_model_loss = DreamerModelLoss(world_model, cfg).to(device)
actor_loss = DreamerActorLoss(
    actor_model, value_model, model_based_env, cfg
).to(device)
value_loss = DreamerValueLoss(value_model).to(device)
```
This 3 losses module will each enable training the 5 models above.

#### Usecase of model_based_env
Our Env allow us to generate new data.
To do so we can do:
```python
td  = model_based_env.rollout(
    max_steps=self.cfg.imagination_horizon,
    policy=self.actor_model,
)
```
By default, the env will get reseted to have defaults states of zeros. The better way to sample from it is to have a previous world state (lets call it td)
```python
td  = model_based_env.rollout(
    max_steps=self.cfg.imagination_horizon,
    policy=self.actor_model,
    auto_reset=False,
    tensordict=td
)
```

## Motivation and Context

Dreamer is very data efficient and will enable easy use of model based methods.

## Types of changes

## Checklist


torchrl/envs/mb_envs/dreamer.py
```diff
@@ -0,0 +1,258 @@
...
+        batch_size=None,
+    ):
+        super().__init__(
+            world_model=DreamerModel(
+                obs_depth=obs_depth,
+                rssm_hidden=rssm_hidden,
+                rnn_hidden_dim=rnn_hidden_dim,
+                state_dim=state_dim,
+            ),
+            reward_model=TensorDictModule(
```
vmoens, line 32: I would move these out of the call to `__init__` for clarity

torchrl/envs/mb_envs/dreamer.py
```diff
@@ -0,0 +1,258 @@
...
+        self.inference_world_model = self.word_model.select_subsequence(
+            in_keys=["initial_state", "initial_rnn_hidden", "action"],
+            out_keys=["prior_states", "prior_rnn_hiddens"],
+        )
+        self.inference_reward_model.in_keys = [
+            "prior_states",
+            "beliefs",
+        ]
+
+    def _set_optimizer(self) -> torch.optim.Optimizer:
```
vmoens, line 54: I'm a bit against having module.optimizer 
Think of this: how would you serialize this object? 
Also, what's the single responsibility of your object? Transforming data? Optimising? To me it's mixing up things that are very different in nature and it's against the single responsibility principle. Do you think we can move it out?

torchrl/envs/mb_envs/dreamer.py
```diff
@@ -0,0 +1,258 @@
...
+        return torch.optim.Adam(
+            list(self.word_model.parameters()) + list(self.reward_model.parameters()),
+            lr=1e-3,
```
vmoens, line 57: These should be hyperparameters no?

torchrl/envs/mb_envs/dreamer.py
```diff
@@ -0,0 +1,258 @@
...
+            eps=1e-4,
+        )
+
+    def loss(self, tensordict: TensorDict) -> torch.Tensor:
+        B, T, D = tensordict["prior_means"].shape
+        flat_prior_means = tensordict["prior_means"].reshape(B * T, D)
```
vmoens, line 63: It's the typical case where I would use tensordict.view(T*B, -1)
And then get("key1") and "key2"

torchrl/envs/mb_envs/dreamer.py
```diff
@@ -0,0 +1,258 @@
...
+        flat_prior_stds = tensordict["prior_stds"].reshape(B * T, D)
+        flat_prior = d.Normal(flat_prior_means, flat_prior_stds)
+        flat_posterior_means = tensordict["posterior_means"].reshape(B * T, D)
```
vmoens, line 66: Same, we can wrap 4 lines of code in one using tensordict.view

torchrl/envs/mb_envs/dreamer.py
```diff
@@ -0,0 +1,258 @@
...
+        flat_posterior_stds = tensordict["posterior_stds"].reshape(B * T, D)
+        flat_posterior = d.Normal(flat_posterior_means, flat_posterior_stds)
+        kl_loss = kl_divergence(flat_posterior, flat_prior).mean()
```
vmoens, line 69: You don't need to change the shape to do that do you? It would work with [B, T] I think

torchrl/envs/mb_envs/dreamer.py
```diff
@@ -0,0 +1,258 @@
...
+        reco_loss = (
+            0.5
+            * F.mse_loss(
+                tensordict["observation"],
```
vmoens, line 73: Better to use get(key) when you can, it's slightly faster

torchrl/envs/mb_envs/dreamer.py
```diff
@@ -0,0 +1,258 @@
...
+    def loss(self, tensordict: TensorDict) -> torch.Tensor:
+        B, T, D = tensordict["prior_means"].shape
+        flat_prior_means = tensordict["prior_means"].reshape(B * T, D)
+        flat_prior_stds = tensordict["prior_stds"].reshape(B * T, D)
+        flat_prior = d.Normal(flat_prior_means, flat_prior_stds)
+        flat_posterior_means = tensordict["posterior_means"].reshape(B * T, D)
+        flat_posterior_stds = tensordict["posterior_stds"].reshape(B * T, D)
+        flat_posterior = d.Normal(flat_posterior_means, flat_posterior_stds)
+        kl_loss = kl_divergence(flat_posterior, flat_prior).mean()
+        reco_loss = (
```
vmoens, line 70: We could make this block more readable by splitting it in single ops

torchrl/envs/mb_envs/dreamer.py
```diff
@@ -0,0 +1,258 @@
...
+            0.5
+            * F.mse_loss(
+                tensordict["observation"],
+                tensordict["reco_observation"],
+                reduction="none",
+            )
+            .mean(dim=[0, 1])
+            .sum()
+        )
+        reward_loss = 0.5 * F.mse_loss(
```
vmoens, line 80: In general i like leaving the option of using smooth l1 or l1 or others. You can look at how we do it in other losses.

torchrl/envs/mb_envs/dreamer.py
```diff
@@ -0,0 +1,258 @@
...
+        ]
+
+    def _set_optimizer(self) -> torch.optim.Optimizer:
+        return torch.optim.Adam(
+            list(self.word_model.parameters()) + list(self.reward_model.parameters()),
+            lr=1e-3,
+            eps=1e-4,
+        )
+
+    def loss(self, tensordict: TensorDict) -> torch.Tensor:
```
vmoens, line 61: It's an env, but it has a loss. Is it intended?
In torchrl we try to split what is a loss from what is an env. Could we do this here?

In general the loss module will contain the part of the model that is not needed for inference (eg value networks etc). I think dreamer could follow that too: the reward model could be in the loss module for instance.

vmoens (commented):
Some high level comments.
I would rather have a DreamerLoss module that contains all that is not needed for inference than one big env that does everything. Do you agree?
Have a look at how we do it for SAC and REDQ for instance.

torchrl/envs/mb_envs/dreamer.py
nicolas-dufour: What do you mean by this? MBEnv requires to have a world model and a reward as an init. And you cant declare nn.Module before running __init__

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,166 @@ def forward(
...
+        self.rnn_hidden_dim = rnn_hidden_dim
+
+    def forward(self, state, belief):
+        if belief is None:
+            *batch_size,_ = state.shape
+            belief = torch.zeros(*batch_size, self.rnn_hidden_dim, device=state.device)
+        hidden = self.backbone(state, belief)
+        loc = self.loc_linear(hidden)
+        scale = self.scale_linear(hidden)
+        if self.training:
```
vmoens, line 1066: Let's use exploration mode for that

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,166 @@ def forward(
 
         input = self.mlp(input)
         return self._lstm(input, hidden0_in, hidden1_in)
+
+
+class TanhActor(nn.Module):
```
vmoens, line 1044: Let's use a ProbabilisticTDModule with a TanhNormal instead

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,166 @@ def forward(
...
+        self, out_features=None, depth=None, num_cells=None, activation_class=None, rnn_hidden_dim=200
+    ):
+        super().__init__()
+        self.backbone = MLP(
+            out_features=out_features,
+            depth=depth,
+            num_cells=num_cells,
+            activation_class=activation_class,
+        )
+        self.loc_linear = nn.Linear(out_features, out_features)
```
vmoens, line 1055: Let's use NormalParamsWrapper instead

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,166 @@ def forward(
...
+        latent = self.state_to_latent(torch.cat([state, rnn_hidden], dim=-1))
+        *batch_sizes, D = latent.shape
+        latent = latent.reshape(-1, D, 1, 1)
+        obs_decoded = self.decoder(latent)
+        _, C, H, W = obs_decoded.shape
+        obs_decoded = obs_decoded.reshape(*batch_sizes, C, H, W)
+        return obs_decoded
+
+class RSSMPriorRollout(nn.Module):
+    def __init__(self, RSSMPrior):
```
vmoens, line 1122: This is confusing: it has the same name as the class. Variable names don't have caps by convention

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,166 @@ def forward(
...
+        if state is None:
+            state = torch.zeros(*batch_size, self.state_dim, device=action.device)
+        if rnn_hidden is None:
+            rnn_hidden = torch.zeros(*batch_size, self.rnn_hidden_dim, device=action.device)
+
+        action_state = self.action_state_projector(torch.cat([state, action], dim=-1))
+        rnn_hidden = self.rnn(action_state, rnn_hidden)
+        belief = rnn_hidden
+        prior = self.rnn_to_prior_projector(belief)
+        prior_mean = self.prior_mean(prior)
```
vmoens, line 1172: Let's use NormalParamWrapper instead

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,166 @@ def forward(
...
+        prior_std = F.softplus(self.prior_std(prior)) + self.min_std
+        prior_state = d.Normal(prior_mean, prior_std).rsample()
```
vmoens, line 1174: Let's not build a normal but do 
loc+sd*randn
Building dists is expensive

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,166 @@ def forward(
...
+        self.RSSMPrior = RSSMPrior
+        self.rnn_hidden_dim = RSSMPrior.rnn_hidden_dim
+    def forward(self, action):
+        prior_means = []
+        prior_stds = []
+        prior_states = []
+        beliefs = []
+        prior_state, belief= None, None 
+        for i in range(action.shape[1]):
+            prior_mean, prior_std, prior_state, belief = self.RSSMPrior(prior_state, belief, action[:, i])
```
vmoens, line 1133: I think your RNN is completely deterministic no? Give then same input it will give twice the same output. So I would suggest to use the batched version instead, it'll be considerably faster

vmoens (commented):
Let's use the TorchRL primitives instead.
Also let's use LSTM(batched_input)
With batched_input with size [B, T] instead of looping over the LSTM. It uses CuDNN and it's way way faster.
Let's avoid building distributions unless it's really necessary.

torchrl/envs/mb_envs/dreamer.py
vmoens: I just meant
No 
```
fun(a=makestuff(...), b=makeotherstuff(...))
```

But
```

a=makestuff(...)
b=makeotherstuff(...)

fun(a, b)
```

torchrl/envs/model_based.py
```diff
@@ -0,0 +1,163 @@
...
+    def _set_seed(self, seed: Optional[int]) -> int:
+        raise NotImplementedError
+
+    def to(self, device: DEVICE_TYPING) -> ModelBasedEnv:
+        super().to(device)
+        self.world_model.to(device)
+        return self
+
+
+class DreamerEnv(ModelBasedEnv):
```
vmoens, line 115: Not sure if this belongs here or in examples... let's think about it

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,166 @@ def forward(
...
+            nn.ReLU(),
+            nn.Conv2d(depth * 2, depth * 4, 4, stride=2),
+            nn.ReLU(),
+            nn.Conv2d(depth * 4, depth * 8, 4, stride=2),
+            nn.ReLU(),
+        )
+
+    def forward(self, observation):
+        *batch_sizes, C, H, W = observation.shape
+        observation = observation.reshape(-1, C, H, W)
```
vmoens, line 1088: You can use flatten for this

vmoens: Also let's use view and not reshape whenever we can

torchrl/modules/models/models.py
nicolas-dufour: No, one of the two outputs is sampled. We could however do the sampling as a secondary step to make it deterministic

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,175 @@
...
+                    "loss_reco": reco_loss,
+                    "loss_reward": reward_loss,
+                },
+                [],
+            ),
+            tensordict,
+        )
+
+    def kl_loss(self, prior_mean, prior_std, posterior_mean, posterior_std):
+        flat_prior = d.Normal(prior_mean, prior_std)
```
vmoens, line 101: Let's write the kl explicitly it'll be much faster

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,175 @@
...
+        with torch.no_grad():
+            tensordict = tensordict.select("next_prior_state", "next_belief")
+
+            tensordict.batch_size = [
+                tensordict.shape[0],
+                tensordict.get("next_belief").shape[1],
+            ]
+            tensordict.rename_key("next_prior_state", "prior_state")
+            tensordict.rename_key("next_belief", "belief")
+            tensordict = tensordict.reshape(-1).detach()
```
vmoens, line 138: Avoid reshape and use view

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,175 @@
...
+                tensordict=tensordict,
+            )
+            with hold_out_net(self.value_model):
+                tensordict = self.value_model(tensordict)
+        lambda_target = self.lambda_target(
+            tensordict.get("reward"), tensordict.get("predicted_value")
+        )
+        actor_loss = -lambda_target.mean()
+        with torch.no_grad():
+            value_td = tensordict.clone().detach()
```
vmoens, line 153: Detach under no_grad?

torchrl/objectives/costs/utils.py
```diff
@@ -338,3 +339,26 @@ def next_state_value(
...
+    Log Likelihood
+    Computes the log likelihood of the observation given the target
+    """
+
+    def __init__(self, reduction: str = "mean"):
+        super().__init__()
+        self.reduction = reduction
+
+    def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+        log_likelihood = -d.Normal(input, 1).log_prob(target)
```
vmoens, line 355: Let's not use that but distance_loss

vmoens (commented):
Some comments to improve efficiency

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,164 @@ def forward(
...
+        if belief is None:
+            *batch_size,_ = state.shape
+            belief = torch.zeros(*batch_size, self.rnn_hidden_dim, device=state.device)
+        hidden = self.backbone(state, belief)
+        loc = self.loc_linear(hidden)
+        scale = self.scale_linear(hidden)
+        if self.training:
+            action = TanhNormal(loc=loc, scale=scale).rsample()
+        else:
+            action = torch.tanh(loc)
```
vmoens, line 1069: The expected value of a tanh-normal distribution is not the tanh of the mean of the normal. Let's use ProbabilisticActor with exploration_mode="mean"!

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,164 @@ def forward(
...
+    def forward(self, state, rnn_hidden):
+        latent = self.state_to_latent(torch.cat([state, rnn_hidden], dim=-1))
+        *batch_sizes, D = latent.shape
+        latent = latent.reshape(-1, D, 1, 1)
+        obs_decoded = self.decoder(latent)
+        _, C, H, W = obs_decoded.shape
+        obs_decoded = obs_decoded.reshape(*batch_sizes, C, H, W)
+        return obs_decoded
+
+class RSSMPriorRollout(nn.Module):
```
vmoens, line 1121: To me this would work perfectly with tensordictmodule. 
That would permit us to preallocate the tensors of the rollout, which should be more efficient than stacking the outputs.

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,164 @@ def forward(
...
+        prior_means = torch.stack(prior_means, dim=1)
+        prior_stds = torch.stack(prior_stds, dim=1)
+        prior_states = torch.stack(prior_states, dim=1)
+        beliefs = torch.stack(beliefs, dim=1)
+        return prior_means, prior_stds, prior_states, beliefs
+
+class RSSMPrior(nn.Module):
+    def __init__(self, hidden_dim=200, rnn_hidden_dim=200, state_dim=20):
+        super().__init__()
+        self.min_std = 0.1
```
vmoens, line 1147: Let's make this a hyperparam

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,164 @@ def forward(
...
+            state = torch.zeros(*batch_size, self.state_dim, device=action.device)
+        if rnn_hidden is None:
+            rnn_hidden = torch.zeros(*batch_size, self.rnn_hidden_dim, device=action.device)
+
+        action_state = self.action_state_projector(torch.cat([state, action], dim=-1))
+        rnn_hidden = self.rnn(action_state, rnn_hidden)
+        belief = rnn_hidden
+        prior = self.rnn_to_prior_projector(belief)
+        prior_mean = self.prior_mean(prior)
+        prior_std = F.softplus(self.prior_std(prior)) + self.min_std
```
vmoens, line 1173: This is biased, unlike NormalParamWrapper

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,164 @@ def forward(
...
+    def forward(self, belief, obs_embedding):
+        if obs_embedding.dim() == 1:
+            obs_embedding = obs_embedding.unsqueeze(0)
+        if belief is None:
+            *batch_sizes, _ = obs_embedding.shape
+            belief = torch.zeros(*batch_sizes, self.hidden_dim, device = obs_embedding.device)
+        posterior = self.obs_rnn_to_post_projector(
+            torch.cat([belief, obs_embedding], dim=-1)
+        )
+        post_mean = self.post_mean(posterior)
```
vmoens, line 1199: Let's use NormalParamWrapper here too

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+        loss = (
+            self.lambda_kl * kl_loss
+            + self.lambda_reco * reco_loss
+            + self.lambda_reward * reward_loss
+        )
+        return (
+            TensorDict(
+                {
+                    "loss_world_model": loss,
```
vmoens, line 90: Here you are returning 4 losses, one of them being a sum of the others. To make sure that we don't do anything stupid like re-summing the losses (which is what the trainer will do) you should either return one loss only (but then we won't be able to log each of them individually) or return the decomposed loss only.

torchrl/objectives/costs/utils.py
```diff
@@ -338,3 +339,26 @@ def next_state_value(
     rewards = rewards.to(torch.float)
     target_value = rewards + (gamma ** steps_to_next_obs) * target_value
     return target_value
+
+
+class LogLikelihood(nn.Module):
```
vmoens, line 344: Isn't this name too generic?

torchrl/trainers/helpers/envs.py
```diff
@@ -5,6 +5,7 @@
 
 from dataclasses import dataclass
 from dataclasses import field as dataclass_field
+from email.mime import image
```
vmoens, line 8: Why do we need this?

torchrl/objectives/costs/dreamer.py
nicolas-dufour: If not detach we have graph optimization overlap

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,310 @@
...
+                scaler.scale(value_loss_td["loss_value"]).backward()
+
+                scaler.unscale_(world_model_opt)
+                clip_grad_norm_(world_model.parameters(), cfg.grad_clip)
+                scaler.unscale_(actor_opt)
+                clip_grad_norm_(actor_model.parameters(), cfg.grad_clip)
+                scaler.unscale_(value_opt)
+                clip_grad_norm_(value_model.parameters(), cfg.grad_clip)
+
+                scaler.step(world_model_opt)
```
vmoens, line 283: I think we should do 

loss1 = ...
Optim1.step()

Loss2 = ...
Optim2.Step()
Etc

Like this we allow the gpu to free mem when calling backward

test/mocking_classes.py
```diff
@@ -344,3 +348,60 @@ class DiscreteActionConvPolicy(DiscreteActionVecPolicy):
...
+    """
+
+    def __init__(
+        self,
+        device="cpu",
+        dtype=None,
+        batch_size=None,
+    ):
+        super(DummyModelBasedEnv, self).__init__(
+            WorldModelWrapper(
```
vmoens, line 366: Let's avoid building thing inside a caller, it's just syntax but it feels messy

torchrl/modules/models/models.py
```diff
@@ -1032,3 +1039,181 @@ def forward(
...
+        *batch_sizes, C, H, W = observation.shape
+        if len(batch_sizes) == 0:
+            end_dim=0
+        else:
+            end_dim = len(batch_sizes)-1
+        observation = torch.flatten(
+            observation, start_dim=0, end_dim=end_dim
+        )
+        obs_encoded = self.encoder(observation)
+        latent = obs_encoded.reshape(*batch_sizes, -1)
```
vmoens, line 1100: Do we need reshape or does view work?

nicolas-dufour: View does not work in this case

examples/dreamer/dreamer.py
nicolas-dufour: I was doing it like that before, however the pb is that autocast struggle in this context. Indeed according to pytorch doc only a single scaler can be created and you cannot scale again after unscale_ . I felt that 16bit precision would be better to have

test/mocking_classes.py
nicolas-dufour: I'll changed that in the MBEnv PR instead

torchrl/envs/model_based.py
luisenp: FWIW, I think it seems more appropriate in an examples folder.

torchrl/envs/model_based.py
```diff
@@ -0,0 +1,181 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
luisenp, line 1: Seems like some of the changes in this PR belong in #333? It would be cleaner for code review and log history to have this PR only include the dreamer specific implementations of the abstractions provided in #333.

torchrl/modules/models/models.py
nicolas-dufour: You mean not having RSSMPriorRollout as an nn.Module but doing the loop over a TDModule? But then how would you integrate this with a TDSequence?

torchrl/objectives/costs/dreamer.py
nicolas-dufour: From what i 've seen from trainer it doesnt retrieve losses if it doesnt start with "loss_". So renaming loss_kl by kl and so on would do the trick no?

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,355 @@
...
+
+                        stacked_pixels = torch.cat([true_pixels, reco_pixels, imagine_pxls], dim=-1)
+                        logger.log_video(
+                            "Pixels reconstruction and imagination", stacked_pixels.detach().cpu().numpy()
+                        )
+
+
+
+def recover_pixels(pixels, stats):
+    return (255 * (pixels - stats["loc"]) / stats["scale"]).clamp(min=0, max=255)
```
vmoens, line 353: this should read 
```
    return (255 * (pixels * stats["scale"] + stats["loc"])).clamp(min=0, max=255)
```

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,355 @@
...
+                                world_model_td.get("next_belief").shape[1],
+                            ]
+                            world_model_td.rename_key("posterior_states", "prior_state")
+                            world_model_td.rename_key("next_belief", "belief")
+                            world_model_td = model_based_env.rollout(
+                                max_steps=true_pixels.shape[1],
+                                policy=actor_model,
+                                auto_reset=False,
+                                tensordict=world_model_td[:, 0],
+                            ).detach()
```
vmoens, line 342: this is already under a no_grad(), no need to detach

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,355 @@
...
+                        imagine_pxls = recover_pixels(model_based_env.decode_obs(world_model_td)["reco_pixels"], stats)
+
+                        stacked_pixels = torch.cat([true_pixels, reco_pixels, imagine_pxls], dim=-1)
+                        logger.log_video(
```
vmoens, line 346: Why does t appear in the log that we have way more reconstructions than actual pixels? Storing pixels is heavy and can quickly overload the disk

examples/dreamer/dreamer.py
vmoens: image
this is after the fix and before

examples/dreamer/dreamer.py
vmoens: Can't you scale / unscale multiple times? What would be the difference between that and scaling/unscaling through a loop?

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,355 @@
...
+            for j in range(cfg.optim_steps_per_batch):
+                # sample from replay buffer
+                sampled_tensordict = replay_buffer.sample(cfg.batch_size).to(device)
+
+                with autocast(dtype=torch.float16):
+                    model_loss_td, sampled_tensordict = world_model_loss(
+                        sampled_tensordict
+                    )
+                    if cfg.record_video:
+                        world_model_td = sampled_tensordict.clone().select(
```
vmoens, line 285: careful here: you clone the whole thing then select, I would do the opposite

examples/dreamer/dreamer.py
vmoens:
```python
                with autocast(dtype=torch.float16):
                    model_loss_td, sampled_tensordict = world_model_loss(
                        sampled_tensordict
                    )
                    if cfg.record_video:
                        world_model_td = sampled_tensordict.clone().select(
                            "pixels", "reco_pixels", "posterior_states", "next_belief"

                        )[:4].detach()
                scaler1.scale(model_loss_td["loss_world_model"]).backward()
                scaler1.unscale_(world_model_opt)
                clip_grad_norm_(world_model.parameters(), cfg.grad_clip)
                scaler1.step(world_model_opt)
                world_model_opt.zero_grad()
                scaler1.update()

                with autocast(dtype=torch.float16):
                    actor_loss_td, sampled_tensordict = actor_loss(sampled_tensordict)
                scaler2.scale(actor_loss_td["loss_actor"]).backward()
                scaler2.unscale_(actor_opt)
                clip_grad_norm_(actor_model.parameters(), cfg.grad_clip)
                scaler2.step(actor_opt)
                actor_opt.zero_grad()
                scaler2.update()

                with autocast(dtype=torch.float16):
                    value_loss_td, sampled_tensordict = value_loss(sampled_tensordict)
                scaler3.scale(value_loss_td["loss_value"]).backward()
                scaler3.unscale_(value_opt)
                clip_grad_norm_(value_model.parameters(), cfg.grad_clip)
                scaler3.step(value_opt)
                value_opt.zero_grad()
                scaler3.update()
```

examples/dreamer/dreamer.py
vmoens: you should also add to(torch.uint8) at the end to remove the warning

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,355 @@
...
+    path.mkdir(exist_ok=True)
+
+    scaler = GradScaler()
+    torch.cuda.empty_cache()
+    for i, tensordict in enumerate(collector):
+
+        # update weights of the inference policy
+        collector.update_policy_weights_()
+
+        if r0 is None:
```
vmoens, line 263: we should log the training rewards somewhere

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,355 @@
...
+                scaler.step(actor_opt)
+                actor_opt.zero_grad()
+
+                scaler.step(value_opt)
+                value_opt.zero_grad()
+
+                scaler.update()
+
+                with torch.no_grad(), set_exploration_mode("mode"):
+                    td_record = record(None)
```
vmoens, line 315: on my end this step does not log any video. Perhaps the way the global_step is indicated in TensorBoard conflicts with the Wandb api?

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,355 @@
...
+
+                scaler.step(actor_opt)
+                actor_opt.zero_grad()
+
+                scaler.step(value_opt)
+                value_opt.zero_grad()
+
+                scaler.update()
+
+                with torch.no_grad(), set_exploration_mode("mode"):
```
vmoens, line 314: I personally prefer to have this out of the inner training loop. The reason is that it's easier to control the number of collection steps than the number of training steps, and small changes in the config can have a great impact in the number of evaluation data collections, which has a great physical memory cost and compute time cost.

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,355 @@
...
+
+        # update weights of the inference policy
+        collector.update_policy_weights_()
+
+        if r0 is None:
+            r0 = tensordict["reward"].mean().item()
+        pbar.update(tensordict.numel())
+        if "mask" in tensordict.keys():
+            current_frames = tensordict["mask"].sum()
+            tensordict = tensordict[tensordict.get("mask").squeeze(-1)]
```
vmoens, line 268: here you're squashing your tensordict, is it intended? IIRC we want to store trajectories no? 
From what I see in the logs we never access that but I would remove it anyway (ie just keep the second block, after the "else:")

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,355 @@
...
+        else:
+            current_frames = tensordict.numel()
+            collected_frames += current_frames
+        tensordict = tensordict.reshape(-1, cfg.batch_length)
```
vmoens, line 272: Why do we need this? Doesn't the tensordict already have this size?
This will break if there is a "mask" key in the tensordict (see my comment above)
Also reshape should be replaced by 'view' no?

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,361 @@
...
+from torchrl.trainers.helpers.envs import (
+    correct_for_frame_skip,
+    get_stats_random_rollout,
+    parallel_env_constructor,
+    transformed_env_constructor,
+    EnvConfig,
+)
+from torchrl.trainers.helpers.models import (
+    make_dreamer,
+    DreamerConfig,
```
vmoens, line 35: Why importing this when we redefine it after?

examples/dreamer/dreamer.py
nicolas-dufour: Thanks!

examples/dreamer/dreamer.py
nicolas-dufour: I've seen memory explosion without the no_grad. Tell me if i'm wrong but no grad makes sure to not collect gradients but it doesn't detach elements that are already in the graph no?

examples/dreamer/dreamer.py
nicolas-dufour: "way more reconstructions than actual pixels?" -> What do you mean by this?

examples/dreamer/dreamer.py
nicolas-dufour: Was doing the opposite before but this was changing the original tensordict, a lot of keys were missing for the actor part after, so that's why i reverted it.

examples/dreamer/dreamer.py
nicolas-dufour: No, the tensordict has the size of the collected batch which is n_workers x max_frames_per_traj with max_frames_per_traj=1000, but then we want to use tensors of size B x batch_length with batch_length = 50 in dreamer default

vmoens: are we sure that the resulting tensordict will always be properly shaped? Also that might break no, if batch_length does not match the size of the collected data. e.g. what happens if max_frames_per_traj=789 and batch_length=25?

examples/dreamer/dreamer.py
vmoens: no select does not change the original tensordict! Or there is a serious bug!

vmoens:
Can you move the `biased_softplus` to a separate PR?

torchrl/modules/tensordict_module/exploration.py
```diff
@@ -224,6 +224,7 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
         tensordict = self.td_module.forward(tensordict)
         if exploration_mode() == "random" or exploration_mode() is None:
             out = tensordict.get(self.action_key)
+            out = self._add_noise(out)
```
vmoens, line 227: Weird stuff, can you create a PR with this?

vmoens:
Here's a list of refactorings (mostly formatting) that is like to see in this PR
- [ ] place all models in a separate file (maybe mode_based_models.py?) 
- [ ] more docstrings (I made an ex for the rollout prior)
- [ ] names that match the paper or docstrings that tell users "this is the representation model, this is what is called "state" in the paper etc.)
- [ ] tests for the loss (that should have been the first thing we did, tests come first!)

nicolas-dufour:
About renaming of models. I don't think we can match the paper names since in the paper, a single model combines the utilities of RSSMPosterior, RSSMPrior and RSSMPriorRollout. However the RSSM prefix correspond to the paper name so it should be easy to understand for the user. I'll make sure the docstrings are explicit enough so that the user isn't confused. I'll add a link to the planet paper instead of dreamer for RSSM since Dreamer is not very explicit about RSSM which was introduced by Planet

vmoens:
> However the RSSM prefix correspond to the paper name so it should be easy to understand for the user

I've been spending my weekend trying to make the parallel between the code and the paper and I can tell you it is all but straightforward. Ok for not renaming but we need to do better in the docstrings and comments in the code to make the parallel explicit.

> I'll add a link to the planet paper instead of dreamer for RSSM since Dreamer is not very explicit about RSSM which was introduced by Planet

sounds great!

torchrl/envs/transforms/transforms.py
```diff
@@ -1584,15 +1584,16 @@ def __init__(
...
-        if (
-            ("reward" in self.keys_in)
-            or ("action" in self.keys_in)
-            or ("reward" in self.keys_in)
-        ):
-            raise RuntimeError(
-                "Concatenating observations and reward / action / done state "
-                "is not allowed."
-            )
+        # TODO: check this: do we want to allow it?
```
vmoens, line 1587: I think that for now we can remove this
The idea was originally that it would be dangereous to concatenate reward and action for instance, but upon reflection i guess that if someone asks for that either she knows what she's doing or she's just far of understanding anything of this class. So I think we can remove that block

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+        out_features (int): Number of output features.
+        depth (int): Number of hidden layers.
+        num_cells (int): Number of hidden units per layer.
+        activation_class (nn.Module): Activation class.
+        rnn_hidden_size (int): Size of the hidden state of the RNN in the RSSM module.
+    """
+
+    def __init__(
+        self,
+        out_features=None,
```
vmoens, line 42: why the default None? Can `DreamerActor` run without `out_feature`?

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    It output the mean and the scale of the action distribution.
+
+    Reference: https://arxiv.org/abs/1912.016034
+
+    Args:
+        out_features (int): Number of output features.
+        depth (int): Number of hidden layers.
+        num_cells (int): Number of hidden units per layer.
+        activation_class (nn.Module): Activation class.
+        rnn_hidden_size (int): Size of the hidden state of the RNN in the RSSM module.
```
vmoens, line 37: this is optional right? I guess the other are not (see my comment below)

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+
+class ObsDecoder(nn.Module):
+    """Observation decoder network.
+
+    Takes the deterministic state and the stochastic belief and decodes it into a pixel observation.
+
+    Reference: https://arxiv.org/abs/1803.10122
+
+    Args:
+        depth (int): Number of hidden units in the last layer.
```
vmoens, line 112: this is optional

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+        obs_decoded = self.decoder(latent)
+        _, C, H, W = obs_decoded.shape
+        obs_decoded = obs_decoded.view(*batch_sizes, C, H, W)
+        return obs_decoded
+
+
+class RSSMRollout(nn.Module):
+    """Rollout the RSSM network.
+
+    Given a set of encoded observations and actions, this function will rollout the RSSM network to compute all the intermediate
```
vmoens, line 145: it's not a function but a module

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    states and beliefs.
```
vmoens, line 146: believes

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    Here, we use the previous posterior as the prior for the next time step. At the first time step, we use the an empty prior to start the rollout.
```
vmoens, line 147: The previous posterior is used as the prior for the next time step

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    At the end, we end up with all the intermediate states and beliefs.
```
vmoens, line 148: Let's reformulate this a bit no? Like "The forward method returns a stack of all intermediate states and believes" or something like it?

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+            posterior_means,
+            posterior_stds,
+            posterior_states,
+        )
+
+
+class RSSMPrior(nn.Module):
+    """The prior network of the RSSM.
+
+    This network takes the previous state and belief and the action and outputs the next state and belief.
```
vmoens, line 227: this sentence makes no sense to me
the previous state and belief + current action?
takes = takes as input?
I would split it in two
```
This network takes as input the previous state and belief and the current action.
It outputs the next state and belief, as well as the parameters of the state distribution
```

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    State is by construction stochastic and belief is deterministic. In the paper this is called the
+    deterministic state and stochastic state respectively. We prefer to call them belief and state for clarity.
```
vmoens, line 229: We prefer to call them belief and state for clarity.

this sounds a bit presumptuous IMO. Can we remove it? I don't think it brings much.

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    Furthermore, the paper refers to both prior and posterior as state. We will disntinguish them by calling
```
vmoens, line 230: `Furthermore, the paper refers to both prior and posterior as state` 
I don't think so: prior and posterior are distributions over state, this is perfectly legit.

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    the prior state and posterior state.
```
vmoens, line 231: `We distinguish states sampled according prior and posterior distribution for clarity.`

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    The RSSMPrior is a transition model in the sense that it takes the previous state and belief and the action
```
vmoens, line 232: that's the same sentence as above no?

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    and outputs the next state and belief. It contains a recurrent network that is used to model the temporal
+    dynamics of the state.
+    Other than the state and the belief, the RSSMPrior also outputs the mean and standard deviation of the
```
vmoens, line 235: I referred to that above, so we could remove it here

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    state distribution. This is useful for the loss function.
+
+    Reference: https://arxiv.org/abs/1811.04551
+
+    Args:
```
vmoens, line 240: all those args seem to have default values

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+        hidden_dim (int): Number of hidden units in the linear network. Input size of the recurrent network.
+        rnn_hidden_dim (int): Number of hidden units in the recurrent network. Also size of the belief.
+        state_dim (int): Size of the state.
+        action_spec (TensorSpec): Action spec.
```
vmoens, line 244: perhaps mention that it can be `None` and what will happen in that case?

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+        prior_mean, prior_std = self.rnn_to_prior_projector(belief)
+        prior_std = prior_std + 0.1
+        prior_state = prior_mean + torch.randn_like(prior_std) * prior_std
+        return prior_mean, prior_std, prior_state, belief
+
+
+class RSSMPosterior(nn.Module):
+    """The posterior network of the RSSM.
+
+    This network takes the belief and the encoded observation and outputs the associated posterior state.
```
vmoens, line 286: let's see my comment about this above. We can split this sentence in 2 parts I believe

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+    It also outputs the mean and standard deviation of the state distribution. This is useful for the loss function.
```
vmoens, line 287: `useful` makes it sound like this is optional. I'd rather say `which are inputs to the loss function` or similar.

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,316 @@
...
+
+    Reference: https://arxiv.org/abs/1811.04551
+
+    Args:
+        hidden_dim (int): Number of hidden units in the linear network.
```
vmoens, line 292: these have default values, let's make them optional

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,295 @@
...
+from torchrl.objectives.costs.common import LossModule
+from torchrl.objectives.costs.utils import hold_out_net, distance_loss
+from torchrl.objectives.returns.functional import vec_td_lambda_return_estimate
+
+
+class DreamerModelLoss(LossModule):
+    """Dreamer Model Loss
+
+    Computes the loss of the dreamer world model. The loss is composed of the kl divergence between the prior and posterior of the RSSM,
+    the reconstruction loss between the observation and the reconstructed observation and the reward loss between the true reward and the predicted reward.
```
vmoens, line 23: can we reformulate `loss between X and Y`? Per se a loss is not a distance or a divergence...

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,295 @@
...
+
+    Reference: https://arxiv.org/abs/1912.01603
+
+    Args:
+        world_model (TensorDictModule): the world model.
+        cfg (DictConfig): the config file.
+        lambda_kl (float): the weight of the kl divergence loss.
```
vmoens, line 30: from here everything has default values, let's make them optional

torchrl/objectives/costs/dreamer.py
vmoens: yeah but you sum it above!

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,295 @@
...
+        loss = (
+            self.lambda_kl * kl_loss
+            + self.lambda_reco * reco_loss
+            + self.lambda_reward * reward_loss
+        )
+        return (
+            TensorDict(
+                {
+                    "loss_world_model": loss,
+                    "kl_model_loss": kl_loss,
```
vmoens, line 106: can we detatch these values, since one of them is already the aggregation of the three?

the preferred (ie more consistent with other loss modules) implementation IMO would be to return this
```
            TensorDict(
                {
                    "loss_kl": self.lambda_kl * kl_loss,
                    "reco_model_loss": self.lambda_reco * reco_loss,
                    "reward_model_loss": self.lambda_reward * reward_loss,
...
```
and do the sum later on

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,295 @@
...
+            kl = kl.sum(-1).mean().clamp_min(self.free_nats)
+        else:
+            kl = kl.sum(-1).clamp_min(self.free_nats).mean()
+        return kl
+
+
+class DreamerActorLoss(LossModule):
+    """Dreamer Actor Loss
+
+    Computes the loss of the dreamer actor. The actor loss is computed as the negative lambda return average.
```
vmoens, line 138: The actor loss is computed as the negative average lambda return.

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,295 @@
...
+
+    Reference: https://arxiv.org/abs/1912.01603
+
+    Args:
+        actor_model (TensorDictModule): the actor model.
+        value_model (TensorDictModule): the value model.
+        model_based_env (DreamerEnv): the model based environment.
+        cfg (DictConfig): the config file.
+        gamma (float, optional): the discount factor.
```
vmoens, line 147: let's not forget about default values, it's important when users call `help(class)`

vmoens: lambda is also a discount factor, no?

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,295 @@
...
+    """Dreamer Value Loss
+
+    Computes the loss of the dreamer value model. The value loss is computed as the mean squared error between the predicted value and the lambda target.
+
+    Reference: https://arxiv.org/abs/1912.01603
+
+    Args:
+        value_model (TensorDictModule): the value model.
+        value_loss (str, optional): the loss to use for the value loss.
+        gamma (float, optional): the discount factor.
```
vmoens, line 239: lambda is also a discount factor

vmoens (commented):
I left some comments here and there, let's not forget about `optional` and `default` values in the args.
Thanks for the description of RSSMPrior, much clearer now

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,289 @@
...
+            tensordict.rename_key("posterior_state", "prior_state")
+            tensordict = tensordict.view(-1).detach()
+        with hold_out_net(self.model_based_env), set_exploration_mode("random"):
+            tensordict = self.model_based_env.rollout(
+                max_steps=self.imagination_horizon,
+                policy=self.actor_model,
+                auto_reset=False,
+                tensordict=tensordict,
+            )
+            tensordict = step_tensordict(
```
vmoens, line 182: Thread:
(1) here we do `step_tensordict`
Hence the `next_prior_state` and `next_belief` become `prior_state` and `belief`
(by the way, I would rename tensordict in `next_tensordict` for clarity)

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,289 @@
...
+                tensordict,
+                keep_other=True,
+                exclude_reward=False,
+                exclude_action=False,
+            )
+            with hold_out_net(self.value_model):
+                tensordict = self.value_model(tensordict)
```
vmoens, line 189: (2) Now we pass this through the value model, hence we have the value of the next state

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,289 @@
...
+                    "loss_actor": actor_loss,
+                },
+                batch_size=[],
+            ),
+            tensordict.detach(),
+        )
+
+    def lambda_target(self, reward: torch.Tensor, value: torch.Tensor) -> torch.Tensor:
+        done = torch.zeros(reward.shape, dtype=torch.bool, device=reward.device)
+        return vec_td_lambda_return_estimate(
```
vmoens, line 218: (3) now we take the values from the second and on (hence the next_next_state) + we discard one datapoint)
I think it's wrong, we should just take value and reward and done, without indexing them

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,289 @@
...
+        with hold_out_net(self.model_based_env), set_exploration_mode("random"):
+            tensordict = self.model_based_env.rollout(
+                max_steps=self.imagination_horizon,
+                policy=self.actor_model,
+                auto_reset=False,
+                tensordict=tensordict,
+            )
+            tensordict = step_tensordict(
+                tensordict,
+                keep_other=True,
```
vmoens, line 184: this is dangereous:
say you `next_prior_state` is not there but you have `prior_state`
Then you will think you're working with the `next_prior_state` but you're not.
If would turn this off if possible. Otherwise, can you tell which value needs to be kept?

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,289 @@
...
+                exclude_reward=False,
+                exclude_action=False,
+            )
+            with hold_out_net(self.value_model):
+                tensordict = self.value_model(tensordict)
+
+        lambda_target = self.lambda_target(
+            tensordict.get("reward"), tensordict.get("predicted_value")
```
vmoens, line 192: here reward comes from a tensordict that has gone through `step_tensordict`, I don't think it's intentional

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,289 @@
...
+        )
+        tensordict = tensordict[:, :-1]
```
vmoens, line 194: why do we need this?

vmoens (commented):
I have not done a full review but i'd suggest to be extra careful with the time indicators, e.g. `next_` and `prev_`.
Also I think the `DreamerActorLoss` is wrong, see my comments

torchrl/objectives/costs/dreamer.py
nicolas-dufour: Why would next_prior_state not be there? step tensordict is anyway called between each rollout step so if we don't have this, it would also be confusing.

torchrl/objectives/costs/dreamer.py
nicolas-dufour: Because we don't have lambda target for the last t

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,289 @@
...
+        tensordict.set("lambda_target", lambda_target)
+
+        if self.discount_loss:
+            discount = self.gamma * torch.ones_like(
+                lambda_target, device=tensordict.device
+            )
+            discount[:, 0] = 1
+            discount = discount.cumprod(dim=1).detach()
```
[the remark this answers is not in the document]
nicolas-dufour: yeah i just translated the no grad from tensorflow but you are right it does not makes sense

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,305 @@
...
+        rnn_hidden_size (int, optional): Size of the hidden state of the RNN in the RSSM module.
+    """
+
+    def __init__(
+        self,
+        out_features,
+        depth=4,
+        num_cells=200,
+        activation_class=nn.ELU,
+        rnn_hidden_dim=200,
```
vmoens, line 46: let's remove this I don't think it's being used

torchrl/modules/models/model_based_models.py
```diff
@@ -0,0 +1,310 @@
...
+        """
+        prior_means = []
+        prior_stds = []
+        prior_states = []
+        beliefs = []
+        posterior_means = []
+        posterior_stds = []
+        posterior_states = []
+
+        for i in range(actions.shape[1]):
```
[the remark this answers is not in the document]
vmoens: I don't get that
`belief` is the result of `f(action_t, belief_t)` so it's `next_belief` (or `belief_{t+1}`), as `next_observation` is the result of `f(observation_t, action_t)` no?

torchrl/trainers/helpers/models.py
```diff
@@ -1148,6 +1164,250 @@ def make_redq_model(
     return model
 
 
+def make_dreamer(
```
vmoens, line 1167: can we add docstrings here?

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,268 @@
...
+        done = torch.zeros(reward.shape, dtype=torch.bool, device=reward.device)
+        return vec_td_lambda_return_estimate(
+            self.gamma, self.lmbda, value, reward, done
+        )
+
+
+class DreamerValueLoss(LossModule):
+    """Dreamer Value Loss
+
+    Computes the loss of the dreamer value model. The value loss is computed as the mean squared error between the predicted value and the lambda target.
```
vmoens, line 209: not necessarily L2 since your loss is an arg of the constructor
maybe we just want to talk about the distance loss or smth?

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def __init__(
+        self,
+        actor_model: TensorDictModule,
+        value_model: TensorDictModule,
+        model_based_env: DreamerEnv,
+        imagination_horizon: int = 15,
+        gamma: int = 0.99,
+        lmbda: int = 0.95,
+        discount_loss: bool = False,  # for consistency with paper
```
vmoens, line 151: I turned this to False but we can experiment if things still run when it's true

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,268 @@
...
+        self.model_based_env = model_based_env
+        self.imagination_horizon = imagination_horizon
+        self.gamma = gamma
+        self.lmbda = lmbda
+        self.discount_loss = discount_loss
+
+    def forward(self, tensordict: TensorDict) -> torch.Tensor:
+        with torch.no_grad():
+            tensordict = tensordict.select("state", "belief")
+            tensordict = tensordict.reshape(-1)  # 50 x 50 -> 2500
```
vmoens, line 165: I forgot to remove the comment here

torchrl/objectives/costs/dreamer.py
```diff
@@ -0,0 +1,268 @@
...
+    Reference: https://arxiv.org/abs/1912.01603
+
+    Args:
+        world_model (TensorDictModule): the world model.
+        lambda_kl (float, optional): the weight of the kl divergence loss. default: 1.0.
+        lambda_reco (float, optional): the weight of the reconstruction loss. default: 1.0.
+        lambda_reward (float, optional): the weight of the reward loss. default: 1.0
+        reco_loss (str, optional): the reconstruction loss. default to l2.
+        reward_loss (str, optional): the reward loss. default to l2.
+        free_nats (int, optional): the free nats.
```
vmoens, line 36: default missing

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,292 @@
...
+        prior_mean, prior_std = self.rnn_to_prior_projector(belief)
+        state = prior_mean + torch.randn_like(prior_std) * prior_std
+        return prior_mean, prior_std, state, belief
+
+
+class RSSMPosterior(nn.Module):
+    """The posterior network of the RSSM.
+
+    This network takes as input the belief and the associated encoded observation.
+    It outputs the next posterior state and belief, as well as the parameters of the posterior state distribution.
```
vmoens, line 263: > It outputs the next posterior state and belief

not really, there is no belief. Also a more appropriate way of saying this would be "It returns the parameters of the posterior as well as a state sampled according to this distribution"

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,292 @@
...
+                _tensordict = update_values[..., t + 1].update(_tensordict)
+
+        return torch.stack(tensordict_out, tensordict.ndimension() - 1).contiguous()
+
+
+class RSSMPrior(nn.Module):
+    """The prior network of the RSSM.
+
+    This network takes as input the previous state and belief and the current action.
+    It outputs the next prior state and belief, as well as the parameters of the prior state distribution.
```
vmoens, line 211: outputs -> returns

vmoens: `It returns the parameters of the prior state distribution, a state tensor sampled according to this distribution and a belief tensor (a deterministic transformation of the inputs).`

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,292 @@
...
+    State is by construction stochastic and belief is deterministic. In the "Dream to control" paper, these are called "deterministic state " and "stochastic state", respectively.
+    We distinguish states sampled according prior and posterior distribution for clarity.
```
vmoens, line 213: Not true anymore (+ not necessarily clearer since we sometimes have to rename them)

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,292 @@
...
+
+    Reference: https://arxiv.org/abs/1811.04551
+
+    Args:
+        hidden_dim (int, optional): Number of hidden units in the linear network. Input size of the recurrent network.
+        rnn_hidden_dim (int, optional): Number of hidden units in the recurrent network. Also size of the belief.
+        state_dim (int, optional): Size of the state.
+        action_spec (TensorSpec, optional): Action spec. If None an error will be raised when initializing.
```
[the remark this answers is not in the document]
nicolas-dufour: What default should we have then?

test/mocking_classes.py
```diff
@@ -712,6 +712,87 @@ def _reset(self, tensordict: TensorDict, **kwargs) -> TensorDict:
         return td
 
 
+class MockPixelEnv(EnvBase):
```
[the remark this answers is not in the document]
vmoens: Can you try with `ContinuousActionConvMockEnv` instead?

test/test_cost.py
```diff
@@ -1564,6 +1584,257 @@ def test_reinforce_value_net(self, advantage, gradient_mode, delay_value):
...
+        actor_model = self._create_actor_model(10, 5).to(device)
+        value_model = self._create_value_model(10, 5).to(device)
+        loss_module = DreamerActorLoss(actor_model, value_model, mb_env).to(device)
+        loss_td, _ = loss_module(tensordict)
+        assert loss_td.get("loss_actor") is not None
+        loss = loss_td.get("loss_actor")
+        loss.backward()
+        actor_model.zero_grad()
+
+    @pytest.mark.parametrize("device", get_available_devices())
```
vmoens, line 1826: the device decorator can be moved out of the class

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+]
+
+
+class DreamerActor(nn.Module):
+    """Dreamer actor network.
+
+    This network is used to predict the action distribution given the
+    the stochastic state and the deterministic belief at the current
+    time step.
+    It output the mean and the scale of the action distribution.
```
vmoens, line 31: it outputs

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+
+    Reference: https://arxiv.org/abs/1912.016034
+
+    Args:
+        out_features (int): Number of output features.
+        depth (int, optional): Number of hidden layers.
+        num_cells (int, optional): Number of hidden units per layer.
+        activation_class (nn.Module, optional): Activation class.
+        rnn_hidden_size (int, optional): Size of the hidden state of the RNN in the RSSM module.
```
vmoens, line 40: this does not match the `__init__`

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+    ):
+        super().__init__()
+        self.backbone = NormalParamWrapper(
+            MLP(
+                out_features=2 * out_features,
+                depth=depth,
+                num_cells=num_cells,
+                activation_class=activation_class,
+            ),
+            scale_mapping="biased_softplus_5.0_1e-4",
```
vmoens, line 58: that should be parametrizable

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+
+class ObsEncoder(nn.Module):
+    """Observation encoder network.
+
+    Takes an pixel observation and encodes it into a latent space.
+
+    Reference: https://arxiv.org/abs/1803.10122
+
+    Args:
+        depth (int, optional): Number of hidden units in the first layer.
```
vmoens, line 74: default is not stated

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+    time step.
+    It output the mean and the scale of the action distribution.
+
+    Reference: https://arxiv.org/abs/1912.016034
+
+    Args:
+        out_features (int): Number of output features.
+        depth (int, optional): Number of hidden layers.
+        num_cells (int, optional): Number of hidden units per layer.
+        activation_class (nn.Module, optional): Activation class.
```
vmoens, line 39: default is not stated

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+    the stochastic state and the deterministic belief at the current
+    time step.
+    It output the mean and the scale of the action distribution.
+
+    Reference: https://arxiv.org/abs/1912.016034
+
+    Args:
+        out_features (int): Number of output features.
+        depth (int, optional): Number of hidden layers.
+        num_cells (int, optional): Number of hidden units per layer.
```
vmoens, line 38: default is not stated

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+    This network is used to predict the action distribution given the
+    the stochastic state and the deterministic belief at the current
+    time step.
+    It output the mean and the scale of the action distribution.
+
+    Reference: https://arxiv.org/abs/1912.016034
+
+    Args:
+        out_features (int): Number of output features.
+        depth (int, optional): Number of hidden layers.
```
vmoens, line 37: default is not stated

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+
+class ObsDecoder(nn.Module):
+    """Observation decoder network.
+
+    Takes the deterministic state and the stochastic belief and decodes it into a pixel observation.
+
+    Reference: https://arxiv.org/abs/1803.10122
+
+    Args:
+        depth (int, optional): Number of hidden units in the last layer.
```
vmoens, line 110: default is not stated

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+
+    Given a set of encoded observations and actions, this module will rollout the RSSM network to compute all the intermediate
+    states and beliefs.
+    The previous posterior is used as the prior for the next time step.
+    The forward method returns a stack of all intermediate states and beliefs.
+
+    Reference: https://arxiv.org/abs/1811.04551
+
+    Args:
+        rssm_prior (RSSMPrior): Prior network.
```
vmoens, line 151: I think their type is more generally a TensorDictModule

torchrl/modules/models/model_based.py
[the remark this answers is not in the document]
vmoens: No default, we should move it before the kwargs

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+    """The prior network of the RSSM.
+
+    This network takes as input the previous state and belief and the current action.
+    It returns the next prior state and belief, as well as the parameters of the prior state distribution.
+    State is by construction stochastic and belief is deterministic. In "Dream to control", these are called "deterministic state " and "stochastic state", respectively.
+
+    Reference: https://arxiv.org/abs/1811.04551
+
+    Args:
+        hidden_dim (int, optional): Number of hidden units in the linear network. Input size of the recurrent network.
```
vmoens, line 217: default is not stated

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+        rnn_hidden_dim (int, optional): Number of hidden units in the recurrent network. Also size of the belief.
```
vmoens, line 218: default is not stated

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+        state_dim (int, optional): Size of the state.
```
vmoens, line 219: default is not stated

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+        # Prior
+        self.rnn = nn.GRUCell(hidden_dim, rnn_hidden_dim)
+        self.action_state_projector = nn.Sequential(nn.LazyLinear(hidden_dim), nn.ELU())
+        self.rnn_to_prior_projector = NormalParamWrapper(
+            nn.Sequential(
+                nn.Linear(hidden_dim, hidden_dim),
+                nn.ELU(),
+                nn.Linear(hidden_dim, 2 * state_dim),
+            ),
+            scale_lb=0.1,
```
vmoens, line 238: shouldn't this be parametrizable too?

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+class RSSMPosterior(nn.Module):
+    """The posterior network of the RSSM.
+
+    This network takes as input the belief and the associated encoded observation.
+    It returns the parameters of the posterior as well as a state sampled according to this distribution.
+
+    Reference: https://arxiv.org/abs/1811.04551
+
+    Args:
+        hidden_dim (int, optional): Number of hidden units in the linear network.
```
vmoens, line 267: default is not stated

torchrl/modules/models/model_based.py
```diff
@@ -0,0 +1,291 @@
...
+        state_dim (int, optional): Size of the state.
```
vmoens, line 268: default is not stated

vmoens (approved):
Great! I think we'll be ready to merge once we figure out how to improve the tests and fix the various docstrings

examples/dreamer/dreamer.py
```diff
@@ -0,0 +1,493 @@
+import dataclasses
```
luisenp, line 1: I took a quick pass at the code, and for now i only have two high level comments. The first one is minor, but the second I think is important:

- This PR is really large, and from what I can see it has a lot of independent small features that could have been merged as separate PRs. 
- There is a lot of boilerplate code in `dreamer.py` that makes it hard to follow the logic easily. I would imagine that one of the main goals of this module is to illustrate how to use TorchRL components to write sophisticated algorithms (is this correct?). However, there is a lot of code here for setting configurations, video recording, logging, etc., all intermingled with the RL code, and I argue this reduces the mental space that a newcomer will have for recognizing the TorchRL features that are enabling this Dreamer implementation. For me, the closer this file would look to readable pseudocode, the better job it'd be doing at selling the library. 

My suggestion is then to move a lot of this boilerplate to utility modules, imported with easily interpretable function names, and also adding a lot of comments in the code guiding users through the important features used in the implementation.

vmoens (approved):
LGTM

vmoens (approved):
LGTM

PR #344, opened 2022-08-06
[Test] pass the OS error in case the file isn't closed

tongbaojia:
## Description

The `OSError` could be raised randomly depending on the CI testing machine. Causing tests not to pass.

Try again the test in case this exception is raised.


vmoens (approved):
LGTM! Thanks for this

PR #356, opened 2022-08-10
[BugFix] Add lock to vec norm transform

### LINKED ISSUE #312: [BUG] missing lock for vecnorm transform
## Describe the bug

VecNorm goes into racing issues because updates to the shared tensors are not locked.

## To Reproduce

VecNorm tests are flacky because of racing operations between workers.

## Suggested solution

A lock should be introduced in vecnorm. For context, vecnorm contains tensors that are placed in shared memory (or equivalently on cuda). To make sure that all vecnorm instances have access to the same shared tensor, we create a vecnorm copy in main using the `EnvCreator` class. This original instance [creates the buffers](https://github.com/facebookresearch/rl/blob/2f2105e2d7f7ae844b740af44a0853c5b8a1f032/torchrl/envs/transforms/transforms.py#L1824) that will store the running stats in shared memory. Then, we link each vecnorm instance on each process to the copy on main by [passing a `state_dict` to each subprocess](https://github.com/facebookresearch/rl/blob/2f2105e2d7f7ae844b740af44a0853c5b8a1f032/torchrl/envs/env_creator.py#L128) (note that [we use `get_extra_state`](https://github.com/facebookresearch/rl/blob/2f2105e2d7f7ae844b740af44a0853c5b8a1f032/torchrl/envs/transforms/transforms.py#L1888-L1896) to pass a TensorDict that contains all the shared tensors). Have a look at [VecNorm docstrings](https://github.com/facebookresearch/rl/blob/2f2105e2d7f7ae844b740af44a0853c5b8a1f032/torchrl/envs/transforms/transforms.py#L1845-L1854) and [tests](https://github.com/facebookresearch/rl/blob/2f2105e2d7f7ae844b740af44a0853c5b8a1f032/test/test_transforms.py#L186) for more info.
Ideally, the solution should pass / get the lock in `get_extra_state` and `set_extra_state`.

## Checklist

- [x] I have read the [documentation](https://github.com/facebookresearch/rl/tree/main/docs/) (**required**)

jaschmid-fb:
## Description

Implemented Lock support in VecNorm
The lock parameter is optional, to avoid overhead if no locking is needed, if provided the VecNorm will acquire it to make the updates multiprocessing safe

Updated tests:
test_vecnorm_parallel
test_vecnorm_parallel_auto
test_parallelenv_vecnorm

As lock is stored in extra state, fixed up some locations in code making assumptions of the layout/contents of the state_dicts, this code seemed internal or part of tests so I don't think the change should be breaking in general

## Motivation and Context

close #312

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -1707,6 +1707,7 @@ class VecNorm(Transform):
         self,
         keys_in: Optional[Sequence[str]] = None,
         shared_td: Optional[TensorDictBase] = None,
+        lock: mp.Lock = None,
         decay: float = 0.9999,
         eps: float = 1e-4,
     ) -> None:
@@ -1732,10 +1733,14 @@ class VecNorm(Transform):
                         f"with keys {shared_td.keys()}"
                     )
 
+        self.lock = lock
         self.decay = decay
         self.eps = eps
 
     def _call(self, tensordict: TensorDictBase) -> TensorDictBase:
+        if self.lock is not None:
+            self.lock.acquire()
+
         for key in self.keys_in:
             if key not in tensordict.keys():
                 continue
@@ -1746,6 +1751,10 @@ class VecNorm(Transform):
             )
 
             tensordict.set_(key, new_val)
+
+        if self.lock is not None:
+            self.lock.release()
+
         return tensordict
 
     def _init(self, tensordict: TensorDictBase, key: str) -> None:
@@ -1885,11 +1894,13 @@ class VecNorm(Transform):
             return td_select.memmap_()
         return td_select.share_memory_()
 
-    def get_extra_state(self) -> TensorDictBase:
-        return self._td
+    def get_extra_state(self) -> OrderedDict:
+        return OrderedDict([("lock", self.lock), ("td", self._td)])
 
-    def set_extra_state(self, td: TensorDictBase) -> None:
-        if not td.is_shared():
+    def set_extra_state(self, state: OrderedDict) -> None:
+        self.lock = state["lock"]
+        td = state["td"]
+        if td is not None and not td.is_shared():
             raise RuntimeError(
                 "Only shared tensordicts can be set in VecNorm transforms"
             )
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import multiprocessing as mp
 from copy import deepcopy, copy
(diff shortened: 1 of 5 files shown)


torchrl/envs/transforms/transforms.py
```diff
@@ -1707,6 +1708,7 @@ def __init__(
         self,
         keys_in: Optional[Sequence[str]] = None,
         shared_td: Optional[TensorDictBase] = None,
+        lock: mp.Lock = (mp.Lock()),
```
vmoens, line 1711: Can we do this instead

```python
...
    lock: Optional[mp.Lock] = None
...
if lock is None:
    lock = mp.Lock()
```

We usually try to avoid instantiating stuff in kwargs

torchrl/envs/vec_env.py
```diff
@@ -798,6 +799,19 @@ def to(self, device: DEVICE_TYPING):
         return self
 
 
+def recursively_strip_locks_from_state_dict(state_dict: OrderedDict) -> OrderedDict:
+    return OrderedDict(
```
vmoens, line 803: same: do we want to use OrderedDict?

vmoens (approved):
omg what a contribution @jachmid-fb !!
LGTM, can you have a look at the couple of comments I left?

torchrl/envs/transforms/transforms.py
```diff
@@ -1885,11 +1894,13 @@ def build_td_for_shared_vecnorm(
             return td_select.memmap_()
         return td_select.share_memory_()
 
-    def get_extra_state(self) -> TensorDictBase:
-        return self._td
+    def get_extra_state(self) -> OrderedDict:
+        return OrderedDict([("lock", self.lock), ("td", self._td)])
```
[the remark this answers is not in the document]
jaschmid-fb: There are a few places in the code where we process state_dicts, for example we call share_memory_ to make shared versions of existing state_dicts. I ran into issues since those pieces of code assumed that state_dicts were recursive OrderedDicts, so inserting a dict rather than an OrderedDict caused failures.

For example EnvCreator.share_memory

It's fixable but will expand the scope of the change, since this OrderedDict was very small this seemed easier but I can go and fix it if you want, please advise.

torchrl/envs/vec_env.py
jaschmid-fb: Same as above, existing code assumes state_dict is an OrderedDict, I have changed a few places in code that assumed this (in the tests) but there are others remaining, not sure if cleaning this up is the objective of this change, but I can do so. It might also make another good bootcamp task.

Issue #351, opened 2022-08-08
[Feature Request] Implement `__eq__` for `TensorSpec`


vmoens:
## Motivation

When comparing environments, it can be useful to assess if all specs are equal. To do this, we'd like to have an `__eq__` method for each `TensorSpec` subclass.

## Solution

The `__eq__` should return True if the class is identical, if the shape is the same and if the bounds are identical too (if any).
It should return False in all other cases.

This feature should be tested for each spec (test against another class, test against same class but different meta-data and test against same copy of the same object).

vmoens:
Closes as of #358

PR #358, opened 2022-08-10
[Feature] Implement __eq__ for TensorSpec

omikad:
## Description

This pull request has the following changes:
* Add tests to cover equality check for all the `TensorSpec` subclasses. Equality function returns True if all the fields are the same and class is the same.
* Since majority of subclasses and their dependencies are marked with `@dataclass` - there was no need to implement `__eq__` in many of them. So this PR only adds `__eq__` implementation to a few exceptions.
* Add `__eq__` to the class `ContinuousBox`. Default `__eq__` implementation doesn't work as expected for multi-dimensional tensor comparison. If you have two tensors with more than 1 value then their equality returns not a boolean, but a tensor of the same dimension, which raises "Truth value is ambiguous" error. This PR uses `torch.equal` method instead
* Add `__eq__` to the class `OneHotDiscreteTensorSpec`, since it has an additional attribute `use_register`, and it is not covered by default `__eq__` method provided by `@dataclass` attribute
* Add `__eq__` to the class `CompositeSpec` since it has no attribute `@dataclass`.

## Motivation and Context

This change solves #351 issue.

### DIFF
--- a/test/test_tensor_spec.py
+++ b/test/test_tensor_spec.py
@@ -394,116 +394,118 @@ class TestEquality:
         return ts_to
 
     def test_equality_bounded(self):
-        MIN = 10
-        MAX = 100
-        DEVICE = "cpu"
-        DTYPE = torch.float16
+        minimum = 10
+        maximum = 100
+        device = "cpu"
+        dtype = torch.float16
 
-        ts = BoundedTensorSpec(MIN, MAX, DEVICE, DTYPE)
+        ts = BoundedTensorSpec(minimum, maximum, device, dtype)
 
-        ts_same = BoundedTensorSpec(MIN, MAX, DEVICE, DTYPE)
+        ts_same = BoundedTensorSpec(minimum, maximum, device, dtype)
         assert ts == ts_same
 
-        ts_other = BoundedTensorSpec(MIN + 1, MAX, DEVICE, DTYPE)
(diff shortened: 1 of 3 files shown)


test/test_tensor_spec.py
```diff
@@ -383,6 +383,302 @@ def test_nested_composite_spec(self, is_complete, device, dtype):
...
+    def _ts_make_all_fields_equal(ts_to, ts_from):
+        ts_to.shape = ts_from.shape
+        ts_to.space = ts_from.space
+        ts_to.device = ts_from.device
+        ts_to.dtype = ts_from.dtype
+        ts_to.domain = ts_from.domain
+        return ts_to
+
+    def test_equality_bounded(self):
+        MIN = 10
```
vmoens, line 397: can we rename the caps variables in non caps?
we usually use that for global variables

vmoens (approved):
LGreatTM! 
Thanks for this, it's really useful!
Once the small formatting issue fixed we'll be ready to merge

test/test_tensor_spec.py
omikad: Sure, one moment

omikad: Done! @vmoens  please check

vmoens (approved):
Brilliant!

PR #360, opened 2022-08-11
[Feature] Support tensor-based decay in TD-lambda

### LINKED ISSUE #355: [Feature Request] Support custom, tensor-based decay in TD(lambda)
## Motivation

Currently, the vectorised TD(lambda) function only supports scalar gamma values.
However, in some cases we'll want to provide a gamma vector: for instance, the [`MultiStep`](https://github.com/facebookresearch/rl/blob/d295fae9f532862d60e28ca1f509ad1183cf6a56/torchrl/data/postprocs/postprocs.py#L109) class computes a gamma tensor of the same size as the reward that is to be used to compute decayed reward sums.

## Solution

The proposed solution is to allow [`_custom_conv1d`](https://github.com/facebookresearch/rl/blob/266ddff05ce0c31ba526c0124e0f025ff5273286/torchrl/objectives/returns/functional.py#L167) to receive filters (i.e. stack of gamma values) with a certain batch size. The conv1d operation does not support batching but we can do something like this:

```python
if filter.ndimension() > 2:  # then filter has a batch size
    filter = filter.reshape(batch_size, 1, 1, T)
    functorch.vmap(torch.conv1d, (None, 0))(val_pad, filter)  # (None, 0) tells vmap to expand val_pad and not filter
else:
   do_as_before()
```

In this implementation, `functorch` will do the vectorised map of filter over val_pad (i.e. expand val_pad and execute conv1d over it).

Naturally, the creation of the filter and the reshaping will need to be adapted accordingly.

Tests should be added to account for this feature and placed in `test_costs.py`.
For instance:

```python
batch = 10
time = 11
next_value = torch.randn(batch, time, 1)
done = torch.zeros(batch, time, 1, dtype=torch.bool)
reward = torch.randn(batch, time, 1)

gamma = 0.99
lmbda = 0.95

gamma_tensor = torch.full(gamma, (batch, time, 1))

v1 = vec_td_lambda_return_estimate(gamma, lmbda, next_state_value, reward, done)
v2 = vec_td_lambda_return_estimate(gamma_tensor, lmbda, next_state_value, reward, done)

assert_isclose(v1, v2)

# same with last done being true
done[:, -1] = True  # terminating trajectory
gamma_tensor[:, -1] = 0.0

v1 = vec_td_lambda_return_estimate(gamma, lmbda, next_state_value, reward, done)
v2 = vec_td_lambda_return_estimate(gamma_tensor, lmbda, next_state_value, reward, done)

assert_isclose(v1, v2)

```

tcbegley:
## Description

This PR adds support for tensor-based decay to the vectorised TD-lambda objective function. I've added tests as suggested in the linked issue (#355).

There is one area which I will mark with an inline comment that needs careful checking as I'm not 100% sure I haven't misunderstood.

## Motivation and Context

Fixes #355 

## Types of changes

## Checklist

- [ ] ~My change requires a change to the documentation.~
- [ ] ~I have updated the documentation accordingly.~ N/A

### DIFF
--- a/torchrl/objectives/returns/functional.py
+++ b/torchrl/objectives/returns/functional.py
@@ -205,7 +205,7 @@ def _custom_conv1d(tensor, filter):
         # rather than compute them all and discard, we stack just the slices
         # of val_pad that we care about, and apply the filter manually
         batched_val_pad = torch.stack(
-            [val_pad[..., i:i + filter.shape[-1]] for i in range(tensor.shape[-1])],
+            [val_pad[..., i : i + filter.shape[-1]] for i in range(tensor.shape[-1])],
             dim=1,
         )
 
--- a/torchrl/objectives/returns/functional.py
+++ b/torchrl/objectives/returns/functional.py
@@ -199,7 +199,7 @@ def _custom_conv1d(tensor, filter):
     if filter.ndimension() > 2:
         # filter will have shape batch_dims x timesteps x timesteps x 1
         # reshape to batch_dims x timesteps x 1 x 1 x timesteps ready for convolving
-        filter = filter.reshape(*filter.shape[:-2], 1, 1, filter.shape[-2])
+        filter = filter.view(*filter.shape[:-2], 1, 1, filter.shape[-2])
 
         # add dimension to val-pad as we are going to vectorise over the first
         # axis (batch) since we have different filter per batch
--- a/torchrl/objectives/returns/functional.py
+++ b/torchrl/objectives/returns/functional.py
@@ -296,7 +296,7 @@ def vec_td_lambda_return_estimate(gamma, lmbda, next_state_value, reward, done):
     lambdas = torch.cumprod(lambdas, -2)
 
     first_below_thr = gammas < 1e-7
-    if gammas.ndimension() > 2:
+    while first_below_thr.ndimension() > 2:
         # if we have multiple gammas, we only want to truncate if _all_ of
         # the geometric sequences fall below the threshold
         first_below_thr = first_below_thr.all(axis=0)
--- a/torchrl/objectives/returns/functional.py
+++ b/torchrl/objectives/returns/functional.py
@@ -5,7 +5,6 @@
 
 from typing import Tuple
 
-import functorch
 import torch
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


torchrl/objectives/returns/functional.py
```diff
@@ -195,9 +196,31 @@ def _custom_conv1d(tensor, filter):
...
+        # TODO: there's a lot of redundancy here since we only care about the
+        # diagonal entries. how to simplify?
+        conv1d_batch = functorch.vmap(torch.conv1d, (None, 0))
+
+        # vectorise over batch dimension
+        out = functorch.vmap(conv1d_batch)(val_pad, filter)
+
+        # out has dimensions batch x timesteps x 1 x 1 x timesteps
+        # we only care about the diagonal timestep entries
+        out = torch.diagonal(out, dim1=-4, dim2=-1).squeeze(-3)
```
tcbegley, line 219: This section, which I've tried to add plenty of comments to, is where I would particularly like some guidance as to whether I have got things right. Even if this is right, it seems like there's some redundancy which could likely be made more efficient

Suppose this has been called from line 314 `rs = _custom_conv1d(reward, gammas * lambdas)`. Thus `tensor` has dimensions `batch x 1 x timesteps` and `filter` has dimensions `batch x timesteps x timesteps x 1` (before reshaping). There are two dimensions length `timesteps` because we have one `gamma` per batch and timestep (call it $\gamma_{bt}$), and each of those we calculate the decay sequence $\gamma_{bt}, \gamma^2_{bt}, \dots$.

The return value has the same dimensions as `tensor` (`batch x 1 x timesteps`). My understanding is that it should contain values like

$$
    \text{out}[b, 0, t] = r_t + \lambda \gamma_{bt} r_{t+1} + \lambda^2\gamma^2_{bt}r_{t+2} + \dots
$$

These can still be calculated using a convolution, but notice that because `t` appears both in the gamma index and the offset of the rewards, we only actually care about the diagonal values of the full vectorised convolution.

Does that seem right? If so I did think for a while about whether it could be done more directly but didn't solve it yet.

If it doesn't seem right, pointers are appreciated!

torchrl/objectives/returns/functional.py
```diff
@@ -195,9 +196,31 @@ def _custom_conv1d(tensor, filter):
         -1,
     )
 
-    # shape = val.shape
-    filter = filter.squeeze(-1).unsqueeze(0).unsqueeze(0)  # 1 x 1 x T
-    out = torch.conv1d(val_pad, filter)
+    if filter.ndimension() > 2:
+        # filter will have shape batch_dims x timesteps x timesteps x 1
+        # reshape to batch_dims x timesteps x 1 x 1 x timesteps ready for convolving
+        filter = filter.reshape(*filter.shape[:-2], 1, 1, filter.shape[-2])
```
vmoens, line 202: we could use view instead (perhaps?)

vmoens (commented):
Let's keep iterating on this, I think we're on the right track!

torchrl/objectives/returns/functional.py
tcbegley: Good idea, have substituted.

torchrl/objectives/returns/functional.py
```diff
@@ -195,9 +196,31 @@ def _custom_conv1d(tensor, filter):
...
+
+        # add dimension to val-pad as we are going to vectorise over the first
+        # axis (batch) since we have different filter per batch
+        val_pad = val_pad.unsqueeze(1)
```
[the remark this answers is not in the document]
tcbegley: If I remove this I get `RuntimeError: wrong output shape` from [here](https://github.com/facebookresearch/rl/blob/da3e63d522c73ed7226912a18a3c5c1abeadd450/torchrl/objectives/returns/functional.py#L225-L226) 😕 

I will leave it for now as it may in any case become redundant when the other logic changes.

torchrl/objectives/returns/functional.py
tcbegley: This runs for me and gives the expected shape. However I'm a bit unclear on the inputs. We are allowing the user to specify a `gamma` per batch _and_ per timestep right? So would that not imply that `filter` should have dimensions for batch and timesteps (corresponding to the different `gamma`s) _and_ another dimension for the geometric decay over subsequent timesteps? Or have I misunderstood that?

The above is the assumption I made in my first pass, which means I'm not sure how to compare my implementation to your snippet?

torchrl/objectives/returns/functional.py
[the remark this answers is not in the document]
tcbegley: This is no longer relevant with switch to `einsum`

torchrl/objectives/returns/functional.py
tcbegley: This discussion no longer relevant after switch to `einsum` which takes care of the redundancy.

vmoens (approved):
This PR went above everything we could have wished for!
Thanks for working on this.

PR #364, opened 2022-08-13
[POC] Using TensorDict for functorch

vmoens:
## Description

We propose a notebook where the swapping and storage of parameters is achieved via TensorDict.
(ipynb is a fairly bad way of doing review, any idea to improve the format of the POC is welcome)

### DIFF
--- a/_dump/tensordict_functorch.ipynb
+++ b/_dump/tensordict_functorch.ipynb
@@ -4,11 +4,7 @@
    "cell_type": "code",
    "execution_count": 1,
    "id": "c675b892-6412-4758-bd05-c6919740124f",
-   "metadata": {
-    "pycharm": {
-     "name": "#%%\n"
-    }
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "from torchrl.data import TensorDict\n",
@@ -25,11 +21,7 @@
    "cell_type": "code",
    "execution_count": 2,
    "id": "996c25c2-bb72-447b-b2e8-25e0367094c1",
-   "metadata": {
-    "pycharm": {
-     "name": "#%%\n"
-    }
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "from functorch._src.vmap import _add_batch_dim, tree_unflatten, tree_flatten"
@@ -39,11 +31,7 @@
    "cell_type": "code",
    "execution_count": 3,
    "id": "a1c4caac-3160-4a06-a0d5-8ece79faabe0",
-   "metadata": {
-    "pycharm": {
-     "name": "#%%\n"
-    }
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "class FunctionalModule(nn.Module):\n",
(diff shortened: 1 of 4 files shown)


zou3519:
cc @samdow as well

zou3519:
> (ipynb is a fairly bad way of doing review, any idea to improve the format of the POC is welcome)

I'm going through the notebook now. I wish there a way to comment on specific parts of it. Is it possible to download the ipynb somehow as a .py file and have it inline the output as comments?

vmoens:
Yes, Let me do that!

vmoens:
@zou3519 I'm curious about your thoughts about this.

The fact that we can stack tensordicts together makes it easy for us in RL to use the same functional model across multiple tasks, which is pretty cool for multi-task or multi-agent learning. It would greatly simplify our API: here is a sketch where I create a common backbone, but then the two different tasks both have a separate sets of weights. 

```python
# we have 2 tasks: stand and walk. The input is still a state of the same dimension, and the actions have the same size too.
# we'd like to have a policy for each with a common backbone between the two.

env1 = lambda: DMControlEnv("humanoid", "walk")
env2 = lambda: DMControlEnv("humanoid", "stand")

# run those two envs in parallel
env = ParallelEnv(2, [env1, env2])

# policy: backbone + 2 task-specific networks
backbone = nn.LazyLinear(12)  # hidden state of size 12
net1 = nn.Linear(12, env.action_spec.shape[-1])
net2 = nn.Linear(12, env.action_spec.shape[-1])

# make all of this functional
net_total = nn.Sequential(backbone, net1)
fnet, _ = make_functional(net_total)
_, params_backbone = make_functional(backbone)
_, params_net1 = make_functional(net1)
_, params_net2 = make_functional(net1)

optimizer = torch.optim.Adam(list(params_backbone.values()) + list(params_net1.values()) + list(params_net2.values()), lr=lr)

# building the parameters is easy: the backbone just needs to be expanded and the task-specific parameters just need to be stacked
params_tensordict = TensorDict(
    {"0": params_backbone.expand(2),  # these parameters are shared for both tasks
    "1": torch.stack([params_net1, params_net2], 0)}, # there params are not shared
    batch_size=[2]
)
policy = TensorDictModule(fnet, in_keys=["observation"], "out_keys"=["action"])

# test
tensordict = env.reset()
tensordict = env.step(policy(tensordict), vmap=True)

```
The same code using lists would be quite burdensome IMO. Note that the `optimizer` would work fine even if we have a `stack` operation in the input tensordict as stacking TensorDicts together is done in a Lazy fashion (i.e. the stack only occurs when `stacked_tensordict.get(key)` is called).

I can update this PR to import that functionality in TorchRL as a temporary patch to our library.

I'd be glad to have your perspective on this. I guess that from ours it would be convenient to have such functionality but I don't want TorchRL to "sort of" depend on functorch with a bunch of monkey-patched version of its functionalities either.

(As a side note, many have pointed that TensorDict is not an RL-specific feature and we should probably consider moving it to a separate repo (a bit like masked tensors), but from there to have functorch depend on it there is still a long way.)

vmoens:
Closing as of #387

PR #365, opened 2022-08-17
[Feature] Vectorized GAE

vmoens:
## Description

Implements a vectorized version of GAE.

cc @tcbegley


test/test_cost.py
```diff
@@ -1713,6 +1715,28 @@ def test_tdlambda(device, gamma, lmbda, N, T):
...
+    state_value = torch.randn(*N, T, 1, device=device)
+    next_state_value = torch.randn(*N, T, 1, device=device)
+
+    r1 = vec_generalized_advantage_estimate(
+        gamma, lmbda, state_value, next_state_value, reward, done
+    )
+    r2 = generalized_advantage_estimate(
+        gamma, lmbda, state_value, next_state_value, reward, done
+    )
+    torch.testing.assert_close(r1, r2, rtol=1e-4, atol=1e-4)
```
Benjamin-eecs, line 1737: wondering the atol and rtol used in here is the standard used in tests of torch.

vmoens: It's a good point, we could default to the same as everywhere else.
Thing is that we truncate the computation when gamma^n gives a negligible number: this may slightly skew the results (hence the loose tolerance for the tests).

vmoens:
blocked by #401 and #400

Issue #367, opened 2022-08-18
[BUG] TensorDict print output seems to be incorrect


ezhang7423:
## Describe the bug

When creating a nested tensordict, indexing into the larger dict prints the wrong shape.
## To Reproduce

Steps to reproduce the behavior.

```
 from torchrl.data import TensorDict

z = TensorDict({str(k): TensorDict({k: torch.arange(300).reshape(150, 2) for k in string.ascii_lowercase}, batch_size=[150, 2]) for k in [0, 1, 2, 3, 4]}, batch_size=[150])`
print(z[0])
TensorDict(
    fields={
        0: TensorDict(
            fields={
                a: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                b: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                c: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                d: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                e: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                f: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                g: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                h: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                i: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                j: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                k: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                l: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                m: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                n: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                o: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                p: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                q: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                r: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                s: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                t: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
                u: Tensor(torch.Size([150, 2, 1]), dtype=torch.int64),
[...]

PR #370, opened 2022-08-19
[BugFix] Representation of indexed nested tensordict

vmoens:
## Description

Fix for #367 
This code snippet should display tensors of size [2, 1] and not [150, 2, 1]:

```python
from torchrl.data import TensorDict
import torch
import string
z = TensorDict({str(k): TensorDict({k: torch.arange(300).reshape(150, 2) for k in string.ascii_lowercase}, batch_size=[150, 2]) for k in [0, 1, 2, 3, 4]}, batch_size=[150])
print(z[0])
```

TODO:
- [ ] Check other items_meta() in the code and make sure they are called on restriced occasions (expensive to create MetaTensor)
- [ ] Write tests for tensordict representation (#380)

cc @ezhang7423

### DIFF
--- a/torchrl/data/tensordict/tensordict.py
+++ b/torchrl/data/tensordict/tensordict.py
@@ -1542,7 +1542,11 @@ dtype=torch.float32)},
         # if return_simple_view and not self.is_memmap():
         return TensorDict(
             source={key: item[idx] for key, item in self.items()},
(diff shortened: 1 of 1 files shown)


ezhang7423:
Thank you!! This is awesome.

PR #384, opened 2022-08-30
[Feature]: Support for planners and CEM

nicolas-dufour:
## Description

This PR adds support for MPC planners and in particular the CEM planner (The cross-entropy method for optimization, Botev et al 2013)

## Motivation and Context

Planners are a key component of RL and is needed for methods such as PlaNet and PETS.

## Types of changes

## Checklist


torchrl/modules/planners/mpc.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        Perform the MPC planning step.
+        Args:
+            tensordict (TensorDict): The TensorDict to perform the planning step on.
+        Returns:
+            TensorDict: The TensorDict with the action added.
+        """
+        raise NotImplementedError()
+
+    def forward(
```
vmoens, line 53: why can't we use the plain forward here?
There's little interest in inheriting from TensorDictModule if we're rewriting fundamental methods.
I thought re-writing _call_module only was sufficient?
(if it's for the functionality side of things, I guess we can just raise errors if vmap or params are passed?)

torchrl/modules/planners/mpc.py
```diff
@@ -0,0 +1,186 @@
...
+            tensordict,
+            tensors,
+            tensordict_out,
+        )
+        return tensordict_out
+
+
+class CEMPlanner(MPCPlannerBase):
+    """
+    CEMPlanner Module. This class inherits from TensorDictModule.
```
vmoens, line 73: after the one-liner presentation, always leave a blank line. Otherwise the docs are kinda messed up

torchrl/modules/planners/mpc.py
```diff
@@ -0,0 +1,186 @@
...
+
+import torch
+from torchrl.data.tensordict.tensordict import TensorDictBase
+from torchrl.envs import EnvBase
+from torchrl.modules import TensorDictModule
+
+
+class MPCPlannerBase(TensorDictModule, metaclass=abc.ABCMeta):
+    """
+    MPCPlannerBase Module. This class inherits from TensorDictModule. This is an abstract class and must be implemented by the user.
```
vmoens, line 16: let's add a blank line after the concise, one liner presentation of the module.
Do we need to mention what this class inherits from here? 
Perhaps we could talk about it later if it's not a core feature of it.

torchrl/modules/planners/mpc.py
```diff
@@ -0,0 +1,186 @@
...
+                flatten_batch_size * self.num_candidates,
+                self.planning_horizon,
+                *self.action_spec.shape,
+            )
+            optim_td = expanded_original_td.clone()
+            policy = PrecomputedActionsSequentialSetter(actions)
+            optim_td = self.env.rollout(
+                max_steps=self.planning_horizon,
+                policy=policy,
+                auto_reset=False,
```
vmoens, line 150: don't we want to reset the env?
Otherwise we'll keep on interating like nothing happened and the actions at time t won't match from one iter of the loop to the other

vmoens (commented):
Great feature, I love it
LGTM for the implementation, just a few comments before we can land this.

The logic seems to make sense. I'd like to see a test of this too, across mocking and real envs if possible (can be done with SerialEnv, ParallelEnv is paradoxically slower with tests because it takes some time to launch).

torchrl/modules/planners/mpc.py
nicolas-dufour: The problem is that the original __call_modules takes as input tensors and ours takes tensordicts. Do we don't want to perform ```tensors = tuple(tensordict.get(in_key, None) for in_key in self.in_keys)```

torchrl/modules/planners/mpc.py
nicolas-dufour: We don 't want to reset it since we want the rollout to happen at the given observation state that we are optimizing for. However you are right that after one optim step we need to reset it to the state at the time we optimize for... This is a complex one. It is trivial for MB env since they are stateless but this is a big problem for gym envs that are stateful... Maybe we should just make a copy of the env and reset env to that copy after the rollout?

torchrl/modules/planners/mpc.py
```diff
@@ -0,0 +1,186 @@
...
+        self.planning_horizon = planning_horizon
+        self.optim_steps = optim_steps
+        self.num_candidates = num_candidates
+        self.num_top_k_candidates = num_top_k_candidates
+        self.reward_key = reward_key
+
+    def planning(self, tensordict: TensorDictBase) -> torch.Tensor:
+        batch_size = tensordict.batch_size
+        expanded_original_td = (
+            tensordict.clone().expand(*batch_size, self.num_candidates).flatten()
```
vmoens, line 115: why do we clone here?

torchrl/modules/planners/mpc.py
vmoens: > We don 't want to reset it since we want the rollout to happen at the given observation state that we are optimizing for.

So what I get from your comment is this (correct me if i'm wrong)
you have an env at state _T_, and you want to plan the action from _T_ to _T_ + _t_.
To do that you run this optimization loop several times.
If you reset, you're set back at state _0_ < _T_, which you don't want. Resetting is not an option, but you'd like to reset at time _T_, right?

This can't be done with gym and it's not a use case that is supported by the library. The assumption of gym envs is that you can't reset states where you want them to be (at least not natively). You should not do that, and each env will have its own way of preventing or allowing you to do that, i.e. there is no generic way of setting a state in gym.

What we should do is raise an error if the environment is stateful. This can be checked by assessing if the `input_spec` keys match the `observation_spec` keys (without the `"next_" prefixes). If they don't you can assume that the env is stateful (bc it outputs stuff that was not its input).

Or we could add a flag `is_stateful` to the envs, but if you're doing that I would suggest to write that in a separate PR.

torchrl/modules/planners/mpc.py
vmoens: Ok, perhaps we don't need the _call_module then? Why do we need this?
```python
    def _call_module(self, tensordict: TensorDictBase) -> torch.Tensor:
        return self.planning(tensordict)
```

torchrl/modules/planners/mpc.py
```diff
@@ -0,0 +1,186 @@
...
+        tensors = (action,)
+        tensordict_out = self._write_to_tensordict(
+            tensordict,
+            tensors,
+            tensordict_out,
+        )
+        return tensordict_out
+
+
+class CEMPlanner(MPCPlannerBase):
```
vmoens, line 71: Are we 100% sure we need the super class here?
I know it's good practice to do that, but I just want to make sure that we're not opening the door to a highly hierarchical and intricate set of MPC classes.
Are we confident that a new MPC class would be codable using this structure?

nicolas-dufour: Maybe instead of MPCPlanner we could just call it PlannerBase? So that it's not tied to MPC planning methods??

torchrl/modules/planners/mpc.py
nicolas-dufour: Your right it's not necessary

torchrl/modules/planners/mpc.py
nicolas-dufour: > So what I get from your comment is this (correct me if i'm wrong)
you have an env at state T, and you want to plan the action from T to T + t.
To do that you run this optimization loop several times.
If you reset, you're set back at state 0 < T, which you don't want. Resetting is not an option, but you'd like to reset at time T, right?

Its almost that, we only optimize for action T+1 but we look in the future until action T+t

Ok i see so, we should accept that this planning algorithm won't work on gym like envs then. I don't really get the `input_spec` `observation_spec` trick. Like for example in MB_envs the keys won't match (input spec will have actions and observation will not!

torchrl/modules/planners/mpc.py
vmoens: Sure why not
Does it have enough functionality such that it's not trivial, and not too much such that it's a good abstraction?
Just want to make sure we're not creating an abstract method for the sake of it

torchrl/modules/planners/mpc.py
vmoens: yeah that's a detail, you can remove action (it'll always be there)
In general that will work. If it doesn't someome will raise an issue and we'll deal with it

test/test_modules.py
```diff
@@ -299,6 +301,33 @@ def test_lstm_net_nobatch(device, out_features, hidden_size):
     torch.testing.assert_close(tds_vec["hidden1_out"][-1], tds_loop["hidden1_out"][-1])
 
 
+class TestPlanner:
+    @pytest.mark.skipif(not _has_gym, reason="no gym")
+    @pytest.mark.parametrize("device", get_available_devices())
+    @pytest.mark.parametrize("env_name", ["Pendulum-v1"])
+    def test_CEM_model_free_env(self, device, env_name, seed=0):
+        env = GymEnv(env_name, frame_skip=1, device=device)
```
vmoens, line 309: can't we use our mocking envs? They're stateless, + gym should fail with this (and it restricts the number of tests where this will run)

torchrl/modules/planners/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from .common import *
+from .cem import *
```
vmoens, line 7: missing empty line
you probably forgot to run pre-commit before pushing

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,133 @@
...
+                device=td.device,
+                dtype=self.env.action_spec.dtype,
+            )
+            actions = actions.view(
+                flatten_batch_size * self.num_candidates,
+                self.planning_horizon,
+                *self.action_spec.shape,
+            )
+            actions = self.env.action_spec.project(actions)
+            optim_td = expanded_original_td.clone()
```
vmoens, line 92: I guess that here you want to_tensordict instead?
clone for tensordicts doesn't work the way it does for torch.Tensor (it doesn't always materialize the lazy operations etc)

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,133 @@
...
+        for _ in range(self.optim_steps):
+            actions = actions_means + actions_stds * torch.randn(
+                flatten_batch_size,
+                self.num_candidates,
+                self.planning_horizon,
+                *self.action_spec.shape,
+                device=td.device,
+                dtype=self.env.action_spec.dtype,
+            )
+            actions = actions.view(
```
vmoens, line 86: I think that here you could use flatten instead, which would be clearer

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,133 @@
...
+            optim_td = self.env.rollout(
+                max_steps=self.planning_horizon,
+                policy=policy,
+                auto_reset=False,
+                tensordict=optim_td,
+            )
+            rewards = (
+                optim_td.get(self.reward_key)
+                .sum(dim=1)
+                .reshape(flatten_batch_size, self.num_candidates)
```
vmoens, line 103: I guess that before the reshape your rewards have shape
[flatten_batch_size, num_candidates, 1]
hence you could just squeeze the last dim

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,133 @@
...
+            )
+            _, top_k = rewards.topk(self.num_top_k_candidates, dim=1)
+            top_k += (
+                torch.arange(0, flatten_batch_size, device=td.device).unsqueeze(
```
vmoens, line 107: this can be pre-computed out of the loop

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,133 @@
...
+                    1
+                )
+                * self.num_candidates
+            )
+            best_actions = actions.view(
```
vmoens, line 112: do we need to view, or can't we just use squeeze here too?

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,133 @@
...
+                flatten_batch_size,
+                self.num_candidates,
+                self.planning_horizon,
+                *self.action_spec.shape,
+            )[torch.arange(flatten_batch_size), top_k]
+            actions_means = best_actions.mean(dim=1, keepdim=True)
+            actions_stds = best_actions.std(dim=1, keepdim=True)
+        return (actions_means[:, :, 0]).view(*batch_size, *self.action_spec.shape)
```
vmoens, line 120: I would use unflatten here, instead of view

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,67 @@
...
+            raise ValueError("params not supported")
+        action = self.planning(tensordict)
+        action  = self.action_spec.project(action)
+        tensors = (action,)
+        tensordict_out = self._write_to_tensordict(
+            tensordict,
+            tensors,
+            tensordict_out,
+        )
+        return tensordict_out
```
vmoens, line 67: missing empty line

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,67 @@
...
+        self,
+        tensordict: TensorDictBase,
+        tensordict_out: Optional[TensorDictBase] = None,
+        **kwargs,
+    ) -> TensorDictBase:
+        if "params" in kwargs or "vmap" in kwargs:
+            raise ValueError("params not supported")
+        action = self.planning(tensordict)
+        action  = self.action_spec.project(action)
+        tensors = (action,)
```
vmoens, line 61: I think we can directly pass (actions,) to the method below

torchrl/modules/planners/mpc.py
nicolas-dufour: Yeah i think so. Unless at some point a TensorDictModule variant supports td -> tensor module functions, i think it's useful for planners to have this abstraction.

test/test_modules.py
nicolas-dufour: Yes this was a placeholder that wasn't working anyway. I will use the mocking env instead. Once MBEnv is merged into main will also add one with DummyMBEnv

torchrl/modules/planners/cem.py
nicolas-dufour: No, it actually as shape [flatten_batch_size * self.num_candidates, 1] that' s why we reshape

torchrl/modules/planners/cem.py
nicolas-dufour: view is necessary since actions where of shape [flatten_batch_size*self.num_candidates, self.planning_horizon,     *self.action_spec.shape]

torchrl/modules/planners/cem.py
nicolas-dufour: We actually need view since we are unflattening but also unsqueezing

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,130 @@
...
+from torchrl.data.tensordict.tensordict import TensorDictBase
+from torchrl.envs import EnvBase
+from torchrl.modules.planners import MPCPlannerBase
+
+__all__ = ["CEMPlanner"]
+
+
+class CEMPlanner(MPCPlannerBase):
+    """
+    CEMPlanner Module. This class inherits from TensorDictModule.
```
vmoens, line 17: this should be 

```
"""CEMPlanner Module. This class inherits from TensorDictModule.

etc
"""
```

Also I would not say in the first line that it inherits from `TensorDictModule`, is it so important that it must be in the headline?

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Provided a TensorDict, this module will perform a CEM planning step.
```
vmoens, line 19: Let's flip this

`This module will perform a CEM planning step when given a TensorDict containing XX and YY`

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,130 @@
...
+    The CEM planning step is performed by sampling actions from a Gaussian distribution with zero mean and unit variance.
+    The actions are then used to perform a rollout in the environment.
+    The rewards are then used to update the mean and standard deviation of the Gaussian distribution.
```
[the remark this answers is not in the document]
vmoens: How do you compute the mean and sd?
(I guess you take the top-k?)
What about "the top-K rewards indicate the indices of the actions to be used to compute a mean and sd of the corresponding actions"

torchrl/modules/planners/cem.py
vmoens: Missing ref (if you have one)

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,130 @@
...
+    The mean and standard deviation of the Gaussian distribution are then used to sample actions for the next planning step.
+    The CEM planning step is repeated for a specified number of steps.
+    At the end, we recover the best action which is the one that maximizes the reward given a planning horizon.
+
+    Args:
+        env (Environment): The environment to perform the planning step on (Can be ModelBasedEnv or EnvBase).
```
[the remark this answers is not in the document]
vmoens: `ModelBasedEnv` is and `EnvBase` no?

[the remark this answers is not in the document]
vmoens: Let's put "`" around the class names

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,130 @@
...
+        planning_horizon (int): The number of steps to perform the planning step for.
+        optim_steps (int): The number of steps to perform the MPC planning step for.
+        num_candidates (int): The number of candidates to sample from the Gaussian distribution.
+        num_top_k_candidates (int): The number of top candidates to use to update the mean and standard deviation of the Gaussian distribution.
```
vmoens, line 32: don't we want a default value for this and num_candidates (the answer can be no :) )

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,130 @@
...
+        reward_key (str): The key in the TensorDict to use to retrieve the reward.
+        action_key (str): The key in the TensorDict to use to store the action.
+
+    Returns:
+        TensorDict: The TensorDict with the action added.
```
vmoens, line 37: Returns: a TensorDict with a new "action" key representing XX

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,130 @@
...
+        action_key: str = "action",
+    ):
+        super().__init__(env=env, action_key=action_key)
+        self.planning_horizon = planning_horizon
+        self.optim_steps = optim_steps
+        self.num_candidates = num_candidates
+        self.num_top_k_candidates = num_top_k_candidates
+        self.reward_key = reward_key
+
+    def planning(self, td: TensorDictBase) -> torch.Tensor:
```
vmoens, line 57: let's avoid td whenever possible, it's confusing for the RL community

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,130 @@
...
+        batch_size = td.batch_size
+        expanded_original_td = (
+            td.unsqueeze(-1)
```
vmoens, line 60: why not 
```
expanded_original_td = td.unsqueeze(-1).expand(*batch_size, self.num_candidates).view(-1).to_tensordict()
```

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,130 @@
...
+            .contiguous()
+            .view(-1)
+        )
+        flatten_batch_size = batch_size.numel()
+        actions_means = torch.zeros(
+            flatten_batch_size,
+            1,
+            self.planning_horizon,
+            *self.action_spec.shape,
+            device=td.device,
```
vmoens, line 71: FYI
Soon we'll have tensordicts with device=None
No need to change anything here but keep it in mind when you call tensordict.device

torchrl/modules/planners/cem.py
vmoens, line 57: Do we want to prop gradients here, or do we want to set a `no_grad` decorator?

torchrl/modules/planners/cem.py
```diff
@@ -0,0 +1,130 @@
...
+
+            best_actions = actions.unflatten(
+                0, (flatten_batch_size, self.num_candidates)
+            )
+            best_actions = best_actions[
+                torch.arange(flatten_batch_size).unsqueeze(1), top_k
+            ]
+            actions_means = best_actions.mean(dim=1, keepdim=True)
+            actions_stds = best_actions.std(dim=1, keepdim=True)
+        return (actions_means[:, :, 0]).reshape(*batch_size, *self.action_spec.shape)
```
vmoens, line 117: why the parenthesis?

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,72 @@
...
+from torchrl.data.tensordict.tensordict import TensorDictBase
+from torchrl.envs import EnvBase
+from torchrl.modules import TensorDictModule
+
+__all__ = ["MPCPlannerBase"]
+
+
+class MPCPlannerBase(TensorDictModule, metaclass=abc.ABCMeta):
+    """
+    MPCPlannerBase Module. This is an abstract class and must be implemented by the user.
```
vmoens, line 19: you can just say it is an abstract class.
Also I would remove this from the headline. Or just say "MPCPlannerBase base module" which says it all

vmoens: Don't forget: no new line between `"""` and the next word (i'm in the process of correcting the whole lib for these so I'd appreciate if new PRs don't have that)

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,72 @@
...
+
+    This class inherits from TensorDictModule. Provided a TensorDict, this module will perform a Model Predictive Control (MPC) planning step.
+    At the end of the planning step, the MPCPlanner will return the action that should be taken.
```
vmoens, line 22: will return a proposed action? or smth like this?

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Args:
+        env (Environment): The environment to perform the planning step on (Can be ModelBasedEnv or EnvBase).
```
vmoens, line 25: What's an Environment here?
Don't forget the "`" when you're pointing to existing classes

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,72 @@
...
+        action_key (str): The key in the TensorDict to use to store the action.
```
vmoens, line 26: if there's a default value in torchrl we consider this as an optional arg

vmoens: the key that will point to the computed action?

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Returns:
+        TensorDict: The TensorDict with the action added.
```
vmoens, line 29: outputs are unnamed in python

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    def __init__(
+        self,
+        env: EnvBase,
+        action_key: str = "action",
+    ):
+        # Check if env is stateless
+        if env.batch_locked:
+            raise ValueError("Environment is not stateless")
```
vmoens, line 39: Let's give a more detailed error here, like why is that a problem?

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,72 @@
...
+        out_keys = [action_key]
+        in_keys = list(env.observation_spec.keys())
+        super().__init__(env, in_keys=in_keys, out_keys=out_keys)
+        self.env = env
+        self.action_spec = env.action_spec
+
+    @abc.abstractmethod
+    def planning(self, td: TensorDictBase) -> torch.Tensor:
+        """
+        Perform the MPC planning step.
```
vmoens, line 49: No new line before but empty line after

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,72 @@
...
+        Args:
+            td (TensorDict): The TensorDict to perform the planning step on.
+        Returns:
```
vmoens, line 52: missing new line

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,72 @@
...
+            TensorDict: The TensorDict with the action added.
```
vmoens, line 53: outputs are unnamed

torchrl/modules/planners/common.py
```diff
@@ -0,0 +1,72 @@
...
+        raise NotImplementedError()
+
+    def forward(
+        self,
+        tensordict: TensorDictBase,
+        tensordict_out: Optional[TensorDictBase] = None,
+        **kwargs,
+    ) -> TensorDictBase:
+        if "params" in kwargs or "vmap" in kwargs:
+            raise ValueError("params not supported")
```
vmoens, line 64: let's give a better error here

test/test_modules.py
```diff
@@ -326,6 +328,31 @@ def test_func_transformer(self):
         torch.testing.assert_close(fmodule(params, buffers, x, x), module(x, x))
 
 
+class TestPlanner:
+    @pytest.mark.parametrize("device", get_available_devices())
+    @pytest.mark.parametrize("batch_size", [3, 5])
+    def test_CEM_model_free_env(self, device, batch_size, seed=1):
+        env = MockBatchedUnLockedEnv(device=device)
+        env.set_seed(seed)
```
vmoens, line 336: torch.manual_seed(seed)

test/test_modules.py
```diff
@@ -326,6 +328,31 @@ def test_func_transformer(self):
...
+        td = env.reset(TensorDict({}, batch_size=batch_size)).to(device)
+        td_copy = td.clone()
+        td = planner(td)
+        assert td.get("action").shape[1:] == env.action_spec.shape
+
+        assert env.action_spec.is_in(td.get("action"))
+
+        for key in td.keys():
+            if key != "action":
+                assert torch.allclose(td[key], td_copy[key])
```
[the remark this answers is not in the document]
nicolas-dufour: what should i test them for? I'm already testing that they exit and have the right shape but i don't really have any information that could allow to test for the values

torchrl/modules/planners/cem.py
nicolas-dufour: no, unless we want to use PlaNet defaults

torchrl/modules/planners/cem.py
```diff
@@ -13,28 +13,26 @@
...
-    At the end, we recover the best action which is the one that maximizes the reward given a planning horizon.
+
+    A call to the module returns the actions that empirically maximised the returns given a planning horizon
 
     Args:
-        env (Environment): The environment to perform the planning step on (Can be ModelBasedEnv or EnvBase).
-        planning_horizon (int): The number of steps to perform the planning step for.
-        optim_steps (int): The number of steps to perform the MPC planning step for.
-        num_candidates (int): The number of candidates to sample from the Gaussian distribution.
+        env (EnvBase): The environment to perform the planning step on (can be ```ModelBasedEnv``` or ```EnvBase```).
```
vmoens, line 29: not sure we want three ` here, that will create a block no? Just one would suffice I think

torchrl/modules/planners/common.py
```diff
@@ -15,18 +15,16 @@
...
 class MPCPlannerBase(TensorDictModule, metaclass=abc.ABCMeta):
-    """
-    MPCPlannerBase Module. This is an abstract class and must be implemented by the user.
+    """MPCPlannerBase Module.
+    
+    This is an abstract class.
 
     This class inherits from TensorDictModule. Provided a TensorDict, this module will perform a Model Predictive Control (MPC) planning step.
-    At the end of the planning step, the MPCPlanner will return the action that should be taken.
+    At the end of the planning step, the MPCPlanner will return a proposed action
```
vmoens, line 23: a period is missing here

torchrl/modules/planners/common.py
```diff
@@ -15,18 +15,16 @@
...
 
     Args:
-        env (Environment): The environment to perform the planning step on (Can be ModelBasedEnv or EnvBase).
-        action_key (str): The key in the TensorDict to use to store the action.
-
-    Returns:
-        TensorDict: The TensorDict with the action added.
+        env (EnvBase): The environment to perform the planning step on (Can be ModelBasedEnv or EnvBase).
+        action_key (str, optional): The key that will point to the computed action
```
vmoens, line 27: period missing

torchrl/modules/planners/common.py
```diff
@@ -36,21 +34,20 @@ def __init__(
...
         super().__init__(env, in_keys=in_keys, out_keys=out_keys)
         self.env = env
         self.action_spec = env.action_spec
+        self.to(env.device)
 
     @abc.abstractmethod
     def planning(self, td: TensorDictBase) -> torch.Tensor:
-        """
-        Perform the MPC planning step.
+        """Perform the MPC planning step.
```
vmoens, line 47: Performs

torchrl/modules/planners/common.py
```diff
@@ -61,7 +58,7 @@ def forward(
         **kwargs,
     ) -> TensorDictBase:
         if "params" in kwargs or "vmap" in kwargs:
-            raise ValueError("params not supported")
+            raise ValueError("MPCPlannerBase does not support params or vmap for now.")
```
vmoens, line 61: MPCPlannerBase does not currently support functional programming

PR #385, opened 2022-08-31
[Feature] static_seed flag for envs, vectorized envs and collectors

vmoens:
## Description

Allows to seed multiple environments with the same seed.
This is interesting in contexts where we need to test multiple strategies across the same init (e.g. meta-RL and MCP planner)

cc @Benjamin-eecs @nicolas-dufour


Benjamin-eecs (approved):
LGTM! I think flag `static_seed` is great

PR #386, opened 2022-08-31
[Feature] Make torchrl runnable without functorch and with gym==0.13

vmoens:
## Description

Ensures backward compatibility with torch 1.10.0 and gym 0.13.
Motivated by projects that work with [polymetis](https://github.com/facebookresearch/fairo/tree/main/polymetis).


torchrl/objectives/common.py
```diff
@@ -55,6 +70,31 @@ def convert_to_functional(
         expand_dim: Optional[int] = None,
         create_target_params: bool = False,
         compare_against: Optional[List[Parameter]] = None,
+    ) -> None:
+        if _has_functorch:
```
vmoens, line 74: let's add an arg to give the user control over that

torchrl/modules/tensordict_module/common.py
```diff
@@ -235,6 +247,14 @@ def __init__(
 
     @property
     def is_functional(self):
+        if not _has_functorch:
```
vmoens, line 250: let's make sure that we also cover cases where the user has explicitly chosen to use the native implementation

torchrl/modules/tensordict_module/common.py
```diff
@@ -17,16 +17,28 @@
...
+    from functorch import FunctionalModule, FunctionalModuleWithBuffers, vmap
+    from functorch._src.make_functional import _swap_state
 
     _has_functorch = True
 except ImportError:
-    _has_functorch = False
+    print(
+        "failed to import functorch. TorchRL's features that do not require "
+        "functional programming should work, but functionality and performance "
+        "may be affected. Consider installing functorch and/or upgrating pytorch."
```
vmoens, line 35: update msg

torchrl/envs/libs/gym.py
```diff
@@ -276,9 +277,16 @@ class GymEnv(GymWrapper):
 
     """
 
-    def __init__(self, env_name, disable_env_checker=True, **kwargs):
+    def __init__(self, env_name, disable_env_checker=None, **kwargs):
```
vmoens, line 280: Let's put disable_env_checker in the docstring

torchrl/data/tensordict/utils.py
```diff
@@ -174,6 +174,39 @@ def convert_ellipsis_to_idx(idx: Union[Tuple, Ellipsis], batch_size: List[int]):
     return new_index
 
 
+def _copy(self: List[int]):
```
vmoens, line 177: that should be moved inside the fun def

PR #387, opened 2022-08-31
[Feature] In-house functional modules for TorchRL using TensorDict

vmoens:
## Description

Allows to use in-house functional modules using TensorDict.
This PR brings the following feature:
- It makes it easy to stack params together (e.g. in multi-task settings), or expand params (for parameter sharing)
- it makes it easy to build one's own set of parameters:
    ```python
    # make all of this functional
    net_total = nn.Sequential(backbone, net1)
    fnet, _ = make_functional(net_total)
    _, params_backbone = make_functional(backbone)
    _, params_net1 = make_functional(net1)
    _, params_net2 = make_functional(net1)

    # building the parameters is easy: the backbone just needs to be expanded and the task-specific parameters just need to be stacked
    params_tensordict = TensorDict(
        {"0": params_backbone.expand(2),  # these parameters are shared for both tasks
        "1": torch.stack([params_net1, params_net2], 0)}, # there params are not shared
        batch_size=[2]
    )
    # should work
    fnet(params_tensordict, data)
    ```
- for some users that rely on torch 1.10 (because of [fairo/polymetis](https://github.com/facebookresearch/fairo/tree/main/polymetis)), and because functional programming is awesome, we can't use functorch (hopefully this won't be a longstanding issue)

For context:
#364


torchrl/modules/functional_modules.py
```diff
@@ -0,0 +1,227 @@
...
+        )
+        try:
+            return self.stateless_model(*args, **kwargs)
+        finally:
+            # Remove the loaded state on self.stateless_model
+            if _RESET_OLD_TENSORDICT:
+                _swap_state(self.stateless_model, old_state)
+
+
+class FunctionalModuleWithBuffers(nn.Module):
```
zou3519, line 135: So it turns out PyTorch has two ways of using "functional modules". The first is the one that you're aware of, [functorch's make_functional](https://pytorch.org/functorch/stable/generated/functorch.make_functional.html). The second is [torch.nn.utils.stateless.functional_call](https://pytorch.org/docs/stable/generated/torch.nn.utils.stateless.functional_call.html).

As a part of the functorch merge into pytorch we've been tasked with figure out how to deduplicate these APIs. So I'm curious what your thoughts on the two APIs are.

torchrl/modules/functional_modules.py
```diff
@@ -0,0 +1,227 @@
...
+        setattr(model, name, None)
+        tensordict[name] = param
+    for name, module in model.named_children():
+        module_tensordict = extract_weights(module)
+        if module_tensordict is not None:
+            tensordict[name] = module_tensordict
+    if len(tensordict.keys()):
+        return tensordict
+    else:
+        return None
```
zou3519, line 188: For TensorDict integration with `functorch.make_functional`, it looks like you've taught the API how to handle loading and unloading state from a TensorDict. Furthermore, make_functional returns a TensorDict instead of a list.

I like how we've changed make_functional to return a dictionary-like thing instead of a list. One question I have is if we should have make_functional return a TensorDict, a dictionary, or give users an option between those.

It depends on how much of a "PyTorch core" feature we think TensorDict is. As a litmus test: 
- Right now, given an nn.Module, one can get a state_dict from it. This state_dict is a regular dictionary. Do you think it should be a TensorDict?

torchrl/modules/functional_modules.py
zou3519: It should be feasible to get torch.nn.utils.stateless.functional_call to work with TensorDict as well; we would just modify the API to accept a TensorDict along with dictionaries (maybe it just works today)

torchrl/modules/functional_modules.py
```diff
@@ -0,0 +1,227 @@
...
+            arg
+            if in_dim is None
+            else arg.apply(lambda _arg: _add_batch_dim(_arg, in_dim, vmap_level))
+            if isinstance(arg, TensorDictBase)
+            else _add_batch_dim(arg, in_dim, vmap_level)
+            for in_dim, arg in zip(flat_in_dims, flat_args)
+        ]
+        return tree_unflatten(batched_inputs, args_spec)
+
+    functorch._src.vmap._create_batched_inputs = _create_batched_inputs
```
zou3519, line 101: The usual way to get a custom data structure to work with vmap is to register pytree definitions for it. This consists of writing two functions, a "flatten" (which, given a Data Structure, returns a list containing all the elements and some metadata), and an "unflatten" (which, given a list and some metadata reconstructs the Data Structure).

For example, [here is the definition for dict](https://github.com/pytorch/pytorch/blob/5e9c26c8e25e5d5be18ff98b7808b674b1e7a0a5/torch/utils/_pytree.py#L47-L51).

Any Data Structure that has a pytree definition automatically works with vmap. Does it make sense to do this for TensorDict?

torchrl/modules/functional_modules.py
```diff
@@ -0,0 +1,227 @@
+from copy import deepcopy
+
+import torch
+from torch import nn
+
+from torchrl.data import TensorDict
+from torchrl.data.tensordict.tensordict import TensorDictBase
```
zou3519, line 7: Some quick questions to better understand TensorDict:
- Are all tensors required to be on the same device? What about other properties, like dtype?
- Does TensorDict optimize storage? Will it try to pack the tensors together in memory?

zou3519 (commented):
I went through the examples in the original PR and here. TensorDict is cool! Some high-level comments:
1. we should be able to get TensorDict to work with vmap without monkeypatching vmap, by registering it as a pytree. As a side note, this would also make it work with FX tracing.
2. we should change functorch's make_functional to return a TensorDict or a dict. Returning a list is not fun to work with, as you've discovered
3. PyTorch actually has two "functional module" APIs and we've been tasked with de-duplicating them so I'm curious if you have thoughts

In terms of shipping this PR, I'd be happy if we could resolve (1) so that changes to functorch's vmap don't end up breaking torchrl

torchrl/modules/functional_modules.py
vmoens: Device: yes as of now it does require all tensors to be on the same device. I think I see where you're going with this though: models can be split across devices. What we want to support is tensordict.to(device) but we could also have unset devices:
TensorDict({"a": torch.randn(3, device=0)}, []) # unset

TensorDict({"a": torch.randn(3, device=0)}, [], device=0) # set

In the second case all tensors must be on the same device, in the first they can be placed on different ones.

Dtype: no we don't require that by design. We want to store things like masks, booleans, integers (eg discrete actions) etc.  We also want to close the door to algebraic operations as I don't see the value of doing multiplications over tensors of different dtypes, so to say.
If you want to map all tensors to a dtype, you can do
TensorDict.apply_(lambda x: x.to(dtype))

Storage: since the dtype and shape are different across tensors I don't really see how we could optimise storage but I'm happy to hear your thoughts!

torchrl/modules/functional_modules.py
vmoens: Seems like functional_call would work with tensordict!
I like that buffers and params are split though: in meta RL you usually optimise either one or the other using dedicated algos. Merging them together is asking the user to split them thereafter which is impractical.

On the functorch side I don't really see the value of getting the params and not the buffers though (FunctionalModule va FunctionalModuleWithBuffers). You can get them and do nothing with it no?

zou3519: Sorry, to clarify, I meant that we have two sets of functional module APIs: there is {functorch.make_functional, functorch.make_functional_with_buffers} and {torch.nn.utils.stateless.functional_call}. We might deprecate one of the two sets. If we deprecate the functorch set, then I'm not quite sure how the torchrl code will look after

zou3519: >  in meta RL you usually optimise either one or the other using dedicated algos

Out of curiosity, what are the algorithms that optimize buffers?

vmoens: Stuff that optimise the gamma and lambda decay, I can find the papers

vmoens: > Sorry, to clarify, I meant that we have two sets of functional module APIs: there is {functorch.make_functional, functorch.make_functional_with_buffers} and {torch.nn.utils. stateless.functional_call}. We might deprecate one of the two sets. If we deprecate the functorch set, then I'm not quite sure how the torchrl code will look after

That's what I understood. I was saying that I like less stateless.functional_call because it does not split params and buffers

vmoens: > >  in meta RL you usually optimise either one or the other using dedicated algos
> 
> 
> 
> Out of curiosity, what are the algorithms that optimize buffers?

Here is one

https://proceedings.neurips.cc/paper/2018/file/2715518c875999308842e3455eda2fe3-Paper.pdf

torchrl/modules/functional_modules.py
vmoens: Let me be bold here: yes I think it should.
You can easily transform it to a tree structure:
TensorDict.unflatten_keys() will turn a flat tensordict in one that has a structure identical to you neural net. It's much easier to construct and to manipulate.

IMO it's an ideal structure for that, for transforms (eg when you want to transform a mask and an image in segmentation tasks etc), you can abstract away lots of stuff with it.

I think it does not have its place in TorchRL but right now that the best I can do with it. 
I think many usages people have of Raggedtensors from torchrec would also be simpler with tensordict (but I'm biased obviously).

@shagunsodhani suggested that we could do a separate repo under pytorch for it, I'm not against the idea. There's a workplace post on our group and tutorials on torchrl about it if you're interested.

torchrl/modules/functional_modules.py
vmoens: Sure I should try that out! Thanks for the tip

PR #399, opened 2022-09-07
[Feature] Added `batch_lock` attribute in EnvBase

nicolas-dufour:
## Description

This PR solves the problem to distinguish if batch size of an env is flexible or not

## Motivation and Context

The motivation is that stateful envs have different behavious than stateless envs with respect to batch_size. Where as stateful envs have a hard constraint on env.batch_size == tensordict.batch_size it's not the case of stateless envs. Therefore we add batch_locked attribute that allow or not batch flexibility in the env

## Types of changes

## Checklist


torchrl/envs/common.py
```diff
@@ -220,6 +220,7 @@ def __init__(
     @classmethod
     def __new__(cls, *args, **kwargs):
         cls._inplace_update = True
+        cls.batch_locked = True
```
vmoens, line 223: This should not be modified, so let's turn in into a property and raise an error when someone wants to set this attribute

torchrl/envs/gym_like.py
```diff
@@ -77,6 +77,7 @@ class GymLikeEnv(_EnvWrapper):
 
     @classmethod
     def __new__(cls, *args, **kwargs):
+        cls.batch_locked = True
```
vmoens, line 80: careful here: you're setting this in `__new__` then call super, which I think will override what you just wrote.
I would put a kwarg in the `__new__` of the parent class which defaults to True and for all the children pass that value upward.

vmoens (changes requested):
Perhaps we want to check that our envs have the attribute we want somewhere in the tests? 
(like in test_env, create one gym and one model-based and assess whether they both have the correct `batch_locked`?)
Also we should test that this can't be set once that is implemented

torchrl/envs/gym_like.py
nicolas-dufour: I thought it was better to set it in the init so that it's persistent if we change the default of EnvBase at some point

nicolas-dufour:
I've added tests for gym. I'll add the same test for MB_env in the MBEnv PR #333

test/test_env.py
```diff
@@ -992,6 +992,32 @@ def test_steptensordict(
...
 
 
+@pytest.mark.skipif(not _has_gym, reason="no gym")
+@pytest.mark.parametrize("env_name", ["Pendulum-v1", "ALE/Pong-v5"])
+@pytest.mark.parametrize("frame_skip", [1, 4])
+def test_batch_lock_gym(env_name, frame_skip, seed=0):
+    env = GymEnv(env_name, frame_skip=frame_skip)
+    assert env.batch_locked
+
+    try:
```
vmoens, line 1002: you can do 
```
with pytest.mark.raise
```

There are a couple of examples in the tests if you'd like to have a look

test/test_env.py
```diff
@@ -992,6 +992,32 @@ def test_steptensordict(
...
+        env.batch_locked = False
+        raise AssertionError("Should not be able to set batch_locked to False")
+    except RuntimeError:
+        pass
+    td = env.reset()
+    td["action"] = env.action_spec.rand()
+    td_expanded = td.expand(2)
+    td = env.step(td)
+
+    try:
```
vmoens, line 1012: same here

torchrl/envs/common.py
```diff
@@ -408,7 +420,9 @@ def set_state(self):
         raise NotImplementedError
 
     def _assert_tensordict_shape(self, tensordict: TensorDictBase) -> None:
-        if tensordict.batch_size != self.batch_size:
+        if tensordict.batch_size != self.batch_size and (
+            self._batch_locked or self.batch_size != torch.Size([])
```
vmoens, line 424: can we use `batch_locked` here?

torchrl/envs/gym_like.py
```diff
@@ -75,6 +76,19 @@ class GymLikeEnv(_EnvWrapper):
...
+        *args,
+        dtype: Optional[np.dtype] = None,
+        device: DEVICE_TYPING = "cpu",
+        batch_size: Optional[torch.Size] = None,
+        **kwargs,
+    ):
+        super().__init__(
+            *args, dtype=dtype, device=device, batch_size=batch_size, **kwargs
+        )
+        self._batch_locked = True
```
vmoens, line 90: I guess we should use `__new__` for that no?

torchrl/envs/common.py
```diff
@@ -198,6 +198,8 @@ def __init__(
             self.device = torch.device(device)
         self._is_done = None
         self.dtype = dtype_map.get(dtype, dtype)
+        if "_batch_locked" not in self.__dict__:
```
vmoens, line 201: why not in `__new__` since it's an immutable property?

vmoens (changes requested):

torchrl/envs/common.py
nicolas-dufour: We are using it: 
   ``` self._batch_locked or self.batch_size != torch.Size([]) ```

test/test_env.py
```diff
@@ -992,6 +992,24 @@ def test_steptensordict(
...
 
 
+@pytest.mark.skipif(not _has_gym, reason="no gym")
+@pytest.mark.parametrize("env_name", ["Pendulum-v1", "ALE/Pong-v5"])
+@pytest.mark.parametrize("frame_skip", [1, 4])
+def test_batch_lock_gym(env_name, frame_skip, seed=0):
+    env = GymEnv(env_name, frame_skip=frame_skip)
+    assert env.batch_locked
+
+    with pytest.raises(RuntimeError):
```
vmoens, line 1002: can we check the reason
I believe it's `raises(RuntimeError, reason="error message")`

torchrl/envs/common.py
```diff
@@ -220,8 +220,18 @@ def __init__(
     @classmethod
     def __new__(cls, *args, **kwargs):
         cls._inplace_update = True
+        if not hasattr(cls, "_batch_locked"):
```
vmoens, line 223: Why can't we do this
```
    @classmethod
    def __new__(cls, *args, _batch_locked=True, **kwargs):
        cls. _batch_locked = _batch_locked
```
and then for all subclasses
```
    @classmethod
    def __new__(cls, *args, **kwargs):
        super().__new__(_batch_locked = what_you_want_it_to_be)
```

Perhaps you have tried but it does not work? Curious about this

torchrl/envs/gym_like.py
```diff
@@ -78,6 +78,7 @@ class GymLikeEnv(_EnvWrapper):
     @classmethod
     def __new__(cls, *args, **kwargs):
         cls._info_dict_reader = None
+        cls._batch_locked = True
         return super().__new__(cls, *args, **kwargs)
```
vmoens, line 82: same here I think it makes more sense to pass this as an arg rather than setting it and then checking if it's not set

torchrl/envs/transforms/transforms.py
```diff
@@ -312,6 +312,8 @@ def __init__(
         device = kwargs["device"]
         super().__init__(**kwargs)
         self._set_env(env, device)
+        self._batch_locked = env._batch_locked
```
vmoens, line 315: I don't think it's appropriate to put this in `__init__`

vmoens (commented):
I added a couple more comments, can you address these?

torchrl/envs/common.py
nicolas-dufour: Haven't tried this way, checking for the cls to have the arge felt more intuitive for me since it's the place the parameters are set. It's also more robust to changes in the default EnvBase. However if you prefer to have it like above tell me and i will make the changes

torchrl/envs/transforms/transforms.py
nicolas-dufour: Yes i agree i didn't really now how to set it since we only get the env info in the init but _new_ is called before. How do you think we should proceed?

vmoens: oh yeah you're right
what about

```
@property
def batch_locked(self):
    return self.base_env.batch_locked
```

We should also consider what to do with `_BatchedEnv`

nicolas-dufour: Ok but we still need to overwrite _batch_locked, no? We don't want it to keep stuck at True if some reason we re-use the attribute

test/test_env.py
nicolas-dufour: pytest doesn't support reason arg for raises. The only other arg supported is match to do a regex on the raised error message, is this what you expected

.circleci/unittest/linux/scripts/environment.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
     - hypothesis
     - future
     - cloudpickle
-    - gym
+    - gym==0.25.1
```
vmoens, line 11: Can we revert this?

torchrl/envs/transforms/transforms.py
vmoens: what do you mean? That a transformed env that has a batch_locked base env could itself be non batched_locked?

vmoens:
I think tests are failing because you have not merged main into your branch recently but i might be wrong

vmoens (commented):
I'd like to know if you consider that TransformedEnv should make _batch_locked mutable? Can we also remove the commented lines?

nicolas-dufour:
Why would we make TransformedEnv being able to change the batch_locked. At the end of the day, base_env will still have that constraint

torchrl/envs/transforms/transforms.py
nicolas-dufour: Ok i didn't understood what you said. What i mean is that _batch_locked will always be True for transformed env

vmoens: Oh why?

nicolas-dufour: I've just realized that this is not an issue anymore with the parameter passing in _new_. Was an issue before if we did
```
@property
def batch_locked(self):
    return self.base_env.batch_locked
```

test/mocking_classes.py
```diff
@@ -126,6 +126,139 @@ def rand_step(self, tensordict: Optional[TensorDictBase] = None) -> TensorDictBa
...
+
+        return TensorDict(
+            {"reward": n, "done": done, "next_observation": n}, batch_size
+        )
+
+    def rand_step(self, tensordict: Optional[TensorDictBase] = None) -> TensorDictBase:
+        return self.step(tensordict)
+
+
+class MockBatchedUnLockedEnv(EnvBase):
```
vmoens, line 195: does this inherit from another env? Can we inherit things like set_seed etc?
It feel that only the `__new__` method changes, am I right?
Perhaps we could just inherit from the other mocking class and change the `__new__` for he non-batch_locked version?

test/test_env.py
```diff
@@ -992,6 +994,126 @@ def test_steptensordict(
...
+    td = env.step(td)
+
+    with pytest.raises(
+        RuntimeError, match="Expected a tensordict with shape==env.shape, "
+    ):
+        env.step(td_expanded)
+
+
+@pytest.mark.parametrize("device", get_available_devices())
+def test_batch_locked_transformed(device):
```
vmoens, line 1049: Since it involves transforms, could we move this to test_transforms.py?

test/test_env.py
```diff
@@ -992,6 +994,126 @@ def test_steptensordict(
...
+    assert not env.batch_locked
+
+    with pytest.raises(RuntimeError, match="batch_locked is a read-only property"):
+        env.batch_locked = False
+    td = env.reset()
+    td["action"] = env.action_spec.rand(env.batch_size)
+    td_expanded = td.expand(2).clone()
+    td = env.step(td)
+
+    try:
```
vmoens, line 1087: why do we need a try / except here? Isn't the error enough, if there is one? Plus here you capture the RuntimeError but then on the CI we won't see it, we will only see a generic message

test/test_env.py
```diff
@@ -992,6 +994,126 @@ def test_steptensordict(
...
+        ),
+    )
+    assert not env.batch_locked
+
+    with pytest.raises(RuntimeError, match="batch_locked is a read-only property"):
+        env.batch_locked = False
+
+    td = env.reset()
+    td["action"] = env.action_spec.rand(env.batch_size)
+    td_expanded = td.expand(2, 2).reshape(-1).to_tensordict()
```
vmoens, line 1108: perhaps reshape => view > to_tensordict? reshape creates one more copy of the data

torchrl/envs/common.py
```diff
@@ -218,10 +228,19 @@ def __init__(
...
     @classmethod
-    def __new__(cls, *args, **kwargs):
+    def __new__(cls, *args, _batch_locked=True, **kwargs):
         cls._inplace_update = True
+        cls._batch_locked = _batch_locked
         return super().__new__(cls)
 
+    @property
+    def batch_locked(self) -> bool:
+        return self._batch_locked
```
vmoens, line 238: we should document what batch_locked means here. A proper docstring would be nice

vmoens (approved):
Sorry about this, still a couple of fix to be made before landing

test/test_env.py
nicolas-dufour: It does not work with view after expand

torchrl/envs/common.py
```diff
@@ -218,10 +228,22 @@ def __init__(
...
-    def __new__(cls, *args, **kwargs):
+    def __new__(cls, *args, _batch_locked=True, **kwargs):
         cls._inplace_update = True
+        cls._batch_locked = _batch_locked
         return super().__new__(cls)
 
+    @property
+    def batch_locked(self) -> bool:
+        """
+        Whether the environnement can be used with a batch size different from the one it was initialized with or not.
```
vmoens, line 239: """Whether the environnement can be used with a batch size different from the one it was initialised with."""

Also, I would add that this is an immutable property. 

And we don't get if `True` means that the batch size can be different of not. The proper formulation would be "returns True if the env blabla"

Can you provide an example in the docstring?

test/mocking_classes.py
```diff
@@ -126,6 +126,77 @@ def rand_step(self, tensordict: Optional[TensorDictBase] = None) -> TensorDictBa
...
+            action=NdUnboundedContinuousTensorSpec((1,)),
+            observation=NdUnboundedContinuousTensorSpec((1,)),
+        )
+        self.observation_spec = CompositeSpec(
+            next_observation=NdUnboundedContinuousTensorSpec((1,))
+        )
+        self.reward_spec = NdUnboundedContinuousTensorSpec((1,))
+        self.counter = 0
+
+    def set_seed(self, seed: int) -> int:
```
vmoens, line 143: can we do set_seed = MockSerialEnv.set_seed or something like that, not to re-write entire methods? This does not seem to depend on batch size.

test/mocking_classes.py
```diff
@@ -126,6 +126,77 @@ def rand_step(self, tensordict: Optional[TensorDictBase] = None) -> TensorDictBa
...
+        assert seed >= 1
+        self.seed = seed
+        self.counter = seed % 17  # make counter a small number
+        self.max_val = max(self.counter + 100, self.counter * 2)
+        return seed_generator(seed)
+
+    def _step(self, tensordict):
+        self.counter += 1
+        n = (
+            torch.full(tensordict.batch_size, self.counter)
```
vmoens, line 153: I'm confused: if ti's batch_lokced, the env batch size indicated the size of the output, right?
Why are we taking it from the tensordict? I guess we assume they're the same but from a design perspective it should be env.batch_size no? Perhaps it's because you wand the non-locked class to inherit from this one? If that's the case, can you leave a comment explaining where this comes from?

test/test_env.py
```diff
@@ -992,6 +994,52 @@ def test_steptensordict(
         assert out is next_tensordict
 
 
+@pytest.mark.parametrize("device", get_available_devices())
+def test_batch_locked(device):
```
vmoens, line 998: I feel there are 3 different blocks in there, can we split this in 3 tests?

vmoens (approved):
almost there, a couple of minor things

vmoens:
@nicolas-dufour have a look at my changes, if you're ok with them we can merge

test/mocking_classes.py
```diff
@@ -140,10 +142,13 @@ def __init__(self, device, batch_size=None):
         self.reward_spec = NdUnboundedContinuousTensorSpec((1,))
         self.counter = 0
 
-        self.set_seed = MockSerialEnv.set_seed
+    set_seed = MockSerialEnv.set_seed
+    rand_step = MockSerialEnv.rand_step
 
     def _step(self, tensordict):
         self.counter += 1
+        if self.batch_locked:
```
nicolas-dufour, line 150: This is actually done inside the step of batch env, so no need to redo it

vmoens: Oh ok!

PR #408, opened 2022-09-08
[Feature] Create a Squeeze transform and update Unsqueeze transform

### LINKED ISSUE #393: [Feature Request] Create a Squeeze transform from the Unsqueeze transform
## Motivation

Squeeze and unsqueeze are mutually reversible operations.
We already have an unsqueeze transform for our environments. However, it does not yet have the capacity to apply the transform in reverse mode (like, for instance, DoubleToFloat does).

## Solution

We could implement an inverse operation for Unsqueeze that would squeeze the input, then design a Squeeze transform that would be the flipped version of the unsqueeze (i.e. the `__init__` method would flip `keys_in` and `keys_inv_in`).

In the two classes, if `action` is in the `keys_inv_in` then the spec should be transformed accordingly (see `DoubleToFloat` for an example), same for all the keys in `input_spec`.

reachsumit:
## Description

Create a Squeeze transform and update Unsqueeze transform

## Motivation and Context

This code change is targeted to address the issue#393. "Create a Squeeze transform from the Unsqueeze transform".
close #393 

## Types of changes

- [x ] New feature (non-breaking change which adds core functionality)

Changes made:
* Implemented an inverse operation for Unsqueeze that would squeeze the input.
* Implemented a Squeeze transform that would be the flipped version of the unsqueeze.
* Wrote 3 tests: 1 for the unsqueeze inverse operation, one for the squeeze forward, and one for the squeeze inverse.
* Applied appropriate spec transformations if `action` is in the `keys_inv_in` and for all the keys in input_spec.

No `pre-commit` or `pytest` issues.

## Checklist

- [x ] I have read the [CONTRIBUTION](https://github.com/facebookresearch/rl/blob/main/CONTRIBUTING.md) guide (**required**)


vmoens (commented):
LGTM
I changed the gym version to provisionally fix the CI, let's see if tests pass! 
I'll give it a deeper look then and approve in case it works ok :)

test/test_transforms.py
```diff
@@ -443,6 +444,116 @@ def test_unsqueeze(self, keys, size, nchannels, batch, device, unsqueeze_dim):
...
+            {
+                key: torch.randn(*batch, *size, nchannels, 16, 16, device=device)
+                for key in keys_total
+            },
+            batch,
+        )
+
+        unsqueeze.inv(td)
+
+        expected_size = [*size, nchannels, 16, 16]
```
vmoens, line 476: Nit: we could create the `torch.Size` here

reachsumit: Thank you for looking into this. 'torch.Size' object doesn't seem to support item deletion, which is conditionally required on line 481.

PR #409, opened 2022-09-08
[BugFix] Updated TensorDict.expand to work as Tensor.expand

### LINKED ISSUE #398: [BUG] TensorDict.expand does not work as Tensor.expand
## Describe the bug

Currently, TensorDict.expand does not work as expected:
```python
>>> import torch
>>> tensor = torch.arange(6).view(3, 2)
>>> tensor.expand(2, 3, 2)
tensor([[[0, 1],
         [2, 3],
         [4, 5]],

        [[0, 1],
         [2, 3],
         [4, 5]]])
>>> t = TensorDict({"a": tensor}, [3])
>>> t.expand(1)["a"]  # gets the same result as above
tensor([[[0, 1],
         [2, 3],
         [4, 5]],

        [[0, 1],
         [2, 3],
         [4, 5]]])

```

However, one would expect that this expand would take the argument `2, 3` and not just `2`.

Here is a list of the expected behaviours:
```python
>>> t = TensorDict({"a": tensor}, [3])
>>> t.expand(2, 3)  # should return a tensordict of size [2, 3]. Currently it will happen the 2, 3 dimensions, resulting in size [2, 3, 3]
>>> t = TensorDict({"a": tensor}, [1])
>>> t.expand(2, 3)  # should return a tensordict of size [2, 3]. Currently it will happen the 2, 3 dimensions, resulting in size [2, 3, 1]
>>> t = TensorDict({"a": tensor}, [2])
>>> t.expand(2, 3)  # should break, as the last dim is 2 but the indicated by expand is [2, 3]. 
```

This will require a bit of refactoring in the code, including in the tutorials and tests.
A quick look through all the `.expand` in the lib is advised. In general, the current calls to `tensordict.expand(*new_dims)` can simply be replaced by `tensordict.expand(*new_dims, *tensordict.shape)` after the refactoring.

AnshulSehgal:
## Description

1. Updated the TensorDict.expand to work like tensor.expand
2. Updated the usage of TensorDict.expand in other files to take  (*new_shape, *tensordict.batch_size) as args instead of (*new_shape)
3. Updated test cases and added a new test case to test the singleton dim expand case
4. Updated the tutorial ipython notebook to reflect the updated usage

Cause of the issue #398 :
Currently in the function when tensor.expand is called the argument passed is concatenation of new shape and the shape of the tensor stored as value in the key value pair and the batch_size of the resultant TensorDict is also a concatenation of new shape and current batch size.
Because of the above setup the tensorDict and tensors stored as values are always expanded to larger number of dimensions, the new ones (same as shape passed to the expand function) appended at the front

Expected behavior after my change
The shape passed as argument should be the final shape of the TensorDict after expansion and the function is updated to take care of following 3 scenarios

1. new shape is same as old shape: Example new_shape = [3,2] old_shape = [3,2] In this case nothing changes 
2. new shape has same length as old shape but different values: This can have two cases
      a. singleton dimension in old_shape: Example new_shape [3,2] old_shape = [3,1]. In this case the expanded TensorDict should have the shape [3,2] and all tensors inside the dict should also change the shape from [3,1,x,x] to [3,2,x,x]
      b. no singleton dimension in old shape: Should throw an error
3. new shape has length greater than old shape: Example new_shape = [3, 3, 2] old_shape = [3, 2]. In this case the expanded TensorDIct should have the shape [3, 3, 2] and all tensors stored in the dict as values should have the shape [3,3,2,x,x,..]. Another example here would be new_shape [3, 3, 2] old_shape [1, 2]. In this case also the output should be same as above but the size of 2nd dimension would change from 1 to 3 for TensorDict and all tensors within it 

## Motivation and Context

 `close #398` if this solves the issue #398 

## Types of changes

## Checklist


vmoens (approved):
High-level comment: better to create a separate branch on your fork and not work on main :)
I don't mind merging this PR but I wanted to change something to fix the CI and run the tests, unfortunately I can't do that if you're working on main

Otherwise LGTM

I was wondering if we should test that the expanded shape is compatible with the old one?
We'll get an error if it doesn't anyway but perhaps things could be clearer for the user if we indicate what has gone wrong before the expand operation is applied on the tensors.

Let's way for the CI to pass (I'll ping you when we can make it work) and see if we want to implement this exception before merging.

vmoens:
I made a PR on your fork to fix the tests!

vmoens:
@AnshulSehgal there is an expand that has not been corrected in the `redq.py` loss, can you patch that one too?

AnshulSehgal:
Done!

PR #412, opened 2022-09-09
[Feature] lock tensordict when calling `share_memory_()`

### LINKED ISSUE #125: [WIP] `TensorDict.is_shared()` and updates of the object
When a tensordict is shared in memory, all of its tensors have a storage that is changed accordingly.

We need to decide what happens when a tensor is added to the tensordict. Here are the various use-cases:

1. The tensordict is shared, the tensor is not, the key already exists, inplace
2. The tensordict is shared, the tensor is not, the key already exists, out-of-place
3. The tensordict is shared, the tensor is not, the key does not exist
4. The tensordict is on cuda (hence it is shared) and all of the above

fdabek1:
## Description

Added `lock=True` flag to both `share_memory_` and `memmap_` methods of TensorDict classes. And when calling the `set` method this only allows the key to be updated if the key already exists and `inplace` is true.

This required a couple changes to existing code/tests:

- Added `lock=False` to existing tests to let them continue to pass.
- Removed `is_locked` checks from `set_`, `set_at_`, and `_stack_onto_` methods due to the issue #125 stating that these methods should ignore the lock.

## Motivation and Context

`Close #125`

## Types of changes

## Checklist


vmoens:
LGTM, 
one last change before we accept this and merge it:
We can write `tensordict.set(value, inplace=True)` (see [here](https://github.com/facebookresearch/rl/blob/12c351f66893edbc7a5103c6602637953de717da/torchrl/data/tensordict/tensordict.py#L2016-L2017))

In that case either (1) the key-tensor pair is present and then is written in place or (2) the key is missing and we fall back onto regular `tensordict.set()`. If (1), I think we'd like for lock to be bypassed no?

fdabek1:
> LGTM, one last change before we accept this and merge it: We can write `tensordict.set(value, inplace=True)` (see [here](https://github.com/facebookresearch/rl/blob/12c351f66893edbc7a5103c6602637953de717da/torchrl/data/tensordict/tensordict.py#L2016-L2017))
> 
> In that case either (1) the key-tensor pair is present and then is written in place or (2) the key is missing and we fall back onto regular `tensordict.set()`. If (1), I think we'd like for lock to be bypassed no?

I don't believe that `tensordict.set(value, inplace=True)` seems to be a valid call.  The `key` and `value` parameters are both required (no default values).  Additionally, looking a couple lines above the linked code, it uses the `key` parameter for defining the `present` value.

vmoens:
> > LGTM, one last change before we accept this and merge it: We can write `tensordict.set(value, inplace=True)` (see [here](https://github.com/facebookresearch/rl/blob/12c351f66893edbc7a5103c6602637953de717da/torchrl/data/tensordict/tensordict.py#L2016-L2017))
> > In that case either (1) the key-tensor pair is present and then is written in place or (2) the key is missing and we fall back onto regular `tensordict.set()`. If (1), I think we'd like for lock to be bypassed no?
> 
> I don't believe that `tensordict.set(value, inplace=True)` seems to be a valid call. The `key` and `value` parameters are both required (no default values). Additionally, looking a couple lines above the linked code, it uses the `key` parameter for defining the `present` value.

I must have been very tired when I wrote that haha
yes, it should have read `tensordict.set(*key*, value, inplace=True)`

fdabek1:
> > > LGTM, one last change before we accept this and merge it: We can write `tensordict.set(value, inplace=True)` (see [here](https://github.com/facebookresearch/rl/blob/12c351f66893edbc7a5103c6602637953de717da/torchrl/data/tensordict/tensordict.py#L2016-L2017))
> > > In that case either (1) the key-tensor pair is present and then is written in place or (2) the key is missing and we fall back onto regular `tensordict.set()`. If (1), I think we'd like for lock to be bypassed no?
> > 
> > 
> > I don't believe that `tensordict.set(value, inplace=True)` seems to be a valid call. The `key` and `value` parameters are both required (no default values). Additionally, looking a couple lines above the linked code, it uses the `key` parameter for defining the `present` value.
> 
> I must have been very tired when I wrote that haha yes, it should have read `tensordict.set(*key*, value, inplace=True)`

Haha no worries.

Are there any circumstances that we should be bypassing the lock?  I believe all of the circumstances seem to be handled.

torchrl/data/tensordict/tensordict.py
```diff
@@ -2730,7 +2733,8 @@ def set(
         _run_checks: bool = True,
     ) -> TensorDictBase:
         if self.is_locked:
-            raise RuntimeError("Cannot modify immutable TensorDict")
+            if not inplace or key not in self.keys():
```
vmoens, line 2736: iterating through the keys can be a tedious operation when repeated frequently. We try to avoid doing that twice per function call by creating a set at the beginning. Can you move that before your call and check the presence in the set (rather than in the iterator method)?

vmoens (approved):
LGTM and tests are passing, which is awesome! Once you've addressed that minor comment regarding `set(self.keys())` we're good to go!

torchrl/data/tensordict/tensordict.py
fdabek1: I moved the local `keys` variable from below up to remove the duplicate call.  Also wrapped other calls of `self.keys()` with `set()` just in case.

fdabek1:
I updated and responded to the minor comment above.  Good catch!

torchrl/data/tensordict/tensordict.py
```diff
@@ -1991,7 +1995,8 @@ def set(
         and if the key already exists, set will call set_ (in place setting).
         """
         if self.is_locked:
-            raise RuntimeError("Cannot modify immutable TensorDict")
+            if not inplace or key not in set(self.keys()):
```
vmoens, line 1998: you don't need to create a set every time :)
The idea is this, say you have a function that does this

```
if key in self.keys():
   do_stuff()
elif key is not in self.keys() and foo:
    val = bar
```
then you're iterating twice over the `keys()` iterator, which can be costly. Then we create a set once and avoid calling the iterator.
If `keys()` is called just once it's actually faster to leave it as it is (because creating a set is not free and the condition `if key in self.keys()` will stop whenever a matching keys is found, which can happen before all keys have been scanned)

vmoens:
> I updated and responded to the minor comment above. Good catch!

Sorry I think I mislead you with this, there is only a couple of places where the set makes sense, have a look at my comment

fdabek1:
> Sorry I think I mislead you with this, there is only a couple of places where the set makes sense, have a look at my comment

I realized afterwards my mistake.  Just updated to remove the extra `set()` calls.

PR #428, opened 2022-09-12
[Feature] New `biased_softplus` semantic to allow for minimum scale setting

nicolas-dufour:
## Description

This PR allow user to set a different min_std in the mapping function.

## Motivation and Context

This PR allows to have a scale mapping such as we can pass the following string: "biased_sofplus_{bias}_{min_std}".
"biased_sofplus_{bias}" is also still supported.

## Types of changes

## Checklist


torchrl/modules/utils/mappings.py
```diff
@@ -65,7 +65,11 @@ def mappings(key: str) -> Callable:
 
     Args:
         key (str): one of "softplus", "exp", "relu", "expln",
-        or "biased_softplus".
+        or "biased_softplus". If key beggins with "biased_softplus",
```
vmoens, line 68: "if the key"

torchrl/modules/utils/mappings.py
```diff
@@ -65,7 +65,11 @@ def mappings(key: str) -> Callable:
...
+        then the key needs to take the following form:
```
vmoens, line 69: then it needs

torchrl/modules/utils/mappings.py
```diff
@@ -65,7 +65,11 @@ def mappings(key: str) -> Callable:
...
+        "biased_softplus_{bias}" where bias is a float and the bias term of the biased softplus.
```
vmoens, line 70: `"biased_softplus_{bias}"` (in code style)
floating point number

where `"bias"` can be converted to a floating point number that will be used to bias the softplus function.

torchrl/modules/utils/mappings.py
```diff
@@ -65,7 +65,11 @@ def mappings(key: str) -> Callable:
...
+        or "biased_softplus_{bias}_{min_val}" where bias is a float and the bias term,
```
vmoens, line 71: Alternatively, the `"biased_softplus_{bias}_{min_val}"` syntax can be used. In that case, the additional `"min_val"` term is a floating point number that will be used to encode the minimum value of the softplus transform.
In practice, the equation used is `softplus(x + bias) + min_val`, where `bias` and `min_val` are values computed such that the conditions above are met.

torchrl/modules/utils/mappings.py
```diff
@@ -81,6 +85,15 @@ def mappings(key: str) -> Callable:
     if key in _mappings:
         return _mappings[key]
     elif key.startswith("biased_softplus"):
-        return biased_softplus(float(key.split("_")[-1]))
+        stripped_key = key.replace("biased_softplus_", "").split("_")
```
vmoens, line 88: why not just splitting (and doing one op only)?

vmoens (commented):
LGTM, let's address the small changes.

On the long term we should find a better way to code this than a string but since it builds up on the existing API i'm ok with it for now.

vmoens (approved):
LGTM thanks

PR #438, opened 2022-09-13
[Refactor] Refactoring `set*()` methods for `TensorDictBase` class

### LINKED ISSUE #431: [Feature Request] Fix calls to `_process_tensor` in `tensordict.set`, `tensordict.set_` and `tensordict.set_at_`
## Motivation

The purpose of `_process_tensor` is to map the tensor to a tensor, possibly on the right device, when calling `set_`, `set_at_` and `set`. Calls to `_process_tensor` are inconsistent across `TensorDict` classes.
We should fix that.

Here are the requirements for the fix: for each call to `set` / `set_` or `set_at_` there should be one and only one call to `_process_tensor` as this can be expensive. A call to `_process_tensor` should also be avoidable if needed with a flag to the call to the method (those checks can be expensive and we would like to skip them from time to time).

Heres' how to test the fix: we should test that we can set a numpy array of the right size in each tensordict class. We should also test that this can be done in place (`set_`) and at the desired index (`set_at_`). We could also test this for a list. I believe we already test that for lists

zeenolife:
## Description

For all the `TensorDictBase` child classes, refactoring `set*()` methods, so that `dict -> TensorDict` casting happens only in `_process_input()`. 

## Motivation and Context
Removing duplicate code and making it consistent across the child classes.
close #431 

## Types of changes

## Checklist

### DIFF
--- a/torchrl/data/tensordict/tensordict.py
+++ b/torchrl/data/tensordict/tensordict.py
@@ -2073,12 +2073,14 @@ class TensorDict(TensorDictBase):
         if not isinstance(key, str):
             raise TypeError(f"Expected key to be a string but found {type(key)}")
 
-        # TODO Handle ._is_shared
         if self._is_shared is None:
             try:
                 self._is_shared = value.is_shared()
             except NotImplementedError:
                 # when running functorch, a NotImplementedError may be raised
+                pass
+            except AttributeError:
+                # when setting a value of type dict
                 pass
         if self._is_memmap is None:
             self._is_memmap = isinstance(value, MemmapTensor)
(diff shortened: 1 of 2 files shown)


torchrl/data/tensordict/tensordict.py
```diff
@@ -2195,17 +2197,14 @@ def _stack_onto_at_(
...
-                value, batch_size=self.batch_size, device=self.device_safe()
-            )
         if not isinstance(key, str):
             raise TypeError(f"Expected key to be a string but found {type(key)}")
 
         # do we need this?
-        # value = self._process_tensor(
-        #     value, check_tensor_shape=False, check_device=False
-        # )
+        if not isinstance(value, _accepted_classes):
```
vmoens, line 2204: For record: I guess we want to process the input only if it's not of the right class, and not check the device

I think the previous logic was: if you're calling set_at_ you're basically doing something like `tensor.copy_(smth)`. The only requirement of tensor.copy is "The src tensor must be [broadcastable](https://pytorch.org/docs/stable/notes/broadcasting.html#broadcasting-semantics) with the self tensor. It may be of a different data type or reside on a different device." (see [here](https://pytorch.org/docs/stable/generated/torch.Tensor.copy_.html)) so I guess we're fine with this

Issue #396, opened 2022-09-07
[Feature Request] Test loggers video saving


vmoens:
## Motivation

We currently don't test the logging of videos through the loggers.

## Solution

Add a test_log_video to `TestTensorboard`, `TestCSVLogger` and `TestWandbLogger`.
If possible, we should test that a very simple video (white and black frames for instance) is logged properly. 

In each logger, we could add a shape check in `log_video`: the shape should be [(N), T, C, H, W] with C=3 or 1

For all loggers except csv, we should check that fps can be passed and that it has an impact on the logged video.

PR #439, opened 2022-09-13
[Tests] Test loggers video saving

bashnick:
https://github.com/facebookresearch/rl/issues/396

## Description

Added the tests for the log_video methods in Tensorboard, CSVLogger and WandbLogger.

## Motivation and Context

Fixes an open issue, #396 

## Types of changes

## Checklist

### DIFF
--- a/test/test_loggers.py
+++ b/test/test_loggers.py
@@ -53,7 +53,7 @@ class TestTensorboard:
             # C - number of image channels (e.g. 3 for RGB), H, W - image dimensions.
             # the first 64 frames are black and the next 64 are white
             video = torch.cat(
-                (torch.zeros(64, 1, 32, 32), torch.mul(torch.ones(64, 1, 32, 32), 255))
+                (torch.zeros(64, 1, 32, 32), torch.full((64, 1, 32, 32), 255))
             )
             video = video[None, :]
             for i in range(3):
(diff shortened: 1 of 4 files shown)


test/test_loggers.py
```diff
@@ -42,6 +42,48 @@ def test_log_scalar(self, steps):
...
+        torch.manual_seed(0)
+        with tempfile.TemporaryDirectory() as log_dir:
+            exp_name = "ramala"
+            logger = TensorboardLogger(log_dir=log_dir, exp_name=exp_name)
+
+            # creating a sample video (T, C, H, W), where T - number of frames,
+            # C - number of image channels (e.g. 3 for RGB), H, W - image dimensions.
+            # the first 64 frames are black and the next 64 are white
+            video = torch.cat(
+                (torch.zeros(64, 1, 32, 32), torch.mul(torch.ones(64, 1, 32, 32), 255))
```
vmoens, line 56: you can use `torch.full` here for a tensor full of 255

vmoens (commented):
LGTM, do you want to test that the video looks like what it should? Also any idea of how to test the fps for tensorboard?

test/test_loggers.py
bashnick: yes, thanks!

bashnick: Re the "fps for tensorboard", the problem is that the return format for the video is an encoded_string, which has the same length for any video. I didn't find a nice method to convert it back to tensor or video object.
Re the "want to test that the video looks like what it should": 
- I am testing it in the CSVLogger
- For the WandbLogger and Tensorboard haven't found a nice method to do that.

vmoens: Ok I think we're good to go as it is!

vmoens (approved):
LGTM thanks for this

PR #442, opened 2022-09-14
Revert "[BugFix] Fixed Additive noise"

vmoens:
Reverts facebookresearch/rl#441


vmoens:
@nicolas-dufour can we fix the tests too? Thanks

PR #443, opened 2022-09-14
[Feature ]Add transform_observation_spec _R3MNet

### LINKED ISSUE #407: [Feature Request] Create a tensorspec for the R3M output
## Motivation

R3M (torchrl.envs.transforms.r3m) is a generic transform layer that will embed an image in a tensor, easier to work with using a policy or value function.
In TorchRL, each env has am `observation_spec` attribute that contains `tensor_spec` that describe what the observations are (they behave like dictionaries, and all of their keys must start with `"next_"`).

Here is how the observation_spec should behave:
1. We should add one TensorSpec for the output of R3M (the vector). It should have the size that is expected given the resnet used (i.e. an unbounded continuous tensorspec of size 2048 IIRC for resnet50).
2. We should delete the tensorspec of the pixels if it disappears (which will happen unless `tensor_pixels_keys` is set). e.g. when doing `R3MTransform("resnet50", keys_in=["next_pixels"])`, we should not see the "next_pixels" key in the `observation_spec` of the TransformedEnv once R3M has been applied since this key-value pair has disappeared. Instead, we should see the spec of "next_r3m_vec".
3. If the value of "next_pixels" is kept somewhere (using `tensor_pixels_keys`) we should duplicate the value of the `"next_pixels"`'s TensorSpec onto the key(s) indicated by `tensor_pixels_keys`.

ymwdalex:
### Define transform_observation_spec function in _R3MNet class
1. if del_keys is True, we delete all items in the input observation_spec
2. We iterate the keys_out, for each output key, we generate NdUnboundedContinuousTensorSpec objects with corresponding ResNet output dimension as shape

### Add test cases
One test case TestR3MNet.test_r3mnet_transform_observation_spec is added. In the test, we set up these parameters
- models: resnet 18/34/50
- del_keys: True or False
- keys_in: [["next_pixels"], ["next_pixels_1", "next_pixels_2", "next_pixels_3"]]
- keys_out: [["next_r3m_vec"], ["next_r3m_vec_1", "next_r3m_vec_2", "next_r3m_vec_3"]]

Then we compare the new observation_spec with the expected ones.

## Motivation and Context
close #407 

## Types of changes

## Checklist


ymwdalex:
Hi @vmoens, I got the message for the workflow:
```
1 workflow awaiting approval
First-time contributors need a maintainer to approve running workflows. [Learn more.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)
```
Could you please approve this workflow? Thanks!

test/test_transforms.py
```diff
@@ -1294,6 +1296,57 @@ def test_r3m_parallel(self, model, device):
...
+        [["next_pixels"], ["next_pixels_1", "next_pixels_2", "next_pixels_3"]],
+    )
+    @pytest.mark.parametrize(
+        "out_keys",
+        [["next_r3m_vec"], ["next_r3m_vec_1", "next_r3m_vec_2", "next_r3m_vec_3"]],
+    )
+    def test_r3mnet_transform_observation_spec(
+        self, in_keys, out_keys, del_keys, device, model
+    ):
+        in_keys = ["next_pixels"]
```
vmoens, line 1314: if `in_keys` is reset, perhaps let's remove it from the args? Or keep it in the arge but not reset it

test/test_transforms.py
```diff
@@ -1294,6 +1296,57 @@ def test_r3m_parallel(self, model, device):
...
+        if del_keys:
+            exp_ts = CompositeSpec(
+                **{
+                    key: NdUnboundedContinuousTensorSpec(r3m_net.outdim, device)
+                    for key in out_keys
+                }
+            )
+
+            observation_spec_out = r3m_net.transform_observation_spec(observation_spec)
+            for key in out_keys:
```
vmoens, line 1331: It would be good to test that the observation_spec matches what's expected, i.e things that have to be deleted aren't there anymore

test/test_transforms.py
```diff
@@ -1294,6 +1296,57 @@ def test_r3m_parallel(self, model, device):
...
+        else:
+            ts_dict = {}
+            for key in in_keys:
+                ts_dict[key] = observation_spec[key]
+            for key in out_keys:
+                ts_dict[key] = NdUnboundedContinuousTensorSpec(r3m_net.outdim, device)
+            exp_ts = CompositeSpec(**ts_dict)
+
+            observation_spec_out = r3m_net.transform_observation_spec(observation_spec)
+            for key in out_keys:
```
vmoens, line 1344: shouldn't we iterate over in_keys too?

test/test_transforms.py
```diff
@@ -1294,6 +1296,57 @@ def test_r3m_parallel(self, model, device):
         del transformed_env
 
 
+@pytest.mark.parametrize("device", get_available_devices())
+@pytest.mark.parametrize("model", ["resnet18", "resnet34", "resnet50"])
+class TestR3MNet:
```
vmoens, line 1301: can't we merge this with the above class, and keep a single test class for R3M?

test/test_transforms.py
```diff
@@ -1294,6 +1296,57 @@ def test_r3m_parallel(self, model, device):
...
+            for key in out_keys:
+                ts_dict[key] = NdUnboundedContinuousTensorSpec(r3m_net.outdim, device)
+            exp_ts = CompositeSpec(**ts_dict)
+
+            observation_spec_out = r3m_net.transform_observation_spec(observation_spec)
+            for key in out_keys:
+                assert observation_spec_out[key].shape == exp_ts[key].shape
+                assert observation_spec_out[key].dtype == exp_ts[key].dtype
+                assert observation_spec_out[key].device == exp_ts[key].device
+
```
vmoens, line 1348: I would add one more test in the same class, where we create a transformed env with R3M and check that the expected observation_spec matches what we get from rollouts.

Something like
```python
    @pytest.mark.parametrize("tensor_pixels_key", [None, ["funny_key"]])
    def test_r3m_spec_against_real(self, model, tensor_pixels_key, device):
        keys_in = ["next_pixels"]
        keys_out = ["next_vec"]
        r3m = R3MTransform(
            model,
            keys_in=keys_in,
            keys_out=keys_out,
            tensor_pixels_keys=tensor_pixels_key,
        )
        base_env = DiscreteActionConvMockEnvNumpy().to(device)
        transformed_env = TransformedEnv(base_env, r3m)
        expected_keys = list(transformed_env.input_spec.keys()) + \
                        list(transformed_env.observation_spec.keys()) + \
                        [key.strip("next_") for key in transformed_env.observation_spec.keys()] + \
                        ["reward"] + ["done"]
        assert set(expected_keys) == set(transformed_env.rollout(3).keys())
```

vmoens (commented):
LGTM, let's clean up the tests and land this!

test/test_transforms.py
ymwdalex: Thanks for pointing out. Fixed

test/test_transforms.py
ymwdalex: Sorry for the careless mistake. Fixed

test/test_transforms.py
ymwdalex: Thanks for pointing out. Fixed

ymwdalex:
> LGTM, let's clean up the tests and land this!

Thanks for your comments! I have improved the code accordingly, rebased to the latest main branch, and update the branch.

vmoens (approved):
LGTM thanks!

PR #445, opened 2022-09-14
[Feature] ParallelEnv always creates and EnvCreator for compatibility with collectors

vmoens:
## Description

When setting a ParallelEnv in a remote data collector and not embedding the env maker in an EnvCreator, an error is raised.
This PR solves this issue.


vmoens:
Should revamp this from scratch

PR #446, opened 2022-09-14
[CI, Doc] Update functorch source installation command

zou3519:
We changed how one installs functorch: functorch is now included in the PyTorch binary. This PR removes the old installation command which effectively does nothing right now. We'll be removing support for the old installation command after we migrate people off of it.

Test Plan:
- wait for CI

### DIFF
--- a/.circleci/unittest/linux/scripts/install.sh
+++ b/.circleci/unittest/linux/scripts/install.sh
@@ -33,12 +33,6 @@ else
     pip3 install --pre torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/nightly/cu113
 fi
 
-printf "Installing functorch\n"
-pip3 install ninja  # Makes the build go faster
-#pip3 install "git+https://github.com/pytorch/functorch.git"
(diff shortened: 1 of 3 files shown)


zou3519:
Looks like PyTorch didn't publish a nightly binary for mac yet, so this PR will have to wait until then

vmoens:
Wonderful! We'll need to update the readme.md and contributing.md accordingly I guess.
Will it work with nightly on M1 too?

zou3519:
> EDIT: macos still failing :/

Looks like the torchtext/torchvision nightlies have the 20220913 torch binary as a dependency. Should be fixed tomorrow, so will wait another day.

> Wonderful! We'll need to update the readme.md and contributing.md accordingly I guess.

Yup I can do that in this PR.

> Will it work with nightly on M1 too?

It should, but we have not tested functorch extensively on M1 yet. There's a high probability that it works correctly on M1 but we'll add explicit tests soon.

zou3519:
> Will it work with nightly on M1 too?

Confirmed that installing PyTorch from source on an M1 machine also installs functorch. So yes, we have M1 support now.

PR #448, opened 2022-09-14
[Feature]: Added TensorDictDefaultInitializer

nicolas-dufour:
## Description

This PR adds a module that can initialize a set of tensors to default values and add it to the given TD
## Motivation and Context

This change is required because we might want to init some tensors inside of a TDSequence. For example if the td is given from an env, we can want to add some other tensors. This could require to make changes to core methods such as collectors, rollouts et... By doing this, we can directly do the change inside a TDSequence

## Types of changes

## Checklist

### DIFF
--- a/test/test_tensordictmodules.py
+++ b/test/test_tensordictmodules.py
@@ -7,6 +7,7 @@ import argparse
 
 import pytest
 import torch
+from _utils_internal import get_available_devices
 from functorch import make_functional, make_functional_with_buffers
 from torch import nn
 from torchrl.data import TensorDict
@@ -21,6 +22,7 @@ from torchrl.modules import (
     TanhNormal,
     NormalParamWrapper,
 )
+from torchrl.modules.tensordict_module.initializer import TensorDictDefaultInitializer
 from torchrl.modules.tensordict_module.probabilistic import (
     ProbabilisticTensorDictModule,
 )
@@ -133,7 +135,7 @@ class TestTDModule:
                     dist_param_keys=dist_param_keys,
                     out_key_sample=["out"],
                     safe=safe,
-                    **kwargs
+                    **kwargs,
                 )
             return
         else:
@@ -143,7 +145,7 @@ class TestTDModule:
                 dist_param_keys=dist_param_keys,
                 out_key_sample=["out"],
                 safe=safe,
-                **kwargs
+                **kwargs,
             )
 
         td = TensorDict({"in": torch.randn(3, 3)}, [3])
@@ -249,7 +251,7 @@ class TestTDModule:
                     dist_param_keys=["loc", "scale"],
                     out_key_sample=["out"],
                     safe=safe,
-                    **kwargs
+                    **kwargs,
                 )
             return
         else:
@@ -259,7 +261,7 @@ class TestTDModule:
                 dist_param_keys=["loc", "scale"],
                 out_key_sample=["out"],
                 safe=safe,
-                **kwargs
+                **kwargs,
             )
 
         td = TensorDict({"in": torch.randn(3, 3)}, [3])
@@ -312,7 +314,7 @@ class TestTDModule:
                     dist_param_keys=["loc", "scale"],
                     out_key_sample=["out"],
                     safe=safe,
-                    **kwargs
+                    **kwargs,
                 )
             return
         else:
@@ -322,7 +324,7 @@ class TestTDModule:
                 dist_param_keys=["loc", "scale"],
                 out_key_sample=["out"],
                 safe=safe,
-                **kwargs
+                **kwargs,
             )
         tensordict_module, (
             params,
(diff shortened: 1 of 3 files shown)


test/test_tensordictmodules.py
```diff
@@ -1494,7 +1496,7 @@ def test_vmap_probabilistic(self, safe, spec_type):
                 out_key_sample=["out"],
                 dist_param_keys=["loc", "scale"],
                 safe=safe,
-                **kwargs
+                **kwargs,
```
vmoens, line 1499: why are all these changes necessary?

test/test_tensordictmodules.py
```diff
@@ -1701,6 +1703,100 @@ def test_subsequence_weight_update(self):
...
+    @pytest.mark.parametrize("batch_size", [[], [3], [5], [2, 3]])
+    @pytest.mark.parametrize("initializer", [torch.zeros, torch.ones, torch.randn])
+    @pytest.mark.parametrize("num_new_tensors", [1, 2, 3])
+    def test_td_initializer(self, device, batch_size, initializer, num_new_tensors):
+        td = TensorDict(
+            {"a": torch.randn(*batch_size, 3), "b": torch.randn(*batch_size, 4)},
+            batch_size,
+        ).to(device)
+        td_ref = td.clone()
+        default_td = {
```
vmoens, line 1718: why is it named `*_td`?

test/test_tensordictmodules.py
```diff
@@ -1701,6 +1703,100 @@ def test_subsequence_weight_update(self):
...
+            f"c_{i}": {"initializer": initializer, "shape": [2]}
+            for i in range(num_new_tensors)
+        }
+        initializer = TensorDictDefaultInitializer(default_td).to(device)
+        td = initializer(td)
+
+        assert torch.allclose(td["a"], td_ref["a"])
+        assert torch.allclose(td["b"], td_ref["b"])
+        for i in range(num_new_tensors):
+            assert f"c_{i}" in td.keys()
```
vmoens, line 1728: Ideally I guess you'd like to test the value

test/test_tensordictmodules.py
```diff
@@ -1701,6 +1703,100 @@ def test_subsequence_weight_update(self):
...
+            assert td[f"c_{i}"].shape == torch.Size([*batch_size, 2])
+            assert td[f"c_{i}"].device == device
+
+    @pytest.mark.parametrize("batch_size", [[], [3], [5], [2, 3]])
+    @pytest.mark.parametrize("initializer", [torch.zeros, torch.ones, torch.randn])
+    @pytest.mark.parametrize("num_new_tensors", [1, 2, 3])
+    def test_td_initializer_no_device(self, batch_size, initializer, num_new_tensors):
```
vmoens, line 1735: why do we need this? What info does it bring wrt the previous one?

torchrl/modules/tensordict_module/initializer.py
```diff
@@ -0,0 +1,112 @@
...
+        params: Optional[Union[TensorDictBase, List[Tensor]]] = None,
+        buffers: Optional[Union[TensorDictBase, List[Tensor]]] = None,
+        **kwargs,
+    ) -> TensorDictBase:
+        batch_size = tensordict.batch_size
+        device = tensordict.device_safe()
+        tensors = self.module(self.defaults_init, batch_size, device)
+        if not self.reinit:
+            tensors = {k: v for k, v in tensors.items() if k not in tensordict}
+        new_tensors_td = TensorDict(tensors, batch_size=batch_size)
```
vmoens, line 108: creating a TD is more expensive than a dict, perhaps we can just work with a dict?

vmoens (commented):
I see the utility in the case of Dreamer, but is there any other module that can have no input and has to figure out what to do?
I feel that this is a bit overkilling it, and if there are no other usages on the horizon I would put it on hold. I feel that users can create a non-generic TensorDictModule that does this if they need to.
TensorDictModule is restrictive but I guess it's fine: it can't read strings, its inputs should only be tensors etc. We should probably figure out how to make this more modular but I guess we also want to minimize the number of primitives we're supporting in case there's a need to refactor the whole class.

In the case that is ours (Dreamer) I feel that the problem is more of a design question:
the policy depends on the action at t-1, so even for the first action we need to have a 0th action available. I think a Transform could do that. We could do a transform that creates a tensor of arbitrary key, size, dtype and with a certain value that will be placed on the TensorDict when reset. There are multiple use cases for this: setting a 0th action, setting a 0th hidden state, setting an env-specific indicator (e.g. the task in multi-task settings, or the agent in multi-agent)

I'm not saying we should close this but I'd rather put it on hold until we have a stronger signal that this is needed.

Happy to hear your thoughts about it

nicolas-dufour:
Closing this, can be reopen if a need for this kind of tensordictmodule primitive arises

PR #456, opened 2022-09-15
[Feature]: `TensorDictPrimer` transform

nicolas-dufour:
## Description

This PR add a ForceTensorReset transform.

## Motivation and Context

ForceTensorReset allows to set or reset to default values some given vectors
## Types of changes

## Checklist


vmoens:
Lint!

vmoens:
`ForceTensorReset` is not very indicative IMO
What about 
- `DefaultTensorRest`
- `DefaultValueReset`
- `TensorDictPrimerReset`
- `TensorDictPrimer`
- `TensorDictPrimerTransform`

vmoens:
Added `TensorDictPrimer`
Let me know if that's a suitable solution. The big advantage is that the spec is registered in the env, and we can execute rollouts in parallel without problem as the tensordict will be properly pre-allocated.
If you're fine with it, can you remove `ForceTensorReset` as well as the tests?
You can rename the class if you feel like it :)

nicolas-dufour:
The transform is still needed to properly reset the states. Thank you for the changes to rollout!. I'll rename it to DefaultTensorReset if that's ok with you

vmoens:
Have you looked at my solution? It does reset things. What does ForceTensorReset do that TensorDictPrimer does not?

nicolas-dufour:
I'm not sure i understand how i can set vectors that aren't given by the env with TensorDictPrimer. Like prior_state and belief

vmoens:
Look at the docstring :)

nicolas-dufour:
Ah it's with the kwargs. Ok i will remove my tests and transform then

PR #468, opened 2022-09-20
Making the EnvBase.action_spec property point to _input_spec["action"]

### LINKED ISSUE #405: [Feature Request] Remove `env._action_spec` and make `env.action_spec` point to `env.input_spec["action"]`
## Motivation

We keep two copies of "action_spec", one in `env._action_spec` and one in `env.input_spec["action"]`.
When applying transformations, we change both independently and they can cease to point to the same object.

We should refactor this such that `env.action_spec` always points to `env.input_spec["action"]`. When setting `env.action_spec = smth`, under the hood the env should take care of setting the action spec to `env.input_spec`.

## Solution

All refs to `env._action_spec` must be removed. 
A pointer should be implemented from `def action_spec(self)` to `self.input_spec["action"]` and from the setter to the `input_spec` setter.

In the transforms (torchrl.envs.transforms), we just need to keep the transformation of `action_spec` (not the transformation of the "action_spec" value from "input_spec"), i.e. we want the action spec to be transformed only once.

JulienB-78:
## Description

EnvBase._action_spec has been remove and the action_spec property now points to _input_spec["action"]

## Motivation and Context

Close #405

- [x] I have raised an issue to propose this change ([required](https://github.com/facebookresearch/rl/issues) for new features and bug fixes)

## Types of changes

## Checklist


torchrl/envs/transforms/transforms.py
```diff
@@ -358,21 +344,12 @@ def observation_spec(self) -> TensorSpec:
...
-        if self._action_spec is None or not self.cache_specs:
-            action_spec = self.transform.transform_action_spec(
-                deepcopy(self.base_env.action_spec)
-            )
-            if self.cache_specs:
-                self._action_spec = action_spec
-        else:
-            action_spec = self._action_spec
-        return action_spec
+        """Action spec of the transformed environment"""
```
vmoens, line 347: Do you think we still need this, as TransformedEnv inherits from EnvBase I guess the end result will be the same

torchrl/envs/transforms/transforms.py
```diff
@@ -539,9 +516,22 @@ def __setattr__(self, key, value):
         if isinstance(value, Transform):
             value.set_parent(self)
         if isinstance(propobj, property):
-            if propobj.fset is None:
+            ancestors = list(__class__.__mro__)[::-1]
```
vmoens, line 519: I wonder what this change does? What's the reason for this?

torchrl/envs/transforms/transforms.py
```diff
@@ -539,9 +516,22 @@ def __setattr__(self, key, value):
...
+            while isinstance(propobj, property):
+                if propobj.fset is not None:
+                    return propobj.fset(self, value)
+                propobj = getattr(ancestors.pop(), key, None)
+            else:
                 raise AttributeError(f"can't set attribute {key}")
-            return propobj.fset(self, value)
+
+        # if isinstance(propobj, property):
```
vmoens, line 527: do we want to keep this commented block?

vmoens (approved):
LGTM, just a couple of comments for clarification

torchrl/envs/transforms/transforms.py
JulienB-78: If the logic with cache_specs is not mandatory, we could indeed removing it. As a side not this property as no setter so, I am not mistaken, action_spec cannot be be set in a TranformedEnv. (I don't think the setter is inherited if the property is redefined in the child class)

torchrl/envs/transforms/transforms.py
JulienB-78: The way it was done before did not work if the property was inherited. getattr(self.__class__, key, None) return None if the property is not defined in self.__class__ but in own of its parents.

torchrl/envs/transforms/transforms.py
JulienB-78: No, that should go away.

PR #471, opened 2022-09-21
Struct hydra collector

vmoens:
## Description

Create configs for collectors.


nicolas-dufour (commented):
I would not do this like this. The "hydra way" of doing this is to have the following:
1 config.yaml with the following
```
defaults:
    collector:  sync_single
num_workers: 0
```
Then in a folder named collector we have sync_single.yaml of the form:
```
 _target_: torchrl.collectors.SyncDataCollector
  device: "cpu"
  passing_device: "cpu"
  # device on which the data collector should store the trajectories to be passed to this script.
  # If the collector device differs from the policy device (cuda:0 if available), then the
  # weights of the collector policy are synchronized with collector.update_policy_weights_().
  pin_memory: False
  # if True, the data collector will call pin_memory before dispatching tensordicts onto the passing device
  init_with_lag: False
  # if True, the first trajectory will be truncated earlier at a random step. This is helpful
  # to desynchronize the environments, such that steps do no match in all collected
  # rollouts. Especially useful for online training, to prevent cyclic sample indices.
  frames_per_batch: 200
  # number of steps executed in the environment per collection.
  # This value represents how many steps will the data collector execute and return in *each*
  # environment that has been created in between two rounds of optimization
  # (see the optim_steps_per_batch above).
  # On the one hand, a low value will enhance the data throughput between processes in async
  # settings, which can make the accessing of data a computational bottleneck.
  # High values will on the other hand lead to greater tensor sizes in memory and disk to be
  # written and read at each global iteration. One should look at the number of frames per second
  # in the log to assess the efficiency of the configuration.
  total_frames: 5000000
  # total number of frames collected for training. Does account for frame_skip (i.e. will be
  # divided by the frame_skip). Default=5e6.
  seed: 42
  # seed used for the environment, pytorch and numpy.
  exploration_mode: "random"
  # exploration mode of the data collector
```
and the same structure for the rest of the files
then, we always load the same config.yaml but we then can change the collector with overides. In the command line you can do python main.py collector = the_collector_you want. In python we can also override using hydra.compose with the overrides argument

shagunsodhani (approved):

vmoens:
> ```
> num_workers: 0
> ```

But the `num_workers` is tied to the collector config, though it's not an argument of the collector. How can I make sure that if someone just says she wants the async collector the number of workers follows accordingly? Because in your example if you want the async collector you need to do
```
python foo.py collector=async.yaml num_workers=2
```
but hydra should be able to figure out that num_workers is 2 if this config is being used...

shagunsodhani:
> But the `num_workers` is tied to the collector config, though it's not an argument of the collector. 

In what sense is it tied then ?

vmoens:
> > But the `num_workers` is tied to the collector config, though it's not an argument of the collector.
> 
> In what sense is it tied then ?

When you're creating the collector, you tell it where the policy has to be run like this
```
devices=[cuda:1, cuda:0]
```
which means: run the first process on cuda 1 and the second on cuda 0.
Another valid devices arg would be
```
devices=cuda:0
make_env_fn=[make_env1, make_env2]
```
In that case the collector "knows" that this means that all policies must be run on cuda:0. It also knows that 2 envs have to be created and run in parallel processes.
In those two cases, the `num_workers` is going to be 2.

When we create the collector, we need to pass a list of env constructor that depends on num_workers.

Those things are a bit entangled. In summary, we need the external world to know how many processes are needed too. But `num_workers` must match the args of the collector, otherwise things will collide.

nicolas-dufour:
> > ```
> > num_workers: 0
> > ```
> 
> But the `num_workers` is tied to the collector config, though it's not an argument of the collector. How can I make sure that if someone just says she wants the async collector the number of workers follows accordingly? Because in your example if you want the async collector you need to do
> 
> ```
> python foo.py collector=async.yaml num_workers=2
> ```
> 
> but hydra should be able to figure out that num_workers is 2 if this config is being used...

@vmoens i had this issue a while ago and raised it to the devs. Here is there answer: https://github.com/facebookresearch/hydra/issues/1774

nicolas-dufour (commented):
LGTM. The only issue to resolve is to tie num_workers to the required collector.
There is a hacky way to do it. add it as a kwarg to the different collector functions and then we can access it with cfg.collector.num_workers.
or we can have inside each collector file the following structure:
```
num_workers: 0
constructor: 
   _target_: torchrl.collectors.SyncDataCollector
  args: blablalba
```
and then do instantiate(cfg.collector.constructor) instead

vmoens:
why not... in this particular case I think we could also refactor the collector to store num_workers and do stuff with it, then every piece of code would access this through `collector.num_workers` and that's it
so I'd suggest to leave it as is for now

PR #472, opened 2022-09-21
Structured hydra env

vmoens:
## Description

Implements envs and transforms instantiation from config


examples/configs/env/dmcontrol_pixels.yaml
```diff
@@ -0,0 +1,12 @@
+# by default we map the double-prec tensors to single-prec
+_target_: torchrl.envs.TransformedEnv
+env:
+  _target_: torchrl.envs.libs.dm_control.DMControlEnv
+  env_name: cheetah
```
nicolas-dufour, line 5: I think the envs should also be define with a default and a dir with configs for every env. I also don't think we need to have 2 different files for from pixels and state since the only attribute to change is from pixels which is easy to override.

nicolas-dufour: maybe we can have a dir called env_base and then we can do 
```yaml
_target_: torchrl.envs.TransformedEnv
env: ${env_base}
```

vmoens: > I think the envs should also be define with a default and a dir with configs for every env. I also don't think we need to have 2 different files for from pixels and state since the only attribute to change is from pixels which is easy to override.

not sure about these points
For the first, I think that this is not scalable: there are tens of gym envs and dm_control envs and we don't want to write a file for each. 
For the second point I'm not sure that this will stay as simple on the long run. But happy to chat more about this :)

vmoens: > I think the envs should also be define with a default and a dir with configs for every env. I also don't think we need to have 2 different files for from pixels and state since the only attribute to change is from pixels which is easy to override.

Still about this: the first point seems to indicate we need more files (but they will contain a single or a couple of lines: `env_name: HalfCheetah-v4` for instance). The second point to the fact that we already have too many files. I guess we don't want to have too many files to choose from no?

nicolas-dufour: My point about not having a pixel file and a state file is more about avoinding the redundancy since the change in config is quite explicit (it's a boolean, with a variable name which is quite explicit). Also when what happens when doing embodied AI for example?

For me having separate files for each env is more beginner-friendly. If you are not familiar with gym you don't know the exact syntax of the envs. I don't see the problem in having multiple file. Also this is quite useful if we beggin to have env specific configs (ex frame_skip can have default values which are "standard".

PR #485, opened 2022-09-23
[Feature] Replaced `device_safe()` with `device`

### LINKED ISSUE #455: [Feature Request] Removing `TensorDict.device_safe()` in favour of `TensorDict.device`
## Motivation

TensorDict device can now be None even when populated in a great move from @tcbegley
The current TensorDict's API will return an error in one calls `td.device` when None is set. To get this device, one should call `td.safe_device()`.

Pros:
- Users are notified that they're playing with fire
- consistent with torch in that `device` is always a `torch.device` if it exists.
- `tensor.to(tensordict.device)` won't work if the device is not set 
    -> but if the goal is to put the tensor in the tensordict, then `tensordict[key] = tensor` will already do the casting if needed

Cons:
- Inconsistent with torch
    - no code re-usability (a call to `tensor.device` must be replaced by `td.device_safe()`
    - requires some kind of learning from the users
- Pollutes the code with a lot of `if hasattr(obj, "device_safe")` and such
- In many cases a call to `tensordict.device != torch.device(smth)` is actually legit: if the device is not set, we can expect that things may break. Most of the time this will be followed by `tensordict.to(smth)` which will be an (almost) no-op if all tensors are already on device.

## Solution

Removing `device_safe` and replacing it by `device`.

ordinskiy:
`device_safe()` was removed from `TensorDict` classes. It was replaced with `device`, which no longer throws an exception when `_device` is `None`.

## Description

- `device_safe()` method was removed from all `TensorDict` classes.
- `TensorDict.device` property no longer throws an exception and returns `None` if `self._device` is `None`.
- `SavedTensorDict` implicitly copies `device` of the source dictionary, unless `device` is provided explicitly. 

## Motivation and Context

close #455  

## Types of changes

## Checklist


ordinskiy:
Failed CI reports next error:
```
AttributeError: module 'functorch._C' has no attribute 'is_batchedtensor'
```
There is no direct connection between changes in this pull request with this functionality. I don't see related open issues either. What is the the procedure in this case?

vmoens:
> Failed CI reports next error:
> 
> ```
> AttributeError: module 'functorch._C' has no attribute 'is_batchedtensor'
> ```
> 
> There is no direct connection between changes in this pull request with this functionality. I don't see related open issues either. What is the the procedure in this case?

you can just merge main, this has just been fixed

torchrl/data/tensordict/tensordict.py
```diff
@@ -1921,17 +1904,8 @@ def batch_dims(self, value: COMPATIBLE_TYPES) -> None:
...
-        if device is None:
-            raise RuntimeError(
-                "Querying a tensordict device when it has not been set is not "
-                "permitted. Use `tensordict.device_safe()` return the device "
-                "value or `None` if the device is not set.\n\nSet the device "
-                "upon creation using the `device=dest` keyword, or "
-                "subsequently using `tensordict.to(device)."
-            )
-        return device
+    def device(self) -> Union[None, torch.device]:
```
vmoens, line 1907: Could you add a note here in the docstring, mentioning that we allow device to be None if it is not specified upon construction or when calling `tensordict.to(...)`?

vmoens (approved):
Wonderful! Thanks a million for this
Just a small comment in the call to the device property, having a docstring there would be a great plus

torchrl/data/tensordict/tensordict.py
ordinskiy: Added the comment:
        """Returns `None` if device hasn't been provided in the constructor
        or set via `tensordict.to(device)`.
        """

PR #489, opened 2022-09-26
[Refactoring] Minor tweaks to recorder and logger

nicolas-dufour:
## Description

Minor changes to recorder and logger.

## Checklist

### DIFF
--- a/torchrl/trainers/trainers.py
+++ b/torchrl/trainers/trainers.py
@@ -932,7 +932,6 @@ class Recorder:
     def __call__(self, batch: TensorDictBase) -> Dict:
         out = None
         if self._count % self.record_interval == 0:
-            torch.cuda.empty_cache()
             with set_exploration_mode(self.exploration_mode):
                 if isinstance(self.policy_exploration, torch.nn.Module):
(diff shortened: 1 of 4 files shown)


torchrl/trainers/trainers.py
```diff
@@ -933,12 +932,11 @@ def __init__(
     def __call__(self, batch: TensorDictBase) -> Dict:
         out = None
         if self._count % self.record_interval == 0:
+            torch.cuda.empty_cache()
```
vmoens, line 935: I think we added this because of weird cuda mem issues which I believe are solved now. Can we remove this and try to train something like dreamer while looking at the mem on devices? 
Asking this because this operation takes a considerable amount of time :)

vmoens (commented):
Let's see what we should do about the `empty_cache()`, otherwise LGTM

torchrl/trainers/trainers.py
nicolas-dufour: yes i agree let's remove this

vmoens (approved):

PR #490, opened 2022-09-26
[Feature]: Deactivate typechecks in envs

nicolas-dufour:
## Description

Allows to deactivate typechecks in envs
## Motivation and Context

We want to be able to deactivates typechecks since it can block code execution when the dtype gets changes (ex autocast to float16)

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/model_based/common.py
+++ b/torchrl/envs/model_based/common.py
@@ -97,6 +97,7 @@ class ModelBasedEnvBase(EnvBase, metaclass=abc.ABCMeta):
         device (torch.device, optional): device where the env input and output are expected to live
         dtype (torch.dtype, optional): dtype of the env input and output
         batch_size (torch.Size, optional): number of environments contained in the instance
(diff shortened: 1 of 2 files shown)


torchrl/envs/model_based/common.py
```diff
@@ -118,7 +118,7 @@ def __init__(
         batch_size: Optional[torch.Size] = None,
     ):
         super(ModelBasedEnvBase, self).__init__(
-            device=device, dtype=dtype, batch_size=batch_size
+            device=device, dtype=dtype, batch_size=batch_size, run_type_checks=False
```
vmoens, line 121: Let's make this an art that defaults to false no?

nicolas-dufour: an art?

nicolas-dufour: I guess you meant i put it as an attribute in the init

vmoens (approved):
LGTM

PR #500, opened 2022-09-30
[Refactoring] Import at root to enable vmap monkey-patching

vmoens:
## Description

- monkey-patches a few more functorch functions
- implements tests for `nn.Module` vmap with tensordicts

Following suggestions from @zou3519 we remove and add a batch dimension when vmapping over a tensordict.
Some commended code shows attempts to register TensorDIct in pytree but that failed + it requires to decompose and recompose the tensordict multiple times, while doing complex operations to keep track of the modified batch size. Besides the fact that it did not work, it also seemed suboptimal from a compute time perspective.

### DIFF
--- a/test/_utils_internal.py
+++ b/test/_utils_internal.py
@@ -8,7 +8,7 @@ import os
 # Get relative file path
 # this returns relative path from current file.
 import torch.cuda
-from torchrl import seed_generator
+from torchrl._utils import seed_generator
 
 
 def get_relative_path(curr_file, *path_components):
--- a/test/mocking_classes.py
+++ b/test/mocking_classes.py
@@ -6,7 +6,7 @@ from typing import Optional
 
 import torch
 import torch.nn as nn
-from torchrl import seed_generator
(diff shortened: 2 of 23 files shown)


vmoens:
cc @nicolas-dufour 
allows for this kind of behaviour
```python
        class MyModule(torch.nn.Module):
            def forward(self, tensordict, tensor):
                a = tensordict["a"]
                return (
                    TensorDict(
                        {"a": a}, tensordict.batch_size, device=tensordict.device
                    ),
                    tensor,
                )

        tensor = torch.randn(3).expand(4, 3)
        tensordict = TensorDict({"a": torch.randn(3, 1)}, [3])
        out = vmap(MyModule(), (None, 0))(tensordict, tensor)

        assert out[0].shape == torch.Size([4, 3])
        assert out[1].shape == torch.Size([4, 3])
        assert out[0]["a"].shape == torch.Size([4, 3, 1])
```

allowing for nn.Module with arbitrary TensorDict inputs

PR #504, opened 2022-10-03
[Feature] Add info dict key-spec pairs to observation_spec

### LINKED ISSUE #453: [Feature Request] Add info dict key-spec pairs to `env.observation_spec` if they're provided to a gym environment through `set_info_dict_reader`
## Motivation

gym's environments can pass some more info through a dictionary `info`:
`obs, reward, done, terminated = env.step(action)`.
This dictionary can contain additional information about the state of the env: time spent doing the task, whether the task is solved etc.
We wrap the obs, reward and done in the output `TensorDict`. We can also place the info if they're numerical values (see #234):
```python
        >>> from torchrl.envs.libs.gym import GymWrapper
        >>> from torchrl.envs import default_info_dict_reader
        >>> reader = default_info_dict_reader(["my_info_key"])
        >>> # assuming "some_env-v0" returns a dict with a key "my_info_key"
        >>> env = GymWrapper(gym.make("some_env-v0"))
        >>> env.set_info_dict_reader(info_dict_reader=reader)
        >>> tensordict = env.reset()
        >>> tensordict = env.rand_step(tensordict)
        >>> assert "my_info_key" in tensordict.keys()
```

Problem is that we also register what are the expected observations (domain, dtype, device etc) in an `observation_spec` attribute. It is important for us to know what to expect as output from the env. However the info is not yet registered in `observation_spec`. This is a problem for parallel environments as we would wish to pre-allocate the tensors that we place in share memory to pass information from one process to another: for that, we use the `observation_spec` which is cheaper than resetting the env.

## Solution

We should be able to provide the specs corresponding to the info keys in the `default_info_dict_reader`, and the `set_info_dict_reader` should be able to read them. If no spec is provided, a default unidimensional, floating-point unbounded spec should be assumed.

tcbegley:
## Description

Specs are optionally provided via the `info_spec` property of the `info_dict_reader` (all readers should inherit from `BaseInfoDictReader`). When the `info_dict_reader` is set, `GymLikeEnv` recovers the specs and adds them to `observation_spec`.

I would appreciate some guidance on how / where best to test these changes.

## Motivation and Context

Closes #453 

## Types of changes

## Checklist


torchrl/envs/gym_like.py
```diff
@@ -39,11 +57,30 @@ class default_info_dict_reader:
...
+                    "If specifying specs for info keys with a sequence, the "
+                    "length of the sequence must match the number of keys"
+                )
+            self._info_spec = dict(zip(self.keys, spec))
+        else:
+            if spec is None:
+                spec = {}
+
+            self._info_spec = {
+                key: spec.get(key, UnboundedContinuousTensorSpec()) for key in self.keys
```
tcbegley, line 81: MyPy gives me a warning here because `TensorSpec` has abstract method `index` which `UnboundedContinuousTensorSpec` does not define.

test/test_env.py
```diff
@@ -1121,6 +1123,17 @@ def test_batch_unlocked_with_batch_size(device):
...
 
+@pytest.mark.skipif(not _has_gym, reason="no gym")
+def test_info_dict_reader(seed=0):
+    import gym
+
+    env = GymWrapper(gym.make("HalfCheetah-v4"))
+    env.set_info_dict_reader(default_info_dict_reader(["x_position"]))
+
+    assert "x_position" in env.observation_spec.keys()
+    assert isinstance(env.observation_spec["x_position"], UnboundedContinuousTensorSpec)
```
tcbegley, line 1134: is there something else I could be checking here?

vmoens: you could do `assert env.observation_spec["x_position"].is_in(tensordict["x_position"])`

vmoens: it's essentially testing that x_position is what it's supposed to be 
Another thing we could do is set up a screwed up tensor spec (e.g. one that has wrong shape or wrong dtype) and check that this assertion raises an error

tcbegley: I just added some additional tests as you suggested, although it wasn't raising an error at any stage, just returning false for 

```python
env.observation_spec["x_position"].is_in(tensordict["x_position"])
```

Would you have expected an exception at some point?

vmoens: no but yes if you put an assert if front of it :p

test/test_env.py
```diff
@@ -1121,6 +1123,34 @@ def test_batch_unlocked_with_batch_size(device):
...
+    env2.set_info_dict_reader(
+        default_info_dict_reader(
+            ["x_position"], spec={"x_position": OneHotDiscreteTensorSpec(5)}
+        )
+    )
+
+    tensordict2 = env2.reset()
+    tensordict2 = env2.rand_step(tensordict2)
+
+    assert not env.observation_spec["x_position"].is_in(tensordict["x_position"])
```
vmoens, line 1151: oddly this seems to pass, I wonder why it is...

tcbegley: I think it's just a dumb typo: `env` -> `env2` 

>_<

vmoens (approved):
LGTM, thanks for this contribution!

PR #511, opened 2022-10-03
[BugFix] making first_dim and last_dim negative in FlattenObservation when a parent is set

### LINKED ISSUE #510: [BUG] FlattenObservation specs cannot be retrieved when first_dim!=0
## Describe the bug

`FlattenObservation` transformed features and specs conflicts when the environment has a batch size and the dimensions are greater than 0.

## To Reproduce

Steps to reproduce:

```python
>>> env = TransformedEnv(ParallelEnv(2, lambda:DMControlEnv("cheetah", "run", from_pixels=True)), FlattenObservation(1, -1, keys_in=["next_pixels"]))
>>> env.observation_spec["next_pixels"].shape
torch.Size([240, 960])
>>> env.reset()
TensorDict(
    fields={
        done: Tensor(torch.Size([2, 1]), dtype=torch.bool),
        pixels: Tensor(torch.Size([2, 230400]), dtype=torch.uint8),
        position: Tensor(torch.Size([2, 8]), dtype=torch.float64),
        velocity: Tensor(torch.Size([2, 9]), dtype=torch.float64)},
    batch_size=torch.Size([2]),
    device=cpu,
    is_shared=False)
 ```
Hence the flattened `pixels` don't have the expect shape.

## Expected behavior

The transformed feature shape and the transformed observation spec should match.

### PR DESCRIPTION
**Pull request by vmoens**
## Description

Solves and closes #510

### DIFF
--- a/torchrl/trainers/helpers/envs.py
+++ b/torchrl/trainers/helpers/envs.py
@@ -129,7 +129,7 @@ def make_env_transforms(
         env.append_transform(Resize(cfg.image_size, cfg.image_size))
         if cfg.grayscale:
             env.append_transform(GrayScale())
-        env.append_transform(FlattenObservation(first_dim=0))
+        env.append_transform(FlattenObservation())
         env.append_transform(CatFrames(N=cfg.catframes, keys_in=["next_pixels"]))
         if stats is None:
             obs_stats = {"loc": 0.0, "scale": 1.0}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: torchrl/trainers/helpers/envs.py [hunk 1]
```diff
@@ -129,7 +129,7 @@ def make_env_transforms(
         env.append_transform(Resize(cfg.image_size, cfg.image_size))
         if cfg.grayscale:
             env.append_transform(GrayScale())
-        env.append_transform(FlattenObservation(first_dim=batch_dims))
+        env.append_transform(FlattenObservation(first_dim=0))
```
vmoens, line 132: @nicolas-dufour perhaps we could keep the default values here?

PR #515, opened 2022-10-04
[BugFix] Changing the dm_control import to fail if not installed

### LINKED ISSUE #506: [BUG] importing `dm_control` environments when the library is not installed does not display the right error message
## Describe the bug

We allow users to load the file `envs/libs/dm_control.py` even if `dm-control` is not installes. When creating an object in this situation, the error message now is that the environment configuration cannot be found. An `ImportError` should be raised instead (after a check for `_has_dmc` has been made).

zeenolife:
## Description

This PR handles the import of the `torchrl.envs.libs.dm_control`, when the `dm_control` library is not installed. Specifically, it raises the `ImportError`.

## Motivation and Context

close #506 

## Types of changes

## Checklist

- [ ] ? My change requires a change to the documentation.
- [ ] ? I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] ? I have updated the documentation accordingly.

### DIFF
--- a/torchrl/envs/libs/dm_control.py
+++ b/torchrl/envs/libs/dm_control.py
@@ -35,8 +35,9 @@ try:
 
     _has_dmc = True
 
-except ImportError:
-    _has_dmc = False
+except ImportError as error:
+    print(error)
+    raise
 
 __all__ = ["DMControlEnv", "DMControlWrapper"]
 
--- a/torchrl/envs/libs/dm_control.py
+++ b/torchrl/envs/libs/dm_control.py
@@ -35,7 +35,7 @@ try:
 
     _has_dmc = True
 
-except ImportError as error:
+except ImportError:
     _has_dmc = False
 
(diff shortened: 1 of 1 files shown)


torchrl/envs/libs/dm_control.py
```diff
@@ -272,6 +272,11 @@ class DMControlEnv(DMControlWrapper):
     """
 
     def __init__(self, env_name, task_name, **kwargs):
+        if not _has_dmc:
+            raise ImportError(
+                "dm_control python package was not found."
+                "Please install this dependency."
+            )
```
zeenolife, line 279: I think it's better to check it with `_has_dmc` compared to `self.available_envs`, because it's easier to track why `_has_dmc` is `False`

vmoens: you're absolutely right!

vmoens (approved):
LGTM thanks for this, let's wait a couple of minutes for the tests to pass

PR #521, opened 2022-10-05
[BugFix] Restore missing keys in data collector output

### LINKED ISSUE #505: [BUG] data collectors may omit some keys in the output tensordict
## Describe the bug

Remote data collector will populate the output tensordict with the data that come with the first collection in the environment. There are several issues with this, the main being that if we use a value of `init_random_frames` > 0, the initial tensordict will just be made out of the environment keys + an action key + some utils from the collector (trajectory id and such).

## To Reproduce

We can create a policy that uses an RNN to reproduce this.
```python
from torchrl.collectors import MultiaSyncDataCollector
from torchrl.modules import TensorDictModule, LSTMNet
import torch

out_features=1
batch = 7
time_steps = 6
hidden_size = 12
net = LSTMNet(
    out_features,
    {"input_size": hidden_size, "hidden_size": hidden_size},
    {"out_features": hidden_size},
)

policy = TensorDictModule(net, ["observation", "hidden1", "hidden2"], ["action", "hidden1", "hidden2", "next_hidden1", "next_hidden2"])

env_maker = lambda: GymEnv("Pendulum-v1")

policy(env_maker().reset())
```

### Case 1: no `init_random_frames`

Here everything works as expected
```python
c = MultiaSyncDataCollector([env_maker], policy, total_frames=200, frames_per_batch=20, init_random_frames=0)

for i, b in enumerate(c):
    print(b)
    if i == 5:
        # by now we should have a tensordict that is the output of the policy
        break
```
Every tensordict look like this

```
TensorDict(
    fields={
        action: Tensor(torch.Size([1, 20, 1]), dtype=torch.float32),
        done: Tensor(torch.Size([1, 20, 1]), dtype=torch.bool),
        hidden1: Tensor(torch.Size([1, 20, 1, 12]), dtype=torch.float32),
        hidden2: Tensor(torch.Size([1, 20, 1, 12]), dtype=torch.float32),
        mask: Tensor(torch.Size([1, 20, 1]), dtype=torch.bool),
        next_hidden1: Tensor(torch.Size([1, 20, 1, 12]), dtype=torch.float32),
        next_hidden2: Tensor(torch.Size([1, 20, 1, 12]), dtype=torch.float32),
        next_observation: Tensor(torch.Size([1, 20, 3]), dtype=torch.float32),
        observation: Tensor(torch.Size([1, 20, 3]), dtype=torch.float32),
        reward: Tensor(torch.Size([1, 20, 1]), dtype=torch.float32),
        step_count: Tensor(torch.Size([1, 20, 1]), dtype=torch.int32),
        traj_ids: Tensor(torch.Size([1, 20, 1, 1]), dtype=torch.int64)},
    batch_size=torch.Size([1, 20]),
    device=cpu,
    is_shared=False)
```

### Case 2: `init_random_frames` > 0

When running this
```python
c = MultiaSyncDataCollector([env_maker], policy, total_frames=200, frames_per_batch=20, init_random_frames=50)

for i, b in enumerate(c):
    print(b)
    if i == 5:
        # by now we should have a tensordict that is the output of the policy
        break
```
every tensordict look like this

```
TensorDict(
    fields={
        action: Tensor(torch.Size([1, 20, 1]), dtype=torch.float32),
        done: Tensor(torch.Size([1, 20, 1]), dtype=torch.bool),
        mask: Tensor(torch.Size([1, 20, 1]), dtype=torch.bool),
        next_observation: Tensor(torch.Size([1, 20, 3]), dtype=torch.float32),
        observation: Tensor(torch.Size([1, 20, 3]), dtype=torch.float32),
        reward: Tensor(torch.Size([1, 20, 1]), dtype=torch.float32),
        step_count: Tensor(torch.Size([1, 20, 1]), dtype=torch.int32),
        traj_ids: Tensor(torch.Size([1, 20, 1, 1]), dtype=torch.int64)},
    batch_size=torch.Size([1, 20]),
    device=cpu,
    is_shared=False)
```
i.e. the `hidden` and `next_hidden` keys are missing **even when we have passed the number of frames indicated by `init_random_frames`**.

### Case 3: `init_random_frames` > 0 with `SyncDataCollector`

`SyncDataCollector` is the simplest data collector: it does not spawn other processes. 

```python
c = SyncDataCollector(env_maker, policy, total_frames=200, frames_per_batch=20, init_random_frames=50)

for i, b in enumerate(c):
    print(b)
    if i == 5:
        # by now we should have a tensordict that is the output of the policy
        break
```
The first three tensordicts look like this
```
TensorDict(
    fields={
        action: Tensor(torch.Size([1, 20, 1]), dtype=torch.float32),
        done: Tensor(torch.Size([1, 20, 1]), dtype=torch.bool),
        mask: Tensor(torch.Size([1, 20, 1]), dtype=torch.bool),
        next_observation: Tensor(torch.Size([1, 20, 3]), dtype=torch.float32),
        observation: Tensor(torch.Size([1, 20, 3]), dtype=torch.float32),
        reward: Tensor(torch.Size([1, 20, 1]), dtype=torch.float32),
        step_count: Tensor(torch.Size([1, 20, 1]), dtype=torch.int32),
        traj_ids: Tensor(torch.Size([1, 20, 1, 1]), dtype=torch.int64)},
    batch_size=torch.Size([1, 20]),
    device=cpu,
    is_shared=False)
```
after that, they start being populated with the hidden keys:

```
TensorDict(
    fields={
        action: Tensor(torch.Size([1, 20, 1]), dtype=torch.float32),
        done: Tensor(torch.Size([1, 20, 1]), dtype=torch.bool),
        hidden1: Tensor(torch.Size([1, 20, 1, 12]), dtype=torch.float32),
        hidden2: Tensor(torch.Size([1, 20, 1, 12]), dtype=torch.float32),
        mask: Tensor(torch.Size([1, 20, 1]), dtype=torch.bool),
        next_hidden1: Tensor(torch.Size([1, 20, 1, 12]), dtype=torch.float32),
        next_hidden2: Tensor(torch.Size([1, 20, 1, 12]), dtype=torch.float32),
        next_observation: Tensor(torch.Size([1, 20, 3]), dtype=torch.float32),
        observation: Tensor(torch.Size([1, 20, 3]), dtype=torch.float32),
        reward: Tensor(torch.Size([1, 20, 1]), dtype=torch.float32),
        step_count: Tensor(torch.Size([1, 20, 1]), dtype=torch.int32),
        traj_ids: Tensor(torch.Size([1, 20, 1, 1]), dtype=torch.int64)},
    batch_size=torch.Size([1, 20]),
    device=cpu,
    is_shared=False)
```

## Proposed solution

The expected behaviour should be that every tensordict has all possible keys **from the first batch** (values that are not valid should be populated with 0s) for all data collectors (note that under the hood every collector will call `SyncDataCollector` hence only this class needs to be corrected).

The reason for this is that we will store the tensordict in a replay buffer and we need to know what key-value pairs will need to be stored in it from the first collection.

What we want is that after instantiating the env, the `self._tensordict_out` object should be populated with all the relevant keys.

The collector should make the following operations:
1. If the policy has an non-empty `policy.spec` attribute **and** all the keys are not None (`policy.spec` is a `CompositeSpec` object that can have values set to None) **and** the keys in the spec match the `out_keys` of the policy,  then we assume that the user has given all the relevant information about it. We just need to do 
```python
self._tensordict_out = env.reset()
self._tensordict_out.update(policy.action_spec.rand())
```
If some values in the `spec` are None or if the policy `out_keys` are not all present in `policy.spec`, we would revert to (2) here below.

2. If the policy does not have a spec, then we should run a mini-rollout in the env with the policy to make sure that we have all the relevant keys:
```python
self._tensordict_out = env.rollout(3, policy)[..., 0]
```

Finally, if at some point there are more keys in the tensordict collected in `SyncDataCollector.rollout` than in the `SyncDataCollector._tensordict_out` object we should raise a warning to the user informing that some data is likely to be missed during collection for the reasons explained above.

```

### PR DESCRIPTION
**Pull request by tcbegley**
## Description

This PR fixes a bug with data collectors that saw some keys being omitted from the output when e.g. random actions were applied during initialisation.

Data collectors will now attempt to determine keys from the policy spec if it is supplied and complete, otherwise they will perform a short rollout with the policy to determine which keys will be returned.

## Motivation and Context

Closes #505 

## Types of changes

## Checklist

I still need to add a test, currently seeing an error on one particular test I don't understand.

### DIFF
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -674,14 +674,17 @@ def test_collector_vecnorm_envcreator(static_seed):
 
 
 @pytest.mark.parametrize("use_async", [False, True])
-@pytest.mark.skipif(torch.cuda.device_count() <= 1, reason="no cuda device found")
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="no cuda device found")
 def test_update_weights(use_async):
-    policy = torch.nn.Linear(3, 4).cuda(1)
+    n_actions = ContinuousActionVecMockEnv().action_spec.shape[-1]
+    policy = TensorDictModule(
+        torch.nn.LazyLinear(n_actions), in_keys=["observation"], out_keys=["action"]
+    )
     collector_class = (
         MultiSyncDataCollector if not use_async else MultiaSyncDataCollector
     )
     collector = collector_class(
-        [lambda: DiscreteActionVecMockEnv()] * 3,
+        [lambda: ContinuousActionVecMockEnv()] * 3,
         policy=policy,
         devices=[torch.device("cuda:0")] * 3,
         passing_devices=[torch.device("cuda:0")] * 3,
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -787,7 +787,6 @@ def test_excluded_keys(collector_class, exclude):
 @pytest.mark.parametrize(
     "collector_class",
     [MultiaSyncDataCollector, MultiSyncDataCollector, SyncDataCollector],
-    # [SyncDataCollector],
 )
 @pytest.mark.parametrize("init_random_frames", [0, 50])
 @pytest.mark.parametrize("explicit_spec", [True, False])
@@ -812,7 +811,7 @@ def test_collector_output_keys(collector_class, init_random_frames, explicit_spe
         "out_keys": ["action", "hidden1", "hidden2", "next_hidden1", "next_hidden2"],
     }
     if explicit_spec:
-        hidden_spec = NdUnboundedContinuousTensorSpec(hidden_size)
+        hidden_spec = NdUnboundedContinuousTensorSpec((1, hidden_size))
         policy_kwargs["spec"] = CompositeSpec(
             action=UnboundedContinuousTensorSpec(),
             hidden1=hidden_spec,
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -25,6 +25,11 @@ from torchrl.collectors.collectors import (
     MultiSyncDataCollector,
     MultiaSyncDataCollector,
 )
+from torchrl.data import (
+    CompositeSpec,
+    NdUnboundedContinuousTensorSpec,
+    UnboundedContinuousTensorSpec,
+)
 from torchrl.data.tensordict.tensordict import assert_allclose_td
 from torchrl.envs import EnvCreator
 from torchrl.envs import ParallelEnv
@@ -782,24 +787,41 @@ def test_excluded_keys(collector_class, exclude):
 @pytest.mark.parametrize(
     "collector_class",
     [MultiaSyncDataCollector, MultiSyncDataCollector, SyncDataCollector],
+    # [SyncDataCollector],
 )
 @pytest.mark.parametrize("init_random_frames", [0, 50])
-def test_collector_output_keys(collector_class, init_random_frames):
+@pytest.mark.parametrize("explicit_spec", [True, False])
+def test_collector_output_keys(collector_class, init_random_frames, explicit_spec):
     from torchrl.envs.libs.gym import GymEnv
 
     out_features = 1
     hidden_size = 12
+    total_frames = 200
+    frames_per_batch = 20
+    num_envs = 3
+
     net = LSTMNet(
         out_features,
         {"input_size": hidden_size, "hidden_size": hidden_size},
         {"out_features": hidden_size},
     )
 
-    policy = TensorDictModule(
-        net,
-        ["observation", "hidden1", "hidden2"],
-        ["action", "hidden1", "hidden2", "next_hidden1", "next_hidden2"],
-    )
+    policy_kwargs = {
+        "module": net,
+        "in_keys": ["observation", "hidden1", "hidden2"],
+        "out_keys": ["action", "hidden1", "hidden2", "next_hidden1", "next_hidden2"],
+    }
+    if explicit_spec:
+        hidden_spec = NdUnboundedContinuousTensorSpec(hidden_size)
+        policy_kwargs["spec"] = CompositeSpec(
+            action=UnboundedContinuousTensorSpec(),
+            hidden1=hidden_spec,
+            hidden2=hidden_spec,
+            next_hidden1=hidden_spec,
+            next_hidden2=hidden_spec,
+        )
+
+    policy = TensorDictModule(**policy_kwargs)
 
     env_maker = lambda: GymEnv("Pendulum-v1")
 
@@ -808,14 +830,14 @@ def test_collector_output_keys(collector_class, init_random_frames):
     collector_kwargs = {
         "create_env_fn": env_maker,
         "policy": policy,
-        "total_frames": 200,
-        "frames_per_batch": 20,
+        "total_frames": total_frames,
+        "frames_per_batch": frames_per_batch,
         "init_random_frames": init_random_frames,
     }
 
     if collector_class is not SyncDataCollector:
         collector_kwargs["create_env_fn"] = [
-            collector_kwargs["create_env_fn"] for _ in range(3)
+            collector_kwargs["create_env_fn"] for _ in range(num_envs)
         ]
 
     collector = collector_class(**collector_kwargs)
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -676,15 +676,20 @@ def test_collector_vecnorm_envcreator(static_seed):
 @pytest.mark.parametrize("use_async", [False, True])
 @pytest.mark.skipif(not torch.cuda.is_available(), reason="no cuda device found")
 def test_update_weights(use_async):
+    def create_env():
+        return ContinuousActionVecMockEnv()
+
     n_actions = ContinuousActionVecMockEnv().action_spec.shape[-1]
     policy = TensorDictModule(
         torch.nn.LazyLinear(n_actions), in_keys=["observation"], out_keys=["action"]
     )
+    policy(create_env().reset())
+
     collector_class = (
         MultiSyncDataCollector if not use_async else MultiaSyncDataCollector
     )
     collector = collector_class(
-        [lambda: ContinuousActionVecMockEnv()] * 3,
+        [create_env] * 3,
         policy=policy,
         devices=[torch.device("cuda:0")] * 3,
         passing_devices=[torch.device("cuda:0")] * 3,
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -786,7 +786,11 @@ def test_excluded_keys(collector_class, exclude):
 @pytest.mark.skipif(not _has_gym, reason="test designed with GymEnv")
 @pytest.mark.parametrize(
     "collector_class",
-    [MultiaSyncDataCollector, MultiSyncDataCollector, SyncDataCollector],
+    [
+        SyncDataCollector,
+        MultiaSyncDataCollector,
+        MultiSyncDataCollector,
+    ],
 )
 @pytest.mark.parametrize("init_random_frames", [0, 50])
 @pytest.mark.parametrize("explicit_spec", [True, False])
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: torchrl/collectors/collectors.py [hunk 1]
```diff
@@ -322,10 +322,56 @@ def __init__(
...
+            else:
+                self._tensordict_out = self._tensordict_out[:1]
+            self._tensordict_out = (
+                self._tensordict_out.expand(*env.batch_size, self.frames_per_batch)
+                .to_tensordict()
+                .zero_()
+                .detach()
+            )
+
+        # in addition to outputs of the policy, we add traj_ids and step_count to
```
tcbegley, line 353: Are there other utils the collector needs that I might have missed?

tcbegley (commented):
I've submitted as a draft because I'm expecting CI to fail on `test/test_collectors:test_update_weights`. 

I've been having a bit of trouble pinning down the issue because it originates in a sub-process, but it seems to be related to

```
TypeError: no implementation found for 'torch.nn.linear' on types that implement __torch_function__: [<class 'torchrl.data.tensordict.tensordict.TensorDict'>]
```

The test defines `policy = torch.nn.Linear(3, 4).cuda(1)`,and it seems the problem arises during `env.rollout(3, policy)` used to determine the keys for `_tensordict_out` because torch doesn't know how to handle

```python
tensordict = policy(tensordict)
```

Does the policy need to be wrapped first or something before doing `env.rollout`?

tcbegley:
Weirdly test I was expecting to fail is passing after all... Could be a quirk of my dev environment? Happy to look into that more if you like, but since the tests are passing in CI perhaps this can be reviewed as is.

torchrl/collectors/collectors.py
```diff
@@ -322,10 +322,56 @@ def __init__(
         self._tensordict.set(
             "step_count", torch.zeros(*self.env.batch_size, 1, dtype=torch.int)
         )
-        self._tensordict_out = TensorDict(
-            {},
-            batch_size=[*self.env.batch_size, self.frames_per_batch],
-            device=self.passing_device,
+
+        # TODO: can we rely on policy having spec and ditch hasattr?
```
vmoens, line 326: I think we can assume that the policy has spec, but those spec can be:
None -> do a rollout
CompositeSpec() -> do a rollout
CompositeSpec(action=None) -> do a rollout
CompositeSpec(action=foo, hidden=None) -> do a rollout
CompositeSpec(action=foo) -> use spec
CompositeSpec(action=foo, hidden=bar) -> use spec

torchrl/collectors/collectors.py
```diff
@@ -322,10 +322,56 @@ def __init__(
...
+            if env.batch_size:
+                self._tensordict_out = self._tensordict_out[..., :1]
+            else:
+                self._tensordict_out = self._tensordict_out[:1]
+            self._tensordict_out = (
+                self._tensordict_out.expand(*env.batch_size, self.frames_per_batch)
+                .to_tensordict()
+                .zero_()
+                .detach()
+            )
```
vmoens, line 351: just to make sure, we should reset the env here

torchrl/collectors/collectors.py
vmoens: I don't think so

torchrl/collectors/collectors.py
```diff
@@ -322,10 +322,56 @@ def __init__(
...
+            # match the out_keys we assume the user has given all relevant information
+            self._tensordict_out = env.reset()
+            self._tensordict_out.update(policy.spec.rand())
+        else:
+            # otherwise, we perform a small number of steps with the policy to
+            # determine the relevant keys with which to pre-populate _tensordict_out.
+            # See #505 for additional context.
+            self._tensordict_out = env.rollout(3, policy)
+            if env.batch_size:
+                self._tensordict_out = self._tensordict_out[..., :1]
```
vmoens, line 343: wouldn't this work in both cases?

tcbegley: I thought so too but was getting an `IndexError` when `batch_size` was empty which led me to this.

torchrl/collectors/collectors.py
tcbegley: Good point, done!

test/test_collector.py
```diff
@@ -769,6 +778,67 @@ def make_env():
...
+    net = LSTMNet(
+        out_features,
+        {"input_size": hidden_size, "hidden_size": hidden_size},
+        {"out_features": hidden_size},
+    )
+
+    policy = TensorDictModule(
+        net,
+        ["observation", "hidden1", "hidden2"],
+        ["action", "hidden1", "hidden2", "next_hidden1", "next_hidden2"],
```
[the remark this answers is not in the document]
tcbegley: Thanks, quite an oversight on my part and actually turned up some issues which I think are now resolved!

torchrl/collectors/collectors.py
```diff
@@ -322,10 +322,56 @@ def __init__(
...
+        if (
+            hasattr(policy, "spec")
+            and policy.spec is not None
+            and all(v is not None for v in policy.spec.values())
+            and set(policy.spec.keys()) == set(policy.out_keys)
+        ):
+            # if policy spec is non-empty, all the values are not None and the keys
+            # match the out_keys we assume the user has given all relevant information
+            self._tensordict_out = env.reset()
+            self._tensordict_out.update(policy.spec.rand())
```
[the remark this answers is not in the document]
tcbegley: I've sort of done this but without `unsqueeze`. I think what I have maybe isn't quite right, but ended up being necessary because scalar values specified via `UnboundedContinuousTensorSpec` already have dimension which meant I was getting a clash after unsqueezing.

However, now I have the spec for the hidden layers with shape `(1, hidden_size)` which doesn't seem quite right either.

torchrl/collectors/collectors.py
```diff
@@ -322,10 +322,76 @@ def __init__(
         self._tensordict.set(
             "step_count", torch.zeros(*self.env.batch_size, 1, dtype=torch.int)
         )
-        self._tensordict_out = TensorDict(
-            {},
-            batch_size=[*self.env.batch_size, self.frames_per_batch],
-            device=self.passing_device,
+
+        # TODO: perhaps check type of policy and raise TypeError if something
```
vmoens, line 326: we can remove this comment
Long term plan is to support any kind of policy so i'm happy with the state of the checks

torchrl/collectors/collectors.py
```diff
@@ -322,10 +322,76 @@ def __init__(
...
+                .zero_()
+                .detach()
+            )
+            env.reset()
+
+        # in addition to outputs of the policy, we add traj_ids and step_count to
+        # _tensordict_out which will be collected during rollout
+        if len(self.env.batch_size):
+            traj_ids = (
+                torch.arange(self.env.batch_size[0])
```
vmoens, line 377: do we need to populate traj_ids too? 
It will be set during the first iteration anyway.
Also I would rather see it full of 0s if we really need it

torchrl/collectors/collectors.py
```diff
@@ -322,10 +322,76 @@ def __init__(
...
+                torch.arange(1)
+                .expand(*self._tensordict_out.batch_size)
+                .unsqueeze(-1)
+                .unsqueeze(-1)
+                .clone()
+            )
+
+        self._tensordict_out.set("traj_ids", traj_ids)
+        self._tensordict_out.set(
+            "step_count", torch.zeros(*self._tensordict_out.batch_size, 1)
```
vmoens, line 394: same thing here, do we need to preset this too?

torchrl/collectors/collectors.py
tcbegley: Yeah, if I remove this I get `KeyError: 'key "traj_ids" not found in TensorDict with keys ...` from `split_trajectories` function.

But I've changed to zeros and simplified

torchrl/collectors/collectors.py
tcbegley: Removing this doesn't cause an exception, but it does lead to tests failing because `step_count` is missing from output.

vmoens (approved):
LGTM thanks for this contribuiton!

PR #522, opened 2022-10-05
[Feature] Add make_tensordict() function

### LINKED ISSUE #518: [Feature Request] Automatic tensordict creator
## Motivation

It would be great to have a helper function that automatically creates a `TensorDict` out of a `dict` with a predefined set of rules.
This should not be the default way of creating TensorDict instances but it should be used in several other helper functions through the library to make the usage of TensorDict-based classes easier.

## Intended usage

The function should work like this:
```python
d = {"a": torch.randn(3, 4, 5), "b": torch.randn(3, 4, 2)}
tensordict = make_tensordict(d)
```
this should return a tensordict with the maximum batch size (in this case `[3, 4]`) and the device (if any) of its content (in this case, "cpu").

It should be possible to parameterize some behaviours, e.g. say how many dimensions should the function look at when determining the batch size, or overriding the device.

This funciton should be implemented in the tensordict `utils.py` file.

sicong-huang:
## Description

Implement new function `make_tensordict()` that can automatically create tensordict from keyword arguments. When batch_size is unspecified, it assumes maximum possible batch_size based on the input source dict. 

## Motivation and Context

Closes #518 

## Types of changes

## Checklist

### DIFF
--- a/test/test_tensordict.py
+++ b/test/test_tensordict.py
@@ -2566,55 +2566,62 @@ def test_memory_lock(method):
     td.set("b", torch.randn(4, 5), inplace=True)
 
 
-@pytest.mark.parametrize("device", get_available_devices())
-def test_make_tensordict(device):
-    a = torch.randn(3, 4)
-    tensordict = make_tensordict({"a": a})
-    assert (tensordict["a"] == a.unsqueeze(-1)).all()
+class TestMakeTensorDict:
+    def test_create_tensordict(self):
+        tensordict = make_tensordict(a=torch.zeros(3, 4))
+        assert (tensordict["a"] == torch.zeros(3, 4, 1)).all()
 
-    tensordict = make_tensordict({"a": torch.randn(3, 4)})
-    assert tensordict.batch_size == torch.Size([3, 4])
-    assert tensordict.device is None
+    def test_tensordict_batch_size(self):
+        tensordict = make_tensordict()
+        assert tensordict.batch_size == torch.Size([])
 
-    tensordict = make_tensordict({"a": torch.randn(3, 4), "b": torch.randn(3, 4, 5)})
-    assert tensordict.batch_size == torch.Size([3, 4])
-    assert tensordict.device is None
+        tensordict = make_tensordict(a=torch.randn(3, 4))
+        assert tensordict.batch_size == torch.Size([3, 4])
 
-    tensordict = make_tensordict({"a": torch.randn(3, 4, 2), "b": torch.randn(3, 4, 5)})
-    assert tensordict.batch_size == torch.Size([3, 4])
+        tensordict = make_tensordict(a=torch.randn(3, 4), b=torch.randn(3, 4, 5))
+        assert tensordict.batch_size == torch.Size([3, 4])
 
-    tensordict = make_tensordict({"a": torch.randn(3, 4), "b": torch.randn(1)})
-    assert tensordict.batch_size == torch.Size([])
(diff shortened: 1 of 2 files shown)


test/test_tensordict.py
```diff
@@ -2565,6 +2566,57 @@ def test_memory_lock(method):
     td.set("b", torch.randn(4, 5), inplace=True)
 
 
+@pytest.mark.parametrize("device", get_available_devices())
```
vmoens, line 2569: Perhaps we could do a `TestMakeTensorDict` class and split these in unit tests?

torchrl/data/tensordict/tensordict.py
```diff
@@ -4732,3 +4732,37 @@ def _expand_to_match_shape(parent_batch_size, tensor, self_batch_dims, self_devi
...
+        batch_size (iterable of int, optional): a batch size for the tensordict.
+        device (torch.device or compatible type, optional): a device for the TensorDict.
+    """
+    if batch_size is None:
+        batch_size = _find_max_batch_size(source)
+    return TensorDict(source, batch_size=batch_size, device=device)
+
+
+def _find_max_batch_size(source: Union[TensorDictBase, dict]) -> list[int]:
+    tensor_data = [tensor for _, tensor in source.items()]
```
vmoens, line 4756: what happens if we have an empty list here?

torchrl/data/tensordict/tensordict.py
```diff
@@ -4732,3 +4732,37 @@ def _expand_to_match_shape(parent_batch_size, tensor, self_batch_dims, self_devi
 #
 #
 # _register_pytree_node(TensorDict, _flatten_tensordict, _unflatten_tensordict)
+
+
+def make_tensordict(
+    source: Union[TensorDictBase, dict],
```
vmoens, line 4738: do you think we could make this
`make_tensordict(a=..., b=..., )` instead than a dict?

test/test_tensordict.py
```diff
@@ -2565,6 +2566,57 @@ def test_memory_lock(method):
...
+def test_make_tensordict(device):
+    a = torch.randn(3, 4)
+    tensordict = make_tensordict({"a": a})
+    assert (tensordict["a"] == a.unsqueeze(-1)).all()
+
+    tensordict = make_tensordict({"a": torch.randn(3, 4)})
```
vmoens, line 2575: Can we switch to `zeros` instead of randn?

torchrl/data/tensordict/tensordict.py
```diff
@@ -4732,3 +4732,37 @@ def _expand_to_match_shape(parent_batch_size, tensor, self_batch_dims, self_devi
...
+    batch_size: Optional[Union[Sequence[int], torch.Size, int]] = None,
+    device: Optional[DEVICE_TYPING] = None,
+) -> TensorDict:
+    """Returns a TensorDict created from the input source.
+    If batch_size is not specified, returns the maximum batch size possible
```
vmoens, line 4743: let's add a new line before this one

torchrl/data/tensordict/tensordict.py
vmoens: you can also call `values()` instead of `items()`

vmoens (commented):
Brilliant! Just a couple of things and we'll be good to go

torchrl/data/tensordict/tensordict.py
sicong-huang: good catch

vmoens (approved):
LGTM thanks for this!

Issue #507, opened 2022-10-03
[Feature Request] All models in `modules.models` should have a `device` kwarg


vmoens:
## Motivation

We can now create most `nn.Module` on device directly by passing a `device` kwarg. However this is not the case in TorchRL's modules. Classes such as `MLP` and `ConvNet` should have a `device` kwarg that is passed to all layers.

PR #524, opened 2022-10-06
[Feature] Implemented `device` argument for `modules.models`

yushiyangk:
## Description

Implemented a `device` argument for modules in `torchrl.modules.models` to allow them to be directly constructed on a particular device, rather than first constructing it then moving it with `.to()`. This follows the behaviour of most modules in `torch.nn`. Updated `test_modules.py` to make use of this.

However, an arbitrary `nn.Module` subclass does not necessarily support the `device` argument, in particular for modules that do not contain parameters or buffers. The utility function added in `torchrl/modules/models/utils.py` addresses this. This functionality is tested in `test_mlp` in `test_modules.py`, parameterised with both `nn.ReLU` (which does not support `device`) and `nn.PReLU` (which does).

## Motivation and Context

Addresses #507

## Types of changes

## Checklist

### DIFF
--- a/test/test_exploration.py
+++ b/test/test_exploration.py
@@ -227,25 +227,23 @@ def test_gsde(
 ):
     torch.manual_seed(0)
     if gSDE:
-        model = torch.nn.LazyLinear(action_dim)
+        model = torch.nn.LazyLinear(action_dim, device=device)
         in_keys = ["observation"]
         module = TensorDictSequential(
             TensorDictModule(model, in_keys=in_keys, out_keys=["action"]),
             TensorDictModule(
-                LazygSDEModule(),
+                LazygSDEModule(device=device),
                 in_keys=["action", "observation", "_eps_gSDE"],
                 out_keys=["loc", "scale", "action", "_eps_gSDE"],
             ),
-        ).to(device)
+        )
         distribution_class = IndependentNormal
         distribution_kwargs = {}
     else:
         in_keys = ["observation"]
-        model = torch.nn.LazyLinear(action_dim * 2)
+        model = torch.nn.LazyLinear(action_dim * 2, device=device)
         wrapper = NormalParamWrapper(model)
-        module = TensorDictModule(
-            wrapper, in_keys=in_keys, out_keys=["loc", "scale"]
-        ).to(device)
+        module = TensorDictModule(wrapper, in_keys=in_keys, out_keys=["loc", "scale"])
         distribution_class = TanhNormal
         distribution_kwargs = {"min": -bound, "max": bound}
     spec = NdBoundedTensorSpec(
--- a/torchrl/modules/models/models.py
+++ b/torchrl/modules/models/models.py
@@ -14,6 +14,7 @@ from torchrl._utils import prod
 from torchrl.data import DEVICE_TYPING
 from torchrl.modules.models.utils import (
     _find_depth,
+    create_on_device,
     LazyMapping,
     SquashDims,
     Squeeze2dLayer,
@@ -56,7 +57,7 @@ class MLP(nn.Sequential):
             an integer is provided, every layer will have the same number of cells. If an iterable is provided,
             the linear layers out_features will match the content of num_cells.
             default: 32;
-        activation_class (Type): activation class to be used.
+        activation_class (Type[nn.Module]): activation class to be used.
             default: nn.Tanh
         activation_kwargs (dict, optional): kwargs to be used with the activation class;
         norm_class (Type, optional): normalization class, if any.
@@ -66,11 +67,12 @@ class MLP(nn.Sequential):
         single_bias_last_layer (bool): if True, the last dimension of the bias of the last layer will be a singleton
             dimension.
             default: True;
-        layer_class (Type): class to be used for the linear layers;
+        layer_class (Type[nn.Module]): class to be used for the linear layers;
         layer_kwargs (dict, optional): kwargs for the linear layers;
         activate_last_layer (bool): whether the MLP output should be activated. This is useful when the MLP output
             is used as the input for another module.
             default: False.
+        device (Optional[DEVICE_TYPING]): device to create the module on.
 
     Examples:
         >>> # All of the following examples provide valid, working MLPs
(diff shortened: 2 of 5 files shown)


test/test_exploration.py
```diff
@@ -227,25 +227,23 @@ def test_gsde(
...
         distribution_class = IndependentNormal
         distribution_kwargs = {}
     else:
         in_keys = ["observation"]
         model = torch.nn.LazyLinear(action_dim * 2)
-        wrapper = NormalParamWrapper(model)
-        module = TensorDictModule(
-            wrapper, in_keys=in_keys, out_keys=["loc", "scale"]
-        ).to(device)
+        wrapper = NormalParamWrapper(model).to(device)
```
vmoens, line 245: I guess that here if we need the device to be passed we'd like to do `NormalParamWrapper(model, device=device)`

torchrl/modules/models/models.py
```diff
@@ -488,8 +523,10 @@ def __init__(
         out_features_value: int = 1,
         mlp_kwargs_feature: Optional[dict] = None,
         mlp_kwargs_output: Optional[dict] = None,
+        device: DEVICE_TYPING = "cpu",
     ):
-        super(DuelingMlpDQNet, self).__init__()
+        super().__init__()
+        self.device = torch.device(device)
```
vmoens, line 529: the problem with this is that when calling `net.to(device)` this attribute won't change

torchrl/modules/models/models.py
```diff
@@ -271,18 +292,20 @@ class ConvNet(nn.Sequential):
...
         norm_kwargs (dict, optional): kwargs to be used with the normalization layers;
         bias_last_layer (bool): if True, the last Linear layer will have a bias parameter.
             default: True;
-        aggregator_class (Type): aggregator to use at the end of the chain.
+        aggregator_class (Type[nn.Module]): aggregator to use at the end of the chain.
             default:  SquashDims;
         aggregator_kwargs (dict, optional): kwargs for the aggregator_class;
         squeeze_output (bool): whether the output should be squeezed of its singleton dimensions.
             default: True.
+        device (DEVICE_TYPING): device to create the module on.
```
vmoens, line 307: we should not set `self.device` in the class as this will be inconsistent when calling `module.to(device)`

torchrl/modules/models/models.py
```diff
@@ -564,8 +607,10 @@ def __init__(
         out_features_value: int = 1,
         cnn_kwargs: Optional[dict] = None,
         mlp_kwargs: Optional[dict] = None,
+        device: DEVICE_TYPING = "cpu",
     ):
-        super(DuelingCnnDQNet, self).__init__()
+        super().__init__()
+        self.device = torch.device(device)
```
vmoens, line 613: same here

torchrl/modules/models/models.py
```diff
@@ -623,6 +668,7 @@ def __init__(self, DQNet: nn.Module):
             not isinstance(DQNet.out_features, Number) and len(DQNet.out_features) > 1
         ):
             raise RuntimeError(self._wrong_out_feature_dims_error)
+        self.device = next(DQNet.parameters()).device
```
vmoens, line 671: same here

torchrl/modules/models/models.py
```diff
@@ -717,9 +769,10 @@ def __init__(
         }
         mlp_net_kwargs = mlp_net_kwargs if mlp_net_kwargs is not None else dict()
         mlp_net_default_kwargs.update(mlp_net_kwargs)
-        self.convnet = ConvNet(**conv_net_default_kwargs)
-        self.mlp = MLP(**mlp_net_default_kwargs)
-        ddpg_init_last_layer(self.mlp[-1], 6e-4)
+        self.device = torch.device(device)
```
vmoens, line 772: same here

torchrl/modules/models/models.py
```diff
@@ -760,8 +820,9 @@ def __init__(self, action_dim: int, mlp_net_kwargs: Optional[dict] = None):
         }
         mlp_net_kwargs = mlp_net_kwargs if mlp_net_kwargs is not None else dict()
         mlp_net_default_kwargs.update(mlp_net_kwargs)
-        self.mlp = MLP(**mlp_net_default_kwargs)
-        ddpg_init_last_layer(self.mlp[-1], 6e-3)
+        self.device = torch.device(device)
```
vmoens, line 823: same here

torchrl/modules/models/models.py
```diff
@@ -837,9 +901,10 @@ def __init__(
         }
         mlp_net_kwargs = mlp_net_kwargs if mlp_net_kwargs is not None else dict()
         mlp_net_default_kwargs.update(mlp_net_kwargs)
-        self.convnet = ConvNet(**conv_net_default_kwargs)
-        self.mlp = MLP(**mlp_net_default_kwargs)
-        ddpg_init_last_layer(self.mlp[-1], 6e-4)
+        self.device = torch.device(device)
```
vmoens, line 904: same here

torchrl/modules/models/models.py
```diff
@@ -896,7 +963,8 @@ def __init__(
             mlp_net_kwargs_net1 if mlp_net_kwargs_net1 is not None else dict()
         )
         mlp1_net_default_kwargs.update(mlp_net_kwargs_net1)
-        self.mlp1 = MLP(**mlp1_net_default_kwargs)
+        self.device = torch.device(device)
```
vmoens, line 966: same here

torchrl/modules/models/models.py
```diff
@@ -948,12 +1016,19 @@ class LSTMNet(nn.Module):
...
+        lstm_kwargs: Dict,
+        mlp_kwargs: Dict,
+        device: DEVICE_TYPING = "cpu",
+    ) -> None:
         super().__init__()
         lstm_kwargs.update({"batch_first": True})
-        self.mlp = MLP(**mlp_kwargs)
-        self.lstm = nn.LSTM(**lstm_kwargs)
-        self.linear = nn.LazyLinear(out_features)
+        self.device = torch.device(device)
```
vmoens, line 1028: same here

torchrl/modules/models/utils.py
```diff
@@ -103,3 +104,24 @@ def _find_depth(depth: Optional[int], *list_or_ints: Sequence):
...
+
+    The new instance is created directly on the device if its constructor supports this.
+
+    Args:
+        module_class (Type[nn.Module]): the class of module to be created.
+        device (DEVICE_TYPING): device to create the module on.
+        *args: positional arguments to be passed to the module constructor.
+        **kwargs: keyword arguments to be passed to the module constructor.
+    """
+    try:
```
vmoens, line 123: try / except are a bit expensive
Even if we don't create modules very often in training loops, we do in the tests, and on the long run this risks to have a negative effect on our test speed.
Do you think we can find some other way of doing this?

torchrl/modules/models/utils.py
```diff
@@ -103,3 +105,25 @@ def _find_depth(depth: Optional[int], *list_or_ints: Sequence):
...
+
+    The new instance is created directly on the device if its constructor supports this.
+
+    Args:
+        module_class (Type[nn.Module]): the class of module to be created.
+        device (DEVICE_TYPING): device to create the module on.
+        *args: positional arguments to be passed to the module constructor.
+        **kwargs: keyword arguments to be passed to the module constructor.
+    """
+    fullargspec = inspect.getfullargspec(module_class.__init__)
```
vmoens, line 124: brilliant!

PR #535, opened 2022-10-07
[BugFix] Fix optional imports

vmoens:
## Description

Fix optional broken optional imports

### DIFF
--- a/test/test_helpers.py
+++ b/test/test_helpers.py
@@ -9,8 +9,14 @@ import dataclasses
 import pytest
 import torch
 from _utils_internal import get_available_devices, generate_seeds
-from hydra import initialize, compose
-from hydra.core.config_store import ConfigStore
+
+try:
+    from hydra import initialize, compose
+    from hydra.core.config_store import ConfigStore
+
+    _has_hydra = True
+except ImportError:
+    _has_hydra = False
 from mocking_classes import (
     ContinuousActionConvMockEnvNumpy,
     ContinuousActionVecMockEnv,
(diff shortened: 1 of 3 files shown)


test/test_helpers.py
```diff
@@ -402,6 +412,7 @@ def test_sac_make(device, gsde, tanh_loc, from_pixels, exploration):
         del proof_environment
 
 
+@pytest.mark.skipif(not _has_hydra, reason="No hydra library found")
```
apbard, line 415: nit: group the skipif lines either before or after parametrize

test/test_loggers.py
```diff
@@ -7,9 +7,23 @@
...
+try:
+    import mlflow
+
+    _has_mlfow = True
+except ImportError:
+    _has_mlfow = False
+
 from torchrl.trainers.loggers.csv import CSVLogger
-from torchrl.trainers.loggers.mlflow import MLFlowLogger, _has_mlflow
+from torchrl.trainers.loggers.mlflow import MLFlowLogger
```
apbard, line 26: if removine _has_mlflow here probably worth doing the same for _has_tb and _has_wandb

test/test_loggers.py
```diff
@@ -7,9 +7,23 @@
...
+    import torchvision
+
+    _has_tv = True
+except ImportError:
+    _has_tv = False
+
+try:
+    import mlflow
+
+    _has_mlfow = True
```
apbard, line 21: typo

test/test_loggers.py
apbard: I mean, to be consistent I would change also lines 27 and 28

vmoens: but how do we get `_has_tb` and `_has_wandb` then? Sorry I probably misunderstood

PR #541, opened 2022-10-08
minor test fix to pass ci

GavinPHR:
Minor changes to configs and config test to fit new function signature

`torchrl.data.replay_buffers.storages.Storage`'s init now has argument `max_size` instead of the previous `size`.
Changing `size` to `max_size in the configs and tests too.

Also skipping tests that require halfcheetah rendering on OSX to avoid `mujoco.FatalError: gladLoadGL error`. 
Those tests pass on my mac. MuJoCo (or its dependencies) might not be installed properly in the test environment, but I couldn't fix it.


vmoens (approved):
LGTM!

PR #546, opened 2022-10-10
[Feature] Allow collectors to accept regular modules as policies

### LINKED ISSUE #544: [Feature Request] Allow data collectors to receive policies that read and write regular tensors
## Motivation

As a modular library, we'd like our components to be as orthogonal to each other as possible.
As such, we should aim at minimising the inter-dependency of the components (at least on what is exposed to the users).
An example of such a restriction is the fact that our data collectors expect policies that are `TensorDictModule` instances.
It should be possible to automatically build a TensorDictModule just by reading the signature of a policy.

Example:
```
policy = MyRegularModule()
collector = datacollector(make_env, policy)
assert isinstance(collector.policy, TensorDictModule) 
assert collector.policy.out_keys == ["action"] 
assert collector.policy.module is policy 
```

We would assume that the first output of the policy is an action (and has the "action" key). 

### Checks when passing the policy
- If the policy is not an nn.Module, there's nothing we can do with it.
- The policy signature must match the environment `observation_spec` (without the `"next_"` prefixes), eg the following code snippet would only work if the env has keys "next_observation" and "next_stuff" in its observation_spec:
    ```
    class MyPolicy(nn.Module):
        def forward(self, observation, stuff):
            return foo()
    ```

Open questions: 
- how can we support policies that have multiple outputs? (e.g. a log_probability, a hidden state etc)?
  We could name these "output1" ... "outputN"?

tcbegley:
## Description

This PR makes changes to the collector classes in order that users can pass simple `nn.Module`s as policies.

It checks that the signature of the `forward` method matches the `next_` entries in the environment's `observation_spec`. The first returned value is assumed to be the action while remaining return values are simply enumerated and labelled `output1`, `output2`, etc.

## Motivation and Context

Closes #544

## Types of changes

## Checklist

### DIFF
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -83,12 +83,33 @@ def recursive_map_to_cpu(dictionary: OrderedDict) -> OrderedDict:
     )
 
 
-def _module_is_tensordict_compatible(module: nn.Module):
-    sig = inspect.signature(module.forward)
-    return (
+def _policy_is_tensordict_compatible(policy: nn.Module):
+    sig = inspect.signature(policy.forward)
+
+    if isinstance(policy, TensorDictModule) or (
         len(sig.parameters) == 1
-        and hasattr(module, "in_keys")
-        and hasattr(module, "out_keys")
+        and hasattr(policy, "in_keys")
+        and hasattr(policy, "out_keys")
+    ):
+        # if the policy is a TensorDictModule or takes a single argument and defines
+        # in_keys and out_keys then we assume it can already deal with TensorDict input
+        # to forward and we return True
+        return True
+    elif not hasattr(policy, "in_keys") and not hasattr(policy, "out_keys"):
+        # if it's not a TensorDictModule, and in_keys and out_keys are not defined then
+        # we assume no TensorDict compatibility and will try to wrap it.
+        return False
+
+    # if in_keys or out_keys were defined but policy is not a TensorDictModule or
+    # accepts multiple arguments then it's likely the user is trying to do something
+    # that will have undetermined behaviour, we raise an error
+    raise TypeError(
+        "Received a policy that defines in_keys or out_keys and also expects multiple "
+        "arguments to self.forward. If the policy is compatible with TensorDict, it "
+        "should take a single argument of type TensorDict to self.forward and define "
+        "both in_keys and out_keys. Alternatively, self.forward can accept arbitrarily "
+        "many tensor inputs and leave in_keys and out_keys undefined and TorchRL will "
+        "attempt to automatically wrap the policy with a TensorDictModule."
     )
 
 
@@ -137,8 +158,10 @@ class _DataCollector(IterableDataset, metaclass=abc.ABCMeta):
                     "env must be provided to _get_policy_and_device if policy is None"
                 )
             policy = RandomPolicy(self.env.action_spec)
-        elif isinstance(policy, nn.Module) and not isinstance(policy, TensorDictModule):
-            if not _module_is_tensordict_compatible(policy):
+        elif isinstance(policy, nn.Module):
+            # TODO: revisit these checks when we have determined whether arbitrary
+            # callables should be supported as policies.
+            if not _policy_is_tensordict_compatible(policy):
                 # policy is a nn.Module that doesn't operate on tensordicts directly
                 # so we attempt to auto-wrap policy with TensorDictModule
                 if observation_spec is None:
@@ -167,11 +190,13 @@ class _DataCollector(IterableDataset, metaclass=abc.ABCMeta):
                     )
                 else:
                     raise TypeError(
-                        "If the supplied policy is a nn.Module, it must either accept "
-                        "a single TensorDict as input and define both in_keys and "
-                        "out_keys, or it must accept tensors that are named "
-                        "consistently with the environment's observation_spec so that "
-                        "it can be automatically wrapped with TensorDictModule."
+                        "Arguments to policy.forward are incompatible with entries in "
+                        "env.observation_spec. If you want TorchRL to automatically "
+                        "wrap your policy with a TensorDictModule then the arguments "
+                        "to policy.forward must correspond one-to-one with entries in "
+                        "env.observation_spec that are prefixed with 'next_'. For more "
+                        "complex behaviour and more control you can consider writing "
+                        "your own TensorDictModule."
                     )
 
         try:
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import abc
+import inspect
 import os
 import queue
 import time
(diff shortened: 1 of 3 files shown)


torchrl/collectors/collectors.py
```diff
@@ -124,6 +128,28 @@ def _get_policy_and_device(
                     "env must be provided to _get_policy_and_device if policy is None"
                 )
             policy = RandomPolicy(self.env.action_spec)
+        elif (
+            observation_spec is not None
+            and isinstance(policy, nn.Module)
+            and not isinstance(policy, TensorDictModule)
```
vmoens, line 134: I guess we could write a helper function to determine if a policy is not reading tensordicts
Here are the possible cases:
- the policy is a TensorDictModule
- the policy is a nn.Module that has in_keys, out_keys and reads tensordicts but it is not a TensorDictModule
- the policy is a nn.Module that reads tensors (and maybe also tensordicts)
In the last case we should wrap it into a TensorDictModule

So what we could do is:
- if it's a TDModule we're good
- If it's not but it has in_keys, out_keys and the signature reads only one value we're good
- Otherwise we should wrap that thing in a TensorDictModule

Another option would be to look at the type-hinting in the forward of the module but I'm not sure we really need that

torchrl/collectors/collectors.py
```diff
@@ -124,6 +128,28 @@ def _get_policy_and_device(
...
+        ):
+            # auto-wrap policy with TensorDictModule
+            sig = inspect.signature(policy.forward)
+            next_observation = {
+                key[5:]: value
+                for key, value in observation_spec.rand().items()
```
vmoens, line 140: no need for rand, we should be good with `observation_spec.items()`

torchrl/collectors/collectors.py
```diff
@@ -124,6 +128,28 @@ def _get_policy_and_device(
...
+            if set(sig.parameters) == set(next_observation):
+                out_keys = ["action"]
+                output = policy(**next_observation)
+
+                if isinstance(output, tuple):
+                    out_keys.extend(f"output{i+1}" for i in range(len(output) - 1))
+
+                policy = TensorDictModule(
+                    policy, in_keys=list(sig.parameters), out_keys=out_keys
+                )
```
vmoens, line 152: shouldn't we raise and exception if a policy is not compatible with the env specs?

torchrl/collectors/collectors.py
tcbegley: Yes, I think this makes sense.

When you say helper function, do you mean pulling the `elif` condition out into a call to e.g. `_module_is_tensordict_compatible(module)` or similar? Or did you mean the actual body of the `elif`?

torchrl/collectors/collectors.py
tcbegley: Sure, makes sense, so user must either pass:

- a `TensorDictModule`
- a `nn.Module` that can read `TensorDicts` as in your comment above
- a `nn.Module` that is compatible with the env and can be auto-wrapped

And if none of those conditions are met, we can reasonably raise a `TypeError`? There's no other case we need to support?

torchrl/collectors/collectors.py
tcbegley: The reason for putting `rand` in there was because a couple of lines down I want to do `output = policy(**next_observation)` to count output arguments and if I attempt to do this just on the spec items I get errors like

```
TypeError: linear: argument 'input' (position 1) must be Tensor, not NdBoundedTensorSpec
```

torchrl/collectors/collectors.py
tcbegley: Had a go at this, let me know if it's what you had in mind.

torchrl/collectors/collectors.py
tcbegley: Added in some exceptions + tests that they get raised when appropriate.

torchrl/collectors/collectors.py
```diff
@@ -81,6 +83,15 @@ def recursive_map_to_cpu(dictionary: OrderedDict) -> OrderedDict:
     )
 
 
+def _module_is_tensordict_compatible(module: nn.Module):
+    sig = inspect.signature(module.forward)
```
vmoens, line 87: I guess I was more thinking of this

a function that returns True in these cases
1. 
```
class MyModule(nn.Module):
    in_keys = ["stuff"]
    out_keys = ["other_stuff"]
    def forward(self, tensordict):
        pass
```
2. TensorDictModule

--- 
returns False in those cases
1. 
```
class MyModule(nn.Module):
    in_keys = ["stuff"]
    out_keys = ["other_stuff"]
    def forward(self, tensordict, tensor):  # tensor is unexpected here
        pass
```

```
class MyModule(nn.Module):
    def forward(self, tensor):
        pass
```

--- 
We could also be super restrictive
Return True iif one of:
- TensorDictmodule
- nn.Module + in_keys + out_keys + one input only (assumed to be of type tensordict)

Return False iif
- _no_ in_keys / out_keys attribute and whatever input

if there are in_keys and/or out_keys but there are multiple inputs or similar, it's likely that the use is trying to do smth that will have an undetermined behaviour.

Other cases: throw an error

(in_keys / out_keys is an indirect indicator of "smth that works with TensorDict" which is always worse than a direct indicator but I guess that's the best we can do right now)

torchrl/collectors/collectors.py
```diff
@@ -81,6 +83,36 @@ def recursive_map_to_cpu(dictionary: OrderedDict) -> OrderedDict:
...
+    # accepts multiple arguments then it's likely the user is trying to do something
+    # that will have undetermined behaviour, we raise an error
+    raise TypeError(
+        "Received a policy that defines in_keys or out_keys and also expects multiple "
+        "arguments to self.forward. If the policy is compatible with TensorDict, it "
+        "should take a single argument of type TensorDict to self.forward and define "
+        "both in_keys and out_keys. Alternatively, self.forward can accept arbitrarily "
+        "many tensor inputs and leave in_keys and out_keys undefined and TorchRL will "
+        "attempt to automatically wrap the policy with a TensorDictModule."
+    )
```
tcbegley, line 113: @vmoens is this what you had in mind in your previous comment?

torchrl/collectors/collectors.py
```diff
@@ -124,6 +158,46 @@ def _get_policy_and_device(
                     "env must be provided to _get_policy_and_device if policy is None"
                 )
             policy = RandomPolicy(self.env.action_spec)
+        elif isinstance(policy, nn.Module):
+            # TODO: revisit these checks when we have determined whether arbitrary
+            # callables should be supported as policies.
```
tcbegley, line 163: More thought is going into whether arbitrary callables should be supported as policies. Once there is a clear answer on this question these checks should perhaps be revisited.

vmoens: yes I think that this is cool
if it's not a nn.Module we assume that it works out of the box

vmoens (approved):
LGTM thanks for this!

PR #556, opened 2022-10-11
[Feature] Structured hydra config: Exploration

GavinPHR:
Adding hydra configs for exploration (noisy neural networks + policy wrappers)


vmoens (commented):
I think this PR is not properly named, its scope is much greater than what it says

Can you explain why we need this big refactoring of the tests (other than clarity)?

Also I'm a bit skeptical about having __init__.py files in the tests, this will probably break somewhere down the line. We can keep them for now but let's keep in mind that we may have to remove them.

The [test_configs_common.py](https://github.com/pytorch/rl/pull/556/files#diff-5a10411b9fd044b577562efc306679b4cd7acb68aacda8735e21a9deb06f2103) file has no test. It should be renamed.
Relative imports are very hard to handle on fbcode so I would rather move things to the test utils we already have.

examples/configs/config.yaml
```diff
@@ -5,4 +5,5 @@ defaults:
   - replay_buffer: circular
   - model: dqn/distributional
   - network: dqn/pixels
+  - exploration: noisy_linear
```
vmoens, line 8: I guess we'd like the default to be more neutral, like linear

examples/configs/exploration/additive_gaussian.yaml
```diff
@@ -0,0 +1,10 @@
+network: linear
+wrapper:
+  _partial_: True
+  _target_: torchrl.modules.tensordict_module.exploration.AdditiveGaussianWrapper
+  sigma_init: 1.0
+  sigma_end: 0.1
+  annealing_num_steps: 1000
```
vmoens, line 7: that's a low number of steps
we should default that to the number of frames, you can do `${collector.total_frames}`

examples/configs/exploration/e_greedy.yaml
```diff
@@ -0,0 +1,9 @@
+network: linear
+wrapper:
+  _partial_: True
+  _target_: torchrl.modules.tensordict_module.exploration.EGreedyWrapper
+  eps_init: 1.0
+  eps_end: 0.1
+  annealing_num_steps: 1000
```
vmoens, line 7: same here

examples/configs/exploration/ou_process.yaml
```diff
@@ -0,0 +1,16 @@
+network: linear
+wrapper:
+  _partial_: True
+  _target_: torchrl.modules.tensordict_module.exploration.OrnsteinUhlenbeckProcessWrapper
+  eps_init: 1.0
+  eps_end: 0.1
+  annealing_num_steps: 1000
```
vmoens, line 7: same here

examples/configs/model/ppo/continuous.yaml
```diff
@@ -1,4 +1,4 @@
-policy_wrapper: normal_param
+network_wrapper: normal_param
```
vmoens, line 1: I understand that we have 2 wrappers for the policy (one for the exploration and the other for the distribution), but network_wrapper seems too general to me: we have multiple networks usually, and several can have wrappers.
Why not `policy_net_wrapper` and `exploration_wrapper`?

script.py
```diff
@@ -0,0 +1,13 @@
+s = """        eps_init: float = 1.0,
```
vmoens, line 1: what's this?

test/test_configs.py
```diff
@@ -436,6 +442,69 @@ def test_redq(self, pixels, independent, continuous):
...
+        "ou_process": OrnsteinUhlenbeckProcessWrapper,
+    }
+
+    @pytest.mark.parametrize("network", ["linear", "noisy_linear"])
+    @pytest.mark.parametrize(
+        "wrapper", [None, "e_greedy", "additive_gaussian", "ou_process"]
+    )
+    def test_exploration(self, network, wrapper):
+        additional_config = [
+            "env=cartpole",
```
vmoens, line 466: you probably don't want to use `AdditiveGaussian` and OU with cartpole
CartPole has a discrete action space, and those explorations are suited for continuous ones.
Perhaps we could use half-cheetah with these?

test/test_configs.py
```diff
@@ -436,6 +442,69 @@ def test_redq(self, pixels, independent, continuous):
...
+        cfg = hydra.compose("config", overrides=exploration_config + additional_config)
+        env = instantiate(cfg.env)
+        transforms = [instantiate(transform) for transform in cfg.transforms]
+        for t in transforms:
+            env.append_transform(t)
+        model_params = instantiate(cfg.model)
+        net_partial = instantiate(cfg.network)
+        actor, qvalue, value = make_model_sac(net_partial, model_params, env)
+        actor(env.reset())
+        assert actor.module.module.layer_class is self._LAYER_CLASS_DICT[network]
```
vmoens, line 489: This whole block is orthogonal to the exploration. 
We can think about it -- perhaps it's better like this -- we might miss incompatibilities with mocking classes.
Otherwise I can help you to make one or two mocking classes to have a test that does not rely on other configs.

vmoens (approved):
LGTM, great job!!
Let's just discuss the few points I mentioned here

script.py
GavinPHR: added by mistake, removed

test/test_configs.py
GavinPHR: Recreated environment from printout, rather than relying on config file

test/test_configs.py
```diff
@@ -436,6 +442,80 @@ def test_redq(self, pixels, independent, continuous):
...
+        from torchrl.envs import TransformedEnv
+        from torchrl.envs.libs.gym import GymEnv
+
+        transforms = T.Compose(
+            T.ObservationNorm(0, 1, ["next_observation"]),
+            T.RewardScaling(0, 1, ["reward"]),
+            T.CatTensors(),
+            T.DoubleToFloat(["next_observation_vector"]),
+        )
+        env = TransformedEnv(GymEnv("HalfCheetah-v4"), transforms)
```
vmoens, line 471: I'd rather use a mock env. Is there a reason not to use it?

PR #571, opened 2022-10-14
[BugFix] Fix SyncDataCollector reset

### LINKED ISSUE #570: [BUG] SyncDataCollector resets all workers instead of one
## Describe the bug

When using the SyncDataCollector with a ParallelEnv, **all** workers are reset if any of the workers is done.

## To Reproduce

```python
import torch
from torchrl.envs import ParallelEnv

if __name__ == '__main__':
    num_workers = 2
    num_steps = 4000

    create_env = lambda: GymEnv(env_name='PongNoFrameskip-v4', frame_skip=4)
    create_parallel_env = lambda: ParallelEnv(num_workers, create_env)

    collector = SyncDataCollector(create_parallel_env, frames_per_batch=num_workers, split_trajs=False)
    iterator = collector.iterator()
    episode_rewards = []
    cumulative_reward = torch.zeros(num_workers)
    for _ in range(num_steps):
        batch = next(iterator).squeeze(1)
        cumulative_reward += batch['reward'].squeeze(-1)
        done = batch['done'].squeeze(-1)
        episode_rewards.extend(cumulative_reward[done].tolist())
        cumulative_reward[done] = 0.
    collector.shutdown()
    print(episode_rewards)
```

This leads to the following episode rewards, although the lowest value for Pong should be -21:
```bash
[-21.0, -37.0, -41.0, -21.0]
```
The reason is that some episodes are reset without notice.

## Expected behavior

Only the workers which are actually done should be reset.

## Reason and Possible fixes

[In this line](https://github.com/pytorch/rl/blob/4862459866153a499a3c7e5d6df4238ffdc29fe8/torchrl/collectors/collectors.py#L588) the ParallelEnv is reset but `"reset_workers"` is ignored.  
A quick fix would be to replace the line with:
```python
self._tensordict.update(self.env.reset(self._tensordict), inplace=True)
```
This leads to the following output for the above code and seems to be correct:
```bash
[-21.0, -21.0, -21.0, -21.0, -21.0, -20.0, -20.0, -21.0]
```
But maybe there is a better solution.

## Checklist

jrobine:
## Description

Fix #570 
`close #570` 

## Types of changes

## Checklist

### DIFF
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -585,7 +585,7 @@ class SyncDataCollector(_DataCollector):
             else:
                 self._tensordict.zero_()
 
-            self._tensordict.update(self.env.reset(), inplace=True)
+            self._tensordict.update(self.env.reset(self._tensordict), inplace=True)
             if self._tensordict.get("done").any():
                 raise RuntimeError(
                     f"Got {sum(self._tensordict.get('done'))} done envs after reset."
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -585,7 +585,7 @@ class SyncDataCollector(_DataCollector):
             else:
                 self._tensordict.zero_()
 
-            self._tensordict.update(self.env.reset(self._tensordict), inplace=True)
+            self.env.reset(self._tensordict)
             if self._tensordict.get("done").any():
                 raise RuntimeError(
                     f"Got {sum(self._tensordict.get('done'))} done envs after reset."
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -25,6 +25,7 @@ from torchrl.collectors.collectors import (
     MultiSyncDataCollector,
     RandomPolicy,
 )
+from torchrl.collectors.utils import split_trajectories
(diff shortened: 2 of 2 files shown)


torchrl/collectors/collectors.py
```diff
@@ -585,7 +585,7 @@ def _reset_if_necessary(self) -> None:
             else:
                 self._tensordict.zero_()
 
-            self._tensordict.update(self.env.reset(), inplace=True)
+            self._tensordict.update(self.env.reset(self._tensordict), inplace=True)
```
vmoens, line 588: I think we don't need the call to update.
env.reset() will update the tensordict.

vmoens (commented):
Thanks! Let's try without the update?

vmoens:
I implemented some tests, let me know if they make sense

jrobine:
Thanks! It seems like the step counts are also incorrect, so I think only assertions 313 and 322 are really necessary.

vmoens:
Do you mean "were incorrect" or are they still incorrect after this PR?

jrobine:
Sorry, they _were_ incorrect

PR #588, opened 2022-10-19
[CI, Tests] Faster tests

vmoens:
## Description

Small tweaks to run faster tests


test/test_collector.py
```diff
@@ -176,7 +176,7 @@ def env_fn(seed):
     else:
 
         def env_fn(seed):
-            env = ParallelEnv(
+            env = SerialEnv(
```
shagunsodhani, line 179: How does going to SerialEnv makes things faster ?

vmoens: ParallelEnv requires multiprocess to be started on `"spawn"` mode, rather than `"fork"`
Hence each new process has to run the whole chain of imports, which takes a considerable amount of time.
In practice, in your app you will only create your ParallelEnv once for all, but with tests we re-instantiate these things continuously. Because of this, some tests take > 30 seconds (because they start multiple processes for the collector which all start multiple processes for the envs).

Switching to the `"fork"` does not work (cuda tensors cannot live in `"fork"` settings and `"fork"` is much more error prone, see [this](https://pythonspeed.com/articles/python-multiprocessing/)). 

SerialEnv starts much quicker, but its execution is obviously slower. Since the launch is the compute bottleneck, it makes sense to use Serial instead of Parallel.

shagunsodhani: Ohh wow! Thanks for the explanation!

vmoens: yeah I tried to do a `SerialEnv` that behaved as much as i could like the parallel but it's not easy to get about.

Issue #538, opened 2022-10-07
[Feature Request] Create a categorical tensorspec and make DQN primitives compatible with non-one-hot encoding


vmoens:
## Motivation

We opted for default one-hot encoding for categorical data because it facilitates the retrieval of action values:

```python
chosen_action_value = action * action_values
```

However, this is not scalable to large action spaces.
We should make it possible to use integers instead.

To make this possible, the following changes are needed:
- [ ] Create a `DiscreteTensorSpec` [similar to `MultOneHotDiscreteTensorSpec`](https://github.com/pytorch/rl/blob/417817abc12240cb8c93d767445fdec5f6c7050c/torchrl/data/tensor_specs.py#L909) + tests
- [ ] Change QValueHook to make it possible to use [a "categorical" space (or similar) instead of "one_hot"](https://github.com/pytorch/rl/blob/497ad8f6304403f358f093f289d4c94ad2ce5878/torchrl/modules/tensordict_module/actors.py#L259)
- [ ] Do the same with [`DistributionalQValueHook`](https://github.com/pytorch/rl/blob/497ad8f6304403f358f093f289d4c94ad2ce5878/torchrl/modules/tensordict_module/actors.py#L357)
- [ ] [Adapt the DQNLoss](https://github.com/pytorch/rl/blob/497ad8f6304403f358f093f289d4c94ad2ce5878/torchrl/objectives/costs/dqn.py#L91-L104) to these changes
- [ ] [Adapt the DistributionalDQNLoss](https://github.com/pytorch/rl/blob/497ad8f6304403f358f093f289d4c94ad2ce5878/torchrl/objectives/costs/dqn.py#L129) to these changes
- [ ] Change the gym specs reader [here](https://github.com/pytorch/rl/blob/497ad8f6304403f358f093f289d4c94ad2ce5878/torchrl/envs/libs/gym.py#L58) to make it possible to choose one spec type or the other. One way to go about this would be to have a global variable that controls which categorical encoding must be used
  ```python
  ONE_HOT_ENCODING = os.environ.get("ONE_HOT_ENCODING", True)
  ```
  Then we can use this in all the code-base to choose one or the other encoding.

All these features should be unit-tested and parametrisable (e.g. with `ONE_HOT_ENCODING`) in the DQN training scripts.

PR #593, opened 2022-10-20
[Feature] Categorical encoding for action space

artkorenev:
## Description

Added alternative to one-hot encoding for action spaces with categorical features.

## Motivation and Context

Implementing feature #538.

## Types of changes

## Checklist
* [x]  I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
* [ ]  My change requires a change to the documentation.
* [x]  I have updated the tests accordingly (_required for a bug fix or a new feature_).
* [ ]  I have updated the documentation accordingly.

Checklist from #538:
* [x]  Create a `DiscreteTensorSpec` [similar to `MultOneHotDiscreteTensorSpec`](https://github.com/pytorch/rl/blob/417817abc12240cb8c93d767445fdec5f6c7050c/torchrl/data/tensor_specs.py#L909) + tests
* [x]  Change QValueHook to make it possible to use [a "categorical" space (or similar) instead of "one_hot"](https://github.com/pytorch/rl/blob/497ad8f6304403f358f093f289d4c94ad2ce5878/torchrl/modules/tensordict_module/actors.py#L259)
* [x]  Do the same with [`DistributionalQValueHook`](https://github.com/pytorch/rl/blob/497ad8f6304403f358f093f289d4c94ad2ce5878/torchrl/modules/tensordict_module/actors.py#L357)
* [x]  [Adapt the DQNLoss](https://github.com/pytorch/rl/blob/497ad8f6304403f358f093f289d4c94ad2ce5878/torchrl/objectives/costs/dqn.py#L91-L104) to these changes
* [x]  [Adapt the DistributionalDQNLoss](https://github.com/pytorch/rl/blob/497ad8f6304403f358f093f289d4c94ad2ce5878/torchrl/objectives/costs/dqn.py#L129) to these changes
* [x]  Change the gym specs reader [here](https://github.com/pytorch/rl/blob/497ad8f6304403f358f093f289d4c94ad2ce5878/torchrl/envs/libs/gym.py#L58) to make it possible to choose one spec type or the other. One way to go about this would be to have a global variable that controls which categorical encoding must be used.

Current implementation seems to couple different parts of TorchRL quite tightly as there is quite a few places where it is assumed that the action space is one-hot. Perhaps, a more generic action-space object that encapsulates manipulation with values and that is shared across losses, actors, hooks, etc. (as a `nn.Module` maybe) might be a worth investment in future.

The mode is turned on by `export CATEGORICAL_ACTION_ENCODING=True` command, and it is turned off by default. Another alternative could be to provide it through the config-file, however this, again, would require a bit more generic solution to configure module that will be passed to all the necessary modules.

### DIFF
--- a/test/test_actors.py
+++ b/test/test_actors.py
@@ -12,12 +12,10 @@ class TestQValue:
             QValueHook(action_space="wrong_value")
         assert "action_space must be one of" in str(exc.value)
 
-
     def test_distributional_qvalue_hook_wrong_action_space(self):
         with pytest.raises(ValueError) as exc:
             DistributionalQValueHook(action_space="wrong_value", support=None)
         assert "action_space must be one of" in str(exc.value)
-
 
     @pytest.mark.parametrize(
         "action_space, expected_action",
@@ -37,7 +35,6 @@ class TestQValue:
         assert (torch.tensor(expected_action, dtype=torch.long) == action).all()
         assert (values == in_values).all()
         assert (torch.tensor([100.0]) == chosen_action_value).all()
-
 
     @pytest.mark.parametrize(
         "action_space, expected_action",
@@ -63,7 +60,6 @@ class TestQValue:
         assert (values == in_values).all()
         assert (torch.tensor([[100.0], [5.0]]) == chosen_action_value).all()
 
-
     @pytest.mark.parametrize(
         "action_space, expected_action",
         (
@@ -71,7 +67,9 @@ class TestQValue:
             ("categorical", 2),
         ),
     )
-    def test_distributional_qvalue_hook_0_dim_batch(self, action_space, expected_action):
+    def test_distributional_qvalue_hook_0_dim_batch(
+        self, action_space, expected_action
+    ):
         support = torch.tensor([-2.0, 0.0, 2.0])
         hook = DistributionalQValueHook(action_space=action_space, support=support)
 
@@ -91,7 +89,6 @@ class TestQValue:
         assert (action == expected_action).all()
         assert values.shape == in_values.shape
         assert (values == in_values).all()
-
 
     @pytest.mark.parametrize(
         "action_space, expected_action",
--- a/test/test_actors.py
+++ b/test/test_actors.py
@@ -1,7 +1,10 @@
 import pytest
 import torch
+from torchrl.modules.tensordict_module.actors import (
+    QValueHook,
+    DistributionalQValueHook,
+)
 
-from torchrl.modules.tensordict_module.actors import QValueHook, DistributionalQValueHook
 
 def test_qvalue_hook_wrong_action_space():
     with pytest.raises(ValueError):
@@ -12,19 +15,22 @@ def test_distributional_qvalue_hook_wrong_action_space():
     with pytest.raises(ValueError):
         DistributionalQValueHook(action_space="wrong_value", support=None)
 
+
 @pytest.mark.parametrize(
     "action_space, expected_action",
     (
         ("one_hot", [0, 0, 1, 0, 0]),
         ("categorical", 2),
-    )
+    ),
 )
 def test_qvalue_hook_0_dim_batch(action_space, expected_action):
     hook = QValueHook(action_space=action_space)
-    
+
     in_values = torch.tensor([1.0, -1.0, 100.0, -2.0, -3.0])
-    action, values, chosen_action_value = hook(net=None, observation=None, values=in_values)
-    
+    action, values, chosen_action_value = hook(
+        net=None, observation=None, values=in_values
+    )
+
     assert (torch.tensor(expected_action, dtype=torch.long) == action).all()
     assert (values == in_values).all()
     assert (torch.tensor([100.0]) == chosen_action_value).all()
@@ -35,17 +41,21 @@ def test_qvalue_hook_0_dim_batch(action_space, expected_action):
     (
         ("one_hot", [[0, 0, 1, 0, 0], [1, 0, 0, 0, 0]]),
         ("categorical", [2, 0]),
-    )
+    ),
 )
 def test_qvalue_hook_1_dim_batch(action_space, expected_action):
     hook = QValueHook(action_space=action_space)
-    
-    in_values = torch.tensor([
-        [1.0, -1.0, 100.0, -2.0, -3.0],
-        [5.0, 4.0, 3.0, 2.0, -5.0],
-    ])
-    action, values, chosen_action_value = hook(net=None, observation=None, values=in_values)
-    
+
+    in_values = torch.tensor(
+        [
+            [1.0, -1.0, 100.0, -2.0, -3.0],
+            [5.0, 4.0, 3.0, 2.0, -5.0],
+        ]
+    )
+    action, values, chosen_action_value = hook(
+        net=None, observation=None, values=in_values
+    )
+
     assert (torch.tensor(expected_action, dtype=torch.long) == action).all()
     assert (values == in_values).all()
     assert (torch.tensor([[100.0], [5.0]]) == chosen_action_value).all()
@@ -56,20 +66,24 @@ def test_qvalue_hook_1_dim_batch(action_space, expected_action):
     (
         ("one_hot", [0, 0, 1, 0, 0]),
         ("categorical", 2),
-    )
+    ),
 )
-def test_distributional_qvalue_hook_0_dim_batch(action_space, expected_action):    
+def test_distributional_qvalue_hook_0_dim_batch(action_space, expected_action):
     support = torch.tensor([-2.0, 0.0, 2.0])
     hook = DistributionalQValueHook(action_space=action_space, support=support)
-    
-    in_values = torch.nn.LogSoftmax(dim=-1)(torch.tensor([
-            [1.0, -1.0, 11.0, -2.0, 30.0],
-            [1.0, -1.0, 1.0, -2.0, -3.0],
-            [1.0, -1.0, 10.0, -2.0, -3.0],
-    ]))
+
+    in_values = torch.nn.LogSoftmax(dim=-1)(
+        torch.tensor(
+            [
+                [1.0, -1.0, 11.0, -2.0, 30.0],
+                [1.0, -1.0, 1.0, -2.0, -3.0],
+                [1.0, -1.0, 10.0, -2.0, -3.0],
+            ]
+        )
+    )
     action, values = hook(net=None, observation=None, values=in_values)
     expected_action = torch.tensor(expected_action, dtype=torch.long)
-    
+
     assert action.shape == expected_action.shape
     assert (action == expected_action).all()
     assert values.shape == in_values.shape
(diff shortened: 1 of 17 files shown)


torchrl/_utils.py
```diff
@@ -117,3 +118,24 @@ def prod(sequence):
         return math.prod(sequence)
     else:
         return int(np.prod(sequence))
+
+
+def get_binary_env_var(key):
```
vmoens, line 123: codecov complains about this being half tested, can we fix that?

torchrl/data/tensor_specs.py
```diff
@@ -37,27 +36,20 @@
...
     "MultOneHotDiscreteTensorSpec",
+    "DiscreteTensorSpec",
     "CompositeSpec",
 ]
 
+from torchrl._utils import get_binary_env_var
 from torchrl.data.tensordict.tensordict import TensorDictBase, TensorDict
 
-_CHECK_IMAGES = os.environ.get("CHECK_IMAGES", False)
+_CHECK_IMAGES = get_binary_env_var("CHECK_IMAGES")
```
vmoens, line 46: thanks for the fix

torchrl/data/tensor_specs.py
```diff
@@ -997,6 +992,72 @@ def _project(self, val: torch.Tensor) -> torch.Tensor:
         return torch.cat([super()._project(_val) for _val in vals], -1)
 
 
+class DiscreteTensorSpec(TensorSpec):
+    """A discrete tensor spec.
+
+    An alternative to OneHotTensorSpec for categorical variables in TorchRL. Instead of
+    using multiplication, categorical variables perform indexing which can speed up
+    computation and reduce memory cost for large categorical variables.
+        >>> batch, size = 3, 4
```
vmoens, line 1001: Can we put that under an `Example:`?

torchrl/data/tensor_specs.py
```diff
@@ -997,6 +992,72 @@ def _project(self, val: torch.Tensor) -> torch.Tensor:
...
+    shape: torch.Size
+    space: DiscreteBox
+    device: torch.device = torch.device("cpu")
+    dtype: torch.dtype = torch.float
+    domain: str = ""
+
+    def __init__(
+        self,
+        n: int,
+        shape: Optional[torch.Size] = torch.Size((1,)),
```
vmoens, line 1026: IIRC we once had a linting issue with this (that seems to have disappeared now).
This is why we usually have a `None` and if `None` then shape is replaced by `Size([1])`.

torchrl/data/tensor_specs.py
```diff
@@ -997,6 +992,72 @@ def _project(self, val: torch.Tensor) -> torch.Tensor:
...
+
+    def rand(self, shape=None) -> torch.Tensor:
+        if shape is None:
+            shape = torch.Size([])
+        return (torch.rand(*shape, *self.shape, device=self.device) * self.space.n).to(
+            torch.long
+        )
+
+    def _project(self, val: torch.Tensor) -> torch.Tensor:
+        return torch.round(val).clamp_(min=0, max=self.space.n - 1)
```
vmoens, line 1042: We should test the project method for this class

torchrl/modules/tensordict_module/actors.py
```diff
@@ -294,6 +306,21 @@ def _mult_one_hot(self, value: torch.Tensor, support: torch.Tensor) -> torch.Ten
...
+        values: torch.Tensor, action: torch.Tensor
+    ) -> torch.Tensor:
+        return (action * values).sum(-1, True)
+
+    @staticmethod
+    def _categorical_action_value(
+        values: torch.Tensor, action: torch.Tensor
+    ) -> torch.Tensor:
+        if len(values.shape) == 1:
+            return values[action].unsqueeze(-1)
```
vmoens, line 320: this seems to be untested

torchrl/modules/tensordict_module/actors.py
```diff
@@ -352,18 +379,23 @@ def __init__(
...
         self.var_nums = var_nums
-        self.fun_dict = {
+        self.action_func_mapping = {
             "one_hot": self._one_hot,
             "mult_one_hot": self._mult_one_hot,
             "binary": self._binary,
+            "categorical": self._categorical,
         }
+        if action_space not in self.action_func_mapping:
+            raise ValueError(
```
vmoens, line 389: this is not covered by the tests

torchrl/modules/tensordict_module/actors.py
```diff
@@ -401,6 +433,10 @@ def _mult_one_hot(self, value: torch.Tensor, support: torch.Tensor) -> torch.Ten
             -1,
         )
 
+    def _categorical(self, value: torch.Tensor, support: torch.Tensor) -> torch.Tensor:
+        value = value = self._support_expected(value, support)
+        return torch.argmax(value, dim=-1).to(torch.long)
```
vmoens, line 438: this seems to be untested

torchrl/objectives/dqn.py
```diff
@@ -95,8 +96,15 @@ def forward(self, input_tensordict: TensorDictBase) -> TensorDict:
             buffers=self.value_network_buffers,
         )
 
+        action = tensordict.get("action")
         pred_val = td_copy.get("action_value")
-        pred_val_index = (pred_val * action).sum(-1)
+
+        if _CATEGORICAL_ACTION_ENCODING:
+            batch_size = action.size(0)
+            pred_val_index = pred_val[range(batch_size), action.squeeze(-1)]
```
vmoens, line 104: I guess we'd like to be able to work with batch sizes that are more than unidimensional.

We can use `gather` (I think)
```python
x = torch.randn(3, 4, 5)
idx = torch.randint(5, (3, 4))
x.gather(-1, idx)
```

This will index `x` along the last dimension.

vmoens: this is not covered by the tests (it should be done in `test_costs.py` I believe).

torchrl/objectives/dqn.py
```diff
@@ -95,8 +96,15 @@ def forward(self, input_tensordict: TensorDictBase) -> TensorDict:
             buffers=self.value_network_buffers,
         )
 
+        action = tensordict.get("action")
         pred_val = td_copy.get("action_value")
-        pred_val_index = (pred_val * action).sum(-1)
+
+        if _CATEGORICAL_ACTION_ENCODING:
```
vmoens, line 102: This is a bit tricky

What if someone explicitly builds a policy that uses categorical actions, but does not set the env variable?

My view of the env variable is more along the line of "a way of easily change a whole training script from categorical to one hot and vice-versa".

But the reversed dependency should not hold, and the environment variable should only be checked when building the network.

One option could be to infer the type of action by looking at the `value_network`: if it has registered the action spec, then we can use it to infer the type of action we will see. Otherwise, we must ask to the user to tell us (via an arg in the constructor) what kind of action we should be expecting.

Open to other suggestions obviously :)

torchrl/objectives/dqn.py
```diff
@@ -198,9 +206,13 @@ def forward(self, input_tensordict: TensorDictBase) -> TensorDict:
             buffers=self.value_network_buffers,
         )  # Log probabilities log p(s_t, ·; θonline)
         action_log_softmax = td_clone.get("action_value")
-        action_expand = action.unsqueeze(-2).expand_as(action_log_softmax)
-        log_ps_a = action_log_softmax.masked_select(action_expand.to(torch.bool))
-        log_ps_a = log_ps_a.view(batch_size, atoms)  # log p(s_t, a_t; θonline)
+
+        if _CATEGORICAL_ACTION_ENCODING:
+            log_ps_a = action_log_softmax[range(batch_size), :, action.squeeze(-1)]
```
vmoens, line 211: same comments as above regarding indexing, usage of env variable and test coverage

torchrl/trainers/helpers/models.py
```diff
@@ -178,6 +181,11 @@ def make_dqn_actor(
     out_features = env_specs["action_spec"].shape[0]
     actor_class = QValueActor
     actor_kwargs = {}
+
+    if _CATEGORICAL_ACTION_ENCODING:
+        actor_kwargs.update({"action_space": "categorical"})
+        out_features = env_specs["action_spec"].space.n
```
vmoens, line 187: perhaps we could have an arg in the function that indicates if categorical has to be used.
If default (e.g. `None`) then it falls back on `_CATEGORICAL_ACTION_ENCODING`.

vmoens (commented):
We'd like to add the new classes to the docs (this is not done automatically atm).
Just add them in docs/source/data.

torchrl/objectives/dqn.py
artkorenev: > What if someone explicitly builds a policy that uses categorical actions, but does not set the env variable?

This is exactly what I meant when mentioned that the current solution couples things a bit too much.

Since actions are passed as pure tensors (even though, wrapped in TensorDict) it is either assumed or configured separately for each class what to do with `action` tensors.

Now I can see two options what we can do here:

1. Whenever we deal with actions, we either pass `action_space` in the constructor or derive it from some other entity (e.g. `value_network` as you proposed). However here we rely on careful configuration of all modules together (in theory, somebody could explicitly configure `value_network` incompatible with an environment). Another drawback is that we "spreading out" the logic for each case across the code base which can make code a harder to understand (I think in case of one-hot/categorical it is fine, but for future approaches it can blow up).
2. We can introduce something like `ActionSpec` which would be an extended version of existing specs, though purely for action processing (deriving actions from `value_network`, etc.). We instantiate the `ActionSpec` once and pass its instance everywhere we need encapsulating work with actions. Although, it is hard for me to estimate what interface `ActionSpec` needs in order to cover all possible cases (I mean globally for all RL cases). And this also would require larger scale refactoring.

tldr; I think 1st option seems better choice for this situation and somewhat follows current code base practices (e.g. `QValueHook`)

torchrl/_utils.py
```diff
@@ -117,3 +118,24 @@ def prod(sequence):
...
+    """Parses and returns the binary enironment variable value.
+
+    If not present in environment, it is considered `False`.
+
+    Args:
+        key (str): name of the environment variable.
+    """
+    val = os.environ.get(key, False)
+    if val in ("0", "False", False):
```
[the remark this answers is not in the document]
artkorenev: I changed it to `"False"` by default. Environment variables cannot be set to something other than `str`.

torchrl/_utils.py
artkorenev: Added tests in order to cover every case.

torchrl/data/tensor_specs.py
artkorenev: Fixed, thx

torchrl/data/tensor_specs.py
artkorenev: Noted, changed to that

torchrl/data/tensor_specs.py
artkorenev: Added tests for `project`, thx

torchrl/modules/tensordict_module/actors.py
artkorenev: Added tests for Q value hooks now

torchrl/modules/tensordict_module/actors.py
artkorenev: Fixed with tests

torchrl/modules/tensordict_module/actors.py
artkorenev: This should be covered now as well

torchrl/objectives/dqn.py
artkorenev: Reworked indexing here, thx

torchrl/objectives/dqn.py
artkorenev: Reworked this as well with a few tweaks since we work with atoms here.

torchrl/trainers/helpers/models.py
artkorenev: I changed it here to be inferred from action spec that is set during the env setup.

torchrl/objectives/dqn.py
artkorenev: Also added tests so should be covered by coverage by now

vmoens (approved):
Great! Amazing work, if fits in the library very well now. I think we're there. 

A couple of things before we merge this
- Can you have a look at the few comments I left?
- Can you have a look at the coverage report (some lines are not covered -- don't worry about things that are not covered but are in the `test` directory).

test/test_actors.py
```diff
@@ -0,0 +1,126 @@
+import pytest
+import torch
+from torchrl.modules.tensordict_module.actors import (
+    QValueHook,
+    DistributionalQValueHook,
+)
+
+
+def test_qvalue_hook_wrong_action_space():
+    with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
artkorenev: Added a message check (had to make it short since it relies on order of the items in dict)

test/test_actors.py
```diff
@@ -0,0 +1,126 @@
...
+        QValueHook(action_space="wrong_value")
+
+
+def test_distributional_qvalue_hook_wrong_action_space():
+    with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
artkorenev: Fixed as in `test_qvalue_hook_wrong_action_space`

test/test_actors.py
```diff
@@ -0,0 +1,126 @@
+import pytest
+import torch
+from torchrl.modules.tensordict_module.actors import (
+    QValueHook,
+    DistributionalQValueHook,
+)
+
+
+def test_qvalue_hook_wrong_action_space():
```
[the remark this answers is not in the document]
artkorenev: > could we put all those `test_qvalue` under a `TestQValue` class?
Fixed, thanks

artkorenev:
Awesome! Thank you for the review!
It seems like the coverage is now fixed completely.

PR #599, opened 2022-10-23
[Feature] Qmix - Gru Net

PaLeroy:
## Description

Creation of the GRUNet to control agents in QMIX

## Types of changes

## Checklist

### DIFF
--- a/torchrl/modules/models/models.py
+++ b/torchrl/modules/models/models.py
@@ -2,7 +2,7 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-
+import warnings
 from numbers import Number
 from typing import Dict, List, Optional, Sequence, Tuple, Type, Union
 
@@ -1206,6 +1206,11 @@ class GRUNet(nn.Module):
             )
         if "bidirectional" in gru_kwargs and gru_kwargs["bidirectional"]:
             raise NotImplementedError("bidirectional GRU is not yet implemented.")
+
+        if not gru_kwargs.get("batch_first", True):
+            warnings.warn(
+                "You set batch_first to False, but GRUNet enforces batch_first to be True, gru_kwargs will be updated."
+            )
 
         mlp_input_kwargs.update({"device": device})
         gru_kwargs.update({"device": device, "batch_first": True})
--- a/torchrl/modules/models/models.py
+++ b/torchrl/modules/models/models.py
@@ -1038,9 +1038,7 @@ class LSTMNet(nn.Module):
 
 
 class GRUNet(nn.Module):
-    """
-    The GRUNet is a neural network composed of a GRU layer encapsulated
-    between two MLPs.
+    """The GRUNet is a neural network composed of a GRU layer encapsulated between two MLPs.
 
     Args:
         mlp_input_kwargs (dict):  kwargs for the MLP before the GRU
@@ -1084,6 +1082,21 @@ class GRUNet(nn.Module):
         last_h: Tensor of shape
             D*num_layers , H_out
             D*num_layers , N, H_out
+
+    Examples:
+        >>> in_features = 11
+        >>> hidden_size = 13
+        >>> out_features = 3
+        >>> batch_size = 5
+        >>> seq_len = 7
+        >>> mlp_input_kwargs = {"in_features": in_features, "out_features": hidden_size}
+        >>> gru_kwargs = { "input_size": hidden_size, "hidden_size": hidden_size}
+        >>> mlp_output_kwargs = {"in_features": hidden_size, "out_features": out_features}
+        >>> net = GRUNet(mlp_input_kwargs, gru_kwargs, mlp_output_kwargs)
+        >>> x_batch = torch.randn(batch_size, seq_len, 11)
+        >>> out_batch , h_batch = net(x_batch)
+        >>> x_no_batch = torch.randn(seq_len, 11)
+        >>> out_no_batch, h_no_batch = net(x_no_batch)
 
     """
 
--- a/torchrl/modules/models/models.py
+++ b/torchrl/modules/models/models.py
@@ -1034,3 +1034,23 @@ class LSTMNet(nn.Module):
 
         input = self.mlp(input)
         return self._lstm(input, hidden0_in, hidden1_in)
+
+
+class GRUNet(nn.Module):
+    """
+    A GruCell encapsulated between two MLP
+    """
+
+    def __init__(
+        self, mlp_input_kwargs: Dict, gru_kwargs: Dict, mlp_output_kwargs: Dict
+    ) -> None:
+        self.mlp_int = MLP(**mlp_input_kwargs)
+        self.gru = nn.GRUCell(**gru_kwargs)
+        self.mlp_out = MLP(**mlp_output_kwargs)
+        super().__init__()
+
+    def forward(self, inputs, hidden_state):
+        x = self.mlp_int(inputs)
+        h = self.gru(x, hidden_state)
+        q = self.mlp_out(h)
+        return q, h
--- a/torchrl/modules/models/models.py
+++ b/torchrl/modules/models/models.py
@@ -1083,7 +1083,7 @@ class GRUNet(nn.Module):
                 (0): Linear(in_features=13, out_features=3, bias=True)
               )
             )
-        >>> x_no_batch = torch.randn(5, 11)
+        >>> x_no_batch = torch.randn(7, 11)
         >>> out_no_batch, h_no_batch = net(x_no_batch)
         >>> print(out_no_batch.shape, h_no_batch.shape)
         torch.Size([7, 3]) torch.Size([1, 13])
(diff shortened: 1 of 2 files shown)


torchrl/modules/models/models.py
```diff
@@ -1034,3 +1035,27 @@ def forward(
...
+
+
+class GRUNet(nn.Module):
+    """
+    MLP -> GRU -> MLP
+
+    """
+
+    def __init__(
+        self, mlp_input_kwargs: Dict, gru_kwargs: Dict, mlp_output_kwargs: Dict
```
vmoens, line 1047: it would be nice to have a `device` argument here, almost all torch modules now support instantiation on device

torchrl/modules/models/models.py
```diff
@@ -1034,3 +1035,27 @@ def forward(
...
+        assert mlp_input_kwargs["out_features"] == gru_kwargs["input_size"]
+        assert mlp_output_kwargs["in_features"] == gru_kwargs["hidden_size"]
+
+        self.mlp_in = MLP(**mlp_input_kwargs)
+        self.gru = nn.GRUCell(**gru_kwargs)
+        self.mlp_out = MLP(**mlp_output_kwargs)
+
+    def forward(self, inputs, hidden_state):
+        mlp_in = self.mlp_in(inputs)
+        hidden_state = self.gru(mlp_in, hidden_state)
```
vmoens, line 1059: I expect that if you give a sequence as input this module will not behave as expected right?
Should we enforce that the input shapes match some requirement? Like no 3d input?
Problem with that 3d inputs are only an indirect indicator that someone is passing tensors that are time-stamped, it might as well be a 2d batch size. Plus, a 1d batch size might be time-stamped too :p 

Perhaps a good docstring would go a long way. Can we put in the docstring of this class what kind of input we expect?

PaLeroy: I never worked with GRU that takes full sequences... It appears that I missed a nice feature.
Indeed, the current implementation will not work well when taking a sequence as input.

I'll rework this as it is done in the GRU, and using GRU instead of GRUCell then (enforcing batch first)

> **input**: tensor of shape :math:`(L, H_{in})` for unbatched input,
          :math:`(N, L, H_{in})` when ``batch_first=True`` containing the features of
          the input sequence.

This should lead to something closer to the LSTMNet already implemented.

PaLeroy:
I changed to GRU instead of GRUCell.
It supports batched or unbatched inputs but the sequence length must be specified -> only 2d and 3d inputs.
Maybe the docstring is a bit overloaded?

torchrl/modules/models/models.py
```diff
@@ -1034,3 +1035,90 @@ def forward(
 
         input = self.mlp(input)
         return self._lstm(input, hidden0_in, hidden1_in)
+
+
+class GRUNet(nn.Module):
+    """
+    The GRUNet is a neural network composed of a GRU layer encapsulated
```
vmoens, line 1042: This should look like this for a nice rendering in the doc

```python
"""The GRUNet is a neural network composed of a GRU layer encapsulated between two MLPs.

...
"""
```

torchrl/modules/models/models.py
```diff
@@ -1034,3 +1035,90 @@ def forward(
...
+        The size of input features in mlp_output must be the same as the last hidden size of the GRU.
+
+    N = batch size
+    L = sequence size
+    mlp_input_in = input size of the MLP_input
+    mlp_input_out = output size of the MLP_input
+    H_in = input size of the GRU
+    H_out = output size of the GRU
+    mlp_output_in = input size of the MLP_output
+    mlp_output_out = output size of the MLP_output
```
vmoens, line 1065: What are those fields? I don't see them anywhere

vmoens (commented):
Almost there! It's still a bit fuzzy to me how we're supposed to work with these modules in a way that handles single steps and trajectories interchangeably (e.g. using the same module at loss compute time and during policy execution while benefitting from the cudnn kernel at loss time).

But it's more an open question than a comment about this PR.
One option could be to use first-class dimensions, I'll explore that and keep you posted.

torchrl/modules/models/models.py
```diff
@@ -1034,3 +1035,90 @@ def forward(
...
+
+
+class GRUNet(nn.Module):
+    """
+    The GRUNet is a neural network composed of a GRU layer encapsulated
+    between two MLPs.
+
+    Args:
+        mlp_input_kwargs (dict):  kwargs for the MLP before the GRU
+        mlp_output_kwargs (dict): kwargs for the MLP after the GRU
```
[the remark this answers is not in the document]
PaLeroy: Another possibility is to have `input_features`, `hidden_size` and `output_features` as non-optional arguments that would create a default GRUNet such as the one in the example I just pushed, and having these 3 param dictionaries as optional and used only if a different network configuration is required.
What do you think?

torchrl/modules/models/models.py
PaLeroy: Only present in the docstring to document sizes of inputs and outputs.
They would become irrelevant with the suggestion I just made to change the arguments of GRUNet.

torchrl/modules/models/models.py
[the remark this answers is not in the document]
vmoens: Sounds great!

torchrl/modules/models/models.py
```diff
@@ -1034,3 +1035,191 @@ def forward(
...
+                "The size of output features in mlp_input must be the same as the input size of the GRU."
+            )
+        if mlp_output_kwargs["in_features"] != gru_kwargs["hidden_size"]:
+            raise ValueError(
+                "The size of input features in mlp_output must be the same as the last hidden size of the GRU."
+            )
+        if "bidirectional" in gru_kwargs and gru_kwargs["bidirectional"]:
+            raise NotImplementedError("bidirectional GRU is not yet implemented.")
+
+        self.device = device
```
vmoens, line 1210: we should avoid having `self.device`
I know we have some remaining in the lib but it's bad practice as more and more models are dispatched over multiple devices nowadays.

torchrl/modules/models/models.py
```diff
@@ -1034,3 +1035,191 @@ def forward(
...
+        mlp_input_kwargs.update({"device": self.device})
+        gru_kwargs.update({"device": self.device, "batch_first": True})
+        mlp_output_kwargs.update({"device": self.device})
+
+        self.mlp_in = MLP(**mlp_input_kwargs)
+        self.gru = nn.GRU(**gru_kwargs)
+        self.mlp_out = MLP(**mlp_output_kwargs)
+
+    def forward(self, x, h_0=None):
+        if 2 > len(x.size()) > 3:
```
vmoens, line 1220: maybe if we have more than 3 dims, we could flatten and then unflatten the first dims (only if the rnn batch_first is True)?

torchrl/modules/models/models.py
```diff
@@ -1034,3 +1035,191 @@ def forward(
...
+              (mlp_in): MLP(
+                (0): Linear(in_features=11, out_features=13, bias=True)
+                (1): ReLU()
+              )
+              (gru): GRU(13, 13, batch_first=True)
+              (mlp_out): MLP(
+                (0): Linear(in_features=13, out_features=3, bias=True)
+              )
+            )
+        >>> net_stacked = GRUNet(
```
vmoens, line 1115: This should look like this
```python
        >>> net_stacked = GRUNet(
        ...     in_features=123,
        ...     hidden_size=456,
        ...     out_features=3,
        ...     mlp_input_kwargs={
        ...         "in_features": 11,
        ...         "out_features": 13,
        ...     },
        ...     gru_kwargs={
        ...         "input_size": 13,
        ...         "hidden_size": 13,
        ...         "num_layers": 2,
        ...     },
        ...     mlp_output_kwargs={
        ...         "in_features": 13,
        ...         "depth": 0
        ...     },
        ... )
```

torchrl/modules/models/models.py
```diff
@@ -1034,3 +1035,191 @@ def forward(
...
+            )
+        >>> x_no_batch = torch.randn(7, 11)
+        >>> out_no_batch, h_no_batch = net(x_no_batch)
+        >>> print(out_no_batch.shape, h_no_batch.shape)
+        torch.Size([7, 3]) torch.Size([1, 13])
+        >>> x_batch = torch.randn(5, 7, 11)
+        >>> out_batch, h_batch = net(x_batch)
+        >>> print(out_batch.shape, h_batch.shape)
+        torch.Size([5, 7, 3]) torch.Size([1, 5, 13])
+        >>> net2 = GRUNet(
```
vmoens, line 1094: see comment below about formatting

vmoens (approved):
Wonderful just a couple of things to fix and we're good!

torchrl/modules/models/models.py
PaLeroy: Since we enforce the gru to be batch first, I guess we can.

PaLeroy:
Finally, I think we can allow input of any dimension.
Like you suggested with dim > 3, we can also unsqueeze here for dim = 1.
As we enforce batch_first, I added the warning if someone tries to use GRUNet without batch_first.

vmoens (approved):
This is absolutely wonderful!
I can help to make a TensorDict wrapper around that, if needed

PR #608, opened 2022-10-25
Add support for null `dim` argument in `TensorDict.squeeze`

### LINKED ISSUE #592: [BUG] `TensorDict.squeeze` does not support an null `dim` argument
## Describe the bug

While `torch.squeeze` can squeeze a tensor of all its singleton dimensions, the squeeze operation over a tensordict does not.
This discrepancy should be solved.

jgonik:
## Description

I began adding support for a null `dim` argument in `TensorDict.squeeze`. However, this code is still buggy as `TensorDict.unsqueeze` is called during the process of creating a SqueezedTensorDict, and `unsqueeze` should not support a null `dim` argument.

## Motivation and Context

This resolves #592.

## Types of changes

## Checklist

### DIFF
--- a/torchrl/data/tensordict/tensordict.py
+++ b/torchrl/data/tensordict/tensordict.py
@@ -1159,7 +1159,7 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
             inv_op_kwargs={"dim": dim},
         )
 
-    def squeeze(self, dim: Optional[int]=None) -> TensorDictBase:
+    def squeeze(self, dim: Optional[int] = None) -> TensorDictBase:
         """Squeezes all tensors for a dimension comprised in between `-td.batch_dims+1` and `td.batch_dims-1` and returns them in a new tensordict.
 
         Args:
@@ -4510,7 +4510,9 @@ class SqueezedTensorDict(_CustomOpTensorDict):
         # dim=1, squeezed_dim=2, [3, 4, 5] [3, 4, 1, 5] [[3, 5], [3, 5], [3, 5], [3, 4]] => unsq 1
         # dim=2, squeezed_dim=2, [3, 4, 5] [3, 4, 1, 5] [[3, 4], [3, 4], ...] => unsq 2
         diff_to_apply = 1 if dim < squeezed_dim else 0
-        list_item_unsqueeze = [item.unsqueeze(squeezed_dim - diff_to_apply) for item in list_item]
+        list_item_unsqueeze = [
+            item.unsqueeze(squeezed_dim - diff_to_apply) for item in list_item
+        ]
         return self._source._stack_onto_(key, list_item_unsqueeze, dim)
 
 
--- a/torchrl/data/tensordict/tensordict.py
+++ b/torchrl/data/tensordict/tensordict.py
@@ -1159,24 +1159,24 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
             inv_op_kwargs={"dim": dim},
         )
 
-    def squeeze(self, dim: int) -> TensorDictBase:
+    def squeeze(self, dim: Optional[int]) -> TensorDictBase:
         """Squeezes all tensors for a dimension comprised in between `-td.batch_dims+1` and `td.batch_dims-1` and returns them in a new tensordict.
 
         Args:
             dim (int): dimension along which to squeeze
 
         """
-        if dim < 0:
+        if dim and dim < 0:
             dim = self.batch_dims + dim
 
-        if self.batch_dims and (dim >= self.batch_dims or dim < 0):
+        if dim and self.batch_dims and (dim >= self.batch_dims or dim < 0):
             raise RuntimeError(
                 f"squeezing is allowed for dims comprised between 0 and "
                 f"td.batch_dims only. Got dim={dim} and batch_size"
                 f"={self.batch_size}."
             )
 
-        if dim >= self.batch_dims or self.batch_size[dim] != 1:
+        if dim and (dim >= self.batch_dims or self.batch_size[dim] != 1):
             return self
         return SqueezedTensorDict(
             source=self,
@@ -4446,8 +4446,8 @@ class UnsqueezedTensorDict(_CustomOpTensorDict):
         True
     """
 
-    def squeeze(self, dim: int) -> TensorDictBase:
-        if dim < 0:
+    def squeeze(self, dim: Optional[int]) -> TensorDictBase:
+        if dim and dim < 0:
             dim = self.batch_dims + dim
         if dim == self.custom_op_kwargs.get("dim"):
             return self._source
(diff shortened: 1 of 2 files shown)


torchrl/data/tensordict/tensordict.py
```diff
@@ -1159,24 +1159,24 @@ def unsqueeze(self, dim: int) -> TensorDictBase:
             inv_op_kwargs={"dim": dim},
         )
 
-    def squeeze(self, dim: int) -> TensorDictBase:
+    def squeeze(self, dim: Optional[int]) -> TensorDictBase:
         """Squeezes all tensors for a dimension comprised in between `-td.batch_dims+1` and `td.batch_dims-1` and returns them in a new tensordict.
 
         Args:
             dim (int): dimension along which to squeeze
```
vmoens, line 1166: let's correct the docstring too

torchrl/data/tensordict/tensordict.py
```diff
@@ -1159,24 +1159,24 @@ def unsqueeze(self, dim: int) -> TensorDictBase:
...
 
         """
-        if dim < 0:
+        if dim and dim < 0:
```
vmoens, line 1169: I guess you'd like to check that `dim` is not `None`
I'd rather write `if dim is not None and dim < 0` instead for clarity.

torchrl/data/tensordict/tensordict.py
```diff
@@ -1159,24 +1159,24 @@ def unsqueeze(self, dim: int) -> TensorDictBase:
...
             dim = self.batch_dims + dim
 
-        if self.batch_dims and (dim >= self.batch_dims or dim < 0):
+        if dim and self.batch_dims and (dim >= self.batch_dims or dim < 0):
```
vmoens, line 1172: same here

torchrl/data/tensordict/tensordict.py
```diff
@@ -1159,24 +1159,24 @@ def unsqueeze(self, dim: int) -> TensorDictBase:
...
             raise RuntimeError(
                 f"squeezing is allowed for dims comprised between 0 and "
                 f"td.batch_dims only. Got dim={dim} and batch_size"
                 f"={self.batch_size}."
             )
 
-        if dim >= self.batch_dims or self.batch_size[dim] != 1:
+        if dim and (dim >= self.batch_dims or self.batch_size[dim] != 1):
```
vmoens, line 1179: same here

torchrl/data/tensordict/tensordict.py
```diff
@@ -4446,8 +4446,8 @@ class UnsqueezedTensorDict(_CustomOpTensorDict):
         True
     """
 
-    def squeeze(self, dim: int) -> TensorDictBase:
-        if dim < 0:
+    def squeeze(self, dim: Optional[int]) -> TensorDictBase:
+        if dim and dim < 0:
```
vmoens, line 4450: same here

torchrl/data/tensordict/tensordict.py
```diff
@@ -4494,10 +4494,9 @@ def _stack_onto_(
         # dim=0, squeezed_dim=2, [3, 4, 5] [3, 4, 1, 5] [[4, 5], [4, 5], [4, 5]] => unsq 1
         # dim=1, squeezed_dim=2, [3, 4, 5] [3, 4, 1, 5] [[3, 5], [3, 5], [3, 5], [3, 4]] => unsq 1
         # dim=2, squeezed_dim=2, [3, 4, 5] [3, 4, 1, 5] [[3, 4], [3, 4], ...] => unsq 2
-        diff_to_apply = 1 if dim < squeezed_dim else 0
-        list_item_unsqueeze = [
-            item.unsqueeze(squeezed_dim - diff_to_apply) for item in list_item
-        ]
+        diff_to_apply = 1 if squeezed_dim and dim < squeezed_dim else 0
```
vmoens, line 4497: same here

torchrl/data/tensordict/tensordict.py
```diff
@@ -4494,10 +4494,9 @@ def _stack_onto_(
...
+        unsqueezed_dim = squeezed_dim - diff_to_apply if squeezed_dim else -1
```
vmoens, line 4498: same here

vmoens (commented):
Instead of checking multiple times if `dim is None` I would create a var

```
dim_is_none = dim is None

if not dim_is_none and dim < 0:
   etc
```

vmoens:
The errors seem to come from `td.set_("a", ...)` which expects the tensor to be unsqueezed before being updated.
The solution is (I guess) that calling `squeeze(None)` will actually lead `None` to become a real dim index.

e.g.
```python
td = TensorDIct({}, [3, 1, 4])
td_sq = td.squeeze()
print(td_sq.custom_op_kwargs["dim"])
# prints 1
```

Do you think it's feasible? 
If there are more than one singleton dim, then we'd have a SqueezedTensorDict of a SqueezedTensorDict.

jgonik:
> The solution is (I guess) that calling `squeeze(None)` will actually lead `None` to become a real dim index.

Wouldn't this break the invariant that calling td.squeeze(dim).unsqueeze(dim) should give you the same TensorDict back? Is that something that we care about maintaining? Also, if None is converted to a real dim arbitrarily, that might lead to behavior that users aren't expecting, especially if they've used unsqueeze in regular PyTorch before

vmoens:
> Wouldn't this break the invariant that calling td.squeeze(dim).unsqueeze(dim) should give you the same TensorDict back? 

If I understand your point, it should also give the same tensordict BUT the dim of unsqueeze must match the dim inferred during squeezing.
Unsqueezing without dim (ie with null dim) is meaningless (you can't know what dim to unsqueeze).

vmoens:
> 
The solution is (I guess) that calling squeeze(None) will actually lead None to become a real dim index.

Wouldn't this break the invariant that calling td.squeeze(dim).unsqueeze(dim) should give you the same TensorDict back? Is that something that we care about maintaining? Also, if None is converted to a real dim arbitrarily, that might lead to behavior that users aren't expecting, especially if they've used unsqueeze in regular PyTorch before

Not sure I'm following.
When users do `torch.ones(2,1,2,1).squeeze()` they expect that this works as `torch.ones(2,1,2,1).squeeze(3).squeeze(1)`. `unsqueeze` with `None` is not allowed IIRC (it would only make sense for tensor without dimension: `torch.tensor(1.0).unsqueeze()` could mean something but in all other cases it's hard to know what to do with this).

vmoens:
> 
The solution is (I guess) that calling squeeze(None) will actually lead None to become a real dim index.

Wouldn't this break the invariant that calling td.squeeze(dim).unsqueeze(dim) should give you the same TensorDict back? Is that something that we care about maintaining? Also, if None is converted to a real dim arbitrarily, that might lead to behavior that users aren't expecting, especially if they've used unsqueeze in regular PyTorch before

Not sure I'm following.
When users do `torch.ones(2,1,2,1).squeeze()` they expect that this works as `torch.ones(2,1,2,1).squeeze(3).squeeze(1)`. `unsqueeze` with `None` is not allowed IIRC (it would only make sense for tensor without dimension: `torch.tensor(1.0).unsqueeze()`).

jgonik:
> `unsqueeze` with `None` is not allowed IIRC

Yes, this was what I was getting at. So should the inferred dims be stored when calling `squeeze` with `dim=None` so that `unsqueeze` knows which dims to use?

vmoens:
> Yes, this was what I was getting at. So should the inferred dims be stored when calling `squeeze` with `dim=None` so that `unsqueeze` knows which dims to use?

No I think `unsqueeze(None)` is not something we want to support, it is not obvious what it should do.
```python
tensordict = TensorDict({}, [1, 2, 3])
assert tensordict.squeeze().unsqueeze(0) is tensordict  # that should work
assert tensordict.squeeze().unsqueeze(1) is tensordict  # that should break
tensordict.squeeze().unsqueeze()  # that should break
```

jgonik:
> > Yes, this was what I was getting at. So should the inferred dims be stored when calling `squeeze` with `dim=None` so that `unsqueeze` knows which dims to use?
> 
> No I think `unsqueeze(None)` is not something we want to support, it is not obvious what it should do.
> 
> ```python
> tensordict = TensorDict({}, [1, 2, 3])
> assert tensordict.squeeze().unsqueeze(0) is tensordict  # that should work
> assert tensordict.squeeze().unsqueeze(1) is tensordict  # that should break
> tensordict.squeeze().unsqueeze()  # that should break
> ```

Got it, I think I misread your first comment about the errors. Thanks for the clarification!

vmoens (approved):
Wonderful thanks for this feature!

PR #614, opened 2022-10-26
[BugFix] ReplayBuffer's storage now signal back when changes happen

### LINKED ISSUE #606: [BUG] Composable replay buffers with shared storage break when a prioritized exist but is not extended
## Context

We have a prototype of composable replay buffer.
If has 4 component: 
- A storage: either a list or some other form of storage that keeps a register of tensors
- A writer that writes stuff in the storage
- A sampler, which you use to get random items from the storage.

Examples: Writer can be a FIFO or something else, the storage can be on disk or in mem, the sampler can be uniform or prioritized (ie weighted)
The final element is replay buffer itself, which handles the 3 previous elements.

One interesting use case is that we can build multiple RB that share a component: for instance you could have 2 replay buffers with same storage but different sampling strategy.

## Describe the bug

Now say I have 2 replay buffers, one with a uniform sampling and another one with a prioritized sampling strategy. We assume both to share the same storage and writer.

If I update the second (prioritized) the storage should be informed that new elements have come in and it should assign them some default weight value.

But if I update the first (uniform) it has no idea that somewhere there is a prioritized sampler attached to the storage.

When I query my prioritized sampler, it will tell me that there are no weights assigned to some elements (or raise a strange issue).

## To Reproduce

If two replay buffers share the same storage, one is prioritized and the other not, extending the one with random sampling breaks the other:
```python
# let's try another thing: extend rb1

n = 100


storage = LazyMemmapStorage(n)
writer = RoundRobinWriter()
sampler0 = RandomSampler()
sampler1 = PrioritizedSampler(max_capacity=n, alpha=0.7, beta=1.1)

rb0 = ReplayBuffer(storage=storage, writer=writer, sampler=sampler0, collate_fn=lambda x: x)
rb1 = ReplayBuffer(storage=storage, writer=writer, sampler=sampler1, collate_fn=lambda x: x)

data = TensorDict({"a": torch.arange(50)}, [50])
rb0.extend(data)

assert len(rb0) == 50
assert len(storage) == 50
assert len(rb1) == 50

rb0.sample(10) # works
rb1.sample(10) # breaks
```

If in the code snippet above, `rb0.extend(data)` is replaced by `rb1.extend(data)` everything works fine.

## Expected behavior

Extending one or the other replay buffer should work.

## Additional context

Ideally the bug fix should consider the following questions:
(1) Similar bugs might rise in the future, and we don't want a solution that is too tailored to Prioritized replay buffers.
Currently, `ReplayBuffer.add` calls `ReplayBuffer._sampler.add()`, but what that should probably be changed.
Perhaps it should be for the storage to call `add` on all the siblings samplers it has: that would involve that when we register a storage in a buffer, the storage should be informed of its new parent buffer, and look for siblings there.
Something like
```
class Storage:
   def __init__(self, parent: ReplayBuffer):
       self.list_of_parents.append(parent)

   def extend(self, data):
       # do stuff
       ...
       # update all samplers
       for parent in self.list_of_parents:
          parent.sampler.extend(data)
```

(2) another thing to keep in mind is whether it would make sense to have different writers for a single storage. I guess someday someone might want this, but it might also be dangerous.
This can be left as an open issue -- it's not directly linked to this one.

paulomarciano:
## Description

The prototype modular Replay Buffer has a bug (#606) when storage is shared across buffers. This happens because the different parts implicitly assume to always be aware of changes in storage, which was no true. This change introduces an API for storage to signal back to the buffers that changes were made. Buffers can then handle this signal however and whenever they see fit.

## Motivation and Context

close #606 

## Types of changes

## Checklist

### DIFF
--- a/torchrl/data/replay_buffers/rb_prototype.py
+++ b/torchrl/data/replay_buffers/rb_prototype.py
@@ -42,8 +42,8 @@ class ReplayBuffer:
         prefetch: Optional[int] = None,
     ) -> None:
         self._storage = storage if storage is not None else ListStorage(max_size=1_000)
+        self._storage.attach(self)
         self._sampler = sampler if sampler is not None else RandomSampler()
-        self._storage.attach(self._sampler)
         self._writer = writer if writer is not None else RoundRobinWriter()
         self._writer.register_storage(self._storage)
 
@@ -155,6 +155,9 @@ class ReplayBuffer:
                 self._prefetch_queue.append(fut)
 
         return ret
+
+    def mark_update(self, index: Union[int, torch.Tensor]) -> None:
+        self._sampler.mark_update(index)
 
 
 class TensorDictReplayBuffer(ReplayBuffer):
--- a/torchrl/data/replay_buffers/rb_prototype.py
+++ b/torchrl/data/replay_buffers/rb_prototype.py
@@ -42,8 +42,8 @@ class ReplayBuffer:
         prefetch: Optional[int] = None,
     ) -> None:
         self._storage = storage if storage is not None else ListStorage(max_size=1_000)
-        self._storage.attach(self)
         self._sampler = sampler if sampler is not None else RandomSampler()
+        self._storage.attach(self._sampler)
         self._writer = writer if writer is not None else RoundRobinWriter()
         self._writer.register_storage(self._storage)
 
@@ -155,17 +155,6 @@ class ReplayBuffer:
                 self._prefetch_queue.append(fut)
 
         return ret
-
-    def mark_update(self, index) -> None:
-        """Marks a given storage index as having changed.
-
-        Derived classes can deal with this however appropriate,
-        forwarding this call to whichever parts are needed.
-
-        Args:
-            index: The modified index from storage.
-        """
-        return self.update_priority(index, self._sampler.default_priority)
 
 
 class TensorDictReplayBuffer(ReplayBuffer):
--- a/torchrl/data/replay_buffers/rb_prototype.py
+++ b/torchrl/data/replay_buffers/rb_prototype.py
@@ -42,6 +42,7 @@ class ReplayBuffer:
         prefetch: Optional[int] = None,
     ) -> None:
         self._storage = storage if storage is not None else ListStorage(max_size=1_000)
+        self._storage.attach(self)
         self._sampler = sampler if sampler is not None else RandomSampler()
         self._writer = writer if writer is not None else RoundRobinWriter()
         self._writer.register_storage(self._storage)
@@ -154,6 +155,17 @@ class ReplayBuffer:
                 self._prefetch_queue.append(fut)
 
         return ret
+
+    def mark_update(self, index) -> None:
+        """Marks a given storage index as having changed.
+
+        Derived classes can deal with this however appropriate,
+        forwarding this call to whichever parts are needed.
+
+        Args:
+            index: The modified index from storage.
+        """
+        return self.update_priority(index, self._sampler.default_priority)
 
 
 class TensorDictReplayBuffer(ReplayBuffer):
--- a/torchrl/data/replay_buffers/storages.py
+++ b/torchrl/data/replay_buffers/storages.py
@@ -41,18 +41,18 @@ class Storage:
     def get(self, index: int) -> Any:
         raise NotImplementedError
 
-    def attach(self, sampler: Any) -> None:
+    def attach(self, buffer: Any) -> None:
         """This function attaches a sampler to this storage.
 
-        Samplers that read from this storage must be included as an attached
+        Buffers that read from this storage must be included as an attached
         entity by calling this method. This guarantees that when data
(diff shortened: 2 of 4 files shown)


torchrl/data/replay_buffers/storages.py
```diff
@@ -29,6 +29,9 @@ class Storage:
 
     def __init__(self, max_size: int) -> None:
         self.max_size = int(max_size)
+        # Prototype feature. RBs that use a given instance of Storage should add
+        # themselves to this set.
+        self.attached_entities = set()
```
vmoens, line 34: do we want this to be public? Is any other object accessing that?

torchrl/data/replay_buffers/storages.py
```diff
@@ -38,11 +41,28 @@ def set(self, cursor: int, data: Any):
...
+        """
+        self.attached_entities.add(buffer)
+
     def __getitem__(self, item):
         return self.get(item)
 
     def __setitem__(self, index, value):
-        return self.set(index, value)
+        ret = self.set(index, value)
+        for i in self.attached_entities:
```
vmoens, line 63: usually (at least in our code base) `i` points to an integer.
Can we rename this?

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -155,6 +156,17 @@ def sample(self, batch_size: int) -> Tuple[Any, dict]:
...
+    def mark_update(self, index) -> None:
+        """Marks a given storage index as having changed.
+
+        Derived classes can deal with this however appropriate,
+        forwarding this call to whichever parts are needed.
+
+        Args:
+            index: The modified index from storage.
+        """
+        return self.update_priority(index, self._sampler.default_priority)
```
vmoens, line 168: That's cool
In the future we will probably want the user to update the priority of the sampler directly, without passing through the replay buffer (otherwise the replay buffer will need to implement custom update methods for all the modules e put in it).

One fix for now could be that the sampler has a `mark_update` method that falls back on `update_priority` for prioritized sampling. If sampling is uniform, `mark_update` is a no-op.
Like this we could remove the `ReplayBuffer.update_priority` method altogether (in a future PR).

Would that make sense?

vmoens (commented):
Awesome and really quick!
Thanks!
Can we discuss about `ReplayBuffer.update_priority` which seems a bit too customize for prioritized sampling?

torchrl/data/replay_buffers/storages.py
paulomarciano: Nope. I can rename it.

torchrl/data/replay_buffers/storages.py
paulomarciano: sure.

torchrl/data/replay_buffers/rb_prototype.py
paulomarciano: It makes sense, yes.

I debated with myself for a while if we should attach samplers to storage or if we should attach RBs to storage, and decided for the latter because it is more extensible: It's possible that we need components other than the sampler to be aware of changes, and it's the RB's duty to coordinate the different moving parts.

The counterpoint to this would be YAGNI. It's simpler if we attach the samplers directly. I can definitely make that change.

By the way, another advantage of having the `mark_update` method instead of directly calling `update_priority` is that samplers can implement the update whenever they feel like it. We could, for example, simply store the modified indexes inside the sampler until the next call to sample, then update everything lazily.

vmoens: > I debated with myself for a while if we should attach samplers to storage or if we should attach RBs to storage, and decided for the latter because it is more extensible: It's possible that we need components other than the sampler to be aware of changes, and it's the RB's duty to coordinate the different moving parts.

I fully agree. I still think we should attach the RB to the storage, not all other children (the parent is the minimal sufficient information).

My point is mostly about generality of the name `update_priority` in the replay buffer. I would rather name that `mark_update`:

```python
class ReplayBuffer:
    def mark_update(self, ...):
        self.sampler.mark_update(...)

class PrioritizedSampler:
    def mark_update(self, ...)
        self.update_priority(...)
   def update_priority(self, ...):
        foo(...)
```

> By the way, another advantage of having the mark_update method instead of directly calling update_priority is that samplers can implement the update whenever they feel like it. We could, for example, simply store the modified indexes inside the sampler until the next call to sample, then update everything lazily.

Yes that would be cool! That would definitely be an advantage (we could also update things in batch when needed, which I would expect to cut the compute time of those ops by a bit)

paulomarciano: Ah yes. Just made that change.

vmoens (approved):
LGTM, thanks for the high-quality work

Can I ask you to create an issue for the lazy implementation of `mark_update` -> `update_priority` we talked about?
I can assign someone else to take care of it

vmoens:
I think not all tests are running now, probably because you create the PR from your main branch an not a separate branch. Not sure what to do in that case. I will run the tests locally and see if it's all working properly.

PR #618, opened 2022-10-27
[Feature] Added `implement_for` decorator

ordinskiy:
## Description

Added the decorator that allows to specify with which version of a given module a function is compatible with. Fitting implementation is selected basing on installed version of the module. 

## Motivation and Context

It is required for bc-braking packages (e.g. "gym"), so we can have several implementations of the same function for different versions of a module. Proposed by @vmoens.

## Types of changes

## Checklist


ordinskiy:
@vmoens I have a question about the following case: when installed version of the module doesn't fit any implementation.
For example: we have gym 0.26 and 2 functions with `implement_for('gym`, '0', '0.24')` and  `implement_for('gym`, '0.24', '.25')`.
Currently, I implemented the decorator, so it just returns the latest declared implementation, but I thought about more strict behaviour, when decorator returns a function, that raises an error (something like "Installed version of 'gym' is not supported), so when the function (with @implement_for) is called this exception is raised.

Which one is preferable?

vmoens:
> Currently, I implemented the decorator, so it just returns the latest declared implementation, but I thought about more strict behaviour, when decorator returns a function, that raises an error (something like "Installed version of 'gym' is not supported), so when the function (with @implement_for) is called this exception is raised.

One option could be that if we use `@implemented_for` we must cover all versions:
```python
@implemented_for("gym", None, "0.13")
def fun():
   foo()
@implemented_for("gym", "0.13", None)
def fun():
   foo()
```
is valid but 
```python
@implemented_for("gym", None, "0.13")
def fun():
   foo()
@implemented_for("gym", "0.13", "0.14")
def fun():
   foo()
```
will return an exception during import.

There are caveats with this (it's an optimistic view on future releases), but at least if nothing is expected to break with a new release our CI will just pass. If something breaks, it means we need one more function implementation.

What do you think?

ordinskiy:
With what I suggested as more strict option there won't be an import time exception. The decorated function will become something like:
```
def fun():
   raise NotImplementedError("Installed version of 'gym' is not supported")
```
So exception will be thrown only if `fun()` is actually called, despite the unsupported version, so it should not break CI if the test is skipped. I was thinking to do the same for missing module.

Such strict behaviour is not typical for python and forcing user to implement "extra" functions might not be a good idea, that is why I decided to implement it like I did. Unless, you think that catching the situation when function is called with unsupported (potentially) version of a module beforehand is more often beneficial than inconvenient.

Also, I just realised that I haven't considered "from_version" as None for some reason, but it does look reasonable to allow it, I will do it.

vmoens:
> So exception will be thrown only if `fun()` is actually called, despite the unsupported version, so it should not break CI if the test is skipped. I was thinking to do the same for missing module.
> 
> Such strict behaviour is not typical for python and forcing user to implement "extra" functions might not be a good idea, that is why I decided to implement it like I did. Unless, you think that catching the situation when function is called with unsupported (potentially) version of a module beforehand is more often beneficial than inconvenient.

My suggestion was a bit in the spirit of `@abc.abstractmethod` where you get an error at loading time if something is not implemented. But an error during execution is equally fine!

ordinskiy:
> > So exception will be thrown only if `fun()` is actually called, despite the unsupported version, so it should not break CI if the test is skipped. I was thinking to do the same for missing module.
> > Such strict behaviour is not typical for python and forcing user to implement "extra" functions might not be a good idea, that is why I decided to implement it like I did. Unless, you think that catching the situation when function is called with unsupported (potentially) version of a module beforehand is more often beneficial than inconvenient.
> 
> My suggestion was a bit in the spirit of `@abc.abstractmethod` where you get an error at loading time if something is not implemented. But an error during execution is equally fine!

Alright, I will change the behaviour, so the decorator will replace the function with a mock that raises the exception by default (until fitting implementation found).
Plus, I missed the case when two functions in different modules have the same name, I will fix it.

test/test_utils.py
```diff
@@ -60,3 +65,51 @@ def test_get_binary_env_var_wrong_value():
...
+def implement_for_test_missing_version():  # noqa: F811
+    return "0.4+"
+
+
+def test_implement_for():
+    assert implement_for_test_func() == "0.3+"
+
+
+def test_implement_for_missing_module():
+    with pytest.raises(ModuleNotFoundError):
```
vmoens, line 109: can we add a `match="msg"` here?

test/test_utils.py
```diff
@@ -60,3 +65,51 @@ def test_get_binary_env_var_wrong_value():
...
+        implement_for_test_missing_module()
+
+
+def test_implement_for_missing_version():
+    with pytest.raises(ModuleNotFoundError):
```
vmoens, line 114: can we add a `match="msg"` here?

test/test_utils.py
```diff
@@ -60,3 +65,51 @@ def test_get_binary_env_var_wrong_value():
     finally:
         if key in os.environ:
             del os.environ[key]
+
+
+# To test from+ range and that this function is correctly selected as the implementation.
+@implement_for("_utils_internal", "0.3")
```
vmoens, line 71: Can we wrap those tests under a `TestVersioning` class or something like that?

torchrl/_utils.py
```diff
@@ -176,3 +179,63 @@ def __repr__(self):
 
 
 _CKPT_BACKEND = _Dynamic_CKPT_BACKEND()
+
+
+class implement_for:
```
vmoens, line 184: Can we add this to the doc under `docs/source/`? We may need to create a special `.rst` file for the top-level utils.

vmoens (approved):
Just a couple of minor fixes and we're good to go! 
Thanks for this!

test/test_utils.py
ordinskiy: Done with `implement_for_test_functions` class.

torchrl/_utils.py
ordinskiy: Added `implement_for` to the doc. @vmoens let me know if you want to add other methods from _utils.py to the documentation.

PR #622, opened 2022-10-27
[Doc] Integrate knowledge base in docs

hatala91:
## Description

With this pull request, I propose to include the existing knowledge base articles in our [documentation](https://pytorch.org/rl/).

Since Sphinx doesn't natively support files outside the source directory, I am generating one file per knowledge base entry to reference it using the `rst`/ `doctool` include directive and `myst` for parsing markdown.

https://user-images.githubusercontent.com/19345882/198412236-3991f8ce-5a31-4499-95c9-75999d97ff3f.mov

## Motivation and Context

Making the knowledge base more accessible by including it in the docs.

## Types of changes

## Checklist


docs/source/knowledge_base.py
```diff
@@ -0,0 +1,50 @@
...
+from pathlib import Path
+from typing import List
+
+FILE_DIR = os.path.dirname(__file__)
+GEN_DIR = "reference/generated/knowledge_base"
+
+
+def _get_file_content(name: str) -> List[str]:
+    """
+    To include markdown files from the knowledge graph in Sphinx, these are linked using the rst include command. This
```
vmoens, line 11: The proper way of writing docstrings should be without a new line after the triple quote, and a short description first
```python
def fun():
    """This function does X.

    A way more elaborate explanation.
    """
    return foo
```

docs/source/knowledge_base.py
```diff
@@ -0,0 +1,50 @@
...
+        "\n",
+        f".. include:: ../../../../../knowledge_base/{name}.md\n",
+        "   :parser: myst_parser.sphinx_\n",
+        "\n",
+    ]
+
+
+def generate_knowledge_base_references(knowledge_base_path: str) -> None:
+    """
+    Sphinx natively doesn't support adding files from outside it's root directory. To include the knowledge base in
```
vmoens, line 29: same as above.

it's -> its

vmoens (approved):
Wonderful! Just a couple of nits

docs/source/knowledge_base.py
hatala91: You are absolutely right, thanks!

docs/source/knowledge_base.py
hatala91: Good catch 👀

PR #641, opened 2022-11-03
[BugFix] Prevent transform parent from being reassigned

### LINKED ISSUE #636: [BUG] Prevent the same transform to be placed twice in a transform list
## Describe the bug

Placing the same transform object twice in a TransformedEnv leads to poorly defined behaviour when calling `transform.parent`.

## To Reproduce

```python
from torchrl.envs import TransformedEnv, RewardScaling, RewardClipping, Compose

env =  TransformedEnv(GymEnv("Pendulum-v1"))
t1 = RewardScaling(0, 1)
t2 = RewardClipping(0.1, 0.5)
env.append_transform(t1)
env.append_transform(t2)
env.append_transform(t1)
print(env.transform[-1])
```

## Expected behavior

This prints an env without transform, but if anything it should show an env with transforms `t1` and `t2`.

## Reason and Possible fixes

Once a transform has a non-null `_parent`, it should not be possible to put it in a transformed env anymore (or there should be a strong reset method that should allow to do that, but it's unclear how it should signal to the parent TransformedEnv that this transform has changed parent).

jasonfkut:
## Description

Once a transform has been assigned a parent, it can't be assigned another parent anymore. A check in the Transform.set_parent method was added which raises an exception if the parent has already been set. A reset_parent method was also added so existing operations such as appending new transforms work with the new check.

## Motivation and Context

Following code should raise an exception because we are reassigning t1's parent:

```
env =  TransformedEnv(GymEnv("Pendulum-v1"))
t1 = RewardScaling(0, 1)
t2 = RewardClipping(0.1, 0.5)
env.append_transform(t1)
env.append_transform(t2)
env.append_transform(t1)
print(env.transform[-1])
```

`close #636`

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -230,7 +230,7 @@ class Transform(nn.Module):
 
     def set_parent(self, parent: Union[Transform, EnvBase]) -> None:
         if self.__dict__["_parent"] is not None:
-            raise AttributeError(f"parent of transform already set")
+            raise AttributeError("parent of transform already set")
         self.__dict__["_parent"] = parent
 
     def reset_parent(self) -> None:
@@ -257,7 +257,7 @@ class Transform(nn.Module):
                     f"Compose parent was of type {type(compose_parent)} but expected TransformedEnv."
                 )
             if compose_parent.transform is not compose:
-                comp_parent_trans = compose_parent.transform
+                comp_parent_trans = copy(compose_parent.transform)
                 comp_parent_trans.reset_parent()
             else:
                 comp_parent_trans = None
@@ -265,9 +265,10 @@ class Transform(nn.Module):
                 compose_parent.base_env,
                 transform=comp_parent_trans,
             )
-            for transform in compose.transforms:
-                if transform is self:
+            for orig_trans in compose.transforms:
+                if orig_trans is self:
                     break
+                transform = copy(orig_trans)
                 transform.reset_parent()
                 out.append_transform(transform)
         elif isinstance(parent, TransformedEnv):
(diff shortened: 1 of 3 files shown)


torchrl/envs/transforms/transforms.py
```diff
@@ -251,15 +256,19 @@ def parent(self) -> EnvBase:
                 raise ValueError(
                     f"Compose parent was of type {type(compose_parent)} but expected TransformedEnv."
                 )
+            if compose_parent.transform is not compose:
+                comp_parent_trans = compose_parent.transform
+                comp_parent_trans.reset_parent()
```
vmoens, line 261: are we sure that the whole transformed env remains unchanged after we call `transform.parent` in this case?

torchrl/envs/transforms/transforms.py
```diff
@@ -312,9 +321,16 @@ def __init__(
                 # we don't use isinstance as some transforms may be subclassed from
                 # Compose but with other features that we don't want to loose.
                 transform = [transform]
+            else:
+                for t in transform:
+                    t.reset_parent()
```
vmoens, line 326: same thing: does this modify the original `TransformedEnv` in any way?

vmoens (commented):
Looks promising
I'm wondering if I did not overlook something there:
When we call `transform.parent` we recreate a `TranformedEnv(base_env, Compose(*list_of_transforms_up_to_self))`, but by doing so we reassign a new `_parent` to each of the transforms before this one!
This will usually go unnoticed:
Say you have 3 transforms, and you call the parent of the second, you'll have
`transform2.parent` => `TransformedEnv(base_env, Compose(transform1))`. The result of `transform1.parent` has not changed BUT the `transform1._parent` has changed.
On the other hand the `_parent` of `transform3 will remain unchanged.

However there may be cases where this behaviour leads to bug (can't think of any on the top of my head). Do you have thoughts on this?

torchrl/envs/transforms/transforms.py
jasonfkut: I was wondering the same thing - Since a new `TransformedEnv` object is created in `transform.parent` we would need to make copies of all the transforms added to this new object in order to keep the original `_parent` attributes pointing to the correct (original) 'TransformedEnv'. Does that make sense to do here?

vmoens:
I guess we could do a shallow copy yes, that cheap but should behave well.

jasonfkut:
Sounds good, I'll make the changes and add a test as well

jasonfkut:
No prob! This is a cool project

PR #645, opened 2022-11-04
[CI] Added CircleCI pipeline to test compatibility across supported gym versions

ordinskiy:
## Description

New CircleCI pipeline that is based on `unittest_linux_olddeps_gpu` pipeline. It runs `test_libs` tests for fixed set of gym versions (0.13, 0.19, 0.25, 0.26). Each gym version has some specifics that are described and handled in `batch_scripts.sh`. The idea is gradually upgrade gym version installing only what is necessary.

## Motivation and Context

We want to make sure that TorchRL stays compatible with supported gym versions. Proposed by @vmoens.

## Types of changes

## Checklist


ordinskiy:
@vmoens please take a look at `batch_scripts.sh` in the PR when you have time, to verify that you are fine with this approach. Just wanted to make sure that I'm on the right path with the pipeline.
I don't particularly like the approach, because there is a chance that some packages are reused from the previous version (like in case of `ale-py`), but I couldn't come with a better way to do this.

vmoens:
Could we make a generic pipeline where the gym version would be a env variable, and pass that to each workflow (e.g. like we do for the python version)?
like this all these pipelines would run in parallel, thereby reducing the compute time

ordinskiy:
> Could we make a generic pipeline where the gym version would be a env variable, and pass that to each workflow (e.g. like we do for the python version)? like this all these pipelines would run in parallel, thereby reducing the compute time

I was considering this approach initially, however I disregarded it, because it seems to me a bit too heavy (an overkill). The part of the pipeline where environment is setup is much longer than part when we install gym and run tests (from what I've seen so far it something like 10+ minutes against seconds for smoke tests and `test_libs`). With the selected tests the compute time will be roughly the same (assuming that all gym pipelines run in parallel), but it will make these tests more reliable due to "cleaner" environment, so if resources is not the problem we can go this way with the pipeline.

vmoens (commented):
LGTM
I merged main into this branch to test the new "next" API
Can we move these tests to `linux_libs` instead of `linux_olddeps`?
It's debatable which one is better, but most of the `linux_libs` will just test `test_libs.py` (as this one does).

ordinskiy:
> Can we move these tests to `linux_libs` instead of `linux_olddeps`?
> It's debatable which one is better, but most of the `linux_libs` will just test `test_libs.py` (as this one does).

Sure, I wasn't so sure where to place it in the first place. I moved the scripts and did some renaming to match what is already there, also changed python version to 3.8 since there is a comment that libs are tested only with 3.8 version.

vmoens (approved):
Great stuff thanks a mil!

PR #647, opened 2022-11-05
[WIP, Refactor] Refactor 'next_' into nested tensordicts

vmoens:
## Description

Refactors the `"next_obs"` convention for steps in MDPs into a `("next", "obs")` nested tensordict approach.

PR #649, opened 2022-11-07
[Refactor] Refactor 'next_' into nested tensordicts

vmoens:
## Description

This is the proper parent PR of #647.


vmoens:
@raphajaner I think we're done with this.

Have a try at the new API. IMO it's much cleaner without those `"next_stuff"` everywhere, and easier to use.
For instance, when creating a transform you don't need to say that the keys is prefixed with `"next_"` anymore, the transform will take care of that for you.
Also the `reset` and `step` methods are much clearly coded.

Example:

```python
>>> from torchrl.envs.libs.gym import GymEnv
>>> from torchrl.envs import TransformedEnv, ObservationNorm, Compose, CatFrames
>>> env = GymEnv("HalfCheetah-v4")
>>> env = TransformedEnv(env, Compose(ObservationNorm(in_keys=["observation"], loc=0.5, scale=1.0, standard_normal=True), CatFrames(N=4, cat_dim=-1, in_keys=["observation"])))
>>> print(list(env.observation_spec.keys())) # specs do not start with next_ anymore
>>> td = env.reset()
>>> print("reset", td)
reset TensorDict(
    fields={
        done: Tensor(torch.Size([1]), dtype=torch.bool),
        observation: Tensor(torch.Size([68]), dtype=torch.float64)},
    batch_size=torch.Size([]),
    device=cpu,
    is_shared=False)
>>> td = env.rollout(4)
>>> print("rollout", td)
rollout TensorDict(
    fields={
        action: Tensor(torch.Size([4, 6]), dtype=torch.float32),
        done: Tensor(torch.Size([4, 1]), dtype=torch.bool),
        next: LazyStackedTensorDict(
            fields={
                observation: Tensor(torch.Size([4, 68]), dtype=torch.float64)},
            batch_size=torch.Size([4]),
            device=cpu,
            is_shared=False),
        observation: Tensor(torch.Size([4, 68]), dtype=torch.float64),
        reward: Tensor(torch.Size([4, 1]), dtype=torch.float32)},
    batch_size=torch.Size([4]),
    device=cpu,
    is_shared=False)
```

@shagunsodhani

PR #653, opened 2022-11-07
[BugFix] GymWrapper does not work with nested observation gym.spaces.Dict #640

### LINKED ISSUE #640: [BUG] GymWrapper does not work with nested observation gym.spaces.Dict
## Describe the bug

Hi All,

First of all: thanks for the great work here!

I think I have encountered a bug in the GymWrapper in ```torchrl.envs.libs.gym.GymWrapper```. When I use a ```gym.Env``` with an observation space with nested ```gym.spaces.Dict```, a KeyError will be thrown since the ```GymLikeEnv.read_obs()``` function does only add "next_" to the first level of Dict but not to nested sub Dicts:
```python
observations = {"next_" + key: value for key, value in observations.items()}
```
Since ```_gym_to_torchrl_spec_transform()``` in ```torchrl.envs.libs.gym``` ends "next_"  in a recursive call to all sub Dicts, the key is missing the necessary "next_". Nested Dict observation spaces are often used (https://www.gymlibrary.dev/api/spaces/#dict), so I guess this is required to work properly.

## To Reproduce
```python
#!/usr/bin/env python
from torchrl.envs.libs.gym import GymWrapper
from gym import spaces, Env
import numpy as np


class CustomGym(Env):
    def __init__(self):
        self.action_space = spaces.Discrete(5)
        self.observation_space = spaces.Dict(
            {
                'sensor_1': spaces.Box(low=0, high=255, shape=(5, 5, 3), dtype=np.uint8),
                'sensor_2': spaces.Box(low=0, high=255, shape=(5, 5, 3), dtype=np.uint8),
                'sensor_3': spaces.Box(np.array([-2, -1, -5, 0]), np.array([2, 1, 30, 1]), dtype=np.float32),
                'sensor_4': spaces.Dict({'sensor_41': spaces.Box(low=0, high=100, shape=(1,), dtype=np.float32),
                                         'sensor_42': spaces.Box(low=0, high=100, shape=(1,), dtype=np.float32),
                                         'sensor_43': spaces.Box(low=0, high=100, shape=(1,), dtype=np.float32)})
            }
        )

    def reset(self):
        return self.observation_space.sample()


if __name__ == '__main__':
    env = CustomGym()
    env = GymWrapper(env)

```

## Reason and Possible fixes
The issue can be fixed by adding a recursive function call to rename also nested observation space Dicts in ```GymLikeEnv.read_obs()``` correctly by adding "next_":
```python

    def read_obs(
        self, observations: Union[Dict[str, Any], torch.Tensor, np.ndarray]
    ) -> Dict[str, Any]:
        """Reads an observation from the environment and returns an observation compatible with the output TensorDict.

        Args:
            observations (observation under a format dictated by the inner env): observation to be read.

        """
        if isinstance(observations, dict):

            def rename(obs):
                return {
                    "next_" + key: rename(value) if isinstance(value, dict) else value
                    for key, value in obs.items()
                }

            observations = rename(observations)
        if not isinstance(observations, (TensorDict, dict)):
            key = list(self.observation_spec.keys())[0]
            observations = {key: observations}
        observations = self.observation_spec.encode(observations)
        return observations
```

The style checker required to not use lambda functions, otherwise the fix could also be as simple as
```python
             rename = lambda obs: {
                "next_" + key: rename(value) if isinstance(value, dict) else value
                for key, value in obs.items()
             }
```
## Checklist

raphajaner:
## Description
Possible bug fix for the issue mentioned in #640 that prevents to use nested `gym.spaces.Dict` due to naming errors with the "next_" word beginning of observations. This bug fix uses recursive functions to rename observations correctly for gym environments with nested `gym.spaces.Dict` spaces. Some ideas for tests have been added.

## Motivation and Context
Close #640 

## Types of changes

## Checklist

### DIFF
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -143,6 +143,51 @@ def test_gym(env_name, frame_skip, from_pixels, pixels_only):
     assert final_seed0 == final_seed2
     assert_allclose_td(tdrollout[0], rollout2, rtol=1e-4, atol=1e-4)
 
+    class NestedObsGymEnv(gym.Env):
+        def __init__(self):
+            self.action_space = gym.spaces.Discrete(5)
+            self.observation_space = gym.spaces.Dict(
+                {
+                    "1": gym.spaces.Discrete(5),
+                    "2": gym.spaces.Dict(
+                        {
+                            "21": gym.spaces.Discrete(5),
+                            "22": gym.spaces.Dict(
+                                {
+                                    "221": gym.spaces.Discrete(5),
+                                    "222": gym.spaces.Discrete(5),
+                                }
+                            ),
+                        }
+                    ),
+                }
+            )
+
+        def reset(self):
+            return self.observation_space.sample()
+
+    nested_obs_env = NestedObsGymEnv()
+    env = GymWrapper(nested_obs_env)
+
+    def assert_obs_name(obs, obs_spec):
+        for key, value in obs_spec.items():
+            assert key in obs.keys()
+            if isinstance(value, dict):
+                assert_obs_name(value, obs[key])
+
+    # internal function returns with "next_"
+    td = env._reset()
+    assert_obs_name(td, env.observation_spec)
+    del td
+
+    # removes "next_"
+    td = env.reset()
+    assert all(["next_" not in key for key in td.flatten_keys().keys()])
+    del td
+
+    env.close()
+    del env
+
 
 @pytest.mark.skipif(not _has_dmc, reason="no dm_control library found")
 @pytest.mark.parametrize("env_name,task", [["cheetah", "run"]])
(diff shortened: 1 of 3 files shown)


vmoens:
Thanks for this!
The solution looks great!

I'm in the process of refactoring the step logic.
The first step is to remove `"next_"` from the observation spec and add it in `env.step` after we call `env._step`, have a look at this: #661
Next I will refactor that in having the next observation in `tensordict["next"]` rather than `tensordict["next_stuff"]`.
This is a long process that involves multiple layers of refactoring, I would expect it to land by the end of week / early next week. 
Would you be willing to wait until them for me to finish this up? 
In the meantime, if you need this feature, you can work from this branch.
Does that sound ok? 
Am I missing anything with this refactoring?

raphajaner:
Sure, sounds perfectly fine and I don't think you're missing anything there!
 
Looking forward to seeing your changes, please let me know if I can be of any help :)

PR #655, opened 2022-11-08
[BugFix] make_trainer possible bug for on-policy cases

albertbou92:
## Description

Possible bug: using the make_trainer helper function in a script can raise an error if you want to use GPU, because mini batches seem to be placed by default in CPU if you don't use a replay buffer. I added a line to register a "process_optim_batch" operation so it places the mini batches to whatever device the networks are in, avoiding this error.

## Motivation and Context

I was trying to run the PPO example on GPU and kept complaining about the data and the weights being in different devices (CPU and GPU).

## Types of changes

## Checklist


vmoens (approved):
LGTM thanks for spotting this!

PR #665, opened 2022-11-10
Relying on the standalone tensordict -- phase 1 updates

tcbegley:
## Description

Updates to #650 TensorDict deprecation in favour of standalone [TensorDict](https://github.com/pytorch-labs/tensordict/). CI should be passing and I've killed remaining references to the TorchRL `TensorDict` class.

- [x] README
- [x] Examples
- [x] Docs

I haven't touched the tutorials as instructed.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -54,6 +54,9 @@ jobs:
       shell: bash
       run: |
         conda run -n build_binary python -m pip install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+    - name: Install tensordict
+      run: |
+        python3 -mpip install git+https://github.com/pytorch-labs/tensordict.git
     - name: Install TorchRL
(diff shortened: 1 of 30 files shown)


tcbegley:
The "old_deps" pipeline is failing, seemingly because an import from `torch.jit._shape_functions` is failing in `tensordict`.

I notice that TorchRL [catches the `ImportError` for this](https://github.com/pytorch/rl/blob/cd488b480d89d1df818bcaaadcef3e16b06434e2/torchrl/data/tensordict/tensordict.py#L39-L42) whereas [`tensordict` does not](https://github.com/pytorch-labs/tensordict/blob/20628b944ce14972fdfcb00101ebbac4e1424f07/tensordict/tensordict.py#L39). Is there a reason this was dropped?

vmoens (approved):
LGTM thanks!
Great that you updated the doctests too, i always forget them personally

PR #669, opened 2022-11-11
[Feature] Auto-compute stats for ObservationNorm

### LINKED ISSUE #565: [Feature Request] Auto-compute stats for ObservationNorm
## Motivation

It's be useful to have a method to compute automatically the stats of an env with observation norm.

## Proposed API
```python
base_env = GymEnv("HalfCheetah-v4", from_pixels=True)
env = TransformedENv(env, ObservationNorm(loc=None, scale=None)
env.transform.init_stats(num_iter=1000)
```

A sketch of `init_stats` would look like this
```python
def init_stats(self, num_iter, reduce_dim=0):
    if len(self.keys_in) > 1:
        raise RuntimeError # we don't know which key to use for init, perhaps that could be an input to this method?
    key = self.keys_in[0]
    if self.loc is not None or self.scale is not None:
        raise RuntimeError("already initialized")
    parent = self.parent
    collected_frames = 0
    while collected_frames < num_iter:
            tensordict = parent.rollout(max_iter = num_iter)
            collected_frames += tensordict.numel()
            loc = tensordict.get(key).mean(reduce_dim)
            scale = tensordict.get(key).std(reduce_dim)

    if self.standard_normal:
        # set as is as transform is (x-loc)/scale
    else:
        # transform to match x*scale - loc
    self.register_buffer("loc", loc)
    self.register_buffer("scale", scale)
    
def _apply_transform(...):
    if self.loc is None:
        raise RuntimeError("transform not initialized")
```

This should be properly tested in `test_transform.py`

romainjln:
## Description

Implement method to compute automatically the stats of an env with an ObservNorm transform

## Motivation and Context

close #565

## Types of changes

## Checklist


torchrl/envs/transforms/transforms.py
```diff
@@ -1279,18 +1279,62 @@ def __init__(
...
+        self.register_buffer("scale", scale)
+
+    def init_stats(
+        self,
+        num_iter: int,
+        reduce_dim: Union[int, Tuple[int]] = 0,
+        key: Optional[str] = None,
+    ) -> None:
+        if self.loc is not None or self.scale is not None:
+            raise RuntimeError(
```
vmoens, line 1303: we could test that we reach this error message by adding one more `init_stats` in the tests just after we call it:
```python
t_env.transform.init_stats(num_iter=11)
with pytest.raises(RuntimeError, match="already initialized"):
    t_env.transform.init_stats(num_iter=11)
```

vmoens (approved):
LGTM, just couple of minor comments

PR #674, opened 2022-11-14
[Feature] Jumanji envs

yingchenlin:
## Description
- Built spec for Jumanji state.
- Convert data between Torch RL TensorDict and Jumaji state.
- Add support for seeding.
- Allow multi-dimensional batch_size.


torchrl/envs/libs/jumanji.py
```diff
@@ -0,0 +1,430 @@
...
+            **{key: _data_to_spec_transform(value) for key, value in state.items()}
+        )
+    else:
+        raise TypeError(f"Unsupported state type {type(state)}")
+
+
+def _jumanji_to_torchrl_data_transform(val, device, batch_size):
+    if isinstance(val, (jax.Array, np.ndarray)):
+        if isinstance(val, jax.Array):
+            val = np.array(val)
```
vmoens, line 104: can't we use as array instead to avoid data copy?

torchrl/envs/libs/jumanji.py
```diff
@@ -0,0 +1,430 @@
...
+    elif isinstance(val, tuple) and hasattr(val, "_fields"):  # named tuples
+        return TensorDict(
+            {
+                k: _jumanji_to_torchrl_data_transform(v, device, batch_size)
+                for k, v in zip(val._fields, val)
+            },
+            device=device,
+            batch_size=batch_size,
+        )
+    elif hasattr(val, "__dict__"):
```
vmoens, line 121: why this one?

torchrl/envs/libs/jumanji.py
```diff
@@ -0,0 +1,430 @@
...
+                for k, v in val.__dict__.items()
+            },
+            device=device,
+            batch_size=batch_size,
+        )
+    else:
+        raise TypeError(f"Unsupported data type {type(val)}")
+
+
+def _torchrl_to_jumanji_state_transform(tensordict: TensorDict, env):
```
vmoens, line 134: can't we use the reverse function of the one above?

.circleci/config.yml
```diff
@@ -402,6 +402,51 @@ jobs:
...
+      - run:
+          name: Install torchrl
+          command: docker run -t --gpus all -v $PWD:$PWD -w $PWD -e UPLOAD_CHANNEL -e CU_VERSION "${image_name}" .circleci/unittest/linux_libs/scripts_jumanji/install.sh
+      - run:
+          name: Run tests
+          command: bash .circleci/unittest/linux_optdeps/scripts/run_test.sh
+      - run:
+          name: Codecov upload
+          command: |
+            bash <(curl -s https://codecov.io/bash) -Z -F linux-outdeps-gpu
```
vmoens, line 443: `linux-outdeps-gpu` should be `linux-jumanji` I guess

vmoens (commented):
Just a couple of comments

torchrl/envs/libs/jumanji.py
yingchenlin: it's because jax array can't be converted to tensor directly. it will cause exception.

torchrl/envs/libs/jumanji.py
yingchenlin: should be using `dataclasses.is_dataclass` and `dataclasses.fields` instead of `__dict__`.

torchrl/envs/libs/jumanji.py
yingchenlin: right, this will be refactored.

.circleci/config.yml
yingchenlin: i will change it

vmoens (approved):
Great stuff! A huge plus for the library!

PR #680, opened 2022-11-15
[Feature] A2C objective class and train example

albertbou92:
## Description

Added an A2C objective class.

I also created the helper functions necessary to run an A2C example, including _make_a2c_loss_, _A2CLossConfig_, _make_a2c_model_, _A2CModelConfig_

Creating a _make_a2c_model_ helper function was not strictly necessary since the models are the same as in PPO. However, I wanted to use less nodes in the hidden layers so I decided to create a _make_a2c_model_ instead of modifying the _make_ppo_model_. The methods can probably be merged in the future if necessary, and the architecture of the networks can be passed as a parameter.

Finally, I played a bit with the parameters int he _canfig.yaml_ file until I found a good enough configuration that learned pretty well in the HalfCheetah-v4 environment.

## Motivation and Context

There is an open issue about A2C, and while it is similar to REINFORCE and PPO which are already in the repo, the objective is not the same. In particular, it has the entropy term (which is not present in REINFORCE) and it does not have the log prob ratio weighting term, the clipping and the KL term present in PPO. 

## Types of changes

## Checklist


vmoens (commented):
Would you mind merging main and trying to solve the issues with the new "next" logic? Let me also know what you think of it :)

albertbou92:
Done! I brought all the changes from main, and now the training script calculates the initial Stats with the key "observation_vector" instead of "next_observation_vector". It should be the same since it is actually the same tensor delayed by 1 timestep. The example script runs without issues.

vmoens:
It feels like you merged each diff independently, which makes a gigantic diff here (over 50 files changed)
Can you run `git merge main` and see what happens?

torchrl/objectives/a2c.py
```diff
@@ -0,0 +1,148 @@
...
+        log_prob = dist.log_prob(action)
+        log_prob = log_prob.unsqueeze(-1)
+        return log_prob, dist
+
+    def loss_critic(self, tensordict: TensorDictBase) -> torch.Tensor:
+        if self.advantage_diff_key in tensordict.keys():
+            advantage_diff = tensordict.get(self.advantage_diff_key)
+            if not advantage_diff.requires_grad:
+                raise RuntimeError(
+                    "value_target retrieved from tensordict does not requires grad."
```
vmoens, line 109: I guess this is a copy paste so it's a self blame comment
"Requires" should read "require"

vmoens (commented):
LGTM overall. The lint test is failing, a pre-commit should solve that. We should consider adding this to the example test pipeline (#687). After that I think we'll be good to go!

torchrl/trainers/helpers/models.py
```diff
@@ -414,6 +410,295 @@ def make_ddpg_actor(
...
+                "A2C learnt from pixels require the shared_mapping to be set to True."
+            )
+        if cfg.lstm:
+            policy_net = LSTMNet(
+                out_features=out_features,
+                lstm_kwargs={"input_size": 64, "hidden_size": 64},
+                mlp_kwargs={"num_cells": [64, 64], "out_features": 64},
+            )
+            in_keys_actor += ["hidden0", "hidden1"]
+            out_keys += ["hidden0", "hidden1", "next_hidden0", "next_hidden1"]
```
vmoens, line 631: I think the last two keys should be `("next", "hidden0")` and `("next", "hidden1")` instead

vmoens (approved):
Just a couple of minor changes and we're good to go!

examples/a2c/config.yaml
```diff
@@ -0,0 +1,41 @@
+# Environment
+env_library: gym  # env_library used for the simulated environment.
+env_name: HalfCheetah - v4  # name of the environment to be created. Default=Humanoid-v2
```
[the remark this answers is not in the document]
albertbou92: yes, sorry it was due to the formatting with ufmt

PR #682, opened 2022-11-16
[Feature] add `standard_normal` for RewardScaling

### LINKED ISSUE #646: [Feature Request] implement a `standard_normal` keyword arg for RewardScaling transform
## Motivation

ObservationNorm transform can be called with standard normal (`(x-loc)/scale`) or regular (`x*scale+loc`). For consistency, we should implement this option for both.

I would keep it `False` by default for now (or default it to true and rename the transform to `RewardNorm`).

adityagandhamal:
## Description

Implement `standard_normal` kwarg for RewardScaling 

## Motivation and Context

Fixes #646 
Brings consistency to RewardScaling wrt ObservationNorm by implementing the `standard_normal` kwarg with the existing regular transform

## Types of changes

## Checklist

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -1312,7 +1312,7 @@ class TestTransforms:
         reward_scaling(td)
         for key in keys_total:
             if standard_normal:
-                assert (td.get(key) == -loc / scale).all()
+                assert (td.get(key) == ((td_copy.get(key) - loc) / scale)).all()
             else:
                 assert (td.get(key) == td_copy.get(key).mul_(scale).add_(loc)).all()
         assert (td.get("dont touch") == td_copy.get("dont touch")).all()
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -1317,11 +1317,7 @@ class TestTransforms:
                 assert (td.get(key) == td_copy.get(key).mul_(scale).add_(loc)).all()
         assert (td.get("dont touch") == td_copy.get("dont touch")).all()
 
-        if len(keys_total) == 0:
-            assert (
-                td.get("reward") == td_copy.get("reward").mul_(scale).add_(loc)
-            ).all()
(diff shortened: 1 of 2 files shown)


torchrl/envs/transforms/transforms.py
```diff
@@ -1522,8 +1531,13 @@ def __init__(
...
 
     def _apply_transform(self, reward: torch.Tensor) -> torch.Tensor:
-        reward.mul_(self.scale).add_(self.loc)
-        return reward
+        if self.standard_normal:
+            loc = self.loc
+            scale = self.scale
+            return (reward - loc) / scale
+        else:
+            reward.mul_(self.scale).add_(self.loc)
```
vmoens, line 1539: I know we were doing it like this before, but I think that in general we should avoid in-place changes (say someone has a reward computed through a model and wants to differentiate through it, things might break).
What about 
```python
reward = rewad * scale + loc
```

test/test_transforms.py
```diff
@@ -1308,13 +1311,13 @@ def test_reward_scaling(self, batch, scale, loc, keys, device):
         td_copy = td.clone()
         reward_scaling(td)
         for key in keys_total:
-            assert (td.get(key) == td_copy.get(key).mul_(scale).add_(loc)).all()
+            if standard_normal:
+                assert (td.get(key) == ((td_copy.get(key) - loc) / scale)).all()
```
vmoens, line 1315: maybe `torch.testing.assert_close(a, b)` instead? to avoid numerical precision errors

vmoens (commented):
Looks great! Can you address those two comments?

test/test_transforms.py
adityagandhamal: Got it!

vmoens (approved):
LGTM thanks for this!

PR #686, opened 2022-11-18
[BugFix] ConvNet forward method with tensors of more than 4 dimensions

albertbou92:
## Description

Collectors and BatchSubSample produce batches with shape (Batch, Length, Channels, Height, Width), but shapes with more than 4 dimensions raise an error when passed through a CNN. 

I just added an adjustment so that for batches with 5 dimensions, then Batch and Length dimension are collapsed before the forward pass and expanded right after it.

## Motivation and Context

I was trying to train on an Atari 2600 environment and the forward pass raised the following error:

> RuntimeError: Expected 3D (unbatched) or 4D (batched) input to conv2d, but got input of size: [8, 32, 4, 84, 84]

## Types of changes

## Checklist

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -146,7 +146,7 @@ def test_convnet(
 ):
     torch.manual_seed(seed)
     batch = 2
-    seq_length = 2
+    seq_len = 2
     convnet = ConvNet(
         in_features=in_features,
         depth=depth,
@@ -166,9 +166,10 @@ def test_convnet(
     )
     if in_features is None:
         in_features = 5
-    x = torch.randn(batch, seq_length, in_features, input_size, input_size, device=device)
+    x = torch.randn(batch, seq_len, in_features, input_size, input_size, device=device)
     y = convnet(x)
-    assert y.shape == torch.Size([batch, seq_length, expected_features])
+    assert y.shape == torch.Size([batch, seq_len, expected_features])
+
 
 @pytest.mark.parametrize(
     "layer_class",
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -146,6 +146,7 @@ def test_convnet(
 ):
     torch.manual_seed(seed)
     batch = 2
+    seq_length = 2
     convnet = ConvNet(
         in_features=in_features,
         depth=depth,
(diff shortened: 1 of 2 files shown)


torchrl/modules/models/models.py
```diff
@@ -463,6 +463,21 @@ def _make_net(self, device: Optional[DEVICE_TYPING]) -> nn.Module:
             layers.append(Squeeze2dLayer())
         return layers
 
+    def forward(self, inputs: torch.Tensor) -> torch.Tensor:
+
+        if len(inputs.shape) == 5:
```
vmoens, line 468: Here's how I would approach this:
```python
*batch, C, L, W = input.shape
if len(batch) > 1:
    inputs = inputs.flatten(0, len(batch))
out = super(ConvNet, self).forward(inputs)
if len(batch) > 1:
    out = out.unflatten(0, batch)
```

vmoens (commented):
Have a look at my comment.
Also we should integrate that feature in the tests of ConvNet.

torchrl/modules/models/models.py
albertbou92: Makes sense this is more general. Just changed it.

Only changed this line, apparently otherwise also colapses the len(batch) dimension which is the number of channels.
`inputs = inputs.flatten(0, len(batch) - 1)
`

test/test_modules.py
```diff
@@ -165,10 +166,9 @@ def test_convnet(
     )
     if in_features is None:
         in_features = 5
-    x = torch.randn(batch, in_features, input_size, input_size, device=device)
+    x = torch.randn(batch, seq_length, in_features, input_size, input_size, device=device)
```
vmoens, line 169: Here's how i would do it
we can add a 
```python
@pytest.mark.parametrize("batch", [tuple(), (2,), (2, 3, 2)])
def test_convnet(..., batch):
    ...
    x = torch.randn(*batch, in_features, input_size, input_size, device=device)
```

like that we test (no batch - batch of dim 1 - batch of dim > 1)

vmoens:
Don't forget to run the lint
```
pre-commit run --all-files
```
See [this doc](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md#formatting-your-code)

albertbou92:
Sorry. Now I did run pre-commit and all tests worked.

vmoens:
Cool! Let's just fix the tests (see comment above) and we're good to go

vmoens (approved):
LGTM!

PR #690, opened 2022-11-21
Prototype transformed replay buffers using extra class.

altre:
## Description

This is a first suggestion showcasing a path to a simple solution for #612 .
Two of the newly added tests still fail (Will provide a seperate issue for this, I think this is a separate problem.)

## Types of changes

 New feature (non-breaking change which adds core functionality)

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.
If you are unsure about any of these, don't hesitate to ask. We are here to help!

- [x] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
- [x] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.

PR #695, opened 2022-11-21
[Feature] Append, init and insert transforms in ReplayBuffer

### LINKED ISSUE #612: [Feature Request] Allow composable replay buffers to accept transforms
## Motivation

We should be able to add transforms to our composable replay buffers.
Adding and removing transforms should be done either:
- at instantiation time (`__init__(self, ..., transfrom):`)
- using `append_transform` as it is done in [`TransformedEnv`](https://github.com/pytorch/rl/blob/61b80f8dbb42de2e58e2725dea93e86f624169e1/torchrl/envs/transforms/transforms.py#L457-L470)
- using `insert_transform` as it is done in [`TransformedEnv`](https://github.com/pytorch/rl/blob/61b80f8dbb42de2e58e2725dea93e86f624169e1/torchrl/envs/transforms/transforms.py#L472-L484)

The transformation of the output should occur after sampling (in `replaybuffer.sample`).

All the `transform_*_spec` should return an error.

Registering parents in a transform should work with a replay buffer too, as it does with envs.

The following use cases should also be considered (and tested):
- appending a transform when none has been created at first
- inserting a transform with a wrongful index (e.g. insert transform at index 5 when there are only 2 transforms) => raise an error?
- Appending a transform when there is only one (i.e. create a `Compose` transform with the first and second transform).

### LINKED ISSUE #692: Cannot create compose on SqueezeTransform, UnsqueezeTransform and FlattenTransform
Some transforms assume to already have a parent when calling set_parent(), in which case creating a new Compose with them fails:

```
>>> Compose(t)
  File "<stdin>", line 1, in <module>
  File "/Users/alanschelten/rl/torchrl/envs/transforms/transforms.py", line 630, in __init__
    t.set_parent(self)
  File "/Users/alanschelten/rl/torchrl/envs/transforms/transforms.py", line 1115, in set_parent
    batch_size = parent.batch_size
AttributeError: 'NoneType' object has no attribute 'batch_size'
>>> t = UnsqueezeTransform(0)
>>> Compose(t)
  File "<stdin>", line 1, in <module>
  File "/Users/alanschelten/rl/torchrl/envs/transforms/transforms.py", line 630, in __init__
    t.set_parent(self)
  File "/Users/alanschelten/rl/torchrl/envs/transforms/transforms.py", line 1115, in set_parent
    batch_size = parent.batch_size
AttributeError: 'NoneType' object has no attribute 'batch_size'
>>> t = FlattenObservation()
>>> Compose(t)
  File "<stdin>", line 1, in <module>
  File "/Users/alanschelten/rl/torchrl/envs/transforms/transforms.py", line 630, in __init__
    t.set_parent(self)
  File "/Users/alanschelten/rl/torchrl/envs/transforms/transforms.py", line 1045, in set_parent
    observation_spec = self.parent.observation_spec
  File "/Users/alanschelten/rl/torchrl/envs/transforms/transforms.py", line 241, in parent
    raise ValueError(
ValueError: Compose parent was of type <class 'NoneType'> but expected TransformedEnv.
```

altre:
## Description

Implemented passing transforms in init, appending and inserting transforms as methods in ReplayBuffer
closes #612 and closes #692. 

Skips any code in transforms that accesses the parent without it being defined.

I think this approach is preferable to: https://github.com/pytorch/rl/pull/690 

Inserting at an out of bounds index throws an error as implemented in Compose. Note that this is not the same behavior as a native python list, which just inserts at the beginning or end of list and doesn't throw.

I'm not quite sure what is meant by transform_*_spec returning an error, since these methods are defined as public on transform subclasses and are not called in this implementation.

We could prevent the use of them by subclassing Compose, but I don't think it's necessary?

## Types of changes

## Checklist


altre:
The remaining failure: FAILED test/test_libs.py::TestCollectorLib::test_collector_run[device1-GymEnv-env_args1-env_kwargs1] - EOFError (https://app.circleci.com/pipelines/github/pytorch/rl/4652/workflows/88023457-c917-479c-bcff-9ebd04e46af6/jobs/114173?invite=true#step-108-3532) does not seem to be related to my changes, I think?

vmoens:
> The remaining failure: FAILED test/test_libs.py::TestCollectorLib::test_collector_run[device1-GymEnv-env_args1-env_kwargs1] - EOFError (https://app.circleci.com/pipelines/github/pytorch/rl/4652/workflows/88023457-c917-479c-bcff-9ebd04e46af6/jobs/114173?invite=true#step-108-3532) does not seem to be related to my changes, I think?

Nope those seem to be flacky tests

test/test_rb.py
```diff
@@ -594,6 +611,118 @@ def test_legacy_rb_does_not_attach():
...
+    pytest.param(partial(RewardScaling, loc=1, scale=2), id="RewardScaling"),
+    DoubleToFloat,
+    VecNorm,
+]
+
+
+@pytest.mark.parametrize("transform", transforms)
+def test_smoke_replay_buffer_transform(transform):
+    rb = rb_prototype.ReplayBuffer(
+        collate_fn=lambda x: torch.stack(x, 0),
```
vmoens, line 698: we should be able to remove the collate_fn as of #688

torchrl/data/replay_buffers/rb_prototype.py
```diff
@@ -30,6 +31,8 @@ class ReplayBuffer:
             samples.
         prefetch (int, optional): number of next batches to be prefetched
             using multithreading.
+        transform (Transform, optional): Transform to be executed when sample() is called.
+            To chain transforms use Compose.
```
vmoens, line 35: maybe 
```
  ... use the :obj:`Compose` class.
```

vmoens (commented):
Let's keep the errors as they were, #697 will implement more informative error messages.

test/test_rb.py
altre: Ok, merged upstream.

test/test_rb.py
```diff
@@ -594,6 +611,118 @@ def test_legacy_rb_does_not_attach():
...
+        transform=transform(in_keys="observation"),
+    )
+
+    td = TensorDict({"observation": torch.randn(3, 3, 3, 16, 1)}, [])
+    rb.add(td)
+    rb.sample(1)
```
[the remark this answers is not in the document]
altre: Well, yes, actually on purpose, I didn't wan't to write tests for each transform, just check if there are any crashes.. I suppose we can ensure it always generally does get called with mock?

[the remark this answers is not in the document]
vmoens: That's ok! I see the actual behaviour is tested later on so it's ok

vmoens (approved):
LGTM!

PR #698, opened 2022-11-22
[BugFix] Fix docs build

tcbegley:
## Description

Docs build was failing because `git-subtree` is not installed in the runner.

Rather than installing , we just use a clone instead and clean up the nested Git repo.


vmoens (approved):
LGTM

PR #700, opened 2022-11-22
[Feature] Migrate to `tensordict.nn.TensorDictModule`

tcbegley:
## Description

This PR adopts `tensordict.nn.TensorDictModule` and its variants and eliminates duplicated code. We don't use `TensorDictModule` as a drop-in replacement since `tensordict` library has no concept of TensorSpecs. Instead we inherit and let the base classes do the heavy lifting, while the corresponding classes in TorchRL largely add spec validation.


tcbegley:
I believe the tests are failing because they require the changes in pytorch-labs/tensordict#66. Not sure what the best way to test this is? Since afaik currently TorchRL does not import anything from `tensordict.nn`, perhaps that PR can be merged first and then these tests, which pull directly from tensordict main, should pass?

torchrl/modules/tensordict_module/common.py
```diff
@@ -88,7 +80,7 @@ def _forward_hook_safe_action(module, tensordict_in, tensordict_out):
                 )
 
 
-class TensorDictModule(nn.Module):
+class TensorDictModule(_TensorDictModule):
```
tcbegley, line 83: I was wondering whether we should consider changing the name here, as it could be confusing if both tensordict and TorchRL have `TensorDictModule` classes. 

I don't have a good name, but I'm thinking something like `TensorDictModuleWithSpec` to convey the difference?

vmoens: Yes that's something I thought about too
Maybe we can drop the TensorDict part (it's implicitly the case since torchrl is all-in into adopting tensordict)
`ModuleWithSpec`? `SpecModule`? `SafeModule`? `RLModule`?

tcbegley: Dropping TensorDict part sounds good to me.

Of your suggestions I'm most drawn to `SafeModule` and `SpecModule` in that order I think?

vmoens: Sorry for the slow answer
We can also think about user experience, and for that RLModule seems good
In ReAgent they have a (improperly named) ModelBase class that all models inherit from. It's not only bad because Model-Based in a subfield in RL, but also because it fails to account for what it is (the base class of all policies and value networks).
SafeModule is also good as the main difference is mainly that putting a spec allows for safe sampling / mapping (it's more about safety than about the spec IMO). In that sense RLModule is less self-explanatory.
TLDR let's go for SafeModule :p

torchrl/modules/tensordict_module/sequence.py
```diff
@@ -235,273 +143,11 @@ def select_subsequence(
...
-            )
-        if tensordict_out is not None:
-            tensordict_out.update(tensordict, inplace=True)
-            return tensordict_out
-        return tensordict
-
-    def __len__(self):
-        return len(self.module)
+        td_sequential = super().select_subsequence(in_keys=in_keys, out_keys=out_keys)
+        return TensorDictSequential(*td_sequential.module)
```
vmoens: I guess the only difference is that it returns an object of the same type as the parent?
Maybe we could modify the original `select_subsequence` in tensordict?

vmoens (approved):
Great stuff! A lot less code: the best code is no code at all

torchrl/modules/tensordict_module/common.py
```diff
@@ -80,8 +80,8 @@ def _forward_hook_safe_action(module, tensordict_in, tensordict_out):
                 )
 
 
-class TensorDictModule(_TensorDictModule):
-    """A TensorDictModule, is a python wrapper around a :obj:`nn.Module` that reads and writes to a TensorDict.
+class SafeModule(TensorDictModule):
+    """A SafeModule, is a python wrapper around a :obj:`nn.Module` that reads and writes to a TensorDict.
```
vmoens, line 84: `An :obj:``SafeModule`` is a :obj:``tensordict.nn.TensorDictModule`` subclass that accepts a :obj:``TensorSpec`` as argument to control the output domain.`

torchrl/modules/tensordict_module/probabilistic.py
```diff
@@ -5,29 +5,27 @@
...
 
 from torchrl.data import TensorSpec
 from torchrl.modules.distributions import Delta
-from torchrl.modules.tensordict_module.common import TensorDictModule
+from torchrl.modules.tensordict_module.common import SafeModule
 
 
-class ProbabilisticTensorDictModule(_ProbabilisticTensorDictModule, TensorDictModule):
+class SafeProbabilisticModule(ProbabilisticTensorDictModule, SafeModule):
     """A probabilistic TD Module.
```
vmoens, line 16: `A :obj:``SafeProbabilisticModule`` is an :obj:``tensordict.nn.ProbabilisticTensorDictModule`` subclass that accepts a :obj:``TensorSpec`` as argument to control the output domain.`

vmoens (approved):
A couple of suggestions for the docstrings
It's ok that things are copied for the rest i guess

torchrl/modules/tensordict_module/sequence.py
tcbegley: Yes you're right.

I think if we change `tensordict.nn.TensorDictSequential.select_subsequence` to return `self.__class__(...)` then we can get rid of this method here.

vmoens (approved):
Cool!

PR #709, opened 2022-11-25
[BugFix] RoundRobinWriter, possible duplicated code in the extend method

albertbou92:
## Description

In the class _RoundRobinWriter_, in the _extend_ method, seems like some of the conditions could be unnecessary. Maybe I am missing something, but if 

`cur_size = self._cursor
`

seems like the last two conditions are duplicated code. The code raises no error, but maybe the extra conditions can be removed.

## Types of changes

## Checklist

### DIFF
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -65,7 +65,7 @@ _has_tv = importlib.util.find_spec("torchvision") is not None
 )
 @pytest.mark.parametrize("writer", [writers.RoundRobinWriter])
 @pytest.mark.parametrize("storage", [ListStorage, LazyTensorStorage, LazyMemmapStorage])
-@pytest.mark.parametrize("size", [3, 100])
+@pytest.mark.parametrize("size", [3, 5, 100])
 class TestPrototypeBuffers:
     def _get_rb(self, rb_type, size, sampler, writer, storage):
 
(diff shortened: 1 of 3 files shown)


vmoens (commented):
LGTM! The implementation was odd, I agree.
Could you add a test in `TestPrototypeBuffers` to check that the cursor moves to the right position in the following cases:

1. Fill the replay buffer with less data than the storage max size (N < M)
2. Fill the replay buffer with as much data as storage max size (N=M), and add some more data in a second time with less data (N<M) => the cursor should be at N
3. Fill the replay buffer directly with more data than the storage max size (N > M), the cursor should be at N-M

vmoens (approved):
LGTM! 
Thanks agains for spotting this and solving it!

albertbou92:
I think the extend method int he ReplayBuffer from replay_buffers.py also can be simplified. I added the same changes I suggested for RoundRobinWriter and also added tests.

vmoens:
> I think the extend method in the ReplayBuffer from replay_buffers.py also can be simplified. I added the same changes I suggested for RoundRobinWriter and also added tests.

Cool thanks for that too!
Let's wait for the tests to pass and we're good to go

PR #715, opened 2022-11-25
[Feature] Use ObservationNorm.init_stats for stats computation in example scripts

### LINKED ISSUE #699: [Feature Request] Use ObservationNorm.init_stats for stats computation in example scripts
## Motivation

We currently use a custom function [`get_stats_random_rollout`](https://github.com/pytorch/rl/blob/b21776f87618777c4b4a3400802a2c7d7a4a6c77/examples/ddpg/ddpg.py#L117-L123) for stats computation in the examples. 
We should use `ObservationNorm.init_stats` instead.

We have 2 options: 
- Rewrite `get_stats_random_rollout` from within to just call the transform method
- Replace every call to `get_stats_random_rollout()` by a call to `ObservationNorm.init_stats`.

The first solution may be more readable but adds one level of coding that the second does not have, and perhaps has little advantages except for conciseness of the training script.

romainjln:
## Description

Refactor examples to use `ObservationNorm.init_stats` instead of `get_stats_random_rollout`

## Motivation and Context

close #699

## Types of changes

## Checklist


vmoens (commented):
I think that here, if we spawn multiple processes, each env on each process will run its own `init_env_steps ` and hence they will all have a different set of summary stats.

We should compute the stats in the main process like we did before (but using the method you provided), then pass these stats to each env.

torchrl/trainers/helpers/envs.py
```diff
@@ -428,6 +429,86 @@ def get_stats_random_rollout(
...
+    if proof_env_is_none:
+        proof_environment = transformed_env_constructor(
+            cfg=cfg, use_env_creator=False
+        )()
+
+    if not hasattr(cfg, "init_env_steps"):
+        raise AttributeError("init_env_steps missing from arguments.")
+
+    if (
+        type(proof_environment.transform) != Compose
```
vmoens, line 464: Why not isinstance?
Why equality and not `is not`?

torchrl/trainers/helpers/envs.py
```diff
@@ -428,6 +429,86 @@ def get_stats_random_rollout(
...
+        and type(proof_environment.transform) != ObservationNorm
```
vmoens, line 465: Same here

torchrl/trainers/helpers/envs.py
```diff
@@ -428,6 +429,86 @@ def get_stats_random_rollout(
...
+        keys = list(proof_environment.observation_spec.keys())
+        key = keys.pop()
+        if len(keys):
+            raise RuntimeError(
+                f"More than one key exists in the observation_specs: {[key] + keys} were found, "
+                "thus generate_stats_from_observation_norm cannot infer which to compute the stats of."
+            )
+
+    obs_norm_transforms = []
+    if type(proof_environment.transform) == Compose:
```
vmoens, line 482: Same

torchrl/trainers/helpers/envs.py
```diff
@@ -428,6 +429,86 @@ def get_stats_random_rollout(
...
+                type(transform) == ObservationNorm
+                and transform.loc is None
+                and transform.scale is None
+            ):
+                obs_norm_transforms.append((idx, transform))
+    else:
+        obs_norm_transforms.append((0, proof_environment.transform))
+
+    stats = []
+    for (idx, transform) in obs_norm_transforms:
```
vmoens, line 494: Upon reflection we could simply take the state dict of the transforms and load it no?
Wouldn't that be simpler than this?
If loc and scale are buffers it could simplify things a bit

vmoens (commented):
LGTM 
Do you think things would be simpler with state dict?
We could 
(1) create a dummy env, compute stats
(2) get the state dict of this dummy env
(3) load the state dict on every env created subsequently

LMK what you think

torchrl/trainers/helpers/envs.py
```diff
@@ -428,6 +446,78 @@ def get_stats_random_rollout(
...
+                transform.init_stats(num_iter=num_iter, key=key)
+    elif (
+        proof_environment.transform.loc is None
+        and proof_environment.transform.scale is None
+    ):
+        proof_environment.transform.init_stats(num_iter=num_iter, key=key)
+
+
+def retrieve_observation_norms_state_dict(proof_environment: TransformedEnv):
+    """Traverse the transforms of the environment and retrieve the ObservationNorm state dicts.
```
vmoens, line 500: Maybe Traverses + retrieves?

Also for code objects we can use :obj:`ObservationNorm` for a cleaner rendering in the doc

torchrl/trainers/helpers/envs.py
```diff
@@ -428,6 +446,78 @@ def get_stats_random_rollout(
     return stats
 
 
+def initialize_observation_norm_transforms(
+    proof_environment: EnvBase,
+    num_iter: int = 1000,
+    key: Union[str, Tuple[str, ...]] = None,
+):
+    """Calling init_stats on all uninitialised ObservationNorms transform of a TransformedEnv.
```
vmoens, line 454: Maybe `Calls :obj:`ObservationNorm.init_stats` on all uninitialized :obj:`ObservationNorm` instances of a :obj:`TransformedEnv`.

vmoens (approved):
LGTM!

Issue #704, opened 2022-11-23
[Feature Request] Create an optimizer hook for the trainer


vmoens:
## Motivation

Optimizers are currently passed as a single object to the optimizer. In some cases, we'd like to have multiple different optimizers for a problem (for instance to update the parameters only every N operations).

Optimizers should be implemented using a hook (for instance `_optimizer_hook` placed in between `_post_loss_hook` and `_post_optim_hook`, and that would receive the loss as input) that would call each optimizer.

Since the loss is a TensorDict of losses, we could even contemplate the possibility of having a backward for each optim hook, and hence each optim hook would do its own backward, optim step and zero_grad over its set of parameters.

We should keep the optimizer in the args of the `Trainer.__init__` method as it's convenient for the users to pass it there, but `optimizer=None` should be supported (then the `_optimizer_hook` would not see anything and would just be a no-op).

Doc: https://pytorch.org/rl/reference/trainers.html#torchrl-trainers-package

vmoens:
Solved by #716

PR #716, opened 2022-11-26
[Feature] Add OptimizerHook

aakhundov:
## Description

The new type of hook, `OptimizerHook`, is added as suggested in #704.

## Motivation and Context

See the suggested change in #704.

## Types of changes

- [x] Documentation (**initial** update in the documentation)

Backward compatibility with the existing `optimizer` parameter of `Trainer.__init__` is respected.
Minor breaking change: previously `grad_norm` key in the `losses_td` is now `grad_norm_0`.

## Checklist

The change made in the documentation is very initial: `optimizer` hook is just briefly mentioned in the list of possible hooks and the `OptimizerHook` class is added to the "Trainer and hooks" table. Probably, wider documentation coverage of the new hook type is needed.

### DIFF
--- a/docs/source/reference/trainers.rst
+++ b/docs/source/reference/trainers.rst
@@ -29,9 +29,9 @@ The :obj:`trainer.train()` method can be sketched as follows:
         ...     self._post_steps_hook()  # "post_steps"
         ...     self._post_steps_log_hook(batch)  #  "post_steps_log"
 
-There are 9 hooks that can be used in a trainer loop: :obj:`"batch_process"`, :obj:`"pre_optim_steps"`,
+There are 10 hooks that can be used in a trainer loop: :obj:`"batch_process"`, :obj:`"pre_optim_steps"`,
 :obj:`"process_optim_batch"`, :obj:`"post_loss"`, :obj:`"post_steps"`, :obj:`"post_optim"`, :obj:`"pre_steps_log"`,
-:obj:`"post_steps_log"` and :obj:`"post_optim_log"`. They are indicated in the comments where they are applied.
+:obj:`"post_steps_log"`, :obj:`"post_optim_log"` and :obj:`"optimizer"`. They are indicated in the comments where they are applied.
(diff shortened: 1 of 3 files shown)


aakhundov:
@vmoens As currently implemented, the new hook's signature is different from other hooks: besides accepting (and returning) a `TensorDict` instance, it also accepts the clip norm parameters and the index `i` in the `trainer._optimizer_ops` list (for distinguishing each hook's `grad_norm_{i}` key added to the returned `TensorDict`).

If we want the optimizer hook to be more similar to other hooks, we could:

1. Add `clip_norm` and `clip_norm_grad` constructor parameters to the `OptimizerHook.__init__`. Then, when creating the "default" optimizer hook from its `optimizer` argument, `Trainer.__init__` will just pass its own `clip_norm` and `clip_norm_grad` arguments as is to the `OptimizerHook.__init__`. Bonus: users will be able to set different gradient clipping configuration for each hook separately.

2. Return `grad_norm` value from the `OptimizerHook.__call__` instead of the updated `TensorDict` parameter. Then in `Trainer._optimizer_hook` we could set the `grad_norm` in the `losses_td` with the respective index, without the need to pass the index to the hook call.

3. Move detaching the `losses_td` outside of `Trainer._optimizer_hook`: to be done by the calling `Trainer.optim_steps` right after the `Trainer._optimizer_hook` returns.

Importantly, even after the above modifications, `Trainer._optimizer_hook` will be different from other `Trainer._*_hook` functions, as it will need to set `grad_norm` in its `TensorDict` argument besides the usual for-loop over the hook instances.

Please let me know how you'd like to proceed. Thanks!

vmoens (approved):
Great stuff!
Thanks a lot for this!

vmoens:
Regarding this

> Importantly, even after the above modifications, Trainer._optimizer_hook will be different from other Trainer._*_hook functions, as it will need to set grad_norm in its TensorDict argument besides the usual for-loop over the hook instances.

It's ok, it's backward compatible and it suits most needs.
Thanks again for this!

PR #719, opened 2022-11-30
[Refactor] Sequential probabilistic changes

tcbegley:
## Description

pytorch-labs/tensordict#88 deprecates the `get_dist` and `get_dist_params` methods of `TensorDictSequential` and hence `torchrl.modules.SafeSequential` which subclasses it. This PR adapts code and tests in TorchRL for that change.

This supercedes an earlier PR https://github.com/tcbegley/rl/pull/5 (rebased onto new commits + targeting TorchRL main rather than a branch on my fork).

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -2921,6 +2921,8 @@ def test_shared_params(dest, expected_dtype, expected_device):
         loss.qvalue_network_params.items(include_nested=True, leaves_only=True)
     ):
         p1 = value
+        if isinstance(loss.actor_network, SafeProbabilisticModule):
+            key = ("module", *key) if isinstance(key, tuple) else ("module", key)
         p2 = loss.actor_network_params[key]
         assert (p1 == p2).all()
         if i == 1:
@@ -2943,6 +2945,8 @@ def test_shared_params(dest, expected_dtype, expected_device):
     for i, (key, qvalparam) in enumerate(
         loss.qvalue_network_params.items(include_nested=True, leaves_only=True)
     ):
+        if isinstance(loss.actor_network, SafeProbabilisticModule):
+            key = ("module", *key) if isinstance(key, tuple) else ("module", key)
         assert qvalparam.dtype is expected_dtype, (key, qvalparam)
         assert qvalparam.device == torch.device(expected_device), key
         assert (qvalparam == loss.actor_network_params[key]).all(), key
--- a/torchrl/modules/tensordict_module/actors.py
+++ b/torchrl/modules/tensordict_module/actors.py
@@ -650,8 +650,22 @@ class ActorValueOperator(SafeSequential):
             value_operator,
         )
 
-    def get_policy_operator(self) -> SafeSequential:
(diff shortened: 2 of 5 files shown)


torchrl/modules/tensordict_module/actors.py
```diff
@@ -650,8 +650,26 @@ def __init__(
...
+                distribution_class=self.module[1].distribution_class,
+                distribution_kwargs=self.module[1].distribution_kwargs,
+                return_log_prob=self.module[1].return_log_prob,
+                cache_dist=self.module[1].cache_dist,
+                n_empirical_estimate=self.module[1].n_empirical_estimate,
+                **kwargs,
+            )
         return SafeSequential(self.module[0], self.module[1])
 
     def get_value_operator(self) -> SafeSequential:
```
tcbegley, line 675: Could `self.module[2]` here be a `SafeProbabilisticModule` and if so, does that mean we need to make a similar change to `get_value_operator`?

test/test_cost.py
```diff
@@ -2921,6 +2921,8 @@ def __init__(self, actor_network, qvalue_network):
         loss.qvalue_network_params.items(include_nested=True, leaves_only=True)
     ):
         p1 = value
+        if isinstance(loss.actor_network, SafeProbabilisticModule):
+            key = ("module", *key) if isinstance(key, tuple) else ("module", key)
```
tcbegley, line 2925: This feels a bit messy, the actor network parameters have a different structure to the q-value network because they've be wrapped with a `SafeProbabilisticModule`. 

I think that extra layer is avoidable, but maybe we could wrap the value network with a `SafeProbabilisticModule` too using `Delta` distributions in order that the parameter structure remains in sync?

torchrl/objectives/redq.py
```diff
@@ -191,16 +190,9 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
-            if isinstance(self.actor_network, TensorDictSequential):
-                sample_key = self.actor_network[-1].sample_out_key[0]
-                tensordict_actor_dist = self.actor_network[-1].build_dist_from_params(
-                    td_params
-                )
-            else:
-                sample_key = self.actor_network.sample_out_key[0]
-                tensordict_actor_dist = self.actor_network.build_dist_from_params(
-                    td_params
-                )
```
tcbegley, line 203: I think under these changes the actor network will never be sequential (it will always be wrapped with a probabilistic module), so this should be safe to delete. Does that seem right?

tcbegley:
@vmoens comment from https://github.com/tcbegley/rl/pull/5 for reference

> It feels like we cleaned up one module but made added a lot of complexity elsewhere.
> What do you think?
> Maybe the the ProbabilisticModule should be constructed above the Sequential in all cases?
> 
> e.g. ActorCritic related objects would not care about it, since they would be wrapped by a ProbabilisticModule anyway.

tcbegley:
Superceded by #728 + #738

PR #725, opened 2022-12-01
[Tests] Omit tests and examples from codecov checks

tcbegley:
## Description

Currently codecov checks are being applied to test and example files. This PR configures `coverage` to only check the test coverage inside `torchrl` itself.

I originally noticed this while working on #719.


tcbegley:
Checking test coverage of tests is actually useful for catching systematic skips.

PR #728, opened 2022-12-07
[Refactor] Adopt prototype ProbabilisticTensorDictModule and ProbabilisticTensorDictSequential

tcbegley:
## Description

This PR adopts the prototype refactored `ProbabilisticTensorDictModule` and `ProbabilisticTensorDictSequential` classes introduced in pytorch-labs/tensordict#104

Currently WIP while I work through the failing tests, but new usage should be clear. Most important changes are in:
- `torchrl/modules/tensordict_module/probabilistic.py` - refactors `SafeProbabilisticModule` and introduces the new `SafeProbabilisticSequential` class
- `torchrl/modules/tensordict_module/actors.py` - note the change to `ActorValueOperator` in particular

Most other changes are just updating usage.

This PR supercedes #719, though I will leave that open for now until it's clear we want to move ahead with the approach laid out in this PR and pytorch-labs/tensordict#104

TODO:
- [x] Update tutorials

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,24 +270,23 @@ The associated [`SafeModule` class](torchrl/modules/tensordict_module/common.py)
     )
     # Wrap the policy module in NormalParamsWrapper, such that the output
     # tensor is split in loc and scale, and scale is mapped onto a positive space
-    policy_module = NormalParamsWrapper(
-        MLP(
-            num_cells=[64, 64],
-            out_features=32,
-            activation=nn.ELU,
-        )
-    )
-    # Wrap the nn.Module in a SafeProbabilisticModule, indicating how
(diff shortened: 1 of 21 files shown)


vmoens:
Regarding the failing tests, I tried to put this after the creation of the ddpg model:
```python
    proc = mp.Process(target=dummyfun, args=(X,))
    proc.start()
    proc.join()
```
where 
```python
def dummyfun(model):
    print(model)
```

It fails when `X=model[0]` but runs when `X=model[0][0]` or `X=model[0][1]`
Hence the problem seem to be related to `ProbabilisticActor`. I'll keep investigating

vmoens (approved):
Neat! thanks for this

PR #734, opened 2022-12-08
[Feature] Multithreaded env

### LINKED ISSUE #591: [Feature Request] Multithreaded env
## Motivation

We have a version of multiprocessed envs, but not multithreaded.
It would be nice to explore how fast these can be, and/or if that would be a viable solution for executing a list of environments simultaneously.

sgrigory:
## Description

Allows creating multithreaded environments based on [EnvPool](https://envpool.readthedocs.io/en/latest/index.html) library. This is an alternative to `ParallelEnv`, which uses multiprocessing parallelism. EnvPool-based approach is faster, but less flexible, as one can only create environments implemented in EnvPool (whereas `ParallelEnv` parallelises arbitrary TorchRL environment)

Changes:
- Add `MultiThreadedEnv` environment to `torchrl/envs/vec_env.py`
- Add benchmarking script to `benchmarks/benchmark_batched_envs.py`
- Add an option for `CompositeSpec` to return only non-nested keys - otherwise `step` didn't work for environments with multiple observations, which have nested observation spec (like [Cheetah](https://github.com/deepmind/dm_control/blob/main/dm_control/suite/cheetah.py) which returns position and velocity)
- Add tests

In this PR I only added [synchronous EnvPool interface](https://envpool.readthedocs.io/en/latest/content/python_interface.html#batch-size), when batch size is equal to the number of workers, but in principle it's possible to add the asynchronous one too.
## Motivation and Context
Closes #591

## Benchmark 

The following benchmark runs `env.rollout(max_steps=1000)` for three types of batched environments: `SerialEnv`, `ParallelEnv`, and `MultiThreadedEnv`:
```
pip install pandas
python benchmarks/benchmark_batched_envs.py
```

  | num_workers_1_cpu | num_workers_4_cpu | num_workers_16_cpu | num_workers_1_cuda | num_workers_4_cuda | num_workers_16_cuda
-- | -- | -- | -- | -- | -- | --
Serial, s | 0.217 | 0.433 | 1.357 | 0.319 | 0.645 | 1.924
Parallel, s | 0.206 | 0.234 | 0.275 | 0.296 | 0.502 | 1.428
Multithreaded, s | 0.13 | 0.153 | 0.161 | 0.187 | 0.211 | 0.242
Gain, % | 36.9 | 34.6 | 41.5 | 36.8 | 58.0 | 83.1

One can see that on CPU `MultiThreadedEnv` is ~40% faster than `ParallelEnv`, depending on the number of workers. Neither of the models is benefiting from GPU in the current setup, or I'm not using it in the right way.
Environment creation time is excluded for all benchmarks.

## Tests

Tests has been added to `tests/test_libs.py::TestEnvPool` and run in a separate pipeline `unittest_linux_envpool_gpu` located in `.cicrleci/unittest/linux_libs/scripts_envpool`.  Many of the tests are the same as for `SerialEnv` and `ParallelEnv`

#### Comparing `_BatchedEnv` and `MultiThreadedEnv` output: different RNGs
One test which is lacking is a check that results from `SerialEnv`/`ParallelEnv` and `MultiThreadedEnv` are the same for the same environment configuration and the seed. This is because they generate random numbers in a different way:
- EnvPool uses C random number generator from `std::uniform_real_distribution` ([here](https://github.com/sail-sg/envpool/blob/494a409db093bd389a5b25d6d709c52d96d1e366/envpool/classic_control/pendulum.h#L55) and [here](
https://github.com/sail-sg/envpool/blob/29d84126955fcd3729d6c94455e89b44791b07dc/envpool/core/env.h#L80))
- `SerialEnv`/`ParallelEnv` with Gym envs take random numbers from Gym, which uses `numpy` RNG ([here](https://github.com/openai/gym/blob/6a04d49722724677610e36c1f92908e72f51da0c/gym/utils/seeding.py) and [here](https://github.com/openai/gym/blob/6a04d49722724677610e36c1f92908e72f51da0c/gym/core.py#L151))

I googled around, but didn't find a good way to use the same random number generator on both sides:
- https://studywolf.wordpress.com/2012/10/02/a-common-random-number-generator-for-c-and-python/
- https://stackoverflow.com/questions/65871948/same-random-numbers-in-c-as-computed-by-python3-numpy-random-rand

## Types of changes

## Checklist


.circleci/unittest/linux/scripts/setup_env.sh
```diff
@@ -86,3 +86,7 @@ fi
 pip install pip --upgrade
 
 conda env update --file "${this_dir}/environment.yml" --prune
+
+if [[ $os == 'Linux' ]]; then
+  pip install envpool
```
vmoens, line 91: For lib such as this one we generally do a single test in a dedicated pipeline of the file test_libs
Have a look at the jumanji pipeline for instance

torchrl/envs/vec_env.py
```diff
@@ -1068,3 +1082,302 @@ def _run_worker_pipe_shared_mem(
...
+        "batch_size",
+        "_dummy_env_str",
+    ]
+
+    def __init__(
+        self,
+        num_workers: int,
+        task_id: str,
+        env_type: str,
+        batch_size: int = None,
```
shagunsodhani, line 1110: Optional[int]

torchrl/envs/vec_env.py
```diff
@@ -1068,3 +1082,302 @@ def _run_worker_pipe_shared_mem(
...
+        create_env_kwargs: Union[dict, Sequence[dict]] = None,
```
shagunsodhani, line 1111: Optional[Union...]

torchrl/envs/vec_env.py
```diff
@@ -1068,3 +1082,302 @@ def _run_worker_pipe_shared_mem(
...
+        return self._batch_size
+
+    @property
+    def device(self) -> torch.device:
+        if self._device is None:
+            self._set_properties()
+        return self._device
+
+    @device.setter
+    def device(self, value: DEVICE_TYPING) -> None:
```
shagunsodhani, line 1196: should this return self ?

sgrigory:
@vmoens @shagunsodhani Feel free to have a look!

torchrl/envs/vec_env.py
```diff
@@ -1076,3 +1089,257 @@ def _run_worker_pipe_shared_mem(
...
+    _verbose: bool = False
+
+    def __init__(
+        self,
+        env: Optional["envpool.python.envpool.EnvPoolMixin"] = None,
+        **kwargs,
+    ):
+        if not _has_envpool:
+            raise ImportError(
+                f"""envpool python package was not found. Please install this dependency.
```
shagunsodhani, line 1106: nitpicking but this error could happen both because `envpool` is missing or because `treevalue` is missing. Maybe we should mention both in the error

torchrl/envs/vec_env.py
```diff
@@ -1076,3 +1089,257 @@ def _run_worker_pipe_shared_mem(
...
+            gym_reset_return_info=True,
+            **create_env_kwargs,
+        )
+        return super()._build_env(env)
+
+    def _set_seed(self, seed: Optional[int]):
+        """Library EnvPool only support setting a seed by recreating the environment."""
+        if seed is not None:
+            self.create_env_kwargs["seed"] = seed
+            self._env = self._build_env(
```
shagunsodhani, line 1333: I am curious if this is an expensive operation and if we should highlight this behavior via logging (debug level)

shagunsodhani (commented):
lgtm

torchrl/envs/vec_env.py
sgrigory: Good point, adjusted the error message

torchrl/envs/vec_env.py
sgrigory: Added a `logging.debug` message

.circleci/unittest/linux_libs/scripts_envpool/environment.yml
```diff
@@ -0,0 +1,24 @@
...
+  - pytorch
+  - defaults
+dependencies:
+  - pip
+  - protobuf
+  - pip:
+    - hypothesis
+    - future
+    - cloudpickle
+    - gym
```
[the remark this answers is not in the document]
sgrigory: Actually I'm testing locally with `gym==0.24.0` and CI uses `gym-0.26.2`, so we might be covered here. I had to set `gym_reset_return_info=True` in `envpool.make` for this to work

torchrl/envs/vec_env.py
```diff
@@ -1076,3 +1092,297 @@ def _run_worker_pipe_shared_mem(
...
+        reset_data = self._env.reset(reset_workers)
+        tensordict_out = self._transform_reset_output(reset_data, reset_workers)
+        self.is_closed = False
+        return tensordict_out
+
+    def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
+        action = tensordict.get("action")
+        # Action needs to be moved to CPU and converted to numpy before being passed to envpool
+        action = action.to(torch.device("cpu"))
+        step_output = self._env.step(action.detach().numpy())
```
shagunsodhani, line 1157: I think this `detach` is redundant since the tensor is already on CPU. Not a big deal though

shagunsodhani (approved):
lgtm

torchrl/envs/vec_env.py
```diff
@@ -1076,3 +1092,297 @@ def _run_worker_pipe_shared_mem(
...
+
+    _verbose: bool = False
+
+    def __init__(
+        self,
+        env: Optional["envpool.python.envpool.EnvPoolMixin"] = None,
+        **kwargs,
+    ):
+        if not _has_envpool:
+            raise ImportError(
```
vmoens, line 1108: Another way is to keep the error in the try/except:
```
IMPORT_ERR_ENVPOOL = err
```

and then do `ImportError(...) from IMPORT_ERR_ENVPOOL`

test/test_libs.py
```diff
@@ -449,6 +458,325 @@ def test_jumanji_consistency(self, envname, batch_size):
...
+            transformed_out=transformed_out,
+            device=device,
+            N=N,
+        )
+
+        assert env_multithreaded.device == torch.device(device)
+        out = env_multithreaded.rollout(max_steps=20)
+        assert out.device == torch.device(device)
+
+        env_multithreaded.close()
```
vmoens, line 777: We should add a simple `check_env_specs(env)` in a separate test. This will assess if the specs are ok.

vmoens (approved):
LGTM! Great work
Can you have a look a the couple of comments I left?

vmoens:
cc @albertbou92 @SebastianAment @Benjamin-eecs

test/test_libs.py
sgrigory: Added a separate test `test_specs`. It works for all envs except for `CheetahRun-v1`. For that env, it seems that 
the observations returned by envpool are inconsistent with the bounding box it returns in spec: the bounding box values is strictly positive, but observations can be both positive and negative. I think I'll submit this to issues in the envpool repo.

```
>>> import envpool
>>> env = envpool.make("CheetahRun-v1", env_type="gym")
>>> env.spec.observation_space
Dict(position:Box([2.22507386e-308 2.22507386e-308 2.22507386e-308 2.22507386e-308
 2.22507386e-308 2.22507386e-308 2.22507386e-308 2.22507386e-308], [1.79769313e+308 1.79769313e+308 1.79769313e+308 1.79769313e+308
 1.79769313e+308 1.79769313e+308 1.79769313e+308 1.79769313e+308], (8,), float64), velocity:Box([2.22507386e-308 2.22507386e-308 2.22507386e-308 2.22507386e-308
 2.22507386e-308 2.22507386e-308 2.22507386e-308 2.22507386e-308
 2.22507386e-308], [1.79769313e+308 1.79769313e+308 1.79769313e+308 1.79769313e+308
 1.79769313e+308 1.79769313e+308 1.79769313e+308 1.79769313e+308
 1.79769313e+308], (9,), float64))
>>> env.reset()
<TreeValue 0x7fa1cf2be8e0>
├── 'position' --> array([[-0.09139574,  0.02779732, -0.06476798, -0.01931019, -0.09387728,
│                           0.00776585, -0.00777442,  0.00666921]])
└── 'velocity' --> array([[ 0.00531753, -0.00294308,  0.00210252,  0.0133823 ,  0.01259834,
                            0.01001254, -0.00010664, -0.01476834, -0.01361733]])
```

torchrl/envs/vec_env.py
sgrigory: Makes sense, done

vmoens:
I think your benchmarks could be improved for parallel env
Starting the workers take a tremendous amount of time, sometimes > 30 secs, but once it's started it should be faster.
The first few rollouts collected should be discarded from the benchmark

torchrl/envs/vec_env.py
sgrigory: Removed `.detach()`, added the decorator. One of them is needed, otherwise it throws "_RuntimeError: Can't call numpy() on Tensor that requires grad. Use tensor.detach().numpy() instead_" even on CPU

sgrigory:
> I think your benchmarks could be improved for parallel env Starting the workers take a tremendous amount of time, sometimes > 30 secs, but once it's started it should be faster. The first few rollouts collected should be discarded from the benchmark

Indeed, that was the case - after I excluded environment creation from the benchmarked time, `ParallelEnv` became faster than `SerialEnv`. Updated the benchmarking script and the PR description

vmoens (approved):
LGTM

PR #746, opened 2022-12-15
[Refactor] Cosmetic improvements to advantage modules

vmoens:
## Description

Some cosmetic improvements to GAE, TDL and TD.
Gathers the in_keys and out_keys automatically, and allows to use the modules without tensordict.

### DIFF
--- a/torchrl/objectives/value/advantages.py
+++ b/torchrl/objectives/value/advantages.py
@@ -3,6 +3,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+from functools import wraps
 from typing import List, Optional, Union
 
 import torch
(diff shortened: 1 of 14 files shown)


torchrl/objectives/value/advantages.py
```diff
@@ -73,22 +88,64 @@ def is_functional(self):
...
+            >>> reward = torch.randn(1, 10, 1)
+            >>> done = torch.zeros(1, 10, 1, dtype=torch.bool)
+            >>> tensordict = TensorDict({"obs": obs, "next": {"obs": next_obs}, "done": done, "reward": reward}, [1, 10])
+            >>> _ = module(tensordict)
+            >>> assert "advantage" in tensordict.keys()
+
+        The module supports non-tensordict (i.e. unpacked tensordict) inputs too:
+
+        Examples:
+            >>> value_net = SafeModule(
```
vmoens, line 137: cc @tcbegley this is something a bit more interesting than TensorDictModule
This module is a regular nn.Module that reads from a tensordict.
You may have a some crazy combination of keys that your value network needs to read, but you'd still be able to use regular tensors with this.

PR #751, opened 2022-12-16
[Feature] RewardSum transform

albertbou92:
## Description

Adds a new Transform class, called RewardSum. Which tracks the cumulative reward of all episodes in progress and adds the information to the tensordict as a new key.

## Motivation and Context

It can be informative to be able to access the training episode rewards. 

e.g. it can be used like this to track the performance during training
```
for batch in collector:
    train_episode_reward = batch["episode_reward"][batch["done"]]
    if batch["episode_reward"][batch["done"]].numel() > 0:
           print(f"train_episode_rewards {train_episode_reward.mean()}")
```

## Types of changes

## Checklist


torchrl/envs/transforms/transforms.py
```diff
@@ -2433,3 +2433,51 @@ def __repr__(self) -> str:
...
+        out_keys: Optional[Sequence[str]] = None,
+    ):
+        in_keys = ["reward"]
+        if out_keys is None:
+            out_keys = ["episode_reward"]
+        super().__init__(in_keys=in_keys, out_keys=out_keys)
+
+    def reset(self, tensordict: TensorDictBase) -> TensorDictBase:
+        """Resets episode rewards."""
+        if "reset_workers" in tensordict.keys():
```
vmoens, line 2459: what happens if we apply reset to an env that is not batched? I don't see the `"epsode_reward"` being reset in that case

torchrl/envs/transforms/transforms.py
```diff
@@ -2433,3 +2433,51 @@ def __repr__(self) -> str:
...
+                    tensordict[out_key][tensordict["reset_workers"]] = 0.0
+
+        return tensordict
+
+    def _call(self, tensordict: TensorDictBase) -> TensorDictBase:
+        """Updates the episode rewards with the step rewards."""
+        # Sanity checks
+        self._check_inplace()
+        for in_key in self.in_keys:
+            if in_key not in tensordict.keys():
```
vmoens, line 2471: what if we have 2 keys (e.g. "reward" and "reward2") and only "reward2" is present?

vmoens: If there can only be one "in_keys" we could check that in the constructor? Then we can just do
```
in_key = self.in_keys[0]
out_key = self.out_keys[0]
```
and avoid relying on for loops in the transform

vmoens (commented):
Sorry for the late review, I was on PTO

Overall LGTM! I like that you code everything in the input-output tensordict. It is not well documented but we should discourage setting attributed to the transform to track call-to-call carry on variables, the tensordict API is made for that.  

Also, because we change the output tensordict by adding a new key, it would be nice to add the out-keys to the `observation_spec` or change the `reward_spec` to a `CompositeSpec` with `"reward"` and `"episode_reward"` keys.
WDYT?

Beyond this PR:
We should drop the inplace attribute of the transforms. In general inplace is a recipe for bugs and PyTorch developers usually recommend not using it. The perf gain is usually marginal and it significantly narrows the number of applications

torchrl/envs/transforms/transforms.py
albertbou92: Now it handles more that one input key (e.g. "reward" and "reward2") and creates outputs based on these names ("episode_reward" and "episode_reward2")

torchrl/envs/transforms/transforms.py
albertbou92: Modified the code to include the non-batched case

albertbou92:
> Sorry for the late review, I was on PTO
> 
> Overall LGTM! I like that you code everything in the input-output tensordict. It is not well documented but we should discourage setting attributed to the transform to track call-to-call carry on variables, the tensordict API is made for that.
> 
> Also, because we change the output tensordict by adding a new key, it would be nice to add the out-keys to the `observation_spec` or change the `reward_spec` to a `CompositeSpec` with `"reward"` and `"episode_reward"` keys. WDYT?
> 
> Beyond this PR: We should drop the inplace attribute of the transforms. In general inplace is a recipe for bugs and PyTorch developers usually recommend not using it. The perf gain is usually marginal and it significantly narrows the number of applications

albertbou92:
> Sorry for the late review, I was on PTO
> 
> Overall LGTM! I like that you code everything in the input-output tensordict. It is not well documented but we should discourage setting attributed to the transform to track call-to-call carry on variables, the tensordict API is made for that.
> 
> Also, because we change the output tensordict by adding a new key, it would be nice to add the out-keys to the `observation_spec` or change the `reward_spec` to a `CompositeSpec` with `"reward"` and `"episode_reward"` keys. WDYT?
> 
> Beyond this PR: We should drop the inplace attribute of the transforms. In general inplace is a recipe for bugs and PyTorch developers usually recommend not using it. The perf gain is usually marginal and it significantly narrows the number of applications

Not sure if I would consider  `episode_reward ` to be a part of the  `observation` or the  `reward` to add them to the spec. In any case I tried to add the key to the reward spec by creating a `CompositeSpec`  and ran into some problems. Should we do it?

torchrl/envs/transforms/transforms.py
```diff
@@ -2433,3 +2433,90 @@ def __repr__(self) -> str:
...
+        #         device=reward_spec.device,
+        #         dtype=reward_spec.dtype
+        #     )
+        #
+        # # Add episode_reward spec to observation spec
+        # reward_spec = CompositeSpec(
+        #     reward=reward_spec,
+        #     episode_reward=episode_reward_spec
+        # )
+        return reward_spec
```
albertbou92, line 2522: I tried to modify the reward_spec, but that generated problem somewhere else in the code.

In any case, is episode_reward part of the reward?

albertbou92:
> > Not sure if I would consider `episode_reward ` to be a part of the `observation` or the `reward` to add them to the spec. In any case I tried to add the key to the reward spec by creating a `CompositeSpec` and ran into some problems. Should we do it?
> 
> It's probably better to put it in the observation_spec. The reason we need it is that the specs are used to create "fake_data" that will be used as buffers in parallel settings. If a key of the env output is not in the env spec, you risk to miss it and an error will be raised when creating that env in parallel. If you want to check that the specs match the data from a rollout, you can use `torchrl.envs.utils.check_env_specs`.

Added the transform_observation_spec method and some code to test it. 
Now the transform accepts a set of in_keys that should be present in the reward specs (This checked when calling transform_observation_spec). If no in_keys are provided it is assumed that the key is simply reward. Also if reward_spec is not a ComposeSpec it is assumed that the only in_key is reward.

torchrl/envs/transforms/transforms.py
```diff
@@ -2433,3 +2433,121 @@ def __repr__(self) -> str:
...
+                    episode_spec = UnboundedContinuousTensorSpec(
+                        device=reward_spec.device,
+                        dtype=reward_spec.dtype,
+                    )
+                episode_specs.update({out_key: episode_spec})
+
+        else:
+
+            # If reward_spec is not a CompositeSpec, the only in_key should be ´reward´
+            assert set(self.in_keys) == {
```
vmoens, line 2530: We should not use asserts in any other place than tests
Here I think a KeyError would be appropriate

torchrl/envs/transforms/transforms.py
```diff
@@ -2433,3 +2433,121 @@ def __repr__(self) -> str:
...
+    def reset(self, tensordict: TensorDictBase) -> TensorDictBase:
+        """Resets episode rewards."""
+        # Non-batched environments
+        if len(tensordict.batch_size) < 1 or tensordict.batch_size[0] == 1:
+            for out_key in self.out_keys:
+                if out_key in tensordict.keys():
+                    tensordict[out_key] = 0.0
+
+        # Batched environments
+        elif "reset_workers" in tensordict.keys():
```
vmoens, line 2473: I guess that the default in the batched cased where no "reset_workers" is present would be to reset them all no?
otherwise in multi-agent settings for instance, a user could do

```
env.reset()
```
an end up with an episode reward that keeps on piling up

torchrl/envs/transforms/transforms.py
```diff
@@ -2433,3 +2433,121 @@ def __repr__(self) -> str:
...
+        for in_key in self.in_keys:
+            if in_key not in tensordict.keys():
+                return tensordict
+
+        # Update episode rewards
+        for in_key, out_key in zip(self.in_keys, self.out_keys):
+            reward = tensordict.get(in_key)
+            if out_key not in tensordict.keys():
+                tensordict.set(
+                    out_key, torch.zeros(*tensordict.shape, 1, dtype=reward.dtype)
```
vmoens, line 2494: let's add `device=reward.device` here

torchrl/envs/transforms/transforms.py
albertbou92: you are right. Now if no "reset_workers" is found in the tensordict for the batched case, all envs are reset

torchrl/envs/transforms/transforms.py
albertbou92: done

torchrl/envs/transforms/transforms.py
albertbou92: ok!

albertbou92:
My code considered the possibility that reward spec was a NdUnboundedContinuousTensorSpec, but I see that class has been removed. Should I consider that reward spec can only be an UnboundedContinuousTensorSpec?

vmoens:
The Nd*Spec classes were confusing so we removed them in favour of their parent class.
Removing the Nd should be enough.

vmoens (commented):
We should also add this transform to the doc in the envs.rst file

vmoens (approved):
Looks like the tests are passing. The only thing missing is the doc in envs.rst

albertbou92:
I added it to the list of transforms in envs.rst, but I am not sure if that is enough

vmoens (approved):

PR #756, opened 2022-12-21
[Feature] Add Step Counter transform

riiswa:
## Description

Add a transform that counts the steps from a reset and sets the done state to True after a certain number of steps.

This transform can be used in parallel / multi env settings, as the counter has the size of the env.batch_size. The "step_counter" key will be added to the output of tensordict when reset or step in the environment.

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -2499,13 +2499,21 @@ class StepCounter(Transform):
     def reset(self, tensordict: TensorDictBase) -> TensorDictBase:
         workers = tensordict.get(
             "reset_workers",
-            default=torch.ones(*tensordict.batch_size, 1, dtype=torch.bool),
+            default=torch.ones(
+                *tensordict.batch_size, 1, dtype=torch.bool, device=tensordict.device
+            ),
         )
         tensordict.set(
             "step_count",
             (~workers)
             * tensordict.get(
-                "step_count", torch.zeros(*tensordict.batch_size, 1, dtype=torch.int64)
+                "step_count",
+                torch.zeros(
+                    *tensordict.batch_size,
+                    1,
+                    dtype=torch.int64,
+                    device=tensordict.device,
+                ),
             ),
         )
         if self.max_steps is not None and self.max_steps <= 0:
@@ -2515,7 +2523,13 @@ class StepCounter(Transform):
     def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
         next_step_count = (
             tensordict.get(
-                "step_count", torch.zeros(*tensordict.batch_size, 1, dtype=torch.int64)
+                "step_count",
+                torch.zeros(
+                    *tensordict.batch_size,
+                    1,
+                    dtype=torch.int64,
+                    device=tensordict.device,
+                ),
             )
             + 1
         )
@@ -2527,7 +2541,15 @@ class StepCounter(Transform):
             )
         return tensordict
 
-    @_apply_to_composite
-    def transform_observation_spec(self, observation_spec: TensorSpec) -> TensorSpec:
-        observation_spec["step_count"] = UnboundedDiscreteTensorSpec(dtype=torch.int64)
+    def transform_observation_spec(
+        self, observation_spec: CompositeSpec
+    ) -> CompositeSpec:
+        if not isinstance(observation_spec, CompositeSpec):
+            raise ValueError(
+                f"observation_spec was expected to be of type CompositeSpec. Got {type(observation_spec)} instead."
+            )
+        observation_spec["step_count"] = UnboundedDiscreteTensorSpec(
+            dtype=torch.int64, device=observation_spec.device
+        )
+        observation_spec["step_count"].space.minimum = 0
         return observation_spec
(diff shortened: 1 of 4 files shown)


riiswa:
@vmoens I'm not sure I fully understood how to edit the observation_spec, in any case here is my PR, I'm waiting for your changes suggestions :)

torchrl/envs/transforms/transforms.py
```diff
@@ -2429,3 +2429,50 @@ def __repr__(self) -> str:
...
+    def _transform_spec(self, spec: TensorSpec) -> None:
+        if isinstance(spec, CompositeSpec):
+            for key in spec:
+                self._transform_spec(spec[key])
+        else:
+            spec.dtype = torch.int64
+
+    @_apply_to_composite
+    def transform_observation_spec(self, observation_spec: TensorSpec) -> TensorSpec:
+        self._transform_spec(observation_spec)
```
vmoens, line 2477: Maybe i'm missing something, but I think you should do something like
```
    def transform_observation_spec(self, observation_spec: TensorSpec) -> TensorSpec:
        observation_spec["step_count"] = UnboundedDiscreteTensorSpec(dtype=torch.int64)
```

torchrl/envs/transforms/transforms.py
```diff
@@ -2429,3 +2429,50 @@ def __repr__(self) -> str:
...
+    """Counts the steps from a reset and sets the done state to True after a certain number of steps.
+
+    Args:
+        max_steps (:obj:`int`, optional): the maximum number of steps to take before setting the done state to True.
+    """
+
+    invertible = False
+    inplace = True
+
+    def __init__(self, max_steps: Optional[int]):
```
vmoens, line 2444: it is optional but it does not have a default value. Maybe None?

torchrl/envs/transforms/transforms.py
```diff
@@ -2429,3 +2429,50 @@ def __repr__(self) -> str:
             f"{self.__class__.__name__}(decay={self.decay:4.4f},"
             f"eps={self.eps:4.4f}, keys={self.in_keys})"
         )
+
+
+class StepCounter(Transform):
+    """Counts the steps from a reset and sets the done state to True after a certain number of steps.
+
+    Args:
+        max_steps (:obj:`int`, optional): the maximum number of steps to take before setting the done state to True.
```
vmoens, line 2438: + add what the default is, and if it is None, what behaviour is expected in that case.

torchrl/envs/transforms/transforms.py
```diff
@@ -2429,3 +2429,50 @@ def __repr__(self) -> str:
...
+
+    invertible = False
+    inplace = True
+
+    def __init__(self, max_steps: Optional[int]):
+        self.max_steps = max_steps
+        super().__init__([])
+
+    def reset(self, tensordict: TensorDictBase) -> TensorDictBase:
+        tensordict.set(
```
vmoens, line 2449: If we want to make this work in parallel settings, we can work with the "reset_workers" key, which indicates what are the workers that need to be reset

torchrl/envs/transforms/transforms.py
```diff
@@ -2429,3 +2429,50 @@ def __repr__(self) -> str:
...
+        )
+        if self.max_steps is not None and self.max_steps <= 0:
+            tensordict.fill_("done", True)
+        return tensordict
+
+    def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
+        next_step_count = tensordict.get("step_count") + 1
+        tensordict.set("step_count", next_step_count)
+        if self.max_steps is not None:
+            tensordict.set(
```
vmoens, line 2460: I think you could just do
```
tensordict["done"] = tensordict["done"] | next_step_count >= self.max_steps
```

vmoens (commented):
Good work!
Have a look at my comments

torchrl/envs/transforms/transforms.py
riiswa: Thanks for all your suggestions! I made the changes, but I would like to know how can I test the `transform_observation_spec` method

vmoens: There's a test function you can run
```
from torchrl.envs.utils import check_env_specs
env = TransformedEnv(GymEnv("Pendulum-v1"), StepCounter(50))
check_env_specs(env)
```

It will run a small rollout + get fake data from your specs, and check that they match.

riiswa: I get this error, I don't really know what I missed,  it comes from the shape of the tensor is []:

```python
test/test_transforms.py:1600 (TestTransforms.test_step_counter_observation_spec)
  File "/Users/waris/Projects/rl/test/test_transforms.py", line 1604, in test_step_counter_observation_spec
    check_env_specs(env)
  File "/Users/waris/Projects/rl/torchrl/envs/utils.py", line 168, in check_env_specs
    fake_tensordict = env.fake_tensordict().flatten_keys(".")
  File "/Users/waris/Projects/rl/torchrl/envs/common.py", line 636, in fake_tensordict
    fake_obs = observation_spec.rand(self.batch_size)
  File "/Users/waris/Projects/rl/torchrl/data/tensor_specs.py", line 1286, in rand
    _dict = {
  File "/Users/waris/Projects/rl/torchrl/data/tensor_specs.py", line 1287, in <dictcomp>
    key: self[key].rand(shape)
  File "/Users/waris/Projects/rl/torchrl/data/tensor_specs.py", line 655, in rand
    r = torch.rand(*shape, *interval.shape, device=interval.device)
TypeError: rand() missing 1 required positional arguments: "size"
```

vmoens: On line 655, try putting the shape in between brackets
Like `rand([*shape1, *shape2],...)`
Also make sure you've merged main into your branch!

vmoens (approved):
LGTM
This should be adapted to the Nd*Spec removal though :)

PR #765, opened 2022-12-24
[BugFix] SyncDataCollector init when device and env_device are different

albertbou92:
## Description

In the init method of the SyncDataCollector class, a small number of steps is taken with the policy to determine the relevant keys of the output TensorDict. When the policy device and the environment device are different, that can raise a RuntimeError since the input provided to the policy is located in the environment device.

This PR only makes sure that the TensorDict provided to the policy is in the policy device, and then moves the output TensorDict to the environment device again. 

## Types of changes

## Checklist

### DIFF
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -1000,6 +1000,7 @@ def test_collector_device_combinations(device, passing_device):
         env = make_make_env("conv")()
         env.set_seed(seed)
         return env
+
     policy = dummypolicy_conv()
 
     collector = SyncDataCollector(
@@ -1018,14 +1019,22 @@ def test_collector_device_combinations(device, passing_device):
     collector.shutdown()
 
     collector = MultiSyncDataCollector(
-        create_env_fn=[env_fn, ],
-        create_env_kwargs=[{"seed": 0}, ],
+        create_env_fn=[
+            env_fn,
+        ],
+        create_env_kwargs=[
+            {"seed": 0},
+        ],
         policy=policy,
         frames_per_batch=20,
         max_frames_per_traj=2000,
         total_frames=20000,
-        devices=[device, ],
-        passing_devices=[passing_device, ],
+        devices=[
+            device,
+        ],
+        passing_devices=[
+            passing_device,
+        ],
         pin_memory=False,
     )
     batch = next(collector.iterator())
@@ -1033,19 +1042,28 @@ def test_collector_device_combinations(device, passing_device):
     collector.shutdown()
 
     collector = MultiaSyncDataCollector(
-        create_env_fn=[env_fn, ],
-        create_env_kwargs=[{"seed": 0}, ],
+        create_env_fn=[
+            env_fn,
+        ],
+        create_env_kwargs=[
+            {"seed": 0},
+        ],
         policy=policy,
         frames_per_batch=20,
         max_frames_per_traj=2000,
         total_frames=20000,
-        devices=[device, ],
-        passing_devices=[passing_device, ],
+        devices=[
+            device,
+        ],
+        passing_devices=[
+            passing_device,
+        ],
         pin_memory=False,
     )
     batch = next(collector.iterator())
     assert batch.device == torch.device(passing_device) or batch["done"].device
     collector.shutdown()
+
 
 @pytest.mark.skipif(not _has_gym, reason="test designed with GymEnv")
 @pytest.mark.parametrize(
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -993,6 +993,60 @@ def test_collector_output_keys(collector_class, init_random_frames, explicit_spe
     del collector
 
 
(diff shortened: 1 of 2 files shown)


vmoens (commented):
Thanks for this! Can you add a test where device and passing_device differ to check non-regression over this bug fix?

albertbou92:
Added tests for all device/passing_device combinations for SyncDataCollector, MultiSyncDataCollector and MultiaSyncDataCollector. 

I noticed that for MultiSyncDataCollector, the output batch TensorDict has its device attribute set to None.
I think the problem can be in the TensorDict.cat method, which returns a TensorDict whose "next" TensorDict has device set to None. As a sanity check, I added a line of code to make sure the MultiSyncDataCollector output TensorDict is cast to the correct device.

vmoens:
@albertbou92 interesting
> I noticed that for MultiSyncDataCollector, the yielded TensorDicts had its device attribute set to None.

Does this happen when the passing devices match too? Or just when there's more than one?
`torch.cat` should conserve the device (since we're sure that all device match). If it doesn't it's a bug. When the devices don't match, we first cast things to cpu before calling cat. It's definitely not ultra fast so we may want to parametrise that feature.

> which returns a TensorDict whose "next" TensorDict has device set to None.
You mean that the tensordict has a device but the nested tensordict doesn't? Weird...

albertbou92:
> @albertbou92 interesting
> 
> > I noticed that for MultiSyncDataCollector, the yielded TensorDicts had its device attribute set to None.
> 
> Does this happen when the passing devices match too? Or just when there's more than one? `torch.cat` should conserve the device (since we're sure that all device match). If it doesn't it's a bug. When the devices don't match, we first cast things to cpu before calling cat. It's definitely not ultra fast so we may want to parametrise that feature.
> 
> > which returns a TensorDict whose "next" TensorDict has device set to None.
> > You mean that the tensordict has a device but the nested tensordict doesn't? Weird...

albertbou92:
> Does this happen when the passing devices match too?

Yes it happens for all device combinations

albertbou92:
> You mean that the tensordict has a device but the nested tensordict doesn't? Weird...

exactly

vmoens:
> > You mean that the tensordict has a device but the nested tensordict doesn't? Weird...
> 
> 
> 
> exactly

Ok then it's a tensordict bug, I'll have a look into it.

albertbou92:
Should I remove the line of code to make sure the MultiSyncDataCollector yielded TensorDicts are cast to the correct device.

We could keep it as a sanity check, won't add overhead if the TensorDicts is already in the correct device.

vmoens:
> Should I remove the line of code that makes sure the MultiSyncDataCollector yielded TensorDicts are cast to the correct device.
> 
> We could keep it as a sanity check, won't add overhead if the TensorDicts is already in the correct device.

Which line of code are we talking about?

By the way: can you check if the device bug with nested tensordicts is still present?

albertbou92:
the device bug with nested TensorDicts seems to be fixed now!
 
In my code I added to the MultiSyncDataCollector _iterator_ method a `out_buffer = out_buffer.to(prev_device)` after concatenating all different dicts to make sure the yielded TensorDict was places to the passing_device. Now that the cat method is fixed might be unnecessary. 
 
I can remove it or we can either keep as a sanity check.

vmoens:
If it's not necessary I would remove it. Those things take little time but they quickly pile up to a consequent overhead.

albertbou92:
ok, removed that checking line.

Now the PR just fixed the bug in  SyncDataCollector when policy device and passing_device are different and adds tests to make sure all collectors work with all device combinations.

test/test_collector.py
```diff
@@ -989,6 +989,83 @@ def test_collector_output_keys(collector_class, init_random_frames, explicit_spe
     del collector
 
 
+@pytest.mark.parametrize("device", ["cuda", "cpu"])
```
vmoens, line 992: You should probably use ˋget_available_devices` here as some tests run on machines that don't have a cuda device

test/test_collector.py
```diff
@@ -989,6 +989,83 @@ def test_collector_output_keys(collector_class, init_random_frames, explicit_spe
...
+@pytest.mark.parametrize("passing_device", ["cuda", "cpu"])
```
vmoens, line 993: Same as above

test/test_collector.py
```diff
@@ -989,6 +989,82 @@ def test_collector_output_keys(collector_class, init_random_frames, explicit_spe
     del collector
 
 
+@pytest.mark.parametrize(
```
vmoens, line 992: Does it make sense to test on cpu only? Maybe we can skip it in that case

albertbou92: ok yes I can remove it in that case

vmoens (approved):

PR #767, opened 2022-12-27
Minor cleaning in BaseEnv classes

matteobettini:
## Description

This cahnge adds annotations for abstract methods and fixes a repeated override

## Motivation and Context

Code cleaning

- [ ] I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

What types of changes does your code introduce? Remove all that do not apply:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds core functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)
- [ ] Example (update in the folder of examples)

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.
If you are unsure about any of these, don't hesitate to ask. We are here to help!

- [X] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] My change requires a change to the documentation.
- [ ] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.

PR #768, opened 2022-12-27
Revert "Minor cleaning in BaseEnv classes"

vmoens:
Reverts pytorch/rl#767


vmoens:
@matteobettini i merged the PR but had to revert it because the tests were failing. Sorry about that. If you could reopen the PR and address the few comments that would be awesome

PR #770, opened 2022-12-28
Making `_set_seed` abstract

matteobettini:
## Description

Making `_set_seed` abstract

## Motivation and Context

In the current implementation of `EnvBase` there is a convention for certain methods to have and abstract counterpart.

`step()` -> `_step()`
`reset()` -> `_reset()`

The same seems to be valid for `set_seed()`, except that `_set_seed()` is not officially abstract.

Following the convention it would be intuitive that inheritors of `EnvBase` would override `_set_seed()`, but there are currenty some inheritors which do not implement `_set_seed()`.  This breaks the intuitive convention.

In this PR we made `_set_seed()` abstract. This caused some inheritors which override `set_seed()` (like `_BatchedEnv` and `TransformedEnv`) to need to implement `_set_seed()` with no action (as these classes act as wrappers).

## Types of changes

## Checklist

I have checked that the tests in test_env.py which I am able to run (without external libraries and cuda) are all passing and linted.


torchrl/envs/transforms/transforms.py
```diff
@@ -453,10 +453,16 @@ def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
 
         return tensordict_out
 
-    def set_seed(self, seed: int, static_seed: bool = False) -> int:
-        """Set the seeds of the environment."""
+    def set_seed(
+        self, seed: Optional[int] = None, static_seed: bool = False
+    ) -> Optional[int]:
+        # This method is not used in transformed environments
```
vmoens, line 459: Is this true? The message matches the one from _set_seed

vmoens: Maybe we could use the docstring above?

Another note: I don't remember why we're not using _set_seed to be honest. I guess it's because the seed iteration would occur twice if `transformed_env._set_seed` would call ˋbase_env.set_seed`, but what if `transformed_env._set_seed` would call ˋbase_env._set_seed`?

torchrl/envs/vec_env.py
```diff
@@ -514,6 +514,10 @@ def close(self) -> None:
     def _shutdown_workers(self) -> None:
         raise NotImplementedError
 
+    def _set_seed(self, seed: Optional[int]):
+        # This method is not used in batched envs
+        return
```
vmoens, line 519: Maybe `pass` instead of `return`? Seems semantically more appropriate to me

vmoens (approved):
LGTM! Let's merge main in this branch to resolve the conflicts

torchrl/envs/transforms/transforms.py
matteobettini: > Is this true? The message matches the one from _set_seed

Oh sorry it was not meant to be present in `set_seed`

> Maybe we could use the docstring above?
> 
> Another note: I don't remember why we're not using _set_seed to be honest. I guess it's because the seed iteration would occur twice if `transformed_env._set_seed` would call ˋbase_env.set_seed`, but what if `transformed_env._set_seed` would call ˋbase_env._set_seed`?

Basically `_set_seed` is only called in the original implementation of `set_seed` in `BaseEnv`. Given that `TransformedEnv` and `BatchedEnv` override `set_seed` (this would not normally be the case for inheritors of `BaseEnv` and is the case just for these two examples as they are wrappers) the inner `_set_seed` is not called anymore and we just implement it with a pass (whatever we implement it with doesn't matter as it is never called)

torchrl/envs/vec_env.py
matteobettini: Yep will fix

matteobettini:
All the comments should have been addressed

Issue #773, opened 2022-12-31
[BUG] `BatchedEnv` crashes when `EnvBase` has non-empty `batch_size`


matteobettini:
## Describe the bug

In the function `get_metadata()` of `_BatchedEnv` the tensordict `expand` method is called in a wrong manner:
```
self.meta_data = get_env_metadata(
                create_env_fn[0], create_env_kwargs[0]
            ).expand(self.num_workers)
```

As per docs, `expand` should be called with an  argument which is the concatenation of the new dims to the old dims

## To Reproduce

Create a `ParallelEnv` from a `BaseEnv` with non-empy batch_size

PR #774, opened 2022-12-31
[BugFix] Batched environments with non empty batch size

matteobettini:
This PR fixes BUG #773

### DIFF
--- a/test/mocking_classes.py
+++ b/test/mocking_classes.py
@@ -7,6 +7,7 @@ from typing import Optional
 import torch
 import torch.nn as nn
 from tensordict.tensordict import TensorDict, TensorDictBase
+
 from torchrl.data.tensor_specs import (
     BinaryDiscreteTensorSpec,
     BoundedTensorSpec,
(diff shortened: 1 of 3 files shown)


vmoens (approved):
LGTM thanks!

vmoens:
@matteobettini can I ask you to add a simple test in test_envs.py to test non regression of this feature?

matteobettini:
I have added the test.

The test tests only the batch size of the `ParallelEnv` compised of envs with batch sizes.

If you for example try to go forward and run a step, you will encounter a series of issues that stem from the fact that when `_BatchedEnv` was created, it was not taken into account that environments can have their own batch sizes. This is also the reason why no Brax environment works when wrapped in a BatchedEnv.

I will open an issue on this as serveral important modifications have to be made to allow a `ParallelEnv` of envs with their own batch sizes

vmoens:
The lint is failing, let's fix this and merge!

matteobettini:
Should be fixed

PR #778, opened 2022-12-31
Allow undounded boxes creation from gym spaces

matteobettini:
Previously only bounded tensors are created from gym specs.

In this PR we create unbounded tensor specs if all the high and low values of the gym Box are inf


vmoens (approved):
LGTM thanks!

PR #782, opened 2023-01-01
[Doc] Doc revamp

vmoens:
## Description

Solves some missing references and enriches the knowledge base.


docs/source/reference/envs.rst
```diff
@@ -195,27 +195,30 @@ in the environment. The keys to be included in this inverse transform are passed
 
     Transform
     TransformedEnv
-    RewardClipping
-    Resize
+    BinarizeReward
```
shagunsodhani, line 198: Curious how are these ordered now ?

vmoens: First the two classes that you must know about, then the rest in alphabetic order

Issue #781, opened 2023-01-01
[Feature Request] MultiDiscreteTensorSpec


matteobettini:
## Motivation

Currently there are the following avaialble specs for discrete spaces:

```
OneHotDiscreteTensorSpec # One hot for one discrete action
MultOneHotDiscreteTensorSpec # One hot for n actions
DiscreteTensorSpec # Categorical integer for one action
```

It would be nice to introduce the remaining missing spec `MultiDiscreteTensorSpec` for n categorical actions.

This would allow the `_gym_to_torchrl_spec_transform` to use this new spec with the parameter `categorical_action_encoding` to decide which space to assign to `gym.spaces.MultIDiscrete` doing somethig like this:

```
    elif isinstance(spec, gym.spaces.multi_discrete.MultiDiscrete):
        return    MultiDiscreteTensorSpec(spec.nvec, device=device)  if categorical_action_encoding
            else MultOneHotDiscreteTensorSpec(spec.nvec, device=device)
```
as it is already done for `gym.spaces.Discrete`.

Furthermore it would be nice for these discrete classes to possess the methods `to_onehot()` and `to_categorical()` respectively.

riiswa:
@matteobettini will you work on it? If not I would like to work on it 😄

matteobettini:
I have a few thing in my hands these days so I do not know when I could start on this if you want to do it go for it :)

PR #783, opened 2023-01-01
[Feature] MultiDiscreteTensorSpec

riiswa:
## Description

The goal of this PR is to add MultiDiscreteTensorSpec for n categorical actions (#781)

Example: 
```python
>>> ts = MultiDiscreteTensorSpec((3,2,3))
>>> ts.is_in(torch.tensor([2, 0, 1]))
True
>>> ts.is_in(torch.tensor([2, 2, 1]))
False
>>> ts.rand()
tensor([0, 1, 2])
>>> ts.rand(torch.Size((2, 2)))
tensor([[[0, 0, 1],
         [1, 0, 2]],

        [[2, 1, 1],
         [2, 1, 1]]])
```

This PR don't support yet `nvec` with several axes.

TODO (<s>still in progress</s> Ready for reviewing):
- [X] Implement MultiDiscreteTensorSpec
- [X] Support for Gym spaces
- [X] Write tests
- [X] Add `to_onehot()` and `to_categorical()`

## Motivation and Context

This will close the issue #781 

## Types of changes

## Checklist


torchrl/data/tensor_specs.py
```diff
@@ -981,6 +989,90 @@ def __eq__(self, other):
     def to_numpy(self, val: TensorDict, safe: bool = True) -> dict:
         return super().to_numpy(val, safe)
 
+    def to_onehot(self) -> OneHotDiscreteTensorSpec:
+        return OneHotDiscreteTensorSpec(self.space.n, self.device, self.dtype)
+
+
+@dataclass(repr=False)
+class MultDiscreteTensorSpec(DiscreteTensorSpec):
```
shagunsodhani, line 997: Nitpicking but probably `MultiDiscreteTensorSpec` seems better than `MultDiscreteTensorSpec`

torchrl/data/tensor_specs.py
```diff
@@ -981,6 +989,90 @@ def __eq__(self, other):
...
+                    space.n,
+                    torch.Size([1, *shape]),
+                    device=self.device,
+                    dtype=self.dtype,
+                )
+                for space in self.space
+            ]
+        ).squeeze()
+        _size = [self._size] if self._size > 1 else []
+        return x.T.reshape([*shape, *_size])
```
vmoens, line 1045: This leads to the following warning if the number of dims of `x` is greater than 2
```
<string>:3: UserWarning: The use of `x.T` on tensors of dimension other than 2 to reverse their shape is deprecated and it will throw an error in a future release. Consider `x.mT` to transpose batches of matricesor `x.permute(*torch.arange(x.ndim - 1, -1, -1))` to reverse the dimensions of a tensor. (Triggered internally at  /Users/distiller/project/conda/conda-bld/pytorch_1646756029501/work/aten/src/ATen/native/TensorShape.cpp:2318.)
```

torchrl/data/tensor_specs.py
```diff
@@ -981,6 +989,90 @@ def __eq__(self, other):
...
+        return torch.cat(
+            [
+                _val.clamp_(min=0, max=space.n - 1).unsqueeze(0)
+                for _val, space in zip(vals, self.space)
+            ],
+            dim=-1,
+        ).squeeze()
+
+    def is_in(self, val: torch.Tensor) -> bool:
+        vals = self._split(val)
```
vmoens, line 1063: we should also check the dtype here
(note to myself: we should check that `is_in` always checks the dtype)

vmoens (approved):
LGTM
I left some minor comments
@matteobettini curious to see if that suits your purpose

matteobettini:
> LGTM I left some minor comments Can you elaborate a bit more the description of the PR?
> 
> @matteobettini curious to see if that suits your purpose

Yep it seems to do what we want.

Maybe one little thig is. with the `to_categorical`and `to_one_hot()` i was meaning something like `to_categorical(tensor)`so that it can also to the value conversion. This for some reason is done inside the `to_numpy()` method currently. We could think of removing this conversion from the numpy methods into their own methods. PS i dont thin the numpy methods should do the conversions

````
spec = torchrl.data.MultOneHotDiscreteTensorSpec([5,10])
sample = spec.rand((2,3))
sample.shape
torch.Size([2, 3, 15])
sample_np =spec.to_numpy(sample)
sample_np.shape
(2, 3, 2)
````

vmoens:
You're right 
This came from the the time where all was one hot, and the default for other libs was categorical. Now that we support both we should keep them separated. 
@riiswa do you want to address that? I can find other people to do it if you'd like

riiswa:
I will work on it :) and I'm also working on multi-dimensional nvec (I forgot this case):

"""
# Gym Example
>> d = MultiDiscrete(np.array([[1, 2], [3, 4]]))
>> d.sample()
array([[0, 0],
       [2, 3]])
"""

Issue #775, opened 2022-12-31
[BUG] Using `sorted()` on the `keys()` attribute of a `CompositeSpec`


matteobettini:
## Describe the bug

In a number of places in the code, the `sorted()`  method is run on the `keys()` attribute of a `CompositeSpec`.

The bug was found when trying to create a `ParallelEnv` for Brax.

Here it fails as for nested specs the outputs of keys can be tuples.

```
env = ParallelEnv(1, lambda: BraxEnv(_get_envs()[0]))
env.rand_step()
```

The error is found is in 
```
  File "/Users/Matteo/PycharmProjects/torchrl/torchrl/envs/vec_env.py", line 426, in _create_td
    self.env_input_keys = sorted(self.input_spec.keys())
```
but I suspect it might not be the only occurrance

## To Reproduce

```python
import torchrl

env = ParallelEnv(1, lambda: BraxEnv(_get_envs()[0]))
env.rand_step()
```

```bash
Traceback (most recent call last):
  File "/Users/Matteo/PycharmProjects/VectorizedMultiAgentSimulator/vmas/examples/torch_rl.py", line 30, in <module>
    env.rand_step()
  File "/Users/Matteo/PycharmProjects/torchrl/torchrl/envs/common.py", line 529, in rand_step
    return self.step(tensordict)
  File "/Users/Matteo/PycharmProjects/torchrl/torchrl/envs/common.py", line 341, in step
    tensordict_out = self._step(tensordict)
  File "/Users/Matteo/PycharmProjects/torchrl/torchrl/envs/vec_env.py", line 32, in decorated_fun
    self._create_td()
  File "/Users/Matteo/PycharmProjects/torchrl/torchrl/envs/vec_env.py", line 426, in _create_td
    self.env_input_keys = sorted(self.input_spec.keys())
TypeError: '<' not supported between instances of 'tuple' and 'str'
```

PR #787, opened 2023-01-03
[BugFix] Sorting nested keys

matteobettini:
## Description

This PR solves issue #775

### DIFF
--- a/torchrl/envs/vec_env.py
+++ b/torchrl/envs/vec_env.py
@@ -40,7 +40,7 @@ def _check_start(fun):
 
 def _sort_keys(element):
     if isinstance(element, tuple):
-        return "".join(element)
+        return "_-|-_".join(element)
     return element
 
 
--- a/torchrl/envs/vec_env.py
+++ b/torchrl/envs/vec_env.py
@@ -1033,6 +1033,8 @@ def _run_worker_pipe_shared_mem(
                 )
             if pin_memory:
                 _td.pin_memory()
+            print("tensordict", tensordict)
+            print("_td", _td)
             tensordict.update_(_td.select(*step_keys, strict=False))
             if _td.get("done"):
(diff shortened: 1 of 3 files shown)


torchrl/envs/vec_env.py
```diff
@@ -39,6 +38,12 @@ def decorated_fun(self: _BatchedEnv, *args, **kwargs):
     return decorated_fun
 
 
+def _sort_keys(element):
+    if isinstance(element, tuple):
+        return "".join(element)
```
vmoens, line 43: let's use a separator that is unlikely to be used in the tensordict
(for instance we may have a key `"nextobs"` and a key `("next", "obs")`)
I sometimes use `"_-|-_"` or similar, take your pick :)

vmoens (commented):
LGTM, let's just make sure that we don't have keys that collide

torchrl/envs/vec_env.py
matteobettini: Thanks, this is very useful toknow. Fixed

vmoens:
@matteobettini since the problem was occuring with Brax I updated the brax wrapper and tests to make sure that things were passing.
In general, it does not really make sense to execute brax in parallel envs (since it's vectorized), but it's a good way to spot bugs.

matteobettini:
> @matteobettini since the problem was occuring with Brax I updated the brax wrapper and tests to make sure that 
things were passing.
> In general, it does not really make sense to execute brax in parallel envs (since it's vectorized), but it's a good way to spot bugs.

@vmoens 
I actually think it is important to use parallel envs with brax (and vmas).

This is because yes they perform batched simulation but parallelenv can furtherly distribute this to the various cpu cores.

This is also why rllib has the 2 parameters: num_workers and num_envs_per_worker

Issue #776, opened 2022-12-31
[BUG] `ParallelEnv` handling of done flag when wrapped envs have non-empty `batch_size`


matteobettini:
## Describe the bug

Currently in `ParallelEnv` the method `_run_worker_pipe_shared_mem` has the following line to check if the environment is done:
```python
_td = env._step(_td)
```
...
```python
if _td.get("done"):
    msg = "done"
else:
    msg = "step_result"
```

The problem is that `_td.get("done")` is a tensor of shape `(*wrapped_env.batch_dim,1)` and thus the if is undefined for a tensor with multiple values.

Furthermore, the `ParallelEnv` uses multiple times the funtions `all()` and `any()` without specified dimensions. This means that it will span over all dimensions (including the environment ones which could be arbitraily many) this has the risk of injecting bias and bugs. This is done for example in the `_reset()` function:

```
self.shared_tensordict_parent.get("done").any()
```

## To Reproduce

Add the instructions ` parallel_env.rand_step()` to the test introduced in #774 

```python
import torchrl
env = MockBatchedLockedEnv(device="cpu", batch_size=torch.Size(env_batch_size))
env.set_seed(1)
parallel_env = ParallelEnv(num_parallel_env, lambda: env)
parallel_env.start()
parallel_env.rand_step()
```

```bash
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniforge/base/envs/torchrl/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/opt/homebrew/Caskroom/miniforge/base/envs/torchrl/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/Matteo/PycharmProjects/torchrl/torchrl/envs/vec_env.py", line 1032, in _run_worker_pipe_shared_mem
    if _td.get("done"):
RuntimeError: Boolean value of Tensor with more than one value is ambiguous
```

## Reason and Possible fixes

The decision on how to handle the done flag in environemnts with non-empty batch sizes is not so trivial.
My suggestion is to remove this logic from the function that is throwing the error and give the done vector to the user as is. 
In case that checks have to be done on the done vector we have to keep in mind its arbitrary shape.

I suggest to remove all the logic checks which use `all()` and `any()` on all dimensions such as 
```
 while self.shared_tensordict_parent.get("done").any():
          if check_count == 4:
[...]

PR #788, opened 2023-01-03
[BugFix] `ParallelEnv` handling of done flag

matteobettini:
## Description

This PR addresses issue #776.

It does two things:

1. The `ParallelEnv` workers will not try to test the "done" flag coming out of a wrapped environment. This worked only when the done flag was without a batch_size. Now, on the other hand, the workers reports the step result as is (the done message was not even used anyway)
2. Very important issue. The workers now call `env.reset()` instead of `env._reset()`. this is because some environments do not return the done key in the tensordict returned by `_reset()`. Since later the worker calls `if _td.get("done").any():` this call fails if the environment does not return the done key in the tensordict returned by `_reset()`. `reset()` on the other hand wraps `_reset()` and automatically adds the defaulte done value, solving the issue.


vmoens:
> 2. Very important issue. The workers now call `env.reset()` instead of `env._reset()`. this is because some environments do not return the done key in the tensordict returned by `_reset()`. Since later the worker calls `if _td.get("done").any():` this call fails if the environment does not return the done key in the tensordict returned by `_reset()`. `reset()` on the other hand wraps `_reset()` and automatically adds the defaulte done value, solving the issue.

I'm curious: can't we use default?
As of now default can't be a callable, so we could do
```python
done = _td.get("done", None)
if done is None:
    done = torch.zeros(_td.shape, dtype=torch.bool, device=device)
```

vmoens (commented):
Some tests are failing, I suspect that it's because of the `_reset` to `reset` change. 
Can we try with the `False` default?

matteobettini:
> Some tests are failing, I suspect that it's because of the `_reset` to `reset` change.
> Can we try with the `False` default?

Yes your options definitely work! I thought that that defaulting was all that reset() deos plus simple checks, but maybe was wrong. Will fix

matteobettini:
Apparently just seeting the default in the if is not enough as later the done is checked again in

```python
 while self.shared_tensordict_parent.get("done").any():
            if check_count == 4:
                raise RuntimeError("Envs have just been reset but some are still done")
```

Thus i now tried as it is done in `reset()` to set the default.

In any case, I plan to propose some bigger modifications to this type of checking of the `done` with `any()`. Because with vectorize enironments it doesn't make sense. So I thing theseparts might be changed in the near future.

torchrl/envs/vec_env.py
```diff
@@ -1004,6 +1002,10 @@ def _run_worker_pipe_shared_mem(
                 raise RuntimeError("call 'init' before resetting")
             # _td = tensordict.select("observation").to(env.device).clone()
             _td = env._reset(**reset_kwargs)
+            _td.set_default(
+                "done",
+                torch.zeros(*_td.batch_size, 1, dtype=torch.bool, device=_td.device),
```
vmoens, line 1007: one issue with set_default is that we're creating the tensor at every single call, thereby allocating memory that is then collected if unused. 
Depending on the kind of env you're running this may or may not cause some overhead. In practice, a boolean tensor is quite cheap so I might be overthinking it.
The solutions might be: 
1. leave it as it is
2. use `torch.zeros(1, ...).expand(*_td.batch_size, 1)` which uses less memory but requires more operations
3. (related to 1) implement a `setdefault` with callable, though this has been discussed by the python community and they [decided not to implement it](https://bugs.python.org/issue31240) for some valid reasons. Since we'd like to stay as close as we can from dict I would not vote for it (besides, we have a `set_default` method when it should be `setdefault` 😱)
4. do a `try / except KeyError` but the issue here is that `try / except` causes an overhead of its own when the except is reached (which should happen less often than not, thereby reversing the memory overhead that occurs say 90%. of the time to a compute time overhead that occurs 10% of the time)
5. check if the key is there, and if not populate. This is basically what set_default does so I would not expect this to be slower, but obviously less 'atomic'.  

Also tensordict do not always have a device (ie device can be None). This is to enable tensordict that collect all the state_dict of a module distributed across several GPUs, for instance. 
Not sure it is relevant here as the tensordict will have the device of the env and the env must have a device, so it's unlikely that it will cause any bug any time soon.

matteobettini: I leave the choice to you. Whatever we choose we should also update the normeal `EnvBase.resert()` then, as this is what is done there https://github.com/pytorch/rl/blob/main/torchrl/envs/common.py#L454

My personal opinion is that all this checks on the done flag should be removed

vmoens (approved):

PR #789, opened 2023-01-03
[Feature] MultiDiscreteTensorSpec nvec with several axes

riiswa:
## Description

I rewrited the code of the PR #783 to support nvec with several axes (Sorry for the double PR I pushed late ^^). Related to #781. 

Example:

```python
>>> ts = MultiDiscreteTensorSpec([[4, 2], [6, 9]])
>>> ts.rand()
tensor([[0, 1],
        [3, 6]])
```

The code is a bit complicated, but it's necessary to generalize to n dimensions, don't hesitate to tell me if you have any questions or if I should add some comments

## Motivation and Context

Improve compatibility with Gym Spaces that have `MultiDiscrete` that support nvec with several axes.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist


riiswa:
Should I add the expected behavior of to_categorical and to_one_hot to this PR ? If so, I will put it in draft form until then

test/test_tensor_spec.py
```diff
@@ -258,22 +259,24 @@ def test_multi_discrete(shape, ns):
     np.random.seed(0)
     ts = MultiDiscreteTensorSpec(ns)
     _real_shape = shape if shape is not None else []
-    _len_ns = [len(ns)] if len(ns) > 1 else []
+    nvec_shape = torch.tensor(ns).size()
+    if nvec_shape == torch.Size([1]):
```
vmoens, line 263: Can't we do
```
@pytest.mark.parametrize(
    "ns",
    [
        5,
        [5, 2, 3],
        [4, 5, 1, 3],
        [[1, 2], [3, 4]],
    ],
)
```
and avoid this?
```
torch.tensor(5)
```
should work just fine.

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1027,77 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
+        x = []
+        for _s in space:
+            if isinstance(_s, BoxList):
+                x.append(self._rand(_s, shape))
+            else:
+                x.append(
+                    torch.randint(
+                        0,
+                        _s.n,
+                        torch.Size([1, *shape]),
```
vmoens, line 1054: I'm a bit puzzled, maybe there's something I'm missing
Can't we just do

```python
                x.append(
                    torch.randint(
                        0,
                        _s.n,
                        shape,
                        device=self.device,
                        dtype=self.dtype,
                    )
                )
        return torch.stack(x, -1)

```
and avoid the permutation in the end?

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1027,77 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
-            ]
-        ).squeeze()
-        _size = [self._size] if self._size > 1 else []
-        return x.T.reshape([*shape, *_size])
 
-    def _split(self, val: torch.Tensor):
-        return [val] if self._size < 2 else val.split(1, -1)
+        x = self._rand(self.space, shape)
+        _size = [] if self.shape == torch.Size([1]) else self.shape
+        return x.permute(*torch.arange(x.ndim - 1, -1, -1)).reshape([*shape, *_size])
```
vmoens, line 1067: if shape = [3, 4, 5]
I understand you end up with a tensor of shape `[*size, 3, 4, 5]` but then you permute it in `[5, 4, 3, *size]` and finally reshape it to `[3, 4, 5, *size]`. This seems wrong: the first dims will be mixed up

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1027,77 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
 
     def _project(self, val: torch.Tensor) -> torch.Tensor:
         if val.dtype not in (torch.int, torch.long):
```
vmoens, line 1070: can we use this class with non integer dtype?
If no, we could just check that `val.dtype` matches the dtype of the object.
If yes, we could check `self.dtype` instead and always call `round`

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1027,77 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
-        return torch.cat(
-            [
-                _val.clamp_(min=0, max=space.n - 1).unsqueeze(0)
-                for _val, space in zip(vals, self.space)
-            ],
-            dim=-1,
-        ).squeeze()
+            val = torch.round(val).type(self.dtype)
+        val = val.unsqueeze(0)
+        for permutation in itertools.product(*[range(d) for d in self.shape]):
```
vmoens, line 1073: Can you explain what this does?
If it's to make sure that `val` with shape `[A, B, C, D]` is bounded by `self.nvec` with shape `[C, D]`, we can just do
```
maxvals = torch.randint(1, 10, (5, 6))
x = torch.randint(1, 10, (3, 4, 5, 6))
x[x >= maxvals] = maxvals.expand_as(x)[x >= maxvals] - 1
```

you can check that after that, all `x` are smaller than the corresponding `maxvals`

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1027,77 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
+        if val.ndim < 1:
+            val = val.unsqueeze(0)
+        val_have_wrong_dim = (
+            self.shape != torch.Size([1])
+            and val.shape[-len(self.shape) :] != self.shape
         )
+        if self.dtype != val.dtype or len(self.shape) > val.ndim or val_have_wrong_dim:
+            return False
+
+        for permutation in itertools.product(*[range(d) for d in self.shape]):
```
vmoens, line 1089: same here, I think we can just do
```python
maxvals = torch.randint(1, 10, (5, 6))
x = torch.randint(1, 10, (3, 4, 5, 6))
assert (x[x < maxvals] & x>0).all()
```

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1027,77 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
-        vals = self._split(val)
-        return self._size == len(vals) and all(
-            [
-                (0 <= _val).all() and (_val < space.n).all()
-                for _val, space in zip(vals, self.space)
-            ]
+        if val.ndim < 1:
+            val = val.unsqueeze(0)
+        val_have_wrong_dim = (
+            self.shape != torch.Size([1])
```
vmoens, line 1083: what is so special about `Size([1])`?

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1027,77 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
+            return False
+
+        for permutation in itertools.product(*[range(d) for d in self.shape]):
+            x = val.unsqueeze(0)[[..., *permutation]]
+            if not ((0 <= x).all() and (x < self.nvec[permutation]).all()):
+                return False
+        return True
 
     def to_onehot(self) -> MultOneHotDiscreteTensorSpec:
+        if len(self.shape) > 1:
```
vmoens, line 1096: what is it's `Size([3])`, in which case the error won't be raised but the size is not [1]

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1027,77 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
+            raise RuntimeError(
+                f"DiscreteTensorSpec with shape != tensor.Size([1]) can't be converted OneHotDiscreteTensorSpec. Got "
+                f"shape={self.shape}."
```
vmoens, line 1099: Can you add `"this could be accomplished via padding or nestedtensors but it is not implemented yet. If you would like to see that feature, please submit an issue of torchrl's github repo."`

vmoens (commented):
I left a bunch of comments.
Not sure they all make sense

test/test_tensor_spec.py
riiswa: Done, I also updated the __init__ method to take scalar for nvec

torchrl/data/tensor_specs.py
riiswa: Yes, I updated the code, and after stacking in the last dimension I just need to a transpose when nvec have many dimension  (between the last dim and the first dim of `*size`)

torchrl/data/tensor_specs.py
riiswa: Something like that ?
```
if not self.dtype.is_floating_point:
    val = torch.round(val)
```

torchrl/data/tensor_specs.py
riiswa: You have understood what I have been trying to do. I didn't know that indexing in this way was possible. Thanks a lot for the tip, the code is much cleaner now!

torchrl/data/tensor_specs.py
riiswa: This is my condition: `((val >= torch.zeros(self.nvec.size())) & (val < self.nvec)).all().item()`

torchrl/data/tensor_specs.py
riiswa: Just an oversight, but it's fixed :)

torchrl/data/tensor_specs.py
riiswa: if `ts.shape == torch.Size([1])`, nvec is a scalar and can be compared to `val` with any shape. For example if val have the shape `torch.Size([4, 5])` and I check if the last dims match, it will be problematic

riiswa:
I tried to rewrite the code in agreement with your comments. Thank you!

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1026,80 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
-        vals = self._split(val)
-        return torch.cat(
-            [
-                _val.clamp_(min=0, max=space.n - 1).unsqueeze(0)
-                for _val, space in zip(vals, self.space)
-            ],
-            dim=-1,
-        ).squeeze()
+        val = val.type(self.dtype)
+        val[val >= self.nvec] = (self.nvec.expand_as(val)[val >= self.nvec] - 1).type(
```
vmoens, line 1078: since you assign it in-place the type casting may not be necessary

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1026,80 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
+        x = self._rand(self.space, shape)
+        if self.nvec.ndim > 1:
+            x = x.transpose(len(shape), -1)
+        return x.squeeze(-1)
 
     def _project(self, val: torch.Tensor) -> torch.Tensor:
-        if val.dtype not in (torch.int, torch.long):
+        val_is_scalar = val.ndim < 1
+        if val_is_scalar:
+            val = val.unsqueeze(0)
```
vmoens, line 1074: why do we unsqueeze and squeeze back? Can't we work with an empty shape?

torchrl/data/tensor_specs.py
vmoens: Got it thanks.

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1026,80 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
-            ]
-        ).squeeze()
-        _size = [self._size] if self._size > 1 else []
-        return x.T.reshape([*shape, *_size])
 
-    def _split(self, val: torch.Tensor):
-        return [val] if self._size < 2 else val.split(1, -1)
+        x = self._rand(self.space, shape)
+        if self.nvec.ndim > 1:
+            x = x.transpose(len(shape), -1)
```
vmoens, line 1068: What does this do? Can you explain?

say `spec.shape = [3, 4]` and `shape=[1, 2]`
you want `x` to have shape `[1, 2, 3, 4]`. From what I understand the output of `_rand` has that shape already.
Why do you invert the dim with size `4` with the one with size `2`?

torchrl/data/tensor_specs.py
```diff
@@ -1014,61 +1026,80 @@ class MultiDiscreteTensorSpec(DiscreteTensorSpec):
...
+        return x.squeeze(-1)
```
vmoens, line 1069: If the shape is `Size([1])`, shouldn't we keep the last dim?

vmoens (commented):
I'm still not sure I grasp everything can you comment a bit further?

torchrl/data/tensor_specs.py
riiswa:
```python
RuntimeError: Index put requires the source and destination dtypes match, got Int for the destination and Long for the source.
```

torchrl/data/tensor_specs.py
riiswa: When val is scalar and I don't unsqueeze and squeeze I get :

```
val[val >= self.nvec] = (self.nvec.expand_as(val)[val >= self.nvec] - 1).type(
RuntimeError: expand(torch.LongTensor{[1]}, size=[]): the number of sizes provided (0) must be greater or equal to the number of dimensions in the tensor (1)
```

torchrl/data/tensor_specs.py
riiswa: It's a oversight, when ts.shape == self.shape == torch.Size([1]), the previous computation add an empty dimension, so I added this case:

```python
if self.shape == torch.Size([1]):
	x = x.squeeze(-1)
return x
```

torchrl/data/tensor_specs.py
riiswa: No, in this case x have a shape of `torch.Size([1, 2, 4, 3])` without the transpose. So I have to invert 4 and 3 (no the one with size 2, because index start with 0).

This is the log of the _rand algorithm (after and before stacking):

```python
[torch.Size([1, 2]), torch.Size([1, 2]), torch.Size([1, 2]), torch.Size([1, 2])]
torch.Size([1, 2, 4])
[torch.Size([1, 2]), torch.Size([1, 2]), torch.Size([1, 2]), torch.Size([1, 2])]
torch.Size([1, 2, 4])
[torch.Size([1, 2]), torch.Size([1, 2]), torch.Size([1, 2]), torch.Size([1, 2])]
torch.Size([1, 2, 4])
[torch.Size([1, 2, 4]), torch.Size([1, 2, 4]), torch.Size([1, 2, 4])]
torch.Size([1, 2, 4, 3])
```

Because at the end I should stack in -2 dim

What do you think about this solution:

```python
    def _rand(self, space: Box, shape: torch.Size, i: int):
		....
                x.append(self._rand(_s, shape, i -1))
		....
        return torch.stack(x, -i)
```

Instead stacking in the last dimension I stack in a dimension according to the depth in the recursive box list, I pass the test with this solution.        

@vmoens

riiswa:
The CI looks broken :/

vmoens:
> The CI looks broken :/

It's weird, can you try pushing an empty commit?
`git commit --allow-empty -m empty`

riiswa:
Done but still broken

vmoens:
I managed to get it to run!

PR #794, opened 2023-01-04
[Refactor] Graduate Replay Buffer prototype

KamilPiechowiak:
## Description

This change replaces ReplayBuffers implementations in replay_buffers.py by composable implementations from rb_prototype.py.
This is a breaking change. ReplayBuffer constructor no longer accepts size parameter. Now it is only provided to the underlying Storage. This change also makes sample return type consistent across ReplayBuffer and its derived classes. It is also a breaking change. That is why many files needed to be updated.

## Motivation and Context

This change allows to create new types of replay buffers by composing samplers and writers.

## Types of changes

## Checklist


README.md
```diff
@@ -335,7 +334,7 @@ The associated [`SafeModule` class](torchrl/modules/tensordict_module/common.py)
     ```python
     from torchrl.objectives import DQNLoss
     loss_module = DQNLoss(value_network=value_network, gamma=0.99)
-    tensordict = replay_buffer.sample(batch_size)
+    tensordict = replay_buffer.sample(batch_size)[0]
```
vmoens, line 337: What's your view on this?
I feel that the `[0]` index is bothersome. With `TensorDictReplayBuffer` we read the second output from `sample` (`info`) and write the content in the tensordict.
Maybe we should just have a `def sample(self, ..., return_info=False)` flag that can be turned on when needed.
Wdyt?

KamilPiechowiak: I also don't like the `[0]` index. However, the version of `TensorDictReplayBuffer` proposed in `rb_prototype.py` also returns 2 values from `sample()`. While migrating I decided to keep this logic. If we change it, we break consistency with parent class `ReplayBuffer`. I don't know if there are cases when users would like to use both `ReplayBuffer` and `TensorDictReplayBuffer` interchangeably in a single piece of code. If this is not the case, we can stop returning info (or return it only if the flag is set).

vmoens: Ok, I see two ways of going about this
1. branch out your branch and have a `sample()` function with a single output, merge it after review in your branch and ship the whole thing. Advantage: we don't bc-break nightly twice
2. ship this PR as it is and solve it in a second time.
Any pref?

vmoens: > I don't know if there are cases when users would like to use both ReplayBuffer and TensorDictReplayBuffer 

It's better if they're consistent, even if they're not interchangeable

KamilPiechowiak: I think that the last approach (1 value by default in base and derived classes) is best.
If you agree, I'll make a new PR with this change.

vmoens: yeah, one value by default unless return_info = True
Thanks for taking care of this!

KamilPiechowiak: I've updated this PR with the required changes.

torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -261,431 +284,217 @@ class PrioritizedReplayBuffer(ReplayBuffer):
...
+        priority = torch.tensor(
+            [self._get_priority(td) for td in data],
+            dtype=torch.float,
+            device=data.device,
+        )
+        self.update_priority(data.get("index"), priority)
+
+    def sample(
+        self, batch_size: int, include_info: bool = False, return_info: bool = False 
+    ) -> TensorDictBase:
```
vmoens, line 409: Can you write a docstring for that?

torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -190,66 +195,84 @@ def extend(self, data: Sequence[Any]):
...
-
-    def sample(self, batch_size: int) -> Any:
+    def sample(self, batch_size: int, return_info: bool = False) -> Any:
         """Samples a batch of data from the replay buffer.
 
+        Uses Sampler to sample indices, and retrieves them from Storage.
+
         Args:
-            batch_size (int): float of data to be collected.
+            batch_size (int): size of data to be collected.
```
vmoens, line 228: return_info is missing

vmoens (approved):
Amazing! Can you look at the 2 comments I left?
Brilliant work I love it

torchrl/data/replay_buffers/replay_buffers.py
KamilPiechowiak: Done.

torchrl/data/replay_buffers/replay_buffers.py
KamilPiechowiak: Fixed.

KamilPiechowiak:
> the lint seems to be failing, can you fix that?

Linting fails because of two things:
- there are type annotations 'Transform' as strings and linter doesn't like it. I did it this way to avoid circular references from envs.transforms. It is possible to remove this linter error by making a global variable  Transform = 'Transform' and use it in annotations but it also doesn't look good. Maybe you have other ideas on how to solve it?
- ReplayBuffer is imported in trainers.py. It is not used there, but it is later imported somewhere else. I'll find this import and import from torchrl.data.

vmoens:
> * there are type annotations 'Transform' as strings and linter doesn't like it. I did it this way to avoid circular references from envs.transforms. It is possible to remove this linter error by making a global variable  Transform = 'Transform' and use it in annotations but it also doesn't look good. Maybe you have other ideas on how to solve it?

if that's the problem you can just put a `# noqa-F821` comment at the end of the lines where the lint complains

KamilPiechowiak:
Reopen.

vmoens (approved):

PR #799, opened 2023-01-06
[Benchmark] Data benchmarks

vmoens:
## Description

Adds the R3M, pong and dmc benchmarks


gtnv:
hi @vmoens, is this still planned? if not, would you like to close it?

Issue #790, opened 2023-01-04
[BUG] Resetting environments with non-empty batch_size


matteobettini:
## Describe the bug

When resetting an environment with non-empty batch size, there is currently no unified way to specify which dimensions to reset.

`ParallelEnv` has a reset key called `"reset_workers"` which is used to choose which workers to reset. The use of this unidimensional key makes `ParallelEnv` crash when `env.batch_size` is not empty.

This is what happens:

```python
def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
    cmd_out = "reset"
    if tensordict is not None and "reset_workers" in tensordict.keys():
        self._assert_tensordict_shape(tensordict) # First assert that the key has the same batch size as the env, let's say [10,4,5,2]
        reset_workers = tensordict.get("reset_workers")
    else:
        reset_workers = torch.ones(self.num_workers, dtype=torch.bool) # If not create one (without respecting the batch size)

    for i, channel in enumerate(self.parent_channels):
        if not reset_workers[i]: # If run on dimension 0 !! This only works if the else branch is taken or env.batch_size is empty
            continue
        channel.send((cmd_out, kwargs)) # Do not even pass the tensordict to the env
```

The problems just in this snippet are:
1. This only works if the else branch is taken or env.batch_size is empty
2. it does not pass the reset tensordict to the env

This is paired with a series of problems in the various reset functions (ParallelEnv and EnvBase) where after reset `done.any()` is called. This is highly problematic as any spans over all dimensions.

## Proposed changes to the API

1. Remove "reset_workers" (which already doesn't work)
2. Introduce the possibility of having a "reset" key in the tensordict given as parameter to the reset functions. This reset key has shape (in the more general case of `ParallelEnv`) `(n_parallel_envs, *env.batch_size)` and is a boolean telling *precisely* which dimensions to reset. ParallelEnv then can call `reset[worker_id].any()` to know if to pass the reset command and key to the worker
3. In the reset function *do not* check `done.any()` but chacke that at least the requested dimensions to reset have done=False. The absence of the "reset" key means resetting all dimensions

## To Reproduce

```python
env = MockBatchedLockedEnv(device="cpu", batch_size=torch.Size(env_batch_size))
env.set_seed(1)
parallel_env = ParallelEnv(num_parallel_env, lambda: env)
parallel_env.start()
reset_td = TensorDict(
    {"reset_workers": torch.full(parallel_env.batch_size, True, device=parallel_env.device)},
[...]

PR #800, opened 2023-01-06
[BugFix] [Feature] "_reset" flag for env reset

matteobettini:
## Description

This PR ddresses issue #790.

The changes replace the "reset_workers" flag (only deigned for ParallelEnvs wrapping environments with emty batch_size) with the "_reset" flag, which spans over all batch_size dimensions.

This allows to more precisely tell the wrapped environemnts which dimensions need to be reset.

In accordace to this, now the `reset()` methods on `EnvBase` and `ParallelEnv` only check that at least the indexes that were flagged to be reset are not done. Instead of checking `assert not done.any()`.


matteobettini:
@vmoens 
I will write a few test for the non-regression of this feature.

In the meantime, if you could have a quick look at the changes I have made to `collectors.py` since I am not super familiar with that part of the codebase yet.

torchrl/envs/common.py
```diff
@@ -457,11 +457,21 @@ def reset(
...
-        if tensordict_reset.get("done").any():
+        if tensordict is not None and "_reset" in tensordict.keys():
+            self._assert_tensordict_shape(tensordict)
+            _reset = tensordict.get("_reset")
+        else:
+            _reset = None
+
+        if (_reset is None and tensordict_reset.get("done").any()) or (
+            _reset is not None and tensordict_reset.get("done")[_reset].any()
+        ):
```
vmoens, line 468: What about 
```python
        if tensordict is not None:
            self._assert_tensordict_shape(tensordict)
        _reset = tensordict.get("_reset", None)

        if (_reset is None and tensordict_reset.get("done").any()) or (
            _reset is not None and tensordict_reset.get("done")[_reset].any()
        ):
```

vmoens (approved):
Minor comment but otherwise LGTM

vmoens (commented):
I did not see but the CI seems broken, let me see what happened

torchrl/envs/common.py
matteobettini: yes thanks!

matteobettini:
The current version of the changes adopts the logic of deleting the _reset flag right after it has been used. `EnvBase`, `ParallelEnv`and `SerialEnv` will delete the "_reset" flag, if present, right after resetting. This implies that external components like collectors just have to worry about setting the flag and not deleting it

matteobettini:
I added some tests, CI seems broken on some http stuff

matteobettini:
While finishing up this PR I stumbled upon an unexpecte behavior that I am not sure is intended.

Imagine having an `EnvBase` where the reset function returns "done" and "observation".

You wrap it in a `ParallelEnv` with 2 workers and first things first you call reset, asking to reset only the first worker. What you will get is observations for both workers, but the ones for the worker which was not reset is random, since that worker did not even call its reset methos (as intended).

Is this expected?

vmoens:
> While finishing up this PR I stumbled upon an unexpecte behavior that I am not sure is intended.
> 
> 
> 
> Imagine having an `EnvBase` where the reset function returns "done" and "observation".
> 
> 
> 
> You wrap it in a `ParallelEnv` with 2 workers and first things first you call reset, asking to reset only the first worker. What you will get is observations for both workers, but the ones for the worker which was not reset is random, since that worker did not even call its reset methos (as intended).
> 
> 
> 
> Is this expected?

Good point, but what would be the expected behaviour?

matteobettini:
> Good point, but what would be the expected behaviour?

The rule I use in vmas is to reset only the ones that need resetting and call `env.observation(agent)` for all envs and all agents. Given that here such an observatio function is not available in all libs I think we have a few options:

## Option 1: call `reset()` for all workers indeendently
This calls reset on all the workers and passes the "_reset" flag. This flag would be all Falses for the workers which didn't need resetting. The envs then have to look at the "_reset" flag and, if the flag is all False, only return the observation without resetting.
 
### Cons

Every envioronment in torchrl has to introduce the logic to handle the _reset flag and this behavior. Might be asking too much

## Option 2: Heterogenous tensordicts

Only return the observation and keys that are actually available. If some workers were not reset, their observation dimension will be 0 and the return tensordict will be heterogeneous. 

## Option 3: Use NaNs or other placeholders.

Instead of putting random values for workers which were not reset, use a convetion of values (nans, inf, zeros).

## Option 4: leave as is

When `reset()` is called after a `step()` the value of observations for workers which were not reset is the last one available from the step. The only case where the behavior is undefined is when reset is called before step only for some workers. We could precisely define what happens in this case in the docs and say that one should only access the values of
```python
_reset = torch.randint(0,2, size=(*parallel_env.batch_size,1), dtype=troch.bool)
reset_td = parallel_env.reset(TensorDict({"_reset":_reset}, batch_size=parallel_env.batch_size))
reset_td[_reset] # Only defined values
```

vmoens:
I agree with you that options 3 and 4 look better than others.
In general, while I see the issue, I would prefer not to adopt a solution that forces users that do not use parallel envs / multi-agent to reset envs in a way that is more complex than needed.

To build the fake_tensordict that serves as buffer for the parallel env, [we use rand](https://github.com/pytorch/rl/blob/569161ee4d62eb264cfb5a64d0578d7021b3701a/torchrl/envs/common.py#L699-L719) with no reason. We could perfectly use NaN, zeros or anything else.

Also, with the collectors we return a "mask" entry that represents what indices of the data is valid. It's a way to make sure that users have access to what was or wasn't the result of a reset. Would you consider something like that?

matteobettini:
That makes sense yes, we can leave as is. Does the mask in the collectors already take into account this?

vmoens:
> That makes sense yes, we can leave as is. Does the mask in the collectors already take into account this problems when resetting parallel workers?

No it just keeps track of padding operations. I was mentioning this as a mechanism to track valid steps.

Another thing to consider is this: if your env supports sub-envs that have not been reset but exist (e.g. in some lib you must reset before doing anything else), you could just append a transform that does some step counting. If an env has been reset, the steps will be strictly greater (or greater or equal?) than 0. If a step is 0 (or -1?), no step has actually been done.

See [this class](https://github.com/pytorch/rl/blob/569161ee4d62eb264cfb5a64d0578d7021b3701a/torchrl/envs/transforms/transforms.py#L2600)

e.g.
```python
base_env = MyEnv(..., n_envs=2)
env = TransformedEnv(base_env, StepCounter())
tensordict = env.reset(TensorDict({"_reset": [True, False]}, [2]))
print(tensordict["steps"])  # prints [0, -1]
```

@riiswa I don't think this behaviour is currently supported, but do you think it would make sense?

matteobettini:
I think this PR can be merged as standalone since it is mainly refactoring "reset_workers" to "_reset" and putting in place some mechanism to handle it. If users want to use "_reset",  it will only work in envs that support it. 

The only libs with sub_envs to my knowledge are vmas and brax. vmas supports it. We could look into enabling support for Brax, but I am not super familiar with jax and the api.

test/test_env.py
```diff
@@ -900,6 +901,76 @@ def env_fn2(seed):
...
+        assert (td["next"]["observation"] == max_steps + 1).all()
+
+        _reset = torch.randint(
+            low=0, high=2, size=(*env.batch_size, 1), dtype=torch.bool
+        )
+        td_reset = env.reset(
+            TensorDict({"_reset": _reset}, batch_size=env.batch_size, device=env.device)
+        )
+        env.close()
+        if _reset.any():
```
vmoens, line 936: How can we make sure that this is reached (I agree there's a low prob that it isn't)
a few idead:
- put a `else: RuntimeError` in the end
- same as above + repeat the test X times (e.g. X=3) if failed with a different seed

test/test_env.py
```diff
@@ -900,6 +901,76 @@ def env_fn2(seed):
         env1.close()
         env2.close()
 
+    @pytest.mark.parametrize("batch_size", [(), (1,), (4,), (32, 5)])
+    @pytest.mark.parametrize("n_workers", [1, 2])
+    def test_parallel_env_reset_flag(self, batch_size, n_workers, max_steps=3):
+        env = ParallelEnv(
```
vmoens, line 907: let's add a `torch.manual_seed(x)` here

test/test_env.py
```diff
@@ -900,6 +901,76 @@ def env_fn2(seed):
...
+            assert (td["next"]["observation"] == i + 1).all()
+
+        td = env.step(
+            TensorDict({"action": action}, batch_size=env.batch_size, device=env.device)
+        )
+        assert (td["done"] == 1).all()
+        assert (td["next"]["observation"] == max_steps + 1).all()
+
+        _reset = torch.randint(
+            low=0, high=2, size=(*env.batch_size, 1), dtype=torch.bool
```
vmoens, line 930: why is the last dim 1? With the latest tensordict version, the dims of a tensor can match the batch size of a tensordict.

torchrl/collectors/collectors.py
```diff
@@ -683,23 +679,27 @@ def reset(self, index=None, **kwargs) -> None:
             # check that the env supports partial reset
             if prod(self.env.batch_size) == 0:
                 raise RuntimeError("resetting unique env with index is not permitted.")
-            reset_workers = torch.zeros(
-                *self.env.batch_size,
+            _reset = torch.zeros(
+                (*self.env.batch_size, 1),
```
vmoens, line 683: same here, do we need the last dim to be 1?

vmoens (approved):
LGTM a few minor issues to solve and we're good to go

test/test_env.py
matteobettini: I used the convention of the done flag which has shape `(*batch_size,1)`. I adopted this convetion for _reset also in all the PR files

test/test_env.py
matteobettini: Good point. Does it make sense now?

test/test_env.py
matteobettini: See here for example

matteobettini: Or the one me and you did recently https://github.com/pytorch/rl/blob/main/torchrl/envs/vec_env.py#L1013

vmoens: For `"done"`, we want the last dim to be 1 because we do this kind of thing:
```python
value = reward + value * gamma * (1-done)
```
we want reward and done to have a last dim of 1 otherwise they will be casted to the size of `value` which is usually the output of a neural net like `nn.Sequential(..., nn.Linear(..., 1))`.
So either we squeeze value or we unsqueeze reward and done.
Squeezing requires more brain power from the users IMO, whereas unsqueezing can be hidden from them.

For `"_reset"` I don't think we need that

matteobettini: There are then some choices we have to make in the `StepCounter`. Because this transoform in instantiation the "step_count" key which has shape (env.batch_dim,1)

So do we want to also make that key loose the 1? or we keep the 1 there and unsqueeze _reset?

vmoens: IMO StepCounter should have steps that match the tensordict batch size.
@riiswa

matteobettini: ok so like it is now in the pr, i agree

matteobettini: In the future it might be worth considering forcing one of the two convenctions for all tensordicts imo. I.e. all keys of shape (batch_size, 1) have to be squeezed. or the opposite. I understand that for reward and done one way might be more comfortable, but I find it confusing that some keys and specs are squeezed and others are not

vmoens (approved):
Great work
I love that the library becomes more and more compatible with multiagent!

test/test_env.py
vmoens: yep it's been under my radar for quite some time. It's a conflicting thing:
Before tensordict was designed such that you could not store tensors with a shape identical to the tensordict (you had to unsqueeze or it was done for your). 
That was surprising to many users, so we dropped that.
But as mentioned above, for reward and done, they interact so much with neural nets that it's pain to work with batched data that does not end with a singleton dimension. 
Maybe unsqueezing in the env is not the solution though, I don't know really. We could delegate that to the value functions and such. My main worry is that it may lead to silent errors, e.g.

```
done = torch.zeros(10)
next_value = torch.ones(10, 1)
value = done*next_value  # silently gives the wrong 10 x 10 tensor
```
it's not only a problem for us but also for the users. That's why I'm not sure we're doing them a favour by removing the unsqueeze. 

But I'm more than happy to keep the conversation going.

@tcbegley was involved in some of this refactoring, he may have a 2 cents to share

PR #803, opened 2023-01-08
[BugFix] Solve R3MTransform init problem

vmoens:
## Description

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/r3m.py
+++ b/torchrl/envs/transforms/r3m.py
@@ -26,6 +26,7 @@ from torchrl.envs.transforms.transforms import (
     Transform,
     UnsqueezeTransform,
 )
+from torchrl.envs import EnvBase
 
 try:
     from torchvision import models
@@ -167,6 +168,9 @@ class R3MTransform(Compose):
              "r3m_vec" is assumed.
         size (int, optional): Size of the image to feed to resnet.
             Defaults to 244.
+        stack_images (bool, optional): if False, the images given in the :obj:`in_keys`
+             argument will be treaded separetely and each will be given a single,
+             separated entry in the output tensordict. Defaults to :obj:`True`.
         download (bool, optional): if True, the weights will be downloaded using
             the torch.hub download API (i.e. weights will be cached for future use).
             Defaults to False.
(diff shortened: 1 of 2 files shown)


torchrl/envs/transforms/r3m.py
```diff
@@ -328,9 +360,36 @@ def device(self):
...
         return self._dtype
 
+    def set_container(self, container: Union[Transform, EnvBase]) -> None:
+        out = super().set_container(container)
+        # try to get the parent: if it exists, initialize. Otherwise, skip
+        try:
+            parent = self.parent
+            has_parent = parent is not None
+        except AttributeError:
+            has_parent = False
```
riiswa, line 370: Maybe you can use the getattr function with a default argument. Something like that:
```python
parent = getattr(self, "parent", None)
has_parent = parent is not None
```

torchrl/envs/transforms/vip.py
```diff
@@ -303,12 +332,34 @@ def device(self):
     def dtype(self):
         return self._dtype
 
+    def set_container(self, container: Union[Transform, EnvBase]) -> None:
+        out = super().set_container(container)
+        # try to get the parent: if it exists, initialize. Otherwise, skip
+        try:
```
riiswa, line 338: Same here

torchrl/envs/transforms/r3m.py
vmoens: thanks for the tip!

vmoens:
Still unhappy with this.
The lazy init is a nightmare, there's a ton of hidden bugs with it.

## What we want to cover:
Some envs return images with dim `BWHC`. Sometimes, we'd like to process multiple images at the same time.
So what we do is that we allow R3M to cat images along the first dim. We check in the parent env what is the dim of the image: if it is a 3d image, we unsqueeze the first dim to then do the cat operation.

## What I think we should cover
We should only cover cases where the input image has size WHC (or BWHC if B is the tensordict batch size).
Then we would simply stack the images along the first dim after the tensordict.ndim. 

In all other cases, we should leave the error as it is, or point the user to a solution.

PR #812, opened 2023-01-10
[BugFix] Fix habitat integration and doc

vmoens:
## Description

Fixes Habitat integration and doc.
The goal is to make a [benchmark](https://github.com/pytorch/rl/pull/799/files) for Habitat as we do for dm_control and Atari.

cc @vincentpierre


vmoens:
Added support for device casting.
You can now call `env = HabitatEnv(..., from_pixels=True, device=3)` and the whole rendering will be executed on the appropriate device. You can also call `env.to(device)`, and the env will be restarted on `device`.

PR #816, opened 2023-01-10
[Feature] Rework `to_one_hot` and `to_categorical` to take a tensor as parameter

riiswa:
## Description

As mentioned in the closed PR #783 we want to update the methods `to_onehot` and `to_categorical` to support value conversion in addition to spec conversion.

This is what currently happens

```pytho
spec = torchrl.data.MultOneHotDiscreteTensorSpec([5,10])
sample = spec.rand((2,3))
sample.shape
torch.Size([2, 3, 15])
sample_np =spec.to_numpy(sample)
sample_np.shape
(2, 3, 2)
```

```python
spec = torchrl.data.MultOneHotDiscreteTensorSpec([5,10])
sample = spec.rand((2,3))
sample.shape
torch.Size([2, 3, 15])
sample_np =spec.to_numpy(sample)
sample_np.shape
(2, 3, 15)

sample_cat = spec.to_categorical(sample)
sample_cat.shape
torch.Size([2, 3, 2])
```

To keep the current behaviour (conversion of the spec), I decided that `val` will be an optional parameter. The documentation will be well written to explain the different behaviours of theses methods.

## Motivation and Context

Also related to the issue #781 

## Types of changes

## Checklist

### DIFF
--- a/test/test_specs.py
+++ b/test/test_specs.py
@@ -341,14 +341,26 @@ def test_discrete_conversion(n, device, shape):
         [4, 5, 1, 3],
     ],
 )
+@pytest.mark.parametrize(
+    "shape",
+    [
+        torch.Size([3]),
+        torch.Size([4, 5]),
+    ],
+)
 @pytest.mark.parametrize("device", get_available_devices())
-def test_multi_discrete_conversion(ns, device):
+def test_multi_discrete_conversion(ns, shape, device):
     categorical = MultiDiscreteTensorSpec(ns, device=device)
     one_hot = MultiOneHotDiscreteTensorSpec(ns, device=device)
 
     assert categorical != one_hot
     assert categorical.to_onehot() == one_hot
     assert one_hot.to_categorical() == categorical
+
+    assert categorical.is_in(one_hot.to_categorical(one_hot.rand(shape)))
+    assert one_hot.is_in(categorical.to_onehot(categorical.rand(shape)))
+
+    categorical.to_onehot()
 
 
 @pytest.mark.parametrize("is_complete", [True, False])
--- a/test/test_specs.py
+++ b/test/test_specs.py
@@ -240,9 +240,9 @@ def test_mult_onehot(shape, ns):
         for _r, _n in zip(rsplit, ns):
             assert (_r.sum(-1) == 1).all()
             assert _r.shape[-1] == _n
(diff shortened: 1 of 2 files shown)


vmoens:
Hey @riiswa can I help to unblock this?

riiswa:
> Hey @riiswa can I help to unblock this?

Hey, I'm a little bit busy in this moment (exams at school). I think the PR is finished, I juste have a little bug on GPU (created tensor are not on the right device) 😄 

I think you can start the PR review, and when I have time I'll fix this little thing (unless you want to do it yourself)

vmoens:
Np i'll have a look

riiswa:
@vmoens I'm finally back, and I think this PR is ready for a review😄
@matteobettini maybe you should also check, because it's your idea

vmoens (commented):
Sorry to shuffle things a little: don't we want two separate methods for these?

like `to_one_hot` and `DiscreteTensorSpec.to(OneHotDiscreteTensorSpec)`?

riiswa:
Yes it's cleaner, but in case of `DiscreteTensorSpec.to(OneHotDiscreteTensorSpec)` what can be the use of the parameter? I mean we will always convert it to the same OneHotDiscreteTensorSpec`. The method can be named `to_one_hot_spec()`

vmoens:
> Yes it's cleaner, but in case of `DiscreteTensorSpec.to(OneHotDiscreteTensorSpec)` what can be the use of the parameter? I mean we will always convert it to the same `OneHotDiscreteTensorSpec`. The method can be named `to_one_hot_spec()`

My idea (but happy to challenge it) was that some day in the future we may code -- say -- a `HashDiscreteSpec` where a hash key is used to code a discrete value (or some other embedding).
Then it may be useful to have a generic `to(SpecType)` method that handles all sorts of conversion.

We could also map `Bounded` to `Unbounded` with the same structure, etc.

By I agree that `to_one_hot_spec` is more explicit.

riiswa:
> Then it may be useful to have a generic `to(SpecType)` method that handles all sorts of conversion.

For the moment I have written a specific function, let's work on the generic version in the future after implementing the `HashDiscreteSpec`!

vmoens (approved):
LGTM thanks for this!

PR #825, opened 2023-01-12
[BugFix] Fix envs specs and info reading

vmoens:
## Description

Introduces some changes in the env API to allow for:
- Reading infos at reset
- Completing reset data when info is missing
- using zero instead of rand for fake_tensordict (@matteobettini)
- better checks in check_env_spec and better error messages

### DIFF
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -636,7 +636,7 @@ class TestVmas:
         )
         for e in [env, wrapped]:
             e.set_seed(0)
-            check_env_specs(e)
+            check_env_specs(e, check_dtype=False)
             del e
 
     @pytest.mark.parametrize("num_envs", [1, 20])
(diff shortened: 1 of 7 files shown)


test/test_libs.py
```diff
@@ -636,7 +636,7 @@ def test_vmas_spec_rollout(
         )
         for e in [env, wrapped]:
             e.set_seed(0)
-            check_env_specs(e)
+            check_env_specs(e, check_dtype=False)
```
vmoens, line 639: @matteobettini I had to turn this of because some float32 are float64 or the opposite. 
Worth having a look on your side. I don't think it's a torchrl issue but not sure about that

matteobettini: Oh yes it was a vmas issue i fixed, as soon as i release the new release on pypi it should be fixed

Issue #807, opened 2023-01-09
[BUG] Collectors crash with enviornments with non-empty `batch_size`


matteobettini:
## Describe the bug

Collectors seem to have been designed to work with enviornments with empty `batch_size`.

Examples of torchrl environments where colllectors do not work are Brax (with a batch_size representing vectorized enviornments) and Vmas (with a batch_size representing vectorized enviornments and number of agents)

For now these are the issues i have identified:

#### Problem
The collectors assume that the number of environments is the numel of the batch size, which is not true in the multi agent case.
#### Solution

frames per batch should be independent of the batch size. It has to achieve the meaning of "frames of dim batch_size per batch"
`self.frames_per_batch = -(-frames_per_batch // self.n_env)` -> `self.frames_per_batch = frames_per_batch `

#### Problem
```python

    def rollout(self) -> TensorDictBase:
        
         ...

        n = self.env.batch_size[0] if len(self.env.batch_size) else 1
        self._tensordict.set("traj_ids", torch.arange(n).view(self.env.batch_size[:1]))
```
the last line respects only the first dimension of the batch size and ignores the other

### 3. `split_trajectories` operates on existing dimensions of the `batch_size`

If trajectories are not split, the tensordict outputted by the collector has the same size of the one from the env with a last dimensions of dim `frames_per_batch` added to it (adopting the solution from problem 1)

If trajectories are split, on the other hand, the last dim is forced to be equal to `max_frames_per_traj` and the rest of that dim is added to dimension 0, without even knowing what dimension 0 is representing.

## My opinion

In my opinion, we should get rid of the `max_frames_per_traj` parameter. This parameter is effectively acting as an horizon. Its logic can be implemented in the environment either naturally or via a transform (return done after x steps). This parameter has no meaning when envs are multi-dimensional becuase if you are rolling out an env with batch_dim (32,26) for 
`max_frames_per_traj`, you are not rolling out a single trajectory, but a batch_size of trajectories which may terminate and be reset at whatever indexes of the batch_size. 

If we do get rid of this parameter (by having it in the done flag), we can get rid of `split_trajectories`, padding, and the relative mask. 

At every iteration of the collector we would then only collect `frames_per_batch` samples of size `env.batch_size` (samples need to have the dim of the env) and this would just add a final dimension to the output tensordict, which will have batch_size `(*env.batch_size, frames_per_batch)`.

[...]

PR #828, opened 2023-01-13
[BugFix] [Feature] Multi-agent collectors

matteobettini:
## Description

This PR introduces the mask, described in #807, which can be used in multi-agent settings to decide what parts of env batch size to consider for collection

## Example

```python
env = ParallelEnv(n_env_workers, VmasEnv("flocking", n_envs, n_agents))
env.batch_size # (n_env_workers, n_agents, n_envs)

# Case 1
collector = MultiSyncDataCollector([env] * n_collector_workers, policy, env_batch_size_mask=(True,True,True), frames_per_batch=frames_per_batch)
for i, data in enumerate(collector):
    data.batch_size  # (n_env_workers, n_agents, n_envs, frames_per_batch / (n_collector_workers * n_env_workers * n_agents * n_envs))

# Case 2
collector = MultiSyncDataCollector([env] * n_collector_workers, policy, env_batch_size_mask=(True,False,True), frames_per_batch=frames_per_batch)
for i, data in enumerate(collector):
   data.batch_size  # (n_env_workers, n_agents, n_envs, frames_per_batch / (n_collector_workers * n_env_workers * n_envs))

```


matteobettini:
TODO:

- Multi-collectors and their docs
- Tests

matteobettini:
@vmoens 

In the current state, this PR works when `split_traj=False`.

Now, we should discuss how we want to change the behavior os `split_traj`. Currently that function is written to take only into consideration envs with 1 domension (parallel envs) and increases that dimension in order to pad the last one.

Let's take an example to discuss new possibilities, here is the setup:
```
my_env.batch_size = (n_env_workers, n_agents, n_envs)
env_batch_size_mask = (True, False, True)

frames_per_batch = frames_per_batch
n_colletor_workers = n_colletor_workers
max_frames_per_traj = max_frames_per_traj
```

Now in case we start collecting, what should the batch_size be at the output? We are interested the case `split_traj=True`
 and `split_traj=False`.

Here are the options for me:

1. For the `split_traj=False` case, we do as we discussed and how is already done in this PR: the size is
```
(n_env_workers, n_agents, n_envs, frames_per_batch / (n_collector_workers * n_env_workers * n_envs))
```
Then, for the case of `split_traj=True` we add an additional dimesion (let's say at the beginning) and the shape becomes 
```
(frames_per_batch / (n_collector_workers * n_env_workers * n_envs * max_frames_per_traj), n_env_workers, n_agents, n_envs, max_frames_per_traj)
```


2. Alternatively, since this adds yet another dimension in an arbitrary place. We could make so that after collection, collectors squash all the batch dims into 1 single dim according to the mask. This would mean that the collectors with `split_traj=False` return shapes of 
```
(frames_per_batch , n_agents) # which would be just (frames_per_batch) without a mask
```
Then, for the case of `split_traj=True` 
```
(frames_per_batch / max_frames_per_traj, n_agents, max_frames_per_traj)
```

Let me know

vmoens:
Related to #845 
Do we actually care about split_trajs (as integrated in the collector).
As stated in the aforementioned, the usage of split_trajs was to make sure that advantage functions and similar would not overlap computation over consecutive but independent trajectories. I now believe that a properly coded advantage module should not suffer from that issue if the `done` state is used properly. 
For the rest (eg custom loss functions that may be buggy on that regard) we could write a function `split_traj` with a limited behaviour (ie assumes a certain input shape, eg always a 2d tensordict or smth along that line) and do something like what you propose in Option 2. We would properly document this for those interested. What do you think?

smorad:
@vmoens I would really like to have time-dimension padding for my recurrent models. Can we keep this as a helper feature somewhere?

vmoens:
Definitely! What we think is to have it as an option, turned off by default

matteobettini:
So @vmoens if we keep `split_traj` but just default it to false I think we should make it work in the context of this PR (aka agnostic to the env batch size using the mask). Should I try to work on option 2 for all collectors?

vmoens:
> So @vmoens if we keep `split_traj` but just default it to false I think we should make it work in the context of this PR (aka agnostic to the env batch size using the mask). Should I try to work on option 2 for all collectors?

That works for me.
FYI as of #848 all value functions work with contiguous trajectories 🥂

matteobettini:
@vmoens
 
Few, it take a while.
We should have collectors working with both split_traj options.

If `split_traj=False` all `frames_per_batch` will be in the first batch size dimension, dimensions that are masked out from batch_size will be after that (if no dimensions are masked out, there is only the first)

If `split_traj=True` all `frames_per_batch`/ max_traj_length will be in the first batch size dimension, dimensions that are masked out from batch_size will be after that and the last dimension is max_traj_length

I also added a warning in case the requested `frames_per_batch`is not a mulriple of the number of batched dimensions used for collection. Because in this case people will find for collected frames than requested. I think this should be an error.

Let me know what you think, I will write tests now.

Just to try it out I usually use scripts like
```python
n_env_workers = 3
n_collector_workers = 2
n_vectorized_envs = 4
n_agents = 5

vmas_env = lambda: VmasEnv(
    "flocking", num_envs=n_vectorized_envs, n_agents=n_agents
)
vmas_parallel_env = lambda: ParallelEnv(n_env_workers, vmas_env)
vmas_policy = TensorDictModule(
    nn.Linear(18, 2), in_keys=["observation"], out_keys=["action"]
)

gym_env = lambda: GymEnv("Pendulum-v1", device="cpu")
gym_parallel_env = lambda: ParallelEnv(n_env_workers, gym_env)
gym_policy = TensorDictModule(
    nn.Linear(3, 1), in_keys=["observation"], out_keys=["action"]
)

envs = [vmas_parallel_env]
policies = [vmas_policy]
split_traj = False
mask = [True, True, True]
frames_per_batch = n_vectorized_envs * n_collector_workers
for env, policy in zip(envs, policies):
    multi_coll = MultiSyncDataCollector(
        [env] * 2,
        policy,
        total_frames=-1,
        max_frames_per_traj=2,
        mask_env_batch_size=mask,
        split_trajs=split_traj,
    )

    for data in multi_coll:
        print("Ending", data)
        break
```

torchrl/collectors/collectors.py
```diff
@@ -526,15 +578,25 @@ def iterator(self) -> Iterator[TensorDictBase]:
...
+
+            # Bring all batch dimensions to the front (only performs computation if it is not already the case)
+            tensordict_out = tensordict_out.permute(self.permute_out_batch_size)
+            # Flatten all batch dimensions into first one and leave unmasked dimensions untouched
+            if len(self.env_batch_size_unmasked) > 0:
+                tensordict_out = tensordict_out.reshape(
+                    self.frames_per_batch * self.n_env, *self.env_batch_size_unmasked
+                )
+            else:
+                tensordict_out = tensordict_out.view(-1).to_tensordict()
```
matteobettini, line 596: I order to move all batch dimensions to the first, I use first a permute and then a reshape. 
This makes it impossible to `return_same_td`.
Is there a workaround to this returning the same td? Or what can we do?

vmoens:
> We should have collectors working with both split_traj options.

Le ultime parole famose 
Let me know how I can help

matteobettini:
> > We should have collectors working with both split_traj options.
> 
> Le ultime parole famose Let me know how I can help

Dovremmo esserci, sto finendo di scrivere l'ultimo test ma la feature è finita

matteobettini:
Ready for review

vmoens:
We still have some errors in split_trajs, do you want me to jump in and have a look?

matteobettini:
Oh I didn't see those, I was blindly only using the tests in `test_collectors` I'll have a look and remark this as draft

matteobettini:
Also attempts to fix #846 with the warning solution

vmoens:
> Also attempts to fix #846 with the warning solution

Ok, then I would re-open the issue (or keep it open) to make this warning an error. WDYT?

matteobettini:
This PR changes also the way frames are counted after collection.

Currently, in the codebase frames are counted either as `mask.sum()` or `batch.numel()`.

While these counting mechanisms will still work for enviornments with non set `mask_env_batch_size` (aka all of the current examples in the code apart from the tests introduced here), they will not give the right number of frames when the mask is set (aka multiagent and so on).

With this PR I have made so that there are new ways to get the number of collected frames.
1. `collector.frames_per_batch` will always return the **exact** number of frames collected (even when the warning is triggered)
2. `batch.batch_size[0]` will be equal to `collector.frames_per_batch` when the `split_traj=False`
3. when `split_traj=True` i still have to think of a simple way to do it without having access to the collector object

vmoens:
Hi @matteobettini 
Sorry to keep waiting with this.
I'm working on some improvements for the collectors, and to avoid multiplying the work on both ends I was thinking to complete that before merging your branch. I hope it's ok!

matteobettini:
yep no probs!

torchrl/collectors/collectors.py
```diff
@@ -293,6 +298,11 @@ class SyncDataCollector(_DataCollector):
             updated. This feature should be used cautiously: if the same tensordict is added to a replay buffer for instance,
             the whole content of the buffer will be identical.
             Default is False.
+        mask_env_batch_size (Sequence[bool], optional): a sequence of bools of the same length as env.batch_size,
+            with a value of True it indicates to consider the corresponding dimension of env.batch_size as part of the
+            batch of environments used to collect frames. A value of False it indicates NOT to consider that dimension
+            as part of the batch of environments used to collect frames (used for agent dimension in multi-agent settings).
+            Default is None (corresponding to all True).
```
vmoens, line 305:
```suggestion
        mask_env_batch_size (Sequence[bool], optional): a sequence of bools of the same length as env.batch_size.
            A value of ``True`` indicates that the corresponding dimension of ``env.batch_size`` is to be included in the computation of the number of frames collected. A value of ``False`` indicates NOT to consider this particular dimension
            as part of the batch of environments used to collect frames (e.g. used for agent dimension in multi-agent settings).
            Defaults to ``True`` for all dims.
```

@matteobettini 
Two questions here:
- What if we provide envs with a different batch size? (It should break when putting the tensordicts together in the sync case but it _should_ work in the async case)...
- What to do with non batch-locked envs (ie envs that are purely stateless and just read and write in a tensordict). Should we consider this case too, and constrain the `mask_env_batch_size` to have a length that match the batch dimensions of the tensordicts?

torchrl/collectors/collectors.py
```diff
@@ -813,6 +911,11 @@ class _MultiDataCollector(_DataCollector):
             in other words, if the env is a multi-agent env, all agents will be
             reset once one of them is done.
             Defaults to `True`.
+        mask_env_batch_size (Sequence[bool], optional): a sequence of bools of the same length as env.batch_size,
+            with a value of True it indicates to consider the corresponding dimension of env.batch_size as part of the
+            batch of environments used to collect frames. A value of False it indicates NOT to consider that dimension
+            as part of the batch of environments used to collect frames (used for agent dimension in multi-agent settings).
+            Default is None (corresponding to all True).
```
vmoens, line 918:
```suggestion
        mask_env_batch_size (Sequence[bool], optional): a sequence of bools of the same length as env.batch_size.
            A value of ``True`` indicates that the corresponding dimension of ``env.batch_size`` is to be included in the computation of the number of frames collected. A value of ``False`` indicates NOT to consider this particular dimension
            as part of the batch of environments used to collect frames (e.g. used for agent dimension in multi-agent settings).
            Defaults to ``True`` for all dims.
```

torchrl/collectors/collectors.py
```diff
@@ -1515,6 +1627,11 @@ class aSyncDataCollector(MultiaSyncDataCollector):
         init_with_lag (bool, optional): if True, the first trajectory will be truncated earlier at a random step.
             This is helpful to desynchronize the environments, such that steps do no match in all collected rollouts.
             default = True
+        mask_env_batch_size (Sequence[bool], optional): a sequence of bools of the same length as env.batch_size,
+            with a value of True it indicates to consider the corresponding dimension of env.batch_size as part of the
+            batch of environments used to collect frames. A value of False it indicates NOT to consider that dimension
+            as part of the batch of environments used to collect frames (used for agent dimension in multi-agent settings).
+            Default is None (corresponding to all True).
```
vmoens, line 1634:
```suggestion
        mask_env_batch_size (Sequence[bool], optional): a sequence of bools of the same length as env.batch_size.
            A value of ``True`` indicates that the corresponding dimension of ``env.batch_size`` is to be included in the computation of the number of frames collected. A value of ``False`` indicates NOT to consider this particular dimension
            as part of the batch of environments used to collect frames (e.g. used for agent dimension in multi-agent settings).
            Defaults to ``True`` for all dims.
```

torchrl/collectors/utils.py
```diff
@@ -25,49 +25,85 @@ def stacked_output_fun(*args, **kwargs):
...
+        ]
+    )
+
+
+def bring_forward_and_squash_batch_sizes(
+    tensordict: TensorDictBase,
+    permute: Sequence[int],
+    batch_size_unmasked: Sequence[int],
+) -> TensorDictBase:
+    """Permutes the batch dimesnions attording to the permute indeces and then squashes all leadning dimesnions apart from batch_size_unmasked."""
```
vmoens, line 104:
```suggestion
    """Permutes the batch dimensions according to the permute indices, then squashes all the leading dimensions except ``batch_size_unmasked``."""
```

torchrl/collectors/collectors.py
matteobettini: > @matteobettini Two questions here:
> 
> * What if we provide envs with a different batch size? (It should break when putting the tensordicts together in the sync case but it _should_ work in the async case)...

The sync case will try to cat them so i think it would break. The async case will just return the one that finishes first i guess?

> * What to do with non batch-locked envs (ie envs that are purely stateless and just read and write in a tensordict). Should we consider this case too, and constrain the `mask_env_batch_size` to have a length that match the batch dimensions of the tensordicts?

In this case the mask will be checked to match env.batch_size. If during execution the batch size changes the mask will not work I guess. We could also uese a mask which is just a list of dimensions to exclude from being considered as part of the batch, something like
```python
batch_excluded_dims=[0,2]
```
Which would work as long as those dimension are part of the batch_size of the non-batch locked env.

PR #833, opened 2023-01-14
Updating Recorder to accomodate "solved" key

ShahRutav:
## Description

Removed the `value.any()` for `solved` key in Recorder

## Motivation and Context

In `mj_envs`,  `solved` key is returned at every step that indicates whether the task has reached the goal. Adding `value.any()` returns a single indicator for all the trajectories. An example use case of `solved` key is [here](https://github.com/vikashplus/mj_envs/blob/77274c05f085d052b7efb65fab9cb15269281f9b/mj_envs/envs/env_base.py#L356)

## Types of changes

## Checklist


vmoens (approved):
LGTM

PR #835, opened 2023-01-15
[BugFIx] Changed "set_count" set in collectors

matteobettini:
## Description

 This one line PR modifies the way the step_count is set in collectors.

With the current
```
step_count += 1
```
The step_count is not set in the case the get method returns a cloned object, which happens if 

```
self._tensordict = env.reset() 
```
Is created from a _reset method that returns LazyStacks

### DIFF
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -22,6 +22,7 @@ from tensordict.nn import TensorDictModule
 from tensordict.tensordict import TensorDict, TensorDictBase
 from torch import multiprocessing as mp
 from torch.utils.data import IterableDataset
+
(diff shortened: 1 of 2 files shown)


torchrl/collectors/collectors.py
```diff
@@ -655,7 +656,7 @@ def rollout(self) -> TensorDictBase:
                     self._tensordict = self.env.step(self._tensordict)
 
                 step_count = self._tensordict.get("step_count")
-                step_count += 1
+                self._tensordict.set_("step_count", step_count + 1)
```
vmoens, line 659: Good catch!

vmoens (approved):
LGTM

PR #850, opened 2023-01-19
[Feature] Vmas to device

matteobettini:
## Description

This PR allows vmas enviornments to be moved from one device to the other when a user calls `env.to(device)` 

In addition, now vmas returns an exact "done" flag after partial reset instead of using the default.

Tests are added to test this feature

and bonus tests for the correct use of "_reset" in vmas


vmoens (approved):
LGTM thanks for this!
Does this change the backend for VMAS (if anything is executed on device)?
For instance, with Habitat it _does_ change the backend, but I don't think it's the case for Brax

matteobettini:
> LGTM thanks for this! Does this change the backend for VMAS (if anything is executed on device)? For instance, with Habitat it _does_ change the backend, but I don't think it's the case for Brax

Yes, it moves the whole simualtion backend among devices and this can be done at any point during execution

PR #868, opened 2023-01-25
[BugFix] Avoid collision of "step_count" key from transform and collector

vmoens:
## Description

`StepCount` and the collectors both assign a `"step_count"` key that it would be wise to use only once.
This PR solves this issue.

cc @riiswa

### DIFF
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -522,8 +522,6 @@ class EnvBase(nn.Module, metaclass=abc.ABCMeta):
                 f"Env {self} was done after reset on specified '_reset' dimensions. This is (currently) not allowed."
             )
         if tensordict is not None:
-            if "_reset" in tensordict.keys():
-                tensordict.del_("_reset")
             tensordict.update(tensordict_reset)
         else:
             tensordict = tensordict_reset
--- a/examples/td3/td3.py
+++ b/examples/td3/td3.py
@@ -273,10 +273,10 @@ def main(cfg: "DictConfig"):  # noqa: F821
         pbar.update(tensordict.numel())
 
         # extend the replay buffer with the new data
-        if "mask" in tensordict.keys():
+        if ("collector", "mask") in tensordict.keys(True):
             # if multi-step, a mask is present to help filter padded values
-            current_frames = tensordict["mask"].sum()
-            tensordict = tensordict[tensordict.get("mask").squeeze(-1)]
+            current_frames = tensordict["collector", "mask"].sum()
+            tensordict = tensordict[tensordict.get(("collector", "mask")).squeeze(-1)]
         else:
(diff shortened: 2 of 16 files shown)


torchrl/envs/common.py
```diff
@@ -522,8 +522,6 @@ def reset(
                 f"Env {self} was done after reset on specified '_reset' dimensions. This is (currently) not allowed."
             )
         if tensordict is not None:
-            if "_reset" in tensordict.keys():
-                tensordict.del_("_reset")
```
vmoens, line 526: @matteobettini do you recall why we need to remove _reset from the td?
The problem is that we call `reset()` in `TransformedEnv` and we need `"_reset"` to be passed to the transforms `reset()` methods. Deleting it make them consider that `"_reset"` is all `True` by default

matteobettini: The idea there is to remove it because if users do not notice it is there and call reset again with the same tensordict it will still only use those key and they might not notice. So i though of making it a consumable that once reaches the inner env is removed.

The problem is that right now the transform reset is called after the env reset. So this consumable logic consumes the _reset flag before reacheing the transforms. There is no workaroud to this right?

vmoens: I might be wrong (please correct me if I am) but isn't is actually useful to keep track of that?

It is also dangerous to keep the "action" entry or anything else FWIW. The real nasty thing is to control `step_mdp` such that its default behaviour does not pass useless info to the next step.

matteobettini: Yes, at the time i made it seemed less error prone to consume it, but we can treat it as we treat actions. And now I realize that transforms reset is called after the env one and they may need it.

matteobettini:
This PR will affect mostly traj_ids right? Aren't we planning to remove the collector step count?

vmoens:
> This PR will affect mostly traj_ids in the long term right? Aren't we planning to remove the collector step count?

yep that should be the plan.
While it's here, let's put it somewhere it doesn't bother anyone.

matteobettini:
Yep, then maybe also the "mask" key should go in the nested collector key

vmoens:
> Yep, then maybe also the "mask" key should go in the nested collector key

Good point

torchrl/envs/common.py
vmoens: If we integrate it, maybe we can remove the underscore. Or leave it (in which case the collector will discard it by default)

matteobettini: I would probably leave it as the private meaning is nice. Collectors will discard it automatically yes but they will also have to explicitly discard it here https://github.com/pytorch/rl/blob/main/torchrl/collectors/collectors.py#L620 i think, otherwise it will be carried to the next step

PR #882, opened 2023-01-30
[Feature] Discrete SAC

BY571:
## Description

Adding a discrete SAC example 

## Motivation and Context

Current SAC implementation only supports continuous action spaces. This PR will add the option to run a discrete SAC example based on the [paper](https://arxiv.org/abs/1910.07207).

Convergence proof tested on CartPole-v1 ([wandb](https://wandb.ai/sebastian-dittert/rl-examples_discrete_sac?workspace=user-sebastian-dittert))
image

## Types of changes

## Checklist


torchrl/objectives/discrete_sac.py
```diff
@@ -0,0 +1,311 @@
...
+except ImportError as err:
+    FUNCTORCH_ERR = str(err)
+    _has_functorch = False
+
+
+class DiscreteSACLoss(LossModule):
+    """Discrete SAC Loss module.
+
+    Args:
+        actor_network (SafeModule): the actor to be trained
```
vmoens, line 40: I think we can just assume this to be a `TensorDictModule`

torchrl/objectives/discrete_sac.py
```diff
@@ -0,0 +1,311 @@
...
+        qvalue_network (SafeModule): a single Q-value network that will be multiplicated as many times as needed.
```
vmoens, line 41: same here

vmoens (commented):
LGTM
Before landing:
- can we move the loss to sac.py? I'd rather have them all in the same place if that makes sense?
- Can we add the loss to the doc?
- Is this supposed to work with gSDE? gSDE is not tailored for discrete action spaces AFAICT

BY571:
> LGTM Before landing:
> 
> * can we move the loss to sac.py? I'd rather have them all in the same place if that makes sense?
> * Can we add the loss to the doc?
> * Is this supposed to work with gSDE? gSDE is not tailored for discrete action spaces AFAICT

Do you mean discrete and continuous sac loss in one objective class or having both losses just in the same file?
I'd prefer to have them in the same class, what do you think? Will have a look at it in the coming days.

Will add it to the doc and also take off the gSDE :)

vmoens:
> I'd prefer to have them in the same class
How much control flow would that entail?
Does it save a lot of code?

I was thinking of having them in the same file. If having them in the same class does not create a monster class I'm happy to consider it.

Will it work with v1 and v2?

BY571:
> I was thinking of having them in the same file. If having them in the same class does not create a monster class I'm happy to consider it.
> 
> Will it work with v1 and v2?

For now, I just added it to the `sac.py` file in objectives. As it only works with v2 it might get messy and as you said would probably create a monster class. Let me know what you think.

I also took off the gSDE from the loss and updated the description of the actor_network to be a `TensorDictModule`.

How can I update the docs?

vmoens (approved):
LGTM

vmoens (approved):
LGTM

torchrl/objectives/sac.py
```diff
@@ -413,3 +414,269 @@ def _alpha(self):
...
+            )
+        else:
+            self.register_parameter(
+                "log_alpha",
+                torch.nn.Parameter(torch.tensor(math.log(alpha_init), device=device)),
+            )
+
+        if target_entropy == "auto":
+            target_entropy = -float(
+                np.log(1.0 / action_spec["action"].shape[0]) * target_entropy_weight
```
vmoens, line 511: careful here: the `[0]` can be the batch size
maybe the last dimension? Or since it's discrete we can check if it's a one hot or a discrete encoding and directly retrieve the number of options from the spec?

vmoens: I think it's the only place where we use `action_spec`
Maybe we could just pass the number of possible actions rather than passing the spec?

BY571: Fixed that and adapted the example script due to recent TorchRL changes. 
However, now some tests fail but I'm at it and hope to resolve them quickly!

PR #893, opened 2023-02-03
[Bugfix] Codecov does not cover multiprocessed tests #879

### LINKED ISSUE #879: [BUG] Codecov does not cover multiprocessed tests
## Describe the bug

Our codecov coverage misrepresents the true coverage of the lib, with the collectors.py being covered at only 33% and vec_envs at 42%.
This can be fixed by instructing it to look at the multiprocess runs:
https://stackoverflow.com/a/36863654

kadeng:
## Description

Fix for #879 - I could verify that the coverage script used in circleci does not properly collect coverage data when multiprocessing is used. In order to fix this, it. was neccessary to change the way the coverage script is invoked. While coverage.py supports collecting coverage information even if the covered process launches subprocesses,it is not supported to pass command line arguments to the script in that case ( since subprocesses don't get that commandline ).

I created a helper script called coverage_run_parallel.py which writes a temporary config file for coverage.py ( cleaned up on exit, including abnormal exit ) and invokes that script in turn, passing the info to use that config via environment variables.

Test Plan:

Running the code coverage scripts locally ( on an M1 Mac without CUDA ), and comparing the scenario before- and after this change, it can be clearly seen that the code coverage increases in some files. For example, coverage of torchrl/collectors/collectors.py increases from 68% to 81%.

Before merging, the CI scripts need to pass and the coverage results should be inspected as part of the review of this PR, given that I could not do a full CI run on my own fork or locally on my machine.

## Motivation and Context

close #879

## Types of changes

- [ X ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ X ] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
- [ X ] I have updated the tests accordingly (*required for a bug fix or a new feature*).

### DIFF
--- a/.circleci/unittest/helpers/coverage_run_parallel.py
+++ b/.circleci/unittest/helpers/coverage_run_parallel.py
@@ -1,0 +1,64 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+When coverage run is launched with --concurrency=multiprocessing, it
+needs a config file for further arguments.
+
+This script is a drop-in wrapper to conveniently pass command line arguments
+nevertheless. It writes temporary coverage config files on the fly and
+invokes coverage with proper arguments
+"""
+
+import os
+import shlex
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+from typing import List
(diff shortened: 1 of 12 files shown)


vmoens (approved):
LGTM thanks for this!

Issue #902, opened 2023-02-09
[BUG] Cannot initialize the transform if parent env is not defined


anthony0727:
## Describe the bug
 
running example throws `Cannot initialize the transform if parent env is not defined`

## To Reproduce

```python dreamer/dreamer.py```

```bash
(py39) {my machine}$ python examples/dreamer/dreamer.py env_name="walker" env_task="walk" env_library="dm_control"
EGL_DEVICE_ID:  1
sys:1: UserWarning: 
'config' is validated against ConfigStore schema with the same name.
This behavior is deprecated in Hydra 1.1 and will be removed in Hydra 1.2.
See https://hydra.cc/docs/1.2/upgrades/1.0_to_1.1/automatic_schema_matching for migration instructions.
/home/anthony/.pyenv/versions/py39/lib/python3.9/site-packages/hydra/main.py:90: UserWarning: 
'config' is validated against ConfigStore schema with the same name.
This behavior is deprecated in Hydra 1.1 and will be removed in Hydra 1.2.
See https://hydra.cc/docs/1.2/upgrades/1.0_to_1.1/automatic_schema_matching for migration instructions.
  _run_hydra(
Using device cuda:0
Error executing job with overrides: ['env_name=walker', 'env_task=walk', 'env_library=dm_control']
Traceback (most recent call last):
  File "/home/anthony/rl/examples/dreamer/dreamer.py", line 111, in main
    initialize_observation_norm_transforms(
  File "/home/anthony/rl/torchrl/trainers/helpers/envs.py", line 485, in initialize_observation_norm_transforms
    transform.init_stats(num_iter=num_iter, key=key)
  File "/home/anthony/rl/torchrl/envs/transforms/transforms.py", line 1508, in init_stats
    raise RuntimeError(
RuntimeError: Cannot initialize the transform if parent env is not defined. 
```

## Expected behavior

example running

## System info

Describe the characteristic of your environment:
0.0.4 1.23.5 3.9.16 (main, Jan 17 2023, 03:16:52) 
[GCC 9.4.0] linux

## Reason and Possible fixes

If you know or suspect the reason for this bug, paste the code lines and suggest modifications.

## Checklist

- [*] I have checked that there is no similar issue in the repo (**required**)
- [*] I have read the [documentation](https://github.com/pytorch/rl/tree/main/docs/) (**required**)
- [*] I have provided a minimal working example to reproduce the bug (**required**)

vmoens:
Thanks for reporting
[...]

PR #903, opened 2023-02-09
[BugFix] Dreamer helpers are broken with batched envs

vmoens:
## Description

Some helper functions in the examples (specifically dreamer) have not been adapted to the new env batch_size convention.

We also drop the convention that vector states are named "observation" for gym envs when they're no accompanied by a "pixels" key and "state" otherwise. Now "observation" always prevails. This may be surprising for users that expect the keys to match those of the gym env but this is even more confusing when switching from env with and without rendering.

We also simplify the recorder construction and prevent non-batched parallel transforms + fix a bug with Noops when parallel envs are used.


vmoens:
Fix #902

PR #930, opened 2023-02-21
[Feature] Distributed data collector (ray)

albertbou92:
## Description

new class DistributedCollector, which allows to do distributed data collection with TorchRL collectors using Ray.

@vmoens This approach required adding a new function to the collectors, one that takes a single step of the iterator. I have done it with SyncDataCollector. Do you think it is fine?  

The PR also includes an example of distributed data collection for the SyncDataCollector case.

## Types of changes

## Checklist


torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+        resources=resources)(cls)
+    c.is_remote = True
+    return c
+
+
+class DistributedCollector(IterableDataset, ABC):
+    """
+    Class to better handle the operations of ensembles of Collectors.
+
+    Contains common functionality across all collectors sets.
```
vmoens, line 77: This is not very explicit.
We should say 
"Like other collectors, this collector is an iterable that yields TensorDict instances with the appropriate number of frames."

We should say what is and what is not supported.

We should say if it's sync, async or depends on the input to the constructor.

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+    remote_config : dict
+        Ray resource specs for the remote collectors.
+    num_collectors: int
+        Total number of collectors in the set (including local collector)
+    """
+
+    def __init__(self,
+                 collector_class,
+                 collector_params,
+                 remote_config=default_remote_config,  # TODO: mutable type as input, not correct
```
vmoens, line 94: the proper way of doing this is 
```suggestion
                 remote_config=None,
...
    if remote_config is None:
        remote_config = default_remote_config
```

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+                 num_collectors=1,
+                 total_frames=1000,  # TODO: is this parameter necessary? it is already specified in each collector.
+                 communication="sync",  # "sync" or "async"
+                 ):
+
+        if not _has_ray:
+            raise RuntimeError(
+                f"ray library not found, unable to create a DistributedCollector. "
+            )
```
vmoens, line 103: add `from RAY_ERR`

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+        num_cpus=num_cpus,
+        num_gpus=num_gpus,
+        memory=memory,
+        object_store_memory=object_store_memory,
+        resources=resources)(cls)
+    c.is_remote = True
+    return c
+
+
+class DistributedCollector(IterableDataset, ABC):
```
vmoens, line 73: Maybe `RayDistributedCollector`

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+        if not _has_ray:
+            raise RuntimeError(
+                f"ray library not found, unable to create a DistributedCollector. "
+            )
+
+        if communication not in ("sync", "async"):
+            raise ValueError(f"Communication parameter in CollectorSet has to be sync or async.")
+
+        # Monkey patching as_remote and print_remote_collector_info to collector class # TODO: is that ok ?
+        collector_class.as_remote = as_remote
```
vmoens, line 109: I think this is ok as long as it works :)

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+
+        self.collected_frames = 0
+        self.total_frames = total_frames
+        self.collector_class = collector_class
+        self.collector_params = collector_params
+        self.num_collectors = num_collectors
+        self.remote_config = remote_config
+        self.communication = communication
+
+        # Create a local instance of the collector class.
```
vmoens, line 120: are we sure we're not actually collecting any data? Is there a smarter way of doing this?

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+        return w
+
+    def add_collectors(self, num_collectors, collector_params):
+        """Create and add a number of remote collectors to the set."""
+        cls = self.collector_class.as_remote(**self.remote_config).remote
+        self._remote_collectors.extend(
+            [self._make_collector(cls, collector_params) for _ in range(num_collectors)])
+
+    def local_collector(self):
+        """Return local collector"""
```
vmoens, line 147:
```suggestion
        """Returns local collector."""
```

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+        ray.wait(pending_samples)
+
+    @staticmethod
+    def _make_collector(cls, collector_params):
+        """Create a single collector instance."""
+        w = cls(**collector_params)
+        return w
+
+    def add_collectors(self, num_collectors, collector_params):
+        """Create and add a number of remote collectors to the set."""
```
vmoens, line 141:
```suggestion
        """Creates and adds a number of remote collectors to the set."""
```

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+        cls = self.collector_class.as_remote(**self.remote_config).remote
+        self._remote_collectors.extend(
+            [self._make_collector(cls, collector_params) for _ in range(num_collectors)])
+
+    def local_collector(self):
+        """Return local collector"""
+        return self._local_collector
+
+    def remote_collectors(self):
+        """Returns list of remote collectors"""
```
vmoens, line 151:
```suggestion
        """Returns list of remote collectors."""
```

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+        return self._remote_collectors
+
+    def stop(self):
+        """Stop all remote collectors"""
```
vmoens, line 155:
```suggestion
        """Stops all remote collectors."""
```

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+            return self.async_iterator()
+
+    def sync_iterator(self) -> Iterator[TensorDictBase]:
+
+        while self.collected_frames < self.total_frames:
+
+            # Broadcast agent weights
+            self.local_collector().update_policy_weights_()
+            state_dict = {"policy_state_dict": self._local_collector.policy.state_dict()}
+            state_dict = ray.put(state_dict)
```
vmoens, line 172: Q: in regular data collectors we leave this to the user or ask explicitly if the update should be done at each iter.
Do we want to do the same here? ie isolating the weight updating in a dedicated method that can or won't be called at each iteration.

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+            for e in self.remote_collectors():
```
vmoens, line 173: let's use a more explicit variable name

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+                e.load_state_dict.remote(**{"state_dict": state_dict, "strict": False})
+
+            # Ask for batches to all remote workers.
+            pending_samples = [e._iterator_step.remote() for e in self.remote_collectors()]
+
+            # Wait for all rollouts
+            samples_ready = []
+            while len(samples_ready) < self.num_collectors:
+                samples_ready, samples_not_ready = ray.wait(
+                    pending_samples, num_returns=len(pending_samples), timeout=0.001)
```
vmoens, line 183: what's the role of timeout here? Is the value standard? Should it be a hyperparam or a global variable?

torchrl/collectors/distributed/distributed_collector.py
```diff
@@ -0,0 +1,232 @@
...
+            self.collected_frames += out_td.numel()
+
+            # Update agent weights
+            self.local_collector().update_policy_weights_()
+            state_dict = {"policy_state_dict": self._local_collector.policy.state_dict()}
+            state_dict = ray.put(state_dict)
+            w.load_state_dict.remote(**{"state_dict": state_dict, "strict": False})
+
+            # Schedule a new collection task
+            future = w._iterator_step.remote()
```
vmoens, line 229: when we reach the end of the loop, we'll still have workers waiting to deliver their data right? Isn't there a risk that they will stay idle forever?

torchrl/collectors/distributed/distributed_collector.py
albertbou92: If timeout is set, the function returns either when the requested number of IDs are ready or when the timeout is reached, whichever occurs first. If it is not set, the function simply waits until that number of objects is ready and returns that exact number of object refs.

Actually for not we can just set timeout=None, but if we want to implement the  preemption mechanism, we will need to check regularly how many of the tasks have ended. We can either set it to a reasonable value of allow the users to define it.

torchrl/collectors/distributed/distributed_collector.py
albertbou92: Yes, I will add some code to make sure remaining tasks are cancelled or waited out.

torchrl/collectors/distributed/distributed_collector.py
albertbou92: yes, we can just provide policy as an input parameter and assume policy will have the latest version of the weights at any moment.

torchrl/collectors/distributed/ray_collector.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def set_seed(self, seed: Union[int, List[int]], static_seed: bool = False) -> List[int]:
+        """Calls set_seed(seed, static_seed) method for each remote collector and results a list of results."""
+        if isinstance(seed, int):
+            seed = [seed]
+        if len(seed) == 1:
+            seed = seed * len(self.remote_collectors())
+        futures = [collector.set_seed.remote(s, static_seed) for s, collector in zip(seed, self.remote_collectors())]
+        results = ray.get(object_refs=futures)
+        return results
```
vmoens, line 298: we should set the seed iteratively
In torchrl seeding is done via a generator that gives you the next seed to use when you need to set it multiple times
Like this you can set different seeds across workers whilst keeping the "deterministic" sequence and have seeds that don't overlap (since the alternative would be seed=0 for worker 0, seed=1 for worker 1 etc. but seeding like this will cause some overlap when you seed with seed=1 for worker 0).
See `seed_generator` for more info.
In practice you should:
- seed worker 0, get the resulting seed
- seed worker 1 with the resulting seed, get the resulting seed
- etc

torchrl/collectors/distributed/ray_collector.py
```diff
@@ -0,0 +1,321 @@
...
+
+            # Retrieve single rollouts
+            out_td = ray.get(future)
+            ray.internal.free(future)  # should not be necessary, deleted automatically when ref count is down to 0
+            self.collected_frames += out_td.numel()
+
+            # Update agent weights
+            policy_weights_local_collector = {"policy_state_dict": self.local_policy.state_dict()}
+            policy_weights_local_collector_ref = ray.put(policy_weights_local_collector)
+            w.load_state_dict.remote(**{"state_dict": policy_weights_local_collector_ref, "strict": False})
```
vmoens, line 268: we should let the user take care of that OR have a param in the constructor that specifies how often we need to update it.
Also: i am working on a param in `update_policy_weights_()` to facilitate this, something like
```
collector.update_policy_weights_(new_weights)
```
but i'm not sure it makes sense.

IMO load_state_dict should only be used to deserialize things, it's a bit inefficient.

In practice what I would see is:

E.g.
```
policy_weights = TensorDict(dict(self.policy.named_parameters()), [])
w.update_policy_weights_(policy_weights)
```

once it's implemented, oc

torchrl/collectors/distributed/ray_collector.py
```diff
@@ -0,0 +1,321 @@
...
+
+            # Schedule a new collection task
+            future = w._iterator_step.remote()
+            pending_tasks[future] = w
+
+            yield out_td
```
vmoens, line 274: this should be moved before the update of the policy weights no?
Like this first the user receives the data, then the weights are updated and things keep on running

torchrl/collectors/distributed/ray_collector.py
```diff
@@ -0,0 +1,321 @@
...
+            # collector.__ray_terminate__.remote()  # This will kill the actor but let pending tasks finish
+            ray.kill(collector)  # This will interrupt any running tasks on the actor, causing them to fail immediately
+
+    def iterator(self):
+        if self.coordination == "sync":
+            return self._sync_iterator()
+        else:
+            return self._async_iterator()
+
+    def _sync_iterator(self) -> Iterator[TensorDictBase]:
```
vmoens, line 206: where do we update the weights here?

vmoens (commented):
now we need some tests 😬

torchrl/collectors/distributed/ray_collector.py
```diff
@@ -0,0 +1,321 @@
...
+        if isinstance(state_dict, OrderedDict):
+            state_dict = [state_dict]
+        if len(state_dict) == 1:
+            state_dict = state_dict * len(self.remote_collectors())
+        for collector, state_dict in zip(self.remote_collectors(), state_dict):
+            collector.load_state_dict.remote(state_dict)
+
+    def shutdown(self):
+        """Finishes processes started by ray.init()."""
+        ray.shutdown()
```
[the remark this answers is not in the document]
albertbou92: yup :)

torchrl/collectors/distributed/ray_collector.py
albertbou92: here:
https://github.com/pytorch/rl/blob/b3a8ec71eeeaf00998ca68ddb9216063a1dd2a1a/torchrl/collectors/distributed/ray_collector.py#L209

torchrl/collectors/distributed/ray_collector.py
albertbou92: makes sense

torchrl/collectors/collectors.py
```diff
@@ -527,6 +527,33 @@ def set_seed(self, seed: int, static_seed: bool = False) -> int:
         """
         return self.env.set_seed(seed, static_seed=static_seed)
 
+    def _iterator_step(self) -> TensorDictBase:
```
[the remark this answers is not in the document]
albertbou92: we can use the next function

torchrl/collectors/distributed/ray_collector.py
albertbou92: we can use the iterative approach

vmoens (commented):
Can we align this with the other distributed collectors?

torchrl/collectors/distributed/ray_collector.py
```diff
@@ -0,0 +1,562 @@
...
+    Returns:
+        A function that creates ray remote class instances.
+    """
+    remote_collector = ray.remote(**remote_config)(cls)
+    remote_collector.is_remote = True
+    return remote_collector
+
+
+class RayDistributedCollector(_DataCollector):
+    """Distributed data collector with Ray (https://docs.ray.io/) backend.
```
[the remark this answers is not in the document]
albertbou92: ok!

torchrl/collectors/distributed/ray.py
```diff
@@ -0,0 +1,637 @@
...
+    ray.remote() when called to create each remote collector actor, including collector compute
+    resources.The sum of all collector resources should be available in the cluster. Refer to Ray
+    documentation for advanced configuration of the ray.remote() method. Default kwargs are:
+
+    {
+        "num_cpus": 1,
+        "num_gpus": 0.2,
+        "memory": 2 * 1024 ** 3,
+    }
+
```
vmoens, line 107: We could do
```suggestion
    >>> kwargs = {
    ...     "num_cpus": 1,
    ...     "num_gpus": 0.2,
    ...     "memory": 2 * 1024 ** 3,
    ... }

```

vmoens:
Sorry for the late reply

> @vmoens This approach required adding a new function to the collectors, one that takes a single step of the iterator. I have done it with SyncDataCollector. Do you think it is fine?

I can't spot it, what is it that you changed?
We have a `collector.next()` method now if that's what you need.

Also why aren't the test following the same pattern than other distributed collectors? I think it's fine if there is a good reason.

vmoens (approved):
LGTM, see my comment in the PR

examples/distributed/collectors/multi_nodes/ray_train.py
vmoens, line 1: Lovely! You went the extra length with this!

vmoens (approved):
Awesome!

PR #934, opened 2023-02-22
[Feature] `torch.distributed` collectors

vmoens:
Distributed data collectors using torch.distributed and rpc

Features:
- `torch.distributed`:
  - sync vs async
  - nccl and gloo tested
  - single and multi-node (mp vs submitit)
  - submitit when jobs cannot launch sub-jobs (eg FAIR cluster)
- RPC
  - sync vs async
  - single and multi node

Check the examples!

TODO:
- [x] RPC: pass tensorpipe options
- [ ] RPC: review the device maps: explicit maps passed to the collector?
- [x] Better match between collector kwargs in mp and distributed settings
- [x] Should we keep the device vs devices in mp collectors?
- [x] Document that using MultiSyncDataCollector will work faster than ParallelEnv bc of higher IO. Also document that higher number of frames per batch is usually faster
- [ ] Benchmark table?
- [x] Document the feature in the README + links to examples
- [x] Keep the sync collector? Simple and efficient, but maybe too much overlap with the generic one.
- [x] remove all prints and turn the verbose off
- [ ] Give example in the doc, maybe a GIF


torchrl/collectors/distributed/distributed_submitit.py
```diff
@@ -0,0 +1,281 @@
...
+    )
+    print("waiting...")
+    rpc.shutdown()
+
+
+class DistributedDataCollector(_DataCollector):
+    def __init__(
+        self,
+        env_makers,
+        policy,
```
albertbou92, line 74: won't policy by in collector_kwargs as well? is there any case in which it will be different?

vmoens:
@chowarfb @shagunsodhani 
This is WIP but if you want to have a look at it and give me some early feedback I'd be happy to read it :)

torchrl/collectors/distributed/generic.py
```diff
@@ -0,0 +1,831 @@
...
+from copy import copy, deepcopy
+from datetime import timedelta
+from typing import OrderedDict
+
+import torch.cuda
+from tensordict import TensorDict
+from torch import multiprocessing as mp, nn
+
+from torchrl.collectors import MultiaSyncDataCollector
+from torchrl.collectors.collectors import (
```
shagunsodhani, line 20: Could consider making `_DataCollector` public.

torchrl/collectors/distributed/generic.py
```diff
@@ -0,0 +1,831 @@
...
+            be executed, they should be dispatched across worker nodes, not
+            subnodes.
+        sync (bool, optional): if ``True``, the resulting tensordict is a stack of all the
+            tensordicts collected on each node. If ``False`` (default), each
+            tensordict results from a separate node in a "first-ready,
+            first-served" fashion.
+        slurm_kwargs (dict): a dictionary of parameters to be passed to the
+            submitit executor.
+        backend (str, optional): must a string "<distributed_backed>" where
+            <distributed_backed> is one of ``"gloo"``, ``"mpi"``, ``"nccl"`` or ``"ucc"``. See
```
shagunsodhani, line 318: Does thie default to gloo because it is mainly designed for CPU jobs ?

torchrl/collectors/distributed/utils.py
```diff
@@ -0,0 +1,150 @@
+import subprocess
+import time
+
+from torchrl.collectors.distributed.default_configs import (
+    DEFAULT_SLURM_CONF,
+    DEFAULT_SLURM_CONF_MAIN,
+    TCP_PORT,
+)
+from torchrl.collectors.distributed.generic import _distributed_init_delayed
+from torchrl.collectors.distributed.rpc import _rpc_init_collection_node
```
shagunsodhani, line 10: Consider making `_distributed_init_delayed` and `_rpc_init_collection_node` public

torchrl/collectors/distributed/generic.py
vmoens: not really
mostly bc I assume that users do not necessarily have a gpu with nccl at hand, and nccl can cause more bugs than gloo.

torchrl/collectors/distributed/generic.py
vmoens: Yep
`DataCollectorBase` would be nice!

PR #942, opened 2023-03-01
[BugFix] Vmas expanded specs

matteobettini:
## Description

After the specs gained a shape, some redundant extends where left in vmas, causing the specs to be potentially extended twice and leading to an error

This PR fixes that


vmoens (approved):
LGTM

PR #960, opened 2023-03-09
[Feature] Contiguous stacking of matching specs

tcbegley:
## Description

This PR changes the behaviour of `torch.stack` when the arguments are `TensorSpec` or `CompositeSpec` variants. It checks for equality among the specs being stacked, and if they all match, it actually performs the stack (rather than returning a lazy stacked spec).

To facilitate this, I have added `squeeze` and `unsqueeze` methods to all specs. Stacking is done by computing the new shape and then calling something like `spec.clone().unsqueeze(stack_dim).expand(new_shape)`. I added `squeeze` since you can't have `unsqueeze` without `squeeze` 😄 

Some of the tests probably need rethinking. I've made small changes to get them to pass, but there is new behaviour here which ideally would be tested for explicitly.

### DIFF
--- a/test/test_specs.py
+++ b/test/test_specs.py
@@ -2273,6 +2273,7 @@ class TestStackComposite:
         assert (r["a"] == 0).all()
 
     @pytest.mark.skipif(not torch.cuda.device_count(), reason="no cuda")
+    @pytest.mark.parametrize("stack_dim", [0, 1, 2, -3, -2, -1])
     def test_to(self, stack_dim):
         c1 = CompositeSpec(a=UnboundedContinuousTensorSpec(shape=(1, 3)), shape=(1, 3))
         c2 = c1.clone()
--- a/test/test_specs.py
+++ b/test/test_specs.py
@@ -18,7 +18,6 @@ from torchrl.data.tensor_specs import (
     CompositeSpec,
     DiscreteTensorSpec,
     LazyStackedCompositeSpec,
-    LazyStackedTensorSpec,
     MultiDiscreteTensorSpec,
     MultiOneHotDiscreteTensorSpec,
     OneHotDiscreteTensorSpec,
@@ -1716,7 +1715,7 @@ class TestStack:
         c1 = BinaryDiscreteTensorSpec(n=n, shape=shape)
         c2 = c1.clone()
         c = torch.stack([c1, c2], stack_dim)
-        assert isinstance(c, LazyStackedTensorSpec)
+        assert isinstance(c, BinaryDiscreteTensorSpec)
         shape = list(shape)
         if stack_dim < 0:
             stack_dim = len(shape) + stack_dim + 1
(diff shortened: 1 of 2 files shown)


test/test_specs.py
```diff
@@ -1716,7 +1715,7 @@ def test_stack_binarydiscrete(self, shape, stack_dim):
         c1 = BinaryDiscreteTensorSpec(n=n, shape=shape)
         c2 = c1.clone()
         c = torch.stack([c1, c2], stack_dim)
-        assert isinstance(c, LazyStackedTensorSpec)
+        assert isinstance(c, BinaryDiscreteTensorSpec)
```
tcbegley, line 1718: All of these tests where we no longer return a `LazyStackedTensorSpec` from `torch.stack` may need a second version that tests the lazy stacked behaviour?

test/test_specs.py
```diff
@@ -2146,36 +2151,14 @@ def test_stack_index_multdim(self, stack_dim):
...
-                    1,
-                    3,
-                    2,
-                ]
-            )
-            assert c_expand.shape == torch.Size([4, 1, 3, 2])
-            assert c_expand.dim == 3
-        else:
-            raise NotImplementedError
-
```
tcbegley, line 2174: I couldn't see how this test could be made to make sense anymore. Even if you instantiate `c` directly as a `LazyStackedCompositeSpec`

```python
c = LazyStackedCompositeSpec(c1, dim=stack_dim)
```

then because `.expand` uses `torch.stack` internally, you still wind up with a regular `CompositeSpec` which breaks the checks.

test/test_specs.py
```diff
@@ -2146,36 +2151,14 @@ def test_stack_index_multdim(self, stack_dim):
...
     @pytest.mark.parametrize("stack_dim", [0, 1, 2, -3, -2, -1])
     def test_stack_expand_multi(self, stack_dim):
         c1 = CompositeSpec(a=UnboundedContinuousTensorSpec(shape=(1, 3)), shape=(1, 3))
-        c2 = c1.clone()
+        c2 = CompositeSpec(
+            a=UnboundedContinuousTensorSpec(shape=(1, 3)),
+            b=UnboundedDiscreteTensorSpec(shape=(1, 3)),
+            shape=(1, 3),
+        )
```
tcbegley, line 2161: A bunch of tests like this I created a second `CompositeSpec` with another entry to force a lazy stack.

test/test_specs.py
vmoens: yep we should probably do that

torchrl/data/tensor_specs.py
```diff
@@ -837,6 +853,39 @@ def expand(self, *shape):
             n=shape[-1], shape=shape, device=self.device, dtype=self.dtype
         )
 
+    def squeeze(self, dim=None):
```
vmoens, line 856: we should overload torch.squeeze and unsqueeze now

vmoens (approved):
LGTM
We can merge this as is and add the missing tests later!

vmoens:
We have some failing tests due to the device changes

PR #963, opened 2023-03-13
[Feature] Multicollector interruptor

albertbou92:
## Description

This PR implements a preemptive mechanism to early stop stragglers in multi sync data collectors. The invalid data can be identified because its trajectory ids are -1.

For now, preemptive mechanism is not compatible with `split_trajs=True`, but this could be easily adapted by ignoring trajectory ids equal to -1 in the `split_trajectories` method.

i.e. 
splits = [(splits == i).sum().item() for i in splits.unique_consecutive() if i != -1] (changing line 39 in collectors/utils.py)
out_splits = rollout_tensordict.view(-1)[traj_ids != -1].split(splits, 0) (changing line 54 in collectors/utils.py)

## Motivation and Context

## Types of changes

## Checklist


torchrl/collectors/collectors.py
```diff
@@ -67,6 +68,44 @@ def __call__(self, td: TensorDictBase) -> TensorDictBase:
         return td.set("action", self.action_spec.rand())
 
 
+class Interruptor:
```
vmoens, line 71: if it's public it should be in the doc

torchrl/collectors/collectors.py
```diff
@@ -67,6 +68,44 @@ def __call__(self, td: TensorDictBase) -> TensorDictBase:
...
+    def stop_collection(self):
+        with self._lock:
+            self._collect = False
+
+    def collection_stopped(self):
+        with self._lock:
+            return self._collect is False
+
+
+class InterruptorManager(SyncManager):
```
vmoens, line 96: ditto

test/test_collector.py
```diff
@@ -1186,6 +1187,72 @@ def weight_reset(m):
...
+            reset_at_each_iter=False,
+            devices="cpu",
+            storing_devices="cpu",
+            split_trajs=False,
+            preemptive_threshold=0.25,
+        )
+
+        for batch in collector:
+            assert (
+                batch["collector"]["traj_ids"][
```
vmoens, line 1249: let's rewrite this in a more straightforward way, it's a bit hard to read
What if none is -1? Shouldn't we test that we have at least one traj_id set to -1 in the batch?

vmoens (commented):
LGTM I left some comments

test/test_collector.py
albertbou92: I simplified the code for more clarity.

Regarding the possibility of having none -1, I have addressed that as well by setting the preemptive threshold to 0.0. This way, all collectors will stop after the first iteration and only the very first frame of each sync collector will be valid. So we know for sure there are -1's in the batch.

vmoens:
Can you merge main in this branch, the tests are failing bc they're looking for a deprecated function in tensordict

vmoens (commented):
Can you have a look at how it would play out with the distributed collectors I integrated yesterday?
We don't necessarily need everything to be fully compatible but I want to make sure that we're not missing some obvious point, e.g. how we handle the traj-ids in the distributed collector may need a bit of refactoring.

(I will merge this sooner than that as the solution is neat and usable as of now)

torchrl/collectors/collectors.py
```diff
@@ -1399,6 +1462,18 @@ def iterator(self) -> Iterator[TensorDictBase]:
 
             i += 1
             max_traj_idx = None
+
+            if self.interruptor:
```
vmoens, line 1466: for clarity can we have `if self.interruptor is not None`?

PR #965, opened 2023-03-13
[Feature] Various improvements to LazyStacked specs

tcbegley:
## Description

This PR makes a few changes to lazy stacked specs:

- Remove unneeded placeholder methods in `LazyStackedTensorSpec`
- Implement the `device` property for `LazyStackedTensorSpec`
- Implement `to_numpy` for the two lazy stacked spec variants.

Would be great if you could check my understanding of what's required for `to_numpy` is correct.

### DIFF
--- a/test/test_specs.py
+++ b/test/test_specs.py
@@ -2045,7 +2045,7 @@ class TestStack:
         shape = (*shape,)
         c1 = UnboundedDiscreteTensorSpec(shape=shape)
         c2 = c1.clone()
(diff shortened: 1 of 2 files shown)


vmoens (commented):
LGTM!
maybe let's add a simple test for to_numpy()?

tcbegley:
I've added a couple of tests, I think this is good to go.

vmoens (approved):
LGTM

PR #981, opened 2023-03-21
[BugFix] make "_reset", "step_count", and other done_based keys follow done_spec

matteobettini:
There are many places where _reset is just intantieated using the batch_size, while in reality it needs to follow the done_spec.

In particular, in transforms, how do we enforce this?

### DIFF
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -689,7 +689,7 @@ class SyncDataCollector(DataCollectorBase):
 
         if not self.reset_when_done:
             done = torch.zeros_like(done)
-        done_or_terminated = done.squeeze(-1) | truncated.squeeze(-1)
+        done_or_terminated = done | truncated
         # keep track of envs that have been done at least once
         if self._has_been_done is None:
             self._has_been_done = done_or_terminated
@@ -699,8 +699,7 @@ class SyncDataCollector(DataCollectorBase):
             # collectors do not support passing other tensors than `"_reset"`
             # to `reset()`.
             if len(self.env.batch_size):
-                self._tensordict.masked_fill_(done_or_terminated, 0)
-                _reset = done_or_terminated.unsqueeze(-1)
+                _reset = done_or_terminated
                 td_reset = self._tensordict.select().set("_reset", _reset)
             else:
                 _reset = None
@@ -714,8 +713,11 @@ class SyncDataCollector(DataCollectorBase):
                 raise RuntimeError(
                     f"Env {self.env} was done after reset on specified '_reset' dimensions. This is (currently) not allowed."
                 )
-            traj_ids[done_or_terminated] = traj_ids.max() + torch.arange(
-                1, done_or_terminated.sum() + 1, device=traj_ids.device
+            traj_done_or_terminated = done_or_terminated.view(
+                self._tensordict.batch_size, -1
+            ).any(-1)
+            traj_ids[traj_done_or_terminated] = traj_ids.max() + torch.arange(
+                1, traj_done_or_terminated + 1, device=traj_ids.device
             )
             self._tensordict.set_(
                 ("collector", "traj_ids"), traj_ids
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -506,7 +506,11 @@ class SyncDataCollector(DataCollectorBase):
         self.reset_when_done = reset_when_done
         self.n_env = self.env.batch_size.numel()
 
-        (self.policy, self.device, self.get_weights_fn,) = self._get_policy_and_device(
+        (
+            self.policy,
+            self.device,
+            self.get_weights_fn,
+        ) = self._get_policy_and_device(
             policy=policy,
             device=device,
             observation_spec=self.env.observation_spec,
@@ -699,6 +703,9 @@ class SyncDataCollector(DataCollectorBase):
             # collectors do not support passing other tensors than `"_reset"`
             # to `reset()`.
             if len(self.env.batch_size):
+                self._tensordict.masked_fill_(
+                    done_or_terminated.view(self._tensordict.batch_size, -1).any(-1), 0
+                )
                 _reset = done_or_terminated
                 td_reset = self._tensordict.select().set("_reset", _reset)
             else:
@@ -714,7 +721,7 @@ class SyncDataCollector(DataCollectorBase):
                     f"Env {self.env} was done after reset on specified '_reset' dimensions. This is (currently) not allowed."
                 )
             traj_done_or_terminated = done_or_terminated.view(
-                self._tensordict.batch_size, -1
+                *self._tensordict.batch_size, -1
             ).any(-1)
             traj_ids[traj_done_or_terminated] = traj_ids.max() + torch.arange(
                 1, traj_done_or_terminated.sum() + 1, device=traj_ids.device
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -506,11 +506,7 @@ class SyncDataCollector(DataCollectorBase):
         self.reset_when_done = reset_when_done
         self.n_env = self.env.batch_size.numel()
 
-        (
-            self.policy,
-            self.device,
-            self.get_weights_fn,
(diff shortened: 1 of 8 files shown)


vmoens:
> There are many places where _reset is just intantieated using the batch_size, while in reality it needs to follow the done_spec.
> 
> In particular, in transforms, how do we enforce this?

you can do `transform.parent.done_spec` and get the `done_spec` as needed. Would that work?

torchrl/envs/transforms/transforms.py
```diff
@@ -3017,6 +3026,7 @@ def reset(self, tensordict: TensorDictBase) -> TensorDictBase:
                 dtype=torch.int64,
                 device=tensordict.device,
             )
+        _reset = _reset.view(tensordict.batch_size, -1).any(-1)
         step_count[_reset] = 0
```
matteobettini, line 3030: This way we are only counting steps within the batch_size.
This is cool but someone may want to count steps on the done.shape

matteobettini:
Init tracker will also create a key that follows done_spec.

torchrl/envs/common.py
```diff
@@ -708,7 +712,7 @@ def policy(td):
                 exclude_action=False,
             )
             if not break_when_any_done and done.any():
-                _reset = done.view(tensordict.shape)
+                _reset = done
```
vmoens, line 715: I guess we'd like to clone it to be 100% sure that any downstream modification won't have surprising effects

torchrl/envs/transforms/transforms.py
matteobettini: So currently, with that line, we are reducing the _reset shape to the batch_size. 
This is because the step_count counts steps in the batch_size and not in the don_spec.shape

But the _reset shape could be bigger (as you could have an agent dimension in the done_spec)
In the current scenario, thus, step_count would only count in the batch_size (and not for each agent for example)

If we make step_count follow the done_spec as well, then we can count also per agent steps

torchrl/envs/common.py
matteobettini: makes sense, i didn't add it as it was not there before but seems needed

torchrl/envs/transforms/transforms.py
vmoens: so essentially
```suggestion
        step_count[_reset.view_as(step_count)] = 0
```
instead of 
```
        _reset = _reset.view(*tensordict.batch_size, -1).any(-1)
        step_count[_reset] = 0
```
right?

matteobettini: Mmh no, that wouldn't work because _reset would need a restuction operation like sum.

What i am proposing is to replace 
```python
  if step_count is None:
          step_count = torch.zeros(
              tensordict.batch_size,
              dtype=torch.int64,
              device=tensordict.device,
          )
```
with 
```python
  if step_count is None:
          step_count = torch.zeros(
              self.parent.done_spec.batch_size,
              dtype=torch.int64,
              device=tensordict.device,
          )
```
so then we can just do 
```python
step_count[_reset] = 0
```

vmoens: lgtm, the only thing that I see is that we'd have a step_count that now has a trailing singleton dim that we currently don't have. This may be a problem down the line.
I'm open to the idea of removing the trailing singleton for done and reward, but before we do that I would not introduce new ones in the midst

matteobettini: It would have the trailing singleton as "is_init", "done" and "_reset" have.  In my view, these 3 keys and more like "step_count", "truncated", etc. should all follow done_spec. Whatever done_spec is (now it happens to have a trailing singleton).

But more than the trailing singleton my point is about multi-agent.

Consider the example where we have 
```python
td.batch_size = (32,64)
td["done"].shape = (32,64, n_agents, 1)
```
then step_count would be
```python
td["step_count"].shape = (32,64)
```
while it  better to have
```python
td["step_count"].shape = (32,64,n_agents,1)
```

matteobettini: the thing is not to assume that done_spec.shape is always `(*batch_size,1)`. If we make this assumption in the codebase, we bring problems down the line

matteobettini: I'll make the modifications to step_count and notify when this pr is ready for review

torchrl/envs/transforms/transforms.py
```diff
@@ -3064,7 +3066,7 @@ def transform_input_spec(self, input_spec: CompositeSpec) -> CompositeSpec:
                 f"input_spec was expected to be of type CompositeSpec. Got {type(input_spec)} instead."
             )
         input_spec["step_count"] = UnboundedDiscreteTensorSpec(
-            shape=input_spec.shape,
+            shape=self.parent.done_spec.shape,
```
matteobettini, line 3069: In the tests, there is not parent env so 

`AttributeError: 'NoneType' object has no attribute 'done_spec'`

vmoens: I think we should default to the tensordict shape if there is not parent env.

matteobettini:
I think we should be ready.

All in all I think this also acted as a cleanup as i got to remove some squeezes and views.

I hope to have  found all occurrences of done-related keys. Let's hope I dodn't miss any

torchrl/collectors/collectors.py
```diff
@@ -699,7 +699,6 @@ def _step_and_maybe_reset(self) -> None:
             # collectors do not support passing other tensors than `"_reset"`
             # to `reset()`.
             if len(self.env.batch_size):
-                self._tensordict.masked_fill_(done_or_terminated, 0)
```
matteobettini, line 702: What was the utility of ths?

It was assuming that the done_spec is expandable to the other key specs

vmoens (approved):
LGTM

torchrl/envs/transforms/transforms.py
```diff
@@ -1718,18 +1719,23 @@ def __init__(
...
-            _reset = torch.ones(
-                tensordict.batch_size,
-                dtype=torch.bool,
-                device=tensordict.device
-                if tensordict.device is not None
-                else torch.device("cpu"),
+        _reset = (
+            tensordict.get(
+                "_reset",
+                torch.ones(
```
vmoens, line 1725: let's not do that. The None has a reason: we don't want to allocate tensors that we don't use. By setting default to None, we can then allocate it only when needed

torchrl/envs/transforms/transforms.py
```diff
@@ -1718,18 +1719,23 @@ def __init__(
...
+                    if self.parent
+                    else tensordict.batch_size,
+                    dtype=torch.bool,
+                    device=tensordict.device
+                    if tensordict.device is not None
+                    else torch.device("cpu"),
+                ),
             )
+            .view(*tensordict.batch_size, -1)
+            .any(-1)
```
vmoens, line 1736: why not `any(range(tensordict.batch_dims, tensor.ndim))` or similar, instead of doing 2 ops?

torchrl/envs/transforms/transforms.py
```diff
@@ -1753,6 +1759,8 @@ def _make_missing_buffer(self, data, buffer_name):
     def _call(self, tensordict: TensorDictBase) -> TensorDictBase:
         """Update the episode tensordict with max pooled keys."""
         _reset = tensordict.get("_reset", None)
+        if _reset is not None:
+            _reset = _reset.view(*tensordict.batch_size, -1).any(-1)
```
vmoens, line 1763: same as above

torchrl/envs/transforms/transforms.py
matteobettini: Got it thanks!

torchrl/envs/transforms/transforms.py
matteobettini: any does not support mult dim, but can use sum! Thanks

torchrl/envs/transforms/transforms.py
matteobettini: yep ty

vmoens:
the tests with vmas are broken and it seems to be related to this PR :)

Issue #943, opened 2023-03-01
[Feature Design] Multi-agent API for TorchRL


matteobettini:
Hello,

This is the official issue for discussing the multi-agent API of torchrl for environments and collectors.

There has been a lot of discussion on wether the agent dimension should be part of the `batch_size` or not. 

In this design doc we propose an API in which the agent dimension is added to the specs, but not to the `batch_size` of an environment. This allows to have part of the specs which are shared by all agents, like a global state, a global reward or a global done. It also allows to have these attributes as per-agent attributes.

```python
n_workers = 4
n_vectorized_envs = 32
batch_size = (n_workers, n_vectorized_envs)

n_agents = 2
n_obs_per_agent = 6
n_state_features = 18

state_spec = UnboundedContinuousTensorSpec(
    torch.Size((
        n_state_features,
    ))
) # shape = (n_state_features,)
obs_spec = UnboundedContinuousTensorSpec(
    torch.Size((
        n_obs_per_agent,
    ))
)  # shape = (n_obs_per_agent,)
action_spec = UnboundedContinuousTensorSpec(
    torch.Size((2,)),
)  # shape = (2,)
reward_spec = UnboundedContinuousTensorSpec(
    torch.Size((1,)),
) # shape = (1,)
done_spec = UnboundedContinuousTensorSpec(
    torch.Size((1,)),
)  # shape = (1,)

# Create multi-agent specs. In this case specs are identical for all agents but they could differ
multi_agent_obs_spec = torch.stack([obs_spec.clone() for _ in range(n_agents)], dim=0) # UnboundedContinuousTensorSpec with shape = (n_agents, n_obs_per_agent)
multi_agent_action_spec = torch.stack([action_spec.clone() for _ in range(n_agents)], dim=0)  # UnboundedContinuousTensorSpec with  shape = (n_agents, 2)
multi_agent_reward_spec = torch.stack([reward_spec.clone() for _ in range(n_agents)], dim=0)  # UnboundedContinuousTensorSpec with shape = (n_agents, 1)
multi_agent_done_spec = torch.stack([done_spec.clone() for _ in range(n_agents)],dim=0)  # UnboundedContinuousTensorSpec with shape = (n_agents, 1)

input_spec = CompositeSpec(
    action=multi_agent_action_spec
).expand(batch_size) # 
# input_spec.shape = (*batch_size),
# input_spec["action"].shape = (*batch_size, n_agents, 2)

output_spec = CompositeSpec(
[...]

PR #983, opened 2023-03-21
[Feature] Multiagent API in vmas

matteobettini:
Implementation of #943 in vmas

The agent dimension has been moved outside the batch_size.

I will use this simulator to implement a multi-agent training example in torch rl


matteobettini:
@vmoens  this should be ready.

vmoens (approved):
LGTM

PR #992, opened 2023-03-27
[Feature] Added support for vector-based rewards from environments in MO-Gymnasium

dennismalmgren:
## Description

I made a small edit to the spec creation code for gym-like environments to support generating vector-based zero-rewards, which multi-objective environments in MO-Gymnasium uses.

## Motivation and Context

When interacting directly with gym code, the GymWrapper code supports vector-based rewards from environments in the MO-Gymnasium library. However, there is an imaginary reward generated by torchrl on environment resets, and this causes issues when doing rollouts for example, when everything gets stacked in one tensordict. This small fix allows rollout code to handle MO-Gymnasium environments by letting the imaginary reward conform to the shape that the environment uses (but not to any value limitations in MO-gym or regular gyms).

I'm not entirely sure how to classify the issue (bug/feature), it depends on whether MO-gym is considered out of scope for the gym wrapper or not. There are other interacting aspects, such as re-using 'categorical_action_encoding', but this is also done for observations and should probably be reviewed separatly.

## Types of changes

## Checklist


vmoens (commented):
Thanks for this! I love the idea
Some gym envs do not have a reward_space so I added a check for that.
Is this something we could cover in our tests?

dennismalmgren:
Thanks!
I could try to create some tests but I looked through the suite and couldn’t find any obvious existing place for it. If you have an idea I’ll go ahead.

vmoens:
We can add mo-gymnasium as a dependency and include this env in the tests
We just need to have one instance of that created in `test_libs.py` and tested with `check_env_specs(env)`.
The ideal place to install this would be in the `script_gyms/batch_scripts` directory, under the latest version of gymnasium (ie we should install mo-gymnasium after we install the latest gymnasium [here](https://github.com/pytorch/rl/blob/e66db532a450476c7cbd8882bdf9971ae37d6e13/.circleci/unittest/linux_libs/scripts_gym/batch_scripts.sh#L119-L144)).
I can do that or you can give it a shot if you'd like.

vmoens:
Some updates:
I'm trying to find my way out with mo-gymnasium.
Basically they re-provide all the specs from gym but under mo-gymnasium which breaks the instance checks we have. 
So I should try to do these instance checks with some context variable that conditions which lib we're talking about... 
The gym ecosystem is 👌

vmoens:
@dennismalmgren We're still working on a better integration of gym for this, see #1023.
Once this is merged I'll edit this PR with a proper MO env wrapper :)

PR #967, opened 2023-03-13
[WIP] Simpler examples

vmoens:
## Description

- [ ] Simpler examples
- [ ] Simpler DQN tutorial, focused on using DQN loss with trainer --> #979 
- [x] Simpler DDPG tutorial, focuses on writing a DDPG loss module --> #979

PR #998, opened 2023-03-28
[WIP] Simpler IQL example 

BY571:
## Description

Rewrites the IQL online example and adds an offline training example for IQL.

## Motivation and Context

Updates the online IQL example similar to #967. Fixes the "correct_for_frame_skip" function for the nested config and adds a simple offline IQL example.

## Types of changes

## Checklist

### DIFF
--- a/examples/iql/iql_online.py
+++ b/examples/iql/iql_online.py
@@ -2,324 +2,108 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+"""IQL Example.
 
+This is a self-contained example of a IQL training script.
+
+It works across Gym and DM-control over a variety of tasks.
+
+Both state and pixel-based environments are supported.
+
+The helper functions are coded in the utils.py associated with this script.
+
+"""
 
 import hydra
-
-import numpy as np
-import torch
-import torch.cuda
 import tqdm
-from tensordict.nn import TensorDictModule
-from tensordict.nn.distributions import NormalParamExtractor
+from torchrl.trainers.helpers.envs import correct_for_frame_skip
 
-from torch import nn, optim
-from torchrl.collectors import SyncDataCollector
-from torchrl.data import TensorDictPrioritizedReplayBuffer, TensorDictReplayBuffer
-
-from torchrl.data.replay_buffers.storages import LazyMemmapStorage
-from torchrl.envs import EnvCreator, ParallelEnv
-from torchrl.envs.libs.gym import GymEnv
-from torchrl.envs.utils import set_exploration_mode
-from torchrl.modules import MLP, ProbabilisticActor, ValueOperator
-from torchrl.modules.distributions import TanhNormal
-
-from torchrl.objectives import SoftUpdate
-from torchrl.objectives.iql import IQLLoss
-from torchrl.record.loggers import generate_exp_name, get_logger
+from utils import (
+    get_stats,
+    make_collector,
+    make_iql_model,
+    make_iql_optimizer,
+    make_logger,
+    make_loss,
+    make_recorder,
+    make_replay_buffer,
+)
 
 
-def env_maker(env_name, frame_skip=1, device="cpu", from_pixels=False):
-    return GymEnv(
-        env_name, "run", device=device, frame_skip=frame_skip, from_pixels=from_pixels
-    )
-
-
-def make_replay_buffer(
-    prb=False,
-    buffer_size=1000000,
-    buffer_scratch_dir="/tmp/",
-    device="cpu",
-    make_replay_buffer=3,
-):
-    if prb:
-        replay_buffer = TensorDictPrioritizedReplayBuffer(
-            alpha=0.7,
-            beta=0.5,
-            pin_memory=False,
-            prefetch=make_replay_buffer,
-            storage=LazyMemmapStorage(
-                buffer_size,
-                scratch_dir=buffer_scratch_dir,
-                device=device,
-            ),
-        )
-    else:
-        replay_buffer = TensorDictReplayBuffer(
-            pin_memory=False,
-            prefetch=make_replay_buffer,
-            storage=LazyMemmapStorage(
-                buffer_size,
-                scratch_dir=buffer_scratch_dir,
-                device=device,
-            ),
-        )
-    return replay_buffer
-
-
-@hydra.main(version_base=None, config_path=".", config_name="online_config")
+@hydra.main(config_path=".", config_name="online_config")
 def main(cfg: "DictConfig"):  # noqa: F821
 
-    device = (
-        torch.device("cuda:0")
-        if torch.cuda.is_available()
-        and torch.cuda.device_count() > 0
-        and cfg.device == "cuda:0"
-        else torch.device("cpu")
+    cfg = correct_for_frame_skip(cfg)
+    model_device = cfg.optim.device
+
+    state_dict = get_stats(cfg.env)
+    logger = make_logger(cfg.logger)
+    replay_buffer = make_replay_buffer(cfg.replay_buffer)
+
+    actor_network, qvalue_network, value_network = make_iql_model(cfg)
+    policy = actor_network.to(model_device)
+    qvalue_network = qvalue_network.to(model_device)
+    value_network = value_network.to(model_device)
+
+    collector = make_collector(cfg, state_dict=state_dict, policy=policy)
+    loss, target_net_updater = make_loss(
+        cfg.loss, policy, qvalue_network, value_network
     )
+    optim = make_iql_optimizer(cfg.optim, policy, qvalue_network, value_network)
+    recorder = make_recorder(cfg, logger, policy)
 
-    exp_name = generate_exp_name("Online_IQL", cfg.exp_name)
-    logger = get_logger(
-        logger_type=cfg.logger,
-        logger_name="iql_logging",
-        experiment_name=exp_name,
-        wandb_kwargs={"mode": cfg.mode},
-    )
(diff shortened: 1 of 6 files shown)


examples/iql/iql_online.py
```diff
@@ -2,291 +2,138 @@
...
-            value_losses,
-        ) = ([], [], [])
         # optimization steps
-        for _ in range(cfg.frames_per_batch * int(cfg.utd_ratio)):
-            # sample from replay buffer
-            sampled_tensordict = replay_buffer.sample(cfg.batch_size).clone()
-
-            loss_td = loss_module(sampled_tensordict)
+        if collected_frames >= init_random_frames:
+            (actor_losses, q_losses, value_losses) = ([], [], [])
```
vmoens, line 108: One way to do this is
```python
td_log = TensorDict({}, [num_updates])
for i in range(num_updates):
    td_log[i] = loss_td.detach()
    ....
td_log = td_log.apply(lambda x: x.float().mean(), batch_size=[])

```
and then you can just log each value in that tensordict in wandb

vmoens (approved):
LGTM! Thanks for this!
Can you merge main into this?
I will try to launch them once you do!

vmoens:
@BY571 I forgot: Can you add the scripts in the tests too?

BY571:
Updated online and added offline example tests for IQL and also added them for CQL as they were still missing. 
Also had to do some fixes here and there.

vmoens (approved):
LGTM

BY571:
Updated the losses of IQL and CQL so that components can get called independently + small fixes here and there.
Online training results:
image

Also doing some offline runs to check if performance fits with the paper for CQL.

Just saw that some examples failed in the tests when logging metrics - will check about that.

BY571:
CQL offline version achieves now a better performance compared to the paper for hopper-medium (0.58 norm score):
image

Implementation-wise wise should be fine now, lets see what the tests say.

examples/cql/cql_offline.py
```diff
@@ -37,49 +42,95 @@ def main(cfg: "DictConfig"):  # noqa: F821
...
-        loss_vals = loss_module(data)
-        # backprop
-        actor_loss = loss_vals["loss_actor"]
+        # compute loss
+        loss_vals = loss_module(data.clone().to(device))
+
+        if i >= policy_eval_start:
+            actor_loss = loss_vals["loss_actor"]
+        else:
+            actor_loss = loss_vals["loss_actor_bc"]
```
vmoens, line 89: this seems pretty custom, can we give a small comment on why we need it?
Also, wondering how that kind of control flow would fit within a trainer class

examples/cql/offline_config.yaml
```diff
@@ -1,37 +1,36 @@
-# Task and env
+# env and task
 env:
-  name: HalfCheetah-v2
+  name: Hopper-v2
   task: ""
   library: gym
-  exp_name: cql_halfcheetah-medium-v2
+  exp_name: cql_${replay_buffer.dataset}
```
vmoens, line 6: Wow! Is this working?

examples/iql/iql_online.py
```diff
@@ -2,361 +2,188 @@
...
+            log_loss_td = TensorDict({}, [num_updates])
+            for j in range(num_updates):
+                # sample from replay buffer
+                sampled_tensordict = replay_buffer.sample().clone()
+                if sampled_tensordict.device != device:
+                    sampled_tensordict = sampled_tensordict.to(
+                        device, non_blocking=True
+                    )
+                else:
+                    sampled_tensordict = sampled_tensordict.clone()
```
vmoens, line 120: there are more than one clone, do we need them all?

examples/iql/utils.py
```diff
@@ -0,0 +1,290 @@
...
+                scratch_dir=buffer_scratch_dir,
+                device=device,
+            ),
+            batch_size=batch_size,
+        )
+    return replay_buffer
+
+
+def make_offline_replay_buffer(rb_cfg):
+    data = D4RLExperienceReplay(
```
vmoens, line 135: in our examples, we could default to the version that does not require the d4rl library, wdyt?
It's pretty annoying to install and the dataset works without it.

test/test_cost.py
```diff
@@ -4854,7 +4854,7 @@ def _create_seq_mock_data_cql(
         return td
 
     @pytest.mark.parametrize("delay_actor", (True, False))
-    @pytest.mark.parametrize("delay_qvalue", (True, False))
+    @pytest.mark.parametrize("delay_qvalue", (True, True))
```
vmoens, line 4857: typo :)

torchrl/objectives/cql.py
```diff
@@ -225,14 +222,33 @@ class _AcceptedKeys:
...
         action: NestedKey = "action"
         value: NestedKey = "state_value"
         state_action_value: NestedKey = "state_action_value"
         log_prob: NestedKey = "_log_prob"
+        pred_q1: NestedKey = "pred_q1"
+        pred_q2: NestedKey = "pred_q2"
+        priority: NestedKey = "td_error"
+        cql_q1_loss: NestedKey = "cql_q1_loss"
+        cql_q2_loss: NestedKey = "cql_q2_loss"
         priority: NestedKey = "td_error"
```
vmoens, line 252: duplicate

torchrl/objectives/cql.py
```diff
@@ -493,36 +501,62 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+        loss_actor_bc, bc_metadata = self.actor_bc_loss(td_device)
+        loss_actor, actor_metadata = self.actor_loss(td_device)
+        loss_alpha, alpha_metadata = self.alpha_loss(td_device)
+        metadata.update(
+            {
+                **bc_metadata,
+                **cql_metadata,
+                **actor_metadata,
+                **alpha_metadata,
+            }
```
vmoens, line 520: this assumes that all metadata are Dict[Str, Any] but maybe we'll add smth else?
Why not 4 updates? (at least this version will crash if 2 metadata collide)

torchrl/objectives/cql.py
```diff
@@ -493,36 +501,62 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+        )
+        tensordict_reshape.set(
+            self.tensor_keys.priority, metadata.pop("td_error").detach().max(0)[0]
```
vmoens, line 523: when using max(0) we should use `max(0).values` instead of `[0]`

torchrl/objectives/cql.py
```diff
@@ -741,54 +805,57 @@ def _loss_qvalue_v(self, tensordict: TensorDictBase) -> Tuple[Tensor, Tensor]:
...
+                "This could be caused by calling alpha_prime_loss method before cql_loss method."
+            )
+        if cql_q2_loss is None:
+            raise KeyError(
+                f"Couldn't find the cql_q2_loss with key {self.tensor_keys.cql_q2_loss} in the input tensordict. "
+                "This could be caused by calling alpha_prime_loss method before cql_loss method."
+            )
 
-        td_error = abs(q_pred - target_value)
+        alpha_prime = torch.clamp(self.log_alpha_prime.exp(), min=0.0, max=1000000.0)
```
vmoens, line 843: is `1000000.0` a  magic number? What is the purpose of this clamp (to understand how we can code this in a more grounded way)?
No need to clamp an exp to 0 below, we can just do clamp_max

torchrl/objectives/iql.py
```diff
@@ -357,35 +361,37 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
         if shape:
             tensordict.update(tensordict_reshape.view(shape))
         out = {
             "loss_actor": loss_actor.mean(),
             "loss_qvalue": loss_qvalue.mean(),
             "loss_value": loss_value.mean(),
-            "entropy": -td_device.get(self.tensor_keys.log_prob).mean().detach(),
+            "entropy": -tensordict_reshape.get(self.tensor_keys.log_prob)
+            .mean()
+            .detach(),
```
vmoens, line 386: I'd rather pre-compute that before putting it in the dict to avoid overloaded dict construction

vmoens (approved):
Wonderful, thanks a mil!
Just a couple of housekeeping comments

examples/cql/cql_offline.py
BY571: Indeed it's pretty custom to the CQL implementation. Just added it for completeness as it's in the official implementation. I'll add a comment on why it's there.

torchrl/objectives/cql.py
BY571: Indeed the number seems quite random, I took it directly from the official implementation [1](https://github.com/aviralkumar2907/CQL/blob/d67dbe9cf5d2b96e3b462b6146f249b3d6569796/d4rl/rlkit/torch/sac/cql.py#L271).
As far as I could see it in some tests the alpha_prime value can become quite big I guess this is just to make sure its not exploding. But I'll have a look at the paper if they mention something more specific.

examples/iql/iql_online.py
```diff
@@ -117,7 +117,7 @@ def main(cfg: "DictConfig"):  # noqa: F821
                         device, non_blocking=True
                     )
                 else:
-                    sampled_tensordict = sampled_tensordict.clone()
+                    sampled_tensordict = sampled_tensordict
```
vmoens, line 120: this can be removed :)

vmoens (approved):
Good one minor thing to edit and we're good I think

vmoens (approved):
Awesome let's proceed!

PR #1004, opened 2023-03-29
[Example] PPO simplified example

albertbou92:
## Description

Simplified PPO code example.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/.circleci/unittest/linux_examples/scripts/run_test.sh
+++ b/.circleci/unittest/linux_examples/scripts/run_test.sh
@@ -91,7 +91,7 @@ python .circleci/unittest/helpers/coverage_run_parallel.py examples/ppo/ppo.py \
   env.num_envs=1 \
   collector.total_frames=48 \
   collector.frames_per_batch=16 \
-  collector.collector_devices=cuda:0 \
+  collector.collector_device=cuda:0 \
   loss.mini_batch_size=10 \
   loss.ppo_epochs=1 \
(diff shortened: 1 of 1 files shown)


examples/ppo/config2.yaml
vmoens, line 1: what is config2? Do we have better config names?

vmoens (approved):
Superb!
If you have some training results, like an image of training curve or smth, maybe we could put that in this folder in a .md file? Wdyt?

examples/ppo/config2.yaml
albertbou92: Yes I have results! I can totally add them. 

Regarding config2, I was using as a reference for testing the original PPO paper (https://arxiv.org/abs/1707.06347), so I tested the config for one Atari environment and one MuJoCo environment. config2 is the configuration for HalfCheetah with the hyperparameters of the original paper.

vmoens (commented):
another thing: let's maybe move the imports in the main function?
Like that we can use hydra + submitit/slurm to launch jobs
The idea is that from a jump node you do
python ppo.py blabla
and submitit starts the jobs for you
but for that we need to have a script with minimal imports (you never know how brittle your jump node can be)

vmoens:
I also see that the examples CI are failing, let's fix that

albertbou92:
I think it should be fine now! The remaining tests failing seem to be due to other code

PR #1014, opened 2023-04-01
Advantages pass `time_dim`and docfix

matteobettini:


torchrl/objectives/value/functional.py
```diff
@@ -351,7 +345,6 @@ def td1_return_estimate(
     return returns
 
 
-@_transpose_time
```
matteobettini, line 354: There is no need to wrap the functions which call transpose_time functions.
We just pass the time_dim directly

vmoens: Yep but we need it in the docstrings and hence in the arguments!

matteobettini: What do you mean, it is an argument and also in the docstring.

I just removed it as an argument from the td0 functions, as those do not need time

vmoens: Maybe not the advantages but I would keep it for td1 as in the future we might optimize this function to reduce the unnecessary overhead.

vmoens (approved):
LGTM 
Maybe we can keep the transpose for td1 in case we rewrite the function?

torchrl/objectives/value/functional.py
matteobettini: I am not understanding. All the functions apart from td0 ones still have the argument and the docstring.

torchrl/objectives/value/functional.py
```diff
@@ -403,12 +395,13 @@ def td1_advantage_estimate(
...
         raise RuntimeError("shape of state_value and next_state_value must match")
-    returns = td1_return_estimate(gamma, next_state_value, reward, done, rolling_gamma)
+    returns = td1_return_estimate(
+        gamma, next_state_value, reward, done, rolling_gamma, time_dim=time_dim
+    )
     advantage = returns - state_value
     return advantage
 
 
-@_transpose_time
```
matteobettini, line 411: You mean this one?

matteobettini: The problem is that if I readd this line and then inside the function we do
```
  return vec_td_lambda_return_estimate(
        gamma=gamma,
        next_state_value=next_state_value,
        reward=reward,
        done=done,
        rolling_gamma=rolling_gamma,
        lmbda=1,
        time_dim=time_dim, # this line
    )
```
than it creates a bug because time_dim is not the right one anymore.

On the other hand , if we readd it and remove ` time_dim=time_dim` (like before this pr), we are still error prone because we are relying on the default_dim dim which is something we want to deprecate.

torchrl/objectives/value/functional.py
vmoens: Sorry I misread :)

PR #1015, opened 2023-04-02
[BugFix] Fix in resetting SumReward and StepCounter transforms

albertbou92:
## Description

In the current version of TorchRL, when envs are reset, a single TensorDict with the bool tensor _reset is passed to the env. This causes problems in the SumReward and StepCounter transforms. 

Summarising, for batched environments every time one environment reaches the end of the episode, these transforms reset their count for all envs because its gets the most recent count from the input TensorDict itseld. This PR proposes a possible solution.

Now, the transforms keep track of the latest information for each environment and only reset the count for the envs indicated in _reset.

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -2876,7 +2876,7 @@ class RewardSum(Transform):
             raise RuntimeError(
                 "the out_keys must be specified for non-conventional in-keys in RewardSum."
             )
-
+        self._current_reward_sums = collections.defaultdict()
         super().__init__(in_keys=in_keys, out_keys=out_keys)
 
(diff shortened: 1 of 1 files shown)


vmoens (commented):
Any chance we can keep things stateless?
Also, can we have a non regression test for both transforms?

albertbou92:
We can keep transforms stateless as long as the whole current tensordict is passed on to the env.reset() method
https://github.com/pytorch/rl/blob/main/torchrl/collectors/collectors.py#L757

But doing raises errors in other transforms. Is there any design decision preventing us from passing the whole tensordict?

vmoens:
Can you give me a minimally reproducible example to work with? I think i can fix the collectors to make this work

vmoens:
Fixed by #1021

PR #1021, opened 2023-04-04
[BugFix] Fix collector reset with truncation

vmoens:
## Description

Alternative to #1015

### DIFF
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -534,6 +534,18 @@ class SyncDataCollector(DataCollectorBase):
         self.env: EnvBase = self.env.to(self.device)
         self.max_frames_per_traj = max_frames_per_traj
         if self.max_frames_per_traj > 0:
+            # let's check that there is no StepCounter yet
+            for key in self.env.output_spec.keys(True, True):
+                if isinstance(key, str):
+                    key = (key,)
+                if "truncated" in key:
+                    raise ValueError(
+                        "A 'truncated' key is already present in the environment "
+                        "and the 'max_frames_per_traj' argument may conflict with "
+                        "a 'StepCounter' that has already been set. "
+                        "Possible solutions: Set max_frames_per_traj to 0 or "
+                        "remove the StepCounter limit from the environment transforms."
+                    )
             env = self.env = TransformedEnv(
                 self.env, StepCounter(max_steps=self.max_frames_per_traj)
             )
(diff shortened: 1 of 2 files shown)


torchrl/collectors/collectors.py
```diff
@@ -759,11 +771,12 @@ def _step_and_maybe_reset(self) -> None:
                 _reset = None
                 td_reset = None
             td_reset = self.env.reset(td_reset)
-            self._tensordict.update(td_reset, inplace=True)
-            done = self._tensordict.get("done")
-            if (_reset is None and done.any()) or (
-                _reset is not None and done[_reset].any()
-            ):
+            reset_idx = done_or_terminated.squeeze(-1)
```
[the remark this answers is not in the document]
vmoens: got it
Then what should we do? Reduce done until its shape matches the tensordict's?
```suggestion
            reset_idx = done_or_terminated
            while reset_idx.ndim > self._tensordict.ndim:
                reset_idx = reset_idx.any(-1)
```

torchrl/collectors/collectors.py
```diff
@@ -759,18 +771,18 @@ def _step_and_maybe_reset(self) -> None:
...
-                )
             traj_done_or_terminated = done_or_terminated.sum(
                 tuple(range(self._tensordict.batch_dims, done_or_terminated.ndim)),
                 dtype=torch.bool,
             )
+            self._tensordict.get_sub_tensordict(traj_done_or_terminated).update(
+                td_reset[traj_done_or_terminated], inplace=True
+            )
+            done = self._tensordict[traj_done_or_terminated].get("done")
+            if (_reset is None and done.any()) or (_reset is not None and done.any()):
```
matteobettini, line 782: This check here might be worth keeping like before because
`self._tensordict.get("done")[_reset]`
checks allt the done_sec dims, while
`self._tensordict[traj_done_or_terminated].get("done")`
uses only `traj_done_or_terminated` which is smaller than the _reset

matteobettini: taking your example if `_reset.shape = [3, 4, 5]` but only `_reset[0,0,0] = True`, when you compute `traj_done_or_terminated` you get `[True, False,False]`.

now when you do `self._tensordict[traj_done_or_terminated].get("done")` you get a done of shape `[4,5]` and you check that none of it is True, while you should check only `[0,0]`

matteobettini, line 782: I propose that 
```
 self._tensordict.get_sub_tensordict(traj_done_or_terminated).update(
                td_reset[traj_done_or_terminated], inplace=True
            )
```
should be the only line replced here and 
```
 done = self._tensordict.get("done")
            if (_reset is None and done.any()) or (
                _reset is not None and done[_reset].any()
            ):
```
should be kept

vmoens: I agree with that point but at that point we'll need a deep refactoring of that method anyway and this line is likely to be changed no?
I'm not against the idea of going back to the previous version but I'd like to understand why one is better than the other

matteobettini: Yep they are equivalent, and also equvalent to `not self._tensordict.get("done").any()` which may be the simplest

vmoens: I was thinking the same thing
What about: we take the simplest option (for now) and leave it as is until we figure out what to do with envs that can start with a done or env that keep being executed when done?

matteobettini: The if in this pr has an or which can be removed BTW dunno if you saw it

vmoens: sorry I did not get that

PR #1023, opened 2023-04-05
[BugFix] Fix Gym imports

vmoens:
## Description

Allows to dynamically set the gym backend.

## The problem

The problem this PR addresses is that if `gym` and `gymnasium` are both installed, it can be difficult to build functions that rely on one or the other lib. As of now, we have a `implement_for` function that allows to code multiple different implementations of a same function for multiple backends. However, if two backends exist, we need to arbitrarily choose one over the other.
This PR proposes a `set_gym_backend` decorator to choose which gym backend should be used. For instance, the function  `_gym_to_torchrl_spec_transform` currently gets easily detached from the gym version that should be used. With this decorator, we can be sure that `_gym_to_torchrl_spec_transform` will fetch the proper gym backend given that it reads it from the `gym_backend` function. This is aimed at integrating [MO-gymnasium](#992) for instance, which currently breaks when `_gym_to_torchrl_spec_transform` is called because the `Box` object should come from `mo_gymnasium` and not `gym` or `gymnasium`.  

TODO:
- [x] Test `set_gym_backend`: we need a comprehensive set of tests of this function. One is already sketched in `test_utils`, but the test should be more thorough. One way to test this would be in the gym pipeline to install gymnasium while the latest gym is still installed (see [here](https://github.com/pytorch/rl/blob/2ac7f25c616cbefc7e4f8ce5b596073ff975267a/.circleci/unittest/linux_libs/scripts_gym/batch_scripts.sh#L108-L151)). It is a stretch but many people have gymnasium and gym installed in their virtual env... Otherwise we could test this with mock gym libraries...
- [x] Test `implement_for` with function and methods: we added the possibility of providing a function to `implement_for` rather than a string, to allow `implement_for(gym_backend)`. This functionality should be properly tested.
- [x] Test that `set_gym_backend` leaves the implements unchanged: when exiting, `set_gym_backend` calls the new `implement_for.reset` method which will reset the function links where they belong. This should also be tested properly.
- [x] Add the new functions to the doc


DamienAllonsius:
for the first bullet, have you tried [to mock import this way?](https://stackoverflow.com/questions/8658043/how-to-mock-an-import)

vmoens:
> for the first bullet, have you tried [to mock import this way?](https://stackoverflow.com/questions/8658043/how-to-mock-an-import)

I guess that could work

torchrl/_utils.py
```diff
@@ -228,18 +240,52 @@ class implement_for:
...
+        from_version: str = None,
+        to_version: str = None,
     ):
         self.module_name = module_name
         self.from_version = from_version
         self.to_version = to_version
+        implement_for._setters.append(self)
+
+    @staticmethod
+    def check_version(version, from_version, to_version):
```
DamienAllonsius, line 257: FYI, the previous implementation of check_version had a bug:
`check_version("0.9.0", "0.1.0", "0.21.0")` returned `False`

torchrl/_utils.py
```diff
@@ -255,8 +255,11 @@ def __init__(
 
     @staticmethod
     def check_version(version, from_version, to_version):
-        return (from_version is None or version >= from_version) and (
-            to_version is None or version < to_version
+        def str2tuple(v):
+            return tuple(map(int, v.split(".")))
+
+        return (from_version is None or str2tuple(version) >= str2tuple(from_version)) and (
+            to_version is None or str2tuple(version) < str2tuple(to_version)
```
vmoens, line 262: can't we use packaging.version.parse to compare versions here?

test/test_utils.py
```diff
@@ -134,31 +137,107 @@ def test_implement_for_reset():
...
+                expected_from_version_gymnasium,
+                expected_to_version_gymnasium,
+            ):
+                expected_fn_gymnasium = impfor.fn
 
-    with set_gym_backend(gymnasium):
-        _utils_internal._set_gym_environments()
-        pendulum = GymEnv(_utils_internal.PENDULUM_VERSIONED)
-        check_env_specs(pendulum)
-        assert isinstance(pendulum._env.action_space, gymnasium.spaces.Box)
```
vmoens: It may seem redundant but the reason we tested gymnasium -> gym -> gymnasium was to make sure that the resetting was ok, does that make sense?

vmoens (commented):
I like it, it's much nicer like this!

torchrl/envs/libs/gym.py
```diff
@@ -26,41 +33,88 @@
...
-    _has_gym = True
-except ImportError as err:
-    IMPORT_ERROR = err
-    try:
-        import gymnasium as gym
 
-        _has_gym = True
-    except ImportError as err:
-        IMPORT_ERROR = err
+class set_gym_backend(_DecoratorContextManager):
```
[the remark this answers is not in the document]
vmoens: no it isn't but it's ok for classes that you use as decorator or callables (IMO). You wouldn't think of

```
@SetGymBackend(...)
def foo():
   etc..
```

I think it's ok to keep it gym-specific, the chances that we'll have to handle something of similar complexity with other libs are thin

torchrl/_utils.py
```diff
@@ -252,36 +305,45 @@ def unsupported(*args, **kwargs):
...
-                        f"Using the last queried ({module} with version {version})."
-                    )
-                else:
-                    return implementations[func_name]
+                version = self.import_module(self.module_name)
+                if self.check_version(version, self.from_version, self.to_version):
+                    if VERBOSE:
+                        warnings.warn(
+                            f"Got multiple backends for {func_name}. "
+                            f"Using the last queried ({module} with version {version})."
```
vmoens, line 313: This is the actual behaviour, in contrast with the docstring which is wrong.
I wonder if for now rather than VERBOSE we shouldn't be using RL_WARNINGS? Then mention in the warning that it can be turned off by setting RL_WARNINGS to False...

vmoens (commented):
(I can't approve it since it's my PR)
Looking good so far!
Let's write some more doc (in docstrings and in docs/source/reference/envs.rst) + fix the lint and test reset.

torchrl/_utils.py
DamienAllonsius: Previously, there a was a comment saying "we don't use RL_WARNINGS because it's utterly verbose". It seems indeed that RL_WARNINGS is an environment variable set everywhere through the code. Why don't using the standard logging API?

vmoens: yeah we probably should :)

PR #1027, opened 2023-04-06
[Example] Multiagent examples: MAPPO-IPPO-MADDPG-IDDPG-IQL-QMIX-VDN

matteobettini:
Example of multiagent training algos in vmas


torchrl/objectives/value/advantages.py
```diff
@@ -365,7 +365,7 @@ def is_stateless(self):
         return self.value_network._is_stateless
 
     def _next_value(self, tensordict, target_params, kwargs):
-        step_td = step_mdp(tensordict)
+        step_td = step_mdp(tensordict, keep_other=False)
```
matteobettini, line 368: We need this because in case "next" does not contain the new inputs for the value network, the old ones will be silently brought forward and no error would be thrown. If we set it to False, no state from the previous step is brought forward and we gin perf

matteobettini:
@vmoens Could you give a preliminary review?

In this PR I am introducing the following things:

- a `examples/multiagent` folder containing scripts for all the algos tested in the paper (MAPPO-IPPO-MADDPG-IDDPG-IQL-QMIX-VDN) in vmas. the scripts follow all the same structure are aiming to be extra intiutive. All the algos are trained on-policy for comparison across algos. I am writing a readme for this folder to explain better
- a `objectives/multiagent` folder with a first entry which is a q mixer loss
- a `modules/models/multiagent.py` file which will contain multiagent only models and currently contains the miers for VDN. and QMIX (i could eventually move the multiagent_mlp from `examples/multiagent` here if you find it helpful)

Let me know

examples/multiagent/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+In this folder we provide a set of multi-agent example scripts using the [VMAS](https://github.com/proroklab/VectorizedMultiAgentSimulator) simulator.
+
+<p align="center">
+<img src="../media/marl_vmas.png" width="600px">
+</p>
+
+<center><i>The MARL algorithms contained in the scripts of this folder run on three multi-robot tasks in VMAS.</i></center>
+
+For more details on the experiment setup and the environments please refer to the corresponding section of the appendix in the [TorchRL paper](https://arxiv.org/abs/2306.00577).
```
matteobettini, line 11: Let's make sure links are disabled in the anonymous repo

vmoens: yes they are

examples/multiagent/README.md
```diff
@@ -0,0 +1,59 @@
...
+### Install
+
+First you need to install vmas and the dependencies of the scripts.
+
+Install torchrl and tensordict following repo instructions.
+
+Install vmas and dependencies:
+
+```bash
+pip install vmas
```
vmoens, line 24: We could have a `extras_require={..., "marl": ["vmas", ...]}` in `setup.py`

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
+import time
```
vmoens, line 1: missing header

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
...
+        # Target
+        "tau": 0.005,
+        # Evaluation
+        "evaluation_interval": 20,
+        "evaluation_episodes": 200,
+    }
+
+    model_config = {
+        "shared_parameters": True,
+    }
```
vmoens, line 78: Would be nice to have a hydra config
```
collector:
  frames_per_batch: 60_000
  max_steps: 100
  ...
objective:
  gamma: 0.9
  ...
```

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
...
+        max_steps=max_steps,
+        device=vmas_device,
+        seed=seed,
+        # Scenario kwargs
+        **env_config,
+    )
+
+    env_test = VmasEnv(
+        scenario=scenario_name,
+        num_envs=config["evaluation_episodes"],
```
vmoens, line 94: why is the number of envs equal to the number of eval episodes?

examples/multiagent/iql.py
vmoens, line 1: I would split the construction in some utils, but I'll leave it to you if you prefer not to.
I like having a `make_env` and a `make_loss` that are self contained and possibly testable given a small config

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
...
+        qnet,
+        eps_init=0.3,
+        eps_end=0,
+        annealing_num_steps=int(total_frames * (1 / 2)),
+        action_key=env.action_key,
+        spec=env.action_spec,
+    )
+
+    with set_exploration_type(ExplorationType.RANDOM):
+        qnet_explore(env.reset().to(training_device))
```
vmoens, line 141: I guess you're init lazy modules, 
perhaps add a comment to tell users what you're doing

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
...
+        config.update({"model": model_config, "env": env_config})
+        model_name = ("Het" if not model_config["shared_parameters"] else "") + "IQL"
+        logger = WandbLogger(
+            exp_name=generate_exp_name(env_config["scenario_name"], model_name),
+            project=f"torchrl_{env_config['scenario_name']}",
+            group=model_name,
+            save_code=True,
+            config=config,
+        )
+        wandb.run.log_code(".")
```
vmoens, line 181: here too, i'd use a hydra structured config where we pass `config.logger` to `make_logger`. If `make_logger` returns `None` then it means that we're not logging. You can check other single agent algos for examples

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
...
+
+    total_time = 0
+    total_frames = 0
+    sampling_start = time.time()
+    for i, tensordict_data in enumerate(collector):
+        print(f"\nIteration {i}")
+
+        sampling_time = time.time() - sampling_start
+        print(f"Sampling took {sampling_time}")
```
vmoens, line 190: do we need this?
We could simply log it in wandb

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
...
+
+        tensordict_data["next", "done"] = (
```
vmoens, line 192: in general accessing data is better done via `get` than `__getitem__`

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
...
+
+                optim.step()
+                optim.zero_grad()
+                target_net_updater.step()
+
+        qnet_explore.step(frames=current_frames)  # Update exploration annealing
+        collector.update_policy_weights_()
+
+        training_time = time.time() - training_start
+        print(f"Training took: {training_time}")
```
vmoens, line 228: ditto

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
...
+                    max_steps=max_steps,
+                    policy=qnet,
+                    callback=rendering_callback,
+                    auto_cast_to_device=True,
+                    break_when_any_done=False,
+                    # We are running vectorized evaluation we do not want it to stop when just one env is done
+                )
+
+                evaluation_time = time.time() - evaluation_start
+                print(f"Evaluation took: {evaluation_time}")
```
vmoens, line 266: ditto

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
...
+
+        if log:
+            logger.experiment.log({}, commit=True)
+        sampling_start = time.time()
+    wandb.finish()
+
+
+if __name__ == "__main__":
+    for seed in [0]:
+        train(seed)
```
vmoens, line 283: this should change under the hydra version

examples/multiagent/maddpg_iddpg.py
```diff
@@ -0,0 +1,310 @@
+import time
```
vmoens, line 1: missing header

examples/multiagent/maddpg_iddpg.py
vmoens, line 1: most comments of the previous file still apply here

examples/multiagent/mappo_ippo.py
```diff
@@ -0,0 +1,313 @@
+import time
```
vmoens, line 1: missing header

examples/multiagent/iql.py
```diff
@@ -0,0 +1,283 @@
+import time
+
+import torch
+
+import wandb
```
vmoens, line 5: Let's use our wandb logger.
Better: let's make the logger configurable via a `backend` arg in the config

torchrl/objectives/multiagent/qmixer.py
```diff
@@ -0,0 +1,377 @@
+from __future__ import annotations
```
vmoens, line 1: missing header

torchrl/objectives/multiagent/__init__.py
```diff
@@ -0,0 +1 @@
+from .qmixer import QMixerLoss
```
vmoens, line 1: missing header
Also let's recursively add this to the previous `__init__.py`

torchrl/objectives/multiagent/qmixer.py
```diff
@@ -0,0 +1,377 @@
...
+    _GAMMA_LMBDA_DEPREC_WARNING,
+    default_value_kwargs,
+    distance_loss,
+    ValueEstimators,
+)
+from torchrl.objectives.value import TDLambdaEstimator
+from torchrl.objectives.value.advantages import TD0Estimator, TD1Estimator
+
+
+class QMixerLoss(LossModule):
```
vmoens, line 35: Add to doc

vmoens, line 35: I don't think it is tested yet

examples/multiagent/iql.py
matteobettini: Sure, but is it possible to do that in the same file?
I would really like to have every file be self contained (the config too)

examples/multiagent/iql.py
matteobettini: because we use vectorized envs for evaluation (1 vecotrized env per parallel evaluation episode)

examples/multiagent/iql.py
matteobettini: I would really like not to do that.
Those helper functions are what confuses me the most when reading examples, I would prefer not using any functions that abstract things in the scripts.

matteobettini:
thanks for the comments, I'll add them straight away!
the 2 main points from me are

- i am ok with using hydra instead of a python dict, but is there a way for keeping everything in one file? I feel like examples are to be read at a glimpse by basic users and hydra is not super intutive for python beginners
- I would like to avoid wrapping parts of the scripts in custom helper functions. For me that kind of defeats the purpose of a glimpse example

matteobettini:
I have assresses the comments in the examples, when you can have a look.

I moved the models and loss to #1378 so we can merge them first

examples/media/marl_vmas.png
matteobettini, line 1: I would recommend not uploading media to this repo.

The repo size is already huge and this can impact user with limited disk space.

Iwould recommend creating a secundary repo/s3 bucket/ some other hosting service to host the media and just embed links to them

vmoens: I agree, this isn't scalable on the long term.
We need to make sure that whatever we link will remain on the long term
pytorch/tutorials and pytorch/pytorch (as well as vision I think) are pretty big though
Let me see how we're supposed to handle things

matteobettini: yeah i dunno if pytorch has any hosting service or repo, but those would be cool.

It is your call anyway, I was just pointing out that git will not only save the media, but also all their previous versions, and it will all be downloaded upon clone

vmoens (commented):
we must add these examples to the example CI
They should not take more than a minute each to run (ie very small nets and very few iterations). We just want to make sure they run ok
I'll what we can do regarding the images and s3 bucket

matteobettini:
The examples are now in the CI, if we sort out the s3 we are ready to go

vmoens (approved):
LGTM
The figure can be found [here](https://pytorch.s3.amazonaws.com/torchrl/github-artifacts/img/marl_vmas.png)

PR #1034, opened 2023-04-11
[BugFix] Loss function convert_to_functional seems to generate backprop issues with shared policy-value networks

albertbou92:
## Description

At least in PPO (but also probably in the rest of the Objective classes), calling self.convert_to_functional in the following lines when using a policy-value shared architecture causes the value network not to update correctly. Could the current code be stopping the value loss from being propagated to the first layers?
https://github.com/pytorch/rl/blob/main/torchrl/objectives/ppo.py#L113
https://github.com/pytorch/rl/blob/main/torchrl/objectives/ppo.py#L117

This PR removes the calling of self.convert_to_functional. I ran it in a test case for Atari Pong env and as the plots show it solved the issue. Green line is the modified version of the code. However, a proper solution should probably be discussed because I am not sure if simply removing these lines causes problems somewhere else. 

PPO_bug1
PPO_bug2

### DIFF
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -114,7 +114,9 @@ class PPOLoss(LossModule):
         )
         # we want to make sure there are no duplicates in the params: the
         # params of critic must be refs to actor if they're shared
-        self.convert_to_functional(critic, "critic", compare_against=self.actor_params)
+        # self.convert_to_functional(critic, "critic", compare_against=self.actor_params)  # Value network update is not correct
+        # self.convert_to_functional(critic, "critic")  # Value network update is not correct either
+        self.critic = critic
         self.advantage_key = advantage_key
         self.value_target_key = value_target_key
         self.samples_mc_entropy = samples_mc_entropy
@@ -169,7 +171,7 @@ class PPOLoss(LossModule):
             tensordict_select = tensordict.select(*self.critic.in_keys)
             state_value = self.critic(
                 tensordict_select,
-                params=self.critic_params,
+                # params=self.critic_params,
             ).get("state_value")
             loss_value = distance_loss(
                 target_return,


vmoens:
When you backprop the two losses, I guess that in your implementation the gradients for the common modules are a sum of gradient for the two losses, right?
What `compare_against` does is that it feeds the module (in this case the critic) with a detached version of the params used for the other module (in this case the policy). This is what is done in some algos but I understand that it may be suboptimal in other cases, especially if the reset of your critic is very shallow.

Have you tried to just remove the `compare_against`? 

If that works, a solution could be an optional call compare_against with a default to not doing it.

albertbou92:
Some updates. I actually only need to modify this line:  https://github.com/pytorch/rl/blob/main/torchrl/objectives/ppo.py#L117

Here is what works and what does not:
self.convert_to_functional(critic, "critic", compare_against=self.actor_params)  # Value network update is not correct
self.convert_to_functional(critic, "critic")  # Value network update is not correct either
self.critic = critic # This works

So removing the compare_against it does not seem to be the solution.

albertbou92:
> When you backprop the two losses, I guess that in your implementation the gradients for the common modules are a sum of gradient for the two losses, right? What `compare_against` does is that it feeds the module (in this case the critic) with a detached version of the params used for the other module (in this case the policy). This is what is done in some algos but I understand that it may be suboptimal in other cases, especially if the reset of your critic is very shallow.
> 
> Have you tried to just remove the `compare_against`?
> 
> If that works, a solution could be an optional call compare_against with a default to not doing it.

Yes for the code to work the gradients have to be the sum of the two losses

vmoens:
You're right, even without that, there is not param tighing. Let me fix that to have a more predictable behaviour.

vmoens:
Closed by #1037

Issue #16, opened 2022-02-11
[Feature Request] Reward-to-go


vmoens:
Implement reward-to-go (as in [here](https://spinningup.openai.com/en/latest/spinningup/rl_intro3.html#implementing-reward-to-go-policy-gradient))

skandermoalla:
@vmoens Is this issue still open?  @BY571 Does #1038 close it?

vmoens:
Yes this is closed now!

skandermoalla:
Nice! @vmoens Do you know if this handles truncation correctly [(e.g. as introduced in Gymnasium)](https://gymnasium.farama.org/tutorials/gymnasium_basics/handling_time_limits/#truncation)? I.e. allows to give a value for the truncation state, $V(s_{truncation})$, to bootstrap all the "reward-to-go", $G_t$, in that episode from it.
Or maybe, if this is deferred to the learning logic, there is a flag for states belonging to truncated episodes so that one could bootstrap their values (e.g. $\nabla \log \pi(a_t|s_t)[(G_t + (1_{s_t \in truncatedEpisode} * V(s_{truncation})) ...]$).
I'm not aware of the updates since #403.

skandermoalla:
Similarly, for episodes that have not finished yet (typically the last episode in each concurrent environment), is there a way to find those and mask them out in the loss?
Thanks!

BY571:
Yes, when an episode was ended (without done=True) truncated is set true on that last state the transform handles it as if that was the last state of the episode:
```
>>> from torchrl.envs.transforms import Reward2GoTransform
>>> import torch
>>> from tensordict import TensorDict
>>> r2g = Reward2GoTransform(in_keys=["reward"], out_keys=["reward_to_go"])
>>> td = TensorDict({"reward": torch.ones(4,1), "next": {"done":torch.zeros(4,1).to(dtype=bool), "truncated": torch.zeros(4,1).to(dtype=bool)}}, batch_size=())
>>> td["next"]["truncated"][-1]=True
>>> r2g._inv_call(td)["reward_to_go"]
tensor([[4.],
        [3.],
        [2.],
        [1.]])
```
If you want to mask these episodes out completely you might have to set states, reward, actions (etc) to zero. Simply setting truncated to True for all those steps would not work. Then the reward-to-go transform returns only the current reward per step as it expects that each step is a single episode with length=1:

```
[...]

Issue #990, opened 2023-03-24
Reward of 0 added to trajectory beginning


matteobettini:
Currently when rolling out an env or using a collector we get a tensordict 

```pyhton
td = TensorDict({
   "obs": torch.Tensor([T, ...]),
   "action": torch.Tensor([T, ...]),
   "done": torch.Tensor([T, ...]),
   "reward": torch.Tensor([T, ...]),
   "next": TensorDict({
        "obs": torch.Tensor([T, ...]),
        "done": torch.Tensor([T, ...]),
        "reward": torch.Tensor([T, ...]),
    }, batch_size=[T]),
}, batch_size=[T])
```

We can see that this provides a 0 reward at t=0: `td["reward“][0, ...] # is equal to all zeros`

This is a big issue as the value of 0 has been chosen arbitrarily and could damage the learning process on trajectories.

## Solution

The simple way to see a tensordict coming from a rollout is this:

- the keys at the first level are the ones available at t=0: obs, done, action
- the keys under the "next" key are the ones available at t=T: obs, done, reward

Therefore, reward should not be present at the first level (for the same reason "action" is not in "next"!)

vmoens:
Do we even need "done" at the root level?
td["next", "done"] can be true, but the only way for td["done"] to be true is either if we make a step in a done env or if we have a done at init time. Maybe that should be hidden by default?

PR #1038, opened 2023-04-11
[Feature] Reward2go Transform

BY571:
## Description

Adds the Reward2Go Transform.

## Motivation and Context

Reward2Go transform is needed for example for the decision transformer. 
Also, there exists an open issue for the reward-to-go #16 

## Types of changes

## Checklist


BY571:
There might be a critical case when the collected episode did not terminate. As the reward-to-go calculation can only happen with a finished episode. It would be good to hear your opinion on that @vmoens. For now, I added an assertion that gets triggered when the collected batch has no done = True.  

But for the cases where the episode length is > frames_per_batch this would be an issue. Not sure if we can store parts of unfinished trajectories in the transforms.

torchrl/envs/transforms/utils.py
```diff
@@ -11,3 +11,24 @@ def check_finite(tensor: torch.Tensor):
...
+    start_idx = 0
+    r2gs = []
+    for idx in episode_ends:
+        r2gs.append(discounted_cumsum(reward[start_idx:idx], gamma))
+        start_idx = idx
+    return torch.vstack(r2gs)
+
+
+def discounted_cumsum(reward: torch.Tensor, gamma: float):
+    """Compute the discounted cumulative sum of rewards."""
```
vmoens, line 32: Let's indicate the args and focus on what shape is expected for the input

torchrl/envs/transforms/transforms.py
```diff
@@ -3607,3 +3607,56 @@ def transform_input_spec(self, input_spec: CompositeSpec) -> CompositeSpec:
...
+        )
+
+        if not isinstance(gamma, torch.Tensor):
+            gamma = torch.tensor(gamma)
+
+        self.register_buffer("gamma", gamma)
+
+    def _inv_call(self, tensordict: TensorDictBase) -> TensorDictBase:
+        episode_ends = torch.where(tensordict.get(("next", "done")))[0]
+        assert episode_ends.shape[0] > 0, "No episode ends found"
```
vmoens, line 3646: We should not have `assert` in the code base. Here, I would use a `RuntimeError`.

torchrl/envs/transforms/transforms.py
```diff
@@ -3607,3 +3607,56 @@ def transform_input_spec(self, input_spec: CompositeSpec) -> CompositeSpec:
...
+            out_keys_inv=out_keys,
+        )
+
+        if not isinstance(gamma, torch.Tensor):
+            gamma = torch.tensor(gamma)
+
+        self.register_buffer("gamma", gamma)
+
+    def _inv_call(self, tensordict: TensorDictBase) -> TensorDictBase:
+        episode_ends = torch.where(tensordict.get(("next", "done")))[0]
```
vmoens, line 3645: we can also have `("next", "truncated")`.

```suggestion
        done = tensordict.get(("next", "done"))
        truncated = tensordict.get(("next", "truncated"), None)
        if truncated is not None:
            done_or_truncated = done | truncated
        else:
            done_or_truncated = done
        if not done_or_truncated.any(-2).all():
            raise RuntimeError("No episode ends found")
        episode_ends = torch.where(done)[0]
```

torchrl/envs/transforms/transforms.py
vmoens: See my comment above for an implementation that checks that `all` trajectories have a `done` or `truncated` state.
Note: should they all *end* with a truncated or done? In other words, what happens if you cat two trajectories one after another and the second does not end?

torchrl/envs/transforms/transforms.py
```diff
@@ -3607,3 +3607,56 @@ def transform_input_spec(self, input_spec: CompositeSpec) -> CompositeSpec:
...
+class Reward2GoTransform(Transform):
+    """Calculates the reward to go based on the episode reward and a discount factor.
+
+    As the Reward2GoTransform is only an inverse transform the in_keys will be directly used for the in_keys_inv.
+    The reward-to-go can be only calculated once the episode is finished. Therefore, the transform should be applied to the replay buffer
+    and not to the collector.
+
+    Args:
+        in_keys (list of str/tuples of str): the entries to rename
+        gamma (float or torch.Tensor): the discount factor
```
vmoens, line 3621: Add default value.

torchrl/envs/transforms/transforms.py
```diff
@@ -3607,3 +3607,56 @@ def transform_input_spec(self, input_spec: CompositeSpec) -> CompositeSpec:
...
+
+class Reward2GoTransform(Transform):
+    """Calculates the reward to go based on the episode reward and a discount factor.
+
+    As the Reward2GoTransform is only an inverse transform the in_keys will be directly used for the in_keys_inv.
+    The reward-to-go can be only calculated once the episode is finished. Therefore, the transform should be applied to the replay buffer
+    and not to the collector.
+
+    Args:
+        in_keys (list of str/tuples of str): the entries to rename
```
vmoens, line 3620: move down one line.
Add default value.

torchrl/envs/transforms/transforms.py
```diff
@@ -3607,3 +3607,56 @@ def transform_input_spec(self, input_spec: CompositeSpec) -> CompositeSpec:
             if not self.create_copy:
                 del input_spec[in_key]
         return input_spec
+
+
+class Reward2GoTransform(Transform):
+    """Calculates the reward to go based on the episode reward and a discount factor.
+
+    As the Reward2GoTransform is only an inverse transform the in_keys will be directly used for the in_keys_inv.
```
vmoens, line 3615:
```suggestion
    As the :class:`~.Reward2GoTransform` is only an inverse transform the ``in_keys`` will be directly used for the ``in_keys_inv``.
```

torchrl/envs/transforms/transforms.py
```diff
@@ -3607,3 +3607,56 @@ def transform_input_spec(self, input_spec: CompositeSpec) -> CompositeSpec:
...
+    The reward-to-go can be only calculated once the episode is finished. Therefore, the transform should be applied to the replay buffer
+    and not to the collector.
```
vmoens, line 3617: makes me think:
when we register a transform to an env, we call `Transform.set_container`.
What we could do is this
```
def set_container(self, container):
    if isinstance(container, Envbase) or  container.parent is not None:
        raise ValueError(...)
```
That won't prevent someone from doing
```
t = Reward2GoTransform()
t = Compose(t)
env = TransformedEnv(env, t)
```
but it will prevent
```
env = TransformedEnv(env)
env.append_transform(t)
```

test/test_transforms.py
```diff
@@ -3845,6 +3846,102 @@ def test_transform_inverse(self):
...
+        raise pytest.skip("No environment transform for Reward2Go")
+
+    def test_serial_trans_env_check(self):
+        raise pytest.skip("No environment transform for Reward2Go")
+
+    def test_trans_serial_env_check(self):
+        raise pytest.skip("No environment transform for Reward2Go")
+
+    def test_trans_parallel_env_check(self):
+        raise pytest.skip("No environment transform for Reward2Go")
```
vmoens, line 3915: If any of these guys should raise an exception (see below), it should be tested

vmoens (commented):
Good work! thanks for this
See my comments.
In general, I think we should assume that the time dimension is at `-2` instead of `0` and this should be parametrizable for the functional (as it is for our value functions such as gae etc)

vmoens:
@BY571 We should also add the new functions and class to the doc

torchrl/envs/transforms/transforms.py
BY571: That's an open problem. The situation that we can have trajectories that didn't end yet are currently handled as if those ended. But a bigger problem is for a batch that consists of one whole trajectory that didn't end.  For example, the episode length is 2000 but we collect only 500 frames per batch. Which means there will be no True done state. For now, I managed this case with the RuntimeError and wrote explicitly that the frames_per_batch should at least be equal to the episode length. Once properly implemented with the Decision Transformer I will have to test what happens with episodes that didn't end properly (got cut off and didn't end with a done) but I think for most of the environments this will not be a huge problem like gym or mujoco.

torchrl/envs/transforms/utils.py
```diff
@@ -11,3 +11,24 @@ def check_finite(tensor: torch.Tensor):
...
+    discount = torch.pow(gamma, torch.arange(reward.shape[0])).unsqueeze(-1)
+    return torch.cumsum(reward * discount, dim=0).flip(0)
```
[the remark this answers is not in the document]
BY571: This could become tricky when `len(B) >1`. Will need to do some tests.

torchrl/envs/transforms/transforms.py
BY571: added it and also added:

```
    def _apply_transform(self, reward: torch.Tensor) -> torch.Tensor:
        raise ValueError(
            "The Reward2GoTransform is only an inverse transform and can only be applied to the replay buffer and not to the collector or the environment."
        )
```
Now catching both cases or generally, any apply call.

BY571: Okay, my fault when sampling from the replay buffer the apply_transform is called and would trigger the error. So it cant be there.

torchrl/envs/transforms/utils.py
```diff
@@ -11,3 +11,24 @@ def check_finite(tensor: torch.Tensor):
     """Raise an error if a tensor has non-finite elements."""
     if not tensor.isfinite().all():
         raise ValueError("Encountered a non-finite tensor.")
+
+
+def compute_reward2go(
```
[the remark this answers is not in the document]
BY571: moved it!

BY571:
Did some updates, let me know what you think.

vmoens:
I changed the following things:
- I implemented another functional which I believe is more general (supports different shapes) but also should be faster. I had to change the signature but I think it is more suited to what users might expect (?)
- The tests were using batch and time sequence interchangeably, it was a bit confusing. I rewrote them separating the batch and time, but maybe you also want to test how things work with a "flat" replay buffer?
- I added an error message if the reward key was not found. IMO we should implement that for all transforms, as skipping a transform is not ideal (probably due to a key misspelling or else)
- I cloned the tensordict before passing it to the transform when a no-op was expected. In general
```python
td_out = transform(td)
assert (td_out == td).all()
``` 
will always return ``True`` bc modifications are done in place.
```python
td_out = transform(td.clone())
assert (td_out == td).all()
``` 
is more appropriate.
- I removed the xfail bc we don't have any control over when the exception occurs so we're not properly testing that it occurs when we want it to (is it in construction, rollout or check_env_specs)?

LMK what you think!

Attention: the "reward" associated with a "done" lives in the "next" entry!
```
TensorDict({
    "reward": Tensor(...),  # <- not this one!!
    "next": {
        "reward": Tensor(...),  # <- this one yes!!
        "done": Tensor(...),
    }, [])
```

vmoens:
@BY571 comments on this?

BY571:
looking at it and makes all sense. Just wondering about the reward and "next" reward in the tensordict. What's the exact difference? Reward is the one from the previous time step and next is the one obtained from the action taken?

vmoens:
> looking at it and makes all sense. Just wondering about the reward and "next" reward in the tensordict. What's the exact difference? Reward is the one from the previous time step and next is the one obtained from the action taken?

`("next", "reward")` is the one resulting from the action in the tensordict.
Actually I'm thinking about removing the "reward" at the root level altogether. Thoughts @albertbou92 @shagunsodhani  @matteobettini? 
Context: Do we need `tensordict["reward"]` when the result of a step is in `tensordict["next", "reward"]`? That may be confusing (there are 2 rewards, with the 'root' one being the offset version of the 'next' one).
@matteobettini pointed this already in #990 
I would suggest addressing #990 and even go one step further and disable the passing of `"reward"` when using `step_mdp`.

vmoens:
@BY571 I also changed the out_keys:
- the user can specify which out_keys are to be used
- by default, `("next", "reward")` is used such that the reward passed to the loss module is the appropriate one.
Thoughts?

matteobettini:
I stand by my discussion in #990.
And yes losses etc. I think should use next-reward

BY571:
> > looking at it and makes all sense. Just wondering about the reward and "next" reward in the tensordict. What's the exact difference? Reward is the one from the previous time step and next is the one obtained from the action taken?
> 
> `("next", "reward")` is the one resulting from the action in the tensordict. Actually I'm thinking about removing the "reward" at the root level altogether. Thoughts @albertbou92 @shagunsodhani @matteobettini? Context: Do we need `tensordict["reward"]` when the result of a step is in `tensordict["next", "reward"]`? That may be confusing (there are 2 rewards, with the 'root' one being the offset version of the 'next' one). @matteobettini pointed this already in #990 I would suggest addressing #990 and even go one step further and disable the passing of `"reward"` when using `step_mdp`.

Yes, I agree with the fact that it is kind of confusing having those two rewards. In my opinion, there should be only the reward in the tensordict that was received by the action taken at the current step - so the td["next", "reward"]. 
To keep the MDP property (s,a,r,s) inside one tensordict. 
As mentioned in #990 I guess the done at root is also not needed per default. It could be an extra transform that adds those ("reward", "done" and maybe previous "action" that led to root reward) if needed to the root level.

albertbou92:
I also agree.  We could have just ["next"]["reward"] by default and maybe have a Transform to allow having the previous reward available if the user wants to.

vmoens:
@albertbou92 @matteobettini got it, see #1065

PR #1041, opened 2023-04-12
[Refactor] Import envpool locally to avoid importing gym at root level

vmoens:


vmoens:
This PR does the job, although it uncovered a series of failures in envpool dependency. Those will be addressed separately

PR #1048, opened 2023-04-12
[CI] pytest should not skip tests for dependencies

rohitnig:
## Description

pytest --error-for-skips for dependencies

## Motivation and Context

In [1041](https://github.com/pytorch/rl/pull/1041), we forced the tests not to be skipped in the CI for envpool.
This should be the rule for all tests under .circleci/unittest/linux_libs, where pytest should always be called with --error-for-skips. This way, we can make sure that dependencies are always tested.
The situation we want to avoid is the one in which a dependency fails to be installed, and the related tests are all skipped because the library cannot be found

## Types of changes

- New feature (non-breaking change which adds core functionality)
- Example - https://github.com/pytorch/rl/pull/1041

## Checklist

- [ x ] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)


.circleci/unittest/linux_libs/scripts_gym/run_test.sh
```diff
@@ -21,8 +21,8 @@ lib_dir="${env_dir}/lib"
 export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$lib_dir
 export MKL_THREADING_LAYER=GNU
 
-python .circleci/unittest/helpers/coverage_run_parallel.py -m pytest test/smoke_test.py -v --durations 20
+python .circleci/unittest/helpers/coverage_run_parallel.py -m pytest test/smoke_test.py -v --durations 20 --error-for-skips
 python .circleci/unittest/helpers/coverage_run_parallel.py -m pytest test/smoke_test_deps.py -v --durations 20 -k 'test_gym or test_dm_control_pixels or test_dm_control'
-MUJOCO_GL=egl python .circleci/unittest/helpers/coverage_run_parallel.py -m pytest --instafail -v --durations 20 -k 'test_libs'
+MUJOCO_GL=egl python .circleci/unittest/helpers/coverage_run_parallel.py -m pytest --instafail -v --durations 20 -k 'test_libs' --error-for-skips
```
vmoens, line 26: Looking at the errors, we should select tests that contain `Gym` or `gym`
For some reason the tests don't see the cuda device, not sure why...

vmoens (commented):
As anticipated gym tests are the ones that are the most painful to change.
We should select the appropriate tests using `-k "pattern1 AND pattern2"` or similar.
We should also figure out why the cuda device is not found

rohitnig:
The CUDA errors seems like a legacy error. It has been happening in previous builds as well. The envpool test ran and failed with the error of "RuntimeError: Could not infer dtype of dict", as it happened before. Also, the GYM test has run and failed - { test_dm_control_pixels} - Forbidden skipped test - Skipped: Not implemented yet

Issue #875, opened 2023-01-27
[BUG] NoisyLinear uses self.training to turn on/off noise


seermer:
## Describe the bug

https://github.com/pytorch/rl/blob/2fd46326e2eb8d155258feed81dbce1574f527f0/torchrl/modules/models/exploration.py#L131

In PyTorch, `module.training` is used to differentiate training and inference. For example, Normalization layers use training attribute because they should behave differently when training and inference. However, NoisyLinear is not the case, NoisyLinear should stay active both during training and during inference (actually this is the whole point of NoisyLinear, it "learns" noisy parameters during training, and uses noise for exploration during inference). The only time NoisyLinear should be turned off is during policy evaluation/testing. (evaluation is not the same as inference in RL, evaluation should always be inference, while inference can be either evaluation or exploration).

In the current implementation, to correctly allow the policy to explore, `module.training` must be turned on even for inference during exploration, however, turning on training mode will unavoidably affect normalization layers, dropout, etc. Although Dropout is not widely used in RL yet, normalization layers are sometimes being used in recent years (especially in some works that try to use Transformers in RL). I personally think using training attribute, in this case, is a bug.

## Expected behavior

NoisyLinear should be activated during exploration/training, and deactivated during evaluation/testing. Other pytorch modules like normalization should not be impacted by the behavior of NoisyLinear.

## Reason and Possible fixes

possible fix: add an additional function or attribute specifically for NoisyLinear to turn on and off noise (just like how its reset_noise resets noise)

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)
- [x] I have read the [documentation](https://github.com/pytorch/rl/tree/main/docs/) (**required**)
- [ ] I have provided a minimal working example to reproduce the bug (**required**)

vmoens:
That's a good discussion to have. I agree with your suggestion by the way.

In RL the use cases can be a bit different though: you may use your policy
- during inference (collecting data for training)
- during test (collecting data to check learning)
- during training (executing on trained data)
- during simulation (e.g. in model-based or planning)

I'm not sure that the `.train()` vs `.eval()` reflects perfectly this. 
For instance: I think that we want the batch-norm (if any) to be turned off during inference and test, but exploration should be turned on during inference only. 

In TorchRL we use a separate global variable to turn on exploration (`@set_exploration_mode("random")`). 
Maybe we could use that for NoisyLinear?

The other option is to write a function like
```
def set_noisy_off(module):
[...]

PR #1060, opened 2023-04-14
[Feature,Doc] QValue refactoring and QNet + RNN tuto

vmoens:
- init
- amend

## Description

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/torchrl/modules/tensordict_module/actors.py
+++ b/torchrl/modules/tensordict_module/actors.py
@@ -3,10 +3,11 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Optional, Sequence, Tuple, Union
+from typing import Optional, Sequence, Tuple, Union, List
 
 import torch
 from tensordict.nn import TensorDictModule, TensorDictModuleWrapper
+from tensordict import TensorDictBase
 from torch import nn
 
 from torchrl.data.tensor_specs import CompositeSpec, TensorSpec
@@ -311,6 +312,8 @@ class ValueOperator(TensorDictModule):
         )
 
 
+
+
 class QValueHook:
     """Q-Value hook for Q-value policies.
 
@@ -326,6 +329,12 @@ class QValueHook:
         var_nums (int, optional): if ``action_space = "mult_one_hot"``,
             this value represents the cardinality of each
             action component.
+        in_keys (list of str or tuple of str, optional): to be used when hooked on
+            a TensorDictModule. The input key representing the action value. Defaults
+            to ``["action_value"]``.
+        out_keys (list of str or tuple of str, optional): to be used when hooked on
+            a TensorDictModule. The output keys representing the actions, action values
+            and chosen action value. Defaults to ``["action", "action_value", "chosen_action_value"]``.
 
     Examples:
         >>> import torch
@@ -358,6 +367,8 @@ class QValueHook:
         self,
         action_space: str,
         var_nums: Optional[int] = None,
+        in_keys: Union[List[str], List[Tuple[str]]] = None,
+        out_keys: Union[List[str], List[Tuple[str]]] = None,
     ):
         self.action_space = action_space
         self.var_nums = var_nums
@@ -374,19 +385,46 @@ class QValueHook:
             raise ValueError(
                 f"action_space must be one of {list(self.action_func_mapping.keys())}"
             )
+        if in_keys is None:
+            in_keys = ["action_value"]
+        self.in_keys = in_keys
+        if out_keys is None:
+            out_keys = ["action", "action_value", "chosen_action_value"]
+        self.out_keys = out_keys
 
     def __call__(
         self, net: nn.Module, observation: torch.Tensor, values: torch.Tensor
     ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
-        if isinstance(values, tuple):
-            values = values[0]
-        action = self.action_func_mapping[self.action_space](values)
+        td_input = isinstance(values, TensorDictBase)
+        if td_input:
+            action_values = values.get(self.in_keys[0], None)
+            if action_values is None:
+                raise KeyError(
+                    f"Action value key {self.in_keys[0]} not found in {values}."
+                    )
+        else:
+            if isinstance(values, tuple):
+                action_values = action_values[0]
+            else:
+                action_values = values
+
+        action = self.action_func_mapping[self.action_space](action_values)
 
         action_value_func = self.action_value_func_mapping.get(
             self.action_space, self._default_action_value
         )
-        chosen_action_value = action_value_func(values, action)
-        return action, values, chosen_action_value
+        chosen_action_value = action_value_func(action_values, action)
+        if not td_input:
+            return action, action_values, chosen_action_value
+        values.update(
+            dict(
+                zip(
+                    self.out_keys,
+                    (action, action_values, chosen_action_value)
+                    )
+                )
+            )
+        return values
 
     @staticmethod
     def _one_hot(value: torch.Tensor) -> torch.Tensor:
@@ -397,7 +435,11 @@ class QValueHook:
     def _categorical(value: torch.Tensor) -> torch.Tensor:
         return torch.argmax(value, dim=-1).to(torch.long)
 
-    def _mult_one_hot(self, value: torch.Tensor, support: torch.Tensor) -> torch.Tensor:
+    def _mult_one_hot(
+        self,
+        value: torch.Tensor,
+        support: torch.Tensor
+        ) -> torch.Tensor:
         values = value.split(self.var_nums, dim=-1)
         return torch.cat(
             [
@@ -571,11 +613,6 @@ class QValueActor(Actor):
             contains more than one element, the values will be passed in the
             order given by the in_keys iterable.
             Defaults to ``["observation"]``.
-        out_keys (iterable of str): keys to be written to the input tensordict.
-            The length of out_keys must match the
-            number of tensors returned by the embedded module. Using "_" as a
-            key avoid writing tensor to output.
-            Defaults to ``["action"]``.
         spec (TensorSpec, optional): Keyword-only argument.
             Specs of the output tensor. If the module
             outputs multiple output tensors,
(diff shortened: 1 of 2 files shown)


vmoens:
Note: there is still some magic that needs to be done when we pass 2 consecutive trajectories to the RNN, but it can be handled without too much trouble (I think)
@matteobettini I will do that with padding. Don't tell me, I know!

torchrl/modules/tensordict_module/exploration.py
```diff
@@ -447,17 +447,19 @@ def step(self, frames: int = 1) -> None:
...
                 warnings.warn(
                     f"The tensordict passed to {self.__class__.__name__} appears to be "
-                    f"missing the 'step_count' entry. This entry is used to "
+                    f"missing the 'is_init' entry. This entry is used to "
                     f"reset the noise at the beginning of a trajectory, without it "
                     f"the behaviour of this exploration method is undefined. "
                     f"This is allowed for BC compatibility purposes but it will be deprecated soon! "
-                    f"To create a 'step_count' entry, simply append a StepCounter "
-                    f"transform to your environment with `env = TransformedEnv(env, StepCounter())`."
+                    f"To create a 'step_count' entry, simply append an torchrl.envs.InitTracker "
```
matteobettini, line 457: Replace step_count with is_init?

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -0,0 +1,234 @@
...
+        lstm (torch.nn.Module): a :class:`torch.nn.LSTM` instance. For simplicity, this class does not
+            handle the construction of the module.
+        in_keys (list of str): a triplet of strings corresponding to the input value,
+            first and second hidden key.
+        out_keys (list of str): a triplet of strings corresponding to the output value,
+            first and second hidden key.
+            .. note::
+              For a better integration with TorchRL's environments, the best naming
+              for the output hidden key is ``("next", "hidden0")`` or similar, such
+              that the hidden values are passed from step to step during a rollout.
```
matteobettini, line 47: Why are we asking the keys like this?

In my mind i would have 3 key arguments:
- in_keys # in keys apart from hidden state (not constrained to 1)
- out_keys # out keys apart from hidden state (not constrained to 1)
- hidden_keys = (hidden0, hidden1) # this is by default like this and the init automatically adds these as input keys and the version of these with preappended next to the output keys

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -0,0 +1,234 @@
...
+
+    Examples:
+        >>> from torchrl.envs import TransformedEnv, InitTracker
+        >>> from torchrl.envs.libs.gym import GymEnv
+        >>> from torchrl.modules import MLP
+        >>> from torch import nn
+        >>> from tensordict.nn import TensorDictSequential as Seq, TensorDictModule as Mod
+        >>> env = TransformedEnv(GymEnv("Pendulum-v1"), InitTracker())
+        >>> lstm = nn.LSTM(input_size=env.observation_spec["observation"].shape[-1], hidden_size=64, batch_first=True)
+        >>> lstm_module = LSTMModule(lstm, in_keys=["observation", "hidden0", "hidden1"], out_keys=["intermediate", ("next", "hidden0"), ("next", "hidden1")])
```
matteobettini, line 64: With the comment above this becomes
`LSTMModule(lstm, in_keys=["observation"], out_keys=["intermediate"])`

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -0,0 +1,234 @@
...
+            fields={
+                action: Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.float32, is_shared=False),
+                done: Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.bool, is_shared=False),
+                intermediate: Tensor(shape=torch.Size([64]), device=cpu, dtype=torch.float32, is_shared=False),
+                is_init: Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.bool, is_shared=False),
+                next: TensorDict(
+                    fields={
+                        hidden0: Tensor(shape=torch.Size([1, 64]), device=cpu, dtype=torch.float32, is_shared=False),
+                        hidden1: Tensor(shape=torch.Size([1, 64]), device=cpu, dtype=torch.float32, is_shared=False)},
+                    batch_size=torch.Size([]),
```
matteobettini, line 78: Isn't it counterintuitive that 
obs -> intermediate -> action are all at root level but the hidden state is in next?

What again is the reason for this?

matteobettini (commented):
Looks good, I left some questions!

- I like that the 2 hidden states are kept separate and I think that is the best choice
- is_init is used nicely in my opinion. We need a flag that needs when to reset and it does exactly that.

matteobettini:
> EDIT: Problem solved, feature implemented!!

We can't escape from padding in rnns right?

vmoens:
> We can't escape from padding in rnns right?

There is the packed_sequence artifact but i'm not entirely sure that it solves the problem.
My implementation of padding / unpadding is 2x faster than other torch tools btw

torchrl/modules/tensordict_module/rnn.py
vmoens: the hidden is not used at time t but t+1 (obs, intermediate and action are for t and t only). It's probably a bit cleaner to have it in next as this is where it belongs... 

so you pack hidden in next such that you find it back when calling step_mdp
Per se we could avoid that, since I think most calls to step_mdp have keep_other=True (it is the default, it is the case for env.rollout and it is the case for the collectors!).
But if someone by mistake calls step_mdp with keep_other=False she'll loose the hidden, and since we populate the hidden with 0s when not found no error will be raised (outch!). 

So i'd rather avoid that scenario as much as I can!

Now i'm not against the idea of having it out of "next" if you find it's clearer.

matteobettini: I see. So we put it in next because we know for sure that a step_mdp will happen.

vmoens: in the algo yes but it's your choice...

smorad:
Awesome, really looking forward to this!

I think it makes sense to write the recurrent state to `next`. It just might make things tricky with when/who is calling `step_mdp`. Are there any footguns here, e.g. collector calling `step_mdp` and reversing `state, (next, state)` before placing them into the replay buffer?

Would it make sense to use `td[(prev, done)]` instead of `is_init`? Perhaps there are some edge cases I haven't considered, or maybe this is tricky to implement.

Opinion: I think `set_temporal_mode` could be a little confusing. Is this different than inference (collector) mode, or just a synonym for inference? If they are indeed the same, maybe it makes sense to have `if self.eval: ...` in the module forward pass instead of an explicit time-batch/single-step mode.

I would really prefer that `packed_sequence` is not used here. It's brought me nothing but pain.

torchrl/modules/tensordict_module/rnn.py
smorad: I have to double-check, but I think RNNs are usually written as

`y_t, s_t = rnn(x_t, s_{t-1})`

in books/papers. I agree that perhaps `prev_state, state` makes more sense than `state, next_state`.

vmoens: There's what papers say and how it aligns with the rest of the API.
ie. I'm not sure how we could mark the input as being "prev" and the current as being "current" if the only containers we have are "current" and "next". Maybe a proper doc that mirrors what you're saying here?

vmoens:
> Awesome, really looking forward to this!
> 
> I think it makes sense to write the recurrent state to `next`. It just might make things tricky with when/who is calling `step_mdp`. Are there any footguns here, e.g. collector calling `step_mdp` and reversing `state, (next, state)` before placing them into the replay buffer?

If you want to store every `h_{t-1}, s_t, a_t, h_t, s_{t+1}` following the nomenclature above, this will correspond to

```
TensorDict(
    fields={
        hidden: Tensor(shape=torch.Size([4, 16]), device=cpu, dtype=torch.int64, is_shared=False),
        action: Tensor(shape=torch.Size([4, 1]), device=cpu, dtype=torch.int64, is_shared=False),
        observation: Tensor(shape=torch.Size([4, 2]), device=cpu, dtype=torch.bool, is_shared=False),
        next: TensorDict(
            fields={
                hidden: Tensor(shape=torch.Size([4, 16]), device=cpu, dtype=torch.int64, is_shared=False),
                observation: Tensor(shape=torch.Size([4, 2]), device=cpu, dtype=torch.bool, is_shared=False),
            batch_size=torch.Size([4]),
            device=cpu,
            is_shared=False),
    batch_size=torch.Size([4]),
    device=cpu,
    is_shared=False)
``` 

> Opinion: I think `set_temporal_mode` could be a little confusing. Is this different than inference (collector) mode, or just a synonym for inference? If they are indeed the same, maybe it makes sense to have `if self.training: ...` in the module forward pass instead of an explicit time-batch/single-step mode.

I agree it's not ideal.
First, we had a similar discussion [here](https://github.com/pytorch/rl/issues/875) if your want to have a look.

First, let's put the cards on the table. There are various things that are related but not exactly identical:
- Train / eval mode
- Explorative / exploitative mode
- temporal / static mode (eg example here)
- grad on / off mode (eg example here)
When doing inference and collecting data for training you will be in 
  eval + explorative + static + grad off
When testing the policy in "real" setting you will be 
  eval + exploitative + static + grad off
When training the policy you will be in
  train + exploitative* + temporal** + grad on***

* not always true, sometimes you want to sample the action, sometimes not
** not always true: in dreamer you simulate trajectories during training that are fully reparametrized, and you're in static mode bc the RNN feeds itself with the action that results from the previous call.
*** depending on the loss (policy / critic / model in MBRL) you may want to keep some ops out of the graph or just detach some variables (not talking about meta-RL!)

Gradient and exploration are currently handled via "exploration_type" and the torch "grad_mode" environment variables.
Train and eval are handled via the module attribute. I hesitate for temporal, but I can see both cases where you may want to use static / temporal at both training and inference time. I thought that giving the user full power over what happens internally was the easiest option to avoid any kind of confusion.

If it's about naming i'm happy to reconsider the name oc. I'm actually looking forward to have more input on this from the community: what's the best way of approaching all of these "modes" that -- to me -- seem very algorithm-dependant. 
Just think about SARSA vs QMax, where the action selection is completely different. We can easily do that via "exploration_type", but it's neither related to eval/train, gradient off/on or else.

> I would really prefer that `packed_sequence` is not used here. It's brought me nothing but pain.

Got it, it's a headache on our side too to let's just not use it.

smorad:
That tensordict format looks good to me.

Yeah that's a good point with the modes. In that case it's better to err on the side of flexibility than simplicity. I think perhaps a small table in the docs mapping each of these modes to a 1-2 sentence description would help clear up any ambiguities. I do like the way exploration etc. is currently handled. Is `set_recurrent_mode` more specific than `set_temporal_mode`? Could either be `batch_time` or `sequential_time`.

nit: `hidden/state` are super overloaded terms. I think we should be explicit with `recurrent_state_h, recurrent_state_c` or `rstate_h, rstate_c`, which will help new users when they are grepping through the codebase.

PR #1069, opened 2023-04-17
[CI] Remove migrated CircleCI macOS jobs

seemethere:
This removes jobs that were migrated as part of #886, this PR finally removes it


vmoens (approved):
LGTM
If we need to, we can have a single python version tested here

Issue #1000, opened 2023-03-29
[Feature Request]  Decouple losses from targets


matteobettini:
Currently the losses in torchrl compute the value target when forward is called on them.

The problem is that if the loss is called on minibatches, the target will be computed each time for each menibatch.

This is extremely inefficient as targets can be precomputed at the beginning of the training iteration

I am proposing that losses should all have a separate function 
`loss.compute_value_target(tensordict)` which writes the target to the tensordict

```python
batch = rb.sample(60_000)
loss_module.compute_value_target(batch)
for _ in range(n):
    minibatch = subsample(batch, 1000)
    loss_vals = loss_module(minibatch)
```

the forward function of the loss module will then check if the target is present and, if not, will call `loss.compute_value_target(tensordict)`.

Furthermore, in this restructuring, value estimators would be made independent of neural networks and jsut assume that they are given a tensordict with all the desired keys and write in it the new keys.

This whole update would allow to do something like

<img width="647" alt="Screenshot 2023-03-29 at 09 39 58" src="https://user-images.githubusercontent.com/55539777/228477183-53890150-66ac-446c-9125-987f13ee8c30.png">

Here, where line 12 is currently not possible in torch rl

smorad:
So in the algorithm you linked, I believe you need to recompute targets for each minibatch because you are using polyak averaging (line 15). So the targets should change slightly with each update.

That said, I think separation between target computation and value loss would be pretty cool for implementing custom algorithms. Then you could override DDPG/SAC do whatever harebrained target scheme you want (e.g. TQC, etc.). In general, I think it would be beneficial to have more decoupling in the loss functions. That way, we can try small changes without rewriting `SACLoss.forward` and introducing potential bugs.

vmoens:
I agree that's a feature we can work on
And it can be done on a loss-by-loss basis

```
class SACLoss(...):
    def forward(self, td):
        # some preproc
        value_loss = self._value_loss(td_preproc)
        actor_loss = self._actor_loss(td_preproc)
        return TD({...}, []) # as it is now
     def actor_loss(td):
        # some preproc
        actor_loss = self._actor_loss(td_preproc)
        return actor_loss
     def value_loss(td):
        # some preproc
        value_loss = self._value_loss(td_preproc)
        return actor_loss
```
Would this work?

matteobettini:
I think also this could be applyable to all losses

smorad:
[...]

PR #1070, opened 2023-04-18
[Feature] Skip existing for advantage modules

vmoens:
## Description

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -4980,6 +4980,62 @@ class TestAdv:
         td = module(td.clone(False))
         assert td["advantage"].is_leaf
 
+    @pytest.mark.parametrize(
+        "adv,kwargs",
+        [
+            [GAE, {"lmbda": 0.95}],
+            [TD1Estimator, {}],
+            [TDLambdaEstimator, {"lmbda": 0.95}],
+        ],
+    )
+    @pytest.mark.parametrize("has_value_net", [True, False])
+    @pytest.mark.parametrize("skip_existing", [True, False, None])
+    def test_skip_existing(
+        self,
+        adv,
+        kwargs,
+        has_value_net,
+        skip_existing,
+    ):
+        if has_value_net:
+            value_net = TensorDictModule(
+                lambda x: torch.zeros(*x.shape[:-1], 1),
+                in_keys=["obs"],
+                out_keys=["state_value"],
+            )
+        else:
+            value_net = None
+
+        module = adv(
+            gamma=0.98,
+            value_network=value_net,
+            differentiable=True,
+            skip_existing=skip_existing,
+            **kwargs,
+        )
+        td = TensorDict(
+            {
+                "obs": torch.randn(1, 10, 3),
+                "state_value": torch.ones(1, 10, 1),
+                "next": {
+                    "obs": torch.randn(1, 10, 3),
+                    "state_value": torch.ones(1, 10, 1),
+                    "reward": torch.randn(1, 10, 1, requires_grad=True),
+                    "done": torch.zeros(1, 10, 1, dtype=torch.bool),
+                },
+            },
+            [1, 10],
+        )
+        td = module(td.clone(False))
+        if has_value_net and not skip_existing:
+            exp_val = 0
+        elif has_value_net and skip_existing:
+            exp_val = 1
+        elif not has_value_net:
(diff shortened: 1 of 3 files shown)


albertbou92:
What about the other way around? by default not skipping.

In that case we should make sure in the advantage module the value is computed without gradients.

matteobettini:
Could you explain a bit more why your example constitutes a problem?
Wouldn't the value computed during rollout be the same if we just compute it later in the loss?

Also I wonder if this is a small step in the direction of #1000.
Because here we are allowing to precompute values  (maybe on bigger batches) outside losse forward.
And in #1000 we want to allow also precomputing targets (maybe on bigger batches) outside loss forward.

vmoens:
> Could you explain a bit more why your example constitutes a problem? Wouldn't the value computed during rollout be the same if we just compute it later in the loss?
> 
> Also I wonder if this is a small step in the direction of #1000. Because here we are allowing to precompute values (maybe on bigger batches) outside losse forward. And in #1000 we want to allow also precomputing targets (maybe on bigger batches) outside loss forward.
> 
> PS in #1000 advantages would not even have access to neural networks. Only losses will. and thus it will make all this even simpler

you can perfectly not pass a network and it will work fine.
I'd like to keep the networks bc it's much easier, otherwise we are decomposing things and we'll be asking users to do
```
value_net(td)
value_net(td.get("next"))
gae(td)
```
instead of 
```
gae(td)
```

matteobettini:
Yep i didn't see that we could pass None, that works perfectly.

Thinking about this more, if losses start skipping existing keys by default, this may have some bad consequences if collected data (from collectors or buffers) has some of the loss keys of interest (cause they might be old).

But thinking of #1000 we definitely need a way to precomupte keys for losses (values and targets) and make losses use those precomputed keys.

I think it would be fine to use that skip_existing by default also in the losses and if users feed the losses data with old values it is their problem.
```
batch = rb.sample(60_000)
loss_module.compute_value_target(batch)
for _ in range(n):
    minibatch = subsample(batch, 1000)
    loss_vals = loss_module(minibatch) # by default uses the precomputed targets
```

Alternatively, what @albertbou92 says also makes sense, the losses could recompute everything by default and we would need to
```
batch = rb.sample(60_000)
loss_module.compute_value_target(batch)
for _ in range(n):
    minibatch = subsample(batch, 1000)
    loss_vals = loss_module(minibatch, skip_exisitng=True) # we need to enforce it to not recompute the targets
```

I am leaning more towards the first option tho

vmoens:
> Thinking about this more, if losses start skipping existing keys by default, this may have some bad consequences if collected data (from collectors or buffers) has some of the loss keys of interest (cause they might be old).

yes this is why I set it to False, and why I was asking you guys what you were thinking. I guess we'll leave it as it is.

> But thinking of #1000 we definitely need a way to precomupte keys for losses (values and targets) and make losses use those precomputed keys.

if you don't pass the value net to the module an error will be raised bc the entry is missing.

> I think it would be fine to use that skip_existing by default also in the losses and if users feed the losses data with old values it is their problem.

yep that's the idea.
There are 2 things you could do
```
with set_skip_existing(True):
    loss_fn(td)
```
or 
```
loss_fn.make_value_estimator(..., skip_existing=True)
```

matteobettini:
I am more for the first of the 2 since you are not bound to smth at init time.

Regarding the default of skip_existing i am leaning more towards True but I am fine with anything

vmoens:
> I am more for the first of the 2 since you are not bound to smth at init time.

then use that one 😉


> Regarding the default of skip_existing i am leaning more towards True but I am fine with anything

Let's put it that way:
worst that can happen when True: wrong value -> no training -> torchrl sucks
worst that can happen when False: slow training -> good training, but slow -> issue

albertbou92:
> > I am more for the first of the 2 since you are not bound to smth at init time.
> 
> then use that one 😉
> 
> > Regarding the default of skip_existing i am leaning more towards True but I am fine with anything
> 
> Let's put it that way: worst that can happen when True: wrong value -> no training -> torchrl sucks worst that can happen when False: slow training -> good training, but slow -> issue

Yes  I like the idea of having something that works out of the box, even if there is later on margin for improvement in terms of performance and requires being a bit of an advanced user

PR #1076, opened 2023-04-20
[Example] A2C simplified example

albertbou92:
## Description

Add a simplified version of the A2C code example, including some plot results.

## Motivation and Context

## Types of changes

## Checklist


examples/a2c/a2c.py
```diff
@@ -2,164 +2,144 @@
...
+                    auto_reset=True,
+                    auto_cast_to_device=True,
+                    break_when_any_done=True,
+                ).clone()
+                logger.log_scalar(
+                    "reward_testing",
+                    td_test["next"]["reward"].sum().item(),
+                    collected_frames,
+                )
+                actor.train()
```
vmoens, line 142: Nit: If we use the ``Recorder`` class we can do all of this I guess, but it's a bit of a black box so I'm ok with the explicit calls.
Maybe let's we use `td_test["next", "reward"]` when we can (same for all the key indexing in the script) :)

albertbou92: I saw that in Recorder you need to specify a number of steps, and I wanted to record a single test episode, independently of the number of steps. Maybe recorder could accept a number of episodes to records instead of a number steps?

vmoens (approved):

PR #1080, opened 2023-04-24
[BugFix] Target return for sequential td

BY571:
## Description

Target return calculation for sequential tensordict with time dimension did not work properly.

## Motivation and Context

When using a sequential tensordict with time dimension (batch, time, feature)  the reward was subtracted over the time dimension which is wrong. It should only subtract the reward from the most recent target return in time.  

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -1066,7 +1066,10 @@ class TargetReturn(Transform):
         self, reward: torch.Tensor, target_return: torch.Tensor
     ) -> torch.Tensor:
         if self.mode == "reduce":
-            target_return = target_return - reward
+            if len(target_return.shape) == 3:
+                target_return[:, -1] -= reward
+            else:
+                target_return = target_return - reward
             return target_return
         elif self.mode == "constant":
             return target_return


vmoens (commented):
In general the time dimension is the last of a tensordict, but it isn't as easy as that.

It's hard to tell if a tensordict is multidimensional because you're passing a batch of vectorized tds (for instance) or if it has multiple dimensions.
e.g. you could have nested parallel envs
```
ParallelEnv(3, ParallelEnv(4, make_env))
```
(same for distributed data collectors with batched environment)
This which will output tds of shape [3, 4]. The rewards will have shape [3, 4, 1]. So will the target return.
Now if you do a rollout you'll have target returns of shape [3, 4, T, 1].
In general it's hard to know what the time dimension is :( 


There *could* be a way of doing that by having named dimensions in tensordict but I believe that the ["named dimension"](https://pytorch.org/docs/stable/named_tensor.html) of pytorch is not something we want to support for very long (it's a bit messy and hard to maintain). So not sure that integrating it now in tensordict would be wise. But we can talk about it. That way we could also keep track in other modules if the last dim of a tensordict is indeed marked as time or not.
See the discussion [here](https://github.com/pytorch/rl/pull/1060) about using LSTMs for temporal and non temporal data.
@smorad and @matteobettini wdyt of the idea of stamping tensordicts with a "time" dimension flag to instruct modules of what to do with the data in sequential contexts such as this one?
It you be smth like 
```python
td = TensorDict({...}, [3, 4], names=["batch", "time"])
# later in the code
if td.names[-1] == "time":
   foo()
else:
  bar()
```

vmoens:
cc @apbard @tcbegley: curious to hear what you think of having named dims in tensordict

matteobettini:
I think if we introduce named dimensions now the whole thing will become extremely hard to maintain.

the current convention on tensordict key shapes is:

A td has shape *B
A key inside has shape *B, *F

When a time dimension is needed it will be taken as the last of *B (the other dims in B can be many and it is difficult to assume what they are).

*F is also multidim (for example it could have a multiagent dimension)

vmoens:
> I think if we introduce named dimensions now the whole thing will become extremely hard to maintain.

Why so? Not saying it won't but on the RL side the changes seem small to me and not bc-breaking.

We'd just flag the last dimension of the tensordicts that come from the collector and rollout methods with a dim named "time". All others would be kept to None.

Something like appending this in `EnvBase.rollout` and `collector.rollout`:
```python
tensordict = tensordict.refine_names(*[None] * (tensordict.ndim-1), "time")
return tensordict
```

which are our main sources of temporal data. Any reshape / view operation on a tensordict would make the "time" disappear.

Also, I don't see the worry with multidimensional features?

matteobettini:
If we add a flag for the time dimension i feel like users will start requesting flags for other dimensions. Then modules and losses may start introducing more if checks on these things

apbard:
I suspect it won't be easy...can dims be multidimensional?
As mentioned elsewhere having "named" objects tensors or td modules could bring some benefits...but we need to weight correctly pros and cons

vmoens:
> If we add a flag for the time dimension i feel like users will start requesting flags for other dimensions. Then modules and losses may start introducing more if checks on these things

In a way users could add the flags if they feel it's proper. For instance, someone working on an env could name a dim of the tensordict with the name she wants, we'd support that in tensordict. 
That doesn't mean we need to incorporate it in a pervasive way.
I agree I'd be cautious about it though, could be a pandora box

BY571:
I like the idea of named tensors! But if adding them would cause big changes it might not be worth it for now. However, I think it would give new users some clarity regarding the structure and organization of tensordicts and a tool to handle multiple dimensions. 

A general question why would we keep tds with nested parallel envs in the shape of for example (3, 4, 1) as you mentioned above and not reshape them to (3*4, 1)?

tcbegley:
I'm in favour of supporting named dimensions. As Vincent said I don't see why there should be any bc-breaking change here since it would be opt-in, but potentially it makes working with tensordicts more convenient in some places.

The thing that I think we'll need to be careful of is if a library like TorchRL is choosing names for particular dimensions. There's a risk of clashing names chosen by the library and names chosen by the end-user. We'll just have to make sure that any automatic labelling of dimensions inside loss functions etc. is well documented, consistent across the library, and that the names are well chosen.

BY571:
> > A general question why would we keep tds with nested parallel envs in the shape of for example (3, 4, 1) as you mentioned above and not reshape them to (3*4, 1)?
> 
> Because we don't know where the multi-dim comes form and if it has some structural meaning. In regular, single-objective, single task, single agent training, we'd expect that the batch could be squashed to 1 dimension always leaving the time in the second, but the collector does not know that. Per se the collector is oblivious to what the envs are doing, it just stacks temporal data along the last dimension. I'm pretty sure that if from tomorrow we output data that is always 2d we'll make some of our users unhappy!

Thanks for the clarification! 

I also agree with @tcbegley names should be very clear, intuitive, and well-documented if we set some default names in TorchRL. But I think the only named dimension really needed as default would be the time dimension (if existing in the td) and as @vmoens  said the rest could be `None` as default or set by the user.

vmoens:
> I'm in favour of supporting named dimensions. As Vincent said I don't see why there should be any bc-breaking change here since it would be opt-in, but potentially it makes working with tensordicts more convenient in some places.

Exactly! As of now the way I see it is that we would prototype the feature by adding it to tensordict and marking the dim names wherever we see fit. We'd make sure that at no point in the code this feature is mandatory, eg. for the rnn you could still turn it manually on temporal or non-temporal mode. 

Here's a list of papercuts I had in the past days that makes me think that this could be useful:
- One would expect that `buffer.extend(next(collector))` would always do the same thing but it is not the case if the collector has one or many envs (either because the collector or the env is multiprocessed). That means that you expect to get trajectories out of your buffer and then get single steps, without warning of any kind. If we could create a buffer that explicitly looks for the "time" stamp in the data, we would avoid such silent errors. Again this would be a simple opt-in, something like `buffer.append_transform(TimeChecker())`.
- the LSTM example above: avoid duplicating instances depending on the behaviour that we want
- run checks in time-based classes: some losses, advantage functions, etc. in order to make sure that the data that is being passed meets the requirements. Not necessarily raising an exception but a warning that says "you gave me data that is not time-stamped, I'll assume the last dim is the time but make sure you're doing things as expected". The impacted classes would include target returns, random crop, etc. Other classes may also check that there is no time dimension already.

smorad:
What I've seen used in other libraries like `torch_geometric` is the use of an `dim` parameter which uses a sensible default __and indexes from the right__. E.g.

```
def td_error(x, target, dim=-1): # Default can handle arbitrary number of leading/batch dims
  ...
```

In this way, you can have an arbitrary number of leading dims, since like 90% of these functions just operate over the feature or time dimension anyways.

I'm not sure if this applies here, but I think it's an elegant and low-cost change to plumb through whatever other helper functions need this.

vmoens:
Thanks @smorad 
The context here is slightly different though IIUC. The issue that this PR tries to solve is that whether the transform is used in an online setting (ie appended to an environment) or offline (ie appended to a replay buffer) we expect data that is either static or temporal in nature. In both cases, it can be multidimensional and there is no way to tell in general if the last dimension is or isn't "time".

gtnv:
hi @BY571, do you still plan to work on this? if not, can we close it?

PR #1082, opened 2023-04-24
[BugFix] Change default of skip_existing to None

tcbegley:
This is the behaviour stated in the docstring, and the current default value of `False` makes it impossible to override `skip_existing` with the `set_skip_existing` context manager.


vmoens (approved):
LGTM

PR #1093, opened 2023-04-26
[BugFix] Multi-agent exploration

matteobettini:
Currently, the exploration classes assume that the actions have shape `[tensordict.batch_size, n_actions]`

In multiagent settings, the actions have shape `[tensordict.batch_size,  n_agents, n_actions]`

Therfore the exploration classes need to handle this use case.

In this PR we modify the `exploration.py` to consider the last dimension of the action_spec as the n_actions.

The problem here is that we are not able to determine the difference between multi-agent settings and action spec with multiple action dimensions.

For example if an env has action spec:
- `[*batch, n_agents, n_actions]`
- or
- `[*batch, n_actions_1, n_actions_2]`

we are not able to differentiate among the 2 cases and we will take independent exploratory actions for all the dimensions which are not the last.

Another possibility is to just consider all dimensions in the exploration wrappers. 
e.g.
```
torch.rand(out.shape
```
and
```
torch.Size([*tensordict.get(self.key).shape
```
In this case for example we will take a greedy action independently for each action in the space


vmoens:
> For example if an env has action spec:
> 
> * `[*batch, n_agents, n_actions]`
> * or
> * `[*batch, n_actions_1, n_actions_2]`
> 
> we are not able to differentiate among the 2 cases and we will take independent exploratory actions for all the dimensions which are not the last.
> 

I might be missing something, but wouldn't the solution to be to present the data for MARL with the following structure:
```python
data = TensorDict({
   "agent_features": TensorDict({
      "obs": torch.Tensor(B, T, A, F1),
      "action": torch.Tensor(B, T, A, F2),
    }, [B, T, A]),
    "common_features": TensorDict({
      "stuff": torch.Tensor(B, T, F3),
    }, [B, T]),
)
```
where B is our batch, T the time, A the number of agents
In this way the tensordict size always separated batch and feature size.

matteobettini:
Yeah the data is in that structure, but the current exploration is not compatible with that. 

The rand in eps_greedy has to also span over the agent dim

and the pther noise wrapper simply crashes without the proposed changes

matteobettini:
If you are asking why not having those nested tensordict, we currently don’t use those nested keys and forcing them i think is not clean.

ie i still want action and done and pthers to be at the first level of a tensordict in multiagent

matteobettini:
Just to recap:

Currently, multi-agent data is moved around like normal data as
```python
data = TensorDict(
     "obs": torch.Tensor(*B, T, A, F1),
     "action": torch.Tensor(*B, T, A, F2),
     "stuff": torch.Tensor(*B, T, F3), # shared
     "done":  torch.Tensor(*B, T, A, 1),
      batch_size=  [*B, T],
)
```
This works with all the componenets in torchrl almost seamlessly. (For the shared stuff users wiill have to use that as they want in the model input)

Your are proposing to change the data carrier to 
```python
data = TensorDict({
   "agent_features": TensorDict({
      "obs": torch.Tensor(B, T, A, F1),
      "action": torch.Tensor(B, T, A, F2),
    }, [B, T, A]),
    "common_features": TensorDict({
      "stuff": torch.Tensor(B, T, F3),
    }, [B, T]),
)
```

Here are some reasons why I think it is better not to do this:
- if i have a done or reward, these may be shared or not and thus sometimes i will find them under agent_features and sometimes under common_features
- All the modules and collections and env and so on ( the whole codebase) will have to be adapted to this specific format, as now it cannot find the keys in the root level


I think there are better solutions to identify the difference between the agent dimension and the batch and feature dimensions (which can be more seamlessly integrated in the library).

vmoens:
> I think there are better solutions to identify the difference between the agent dimension and the batch and feature dimensions.

I'm happy with whichever solution you think is fit, you're the domain expert.
FWIW, regarding this

> which can be more seamlessly integrated in the library

I think it integrates well with the lib.
For instance the CompositeSpec would do the same thing:
```
CompositeSpec(
  agent_features=CompositeSpec(..., shape=[B, A]),
  common_features=CompositeSpec(..., shape=[B]),
shape=[B]
)
```

> if i have a done or reward, these may be shared or not and thus sometimes i will find them under agent_features and sometimes under common_features

You could have a reward spec
```
reward_spec = CompositeSpec(agent_features=CompositeSpec(reward=SomeSpec(shape=[B, A, R])))
```
or alternatively
```
reward_spec = CompositeSpec(common_features=CompositeSpec(reward=SomeSpec(shape=[B, R])))
```

we'd just need to adapt the code to allow non-standard nested reward keys (a pain be we can talk about it).

torchrl/modules/tensordict_module/exploration.py
```diff
@@ -127,10 +127,7 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
         if exploration_type() == ExplorationType.RANDOM or exploration_type() is None:
             out = tensordict.get(self.td_module.out_keys[0])
             eps = self.eps.item()
-            cond = (torch.rand(tensordict.shape, device=tensordict.device) < eps).to(
-                out.dtype
-            )
-            cond = expand_as_right(cond, out)
+            cond = (torch.rand(out.shape, device=tensordict.device) < eps).to(out.dtype)
```
vmoens, line 130: not sure I understand what this does but I don't think it will work with one-hot encoding no?

matteobettini: i see, you are right.

The problem here was that if the condition had the shape of the tensordict the agents are left out.

We could do something like
```
cond = (torch.rand(out.shape[:-1], device=tensordict.device) < eps).to(out.dtype)
cond = expand_as_right(cond, out)
```
which would work with:
`MultiDiscrete, OneHot, MultiOneHot` 
since they have the hardcoded that the last dimension is the action dimension.

It would not work with `DiscreteTensorSpec`since here the last dimension would be the agents.

Should we uniform `DiscreteTensorSpec` to the other discrete specs ( by making it have the last dim of 1) and go for this solution?

matteobettini: unifroming `DiscreteTensorSpec ` would also get rid of some unsqueezes like here https://github.com/pytorch/rl/blob/main/torchrl/modules/tensordict_module/actors.py#L490

PR #1094, opened 2023-04-26
[Feature] Add inplace for Reward2GoTransform

BY571:
## Description

Adds the inplace option for the Reward2GoTransform. Also fixes transform_output_spec in target_return.

## Motivation and Context

When Reward2GoTransform is combined with TargetReturn transform we want to replace the inference target_returns with the actual received rewards (hindsight reward relabeling). The `inplace` feature allows that. 

Not sure if the test needs to be adapted as the `inplace` would just override existing keys.

## Types of changes

- [X ] New feature (non-breaking change which adds core functionality)

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -1090,15 +1090,15 @@ class TargetReturn(Transform):
             raise ValueError(
                 f"observation_spec was expected to be of type CompositeSpec. Got {type(observation_spec)} instead."
             )
-
-        target_return_spec = BoundedTensorSpec(
-            minimum=-float("inf"),
-            maximum=self.target_return,
-            shape=self.parent.reward_spec.shape,
-            dtype=self.parent.reward_spec.dtype,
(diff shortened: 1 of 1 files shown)


BY571:
Yes, that is exactly what we want. Let me give an example. 
We train an agent for CartPole and we set the target_return to 200. Inputs to our agent would be ["observation", "action", "target_return"]. Now in the rollout, even though the target is 200, the agent only stays alive for 15 steps, meaning for CartPole the reward is 15. We would use the Reward2GoTransform to relabel those target_return values to [15, 14, 13, ...] instead of [200, 199, 198... ](hindsight reward relabeling) so that the context of observations, actions, and actually achieved rewards is correct. As the inputs to the model stay the same we want to directly replace those inference target_returns with the actually received ones for optimization.

vmoens:
Why do we need this to be in-place (sorry I feel like I'm missing something obvious here)

BY571:
With the use of ExcludeTransform I think this PR can be closed :)

PR #1096, opened 2023-04-26
[BugFix] Remove false warnings in losses

vmoens:
cc @matteobettini


matteobettini:
I still get the warning after only doing this
```
 loss_module = DQNLoss(global_qnet, delay_value=True)
    target_net_updater = SoftUpdate(loss_module, eps=1 - config["tau"])
```

vmoens:
> I still get the warning after only doing this
> 
> ```
> loss_module = DQNLoss(global_qnet, delay_value=True)
> target_net_updater = SoftUpdate(loss_module, eps=1 - config["tau"])
> ```

should be solved by now

Issue #1097, opened 2023-04-27
[BUG] KeyError about 'state_value' from PPOLoss


xmaples:
## Describe the bug

A KeyError will be raised from `PPOLoss` when PPO working with `ActorCriticOperator`

This is caught by tensor getting using a hardcoded key name 'state_value', on code line: https://github.com/pytorch/rl/blob/v0.1.0/torchrl/objectives/ppo.py#L127  .
However, `ActorCriticOperator` produces the tensor naming 'state_action_value'.

This may be fixed using a value_key parameter instead of hardcoded key name.

## To Reproduce

Steps:
 1. create `ActorCriticOperator` and PPOLoss (e.g. ClipPPOLoss)
 2. run loss

```python
# reproduce KeyError raised from ppo loss with ActorCriticOperator
""" versions of libraries
torchrl 0.1.0
gym 0.26
torch 2.0
mujoco
"""

from torchrl.envs.libs.gym import GymEnv
from torchrl.envs.transforms import TransformedEnv,DoubleToFloat
from torchrl.modules import ActorCriticOperator, ProbabilisticActor, ValueOperator, Actor
from torchrl.modules import MLP
from torchrl.modules.distributions import TanhNormal
from torchrl.objectives.value import GAE
from torchrl.objectives.ppo import ClipPPOLoss
from torchrl.collectors import SyncDataCollector
from tensordict.nn import TensorDictModule
from tensordict.nn.distributions import NormalParamExtractor
import torch
from torch import nn

env = GymEnv("InvertedDoublePendulum-v4")
print(env.action_spec)
env=TransformedEnv(env, DoubleToFloat(in_keys=["observation"]))

ac_common_module = TensorDictModule(
    module=nn.LazyLinear(out_features=32),
    in_keys=["observation"],
    out_keys=["hidden"]
)
policy_module = ProbabilisticActor(
    module=TensorDictModule(
        module=nn.Sequential(
            nn.LazyLinear(out_features=2*env.action_spec.shape[-1]),
            NormalParamExtractor()
        ),
        in_keys=["hidden"],
        out_keys=["loc","scale"]
    ),
    in_keys=["loc", "scale"],
    distribution_class=TanhNormal,
    distribution_kwargs={
        "min": env.action_spec.space.minimum,
        "max": env.action_spec.space.maximum,
    },
    return_log_prob=True,
)
critic_module = ValueOperator(
    module=MLP(out_features=1, depth=0),
    in_keys=["hidden", "action"],
[...]

PR #1098, opened 2023-04-27
[BugFix] Instruct the value key to PPOLoss

xmaples:
using a `value_key` parameter instead of hardcoded name 'state_value' to get the state value tensor

## Description

using a `value_key` parameter instead of hardcoded name 'state_value' to get the state value tensor

## Motivation and Context

#1097 

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -47,6 +47,8 @@ class PPOLoss(LossModule):
             Defaults to ``"advantage"``.
         value_target_key (str, optional): the input tensordict key where the target state
             value is expected to be written. Defaults to ``"value_target"``.
+        value_key (str, optional): the input tensordict key where the state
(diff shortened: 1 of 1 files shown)


vmoens (approved):
LGMT
Can you sign the CLA?

vmoens:
I believe we should also change the error message because as of now we do t discriminate if o e key is missing or another (the value or the target).
Could you modify the KeyError to have one KeyError for both entries?

vmoens:
Closed by #1124

xmaples:
> I believe we should also change the error message because as of now we do t discriminate if o e key is missing or another (the value or the target). Could you modify the KeyError to have one KeyError for both entries?

missed the message cos of the labor day. I just noticed that you've modified it. nice for you.

PR #1081, opened 2023-04-24
[Feature] Indexing Discrete and OneHot specs

remidomingues:
## Description

Adding indexing support to `DiscreteTensorSpec` and `OneHotDiscreteTensorSpec` specs. 
- [x] Integers
- [x] Lists
- [x] Tuples
- [x] Ranges
- [x] Slices
- [x] None
- [x] Ellipsis
- [x] Numpy arrays
- [x] Tuples

## Motivation and Context

Addressing feature request of adding indexing to specs: https://github.com/pytorch/rl/issues/1051.

Note: although the feature request states that the desired indexing should behave similarly to tensordict, we have found multiple edge cases where the behavior of Numpy and tensordict differ. For all such cases, the behavior of numpy makes the most sense, it may be that these edge cases are tensordict bugs: \[[1](https://github.com/remidomingues/rl/blob/indexing_specs/test/test_specs.py#L2415), [2](https://github.com/remidomingues/rl/blob/indexing_specs/test/test_specs.py#L2422), [3](https://github.com/remidomingues/rl/blob/indexing_specs/test/test_specs.py#L2430), [4](https://github.com/remidomingues/rl/blob/indexing_specs/test/test_specs.py#L2438), [5](https://github.com/remidomingues/rl/blob/indexing_specs/test/test_specs.py#L2449)\].

## Limitations
- The current pull request only adds indexing to `DiscreteTensorSpec` and `OneHotDiscreteTensorSpec`.
We plan to extend the indexing feature in upcoming PRs but thought it might be best to get it approved in smaller chunks.
- For a [very specific edge case](https://github.com/remidomingues/rl/blob/indexing_specs/test/test_specs.py#L2457), containing a combination of multiple slices, lists and integer indices in a very specific order, our current code matches the behavior of Tensordict, but not of numpy. We'd like to fix that but weren't able to understand the rationale behind such indexing.

## Types of changes

- [x] New feature (non-breaking change which adds core functionality)

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.
If you are unsure about any of these, don't hesitate to ask. We are here to help!

- [x] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.

Issue #1051, opened 2023-04-13
[Feature Request] Indexing specs


vmoens:
## Motivation

We can stack specs, but not index them.
Indexing should work as it does for tensordict:
- [ ] integers
- [ ] tensors and lists
- [ ] ranges
- [ ] slices
- [ ] ellipsis
- [ ] None (adds a singleton dimension where needed)

For example:
```python
>>> spec = CompositeSpec(shape=[3, 4])
>>> spec[:].shape
torch.Size([3, 4])
>>> spec[:1].shape
torch.Size([1, 4])
>>> spec[1].shape
torch.Size([4])
>>> spec[..., :2].shape
torch.Size([3, 2])
```

vmoens:
Completed

PR #1105, opened 2023-04-28
[Feature] Indexing specs

remidomingues:
## Description

Add indexing support to remaining specs:
- [x] BinaryDiscreteTensorSpec
- [x] BoundedTensorSpec
- [x] CompositeSpec
- [x] MultiDiscreteTensorSpec
- [x] MultiOneHotDiscreteTensorSpec
- [x] UnboundedContinuousTensorSpec
- [x] UnboundedDiscreteTensorSpec

Already supported by previous PR #1081 :
- [x] DiscreteTensorSpec
- [x] OneHotDiscreteTensorSpec

Note: although indexing & tests have been implemented for BoundedTensorSpec & MultiDiscreteTensorSpec, a NotImplementedError is currently set to prevent a different behavior with the other specs until https://github.com/pytorch/pytorch/issues/100080 is addressed.

## Motivation and Context

Address feature request of adding indexing to specs: https://github.com/pytorch/rl/issues/1051.

## Types of changes

## Checklist

cc @matteobettini

### DIFF
--- a/torchrl/data/tensor_specs.py
+++ b/torchrl/data/tensor_specs.py
@@ -83,6 +83,13 @@ def _default_dtype_and_device(
 
 
 def _validate_idx(shape: list[int], idx: int, axis: int = 0):
+    """Raise an IndexError if idx is out of bounds for shape[axis].
+
+    Args:
+        shape (list[int]): Input shape
+        idx (int): Index, may be negative
+        axis (int): Shape axis to check
+    """
     if idx >= shape[axis] or idx < 0 and -idx > shape[axis]:
         raise IndexError(
             f"index {idx} is out of bounds for axis {axis} with size {shape[axis]}"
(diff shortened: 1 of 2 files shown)


torchrl/data/tensor_specs.py
```diff
@@ -135,7 +135,12 @@ def _slice_indexing(shape: list[int], idx: slice):
     return [n_items] + shape[1:]
 
 
-def _shape_indexing(shape: list[int], idx: SHAPE_INDEX_TYPING):
+def _shape_indexing(
```
vmoens, line 138: this seems pretty crucial and we may refactor that at some point (e.g. using fake tensors)
Can we add a short docstring to say what it is about?

vmoens: we have a similar function in tensordict called `_getitem_batch_size`, is it inspired by that?

vmoens (approved):
Fantastic work, thanks a mil for this!

torchrl/data/tensor_specs.py
remidomingues: Docstring added! `_shape_indexing` is definitely similar to `_getitem_batch_size`. I could use the latter though as it doesn't perform some indexing checks, which are only executed when actually indexing the tensor. I assumed we wanted those checks, hence the reimplementation
```
from tensordict.utils import _getitem_batch_size
from torchrl.data.tensor_specs import _shape_indexing
_getitem_batch_size(torch.Size((3, 2)), 5) # torch.Size([2])
_shape_indexing(torch.Size((3, 2)), 5)  # IndexError: index 5 is out of bounds for axis 0 with size 3
```
If we can work with fake_mode to have both fast shape indexing and such checks, that would definitely be the best of both worlds!

PR #1108, opened 2023-04-28
[WIP] Gradient collector class: an iterator returning model weights

albertbou92:
## Description

Work in progress.

This PR aims to propose a single class to abstract data collection and gradient computation.

This solution might not be very useful in a local training setting but could be use as a base worker for distributed approach like RAPID.

## Motivation and Context

## Types of changes

## Checklist


torchrl/gradient_collector.py
```diff
@@ -0,0 +1,168 @@
...
+        device: torch.device = "cpu",
+    ):
+
+        self.device = device
+        self.updates_per_batch = updates_per_batch
+
+        # Get Actor Critic instance
+        # Could be instantiated passing class + params if necessary
+        self.policy = policy  # TODO: do I really need the policy if we can get the params from objective ?
+        self.policy_params = TensorDict(dict(self.policy.named_parameters()), [])
```
vmoens, line 55: we should add the buffers to this

torchrl/gradient_collector.py
```diff
@@ -0,0 +1,168 @@
...
+        self.critic = critic  # TODO: do I really need the critic if we can get the params from objective ?
+        self.critic_params = TensorDict(dict(self.critic.named_parameters()), [])
```
vmoens, line 57: ditto

torchrl/gradient_collector.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # Get loss instance
+        # Could be instantiated passing class + params if necessary
+        self.objective = objective
+        self.objective_params = TensorDict(dict(self.objective.named_parameters()), [])
```
vmoens, line 62: I expect those params to be the same as policy and critic

torchrl/gradient_collector.py
```diff
@@ -0,0 +1,168 @@
...
+            return None
+
+    def update_policy_weights_(
+            self,
+            policy_weights: Optional[TensorDictBase] = None,
+            critic_weights: Optional[TensorDictBase] = None,
+            objective_weights: Optional[TensorDictBase] = None,
+    ) -> None:
+
+        # TODO: is this the right way to do it ?
```
vmoens, line 108: yeah I think it should work

torchrl/gradient_collector.py
```diff
@@ -0,0 +1,168 @@
...
+        # Update policy and critic
+        self.objective_params.apply(lambda x: x.data).update_(objective_weights)
+        self.policy_params.apply(lambda x: x.data).update_(policy_weights)
+        self.critic_params.apply(lambda x: x.data).update_(critic_weights)
+
+        for name, param in self.objective.named_parameters():
```
vmoens, line 114: in theory your objective params and buffers should match in identity those of the policy and critic
One way to make sure that it's the case would be to register only the objective params and buffers, and during this (in `__init__`) check that they match in identity those of the policy and critic, smth like
```python
if set(policy.parameters()).union(set(critic.parameters())) != objective.parameters():
    raise ValueError
```

torchrl/gradient_collector.py
```diff
@@ -0,0 +1,168 @@
...
+                self.optimizer.zero_grad()
+                loss_sum.backward()
+
+                # Get gradients as a Tensordict
+                objective_params = TensorDict(dict(self.objective.named_parameters()), [])
+                objective_grads = objective_params.apply(lambda x: x.grad)
+                actor_params = TensorDict(dict(self.policy.named_parameters()), [])
+                actor_grads = actor_params.apply(lambda x: x.grad)
+                critic_params = TensorDict(dict(self.critic.named_parameters()), [])
+                critic_grads = critic_params.apply(lambda x: x.grad)
```
vmoens, line 157:
```suggestion
                objective_params = TensorDict(dict(self.objective.named_parameters()), [])
                objective_grads = objective_params.apply(lambda x: x.grad)
```

Only the first two lines should be necessary

Not sure how that would work if the objective_params also contain buffers. Maybe we should do

```
objective_grads = objective_params.apply(lambda x: x.grad if x.grad is not None else torch.zeros_like(x))
```

but it could be unsafe (e.g. if by mistake a param is ignored)

vmoens (commented):
Some early thoughts

PR #1114, opened 2023-05-01
[Nova] Lint job on GHA

osalpekar:
Creating a lint job on GHA via Nova. If this runs well for some time, we can deprecate the old CircleCI job.


osalpekar:
cc @vmoens (I'm unable to add you as a reviewer for some reason, so cc-ing for now)

vmoens (approved):
LGTM

PR #1115, opened 2023-05-02
[Refactor] Nested reward and done specs

vmoens:
## Description

Ensures that the env primitives are compatible with nested reward and done spec.

The reward and done spec are *always* the leaf done and reward. Their location in the output data can be retrieved via the `reward_key` and `done_key` attributed. 
Essentially,
```
assert env.output_spec["done", *env.done_key] is env.done_spec
reset = env.reset()
done = reset[env.done_key]
# for regular envs this will work as expected
done = reset["done"]
```
These changes make it fully bc-compatible (except for the location of the done spec in output spec but we expect most users not to use that feature).

cc @matteobettini have a look at the new test

TODO:
- the nested keys are a bit confusing and instead of 
```python
        CompositeSpec(
            observation: CompositeSpec(
                observation: BoundedTensorSpec(
                    shape=torch.Size([3]),
                    space=ContinuousBox(
                        minimum=Tensor(shape=torch.Size([3]), device=cpu, dtype=torch.float32, contiguous=True),
                        maximum=Tensor(shape=torch.Size([3]), device=cpu, dtype=torch.float32, contiguous=True)),
                    device=cpu,
                    dtype=torch.float32,
                    domain=continuous), device=cpu, shape=torch.Size([])),
            reward: CompositeSpec(
                reward: UnboundedContinuousTensorSpec(
                    shape=torch.Size([1]),
                    space=None,
                    device=cpu,
                    dtype=torch.float32,
                    domain=continuous), device=cpu, shape=torch.Size([])),
            done: CompositeSpec(
                done: DiscreteTensorSpec(
                    shape=torch.Size([1]),
                    space=DiscreteBox(n=2),
                    device=cpu,
                    dtype=torch.bool,
                    domain=discrete), device=cpu, shape=torch.Size([])), device=cpu, shape=torch.Size([]))
```
for the outputspec we could have
```python
        CompositeSpec(
            observation_spec: CompositeSpec(
                observation: BoundedTensorSpec(
                    shape=torch.Size([3]),
                    space=ContinuousBox(
                        minimum=Tensor(shape=torch.Size([3]), device=cpu, dtype=torch.float32, contiguous=True),
                        maximum=Tensor(shape=torch.Size([3]), device=cpu, dtype=torch.float32, contiguous=True)),
                    device=cpu,
                    dtype=torch.float32,
                    domain=continuous), device=cpu, shape=torch.Size([])),
            reward_spec: CompositeSpec(
                reward: UnboundedContinuousTensorSpec(
                    shape=torch.Size([1]),
                    space=None,
                    device=cpu,
                    dtype=torch.float32,
                    domain=continuous), device=cpu, shape=torch.Size([])),
            done_spec: CompositeSpec(
                done: DiscreteTensorSpec(
                    shape=torch.Size([1]),
                    space=DiscreteBox(n=2),
                    device=cpu,
                    dtype=torch.bool,
                    domain=discrete), device=cpu, shape=torch.Size([])), device=cpu, shape=torch.Size([]))
```


matteobettini:
this PoC looks good to me, we can continue the discussion in #1140

vmoens:
@matteobettini All tests are passing, I believe this is ready

matteobettini:
Amazing!

So are the values and the losses stil expecting keys under td["key"] (e.g. action, reward, done)? is there a mechanism in place to them where the keys are?

Or are we doing all that in separate prs?

vmoens:
> Amazing!
> 
> So are the values and the losses stil expecting keys under td["key"] (e.g. action, reward, done)? is there a mechanism in place to them where the keys are?
> 
> Or are we doing all that in separate prs?

On that point of view nothing changes.

Let me take done_spec as example:
```python
>>> env.done_spec = foo
>>> env.output_spec
CompositeSpec(
    "_done_spec": CompositeSpec(
        "done": foo
    )
))
>>> env.done_spec
foo
>>> env.output_spec['_done_spec'][env.done_key]
foo
```
You can also pass a CompositeSpec directly:
```python
>>> env.done_spec = CompositeSpec(done=foo)
>>> # everything here matches the previous code snippet
>>> env.output_spec
CompositeSpec(
    "_done_spec": CompositeSpec(
        "done": foo
    )
))
>>> env.done_spec
foo
>>> env.output_spec['_done_spec'][env.done_key]
foo
```

matteobettini:
No yes what i meant is:

Are losses and modules gonna have access to env.x_key?

Do we implement the infratructure for that in another PR?

vmoens:
Good point.
We will need to let users control that.
Here are a bunch of ideas:
- pass the key names in constructors. This is messy, we'll end up with crazy signatures with tons of args
- sync the loss with an env, smth like `loss.sync_env(env)` which will gather all keys etc. that forces users of the lib to use our envs too, which is not great.
- dedicated methods like `loss.set_keys(reward="my_reward")` and similar. This is compatible with the previous one (ie we can do both)

vmoens:
Oh and yes those things belong to a separate PR

matteobettini:
Yeah out of these the last is the one that makes more sense.
But it think keys that we can set should not only be the ones from the env, but also all the other keys a loss writes.

For example, MADDPG, has a non parametric value key, https://github.com/pytorch/rl/blob/main/torchrl/objectives/ddpg.py#L139

We need to make a sweep and have a systematic way to allow parametric keys

vmoens:
totally agree, we need to brainstorm about that
There needs to be a fine balance between default behaviour and customized one

matteobettini:
I mean in my view each loss should define a list of keys of interest (e.g. reward, done, action, state_value, etc) those can have defaults but users need a quick way to use their own. 

In PPO they are in the constructor https://github.com/pytorch/rl/blob/main/torchrl/objectives/ppo.py#L125 for now

So once we choose hoe to parametrize all loss keys can be in constructor or dedicated method, we unify all losses

PR #1120, opened 2023-05-03
[BugFix] Avoid copying model in losses

vmoens:


vmoens:
I think we still need to copy the model, otherwise we may end up in complicated situations where one model is de-populated from its params and the collector (which uses the same model) ends up with a stateless model.

PR #1121, opened 2023-05-03
[Nova] Remove CircleCI Wheels Builds

osalpekar:
All wheels builds (Linux, MacOS x86, and Windows across all Python versions) already occur on GitHub Actions. The CircleCI  wheels builds are therefore duplicated, and we can remove them safely. This PR removes the build configs from the CircleCI `config.yml` file.

cc @vmoens


vmoens (approved):
LGTM

PR #1122, opened 2023-05-04
[Feature] CatFrames for offline data

vmoens:
## Description

Allows CatFrames to work with offline data.
Upcoming: The tensordict provided *must* have its last dimension marked as time.
As of now the tensordict provided is loosely assumed to have its last dimension being the time dimension.

This is BC-breaking as from now missing in-keys won't be allowed in transforms

cc @BY571


vmoens:
Added padding @BY571

BY571:
> Added padding @BY571

Looks great to me!

PR #1131, opened 2023-05-05
[BugFix] extract the info dict from a list

xmaples:
extract the info dict from the form of obs,info returned from .reset() of a gym env

## Description

extract the info dict from the form of obs,info returned from .reset() of a gym env, so that the function `info_dict_reader` would get an info of type dict {} instead of list of dicts [{}]

## Types of changes


vmoens:
Cool! Do you have an example we could put in the tests under test_libs.py?

xmaples:
here is it:
```python
import gym
import gym_super_mario_bros
from torchrl.envs.libs.gym import GymWrapper
env=gym.make("SuperMarioBros-v0", apply_api_compatibility=True)
env=GymWrapper(env)
def info_reader(info, tensordict):
    print("info: ", info)     # output: info:  [{}] before bugfix
    assert isinstance(info, dict)
   
env.info_dict_reader=info_reader
env.reset()
```

PR #1133, opened 2023-05-05
Linux GPU Brax Unittests

osalpekar:
Moving Linux Brax Unittests to Nova. 

Job is passing: https://github.com/pytorch/rl/actions/runs/5094108150/jobs/9157530612?pr=1133

### DIFF
--- a/.github/workflows/test-linux-brax.yml
+++ b/.github/workflows/test-linux-brax.yml
@@ -28,6 +28,9 @@ jobs:
         export TAR_OPTIONS="--no-same-owner"
         export UPLOAD_CHANNEL="nightly"
 
+        nvidia-smi
+
+
         # docker run -t --gpus all -v $PWD:$PWD -w $PWD -e UPLOAD_CHANNEL -e CU_VERSION -e PYTHON_VERSION pytorch/manylinux-cuda117 .circleci/unittest/linux_libs/scripts_brax/run_all.sh
         bash .circleci/unittest/linux_libs/scripts_brax/run_all.sh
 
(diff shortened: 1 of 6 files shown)


osalpekar:
cc @vmoens - this job is now passing: https://github.com/pytorch/rl/actions/runs/5094108150/jobs/9157530612?pr=1133

.github/workflows/test-linux-brax.yml
```diff
@@ -0,0 +1,30 @@
...
+      - release/*
+  workflow_dispatch:
+
+jobs:
+  unittests:
+    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
+    with:
+      repository: pytorch/rl
+      runner: "linux.g5.4xlarge.nvidia.gpu"
+      docker-image: "pytorch/manylinux-cuda117"
```
vmoens, line 18: looking at the CI report, the GPU is not apparent. Is it normal?

osalpekar:
Test passing at this point but not using GPU. Trying with conda-builder image and GPU pass-through

osalpekar:
@vmoens Getting back to this after some time away :) I see that the CircleCI job on trunk actually also falls back to CPU - see the logs here: https://app.circleci.com/pipelines/github/pytorch/rl/9559/workflows/8a4d7818-94ee-48ec-a6ac-3fa9e9a7c7d8/jobs/299644. I also added a small test in the job to create a torch tensor on GPU and log its device - it indeed creates the tensor on the CUDA device. So the GPUs are apparent from the docker container, it's just that these particular brax tests do not attempt to use them. With verbose logging, I see that this test only attempt to load `libtpu.so` for XLA/TPU support but does not attempt to load `libcuda` at all. I can dig in further later on if needed, but looks like this PR brings the GHA job on par with the trunk CCI job.

vmoens:
ok could be that we need to install jax for cuda? Let me have a look

PR #1134, opened 2023-05-05
Remove old CircleCI Lint

osalpekar:
Looks like the GHA Lint job has been running fine for the past few days, we can probably remove the old CircleCI lint job at this point.


osalpekar:
cc @vmoens

vmoens (approved):
LGTM

PR #1138, opened 2023-05-08
[Nova] Linux Stable Job to GHA

osalpekar:
Moving Linux Stable GPU Unittests to GHA.

Passing job link: https://github.com/pytorch/rl/actions/runs/5814385221/job/15763961829?pr=1138


osalpekar:
@vmoens  Seeing the following errors when using mostly the same scripts/env vars as the previous CCI job:

```
=========================== short test summary info ============================
SKIPPED [1] test/smoke_test_deps.py:25: Not implemented yet
FAILED test/smoke_test_deps.py::test_dm_control - ImportError: Cannot initialize a headless EGL display.
FAILED test/smoke_test_deps.py::test_gym - gymnasium.error.NamespaceNotFound: Namespace ALE not found. Have you installed the proper package for ALE?
``` 

Did some research but don't want to touch anything that would change the test functionality. It seems like `gymnasium` is correctly installed though. Any ideas on resolving these errors?

vmoens:
> Cannot initialize a headless EGL display

I usually rely on the `"nvidia/cudagl:11.4.0-base"` docker image which comes with all the EGL tooling I need. The pytorch docker images we usually use are hard to make work.

> gymnasium.error.NamespaceNotFound: Namespace ALE not found. Have you installed the proper package for ALE?

This should work. Here I read that [it has all been installed](https://github.com/pytorch/rl/actions/runs/4997523323/jobs/8951974740?pr=1138#step:10:2910).
Now it does say just after
 > WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv

which might mean that the libs have been put somewhere else but then gym should not be accessible either...

osalpekar:
Looks like the original Linux Unittests (stable) workflow also used the PyTorch manylinux image, not the cudaGL image. The pip error doesn't seem to be related in this case since other dependencies are found within the conda environment.

vmoens:
If you look at this
https://github.com/pytorch/rl/blob/1e17304cb0732d2f30bef2c15e0e6a0172d9feb8/.circleci/config.yml#L928
we're actually not using the docker to run the tests which makes the GPU apparent (I think)

osalpekar:
I made the generic script run outside of the container. Then we specifically run all the scripts except for `run_tests.sh` inside of the container. Looks like this basically mimics the existing behavior at this point. We're down to just 1 test failing now - looks like gym test is still failing:
```
FAILED test/smoke_test_deps.py::test_gym - gymnasium.error.NamespaceNotFound: Namespace ALE not found. Have you installed the proper package for ALE?
```

.circleci/unittest/linux_stable/scripts/run_test.sh
```diff
@@ -19,6 +19,11 @@ export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$lib_dir
 export MKL_THREADING_LAYER=GNU
 export CKPT_BACKEND=torch
 
+echo "TESTING GYM REGKEYS"
+python -c "import glfw"
```
vmoens, line 23: This seems to fail.
Some context:
If i'm not wrong, GLFW is a backend used by MUJOCO. We usually prefer EGL for headless rendering.
EGL can be either installed manually or we can take some docker image from cuda.
To tell MUJOCO to use EGL instead of GLFW we set MUJOCO_GL=egl and it usually does the job.

osalpekar:
cc @vmoens Got this job to work on GHA - would love your review

.circleci/unittest/linux_stable/scripts/run_all.sh
```diff
@@ -53,16 +55,11 @@ printf "* Installing dependencies (except PyTorch)\n"
...
 
-
-if [[ $OSTYPE == 'darwin'* ]]; then
-  PRIVATE_MUJOCO_GL=glfw
-elif [ "${CU_VERSION:-}" == cpu ]; then
-  PRIVATE_MUJOCO_GL=osmesa
-else
-  PRIVATE_MUJOCO_GL=egl
-fi
+PRIVATE_MUJOCO_GL=osmesa
```
vmoens, line 58: Osmesa is incredibly slow and not sustainable. If the machine has a gpu we need EGL

vmoens (commented):
Have we tried to just copy paste the version that uses the nightlies and install the regular PyTorch instead?

osalpekar:
> Have we tried to just copy paste the version that uses the nightlies and install the regular PyTorch instead?

This is still using the old `install.sh` script, I just consolidated everything into a single `run_all.sh` script. They both use the stable PyTorch version as opposed to nightlies.

.circleci/unittest/linux_stable/scripts/run_all.sh
osalpekar: Can we do this in a follow-up? I haven't got this job work with EGL (with errors like gymnasium regkeys not finding the ALE namespace, Cannot initialize a headless EGL display). Makes sense to use EGL if osmesa is not sustainable, but would like to unblock at least some version of this job in order to complete the CCI deprecation.

Re speed, fwiw this job runs in 30 mins and the old job ran for about 39 minutes on CCI.

PR #1141, opened 2023-05-09
Accelerate GAE

Blonck:
## Description

An optimized vecotrized version for the generalized advantage estimation is used in case gamma and lambda are scalars.

## Motivation and Context

When handling consecutive trajectories of the form

```
```
, vec_generalized_advantage_estimate
needs to build a giant gamma tensor of size [Batch, T, T] with a decayed gamma tensor that suits each trajectory. Thus it needs to allocate a big tensor [B, T, T] and do a heavy matrix multiplication. In case gamma and lambda are scalars, this can be optimized by building a single tensor of the form
```
r_transformed = [
    [r00, r01, r02, r03]
    [r10, r11, 0, 0]
]
```
and applying the gamma filter [r00 + gamma r01 + gamma ** 2 r02 + ..., ro1 + gamma r02 + gamma ** 2 r03 + ...,] to calculate the GAE.

`close #1052 `

## Types of changes

## Checklist

- [x ] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)


Blonck:
Selected the wrong source branch. Close this PR.

PR #1142, opened 2023-05-09
Accelerate GAE

Blonck:
## Description

An optimized vecotrized version for the generalized advantage estimation is used in case gamma and lambda are scalars.

## Motivation and Context

When handling consecutive trajectories of the form

```
```
, `vec_generalized_advantage_estimate`
needs to build a giant gamma tensor of size [Batch, T, T] with a decayed gamma tensor that suits each trajectory. Thus it needs to allocate a big tensor [B, T, T] and do a heavy matrix multiplication. In case gamma and lambda are scalars, this can be optimized by building a single tensor of the form
```
r_transformed = [
    [r00, r01, r02, r03]
    [r10, r11, 0, 0]
]
```
and applying the gamma filter `[r00 + gamma r01 + gamma ** 2 r02 + ..., ro1 + gamma r02 + gamma ** 2 r03 + ...,]` to calculate the GAE.

`close #1052 `

## Types of changes

## Checklist

- [x ] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)


torchrl/objectives/value/functional.py
```diff
@@ -151,19 +214,31 @@ def vec_generalized_advantage_estimate(
     *batch_size, time_steps, lastdim = not_done.shape
 
     value = gamma * lmbda
```
vmoens, line 216: If gamma and lambda are tensors but scalar we should use the fast version

Blonck: I've used
```python
def _to_scalar_if_unique(tensor: Union[float, torch.Tensor]):
    """Return a single scalar if tensor consists of a single unique element, else the source tensor is returned."""
    if isinstance(tensor, torch.Tensor):
        tensor_unique = tensor.unique()
        if tensor_unique.numel() == 1:
            return tensor_unique.item()

    return tensor
```

to catch the case when gamma/lmbda have multiple identical elements, but I'm unsure if it is worth the effort. `.unique()` runtime costs are not high compared to the other code, but if improves the runtime of a case which never occurs in reality, I should probably remove it.

What is your opinion on this?

vmoens: We'll run into issues if gamma carries a gradient. Why not just `if isinstance(gamma, number) or gamma.numel()<=1` (or numpy equivalent)?

vmoens: Oh I see what you did here sorry. I do t thing we should cover this use case. If users provide a tensor full of gammas with the same value they may know what they're doing or be totally off the road, but in either case we should not do an expensive and dangerous operation like unique (same thing as my comment above: if gamma carries a gradient this will break things)

vmoens: My point was merely that your check as it was would treat `gamma = torch.tensor(0.99)` as a fully shaped gamma of the same size of reward and done. If gamma is a tensor but a scalar (it has 1 element or less) it should not be treated as a fully shaped tensor.

Blonck: Done.

benchmarks/test_objectives_benchmarks.py
```diff
@@ -75,6 +75,43 @@ def test_values(benchmark, val_fn, has_lmbda, has_state_value):
...
+    next_values = torch.randn(*size, device=device)
+    reward = torch.randn(*size, device=device)
+    done = torch.zeros(*size, dtype=torch.bool, device=device).bernoulli_(0.1)
+
+    gamma = 0.99
+    if gamma_tensor:
+        gamma = torch.full(size, gamma)
+    lmbda = 0.95
+
+    benchmark(
```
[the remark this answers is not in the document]
Blonck: My guess was to use the automatic calibration is better here, when I have no need of the fine-grained control of benchmark.pedantic. The pytest-benchmark docu says roughly: "don't use pedantic if you don't need it".

torchrl/objectives/value/functional.py
```diff
@@ -114,10 +120,67 @@ def generalized_advantage_estimate(
...
+    All tensors (values, reward and done) must have shape
+    ``[*Batch x TimeSteps x F]``, with ``F`` feature dimensions.
+    """
+    # _gen_num_per_traj and _split_and_pad_sequence need
+    # time dimension at last position
+    if reward.ndim > 2:
+        done = done.transpose(-2, -1)
+        reward = reward.transpose(-2, -1)
+        state_value = state_value.transpose(-2, -1)
+        next_state_value = next_state_value.transpose(-2, -1)
```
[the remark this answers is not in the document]
Blonck: You are right, that does not make sense. In particular, there is the following line in `vec_generalized_advantage_estimate`
```
*batch_size, time_steps, lastdim = not_done.shape
```
which ensures that reward and the other tensors have at least 3 dimensions, so the checks are never executed.

I will remove the checks, although I need to remember why I introduced them in the first place.

vmoens (approved):
LGTM thanks so much for this contribution!

vmoens (approved):
LGTM thanks so much for this contribution!

PR #1146, opened 2023-05-11
[Performance] Vmas vectorized reset

matteobettini:
Performs a vectorized reset in vmas when possible instead of a per env reset


vmoens (approved):
LGTM

PR #1152, opened 2023-05-12
[CI] Concurrency on gha

vmoens:
Cancels workflow jobs for PRs when those of the previous commit aren't finished yet


vmoens:
@osalpekar thoughts on this?

osalpekar:
@vmoens This is a great idea. For reference, here's the concurrency group formatting we use for all the other GHA jobs in the pytorch organization: https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_wheels_linux.yml#L216 (but they have a slightly different naming convention).

PR #1153, opened 2023-05-12
[Performance] Faster GAE

vmoens:


torchrl/objectives/value/functional.py
```diff
@@ -158,7 +159,16 @@ def _fast_vec_gae(
...
     td0_flat = _split_and_pad_sequence(td0, num_per_traj)
 
-    gammalmbdas = torch.ones_like(td0_flat[0])
+    device = done.device
+    if not isinstance(gammalmbda, torch.Tensor):
+        gammalmbda_tensor = torch.tensor(gammalmbda, device=device)
+    else:
+        gammalmbda_tensor = gammalmbda
+    lim = (
+        (torch.tensor(thr, device=device).log() / gammalmbda_tensor.log()).int().item()
```
Blonck, line 168: I guess the `.item()` can be dropped.

In case this code runs on a GPU and gammalmbda is a scalar: would this code not first have to bring single numbers to the GPU do the logarithms and division and bring it back to host memory immediately?
It may introduce a memory barrier, but I've not enough knowledge of the pytorch internals to be sure.

Blonck (approved):
LGTM.

You may consider dropping `.item()` in the limit calculation or calculate the limit within the if/else block and avoiding doing the calculation on the GPU if gamma/lambda are provided not as tensors.

Blonck (approved):
LGTM.

You may consider dropping `.item()` in the limit calculation or calculate the limit within the if/else block and avoiding doing the calculation on the GPU if gamma/lambda are provided not as tensors.

torchrl/objectives/value/functional.py
vmoens: yeah that's right
In practice we would not "bring" anything to gpu but directly create it on gpu. But then we'd bring it on cpu, that's for sure. 
probably negligeable but defo better without it

Blonck (approved):
I would rename `simplefilter` to something like `exponentialDecayFilter`. `simplefilter` could be misleading.

Blonck (approved):
I would rename `simplefilter` to something like `exponentialDecayFilter`. `simplefilter` could be misleading.

torchrl/objectives/value/functional.py
```diff
@@ -38,22 +43,60 @@ def _transpose_time(fun):
...
-            return tuple(transpose_tensor(_out) for _out in out)
-        return fun(*args, time_dim=time_dim, **kwargs)
+                out = transpose_tensor(out)[0]
+                if single_dim:
+                    out = out.squeeze(-2)
+                return out
+            if single_dim:
+                return tuple(transpose_tensor(_out)[0].squeeze(-2) for _out in out)
+            return tuple(transpose_tensor(_out)[0] for _out in out)
+        out = fun(*args, time_dim=time_dim, **kwargs)
```
Blonck, line 91: `out` could be None in case fun has no return values and `out.ndim < 2` will throw an attribute error.

torchrl/objectives/value/functional.py
```diff
@@ -38,22 +43,60 @@ def _transpose_time(fun):
...
 
     @wraps(fun)
     def transposed_fun(*args, time_dim=-2, **kwargs):
         def transpose_tensor(tensor):
-            if isinstance(tensor, (torch.Tensor, MemmapTensor)) and tensor.ndim >= 2:
-                tensor = tensor.transpose(time_dim, -2)
-            return tensor
+            if (
+                not isinstance(tensor, (torch.Tensor, MemmapTensor))
+                or tensor.numel() <= 1
```
Blonck, line 56: Handling `tensor.numel() <= 1` here differently introduces surprising and maybe unwanted behavior. `torch.as_tensr([1])` will not be transformed and has shape [1] (one-dimensional), whereas torch.as_tensor([1, 2])` will be transformed to shape [2, 1] (two-dimensional).

Blonck (changes requested):
Overall LGTM.

@_transpose_time has one bug in case the function has no return value. You may also consider the other remarks.

torchrl/objectives/value/functional.py
vmoens: Which value function would not return anything?
This is tailored for such functions. I suspect that if we want to apply it to anything else we'll run in other issues.

torchrl/objectives/value/functional.py
vmoens: Sequences of a single element are defo an edge case. 
A way out could be to assert that sequences have more than one element in each method, when we check the shapes. 
But giving a `gamma=torch.tensor([1.0])` must be supported

torchrl/objectives/value/functional.py
Blonck: For example, a method of a class which calculates some values and stores them into self. On the other hand, its a "private" method and I guess the risk is rather low.

vmoens: yeah it is private, its scope is well defined (within that file, nowhere else) etc
I don't think we need to make it work with other signatures

torchrl/objectives/value/functional.py
```diff
@@ -152,19 +196,31 @@ def _fast_vec_gae(
...
     td0 = reward + not_done * gamma * next_state_value - state_value
 
     num_per_traj = _get_num_per_traj(done)
     td0_flat = _split_and_pad_sequence(td0, num_per_traj)
 
-    gammalmbdas = torch.ones_like(td0_flat[0])
-    gammalmbdas[1:] = gammalmbda
-    gammalmbdas[1:] = gammalmbdas[1:].cumprod(0)
-    gammalmbdas = gammalmbdas.unsqueeze(-1)
+    if not CPP:
```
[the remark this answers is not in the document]
vmoens: Agreed
TBH it was for debugging and quickly hack through one version or the other.
I think I will move the c++ codes to a dormant PR

vmoens:
Here are the results of the profiling on an A100 cluster
```
Timer unit: 1e-06 s

Total time: 3.98617 s
File: /fsx/users/vmoens/work/rl/torchrl/objectives/value/functional.py
Function: _fast_vec_gae at line 162

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   162                                           @profile
   163                                           def _fast_vec_gae(
   164                                               reward: torch.Tensor,
   165                                               state_value: torch.Tensor,
   166                                               next_state_value: torch.Tensor,
   167                                               done: torch.Tensor,
   168                                               gamma: float,
   169                                               lmbda: float,
   170                                               thr: float = 1e-7,
   171                                           ):
   172                                               """Fast vectorized Generalized Advantage Estimate when gamma and lmbda are scalars.
   173                                           
   174                                               In contrast to `vec_generalized_advantage_estimate` this function does not need
   175                                               to allocate a big tensor of the form [B, T, T].
   176                                           
   177                                               Args:
   178                                                   reward (torch.Tensor): a [*B, T, F] tensor containing rewards
   179                                                   state_value (torch.Tensor): a [*B, T, F] tensor containing state values (value function)
   180                                                   next_state_value (torch.Tensor): a [*B, T, F] tensor containing next state values (value function)
   181                                                   done (torch.Tensor): a [B, T] boolean tensor containing the done states
   182                                                   gamma (scalar): the gamma decay (trajectory discount)
   183                                                   lmbda (scalar): the lambda decay (exponential mean discount)
   184                                                   thr (float): threshold for the filter. Below this limit, components will ignored.
   185                                                       Defaults to 1e-7.
   186                                           
   187                                               All tensors (values, reward and done) must have shape
   188                                               ``[*Batch x TimeSteps x F]``, with ``F`` feature dimensions.
   189                                           
   190                                               """
   191                                               # _gen_num_per_traj and _split_and_pad_sequence need
   192                                               # time dimension at last position
   193      1000       3439.7      3.4      0.1      done = done.transpose(-2, -1)
   194      1000       1571.7      1.6      0.0      reward = reward.transpose(-2, -1)
   195      1000       1416.2      1.4      0.0      state_value = state_value.transpose(-2, -1)
   196      1000       1497.0      1.5      0.0      next_state_value = next_state_value.transpose(-2, -1)
   197                                           
   198      1000        362.2      0.4      0.0      gammalmbda = gamma * lmbda
   199      1000      19214.0     19.2      0.5      not_done = (~done).int()
   200      1000      52267.5     52.3      1.3      td0 = reward + not_done * gamma * next_state_value - state_value
   201                                           
   202      1000     154644.6    154.6      3.9      num_per_traj = _get_num_per_traj(done)
   203      1000     254032.6    254.0      6.4      td0_flat, mask = _split_and_pad_sequence(td0, num_per_traj, return_mask=True)
   204                                           
   205      1000        750.2      0.8      0.0      device = done.device
   206      1000       1035.2      1.0      0.0      if not isinstance(gammalmbda, torch.Tensor):
   207      1000     275789.2    275.8      6.9          gammalmbda_tensor = torch.tensor(gammalmbda, device=device)
   208                                               else:
   209                                                   gammalmbda_tensor = gammalmbda
   210      1000      36622.6     36.6      0.9      lim = int(math.log(thr) / gammalmbda_tensor.log().item())
   211      1000      23164.5     23.2      0.6      gammalmbdas = torch.ones_like(td0_flat[0][:lim])
   212                                           
   213      1000      13821.4     13.8      0.3      gammalmbdas[1:] = gammalmbda
   214      1000      34189.4     34.2      0.9      gammalmbdas[1:] = gammalmbdas[1:].cumprod(0)
   215      1000       3667.3      3.7      0.1      gammalmbdas = gammalmbdas.unsqueeze(-1)
   216                                           
   217      1000    1417091.2   1417.1     35.6      advantage = _custom_conv1d(td0_flat.unsqueeze(1), gammalmbdas)
   218      1000       2815.6      2.8      0.1      advantage = advantage.squeeze(1)
   219                                               # torchaudio version: slightly slower than plain conv1d:
   220                                               #     dtype = td0_flat.dtype
   221                                               #     advantage = torchaudio.functional.lfilter(
   222                                               #         td0_flat.flip(-1),
   223                                               #         torch.tensor([1.0, -gammalmbda], dtype=dtype, device=device),
   224                                               #         torch.tensor([1.0, 0.0], device=device, dtype=dtype),
   225                                               #         clamp=False,
   226                                               #     ).flip(-1)
   227      1000    1668560.8   1668.6     41.9      advantage = _inv_pad_sequence(advantage, num_per_traj, mask=mask).view_as(reward)
   228                                           
   229      1000      13402.4     13.4      0.3      value_target = advantage + state_value
   230                                           
   231      1000       4692.6      4.7      0.1      advantage = advantage.transpose(-1, -2)
   232      1000       1849.9      1.8      0.0      value_target = value_target.transpose(-1, -2)
   233                                           
   234      1000        276.3      0.3      0.0      return advantage, value_target

Total time: 1.653 s
File: /fsx/users/vmoens/work/rl/torchrl/objectives/value/utils.py
Function: _inv_pad_sequence at line 307

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   307                                           @profile
   308                                           def _inv_pad_sequence(
   309                                               tensor: Union[torch.Tensor, TensorDictBase],
   310                                               splits: torch.Tensor,
   311                                               mask: torch.Tensor = None,
   312                                           ):
   313                                               """Inverse a pad_sequence operation.
   314                                           
   315                                               If tensor is of shape [B, T], than splits must be of of shape [B] with all elements
   316                                               and integer between [1, T].
   317                                               The result will be flattened along the batch dimension(s) and must be reshaped into
   318                                               the original shape (if necessary).
   319                                           
   320                                               Examples:
   321                                                   >>> rewards = torch.randn(100, 20)
   322                                                   >>> num_per_traj = _get_num_per_traj(torch.zeros(100, 20).bernoulli_(0.1))
   323                                                   >>> padded = _split_and_pad_sequence(rewards, num_per_traj)
   324                                                   >>> reconstructed = _inv_pad_sequence(padded, num_per_traj)
   325                                                   >>> assert (reconstructed==rewards).all()
   326                                               """
   327      1000       1202.6      1.2      0.1      if splits.numel() == 1:
   328                                                   return tensor
   329                                           
   330      1000        252.2      0.3      0.0      if mask is None:
   331                                                   # int16 supports length up to 32767
   332                                                   dtype = (
   333                                                       torch.int16
   334                                                       if tensor.shape[-1] < torch.iinfo(torch.int16).max
   335                                                       else torch.int32
   336                                                   )
   337                                                   arange = torch.arange(
   338                                                       tensor.shape[-1], device=tensor.device, dtype=dtype
   339                                                   ).unsqueeze(0)
   340                                                   mask = arange < splits.unsqueeze(1)
   341      1000    1651545.3   1651.5     99.9      return tensor[mask]

```
We spend half of the time indexing the tensor with the mask!

Blonck (approved):
LGTM.

PR #1158, opened 2023-05-15
[Performance] Accelerate TD lambda return estimate 

### LINKED ISSUE #1148: [Performance] Accelerate TD lambda return estimate
# Description

The idea is to improve the performance of `vec_td_lambda_return_estimate` in the lambda/gamma are scalars (or tensors with a single unique value).
The vectorized version of TD lambda works by building a cumprod of the gamma decay
```
gamma_cumprod = [1, gamma, gamma ** 2, ...]
```
and applying conv1d to it which results in
```
[r00 + gamma r01 + gamma ** 2 r02 + ....,
 r01 + gamma r02 + gamma ** 2 r03 + ....,
 r0n
]
```

In principal, the same idea used in #1142 should be applicable.
Given consecutive trajectories in the form of:
```
```
the traces are split into:
```
r_transformed = [
    [r00, r01, r02, r03],
    [r10, r11, 0, 0]
]
```
Apply the filter to `r_transformed` to calculate the TD lambda return.

Finally, `vec_td_lambda_return_estimate` should use this case if applicable.
```
if (isinstance(gamma, torch.Tensor) and gamma.numel() > 1):
   # use existing code
elif done.any():
   # use new version
else:
   # use existing code
```

Blonck:
## Description

Implement a version of for `vec_td_lambda_return_estimate` optimized for the case where gamma is a scalar.

## Motivation and Context

In the case gamma is a scalar one can do the same optimization as done in #1141. Instead of constructing a big gamma tensor [B, T, T], one can split and pad the consecutive traces:
```python
```
into
```python
r_transformed = [
    [r00, r01, r02, r03],
    [r10, r11, 0, 0]
]
```
and then apply a gamma filter to this transformed input.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

Closes #1148 

## Checklist

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -4307,34 +4307,49 @@ class TestValues:
         torch.testing.assert_close(v1, v2, rtol=1e-4, atol=1e-4)
 
     @pytest.mark.parametrize("device", get_available_devices())
-    @pytest.mark.parametrize("gamma", [0.5, 0.99, 0.1])
-    @pytest.mark.parametrize("lmbda", [0.1, 0.5, 0.99])
+    @pytest.mark.parametrize("gamma", [0.5, 0.99])
+    @pytest.mark.parametrize("lmbda", [0.25, 0.99])
     @pytest.mark.parametrize("N", [(3,), (7, 3)])
-    @pytest.mark.parametrize("T", [3, 5, 200])
+    @pytest.mark.parametrize("T", [3, 100])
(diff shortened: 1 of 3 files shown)


Blonck:
Performance comparison for shape (16, T, 1) on GPU. Shown is the the ratio runtime(`algorithm`)/runtime(`td_lambda_return_estimate`).

newplot (1)

vmoens (approved):
LGTM thanks for this!

PR #1159, opened 2023-05-16
[Nova] Remove windows Unittests from CCI

osalpekar:
We moved this unittest job to GHA in a previous PR. This removes those jobs from CCI entirely.


osalpekar:
cc @vmoens

vmoens (approved):
LGTM

PR #1161, opened 2023-05-16
Fix bug where ToTensorImage unsqueeze would not update the observation spec

### LINKED ISSUE #1160: [BUG] `ToTensorImage` doesn't update the observation spec properly if `unsqueeze=True`
## Describe the bug

If you use `ToTensorImage` with the unsqueeze parameter set, the observations are unsqueezed correctly. However, the observation spec isn't updated.

## To Reproduce

You can just use any `TransformedEnv` that has a `ToTensorImage` transformation with `unsqueeze=True`. Then, try to create a fake tensordict with `fake_tensordict()`. The dimensions of the observation will be wrong as they don't include the unsqueezed dimension.

```python
import torchrl

base_env = ...
env = TransformedEnv(
    base_env,
    Compose(
        ToTensorImage(in_keys="observation", unsqueeze=True),
        StepCounter(),
    ),
)
print(env.fake_tensordict())
```

## Expected behavior

The fake tensordict should have an extra dimension to account for the unsqueezed dimension.

## System info
`torchrl` was installed through Python Poetry and is running on Python 3.10.8.

```python
import torchrl, numpy, sys
None 1.24.3 3.10.8 (tags/v3.10.8:aaaf517, Oct 11 2022, 16:50:30) [MSC v.1933 64 bit (AMD64)] win32
```

## Additional context

N/A

## Reason and Possible fixes

I will submit a PR that outlines a possible fix.

## Checklist

hyerra:
## Description

If you use `ToTensorImage` with the unsqueeze parameter set, the observations are unsqueezed correctly. However, the observation spec isn't updated. This PR fixes that issue.

## Motivation and Context

This is necessary for methods like `fake_tensordict()` where the generated tensordict depends on the observation spec. This change makes sure the observation dict is updated accordingly so methods like `fake_tensordict()` work properly.

close #1160

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -951,15 +951,17 @@ class ToTensorImage(ObservationTransform):
             *list(range(observation.ndimension() - 3)), -1, -3, -2
         )
         observation = observation.div(255).to(self.dtype)
-        if observation.ndimension() == 3 and self.unsqueeze:
+        if self._should_unsqueeze(observation):
             observation = observation.unsqueeze(0)
         return observation
 
     @_apply_to_composite
     def transform_observation_spec(self, observation_spec: TensorSpec) -> TensorSpec:
         observation_spec = self._pixel_observation(observation_spec)
+        unsqueeze_dim = [1] if self._should_unsqueeze(observation_spec) else []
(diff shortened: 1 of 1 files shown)


vmoens (approved):
LGTM

PR #1164, opened 2023-05-18
[Feature] KL to init setting in PPO

vmoens:
@tcbegley @apbard

### DIFF
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -370,7 +370,7 @@ class ClipPPOLoss(PPOLoss):
             parameters for the actor. These will be used to return a ``"loss_kl_init"``
             differentiable entry in the output loss TensorDict that can be used
             to constrain the policy to stay close to its original configuration.
-            Defaults to 0 (ie no loss).
+            Defaults to 0 (i.e., no KL component in the loss).
             .. note::
               if this KL has to be registered but not used for training, the
               resulting ``"loss_kl_init"`` can be detached after loss
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -572,8 +572,8 @@ class KLPENPPOLoss(PPOLoss):
             will be available and contain the initial configuration of the
             parameters for the actor. These will be used to return a ``"loss_kl_init"``
             differentiable entry in the output loss TensorDict that can be used
-            to constain the policy to stay close to its original configuration.
-            Defaults to 0 (ie no loss).
+            to constrain the policy to stay close to its original configuration.
+            Defaults to 0 (i.e., no loss).
             .. note::
               if this KL has to be registered but not used for training, the
               resulting ``"loss_kl_init"`` can be detached after loss
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -80,7 +80,7 @@ class PPOLoss(LossModule):
             parameters for the actor. These will be used to return a ``"loss_kl_init"``
             differentiable entry in the output loss TensorDict that can be used
             to constrain the policy to stay close to its original configuration.
-            Defaults to 0 (ie no loss).
+            Defaults to 0 (i.e., no KL component in the loss).
             .. note::
               if this KL has to be registered but not used for training, the
               resulting ``"loss_kl_init"`` can be detached after loss
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -369,7 +369,7 @@ class ClipPPOLoss(PPOLoss):
             will be available and contain the initial configuration of the
             parameters for the actor. These will be used to return a ``"loss_kl_init"``
             differentiable entry in the output loss TensorDict that can be used
-            to constain the policy to stay close to its original configuration.
+            to constrain the policy to stay close to its original configuration.
             Defaults to 0 (ie no loss).
             .. note::
               if this KL has to be registered but not used for training, the
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -73,9 +73,9 @@ class PPOLoss(LossModule):
             policy and critic will only be trained on the policy loss.
             Defaults to ``False``, ie. gradients are propagated to shared
             parameters for both policy and critic losses.
-        kl_init_coef (scalar, optional): A coefficient for the KL-divergence
+        kl_coef (scalar, optional): A coefficient for the KL-divergence
             between the current actor configuration and the initial one.
-            If ``kl_init_coef > 0``, a ``initial_actor_params`` attribute
+            If ``kl_coef > 0``, a ``initial_actor_params`` attribute
             will be available and contain the initial configuration of the
             parameters for the actor. These will be used to return a ``"loss_kl_init"``
             differentiable entry in the output loss TensorDict that can be used
@@ -148,7 +148,7 @@ class PPOLoss(LossModule):
         normalize_advantage: bool = False,
         gamma: float = None,
         separate_losses: bool = False,
-        kl_init_coef: float = 0.0,
+        kl_coef: float = 0.0,
     ):
         super().__init__()
         self.convert_to_functional(
@@ -160,12 +160,12 @@ class PPOLoss(LossModule):
             policy_params = list(actor.parameters())
         else:
             policy_params = None
-        self.kl_init_coef = kl_init_coef
+        self.kl_coef = kl_coef
         self.convert_to_functional(
             critic,
             "critic",
             compare_against=policy_params,
-            create_target_params=kl_init_coef > 0,
+            create_target_params=kl_coef > 0,
         )
         self.advantage_key = advantage_key
         self.value_target_key = value_target_key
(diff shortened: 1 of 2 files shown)


torchrl/objectives/ppo.py
```diff
@@ -71,6 +73,19 @@ class PPOLoss(LossModule):
             policy and critic will only be trained on the policy loss.
             Defaults to ``False``, ie. gradients are propagated to shared
             parameters for both policy and critic losses.
+        kl_init_coef (scalar, optional): A coefficient for the KL-divergence
+            between the current actor configuration and the initial one.
+            If ``kl_init_coef > 0``, a ``init_actor_params`` attribute
+            will be available and contain the initial configuration of the
+            parameters for the actor. These will be used to return a ``"loss_kl_init"``
+            differentiable entry in the output loss TensorDict that can be used
+            to constain the policy to stay close to its original configuration.
```
apbard, line 82:
```suggestion
            to constrain the policy to stay close to its original configuration.
```

torchrl/objectives/ppo.py
```diff
@@ -71,6 +73,19 @@ class PPOLoss(LossModule):
...
+            Defaults to 0 (ie no loss).
```
apbard, line 83:
```suggestion
            Defaults to 0 (i.e., no loss).
```

torchrl/objectives/ppo.py
```diff
@@ -323,6 +363,18 @@ class ClipPPOLoss(PPOLoss):
...
```
apbard, line 373:
```suggestion
            Defaults to 0 (i.e., no KL component in the loss).
```

torchrl/objectives/ppo.py
```diff
@@ -323,6 +363,18 @@ class ClipPPOLoss(PPOLoss):
             policy and critic will only be trained on the policy loss.
             Defaults to ``False``, ie. gradients are propagated to shared
             parameters for both policy and critic losses.
+        kl_init_coef (scalar, optional): A coefficient for the KL-divergence
+            between the current actor configuration and the initial one.
+            If ``kl_init_coef > 0``, a ``init_actor_params`` attribute
+            will be available and contain the initial configuration of the
+            parameters for the actor. These will be used to return a ``"loss_kl_init"``
+            differentiable entry in the output loss TensorDict that can be used
+            to constain the policy to stay close to its original configuration.
```
apbard, line 372:
```suggestion
            to constrain the policy to stay close to its original configuration.
```

torchrl/objectives/ppo.py
```diff
@@ -71,6 +73,19 @@ class PPOLoss(LossModule):
             policy and critic will only be trained on the policy loss.
             Defaults to ``False``, ie. gradients are propagated to shared
             parameters for both policy and critic losses.
+        kl_init_coef (scalar, optional): A coefficient for the KL-divergence
+            between the current actor configuration and the initial one.
+            If ``kl_init_coef > 0``, a ``init_actor_params`` attribute
+            will be available and contain the initial configuration of the
+            parameters for the actor. These will be used to return a ``"loss_kl_init"``
```
apbard, line 80: > ``init_actor_params`` attribute
            will be available and contain the initial configuration of the
            parameters for the actor
            
not super clear. what do you mean?

torchrl/objectives/ppo.py
```diff
@@ -163,6 +185,14 @@ def __init__(
             warnings.warn(_GAMMA_LMBDA_DEPREC_WARNING, category=DeprecationWarning)
             self.gamma = gamma
 
+    @property
+    def init_actor_params(self):
```
apbard, line 189: maybe `actor_init_params` is better, otherwise seems an intialisation method

torchrl/objectives/ppo.py
vmoens: The property won't be available otherwise

torchrl/objectives/ppo.py
```diff
@@ -71,6 +73,19 @@ class PPOLoss(LossModule):
             policy and critic will only be trained on the policy loss.
             Defaults to ``False``, ie. gradients are propagated to shared
             parameters for both policy and critic losses.
+        kl_init_coef (scalar, optional): A coefficient for the KL-divergence
```
apbard, line 76: `kl_init_coef` let think that this is an initial value for kl. what about simply `kl_coef`?

vmoens: We have other KL coefs we need to distinguish them no?

torchrl/objectives/ppo.py
vmoens: Right! I usually don't define docstrings for properties bc I don't expect users to know that smth is a property and not a regular attribute (hence they will never call `help(obj.smth)`)
Can we write it in both places?

vmoens:
I think I addressed all your comments @apbard

torchrl/objectives/ppo.py
apbard: we need a name that says its the kl vs the initial/reference model. `initial_model_kl_coeff`? its long but I can't think to anything better right now

also what about passing a "config" (aka a dict/dataclass) instead? this enables to implement advanced kl controllers like [here](https://github.com/lvwerra/trl/blob/25fa1bd8803c9998d95a3c38b2bb27a85235ce7b/trl/trainer/ppo_trainer.py#L269)

torchrl/objectives/ppo.py
apbard: I think here it is not very useful. If you want to keep it I would just write something short as `If kl_init_coef > 0, the init_actor_params will raise`

another option is to make the attribute private.

torchrl/objectives/ppo.py
vmoens: I changed it to kl_coef, I think it's fine. the other one is named beta and it is used differently. There's no risk of clashing IMO

torchrl/objectives/ppo.py
```diff
@@ -652,7 +722,34 @@ def forward(self, tensordict: TensorDictBase) -> TensorDict:
             loss_critic = self.loss_critic(tensordict)
             td_out.set("loss_critic", loss_critic.mean())
 
+        if self.kl_coef:
+            loss_kl_init = self.kl_coef * self._kl_init(dist, tensordict)
+            td_out.set("loss_kl_init", loss_kl_init.mean())
```
apbard, line 727: I will probably change also this to `kl_loss` or `loss_kl`

apbard (commented):
LGTM

torchrl/objectives/ppo.py
vmoens: kl_loss no because we need all losses to start by loss by convention
For the KLPENPPOLoss we already have a non-differentiable "kl" entry so it's confusing if we have loss_kl and kl together, one would naturally expect them to be related when they're not.
Maybe `"loss_actor_shift"`?

apbard: "loss_kl_actor_shift" is nice but my concern is that we have 2 losses that are based on KL. we need to make sure it's clear its the kl computed against:
- ref/initial/starting model
- previous step/iteration model

so would change "kl" entry also

torchrl/objectives/ppo.py
```diff
@@ -454,6 +511,9 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
         if self.critic_coef:
             loss_critic = self.loss_critic(tensordict)
             td_out.set("loss_critic", loss_critic.mean())
+        if self.kl_coef:
+            loss_kl_init = self.kl_coef * self._kl_init(dist, tensordict)
+            td_out.set("loss_kl_init", loss_kl_init.mean())
         td_out.set("ESS", ess.mean() / batch)
         return td_out
```
apbard, line 519: detail what is new/old policy in the following docstring

torchrl/objectives/ppo.py
vmoens: kl_to_init
kl_to_prev

apbard: kl_wrt_init
kl_wrt_prev
?
otherwise good your version with \_to\_

torchrl/objectives/ppo.py
apbard: KLPENPPOLoss one. (you haven't edit that so couldn't comment on those lines)

torchrl/objectives/ppo.py
```diff
@@ -708,7 +708,7 @@ def forward(self, tensordict: TensorDictBase) -> TensorDict:
         td_out = TensorDict(
             {
                 "loss_objective": -neg_loss.mean(),
-                "kl": kl_val.detach().mean(),
+                "kl_to_prev": kl_val.detach().mean(),
```
apbard, line 711: why is not starting with "loss_*"?

vmoens: there's a detach just after

vmoens: true, we could separate that loss too
in this case i did not do it for consistency with clip
such that one can safely do 
```
loss = losses["loss_actor"] + loss["loss_critic"]
```
otherwise the KL module needs one more term.

vmoens: If we want to change that behaviour it has to be in a separate PR bc it is bc breaking

vmoens: also for context non-loss values are metadata
we should probably ppt them in a separate nested td

torchrl/objectives/ppo.py
apbard: also is KL(old_policy, new_policy) correct? here they use the opposite:  https://github.com/lvwerra/trl/blob/main/trl/trainer/ppo_trainer.py#LL960C28-L960C39

vmoens: image
From the paper

gtnv:
hey @vmoens, is this still being implemented? if not, would you like to close it?

PR #1166, opened 2023-05-18
[Refactor] Parallelenv uses queues

vmoens:
cc @matteobettini @Blonck


vmoens:
I took the opportunity to investigate why parallel env is so slow and it is due to casting on device.
So I did the casting all at once.
Together with https://github.com/pytorch-labs/tensordict/pull/389 the speed up is about 4x, almost comparable with the CPU version.
With native CUDA envs it should be better.

PR #1175, opened 2023-05-22
[Refactor] the usage of tensordict keys in loss modules

### LINKED ISSUE #1174: [Feature Request] Refactor key usage of loss modules
We have various loss modules in RL. 
They work as

```
loss_module = LossModule(network, …)
loss_module(data)
```

`data` is a TensorDict instance. The loss will pick up a bunch of keys from it and read them.
Some are defined by network (network will simply do `network(data)`)
For some other operations, it is `LossModule` that will read the keys from the tensordict. For instance:

https://github.com/pytorch/rl/blob/714d6459e7b58f3b5601c26d612211527582327a/torchrl/objectives/ddpg.py#L139
https://github.com/pytorch/rl/blob/d6a466da6b403a6ec87bbb633d0249cbd824475e/torchrl/objectives/ppo.py#L188

What we would like to do is to have a way to tell the loss module where to find these keys, something like

```
loss_module.set_keys(sample_log_prob=“some_other_key”)
```

This set_keys would take a limited number of arguments for each loss module and write a private attribute with the key that points to the value we want.

Like this the loss would be 100% oblivious to choices from the user in terms of key naming, but still have default values for an easier integration.

It will also remove the key names from the `__init__` method which pollutes them.

# Action items

1. For each loss, implement set_keys and document which keys can be written on a case-by-case basis.
2. In each constructor, raise a deprecation warning if users pass a key 
e.g. here 
https://github.com/pytorch/rl/blob/d6a466da6b403a6ec87bbb633d0249cbd824475e/torchrl/objectives/ppo.py#L125-L127
3. write tests for each loss to check that this works as expected
4. We should do the same for the value modules (here https://github.com/pytorch/rl/blob/d6a466da6b403a6ec87bbb633d0249cbd824475e/torchrl/objectives/value/advantages.py)

# Context

To give you some context, the way we will be using that is when users (for instance in MultiAgent RL) are using nested keys:
```
reward = data[‘next’, ‘agents’, ‘reward’]
```
which is unconventional but should be supported.
So we want to be able to tell the loss: look, here the reward is not in `(‘next’, ‘reward’)`

cc @Blonck @matteobettini

Blonck:
## Description

We have various loss modules in RL. 
They work as

```
loss_module = LossModule(network, …)
loss_module(data)
```

These loss modules access the actual data by keys. Some keys are configurable via ctor,

https://github.com/pytorch/rl/blob/3c8197b41e351de2ebefbe2df5d5900d82e43882/torchrl/objectives/ppo.py#L125

others are hardcoded,

https://github.com/pytorch/rl/blob/714d6459e7b58f3b5601c26d612211527582327a/torchrl/objectives/ddpg.py#L139

This is refactored such that all relevant keys can be set via

```
loss_module.set_keys(sample_log_prob=“some_other_key”)
```

The same is done for the advantage modules.
```
advantage_module.set_keys(advantage="other_advantage")
```

Closes #1174 

## Types of changes

## Checklist

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -3954,14 +3954,14 @@ class TestDreamer(LossModuleTestBase):
         loss_fn = DreamerModelLoss(world_model)
 
         default_keys = {
-            "reward_key": "reward",
-            "true_reward_key": "true_reward",
-            "prior_mean_key": "prior_mean",
-            "prior_std_key": "prior_std",
-            "posterior_mean_key": "posterior_mean",
-            "posterior_std_key": "posterior_std",
-            "pixels_key": "pixels",
-            "reco_pixels_key": "reco_pixels",
+            "reward": "reward",
+            "true_reward": "true_reward",
+            "prior_mean": "prior_mean",
+            "prior_std": "prior_std",
+            "posterior_mean": "posterior_mean",
+            "posterior_std": "posterior_std",
+            "pixels": "pixels",
+            "reco_pixels": "reco_pixels",
         }
         self.tensordict_keys_test(loss_fn, default_keys=default_keys)
 
@@ -3979,12 +3979,12 @@ class TestDreamer(LossModuleTestBase):
         )
 
         default_keys = {
-            "belief_key": "belief",
-            "reward_key": "reward",
-            "value_key": "state_value",
-            "done_key": "done",
+            "belief": "belief",
+            "reward": "reward",
+            "value": "state_value",
+            "done": "done",
         }
-        key_mapping = {"value_key": "value_key"}
+        key_mapping = {"value": "value_key"}
         self.tensordict_keys_test(
             loss_fn,
             default_keys=default_keys,
@@ -3997,7 +3997,7 @@ class TestDreamer(LossModuleTestBase):
         loss_fn = DreamerValueLoss(value_model)
 
         default_keys = {
-            "value_key": "state_value",
+            "value": "state_value",
         }
         self.tensordict_keys_test(loss_fn, default_keys=default_keys)
 
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -141,16 +141,10 @@ def get_devices():
 
 
 class LossModuleTestBase:
-    def tensordict_keys_test(
-        self, loss_fn, default_keys, td_est=None, loss_advantage_key_mapping=None
-    ):
+    def tensordict_keys_test(self, loss_fn, default_keys, td_est=None):
         self.tensordict_keys_unknown_key_test(loss_fn)
         self.tensordict_keys_default_values_test(loss_fn, default_keys)
         self.tensordict_set_keys_test(loss_fn, default_keys)
-        if loss_advantage_key_mapping is not None:
-            self.set_advantage_keys_through_loss_test(
-                loss_fn, td_est, loss_advantage_key_mapping
-            )
 
     def tensordict_keys_unknown_key_test(self, loss_fn):
         """Test that exception is raised if an unknown key is set via .set_keys()"""
@@ -185,12 +179,17 @@ class LossModuleTestBase:
     ):
         key_mapping = loss_advantage_key_mapping
         test_fn = deepcopy(loss_fn)
+
+        new_keys = {}
+        for loss_key, (_, new_key) in key_mapping.items():
+            new_keys[loss_key] = new_key
+
+        test_fn.set_keys(**new_keys)
         test_fn.make_value_estimator(td_est)
 
-        for loss_key, advantage_key in key_mapping.items():
-            test_fn.set_keys(**{loss_key: "test1"})
+        for _, (advantage_key, new_key) in key_mapping.items():
             assert (
-                getattr(test_fn.value_estimator.tensor_keys, advantage_key) == "test1"
+                getattr(test_fn.value_estimator.tensor_keys, advantage_key) == new_key
             )
 
 
@@ -875,15 +874,21 @@ class TestDDPG(LossModuleTestBase):
             "state_action_value": "state_action_value",
             "priority": "td_error",
         }
-        # TODO value needs state_action_value as out key for this test
-        # key_mapping = {"state_action_value": "value"}
 
         self.tensordict_keys_test(
             loss_fn,
             default_keys=default_keys,
             td_est=td_est,
-            # loss_advantage_key_mapping=key_mapping,
         )
+
+        value = self._create_mock_value(out_keys=["state_action_value_test"])
+        loss_fn = DDPGLoss(
+            actor,
+            value,
+            loss_function="l2",
+        )
+        key_mapping = {"state_action_value": ("value", "state_action_value_test")}
+        self.set_advantage_keys_through_loss_test(loss_fn, td_est, key_mapping)
 
     @pytest.mark.parametrize(
         "td_est",
@@ -929,7 +934,9 @@ class TestTD3(LossModuleTestBase):
         )
         return actor.to(device)
 
-    def _create_mock_value(self, batch=2, obs_dim=3, action_dim=4, device="cpu"):
+    def _create_mock_value(
+        self, batch=2, obs_dim=3, action_dim=4, device="cpu", out_keys=None
+    ):
         # Actor
         class ValueClass(nn.Module):
             def __init__(self):
@@ -943,6 +950,7 @@ class TestTD3(LossModuleTestBase):
         value = ValueOperator(
             module=module,
             in_keys=["observation", "action"],
+            out_keys=out_keys,
         )
         return value.to(device)
 
@@ -1193,7 +1201,6 @@ class TestTD3(LossModuleTestBase):
     def test_td3_tensordict_keys(self, td_est):
         actor = self._create_mock_actor()
         value = self._create_mock_value()
-        td = self._create_mock_data_td3()
         loss_fn = TD3Loss(
             actor,
             value,
@@ -1204,15 +1211,21 @@ class TestTD3(LossModuleTestBase):
             "state_action_value": "state_action_value",
             "action": "action",
         }
-        # TODO value needs state_action_value as out key for this test
-        # key_mapping = {"state_action_value": "value"}
 
         self.tensordict_keys_test(
             loss_fn,
             default_keys=default_keys,
             td_est=td_est,
-            # loss_advantage_key_mapping=key_mapping,
         )
+
+        value = self._create_mock_value(out_keys=["state_action_value_test"])
+        loss_fn = DDPGLoss(
+            actor,
+            value,
+            loss_function="l2",
+        )
+        key_mapping = {"state_action_value": ("value", "state_action_value_test")}
+        self.set_advantage_keys_through_loss_test(loss_fn, td_est, key_mapping)
 
 
 @pytest.mark.parametrize("version", [1, 2])
@@ -1657,14 +1670,22 @@ class TestSAC(LossModuleTestBase):
             "action": "action",
             "log_prob": "_log_prob",
         }
-        # key_mapping = {"value": "value"}
 
         self.tensordict_keys_test(
             loss_fn,
             default_keys=default_keys,
             td_est=td_est,
-            # loss_advantage_key_mapping=key_mapping,
         )
+
+        value = self._create_mock_value()
+        loss_fn = SACLoss(
+            actor,
+            value,
+            loss_function="l2",
+        )
+
+        key_mapping = {"value": ("value", "state_value_test")}
+        self.set_advantage_keys_through_loss_test(loss_fn, td_est, key_mapping)
 
 
 class TestDiscreteSAC(LossModuleTestBase):
@@ -2020,13 +2041,22 @@ class TestDiscreteSAC(LossModuleTestBase):
             "value": "state_value",
             "action": "action",
         }
-        # key_mapping = {"value": "value"}
         self.tensordict_keys_test(
             loss_fn,
             default_keys=default_keys,
             td_est=td_est,
-            # loss_advantage_key_mapping=key_mapping,
         )
+
+        qvalue = self._create_mock_qvalue()
+        loss_fn = DiscreteSACLoss(
+            actor_network=actor,
+            qvalue_network=qvalue,
+            num_actions=actor.spec["action"].space.n,
+            loss_function="l2",
+        )
+
+        key_mapping = {"value": ("value", "state_value_test")}
+        self.set_advantage_keys_through_loss_test(loss_fn, td_est, key_mapping)
 
 
 @pytest.mark.skipif(
@@ -2522,13 +2552,21 @@ class TestREDQ(LossModuleTestBase):
             "sample_log_prob": "sample_log_prob",
             "state_action_value": "state_action_value",
         }
-        # key_mapping = {"value": "value"}
         self.tensordict_keys_test(
             loss_fn,
             default_keys=default_keys,
             td_est=td_est,
-            # loss_advantage_key_mapping=key_mapping,
         )
+
+        qvalue = self._create_mock_qvalue()
+        loss_fn = REDQLoss(
+            actor_network=actor,
+            qvalue_network=qvalue,
+            loss_function="l2",
+        )
+
+        key_mapping = {"value": ("value", "state_value_test")}
+        self.set_advantage_keys_through_loss_test(loss_fn, td_est, key_mapping)
 
 
 class TestPPO(LossModuleTestBase):
@@ -2994,17 +3032,23 @@ class TestPPO(LossModuleTestBase):
             "sample_log_prob": "sample_log_prob",
             "action": "action",
         }
-        key_mapping = {
-            "advantage": "advantage",
-            "value_target": "value_target",
-            # "value": "value",
-        }
+
         self.tensordict_keys_test(
             loss_fn,
             default_keys=default_keys,
             td_est=td_est,
-            loss_advantage_key_mapping=key_mapping,
         )
+
+        value_key = "state_value_test"
+        value = self._create_mock_value(out_keys=[value_key])
+        loss_fn = loss_class(actor, value, loss_critic_type="l2")
+
+        key_mapping = {
+            "advantage": ("advantage", "advantage_new"),
+            "value_target": ("value_target", "value_target_new"),
+            "value": ("value", value_key),
+        }
+        self.set_advantage_keys_through_loss_test(loss_fn, td_est, key_mapping)
 
     @pytest.mark.parametrize("loss_class", (PPOLoss, ClipPPOLoss, KLPENPPOLoss))
     @pytest.mark.parametrize("advantage", ("gae", "td", "td_lambda", None))
@@ -3027,12 +3071,6 @@ class TestPPO(LossModuleTestBase):
         )
         actor = self._create_mock_actor()
         value = self._create_mock_value(out_keys=[tensor_keys["value"]])
-
-        adv_keys = {
-            key: value
-            for key, value in tensor_keys.items()
-            if key in asdict(GAE._AcceptedKeys()).keys()
-        }
 
         if advantage == "gae":
             advantage = GAE(
@@ -3062,6 +3100,12 @@ class TestPPO(LossModuleTestBase):
         loss_fn = loss_class(actor, value, loss_critic_type="l2")
         loss_fn.set_keys(**tensor_keys)
         if advantage is not None:
+            # collect tensordict key names for the advantage module
+            adv_keys = {
+                key: value
+                for key, value in tensor_keys.items()
+                if key in asdict(GAE._AcceptedKeys()).keys()
+            }
             advantage.set_keys(**adv_keys)
             advantage(td)
         else:
@@ -3336,17 +3380,23 @@ class TestA2C(LossModuleTestBase):
             "value": "state_value",
             "action": "action",
         }
-        key_mapping = {
-            "advantage": "advantage",
-            "value_target": "value_target",
-            # "value": "value",
-        }
+
         self.tensordict_keys_test(
             loss_fn,
             default_keys=default_keys,
             td_est=td_est,
-            loss_advantage_key_mapping=key_mapping,
         )
+
+        value = self._create_mock_value(out_keys=["value_state_test"])
+
+        loss_fn = A2CLoss(actor, value, loss_critic_type="l2")
+
+        key_mapping = {
+            "advantage": ("advantage", "advantage_test"),
+            "value_target": ("value_target", "value_target_test"),
+            "value": ("value", "value_state_test"),
+        }
+        self.set_advantage_keys_through_loss_test(loss_fn, td_est, key_mapping)
 
     @pytest.mark.parametrize("device", get_available_devices())
     def test_a2c_tensordict_keys_run(self, device):
@@ -3392,7 +3442,6 @@ class TestA2C(LossModuleTestBase):
         for name, p in named_parameters:
             if p.grad is not None and p.grad.norm() > 0.0:
                 assert "actor" not in name
-                assert "critic" in name
             if p.grad is None:
                 assert "actor" in name
                 assert "critic" not in name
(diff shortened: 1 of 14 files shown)


torchrl/objectives/ddpg.py
```diff
@@ -48,10 +48,17 @@ def __init__(
         *,
         loss_function: str = "l2",
         delay_actor: bool = False,
-        delay_value: bool = True,
+        delay_value: bool = False,
```
vmoens, line 51:
```suggestion
        delay_value: bool = True,
```

test/test_cost.py
vmoens, line 1: This is very comprehensive!
I'm just worried that because we're not grouping these in the losses themselves, when implementing a new loss we'll need to think about putting tests in 2 different places rather than one...

torchrl/objectives/common.py
```diff
@@ -73,6 +73,39 @@ def __init__(self):
...
+            >>> # initialize the DQN loss
+            >>> actor = torch.nn.Linear(3, 4)
+            >>> dqn_loss = DQNLoss(actor, action_space="one-hot")
+            >>> dqn_loss.set_keys(priority_key="td_error", action_value_key="action_value")
+        """
+        for key, value in kwargs.items():
+            if key not in self.tensordict_keys.keys():
+                raise ValueError(f"{key} not a valid tensordict key")
+            set_value = value if value is not None else self.tensordict_keys[key]
+            setattr(self, key, set_value)
```
vmoens, line 108: I'd rather have these keys contained in a separate container, like a dictionary or else.
I'm afraid that as the number of keys increase, we'll end up with a class with many attributes and no easy access to the list of them all.

Blonck: No, the container is already there. However, using the keys would be involve always one extra step.
```
tensordict.get(self.action_key) -> tensordict.get(self.tensordict_keys["action_key"])
```
From the data organisation it would be the better solution, but the code will be a little more noisy.

vmoens: I agree it's a bit clunky.
@matteobettini do you have an opinion on the matter?

test/test_cost.py
Blonck: For now I would like it to leave it as is. Putting these into the loss tests would lead to a huge duplication of code.

Blonck:
> Is there a way for `tensordict_keys` to be defined at the class level (and not instance)? Such that one can do `PPOLoss.tensordict_keys` without instantiating an object?

That would lead that the behavior of one instance of a loss module would change if `.set_keys()` is called somewhere else. I think that could surprise the user of torchrl. (Although, usually only one loss module is used.)

> Also, what about making `tensordict_keys` a property using @abc.abstractmethod in LossModule parent class, such that we force all new losses to have that attribute?

That is a good idea, I will try to implement this. It should lead to an easier interface.

torchrl/objectives/common.py
matteobettini: I prefer `tensordict.get(self.tensordict_keys["action_key"])`.

Also, would `loss_keys` make sense instead of `tensordict_keys`

vmoens: What about an Enum instead of a dict (to make sure only a finite set of keys is present)?
```
tensordict.get(self.loss_keys.action_key)
```

Blonck:
> Not necessarily.
> Here you clearly separate the default and non-default. The default should be defined at the class level if they're kept separated.

Got it. I've implemented this idea and it looks good. Although, I had to use a abstract staticmethod. Afaik, there is no static property decorator.

matteobettini:
Also one thing to mention is the the new keys have to be transparently reflected in the value_estimators
```
loss = loss()
loss.set_keys()
loss.make_value_estim()
```
and
```
loss = loss()
loss.make_value_estim()
loss.set_keys()
```
Should both work with the keys reflected in the value estimators

vmoens:
> Also one thing to mention is the the new keys have to be transparently reflected in the value_estimators
> 
> ```
> loss = loss()
> loss.set_keys()
> loss.make_value_estim()
> ```
> 
> and
> 
> ```
> loss = loss()
> loss.make_value_estim()
> loss.set_keys()
> ```
> 
> Should both work with the keys reflected in the value estimators

ideally keys of the value estimator should only be set through the value estimator no?

matteobettini:
But the value estimator is created by the loss.

do we want to make users to call set_keys twice?

Blonck:
> But the value estimator is created by the loss.
> 
> do we want to make users to call set_keys twice?
> 
> already calling it once is a added complexity for MARL users

That is a good point. This problem is introduced by allowing to configure the tensordict keys after constructing the losses.
I could solve this by extending `.set_keys(...)` so that it also sets the keys for the value estimator.
Another solution would be to remove `.set_keys(...)` and add a generic constructor argument for all loss modules, e.g., `tensordict_keys`.
If there is no need to configure the keys during runtime, I would prefer the later one, although the code becomes a bit clunky.

Blonck:
The solution with the ctor would roughly look like:
```python
class MyLoss(LossModule):
    def __init__(self, ...., tensordict_keys = {}):
         super().__init__(tensordict_keys=tensordict_keys)
         ...

class LossModule(nn.Module, ABC):
    def __init__(self, tensordict_keys={}):
        #merge default tensordict keys with provided keys
       default_keys = self.default_tensordict_keys()
       for key, value in tensordict_keys.items():
          if key not in default_keys:
              raise ValueError(...)
          default_keys[key] = value
      // create attributes
      ...
```

torchrl/objectives/common.py
Blonck: > What about an Enum instead of a dict (to make sure only a finite set of keys is present)?
This would mean constructing the enum in the base class from data provided by the child class. I guess this is possible, since almost everything is possible in python. However, the syntax would not look like
```
tensordict.get(self.loss_keys.action_key)
```
since the key must be converted to str (or pair of strings). StrEnum would be a solution, but are introduced in python 3.11.

Blonck:
Forwarding the new keys via `.set_keys()` to the value estimator would also require to maintain a mapping from key_name_loss to key_name_value_estimator. In the default case both key names are identical, see `ppo.py`
```python
# from ppo.py
def make_value_estimator(...):
    ...
    value_key = self.value_key
    if value_type == ValueEstimators.TD1:
        self._value_estimator = TD1Estimator(
            value_network=self.critic, value_key=value_key, **hp
        )
```
However, there is also this case `ddpg.py`:
```python
# from ddpg.py
def make_value_estimator(...):
    ...
    value_key = "state_action_value" # <- would correspond to self.state_action_value_key
    if value_type == ValueEstimators.TD1:
        self._value_estimator = TD1Estimator(
            value_network=self.actor_critic, value_key=value_key, **hp
        )
```

Blonck:
Moving the tensordict keys to the ctor would look like this:
https://github.com/Blonck/rl/compare/refactor_loss_keys...Blonck:rl:loss_keys_in_ctor?expand=1

As already mentioned using `.set_keys()`  would need a mapping between the key names of the loss module and the key names of the advantage. In this case, I would remove the code from the loss module base class and add a `.set_keys()` to each loss module separately which translates between the names.
```python
def set_keys(priority_key, state_action_value_key):
    self.priority_key=priority_key
    self.state_action_value_key=state_action_value_key

   self.value_estimator.set_keys(value_key=state_action_value_key)
```

vmoens:
> Another solution would be to remove `.set_keys(...)` and add a generic constructor argument for all loss modules, e.g., `tensordict_keys`.

Hmmm the whole point of this is to avoid polluting the constructor too much. I don't like passing dictionaries and complicated structures to constructors, it makes things quite unclear.

vmoens:
I'm still not sure about storing the keys at root level. 
It's hard to access them all in one go.
Can we have

```
class SACLoss:
    @dataclass
    class _AcceptedKeys:
        reward: str = "reward"
    default_keys = _AcceptedKeys()
    
    @classmethod
    def __new__(cls, ...):
        cls._tensor_keys = self._AcceptedKeys()

    @property
    def tensor_keys(self):
        return self._tensor_keys

    def set_keys(self, **kw):
        for key, value in kw.items():
            if key not in self._AcceptedKeys.__dict__:
                raise ValueError
        self._tensor_keys = self._AcceptedKeys(**kw)

SACLoss.default_keys # gets all the accepted keys and their values
SACLoss.tensor_keys # error
SACLoss(...).tensor_keys
```

Blonck:
> isn't pretty

I agree.

> and quite error prone

I think the alternative has other problems. Making the keys configurable after construction puts a burden on all other code paths to handle these case. Usually, `.set_keys()` will only be called directly after construction, but we should not rely on that. This is a pretty good example.

```python
loss = loss()
loss.make_value_estim()
loss.set_keys()
```

In general, I would keep things immutable which don't have to be mutable. In other languages one uses something like the builder pattern to construct more complex objects, `LossModuleBuilder().set_weights().build()`, but that is very uncommon for python and it would also not align with the pytorch base library.

Nevertheless, for now I would vote for reverting my changes and do the following.
* make ctor keys deprecated (as in the current patch)
* add `.set_keys()` to each loss module
* `.set_keys()` of each loss module calls `.set_keys()` of the advantage modules and maps between the key names manually
* there is no longer the need for a data structure to keep the keys, they become normal attributes `self.loss_keys["value_key"] -> self.value_key` again

This would duplicate a lot of code and also would need additional work whenever new losses or new tensordict keys are introduced. On the other hand the code stays very simple.

Blonck:
The reduce the noise on this branch I'm developing the suggested change here:
https://github.com/pytorch/rl/compare/main...Blonck:rl:refactor_loss_keys_advanced?expand=1

Blonck:
Btw, this code will raise an exception in case the value key name has a non-default value
```
loss = loss()
loss.make_value_estim()
loss.set_keys()
```

because
```
actor = # construct tensordict with non-default value key
loss = loss() # constructing loss with default key names works because the ctor doesn't use any key names
loss.make_value_estim() # will raise an exception because
```
The value estimator checks that the value key is in the value_network.out_keys.
Since the actor uses a non-default value key, but loss and hence the value estimator are instantiated with the default values this will raise an exception.

torchrl/objectives/a2c.py
```diff
@@ -67,24 +69,57 @@ class A2CLoss(LossModule):
...
+    class _AcceptedKeys:
+        """Maintains default values for all configurable tensordict keys.
+
+        This class defines which tensordict keys can be set using '.set_keys(key_name=key_value)' and their
+        default values.
+
+        Attributes:
+        ------------
+        advantage : NestedKey
+            The input tensordict key where the advantage is expected.
```
vmoens, line 82: Should follow the Google docstring formatting

```suggestion
        Attributes:
            advantage (NestedKey): The input tensordict key where the advantage is expected.
```

torchrl/objectives/common.py
```diff
@@ -74,6 +91,41 @@ def __init__(self):
         self.value_type = self.default_value_estimator
         # self.register_forward_pre_hook(_parameters_to_tensordict)
 
+    @abstractmethod
+    def _forward_value_estimator_keys(self, **kwargs) -> None:
+        """Passes updated tensordict keys to the underlying value estimator."""
+        ...
+
+    def _set_deprecated_ctor_keys(self, **kwargs) -> None:
+        """Helper function to set a tensordict key from a ctor and raise a warning simultaneously."""
```
vmoens, line 100:
```suggestion
        """Helper function to set a tensordict key from a constructor and raise a warning simultaneously."""
```

torchrl/objectives/common.py
```diff
@@ -74,6 +91,41 @@ def __init__(self):
...
+        for key, value in kwargs.items():
+            if value is not None:
+                warnings.warn(
+                    f"Setting '{key}' via ctor is deprecated, use .set_keys({key}='some_key') instead.",
```
vmoens, line 104:
```suggestion
                    f"Setting '{key}' via the constructor is deprecated, use .set_keys(<key>='some_key') instead.",
```

torchrl/objectives/ddpg.py
```diff
@@ -39,6 +40,26 @@ class DDPGLoss(LossModule):
...
 
+    @dataclass
+    class _AcceptedKeys:
+        """Maintains default values for all configurable tensordict keys.
+
+        This class defines which tensordict keys can be set using '.set_keys(key_name=key_value)' and their
+        default values
+
+        Attributes:
+        ------------
```
vmoens, line 51: ditto

torchrl/objectives/dqn.py
```diff
@@ -288,21 +304,58 @@ class DistributionalDQNLoss(LossModule):
...
 
+    @dataclass
+    class _AcceptedKeys:
+        """Maintains default values for all configurable tensordict keys.
+
+        This class defines which tensordict keys can be set using '.set_keys(key_name=key_value)' and their
+        default values
+
+        Attributes:
+        ------------
```
vmoens, line 324: Ditto

torchrl/objectives/dreamer.py
```diff
@@ -49,6 +51,44 @@ class DreamerModelLoss(LossModule):
...
 
+    @dataclass
+    class _AcceptedKeys:
+        """Maintains default values for all configurable tensordict keys.
+
+        This class defines which tensordict keys can be set using '.set_keys(key_name=key_value)' and their
+        default values
+
+        Attributes:
+        ------------
```
vmoens, line 62: ditto

torchrl/objectives/dreamer.py
```diff
@@ -49,6 +51,44 @@ class DreamerModelLoss(LossModule):
...
+        reco_pixels : NestedKey
+            The reconstruction pixels is expected to be in the tensordict key ("next", reco_pixels). Defaults to ``"reco_pixels"``.
+        """
+
+        reward: NestedKey = "reward"
+        true_reward: NestedKey = "true_reward"
+        prior_mean: NestedKey = "prior_mean"
+        prior_std: NestedKey = "prior_std"
+        posterior_mean: NestedKey = "posterior_mean"
+        posterior_std: NestedKey = "posterior_std"
```
vmoens, line 86: Not 100% sure these guys belong here.
It's true that they are input to the loss module.
On the other hand, they are specific to our implementation of dreamer: if one is to change their names, it is likely that that person is also changing a lot of other things in the algo (ie it is a highly integrated feature).
Happy to keep them as part of this PR though, too many keys is better than too few.

torchrl/objectives/dreamer.py
```diff
@@ -155,6 +201,32 @@ class DreamerActorLoss(LossModule):
...
 
+    @dataclass
+    class _AcceptedKeys:
+        """Maintains default values for all configurable tensordict keys.
+
+        This class defines which tensordict keys can be set using '.set_keys(key_name=key_value)' and their
+        default values.
+
+        Attributes:
+        ------------
```
vmoens, line 212: ditto

torchrl/objectives/iql.py
```diff
@@ -55,9 +54,41 @@ class IQLLoss(LossModule):
...
 
+    @dataclass
+    class _AcceptedKeys:
+        """Maintains default values for all configurable tensordict keys.
+
+        This class defines which tensordict keys can be set using '.set_keys(key_name=key_value)' and their
+        default values
+
+        Attributes:
+        ------------
```
vmoens, line 70: ditto

torchrl/objectives/ppo.py
```diff
@@ -133,8 +167,17 @@ def __init__(
         normalize_advantage: bool = False,
         gamma: float = None,
         separate_losses: bool = False,
+        advantage_key: str = None,
+        value_target_key: str = None,
+        value_key: str = None,
```
vmoens, line 172: value_key depends on the value network. But if I'm not mistaken, we have not set the value network yet. How does this not produce an exception? What if the value_key has a value that matches the value network output key but is not the one usually expected?

torchrl/objectives/redq.py
```diff
@@ -75,9 +74,41 @@ class REDQLoss(LossModule):
...
 
+    @dataclass
+    class _AcceptedKeys:
+        """Maintains default values for all configurable tensordict keys.
+
+        This class defines which tensordict keys can be set using '.set_keys(key_name=key_value)' and their
+        default values
+
+        Attributes:
+        ------------
```
vmoens, line 91: ditto

torchrl/objectives/reinforce.py
```diff
@@ -66,24 +69,59 @@ class ReinforceLoss(LossModule):
...
 
+    @dataclass
+    class _AcceptedKeys:
+        """Maintains default values for all configurable tensordict keys.
+
+        This class defines which tensordict keys can be set using '.set_keys(key_name=key_value)' and their
+        default values.
+
+        Attributes:
+        ------------
```
vmoens, line 80: ditto

torchrl/objectives/td3.py
```diff
@@ -59,6 +61,29 @@ class TD3Loss(LossModule):
...
 
+    @dataclass
+    class _AcceptedKeys:
+        """Maintains default values for all configurable tensordict keys.
+
+        This class defines which tensordict keys can be set using '.set_keys(key_name=key_value)' and their
+        default values.
+
+        Attributes:
+        ------------
```
vmoens, line 72: ditto

torchrl/objectives/value/advantages.py
```diff
@@ -64,8 +66,44 @@ class ValueEstimatorBase(TensorDictModuleBase):
...
 
+    @dataclass
+    class _AcceptedKeys:
+        """Maintains default values for all configurable tensordict keys.
+
+        This class defines which tensordict keys can be set using '.set_keys(key_name=key_value)' and their
+        default values.
+
+        Attributes:
+        ------------
```
vmoens, line 77: ditto

torchrl/objectives/value/advantages.py
```diff
@@ -192,16 +292,18 @@ class TD0Estimator(ValueEstimatorBase):
...
-        value_target_key (str or tuple of str, optional): the key of the advantage entry.
-            Defaults to "value_target".
-        value_key (str or tuple of str, optional): the value key to read from the input tensordict.
-            Defaults to "state_value".
         skip_existing (bool, optional): if ``True``, the value network will skip
             modules which outputs are already present in the tensordict.
             Defaults to ``None``, ie. the value of :func:`tensordict.nn.skip_existing()`
             is not affected.
+        tensor_keys: (Dict[str, str or tuple of str], optional) Specify tensordict key names for
+            all keys specified in _AcceptedKeys.
```
vmoens, line 300:
```suggestion
            all keys specified in :class:`~._AcceptedKeys`.
```

torchrl/objectives/value/advantages.py
```diff
@@ -366,16 +468,18 @@ class TD1Estimator(ValueEstimatorBase):
...
```
vmoens, line 476:
```suggestion
            all keys specified in :class:`~._AcceptedKeys`. 
```

torchrl/objectives/value/advantages.py
```diff
@@ -544,16 +648,18 @@ class TDLambdaEstimator(ValueEstimatorBase):
...
```
vmoens, line 656:
```suggestion
            all keys specified in :class:`~._AcceptedKeys`.
```

torchrl/objectives/value/advantages.py
```diff
@@ -192,16 +292,18 @@ class TD0Estimator(ValueEstimatorBase):
...
+        advantage_key (str or tuple of str, optional): [Deprecated] the key of the advantage entry.
+            Defaults to "advantage".
```
vmoens, line 302:
```suggestion
            Defaults to ``"advantage"``.
```

torchrl/objectives/value/advantages.py
```diff
@@ -192,16 +292,18 @@ class TD0Estimator(ValueEstimatorBase):
...
+        value_target_key (str or tuple of str, optional): [Deprecated] the key of the advantage entry.
+            Defaults to "value_target".
```
vmoens, line 304:
```suggestion
            Defaults to ``"value_target"``.
```

torchrl/objectives/value/advantages.py
```diff
@@ -192,16 +292,18 @@ class TD0Estimator(ValueEstimatorBase):
...
+        value_key (str or tuple of str, optional): [Deprecated] the value key to read from the input tensordict.
+            Defaults to "state_value".
```
vmoens, line 306:
```suggestion
            Defaults to ``"state_value"``.
```

torchrl/objectives/value/advantages.py
```diff
@@ -366,16 +468,18 @@ class TD1Estimator(ValueEstimatorBase):
...
-        value_key (str or tuple of str, optional): the value key to read from the input tensordict.
-            Defaults to "state_value".
         skip_existing (bool, optional): if ``True``, the value network will skip
             modules which outputs are already present in the tensordict.
             Defaults to ``None``, ie. the value of :func:`tensordict.nn.skip_existing()`
             is not affected.
+        tensor_keys: (Dict[str, str or tuple of str], optional) Specify tensordict key names for
+            all keys specified in _AcceptedKeys.
+        advantage_key (str or tuple of str, optional): [Deprecated] the key of the advantage entry.
+            Defaults to "advantage".
```
vmoens, line 478:
```suggestion
            Defaults to ``"advantage"``.
```

torchrl/objectives/value/advantages.py
```diff
@@ -366,16 +468,18 @@ class TD1Estimator(ValueEstimatorBase):
...
+        value_target_key (str or tuple of str, optional): [Deprecated] the key of the advantage entry.
+            Defaults to "value_target".
```
vmoens, line 480:
```suggestion
            Defaults to ``"value_target"``.
```

torchrl/objectives/value/advantages.py
```diff
@@ -366,16 +468,18 @@ class TD1Estimator(ValueEstimatorBase):
...
+        value_key (str or tuple of str, optional): [Deprecated] the value key to read from the input tensordict.
+            Defaults to "state_value".
```
vmoens, line 482:
```suggestion
            Defaults to ``"state_value"``.
```

torchrl/objectives/value/advantages.py
```diff
@@ -544,16 +648,18 @@ class TDLambdaEstimator(ValueEstimatorBase):
...
-        value_key (str or tuple of str, optional): the value key to read from the input tensordict.
-            Defaults to "state_value".
         skip_existing (bool, optional): if ``True``, the value network will skip
             modules which outputs are already present in the tensordict.
             Defaults to ``None``, ie. the value of :func:`tensordict.nn.skip_existing()`
             is not affected.
+        tensor_keys: (Dict[str, str or tuple of str], optional) Specify tensordict key names for
+            all keys specified in _AcceptedKeys.
+        advantage_key (str or tuple of str, optional): [Deprecated] the key of the advantage entry.
+            Defaults to "advantage".
```
vmoens, line 658:
```suggestion
            Defaults to ``"advantage"``.
```

torchrl/objectives/value/advantages.py
```diff
@@ -544,16 +648,18 @@ class TDLambdaEstimator(ValueEstimatorBase):
...
+        value_target_key (str or tuple of str, optional): [Deprecated] the key of the advantage entry.
+            Defaults to "value_target".
```
vmoens, line 660:
```suggestion
            Defaults to ``"value_target"``.
```

torchrl/objectives/value/advantages.py
```diff
@@ -544,16 +648,18 @@ class TDLambdaEstimator(ValueEstimatorBase):
...
+        value_key (str or tuple of str, optional): [Deprecated] the value key to read from the input tensordict.
+            Defaults to "state_value".
```
vmoens, line 662:
```suggestion
            Defaults to ``"state_value"``.
```

vmoens (commented):
Great work! Must have been quite a headache to come about!
See my comments in the code

vmoens (commented):
Great work! Must have been quite a headache to come about!
See my comments in the code

torchrl/objectives/dreamer.py
Blonck: I had the same thought, but left them in because also for some other tensordict keys in other loss modules the same is true. In the end it felt more consistent to make all input keys configurable, especially it comes with no cost for the user when they just want the default behaviour.

torchrl/objectives/common.py
```diff
@@ -74,6 +91,41 @@ def __init__(self):
...
+        """Set tensordict key names.
+
+        Examples:
+            >>> from torchrl.objectives import DQNLoss
+            >>> # initialize the DQN loss
+            >>> actor = torch.nn.Linear(3, 4)
+            >>> dqn_loss = DQNLoss(actor, action_space="one-hot")
+            >>> dqn_loss.set_keys(priority_key="td_error", action_value_key="action_value")
+        """
+        for key, value in kwargs.items():
```
vmoens, line 119: if we make `_AcceptedKeys` optional, we can raise an exception if it is not present?

vmoens (approved):
Great work!
I'm happy to merge this but I have one more question: we have a tutorial about losses and this feature seems pretty advanced for a regular user that would like to code up a new loss with hard coded keys.
Is it mandatory to code `_AcceptedKeys`?
It's a great tool but maybe we can reserve it for internal usage.

Have you checked that this tutorial is running under this PR?

torchrl/objectives/common.py
```diff
@@ -59,11 +61,26 @@ class LossModule(nn.Module):
...
     default_value_estimator: ValueEstimators = None
     SEP = "_sep_"
 
+    @property
+    def tensor_keys(self) -> _AcceptedKeys:
+        return self._tensor_keys
+
     def __new__(cls, *args, **kwargs):
         cls.forward = set_exploration_type(ExplorationType.MODE)(cls.forward)
+        cls._tensor_keys = cls._AcceptedKeys()
```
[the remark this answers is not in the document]
vmoens: Got it
Up to you for the exception. In a way, if someone writes a loss module then calls set_keys without having written a set of keys they're probably way off the road...

PR #1182, opened 2023-05-22
[CI] Merge Distributed and Linux GPU job

osalpekar:
Runs the distributed tests as part of the Linux CPU/GPU jobs. Confirmed the distributed tests are passing. Failing tests seen on this PR for the Linux jobs are also failing on master.

### DIFF
--- a/.circleci/unittest/linux/scripts/run_test.sh
+++ b/.circleci/unittest/linux/scripts/run_test.sh
@@ -34,7 +34,7 @@ print(model_utils)
 
 python3 .circleci/unittest/helpers/coverage_run_parallel.py -m pytest test/smoke_test.py -v --durations 200
 python3 .circleci/unittest/helpers/coverage_run_parallel.py -m pytest test/smoke_test_deps.py -v --durations 200 -k 'test_gym or test_dm_control_pixels or test_dm_control or test_tb'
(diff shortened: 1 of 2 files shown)


osalpekar:
cc @vmoens

.circleci/unittest/linux/scripts/run_test.sh
```diff
@@ -21,6 +21,6 @@ export CKPT_BACKEND=torch
 
 python .circleci/unittest/helpers/coverage_run_parallel.py -m pytest test/smoke_test.py -v --durations 200
 python .circleci/unittest/helpers/coverage_run_parallel.py -m pytest test/smoke_test_deps.py -v --durations 200 -k 'test_gym or test_dm_control_pixels or test_dm_control or test_tb'
-python .circleci/unittest/helpers/coverage_run_parallel.py -m pytest --instafail -v --durations 200 --ignore test/test_distributed.py --ignore test/test_rlhf.py
+python .circleci/unittest/helpers/coverage_run_parallel.py -m pytest --instafail -v --durations 200
```
vmoens, line 24: (this file isn't used anymore I believe)
test_rlhf should still be excluded

vmoens (approved):
LGTM
Don't we want to also deprecate the distributed workflows while we're at it?

.circleci/unittest/linux/scripts/run_test.sh
osalpekar: Yeah we can probably just delete this file in a follow-up since it's not used anymore.

PR #1186, opened 2023-05-23
[BugFix] Fix TD3 target net

vmoens:


vmoens:
@BY571 I made some more changes and I have a good training curve now
Can you check if it all makes sense?

BY571:
> @BY571 I made some more changes and I have a good training curve now Can you check if it all makes sense?

looking good! Also getting the "old" performance now with the correct delayed values :)

vmoens:
> > @BY571 I made some more changes and I have a good training curve now Can you check if it all makes sense?
> 
> looking good! Also getting the "old" performance now with the correct delayed values :)

Do you have the permission make a review? That would be great!

examples/td3/config.yaml
```diff
@@ -1,21 +1,15 @@
...
 env_name: HalfCheetah-v4
-env_task: ""
 exp_name: "debugging"
-env_library: gym
-record_video: 0
-normalize_rewards_online: 0
-normalize_rewards_online_scale: 5
-normalize_rewards_online_decay: 0.99
 total_frames: 1000000
-frames_per_batch: 1000
```
BY571, line 11: Not a big issue but I might put this back to 1000 as its a common length for MuJoCo envs

examples/td3/config.yaml
```diff
@@ -1,21 +1,15 @@
...
+frames_per_batch: 200
 max_frames_per_traj: 1000
 frame_skip: 1
 from_pixels: 0
 seed: 0
 
 # Collection
-init_random_frames: 25000
```
BY571, line 18: Here also not big issue but exact paper value would be 10k

examples/td3/config.yaml
vmoens: not sure
this is the frames in each batch but not the truncation length.
This should be handled by `max_frames_per_traj`

examples/td3/config.yaml
vmoens: Got it

examples/td3/config.yaml
BY571: Oh, yes! I was thinking of env like halfcheetah. But you are right for Hopper or others where episodes can be less than 1000 so multiple epochs per batch!

BY571:
looks good to me now!

PR #1192, opened 2023-05-24
[BugFix] Fix SAC alpha optim

vmoens:


BY571:
looks good to me and performance verifies it!

PR #1196, opened 2023-05-26
[Feature] KL Transform for RLHF

vmoens:
cc @tcbegley @apbard


torchrl/envs/transforms/rlhf.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # update the in_keys for dispatch etc
+        self.in_keys = self.in_keys + actor.in_keys
+
+        # check that the model has parameters
+        params = make_functional(
+            actor, keep_params=False, funs_to_decorate=["forward", "get_dist"]
+        )
+        self.functional_actor = deepcopy(actor)
+        repopulate_module(actor, params)
```
tcbegley, line 115: Why do we repopulate here when below we seem to only call with `params=self.frozen_params`? Is this so that the caller can continue to use `actor` having supplied it as an argument to the constructor?

tcbegley (approved):
LGTM!

apbard (approved):

PR #1201, opened 2023-05-29
Add support for Unity MLAgents environments.

### LINKED ISSUE #1110: [Feature Request] Unity MLAgents Support
## Motivation

Unity is a popular method for creating game environments for RL algorithms. I was wondering if potentially adding support for [Unity's MLAgents framework](https://unity.com/products/machine-learning-agents) could be considered. Here is some documentation to their low level API that could be useful for supporting Unity games:

https://github.com/Unity-Technologies/ml-agents/blob/develop/docs/Python-LLAPI.md
https://github.com/Unity-Technologies/ml-agents/blob/develop/docs/Python-LLAPI-Documentation.md

## Solution

Adding direct support for Unity MLAgent environments.

## Checklist

hyerra:
## Description

This adds support for Unity MLAgent environments which is a popular method of creating games for reinforcement learning settings.

This pull request is still definitely a work-in-progress, and I would love to get feedback/suggestions on how to make this better. I still have to write unit tests and update documentation, but wanted to get some general input on this work before I started making these.

The code right now should work well for single agent environments. However, we need support for heterogeneous spaces in order for this to work well with multiple agents (or even a single agent with different types of observations). Ideally, we would need some kind of Tuple support.

A big limitation of this code though is that in multi-agent reinforcement learning settings, it requires all of the agents to have the same type of observation and action space because the spec created in _make_specs must be fixed. It would be nice if we could have specs defined on a per-agent level in order to fix this. I'm also open to other suggestions as well! Also, one other important point to note is that the number of agents is not necessarily fixed. New agents may be spawned and sometimes agents might not request a decision at a timestamp (which means there might be more/fewer agents requesting decisions from time to time). This is an issue only for multi-agent settings but not single-agent ones.

## Motivation and Context

`close #1110`

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/libs/unity.py
+++ b/torchrl/envs/libs/unity.py
@@ -1,0 +1,369 @@
+from warnings import warn
+
+import numpy as np
+import torch
+from tensordict.tensordict import TensorDict, TensorDictBase
+from torchrl.data.tensor_specs import (
+    BoundedTensorSpec,
+    CompositeSpec,
+    DiscreteTensorSpec,
+    MultiDiscreteTensorSpec,
+    UnboundedContinuousTensorSpec,
+    UnboundedDiscreteTensorSpec,
+)
+from torchrl.data.utils import numpy_to_torch_dtype_dict
+from torchrl.envs.common import _EnvWrapper
+from torchrl.envs.utils import _classproperty
+
+IMPORT_ERR = None
+try:
+    from mlagents_envs.base_env import ActionSpec, ActionTuple, BaseEnv, ObservationSpec
+    from mlagents_envs.environment import SideChannel, UnityEnvironment
+
+    _has_mlagents = True
+except ImportError as err:
+    _has_mlagents = False
+    IMPORT_ERR = err
+
+
+__all__ = ["UnityWrapper", "UnityEnv"]
+
+
+def _unity_to_torchrl_spec_transform(spec, dtype=None, device="cpu"):
+    """Maps the Unity specs to the TorchRL specs."""
+    if isinstance(spec, ObservationSpec):
+        shape = spec.shape
+        if not len(shape):
+            shape = torch.Size([1])
+        dtype = numpy_to_torch_dtype_dict[dtype]
+        return UnboundedContinuousTensorSpec(shape, device=device, dtype=dtype)
+    elif isinstance(spec, ActionSpec):
+        if spec.continuous_size == len(spec.discrete_branches) == 0:
+            raise ValueError("No available actions")
+        action_mask_spec = None
+        d_spec = c_spec = None
+        if spec.discrete_size == 1:
+            d_spec = DiscreteTensorSpec(
+                spec.discrete_branches[0], shape=[spec.discrete_size], device=device
+            )
+        else:
+            d_spec = MultiDiscreteTensorSpec(
+                spec.discrete_branches, shape=[spec.discrete_size], device=device
+            )
+            # FIXME: Need tuple support as action masks are 2D arrays
+            # action_mask_spec = MultiDiscreteTensorSpec(spec.discrete_branches, dtype=torch.bool, device=device)
+
+        if spec.continuous_size > 0:
+            dtype = numpy_to_torch_dtype_dict[dtype]
+            c_spec = BoundedTensorSpec(
+                -1, 1, (spec.continuous_size,), dtype=dtype, device=device
+            )
+
+        if d_spec and c_spec:
+            return CompositeSpec(discrete=d_spec, continuous=c_spec), action_mask_spec
+        else:
+            return d_spec if d_spec else c_spec, action_mask_spec
+    else:
+        raise TypeError(f"Unknown spec of type {type(spec)} passed")
+
+
+class UnityWrapper(_EnvWrapper):
+    """Unity environment wrapper.
+
+    Examples:
+        >>> env = UnityWrapper(
+        ...     UnityEnvironment(
+        ...         "<<PATH TO UNITY APP>>",
+        ...         side_channels=[],
+        ...         additional_args=[],
+        ...         log_folder=<<PATH TO WHERE LOGS SHOULD BE STORED>>,
+        ...         device=device,
+        ...     )
+        ... )
+    """
+
+    git_url = "https://github.com/Unity-Technologies/ml-agents"
+    libname = "mlagents_envs"
+
+    def __init__(self, env=None, **kwargs):
+        if env is not None:
+            kwargs["env"] = env
+        super().__init__(**kwargs)
+
+    def _init_env(self):
+        self._behavior_names = []
+
+    def _set_seed(self, seed: int | None):
+        warn(
+            "Seeding through _set_seed has not been implemented. Please set the "
+            "seed when you create the environment."
+        )
+
+    @_classproperty
+    def available_envs(cls) -> list[str]:
+        raise NotImplementedError("You must provide your own environments")
+
+    def _build_env(self, env: BaseEnv):
+        if not env.behavior_specs:
+            # Take a single step so that the brain information will be sent over
+            env.step()
+        return env
+
+    def _make_specs(self, env: BaseEnv) -> None:
+        # TODO: Behavior specs are immutable but new ones
+        # can be added if they are created in the environment.
+        # Need to account for behavior specs that are added
+        # throughout the environment lifecycle.
+
+        # IMPORTANT: This assumes that all agents have the same
+        # observations and actions. To change this, we need
+        # some method to allow for different specs depending on
+        # agent.
+        #
+        # A different `Parallel` version of this environment could be
+        # made where the number of agents is fixed, and then you stack
+        # all of the observations together. This design would allow
+        # different observations and actions, but would require
+        # a fixed agent count. The difficulty with implementing a
+        # `Parallel` version though is that not all agents will request
+        # a decision, so the spec would have to change depending
+        # on which agents request a decision.
+
+        first_behavior_name = next(iter(env.behavior_specs.keys()))
+        behavior_unity_spec = env.behavior_specs[first_behavior_name]
+        observation_specs = [
+            _unity_to_torchrl_spec_transform(
+                spec, dtype=np.dtype("float32"), device=self.device
+            )
+            for spec in behavior_unity_spec.observation_specs
+        ]
+        behavior_id_spec = UnboundedDiscreteTensorSpec(1, device=self.device)
+        agent_id_spec = UnboundedDiscreteTensorSpec(1, device=self.device)
+        # FIXME: Need Tuple support here so we can support observations of varying dimensions.
+        # Thus, for now we use only the first observation.
+        self.observation_spec = CompositeSpec(
+            observation=observation_specs[0],
+            behavior_id=behavior_id_spec,
+            agent_id=agent_id_spec,
+        )
+        self.action_spec, self.action_mask_spec = _unity_to_torchrl_spec_transform(
+            behavior_unity_spec.action_spec, dtype=np.int32, device=self.device
+        )
+        self.reward_spec = UnboundedContinuousTensorSpec(shape=[1], device=self.device)
+        self.done_spec = DiscreteTensorSpec(
+            n=2, shape=[1], dtype=torch.bool, device=self.device
+        )
+
+    def __repr__(self) -> str:
+        return (
+            f"{self.__class__.__name__}(env={self._env}, batch_size={self.batch_size})"
+        )
+
+    def _check_kwargs(self, kwargs: dict):
+        if "env" not in kwargs:
+            raise TypeError("Could not find environment key 'env' in kwargs.")
+        env = kwargs["env"]
+        if not isinstance(env, BaseEnv):
+            raise TypeError("env is not of type 'mlagents_envs.base_env.BaseEnv'.")
+        if "frame_skip" in kwargs and kwargs["frame_skip"] != 1:
+            # This functionality is difficult to support because not all agents will request
+            # decisions at each timestep and different agents might request decisions at
+            # different timesteps. This makes it difficult to do things like keep track
+            # of rewards.
+            raise ValueError(
+                "Currently, frame_skip is not supported for Unity environments."
+            )
+
+    def behavior_id_to_name(self, behavior_id: int):
+        raise self._behavior_names[behavior_id]
+
+    def read_reward(self, reward):
+        return self.reward_spec.encode(reward, ignore_device=True)
+
+    def read_obs(self, obs: np.ndarray, behavior_name: str, agent_id: int):
+        behavior_id = self._behavior_names.index(behavior_name)
+        observations = self.observation_spec.encode(
+            {"observation": obs, "behavior_id": behavior_id, "agent_id": agent_id},
+            ignore_device=True,
+        )
+        return observations
+
+    def read_action(self, action):
+        action = self.action_spec.to_numpy(action, safe=False)
+        # We expand the dimensions at the 0 axis because actions are
+        # defined to be 2D arrays with an extra first dimension being
+        # used for the number of agents in the game.
+        if action.ndim == 0:
+            action = np.expand_dims(action, axis=0)
+        if isinstance(self.action_spec, CompositeSpec):
+            action = self.action_spec.to_numpy(action, safe=False)
+            continuous_action = np.expand_dims(action["continuous"], axis=0)
+            discrete_action = np.expand_dims(action["discrete"], axis=0)
+            action = ActionTuple(continuous_action, discrete_action)
+        elif isinstance(self.action_spec, DiscreteTensorSpec | MultiDiscreteTensorSpec):
+            action = np.expand_dims(action, axis=0)
+            action = ActionTuple(None, action)
+        else:
+            action = self.action_spec.to_numpy(action)
+            action = np.expand_dims(action, axis=0)
+            action = ActionTuple(action, None)
+        return action
+
+    def read_action_mask(self, action_mask):
+        # if not self.action_mask_spec:
+        #     return None
+        # return self.action_mask_spec.encode(action_mask)
+        return None
+
+    def read_done(self, done):
+        return self.done_spec.encode(done)
+
+    def _behavior_name_update(self):
+        self._live_behavior_names = list(self.behavior_specs.keys())
+        for k in self._live_behavior_names:
+            if k not in self._behavior_names:
+                # We only add to self._behavior_names if the
+                # behavior name doesn't exist. This helps us
+                # ensure that the index of old behaviors stays
+                # the same and that we don't have duplicate entries.
+                # This is important since we use the index of the behavior
+                # name as an id for that behavior.
+                self._behavior_names.append(k)
+
+    def _batch_update(self, behavior_name):
+        self._current_step_idx = 0
+        self._current_behavior_name = behavior_name
+        self._decision_steps, self._terminal_steps = self.get_steps(behavior_name)
+
+    def _get_next_tensordict(self):
+        num_steps = len(self._decision_steps) + len(self._terminal_steps)
+        if self._current_step_idx >= num_steps:
+            raise ValueError("All agents already have actions")
+        done = False if self._current_step_idx < len(self._decision_steps) else True
+        steps = self._decision_steps if not done else self._terminal_steps
+        agent_id = steps.agent_id[self._current_step_idx]
+        step = steps[agent_id]
+        # FIXME: Need Tuple support here so we can support observations of varying dimensions.
+        # Thus, for now we use only the first observation.
+        obs, reward = step.obs[0], step.reward
+        tensordict_out = TensorDict(
+            source=self.read_obs(
+                obs, behavior_name=self._current_behavior_name, agent_id=agent_id
+            ),
+            batch_size=self.batch_size,
+            device=self.device,
+        )
+        # tensordict_out.set("action_mask", self.read_action_mask(action_mask))
+        tensordict_out.set("reward", self.read_reward(reward))
+        tensordict_out.set("done", self.read_done(done))
+        return tensordict_out
+
+    def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
+        # We step through each agent one at a time, and only perform
+        # an environment step (send the actions to the environment)
+        # once all agents are processed. This is because Unity requires
+        # all agents to have actions before stepping or else the zero
+        # action will be sent.
+        #
+        # In order to step through each agent, we first iterate through
+        # all behaviors, and determine actions for each agent in that behavior
+        # and then repeat the loop until all behavior's and their
+        # agents are accounted for. We then perform an environment step.
+        action = tensordict.get("action")
+        unity_action = self.read_action(action)
+        self.set_action_for_agent(
+            self._current_behavior_name, tensordict.get("agent_id").item(), unity_action
+        )
+        self._current_step_idx += 1
+        try:
+            tensordict_out = self._get_next_tensordict()
+            return tensordict_out.select().set("next", tensordict_out)
+        except ValueError:
+            behavior_id = self._live_behavior_names.index(self._current_behavior_name)
+            # If we have more behaviors to go through, keep continuing. Otherwise step the environment and
+            # then continue again.
+            if behavior_id < len(self._live_behavior_names) - 1:
+                self._current_behavior_name = self._live_behavior_names[behavior_id + 1]
+                self._batch_update(self._current_behavior_name)
+                tensordict_out = self._get_next_tensordict()
+                return tensordict_out.select().set("next", tensordict_out)
+            else:
+                self._env.step()
+                self._behavior_name_update()
+                self._batch_update(self._live_behavior_names[0])
+                tensordict_out = self._get_next_tensordict()
+                return tensordict_out.select().set("next", tensordict_out)
+
+    def _reset(self, tensordict: TensorDictBase | None = None, **kwargs):
+        self._env.reset(**kwargs)
+        self._behavior_name_update()
+        self._batch_update(self._live_behavior_names[0])
+        return self._get_next_tensordict()
+
+
+class UnityEnv(UnityWrapper):
+    """Unity environment wrapper.
+
+    Examples:
+        >>> env = UnityEnv(
+        ...     "<<PATH TO UNITY APP>>",
+        ...     side_channels=[],
+        ...     additional_args=[],
+        ...     log_folder=<<PATH TO WHERE LOGS SHOULD BE STORED>>,
+        ...     device=device,
+        ... )
+    """
+
+    def __init__(
+        self,
+        file_name: str | None = None,
+        seed: int = 0,
+        no_graphics: bool = False,
+        timeout_wait: int = 60,
+        side_channels: list[SideChannel] | None = None,
+        log_folder: str | None = None,
+        **kwargs,
+    ):
+        kwargs["file_name"] = file_name
+        kwargs["seed"] = seed
+        kwargs["no_graphics"] = no_graphics
+        kwargs["timeout_wait"] = timeout_wait
+        kwargs["side_channels"] = side_channels
+        kwargs["log_folder"] = log_folder
+        super().__init__(**kwargs)
+
+    def _check_kwargs(self, kwargs: dict):
+        if "file_name" not in kwargs:
+            raise TypeError("Could not find environment key 'file_name' in kwargs.")
+
+    def _build_env(
+        self,
+        file_name: str | None = None,
+        seed: int = 0,
+        no_graphics: bool = False,
+        timeout_wait: int = 60,
+        side_channels: list[SideChannel] | None = None,
+        log_folder: str | None = None,
+        **env_kwargs,
+    ):
+        if not _has_mlagents:
+            raise RuntimeError(
+                f"Unity MLAgents not found, unable to create environment. "
+                f"Consider downloading and installing Unity MLAgents from"
+                f" {self.git_url}"
+            )
+        self.file_name = file_name
+        return super()._build_env(
+            UnityEnvironment(
+                file_name,
+                seed=seed,
+                no_graphics=no_graphics,
+                timeout_wait=timeout_wait,
+                side_channels=side_channels,
+                log_folder=log_folder,
+                **env_kwargs,
+            )
+        )
+
+    def __repr__(self):
+        return f"{super().__repr__()}(file_name={self.file_name})"


hyerra:
cc: @vmoens @matteobettini

vmoens:
> ## Description
> This adds support for Unity MLAgent environments which is a popular method of creating games for reinforcement learning settings.
> 
> This pull request is still definitely a work-in-progress, and I would love to get feedback/suggestions on how to make this better. I still have to write unit tests and update documentation, but wanted to get some general input on this work before I started making these.
> 
> The code right now should work well for single agent environments. However, we need support for heterogeneous spaces in order for this to work well with multiple agents (or even a single agent with different types of observations). Ideally, we would need some kind of Tuple support.
> 
> A big limitation of this code though is that in multi-agent reinforcement learning settings, it requires all of the agents to have the same type of observation and action space because the spec created in _make_specs must be fixed. It would be nice if we could have specs defined on a per-agent level in order to fix this. I'm also open to other suggestions as well! Also, one other important point to note is that the number of agents is not necessarily fixed. New agents may be spawned and sometimes agents might not request a decision at a timestamp (which means there might be more/fewer agents requesting decisions from time to time). This is an issue only for multi-agent settings but not single-agent ones.
> 
> ## Motivation and Context
> Why is this change required? What problem does it solve? `close #1110`
> 
> * [x]  I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)
> 
> ## Types of changes
> What types of changes does your code introduce? Remove all that do not apply:
> 
> * [x]  New feature (non-breaking change which adds core functionality)
> 
> ## Checklist
> Go over all the following points, and put an `x` in all the boxes that apply. If you are unsure about any of these, don't hesitate to ask. We are here to help!
> 
> * [x]  I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
> * [x]  My change requires a change to the documentation.
> * [ ]  I have updated the tests accordingly (_required for a bug fix or a new feature_).
> * [ ]  I have updated the documentation accordingly.

Thanks so much for this contribution, really appreciated.

A few high level remarks/questions:
- A quick description of how to install MLAgents would be awesome. Along the same line of thoughts: how difficult would it be to write a dedicated CI workflow to test the environments execution? Is there a docker or anything we can use?
- The specs can be stacked together even if they are heterogeneous.
This feature needs a bit of polishing but I believe this kind of use cases is what we need to make it work by fireproofing it!
Here's an example:
```python
from torchrl.data import CompositeSpec, UnboundedContinuousTensorSpec as Spec
import torch

spec1 = CompositeSpec(obs=Spec(shape=(3,)))
spec2 = CompositeSpec(obs=Spec(shape=(4,)))
s = torch.stack([spec1, spec2], 0)
```
The code will surely break here and there when playing with collectors and similar but nothing that can't be fixed! We can work on that, I know @matteobettini is also looking forward to this feature.

- For varying numbers of Agents I'm not sure how to go about that. In the past, I worked in contexts where a simple mask could be used. You would have
```python
>>> env.step(tensordict)
>>> tensordict["next", "valid_mask"]
Tensor([True, False, True])
```
which would say that agent 0 and 2 are present, but 1 isn't. Then you would have
```python
>>> tensordict["next", "agent_feature"]
TensorDict({
    "observation": ...
}, batch_size[3])
```
which contains all the features on a per-agent basis, and you can recover the valid one with
```python
>>> tensordict["next", "agent_feature"][tensordict["next", "valid_mask"]]
```

matteobettini:
This is extremely exciting thank you!

yes, there is a lot of meat to discuss for multiagent.

- as @vmoens said, yes the idea for heterogeneous spaces is to hide heterogeneity using a lazy stack of tensors (a list). This is something high priority for us.

- For agents that do not participate in certain steps, we can use a mask as @vmoens said. Or we could also take a step more and just carry the tensors of the participating ones.

for example, let’s say we have an env with 5 agents. We would normally transport tensors (or lazy stacks) of size [*B, 5,*F]. Now, if at a certain step only 2 agents partake, we could carry only [*B,2,*F] and have a function in the env which tells me what those 2 correspond to. Wdyt?

- for agents being added and removed during execution, i don’t know how we would face this as torchrl specs are static.

hyerra:
Thanks for the feedback everyone, I'll look into this soon and get back to you

hyerra:
Hi @vmoens @matteobettini, thanks for the feedback and sorry for the delay! I've addressed some of the issues and I also have a question with the heterogeneous specs via stacking. Firstly though, to answer some of the questions.

1. Regarding testing, we could use some of the [example environments](https://github.com/Unity-Technologies/ml-agents/tree/develop/Project/Assets/ML-Agents/Examples) provided my MLAgents. We could do something like Unit Tests + integration tests using the example environments there. With regards to the description, would you like to see that in the documentation or would you prefer for me to describe this here?
2. For the stacking, I didn't realize that! I worked on implementing that, but ran into some issues. I'll describe them below.
3. For varying number of agents, I implemented what you said just now. Essentially, the environment kind of behaves like a parallel environment now. All of the agents observations will be recorded and sent to the user now rather than doing it one at a time, and we have a valid mask to accompany this so it is possible to tell which agent requested a decision. The only requirement now though is that all agents MUST request a decision at the first timestep so that the make_specs method can make the proper specs for all agents. After the first timestep though, agents do not need to request a decision and we apply a valid mask to filter which agents actually requested a decision. Also, with regards to @matteobettini's comment of carrying the tensors' forward, I think the problem with this is that you can only provide actions to agent's that requested a decision at that timestep and if you don't supply an action to an agent that requested a decision, [the default 0 action is used](https://github.com/Unity-Technologies/ml-agents/blob/develop/docs/Python-LLAPI.md). This can lead to unexpected behavior because maybe at timestep 1, Agent 1 requests a decision and at timestep 2 Agent 2 requests a decision. We wouldn't be able to supply Agent 1's action at timestep 2 as it didn't request a decision at that timestep. Also, Agent 1 would be given the 0 action which might not be desired.

With regards to the heterogeneous specs, I'm getting a really weird issue that I've been kind of stuck on. I'm not sure if you might have any insights. Essentially, in `make_specs` there's this piece of code:
```python
agent_observation_specs = [
        _unity_to_torchrl_spec_transform(
              spec, dtype=np.dtype("float32"), device=self.device
        )
        for spec in behavior_unity_spec.observation_specs
]
agent_observation_spec = torch.stack(agent_observation_specs, dim=0)
observation_specs[agent_id] = agent_observation_spec
```
Using this code produces the following error:
```
AttributeError: 'TransformedEnv' object has no attribute 'action_spec'. Did you mean: 'action_key'?
```

However, if I change the code slightly so that I include every observation besides the last one, it works. Namely, I need to change the code to be:
```python
agent_observation_specs = [
        _unity_to_torchrl_spec_transform(
              spec, dtype=np.dtype("float32"), device=self.device
        )
        for spec in behavior_unity_spec.observation_specs
]
agent_observation_spec = torch.stack(agent_observation_specs[:-1], dim=0)
observation_specs[agent_id] = agent_observation_spec
```
Notice, the [:-1] on the second to last line. Once this change is made, and I ignore the last observation, I don't get that error. Some important points to note are that the first 3 observations in my case have the same shape, however the last observation has a different shape. Here is a log of the observation specs:
```
[UnboundedContinuousTensorSpec(
    shape=torch.Size([128, 128, 3]),
    space=None,
    device=cpu,
    dtype=torch.float32,
    domain=continuous),
UnboundedContinuousTensorSpec(
    shape=torch.Size([128, 128, 3]),
    space=None,
    device=cpu,
    dtype=torch.float32,
    domain=continuous),
UnboundedContinuousTensorSpec(
    shape=torch.Size([128, 128, 3]),
    space=None,
    device=cpu,
    dtype=torch.float32,
    domain=continuous),
UnboundedContinuousTensorSpec(
    shape=torch.Size([4]),
    space=None,
    device=cpu,
    dtype=torch.float32,
    domain=continuous)]
```

I'm not sure how this relates to the action spec error above though and have been having some trouble debugging that.

hyerra:
@vmoens just wanted to follow up in case this might have gotten lost over the weekend. Also, I noticed in #1027, the VMAS environment is updated to use nested keys, so like Agent -> Observation, Agent -> Action. Should we do something similar here?

matteobettini:
Hey @hyerra, sorry for the delay (i am just starting here working with Vincent). I can help you w this. Feel free to message me also via email if you want to discuss more.

Basically, regarding your last comment: yes! we are moving to that format for multiagent. As you noticed, all the keys the have the agent dimension will be present in under the "agents" nested key. you can look at the vmas.py file in #1027 to see what I mean and uniform unity ML to that. If a key is shared among all agents (like sometimes done and reward are), this will be in the normal place and not under "agents".

With regards to your problem with stacking, here is where what is our view on stacking het tensors and specs:
Tensors or specs can be stacked under the same key (or lazystack) when they are semantically similar (ie. they typically come from similar sources and can be processed by the same neural network). 

For example, you can do

```
torch.stack(
  [UnboundedContinuousTensorSpec(
      shape=torch.Size([128, 128, 3]),
      space=None,
      device=cpu,
      dtype=torch.float32,
      domain=continuous),
  UnboundedContinuousTensorSpec(
      shape=torch.Size([64, 64, 3]),
      space=None,
      device=cpu,
      dtype=torch.float32,
      domain=continuous)]
)
```

because these (eg. images) are sematically similar.
In practice (in the code) we condsider thing to be sematically similar when they have the same number of dims and dtype.

In your case, the last spec is dramatically different from the first 3. The first 3 are more likely images and the last is just a small vector. Therefore you should stack them under different keys with something like

```
stack = torch.stack(
  [
         CompositeSpec(
              {"image":
                      UnboundedContinuousTensorSpec(
                            shape=torch.Size([128, 128, 3]),
                            space=None,
                            device=cpu,
                            dtype=torch.float32,
                            domain=continuous)
               }
         ),
         CompositeSpec(
              {"image":
                      UnboundedContinuousTensorSpec(
                            shape=torch.Size([128, 128, 3]),
                            space=None,
                            device=cpu,
                            dtype=torch.float32,
                            domain=continuous)
               }
         ),
         CompositeSpec(
              {"image":
                      UnboundedContinuousTensorSpec(
                            shape=torch.Size([128, 128, 3]),
                            space=None,
                            device=cpu,
                            dtype=torch.float32,
                            domain=continuous)
               }
         ),
        CompositeSpec(
              {"vec":
                      UnboundedContinuousTensorSpec(
                            shape=torch.Size([4]),
                            space=None,
                            device=cpu,
                            dtype=torch.float32,
                            domain=continuous)
               }
         ),
  ]
)
```

Then when you will do stack[2] you will retrieve an image spec and with stack[-1] you will retrieve a vec spec.

Does this make sense?

I am down to also schedule a brief call if you would like further help

hyerra:
Hi @matteobettini, no worries, thanks for the help! I think your feedback mostly makes sense, maybe I can try to fix it and if we are still running into difficulties we could hop on a call?

Just to make sure I understand, so individual agent observations go under an agents key and become nested. Things that are shared between all agents still remain at the top level.

Also, with regards to the second point about having different keys, I can't necessarily predetermine which index has different observations and know which ones will be similar. For my own environments I could; however, for everyone else they can customize the observations. So like in someone else's environment they might decide to have 4 images, some vector based observations, and maybe a Ray sensor ([MLAgents has a lot!](https://docs.unity3d.com/Packages/com.unity.ml-agents@1.0/api/Unity.MLAgents.Sensors.html)).

In this case would you recommend treating each observation as drastically different and using the keys "observation_1", "observation_2", etc.. and stacking them like that?

matteobettini:
> Hi @matteobettini, no worries, thanks for the help! I think your feedback makes sense, maybe I can try to fix it and if we are still running into difficulties we could hop on a call?

Sure



> Just to make sure I understand, so individual agent observations go under an agents key and become nested. Things that are shared between all agents still remain at the top level.

Yes, that is because tensors which are shared will have a dimension less, while the "agents" tensordict will have a batch_size with a dimension corresponding to the number of agents



> Also, with regards to the second point about having different keys, I can't necessarily predetermine which index has different observations and know which ones will be similar. For my own environments I could; however, for everyone else they can customize the observations. So like in someone else's environment they might decide to have 4 images, some vector based observations, and maybe a Ray sensor ([MLAgents has a lot!](https://docs.unity3d.com/Packages/com.unity.ml-agents@1.0/api/Unity.MLAgents.Sensors.html)).
> 
> In this case would you recommend treating each observation as drastically different and using the keys "observation_1", "observation_2", etc.. and stacking them like that?

Here are some options i can think of:
- One option would be to try and identify the sensor used and then assign it to that name. or use another identification methods and group under the same name obs with same n_dims and dtype (although I am not a super fan of this)
- I am wondering, does ml agent have non-named observations? i.e. sometimes observations come in a dict (this is possible in vmas) so if that is possible in ml agents you could carry those names forward.
- there is always the option of naming them observation_agent_i, as you say 

Form this options my take would be: see if there is any way to get the sensors names from the unity env and if there isn't we can either try to stack them if they are compatible or have a key per agent otherwise

@vmoens i would like to know your take on the (tl;dr) when obs can be heterogenous but we do not have access to their semantic differences

matteobettini:
@hyerra some FYI for you on timelines

For a full multiagent integration we need to support nested parametric keys and heterogeneous tensors across all the library components. These features will currently cause bugs here and there (you can find some issues and prs I opened about this).

I am currently working on the nested side of things and after that i will move on the heterogeneous part.

If you encounter issues on these things please open them in separate github issues and cc me

hyerra:
@matteobettini This makes sense, I made some progress on this yesterday for having nested keys but ran into some issues. I'll keep you posted on what I find!

Also, for similar/different observations, I think maybe option 3 is the best. So the way MLAgents work is there are different behaviors and all of the agents fall under a certain behavior. Each behavior defines the observation and action types/dimensions. For observations specifically, they are collected through "Sensors" in Unity like a Camera Sensor or a Ray Sensor for instance. In their [docs](https://github.com/Unity-Technologies/ml-agents/blob/develop/docs/Python-LLAPI-Documentation.md#decisionstep-objects), only a list of observation specs are returned (no name information) and each element of the list corresponds to an observation a sensor collected.

I think option 3 is the best for the following reasons:
- Yea I agree that option 1 might not be the ideal approach. If we get it wrong, users might get confused as to what's happening. Also, another thing that makes this approach more error-prone is users can actually stack observations temporally in Unity. So for instance I could stack the last two images that were captured which might be useful in some cases. If users did this and stacked 2 images together, the number of channels would grow from 3 to 6. This would lead us to maybe incorrectly classify an observation.
- For number 2, this would be ideal but I don't think we have access to the name information.

Since there's a lot of flexibility, maybe it's best to just do 3 since it's the safest option. If people really did want to stack things together and they knew it was safe, maybe we could offer some sort of Transform for it, but we could explore that later too.

Does this kind of make sense? I can work on having an implantation and sharing issues I find.

matteobettini:
> Also, for similar/different observations, I think maybe option 3 is the best. So the way MLAgents work is there are different behaviors and all of the agents fall under a certain behavior. Each behavior defines the observation and action space for all agents that fall under that behavior. For observations specifically, they are collected through "Sensors" in Unity like a Camera Sensor or a Ray Sensor for instance. In their [docs](https://github.com/Unity-Technologies/ml-agents/blob/develop/docs/Python-LLAPI-Documentation.md#decisionstep-objects), only a list of observation specs are returned (no name information) and each element of the list corresponds to an observation a sensor collected.

If there is a way to get the behavior names maybe that could also be an option, since behaviors I guess are aligned with neural networks.

In any case, I agree, if there is not a way that we can get observation semantics (names) in a principled manner, we are left only with option 3. We can maybe allow in the future for users to have a way to label observations to make things nicer.

Let's start to see if we can reach a working solution this way and then we can iterate.

On the nested side I am fixing a bunch of things so in a week or so that should work easily. Feel free to open any issues or problems you find on that.

Heterogenous stacking and passing to collectors will be a harder problem but I ll eventually get to that too

hyerra:
Sounds good! Also, I was wondering what the best way to encode individual agent observations is. Essentially, I have the observation spec correctly setup now so that it looks something like:

```
CompositeSpec("agents":
   CompositeSpec("observation":
       LazyStack(
          CompositeSpec("observation_1": ...),
          CompositeSpec("observation_2": ...),
       ),
      ....
   )
)
```

Given a list of numpy vectors representing an individual agent's observations:
```
["numpy_agent_1_observation_1", "numpy_agent_1_observation_2",]
```
How do I convert this into a tensordict for that individual agent? I think I have most of it successfully converted, but I'm having difficulty converting the observation part. Essentially, I want to have a tensordict for just that agent, similar to how you do it in VMAS, so something like:
```
agent_td = TensorDict(
     source={
          "agents": {
                "observation": self.read_obs(step.obs),
                 "behavior_name": self.read_behavior(behavior_name_),
                  "reward": self.read_reward(step.reward),
                   ...
            }
     }
)
```
Essentially, I'm not sure how to write the `read_obs` function to do the conversion.

matteobettini:
> Sounds good! Also, I was wondering what the best way to encode individual agent observations is. Essentially, I have the observation spec setup now so that it looks something like:

Mmh ok so you are stacking individual agent observations and also stacking the agents. Why do you need to stack each agent's observations? Couldn't it be simply a Tensordict? Cause otherwise this is getting a little bit convoluted even just to access. 



> How do I convert this into a tensordict for that individual agent? I think I have most of it successfully converted, but I'm having difficulty converting the observation part. Essentially, I want to have a tensordict for just that agent, similar to how you do it in VMAS, so something like:

Supposing that in your case each agent has an individual reward it should look like

```python
agent_i_td = TensorDict(
     source={
          "agents": {
                "observation": TensorDict({
                        "obs_1": Tensor,
                        "obs_2" Tensor,
                   })
                 "reward": Tensor,
            }
     }
```

The problem is that in your case if you have no guarantees that obs are similar (among agents and even within an agents), I don't know if stacking makes sense at all.

Another option (since we have completely 0 guarantees on obs similarity) could be to put all obs in the root td like so
```python
all_agents_td = TensorDict(
     source={
         "agent_1_observation_1": Tensor,
         "agent_1_observation_n": Tensor,
         ....,
         "agent_n_observation_n": Tensor,
         "done": Tensor
          "agents": {
                 "reward": Tensor,
            }, [n_agents,])
     }
```
I dunno what makes more sense.
cc @vmoens  if you have an opinion

hyerra:
> Mmh ok so you are stacking individual agent observations and also stacking the agents. Why do you need to stack each agent's observations? Couldn't it be simply a Tensordict? Cause otherwise this is getting a little bit convoluted even just to access.

Yep, currently I'm stacking the individual agent observations and then stacking the agents together. For the part about it being a Tensordict, I think I'm a little confused. How would the `observation_spec` look like for that?

Also, we could do the second approach as well. However, having everything grouped in the agents key seems nice. I can do either approach. @vmoens do you have a preference?

hyerra:
Hey @matteobettini, I think I'm pretty close to finishing implementation of the solution we discussed earlier, and I'll push my changes soon so we can both take a look and iterate.

However, one thing I noticed is that I think the `_EnvWrapper` is transforming the specs automatically. For instance, for the `done` spec (this also happens for the `reward` spec btw), I might assign it to something like this:
```
CompositeSpec(
    agents: CompositeSpec(
        done: DiscreteTensorSpec(
            shape=torch.Size([4, 1]),
            space=DiscreteBox(n=2),
            device=cpu,
            dtype=torch.bool,
            domain=discrete), device=cpu, shape=torch.Size([4])), device=cpu, shape=torch.Size([]))
```

However, when I do `print(self.done_spec)`, I will get:
```
DiscreteTensorSpec(
    shape=torch.Size([4, 1]),
    space=DiscreteBox(n=2),
    device=cpu,
    dtype=torch.bool,
    domain=discrete)
```

Do we want this behavior or should `_EnvWrapper` not remove the nesting? If we don't want this, I could raise an issue.

hyerra:
Just updated the PR @matteobettini if you want to take a look and let me know what your thoughts are from a high level. It works using a random policy on my end, but still of course we need to do some rigorous testing and have a way to test this in CI on a real environment. Just wanted to get your thoughts on the overall structure and high-level implementation.

matteobettini:
> However, when I do print(self.done_spec)

This is normal. it gives you the leaf spec to make it like single agent.
To retrieve the full nested spec you can do

`env.input_spec["_action_spec"]` (or reward, or done and so on)


> Just updated the PR @matteobettini if you want to take a look and let me know what your thoughts are from a high level. It works using a random policy on my end, but still of course we need to do some rigorous testing and have a way to test this in CI on a real environment. Just wanted to get your thoughts on the overall structure and high-level implementation.

Cool, I ll take a look monday

torchrl/envs/libs/unity.py
```diff
@@ -0,0 +1,421 @@
...
+                            )
+                            for i, spec in enumerate(
+                                behavior_unity_spec.observation_specs
+                            )
+                        }
+                    )
+                    behavior_id_specs[agent_id] = UnboundedDiscreteTensorSpec(
+                        shape=1, device=self.device, dtype=torch.int8
+                    )
+                    agent_id_specs[agent_id] = UnboundedDiscreteTensorSpec(
```
matteobettini, line 147: Can you elaborate a bit more on why we have an agent_is_specs and a behavior_id_spec.

Is this data gonna chage at each step?

torchrl/envs/libs/unity.py
```diff
@@ -0,0 +1,421 @@
...
+
+        for behavior_name, behavior_unity_spec in env.behavior_specs.items():
+            decision_steps, terminal_steps = env.get_steps(behavior_name)
+            for steps in [decision_steps, terminal_steps]:
+                for agent_id in steps.agent_id:
+                    self._agent_id_to_behavior[agent_id] = behavior_name
+
+                    observation_specs[agent_id] = CompositeSpec(
+                        {
+                            f"obs_{i}": _unity_to_torchrl_spec_transform(
```
matteobettini, line 136: this looks like the one we decided yes!

torchrl/envs/libs/unity.py
```diff
@@ -0,0 +1,421 @@
...
+                            )
+                        }
+                    )
+                    behavior_id_specs[agent_id] = UnboundedDiscreteTensorSpec(
+                        shape=1, device=self.device, dtype=torch.int8
+                    )
+                    agent_id_specs[agent_id] = UnboundedDiscreteTensorSpec(
+                        shape=1, device=self.device, dtype=torch.int8
+                    )
+                    action_specs[agent_id] = _unity_to_torchrl_spec_transform(
```
matteobettini, line 150: Why is this not composite like observation? 
I thought in unity agents could have composite action spaces

torchrl/envs/libs/unity.py
```diff
@@ -0,0 +1,421 @@
...
+                    agent_id = int(agent_id)
+                    step = steps[agent_id]
+                    done = False if i == 0 else True
+                    seen_agent_ids.add(agent_id)
+
+                    agent_td = TensorDict(
+                        source={
+                            "observation": self.read_obs(agent_id, step.obs),
+                            "behavior_id": self.read_behavior(agent_id, behavior_name_),
+                            "agent_id": self.read_agent_id(step.agent_id),
```
matteobettini, line 297: Do these change at every step?

matteobettini (commented):
It seems good to me!

See a few comments.

- Also can you add some tests so that we can have them along the way. See the vmas tests to see what kind of things to test.
- Can you give me a few instrucntions on how to instull unity? it seems it requires a paid license, is that so?

torchrl/envs/libs/unity.py
hyerra: No, it won't. However, I wasn't sure if it might be useful for users to have access to this information. I don't have a strong opinion on this, and we can remove it if we don't think it's useful.

The behavior_id might be more important especially since we don't have the information grouped by behaviors yet. (Option 3 in our discussion from earlier).

torchrl/envs/libs/unity.py
hyerra: Ah yep, so I kind of do this dynamically in the `_unity_to_torchrl_spec_transform`. They can have either:
1. continuous action(s) and discrete action(s)
2. continuous action(s)
3. discrete action(s)

In the function, I check to see which scenario it is. In case 1, I use a `CompositeSpec`, however in case 2 and 3 I use a `BoundedTensorSpec` or `DiscreteTensorSpec` respectively. Does this make sense?

torchrl/envs/libs/unity.py
hyerra: Nope, it's fixed.

hyerra:
Also, for your comment about tests, yep for sure, I'll work on this soon! And let me get back to you about the Unity license. Personally, I've just been using an educational/personal license.

torchrl/envs/libs/unity.py
matteobettini: we could have an attribute in the env object.

we do not need to put it in every tensordict or in the specs

torchrl/envs/libs/unity.py
matteobettini: see above

torchrl/envs/libs/unity.py
hyerra: Just updated this!

hyerra:
For the license, do you know if you could use the Personal license [here](https://unity.com/compare-plans)? If so, I think you should be able to download/install Unity from that. They also then have some games [here](https://github.com/Unity-Technologies/ml-agents/tree/develop-pytorch-2-upgrade/Project/Assets/ML-Agents/Examples) we could use for testing. Essentially, it would just involve choosing one of these games and building them I believe. We could then test our integration on the built game. I'll double check to make sure this works though. So far, I've just been testing this on my own custom game environments.

hyerra:
Hi @matteobettini, I just tested the environment out on the 3DBall and Strikers Vs. Goalie environments listed [here](https://github.com/Unity-Technologies/ml-agents/blob/develop/docs/Learning-Environment-Examples.md).

The good news is it works for the 3DBall environment since all of the agents have the same behavior (same observations/actions/etc.). However, it fails for the Strikers Vs. Goalie environment. The Strikers Vs. Goalie environment is more complex and the strikers/goalie have different observations. I think it is crashing because the number of observations between them is different. The striker has 2 sensors (therefore 2 observations) and the goalie only has 1 sensor (1 observation).

Is there a best way for me to share this with you to run? I made a few small modifications to the example environments, like removing their pre-trained networks. One option is maybe I could make a GitHub repo with the games that you can then use to build and run it.

Also, in terms of CI, would we be able to use like a docker instance and just load up the game binaries if we just built the game beforehand?

matteobettini:
Yes a repo with the examples works.

I am now going to try and fix the lazy stacks bugs and i can ping you when thats done and  we can iterate when i finish that and use unity as the final test.

regarding the CI @vmoens will know better what we can and cannot do but we are already working on seeing whats possible

hyerra:
Cool! I just uploaded a repo [here](https://github.com/hyerra/MLAgents-TorchRL-Examples) with the examples. In order to use them, you will want to open the cloned GitHub repo in UnityHub. Then, with the Editor open, you will want to navigate to Assets/Examples/3DBall/Scenes and then double click the 3DBall scene to open it. I attached a picture below:
Screenshot 2023-07-18 at 10 46 26 AM

Then, go to File -> Build Settings and click "Add Open Scenes". Then you can build for your preferred platform and save the game to a preferred location. Here's a picture of what this looks like for me:
Screenshot 2023-07-18 at 11 05 12 AM

You can repeat the same setup for the Strikers Vs. Goalie game. It is located under Assets/Examples/Soccer/Scenes and then double click StrikersVsGoalie.

For the TorchRL code, I've been testing it using a random policy:
```python
import torch
import socket

from contextlib import closing

from torchrl.collectors.collectors import RandomPolicy
from torchrl.envs.libs.unity import UnityEnv


def find_free_port():
    """Finds an open port on the system.

    Returns:
        The open port.
    """
    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
        s.bind(("", 0))
        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        return s.getsockname()[1]


def random_policy():
    """An implementation of a random policy."""
    device = "cpu" if not torch.has_cuda else "cuda:0"

    env = UnityEnv(
        "/Users/hyerra/Desktop/3DBalanceBall.app",
        log_folder="/Users/hyerra/Desktop/UnityLogs",
        # "<PATH TO UNITY GAME>",
        # log_folder="<PATH TO WHERE YOU WANT LOGS>",
        base_port=find_free_port(),
        timeout_wait=60*60*24,
        device=device,
        frame_skip=1,
    )
    policy = RandomPolicy(env.action_spec, action_key=env.action_key)
        env,
        policy,
        frames_per_batch=1,
        total_frames=10_000,
        device=device,
        split_trajs=False,
        exploration_mode="random",
    )

    for batch, data in enumerate(collector):
        pass


if __name__ == "__main__":
    random_policy()
```

Lmk if any questions!

hyerra:
@matteobettini I think I see why the StrikersVsGoalie game is failing now.

Essentially, with the way I coded this up, we are assuming that `obs_0`, `obs_1`, etc. will be the same between agents. However, this is not necessarily the case. For example, maybe `agent_0` has a camera sensor as `obs_0` and lidar as `obs_1` and `agent_1` has a generic vector sensor as `obs_0`. The previous implementation would crash since it was trying to stack tensordicts where `obs_0` had different dimensions between `agent_0` and `agent_1`.

However, if I rename the keys to have a format of `obs_{agent_id}_{obs number}`. Then it works just fine and the game runs properly. I will update the code here to incorporate this fix.

matteobettini:
I see, makes sense!

hyerra:
Sounds good! Also once everything looks good to you, I can work on adding some tests. I think we also have to figure out the best way to run the Unity games in CI

vmoens:
One option (defo not my preferred) is that for some backends we do not use the CI and just run them before the releases. If they run ok we package the wrappers in the release, if not we try to fix it and if we can't it will only be available in the nightlies / github.

hyerra:
There are a couple of options I found online that could work for building games:
1. [GameCI/CircleCI](https://circleci.com/blog/cicd-for-unity-projects/?utm_source=google&utm_medium=sem&utm_campaign=sem-google-dg--uscan-en-dsa-tROAS-auth-nb&utm_term=g_-_c__dsa_&utm_content=&gclid=Cj0KCQjwk96lBhDHARIsAEKO4xYlC462ofZs3B2JEU3zeCMF1lVbOrQZL3HW0WFmIDPzz8qZauY2-CcaAoxKEALw_wcB)
2. [Unity](https://docs.unity.com/devops/en/manual/start-using-devops)

In terms of using the built games, I'm not sure if something like Artifactory to store the games/Docker to run everything could work? This might be overcomplicated though. I'm happy to implement whatever works best.

torchrl/envs/libs/unity.py
```diff
@@ -0,0 +1,385 @@
...
+                    done_specs[agent_id] = DiscreteTensorSpec(
+                        n=2, shape=[1], dtype=torch.bool, device=self.device
+                    )
+                    valid_mask_specs[agent_id] = DiscreteTensorSpec(
+                        n=2, shape=[1], dtype=torch.bool, device=self.device
+                    )
+
+        self.observation_spec = CompositeSpec(
+            {
+                "agents": CompositeSpec(
```
matteobettini, line 159: What do you think about grouping agents by behaviour (as in case 2 of #1463)?

We would still have masks within behaviours for agent dropping in or out.

We could then allow users that do not want agents grouped by behaviour to change this by setting
`env = MLAgentsEnv(group_map={"group_name": [agent_names]}`

For example what you have now would be obtained by providing
`group_map={"agents",["agent_0",...,"agent_n]}`

but the default (whrn the map is not provided) would be groups by behavior

vmoens:
Hi folks
Thanks so much for this work. What's the status of the PR? Any way I can help?

matteobettini:
For the record, there are now more MARL environments in the library to take example from that use the MARL grouping in #1463 (e.g., PettingZoo)

We also have some utils for the grouping mechanism https://github.com/pytorch/rl/blob/fdee6336f271d074c770af8f2b101f7370d8d6c2/torchrl/envs/utils.py#L616 
https://github.com/pytorch/rl/blob/fdee6336f271d074c770af8f2b101f7370d8d6c2/torchrl/envs/utils.py#L691

I believe we could use that API to group agents here by behaviours.

vmoens:
Closed in favour of #2161

PR #1206, opened 2023-05-30
[Algo] Update offpolicy examples

BY571:
## Description

Updating the off-policy examples for ddpg, td3 and sac to simple scripts


examples/ddpg/ddpg.py
```diff
@@ -2,196 +2,166 @@
...
+        collected_frames += current_frames
+
+        # optimization steps
+        if collected_frames >= cfg.init_random_frames:
+            (
+                actor_losses,
+                q_losses,
+            ) = ([], [])
+            for _ in range(
+                int(cfg.env_per_collector * cfg.frames_per_batch * cfg.utd_ratio)
```
vmoens, line 99: we should remove all the `cfg.smth` from the training loop as this can slow down training (it is a known omegaconf issue).

What i usually do:

```python
env_per_collector = cfg. env_per_collector
frames_per_batch = cfg. frames_per_batch
utd_ratio = cfg. utd_ratio

...

            for _ in range(
                int(env_per_collector * frames_per_batch * utd_ratio)
            ):
```

examples/ddpg/utils.py
```diff
@@ -0,0 +1,223 @@
...
+    prb=False,
+    buffer_size=1000000,
+    buffer_scratch_dir="/tmp/",
+    device="cpu",
+    prefetch=3,
+):
+    if prb:
+        replay_buffer = TensorDictPrioritizedReplayBuffer(
+            alpha=0.7,
+            beta=0.5,
```
vmoens, line 94: we could make that parametrizable too

examples/ddpg/utils.py
```diff
@@ -0,0 +1,223 @@
...
+# -----
+
+
+def make_ddpg_agent(train_env, eval_env, device):
+    """Make DDPG agent."""
+    # Define Actor Network
+    in_keys = ["observation"]
+    action_spec = train_env.action_spec
+    actor_net_kwargs = {
+        "num_cells": [256, 256],
```
vmoens, line 129: do we want to let users control that?

examples/ddpg/utils.py
```diff
@@ -0,0 +1,223 @@
...
+        "out_features": action_spec.shape[-1],
+        "activation_class": nn.ReLU,
```
vmoens, line 131: ditto

examples/ddpg/utils.py
```diff
@@ -0,0 +1,223 @@
...
+        module=actor_module,
+        distribution_class=dist_class,
+        distribution_kwargs=dist_kwargs,
+        default_interaction_type=InteractionType.RANDOM,
+        return_log_prob=False,
+    )
+
+    # Define Critic Network
+    qvalue_net_kwargs = {
+        "num_cells": [256, 256],
```
vmoens, line 162: ditto

examples/sac/sac.py
```diff
@@ -2,193 +2,173 @@
...
+        tensordict = tensordict.view(-1)
+        current_frames = tensordict.numel()
+        replay_buffer.extend(tensordict.cpu())
+        collected_frames += current_frames
+
+        # optimization steps
+        if collected_frames >= cfg.init_random_frames:
+            (actor_losses, q_losses, alpha_losses) = ([], [], [])
+            for _ in range(
+                int(cfg.env_per_collector * cfg.frames_per_batch * cfg.utd_ratio)
```
vmoens, line 99: ditto regarding `cfg.`

examples/sac/sac.py
```diff
@@ -2,193 +2,173 @@
...
+                    "entropy": loss_td["entropy"],
+                }
+            )
+        for key, value in train_log.items():
+            logger.log_scalar(key, value, step=collected_frames)
+        if (
+            abs(collected_frames % cfg.eval_iter)
+            < cfg.frames_per_batch * cfg.frame_skip
+        ):
+            with set_exploration_type(ExplorationType.MEAN), torch.no_grad():
```
vmoens, line 153: MEAN is usually a bit slower than MODE
MODE just requires one sample but MEAN can require a lot since the average does not in general have a closed form (ie we take the empirical mean).
Not against it but be aware that this will be slower

examples/sac/utils.py
```diff
@@ -0,0 +1,225 @@
...
+def make_sac_agent(cfg, train_env, eval_env, device):
+    """Make SAC agent."""
+    # Define Actor Network
+    in_keys = ["observation"]
+    action_spec = train_env.action_spec
+    actor_net_kwargs = {
+        "num_cells": [256, 256],
+        "out_features": 2 * action_spec.shape[-1],
+        "activation_class": nn.ReLU,
+    }
```
vmoens, line 127: ditto

examples/td3/utils.py
```diff
@@ -0,0 +1,229 @@
...
+
+def make_td3_agent(train_env, eval_env, device):
+    """Make TD3 agent."""
+    # Define Actor Network
+    in_keys = ["observation"]
+    action_spec = train_env.action_spec
+    actor_net_kwargs = {
+        "num_cells": [256, 256],
+        "out_features": action_spec.shape[-1],
+        "activation_class": nn.ReLU,
```
vmoens, line 131: ditto

vmoens (commented):
LGTM
Would it make sense to have configs a bit more "structured" like
```
collector:
  frames_per_batch: 1000

replay_buffer:
  size: 1_000_000

```

then access the fields through
```
cfg.collector.frames_per_batch
```
?

Also, how deep do we want the config to be? Maybe we could let the users control the number of cells etc?

Don't we need to update the examples worflow in the CI?

vmoens (commented):
LGTM
Would it make sense to have configs a bit more "structured" like
```
collector:
  frames_per_batch: 1000

replay_buffer:
  size: 1_000_000

```

then access the fields through
```
cfg.collector.frames_per_batch
```
?

Also, how deep do we want the config to be? Maybe we could let the users control the number of cells etc?

Don't we need to update the examples worflow in the CI?

examples/td3/config.yaml
```diff
@@ -24,28 +23,21 @@ buffer_size: 1000000
 # Optimization
 utd_ratio: 1.0
 gamma: 0.99
-loss: double
-loss_function: smooth_l1
+loss_function: l2
 lr: 3e-4
 weight_decay: 0.0
```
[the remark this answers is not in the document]
BY571: ill try!

torchrl/objectives/sac.py
```diff
@@ -378,6 +378,7 @@ def _get_value_v2(self, tensordict, _alpha, actor_params, qval_params):
 
         """
         tensordict = tensordict.clone(False)
+        next_tensordict = step_mdp(tensordict)
```
vmoens, line 381: We're doing that below, why is it needed here?

BY571: Oh, I think that was a leftover when I merged main into the branch. Just fixed it :)

examples/ddpg/config.yaml
```diff
@@ -1,35 +1,46 @@
...
+  size: 1000000
+  prb: 0 # use prioritized experience replay
+
+# Optimization
+optimization:
+  utd_ratio: 1.0
+  gamma: 0.99
+  loss_function: smooth_l1
+  lr: 3e-4
+  weight_decay: 0.0
```
vmoens, line 33: let's add some weight decay unless performance is worse with it

examples/sac/config.yaml
```diff
@@ -1,35 +1,50 @@
...
+  size: 1000000
+  prb: 0 # use prioritized experience replay
+
+# Optimization
+optimization:
+  utd_ratio: 1.0
+  gamma: 0.99
+  loss_function: smooth_l1
+  lr: 3e-4
+  weight_decay: 0.0
```
vmoens, line 33: ditto

examples/sac/sac.py
```diff
@@ -2,193 +2,180 @@
...
+                optimizer_critic.step()
+                q_losses.append(q_loss.item())
+                optimizer_actor.zero_grad()
+                actor_loss.backward()
+                optimizer_actor.step()
+                actor_losses.append(actor_loss.item())
+                optimizer_alpha.zero_grad()
+                alpha_loss.backward()
+                optimizer_alpha.step()
+                alpha_losses.append(alpha_loss.item())
```
vmoens, line 131: In this case it makes more sense to add all the losses and do a single backprop.
Doing several can take more compute time

examples/td3/td3.py
```diff
@@ -2,302 +2,126 @@
...
 
                 loss_td = loss_module(sampled_tensordict)
 
                 actor_loss = loss_td["loss_actor"]
                 q_loss = loss_td["loss_qvalue"]
 
                 optimizer_critic.zero_grad()
-                update_actor = i % cfg.policy_update_delay == 0
+                update_actor = j % delayed_updates == 0
                 q_loss.backward(retain_graph=update_actor)
```
vmoens, line 125: Do we need the retain_graphs?
In theory we should not, unless there's a bug. They should be separate graphs.

vmoens (commented):
Almost there!

.circleci/unittest/linux_examples/scripts/run_test.sh
```diff
@@ -163,31 +184,8 @@ python .circleci/unittest/helpers/coverage_run_parallel.py examples/dqn/dqn.py \
...
-  batch_size=10 \
-  frames_per_batch=16 \
-  num_workers=2 \
-  env_per_collector=1 \
-  collector_device=cuda:0 \
-  optim_steps_per_batch=1 \
-  record_video=True \
-  record_frames=4 \
-  buffer_size=120
+  buffer_size=120 \
```
vmoens, line 187:
```suggestion
  replay_buffer.size=120 \
```

vmoens (approved):
Final review

PR #1208, opened 2023-05-30
[Feature, Refactor] Scale in ToTensorImage based on the dtype and new from_int parameter

hyerra:
## Description

In some instances, images might already be in the 0...1 range instead of 0...255 range. This is true in Unity, for example. This PR aims to add some flexibility by only scaling the image if the dtype = torch.uint8. This behavior is consistent with [torchvision's ToTensor transform](https://pytorch.org/vision/stable/generated/torchvision.transforms.ToTensor.html).

I also updated the documentation a little bit so that way the number of channels isn't hardcoded to 3. The code seems like it doesn't rely on the dimension being 3, so I thought that making it more general shouldn't hurt. For example, grayscale images will have 1 channel rather than 3.

## Motivation and Context

See description.

## Types of changes

## Checklist

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -4885,7 +4885,7 @@ class TestToTensorImage(TransformBase):
         assert (td.get("dont touch") == dont_touch).all()
 
         if len(keys) == 1:
-            observation_spec = BoundedTensorSpec(0, 255, (16, 16, 3))
+            observation_spec = BoundedTensorSpec(0, 255, (16, 16, 3), dtype=torch.uint8)
             observation_spec = totensorimage.transform_observation_spec(
                 observation_spec
             )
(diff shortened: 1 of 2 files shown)


hyerra:
Another option is we could determine whether or not scaling is required based on the data type similar to how `to_pil_image` is implemented in torchvision.

vmoens:
Thanks for this @hyerra 
I personally prefer the argument passed to the constructor in this case.

The reason is that someone may use int32 or another dtype for the images. We could leave it to the users to use the proper format but I'm afraid that this decision won't last long.
The argument in the constructor is unambiguous (even if less automatic).

It's a bit of a personal opinion. I'm not totally against looking at the dtype but maybe having both could work?

ie:
Case 1: uint8, from_int=True => div(255)
Case 2: any other type: from_int=True => div(255)
Case 3: uint8, from_int=None (default) => div(255)
Case 4: any other type, from_int=None (default) => no div
Case 5: uint8, from_int=False => no div
Case 6: any other type, from_int=v => no div

Wdyt?

Can we add this to the tests?

vmoens:
Sure is_floating_point + from_int seems to be the best.
It gives the user full control while also covering default reasonable behaviours!

hyerra:
Sounds good, just pushed and added a test!

vmoens (approved):
LGTM
Much better now!

PR #1215, opened 2023-06-01
[Feature] Dispatch for DDPG loss module

Blonck:
## Description

Enable dispatching arguments for the `.forward()` method of the DDPG loss module.

## Types of changes

## Checklist


Blonck:
## Open Questions
This pull request raises some open questions regarding the patch. Some of these questions may also apply to other loss modules.

1. Currently, the PR breaks backward compatibility by renaming the argument input_tensordict to tensordict. This renaming is necessary for @dispatch to function properly.
Potential solutions:
* Accept the change and standardize the argument name. It is unlikely that anyone is using keyword arguments to call .forward() (my preferred option).
* Enhance the functionality of `@dispatch` to allow alternative input names. This could involve adding a parameter that, when set, skips the name validation.

2. The underlying value estimator relies on optional input keys, such as `steps_to_next_obs`, which may be used by `.value_estimate()` if not `None`. The current DQNLoss implementation ignores these keys:
```python
class DQNLoss(LossModule):
     ...

    @dispatch(
        source=[
            "observation",
            ("next", "observation"),
            "action",
            ("next", "reward"),
            ("next", "done"),
        ],
        dest=["loss"],
    )
    def forward(self, tensordict: TensorDictBase) -> TensorDict:
        ...
```
Possible solutions include:
* Simply ignore optional arguments.
* Add optional tensor dict keys to the dispatch argument list, requiring users to provide the argument (which can be set to None).
* Extend `@dispatch` to support optional arguments.

3. Some tensor dict keys used by the value estimator, like `("next", "reward")` and `("next", "done")`, have fixed values despite the advantage module allowing configuration. These keys are also input keys for `.forward()`.
Should they be made configurable by adding them to `_AcceptedKeys` of the loss module in this PR? 

4. Sometimes input/output tensordict keys are dynamic and may depend on the configuration of the loss module. For DDPG loss this is not the case but for example the output keys of the A2CLoss depends on configuration.
One solution would be to use a @property for that. In case of the A2CLoss it would look like:
```python
    @property
    def out_keys(self):
        outs = ["loss_objective"]
        if self.critic_coef:
            outs.append("loss_critic")
        if self.entropy_bonus:
            outs.append("entropy")
            outs.append("loss_entropy")

        return outs
```
Do you see any problems with that approach?

Blonck:
>>> The underlying value estimator relies on optional input keys, such as steps_to_next_obs, which may be used by .value_estimate() if not None. The current DQNLoss implementation ignores these keys:

You can ignore this one. `@dispatch` does not allow any optional arguments, for now. So we need to go with the first solution and ignore optional arguments.

vmoens:
> ## Open Questions
> This pull request raises some open questions regarding the patch. Some of these questions may also apply to other loss modules.
> 
> 1. Currently, the PR breaks backward compatibility by renaming the argument input_tensordict to tensordict. This renaming is necessary for @dispatch to function properly.
>    Potential solutions:
> 
> * Accept the change and standardize the argument name. It is unlikely that anyone is using keyword arguments to call .forward() (my preferred option).
> * Enhance the functionality of `@dispatch` to allow alternative input names. This could involve adding a parameter that, when set, skips the name validation.
> 

I'm open to bc-breaking change in this case

vmoens:
> Simply ignore optional arguments.
> Add optional tensor dict keys to the dispatch argument list, requiring users to provide the argument (which can be set to None).
> Extend @dispatch to support optional arguments.

Two things here:
We can't assume that dispatch will have all the functionalities. I would keep things small scale if possible. 
If there is a key whose presence or absence controls the flow, I think it's best to consider that it is absent for dispatch.
That being said, in some places in the code we do
```python
value = data.get(key, None)
if value is None:
   foo()
else:
   bar()
```
In this case, we could have the user call
```
module(..., steps_to_next_obs=tensor)
```
which will populate the tensordict with that key, and if not the resulting value will be None.
That could require to change the function that reads `"steps_to_next_obs"`.

In summary:
I would suggest to keep the behaviour simple and not include `"steps_to_next_obs"` in the in_keys, but keep track of this in an issue in TorchRL.

vmoens:
> Sometimes input/output tensordict keys are dynamic and may depend on the configuration of the loss module. For DDPG loss this is not the case but for example the output keys of the A2CLoss depends on configuration.
> One solution would be to use a https://github.com/Property for that. In case of the A2CLoss it would look like:

Yes we should be using a property!

Blonck:
> Yes I think they should be part of the `_AcceptedKeys`. Do you have an example of a place where this?

For example ("next", "reward") or better ("next", self.tensor_keys.reward) is used in all advantages.

torchrl/objectives/ddpg.py
```diff
@@ -107,35 +193,56 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
...
+                reward=self._tensor_keys.reward,
+                done=self._tensor_keys.done,
             )
 
-    def forward(self, input_tensordict: TensorDictBase) -> TensorDict:
+    @property
+    def in_keys(self):
+        keys = [
+            ("next", self.tensor_keys.reward),
+            ("next", self.tensor_keys.done),
```
vmoens, line 204: this is a tricky one but the reward key can be nested and `("next", ("other", "key"))` will not work.
I usually check if the key is a str or tuple and act accordingly. We had a private discussion regarding having tensordict working with arbitrarily nested keys but did not reach a consensus yet...

test/test_cost.py
```diff
@@ -920,7 +937,11 @@ def test_ddpg_tensordict_keys(self, td_est):
             value,
             loss_function="l2",
         )
-        key_mapping = {"state_action_value": ("value", "state_action_value_test")}
+        key_mapping = {
+            "state_action_value": ("value", "state_action_value_test"),
+            "reward": ("reward", "reward_test"),
+            "done": ("done", "done_test"),
```
vmoens, line 943: Why not having reward_test and done_test nested? Like this we make sure that we cover the nested keys too.
eg:
```python
("test", "reward_test")
```

test/test_cost.py
```diff
@@ -933,11 +954,13 @@ def test_ddpg_tensordict_run(self, td_est):
         tensor_keys = {
             "state_action_value": "state_action_value_test",
             "priority": "td_error_test",
+            "reward": "reward_test",
+            "done": "done_test",
```
vmoens, line 958: ditto

test/test_cost.py
```diff
@@ -933,11 +954,13 @@ def test_ddpg_tensordict_run(self, td_est):
...
         }
 
         actor = self._create_mock_actor()
         value = self._create_mock_value(out_keys=[tensor_keys["state_action_value"]])
-        td = self._create_mock_data_ddpg()
+        td = self._create_mock_data_ddpg(reward_key="reward_test", done_key="done_test")
```
vmoens, line 963: ditto

vmoens (commented):
Fantastic, really fancy!
On a high level: do you think that the in_keys will be recyclable across modules or will we need to re-code it every time?

vmoens (commented):
Fantastic, really fancy!
On a high level: do you think that the in_keys will be recyclable across modules or will we need to re-code it every time?

torchrl/objectives/ddpg.py
Blonck: Ok, however I would need to adopt way the advantage module allows to configure the tensordict keys in the first place. Currently, the advantage module use some keys like this:
```python
tensordict.set(("next", self.tensor_keys.reward), reward)
```
which would fail if the user provides the reward key as tuple.

Blonck:
> On a high level: do you think that the in_keys will be recyclable across modules or will we need to re-code it every time?

No sure if there is really something common to all loss modules here. For me it seems to dependent on what actually happens inside the loss to recycle something. But I will think about it, maybe I've got some idea.

torchrl/objectives/ddpg.py
Blonck: Solved by unraveling keys in tensordict librarby (https://github.com/pytorch-labs/tensordict/commit/d2edd94bd7b03b508d096fa15d961064d704f638).

vmoens (approved):
Final review

PR #1218, opened 2023-06-01
[BugFix] Fix error message of .set_keys() in advantage modules

### LINKED ISSUE #1216: [BUG] Misleading error message in `advantage_module.set_keys()`
## Describe the bug

[`set_keys(value="new_value_key")`](https://github.com/pytorch/rl/blob/fa45472da29cc1d07b3ae57bdfe094207deeff13/torchrl/objectives/value/advantages.py#L218) expects the the "new_value_key" to be in the `out_keys` of `self.value_network`, but it raises an error complaining about the old key `self.tensor_keys.value`.

## Expected behavior

If it's the new key that should be in the value networks's output then the error message should complain about the new key, not the old key.
Otherwise, if `set_keys` changes the output key of the value network, then the error should be thrown when the old key in not in the value network's out keys.

## Screenshots

image

## Reason and Possible fixes

The feature was introduced very recently in #1175.

## Checklist

Blonck:
## Description

When `.set_keys(action="key_name")` is called with a key_name non-existing in the value_network.out_keys the right error message is generated.

close #1216

## Types of changes

## Checklist

### DIFF
--- a/torchrl/objectives/value/advantages.py
+++ b/torchrl/objectives/value/advantages.py
@@ -218,7 +218,7 @@ class ValueEstimatorBase(TensorDictModuleBase):
                 and (value not in self.value_network.out_keys)
             ):
                 raise KeyError(
-                    f"value key '{self.tensor_keys.value}' not found in value network out_keys"
+                    f"value key '{value}' not found in value network out_keys"
                 )
         if self._tensor_keys is None:
             conf = asdict(self.default_keys)
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -6028,6 +6028,10 @@ class TestAdv:
         module.set_keys(value=value)
         assert module.tensor_keys.value == value
 
+        with pytest.raises(KeyError) as excinfo:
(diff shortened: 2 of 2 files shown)


torchrl/objectives/value/advantages.py
```diff
@@ -218,7 +218,7 @@ def set_keys(self, **kwargs) -> None:
                 and (value not in self.value_network.out_keys)
             ):
                 raise KeyError(
-                    f"value key '{self.tensor_keys.value}' not found in value network out_keys"
+                    f"value key '{value}' not found in value network out_keys"
```
vmoens, line 221: We could also add the values that would be accepted.
I usually have customized error messages such that users don't have to dig into the code to see where the typo is.

```suggestion
                    f"value key '{value}' not found in value network out_keys {self.value_network.out_keys}"
```

vmoens (approved):
LGTM
A very minor comment, feel free to ignore

skandermoalla (approved):
Thanks!

PR #1222, opened 2023-06-01
[Refactor] Make TensorSpec a real class and TensorSpecBase a base class

vmoens:
cc @matteobettini any opinion?


matteobettini:
Do you have two lines on what we are doing here and why? I am not sure I get the full picture from the chagelist

matteobettini:
Is it just making `TensorSpec` a synonim of `UnboundedContinuousTensorSpec `?

vmoens:
> Is it just making `TensorSpec` a synonym of `UnboundedContinuousTensorSpec `?

and creating a base class
it's mildly bc-breaking but I often end up having to write `UnboundedContinuousTensorSpec` to just say "this is a tensor without precise properties"
Would this make sense? Any cons?

matteobettini:
I don’t have a strong opinion about this but two things that might be worth mentioning are:

- this seems quite a big semantic shift for current users
- the general TensorSpec term will assume quite a specific instantiation (taking a specific choice among the continuous and discrete options for a spec)

matteobettini:
Maybe another option is to change UnboundedContinuousTensorSpec to ContinuousTensorSpec and keep TensorSpec as the base

vmoens:
Yep the advantages are
- tensorspecbase is clearer and more aligned with general naming convensions
- it's mildly bc-breaking as you should not inherit from tensorspec anymore. Besides that, we only create a shortcut so for most users nothing will change. Only type annotation could be impacted.

> the general TensorSpec term will assume quite a specific instantiation

What do you mean?

matteobettini:
> What do you mean?

Just that it will be continuous implicitly without having it in the name

vmoens:
Not necessarily since you can change the dtype to int or book. It will be unbounded (within the range of the dtype) though

gtnv:
hi @vmoens, is there more work planned here? if not, can we close it?

PR #1225, opened 2023-06-02
[BugFix] Fix ppo example

vmoens:


torchrl/collectors/collectors.py
```diff
@@ -550,6 +550,7 @@ def __init__(
         else:
             self.policy_weights = TensorDict({}, [])
 
+        print("sending env to device", self.device)
```
skandermoalla, line 553: Maybe enclose this in a `if verbose:` like the other print statements?

examples/ppo/ppo.py
```diff
@@ -96,14 +96,27 @@ def main(cfg: "DictConfig"):  # noqa: F821
...
-        for _ in range(ppo_epochs):
-            for _ in range(frames_in_batch // mini_batch_size):
+        losses = TensorDict(
+            {}, batch_size=[ppo_epochs, -(frames_in_batch // -mini_batch_size)]
+        )
+        for j in range(ppo_epochs):
+            # Compute GAE
+            with torch.no_grad():
+                data = adv_module(data.to(model_device)).cpu()
+
```
skandermoalla, line 106: What's the motivation behind computing the advantages in the epoch loop?
I thought the original implementation of PPO computes then once per data collection step.
image
(https://arxiv.org/pdf/1707.06347.pdf)

skandermoalla: This is also something that I noticed in the PPO tutorial (https://pytorch.org/rl/tutorials/coding_ppo.html).

torchrl/collectors/collectors.py
vmoens: Oooops 😅

examples/ppo/ppo.py
vmoens: Let me have a look at this.
I remember seeing versions of PPO working with the advantage computed at the three stages of the optim (after collection, at each epoch or at each train step). I don't think there is any clear theoretical justification for one or the other but I might be wrong...
I'll investigate this carefully

vmoens: This is what I get with value computation for each batch (green) and for each epoch (red and pink).
The difference is quite striking. 
The only edit I did was to move the GAE computation before and after the epoch loop.

image

skandermoalla: Crazy! Which environment is this?

I guess then the question would be which implementation TorchRL wants to show in its examples/tutorials: the original one or one that works better?
In any case, I believe that just explicitly mentioning this choice of implementation detail would be enough to satisfy the curiosity/surprise of people like me.
(Mentioning some motivation behind it would help but wouldn't be necessary. These plots are cool but more extensive investigation would be needed to make an argument.)

> 
> I don't think there is any clear theoretical justification for one or the other but I might be wrong...
> I'll investigate this carefully

On the theory side, the objective is essentially based on the surrogate objective derived and analyzed in the TRPO paper ([see equation 3 and 14](https://arxiv.org/pdf/1502.05477.pdf)) where the data distribution and the advantages are both coming from the policy that collected the data.

One could say that PPO with advantages computed only at data collection would be the most similar objective, so the straightforward implementation would follow that (as the original one does).

But one could also argue that the theory doesn't extend to ClipPPO anyway (due to clipping), so you're free to play with other things in the objective as well.

And maybe there is some work that analyzes this (?).

skandermoalla: Also for reference, in the [The 37 Implementation Details of Proximal Policy Optimization](https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details/) blog post, advantages used in the actor loss are computed once per data collection step (line 264 in the first code snippet).

skandermoalla: @vmoens https://arxiv.org/abs/2006.05990 discusses recomputing the advantages in section 3.5 and finds that recomputing them at every epoch is better.
I'm not very familiar with their experimental setup though, I think they have data that lags behind the trained policy due to the SEED RL framework.

PR #1236, opened 2023-06-06
[BugFix] step_mdp keep_others=False in losses

vmoens:
cc @Blonck

### DIFF
--- a/torchrl/objectives/deprecated.py
+++ b/torchrl/objectives/deprecated.py
@@ -238,7 +238,7 @@ class REDQLoss_deprecated(LossModule):
         with torch.no_grad():
             selected_q_params = self.target_qvalue_network_params[selected_models_idx]
 
-            next_td = step_mdp(tensordict).select(
+            next_td = step_mdp(tensordict, keep_other=False).select(
                 *self.actor_network.in_keys
             )  # next_observation ->
             # observation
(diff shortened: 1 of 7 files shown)


vmoens:
All failing tests are from non-tensordict so this seems safe to land when the tests are fixed

Blonck:
The proposed changes could potentially result in unpredictable outcomes for torchrl users. Users might employ other TensorDict keys, expecting them to transition from ("next", "some_key") to "some_key" after one step. This anticipated behavior will subtly shift, which may go unnoticed. Users might not realize that the switch from ("next", "some_key") to "some_key" no longer occurs since both will still yield valid values, and there is no certainty that this alteration will cause any code to fail.

However, I'm unsure if this is a real problem or only a theoretical one.

vmoens:
I'm not sure: in which case should things fail?
The `keeps_other` option is there for when there is a context, non time-dependent tensor in the root tensordict that is being copied at every step.
There are two usages for losses:
- the data originates from a TorchRL environment. In that case we just need to make sure that the data in "next" is appropriate. Context tensors will be created via TensorDictPrimer and we should be safe
- the data is created by the users. In that case too, one would expect that "next" contains all the data necessary to be used as input to the model IMO.

Blonck:
You are right. I got the meaning of the keyword wrong. 

I was also wrong with the BC, the input tensordict will be cloned in all loss_module.forward() instances. So I see no problems with this PR.

gtnv:
hi @vmoens, are you planning to continue this? if not, should we close it for now?

PR #1238, opened 2023-06-07
[Feature] ReplayBuffer.empty

vmoens:
cc @tcbegley @apbard

### DIFF
--- a/torchrl/data/replay_buffers/samplers.py
+++ b/torchrl/data/replay_buffers/samplers.py
@@ -28,7 +28,7 @@ class Sampler(ABC):
 
     @abstractmethod
     def sample(self, storage: Storage, batch_size: int) -> Tuple[Any, dict]:
-        raise NotImplementedError
+        ...
 
     def add(self, index: int) -> None:
         return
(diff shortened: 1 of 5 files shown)


torchrl/data/replay_buffers/samplers.py
```diff
@@ -57,6 +59,10 @@ def ran_out(self) -> bool:
         # by default, samplers never run out
         return False
 
+    @abstractmethod
+    def _empty(self):
+        ...
```
tcbegley, line 64: `raise NotImplementedError` would be more consistent with `Sampler.sample`.

tcbegley: I noticed below you're actually favouring Ellipsis over NotImplementedError so perhaps we should change `sample` instead?

tcbegley (approved):
LGTM!

torchrl/data/replay_buffers/samplers.py
tcbegley: Sure, I'm happy with that.

PR #1244, opened 2023-06-08
[Feature] Dispatch SAC loss module

Blonck:
## Description

Enable dispatching arguments for the `.forward()` method of the SAC loss module.

## Types of changes

## Checklist

### DIFF
--- a/torchrl/objectives/sac.py
+++ b/torchrl/objectives/sac.py
@@ -147,7 +147,7 @@ class SACLoss(LossModule):
     This class is compatible with non-tensordict based modules too and can be
     used without recurring to any tensordict-related primitive. In this case,
     the expected keyword arguments are:
-    ``["action", "next_reward", "next_done"]`` + in_keys of the actor, value, and qvalue network
+    ``["action", "next_reward", "next_done"]`` + in_keys of the actor, value, and qvalue network.
     The return value is a tuple of tensors in the following order:
     ``["loss_actor", "loss_qvalue", "loss_alpha", "alpha", "entropy"]`` + ``"loss_value"`` if version one is used.
 
--- a/torchrl/objectives/sac.py
+++ b/torchrl/objectives/sac.py
@@ -10,7 +10,7 @@ from typing import Optional, Tuple, Union
 
 import numpy as np
 import torch
-from tensordict.nn import make_functional, TensorDictModule
+from tensordict.nn import dispatch, make_functional, TensorDictModule
 from tensordict.tensordict import TensorDict, TensorDictBase
 from tensordict.utils import NestedKey
 from torch import Tensor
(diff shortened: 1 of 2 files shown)


torchrl/objectives/sac.py
```diff
@@ -88,6 +88,112 @@ class SACLoss(LossModule):
...
+                loss_qvalue: Tensor(shape=torch.Size([]), device=cpu, dtype=torch.float32, is_shared=False),
+                loss_value: Tensor(shape=torch.Size([]), device=cpu, dtype=torch.float32, is_shared=False)},
+            batch_size=torch.Size([]),
+            device=None,
+            is_shared=False)
+
+    This class is compatible with non-tensordict based modules too and can be
+    used without recurring to any tensordict-related primitive. In this case,
+    the expected keyword arguments are:
+    ``["action", "next_reward", "next_done"]`` + in_keys of the actor, value, and qvalue network
```
vmoens, line 150:
```suggestion
    ``["action", "next_reward", "next_done"]`` + in_keys of the actor, value, and qvalue network.
```

torchrl/objectives/sac.py
```diff
@@ -88,6 +88,112 @@ class SACLoss(LossModule):
...
+        ...     in_keys=["observation"])
+        >>> loss = SACLoss(actor, qvalue, value)
+        >>> batch = [2, ]
+        >>> action = spec.rand(batch)
+        >>> loss_actor, loss_qvlaue, _, _, _, _ = loss(
+        ...     observation=torch.randn(*batch, n_obs),
+        ...     action=action,
+        ...     next_done=torch.zeros(*batch, 1, dtype=torch.bool),
+        ...     next_observation=torch.zeros(*batch, n_obs),
+        ...     next_reward=torch.randn(*batch, 1))
```
vmoens, line 194: For modules that inherit from TensorDIctModuleBase, you can call `module.select_out_keys("loss_actor", "loss_qvalue")` and you should only get those two values out

vmoens (approved):
Wonderful! Some very minor comments

torchrl/objectives/sac.py
Blonck: But this is the example for the @dispatch which returns a list.

vmoens: my point was merely that you could avoid the `_, _, _` by calling `select_out_keys` after constructing the loss

Blonck: Got it. But the loss modules are derived from LossModule which itself is derived from nn.Module, so no `select_out_keys`.

vmoens: Oh I thought we made it inherit from TDMBase, my mistake

PR #1256, opened 2023-06-12
[Feature] Single call to value network in advantages

vmoens:
This PR allows advantages to call once and only once the value model.
If `adv.shifted` is set to `True` and if the params at `t` and `t+1` match, the value net is called only once.
In all other cases, vmap is used to batch the calls to the value net.
 
cc @tcbegley @apbard

### DIFF
--- a/torchrl/objectives/value/advantages.py
+++ b/torchrl/objectives/value/advantages.py
@@ -21,6 +21,7 @@ from tensordict.tensordict import TensorDictBase
 from tensordict.utils import NestedKey
 from torch import nn, Tensor
 
+from torchrl._utils import RL_WARNINGS
 from torchrl.envs.utils import step_mdp
 
 from torchrl.objectives.utils import hold_out_net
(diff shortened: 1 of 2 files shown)


torchrl/objectives/value/advantages.py
```diff
@@ -965,34 +1094,55 @@ def forward(
...
+        #     # we assume that target parameters are not differentiable
+        #     kwargs["params"] = target_params
+        # elif "params" in kwargs:
+        #     kwargs["params"] = kwargs["params"].detach()
+        # if self.value_network is not None:
+        #     with hold_out_net(self.value_network):
+        #         # we may still need to pass gradient, but we don't want to assign grads to
+        #         # value net params
+        #         self.value_network(step_td, **kwargs)
+        # next_value = step_td.get(self.tensor_keys.value)
```
skandermoalla, line 1145: Is the commented-out code left on purpose?

PR #1267, opened 2023-06-13
[BugFix] Parametric `rand_action()` in `BaseEnv`

matteobettini:
`rand_action()` in common needs to be parametericto action key


vmoens (approved):
LGTM

PR #1270, opened 2023-06-13
[BugFix] Batched envs compatibility with custom keys

### LINKED ISSUE #1269: [BUG] VecEnv not compatible with parametric keys
VecEnvs have still paramteric "reward", "done", and "action" keys in their code, so they do not work with parametric keys

https://github.com/pytorch/rl/blob/main/torchrl/envs/vec_env.py#L338
for example

vmoens:
Fixes #1269 
To be merged after #1204 
Tests to be implemented


matteobettini:
We can close this

PR #1272, opened 2023-06-14
[Performance] Faster losses

vmoens:
Depends on https://github.com/pytorch-labs/tensordict/pull/418

vmap is now faster when the tensordict is locked as we can use a cached version of the batched tensordict.
This is especially useful when using lazy stacked tensordicts.

We cache several parameter structures in the losses to make sure that they're not recreated every time the loss is computed, which should improve runtime for most of them.

Anyone can help me with a review?

cc @matteobettini @BY571 @skandermoalla


torchrl/objectives/a2c.py
```diff
@@ -333,14 +334,19 @@ def loss_critic(self, tensordict: TensorDictBase) -> torch.Tensor:
             )
         return self.critic_coef * loss_value
 
+    @property
+    @cache_values
+    def __detach_critic_params(self):
+        return self.critic_params.detach()
+
```
matteobettini, line 341: Here self.critic_params is already using cache so we are only caching the detach operation right?

torchrl/objectives/ddpg.py
```diff
@@ -357,3 +346,23 @@ def make_value_estimator(self, value_type: ValueEstimators = None, **hyperparams
...
+            {
+                "module": {
+                    "0": self.target_actor_network_params,
+                    "1": self.target_value_network_params,
+                }
+            },
+            batch_size=self.target_actor_network_params.batch_size,
+            device=self.target_actor_network_params.device,
+        )
+        return target_params
```
matteobettini, line 363: Same here there is not much benefit in chaching since all components are cached right?

torchrl/objectives/iql.py
```diff
@@ -423,7 +420,7 @@ def _loss_qvalue(self, tensordict: TensorDictBase) -> Tuple[Tensor, Tensor]:
         target_value = self.value_estimator.value_estimate(
             tensordict, target_params=self.target_value_network_params
         ).squeeze(-1)
-        tensordict_expand = vmap(self.qvalue_network, (None, 0))(
+        tensordict_expand = self._vmap_qvalue_networkN0(
```
matteobettini, line 423: what are these N0 or 00 standing for? not super intuitive for me

torchrl/objectives/td3.py
```diff
@@ -313,16 +316,24 @@ def in_keys(self):
...
+    @cache_values
+    def __detach_qvalue_network_params(self):
+        return self.qvalue_network_params.detach()
+
+    @property
+    @cache_values
+    def __stack_actor_params(self):
+        return torch.stack(
+            [self.actor_network_params, self.target_actor_network_params], 0
+        )
```
matteobettini, line 329: In a function like this, what is the benefit of caching? 
Is it just not to redo the stack operation or am I missing smth?

matteobettini (commented):
Looks good. Just trying to understand a bit

torchrl/objectives/a2c.py
vmoens: yep

vmoens: detaching can take quite some time bc we need to rebuild a hugely nested tensordict
caching is virtually memory free and alleviates that

torchrl/objectives/ddpg.py
vmoens: same as before: TD construction is a bottleneck

torchrl/objectives/iql.py
vmoens: None, 0 and 0, 0
because we have 2 different vmaps
other names are welcome! `none_0`?

torchrl/objectives/td3.py
vmoens: Same as before + lazy stacked construction is reaaaally long so we must cache it if we can (have a look at tensordict benchmarks and introduciton of caches https://pytorch-labs.github.io/tensordict/dev/bench/)

Issue #1274, opened 2023-06-14
[BUG] Value keys missing from DQNLoss _AcceptedKeys


matteobettini:
In the DQN loss accepted keys https://github.com/pytorch/rl/blob/main/torchrl/objectives/dqn.py#L115
the 
```
  tensor_keys = {
            "advantage": "advantage",
            "value_target": "value_target",
            "value": "chosen_action_value",
        }
```
are missing.
These are set in https://github.com/pytorch/rl/blob/main/torchrl/objectives/dqn.py#L220 but the user cannot change them.

This is not a case for other losses like ppo 
https://github.com/pytorch/rl/blob/main/torchrl/objectives/ppo.py#L213

Maybe we could sweep all losses for this

cc @Blonck

PR #1285, opened 2023-06-15
[BugFix] DQN loss dispatch respect configured tensordict keys

Blonck:
## Description

Dispatch of `.forward` of the DQN loss module respects configured tensordict keys.

Solves #1274 

## Types of changes

## Checklist

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -480,19 +480,45 @@ class TestDQN(LossModuleTestBase):
             p.data += torch.randn_like(p)
         assert all((p1 != p2).all() for p1, p2 in zip(parameters, actor.parameters()))
 
-    def test_dqn_tensordict_keys(self):
+    @pytest.mark.parametrize(
+        "td_est", [ValueEstimators.TD1, ValueEstimators.TD0, ValueEstimators.TDLambda]
+    )
+    def test_dqn_tensordict_keys(self, td_est):
         torch.manual_seed(self.seed)
         action_spec_type = "one_hot"
         actor = self._create_mock_actor(action_spec_type=action_spec_type)
         loss_fn = DQNLoss(actor)
 
         default_keys = {
+            "advantage": "advantage",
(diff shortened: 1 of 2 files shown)


test/test_cost.py
```diff
@@ -577,37 +603,38 @@ def test_distributional_dqn(
             p.data += torch.randn_like(p)
         assert all((p1 != p2).all() for p1, p2 in zip(parameters, actor.parameters()))
 
-    def test_dqn_notensordict(self):
+    @pytest.mark.parametrize("observation_key", ["observation", "observation2"])
+    @pytest.mark.parametrize("reward_key", ["reward", "reward2"])
+    @pytest.mark.parametrize("done_key", ["done", "done2"])
```
vmoens, line 608: do we skip the action key by design?

vmoens (approved):
LGTM thanks for this
Do we want to cover the action key too in the tests?

test/test_cost.py
Blonck: No it is intended.
Atm, the action key cannot really be configured since it is used in the constructor of the DQN loss via `_find_action_space(action_space)`. 

Either I need to remove the action key from the configurable keys or the following part of the constructor must be moved until `.set_keys()` is called:
```
        if action_space is None:
            # infer from value net
            try:
                action_space = value_network.spec
            except AttributeError:
                # let's try with action_space then
                try:
                    action_space = value_network.action_space
                except AttributeError:
                    raise ValueError(self.ACTION_SPEC_ERROR)
        if action_space is None:
            warnings.warn(
                "action_space was not specified. DQNLoss will default to 'one-hot'."
                "This behaviour will be deprecated soon and a space will have to be passed."
                "Check the DQNLoss documentation to see how to pass the action space. "
            )
            action_space = "one-hot"
        self.action_space = _find_action_space(action_space)
```

vmoens: Got it thanks!

PR #1289, opened 2023-06-15
Linux Olddeps tests to Nova

osalpekar:
Passing job: https://github.com/pytorch/rl/actions/runs/5293966748/jobs/9582760484?pr=1289


.circleci/unittest/linux_olddeps/scripts_gym_0_13/install.sh
```diff
@@ -37,9 +37,9 @@ git submodule sync && git submodule update --init --recursive
 
 printf "Installing PyTorch with %s\n" "${CU_VERSION}"
 if [ "${CU_VERSION:-}" == cpu ] ; then
-    conda install pytorch==1.13.1 torchvision==0.14.1 torchaudio==0.13.1 cpuonly -c pytorch
+    conda install pytorch==2.0.1 torchvision torchaudio cpuonly -c pytorch
```
osalpekar, line 40: Upgraded to using pytorch 2.0.1 since some functions depended on in this job/test used `torch.vmap()`, which is not available in pytorch 1.13.1

vmoens: no we need 1.13 because that's what polymetis is built on :/ 
https://github.com/facebookresearch/fairo/blob/main/polymetis/polymetis/environment.yml

vmoens: We must be compatible with an env where polymetis is installed to have torchrl work on hardware

vmoens, line 40:
```suggestion
    conda install pytorch==1.13.1 torchvision==0.14.1 torchaudio==0.13.1 cpuonly -c pytorch
```

.circleci/unittest/linux_olddeps/scripts_gym_0_13/install.sh
```diff
@@ -37,9 +37,9 @@ git submodule sync && git submodule update --init --recursive
...
 else
-    conda install pytorch==1.13.1 torchvision==0.14.1 torchaudio==0.13.1 pytorch-cuda=11.6 -c pytorch -c nvidia -y
+    conda install pytorch==2.0.1 torchvision torchaudio pytorch-cuda=11.6 -c pytorch -c nvidia -y
```
vmoens, line 42:
```suggestion
    conda install pytorch==1.13.1 torchvision==0.14.1 torchaudio==0.13.1 pytorch-cuda=11.6 -c pytorch -c nvidia -y
```

vmoens:
There is one test that fails. It's a CUDA sync issue.
I solved it on CI by playing with the cuda versions.
Oddly, they seem to match between CCI and GHA so I'm not sure why they fail on one and not the other...

vmoens (approved):

PR #1295, opened 2023-06-19
[BugFix] Fix the bug of `RoundRobinWriter.extend(data)` 

xmaples:
## Motivation and Context

Fix the bug of `RoundRobinWriter.extend(data)` with data.batch_size > twice of storage.max_size.

bug reproduction:
```python
# bug: RoundRobinWriter
from torchrl.data.replay_buffers import TensorDictReplayBuffer,LazyTensorStorage
import torch
td=TensorDict({"a": torch.rand(10,2)}, batch_size=[10])
rb=TensorDictReplayBuffer(storage=LazyTensorStorage(4))
rb.extend(td)   # IndexError: index 4 is out of bounds for dimension 0 with size 4
```

### DIFF
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -187,6 +187,8 @@ class TestComposableBuffers:
                     break
             else:
                 raise RuntimeError("did not find match")
+        data2 = self._get_data(rb_type, size=2 * size + 2)
+        rb.extend(data2)
 
     def test_sample(self, rb_type, sampler, writer, storage, size):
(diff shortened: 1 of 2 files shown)


vmoens (commented):
Great stuff thanks for the fix!
The tests are now failing bc the non-tensordict writer hasn't been changed. I made a PR against your branch with the fix (since you were working on main I could not edit it directly on your branch).

xmaples:
ok, i've accepted.

PR #1309, opened 2023-06-23
[Feature] RLHF dataloading

vmoens:
cc @tcbegley @apbard 

There are 3 TODO i need help with


torchrl/data/rlhf/dataset.py
```diff
@@ -0,0 +1,348 @@
...
+    dataset_name,
+    pre_tokenization_hook=None,
+    from_disk: bool = False,
+):
+    """Loads a text dataset from ``datasets``.
+
+    Args:
+        split (str): One of ``"train"`` or ``"valid"``.
+        dataset_name (str): the name or path of the dataset.
+        pre_tokenization_hook (callable): TODO
```
tcbegley, line 116: `pre_tokenization_hook` is called on the Dataset before tokenization and should return a modified Dataset object. The intended use is for carrying out tasks that require modifying the dataset as a whole as opposed to modifying individual datapoints, for example discarding certain datapoints based on a particular condition. Tokenization and other "elementwise" operations on the data are performed by the process function which is mapped over the dataset.

torchrl/data/rlhf/dataset.py
```diff
@@ -0,0 +1,348 @@
...
+    from_disk=False,
+):
+    """Loads a pre-processed, memory-mapped dataset if it exists, and creates it otherwise.
+
+    Args:
+        split (str): One of ``"train"`` or ``"valid"``.
+        max_length (int): the maximum sequence length.
+        dataset_name (str): the name of the dataset.
+        make_process_fn (callable): a preprocess function.
+        pre_tokenization_hook (callable): TODO
```
tcbegley, line 38: See comment below, this is passed on to the `load_dataset` function.

test/datasets_mini/openai_summarize_comparisons/test/data-00000-of-00001.arrow
tcbegley, line 1: Do we want to commit all of these arrow files etc. to the repo?

vmoens: It's a dummy dataset. I agree it isn't perfect but I'm not sure what choice we have. could we zip it?

vmoens: Vision has something of similar taste
https://github.com/pytorch/vision/tree/main/test/assets
I can put it all in a .pt

test/assets/generate.py
```diff
@@ -0,0 +1,49 @@
...
+"""Script used to generate the mini datasets."""
+from tempfile import TemporaryDirectory
+
+from datasets import Dataset, DatasetDict, load_dataset
+
+from torchrl.data.rlhf.dataset import get_dataloader
+from torchrl.data.rlhf.prompt import PromptData
+
+
+def generate_small_data(comparison=True):
```
apbard, line 15:
```suggestion
def generate_small_dataset(comparison=True):
```

torchrl/data/rlhf/dataset.py
```diff
@@ -0,0 +1,467 @@
...
+        max_length=max_length,
+        tokenizer_fn=tokenizer_fn,
+    )
+    prefix = (split, str(max_length))
+    return dataset_to_tensordict(
+        dataset, data_dir=data_dir, prefix=prefix, valid_mask_key="valid_sample"
+    )[prefix]
+
+
+def load_dataset(
```
apbard, line 119: this is only used internally? should we put a "_" in front?
```suggestion
def _load_dataset(
```

torchrl/data/rlhf/dataset.py
```diff
@@ -0,0 +1,467 @@
...
+
+from tensordict.tensordict import NestedKey
+from torchrl.data import TensorDictReplayBuffer, TensorStorage
+from torchrl.data.replay_buffers import SamplerWithoutReplacement
+from transformers import AutoTokenizer
+
+_has_datasets = importlib.util.find_spec("datasets") is not None
+
+
+def create_or_load_dataset(
```
apbard, line 25: what about calling this just "load_dataset" and maybe add an argument such as "exists" in mkdir?

apbard (approved):
in general I noticed that everything is assuming we use gpt2 models but we are not saying explicitly. E.g. tokenize function

LGTM

test/test_rlhf.py
```diff
@@ -0,0 +1,299 @@
...
+    dataset_path,
+    make_process_fn,
+    pre_tokenization_hook,
+    mocker,
+):
+    # test caching of the values
+    lmemmap_save = deepcopy(TensorDict.load_memmap)
+    mocked_hello = mocker.patch("tensordict.TensorDict.load_memmap")
+    mocked_hello.side_effect = lmemmap_save
+    with zipfile.ZipFile(dataset_path, "r") as zip_ref:
```
tcbegley, line 71: Instead of doing this in every test that loads the data, perhaps we should write a `conftest.py` file that extracts on setup, then cleans up on completion?

test/assets/generate.py
tcbegley, line 1: Should we add logic here to zip the generated data so that the committed `.zip` files are reproducible?

tcbegley (approved):
LGTM, a couple of comments on the generated data / zipping + unzipping it.

torchrl/data/rlhf/reward.py
```diff
@@ -0,0 +1,206 @@
...
+            rejected_data=RewardData(
+                **rejected_data,
+                batch_size=batch_size,
+            ),
+            batch_size=batch_size,
+        )
+
+
+def pre_tokenization_hook(dataset, min_length=5):
+    """Pre-tokenizer for the eward model (comparison) dataset.
```
tcbegley, line 145:
```suggestion
    """Pre-tokenizer for the reward model (comparison) dataset.
```

torchrl/data/rlhf/dataset.py
```diff
@@ -0,0 +1,467 @@
...
+            function which is mapped over the dataset.
+        root_dir (path, optional): the path where the datasets are stored.
+            Defaults to ``"$HOME/.cache/torchrl/data"``
+        from_disk (bool, optional): if ``True``, :func:`datasets.load_from_disk`
+            will be used. Otherwise, :func:`datasets.load_dataset` will be used.
+            Defaults to ``False``.
+
+    The dataset will be stored in ``root/<split>/<max_length>/``
+    Examples:
+        >>> from torchrl.data.rlhf import TensorDictTokenizer        >>> from torchrl.data.rlhf.reward import  pre_tokenization_hook
```
tcbegley, line 61:
```suggestion
        >>> from torchrl.data.rlhf import TensorDictTokenizer        
        >>> from torchrl.data.rlhf.reward import  pre_tokenization_hook
```

vmoens:
> in general I noticed that everything is assuming we use gpt2 models but we are not saying explicitly. E.g. tokenize function

Would the tokenization change with another model?
Any pointer to a place where that would break?

torchrl/data/rlhf/dataset.py
vmoens: Is it? IDK if we want to expose it or not. To me it's a well self-contained function properly documented that could be used somewhere else, but I'm happy to make it private if there's no use.

test/test_rlhf.py
vmoens: Good point yeah
But then how do we tell conftest to do it only when the test_rlhf.py script is executed?

test/assets/generate.py
vmoens: I did not use python for that but we could do it

test/test_rlhf.py
tcbegley: > But then how do we tell conftest to do it only when the test_rlhf.py script is executed?

How about instead of `conftest.py` we add a fixture that has session scope and yields a directory which is the location of the extracted data?

torchrl/data/rlhf/dataset.py
apbard: My suggestion/discussion is whether we could rename this to load_dataset without changing it's current behaviour and add an argument to specify the behaviour we want in case it does not exists.

test/test_rlhf.py
vmoens: I think fixtures are executed in each test, otherwise the same temporary directories would be reused multiple times

torchrl/data/rlhf/dataset.py
apbard: I think in the common case user will want either to load or create. Hence my proposal to keep this internal and expose only the maybe-renamed other function

apbard:
> > in general I noticed that everything is assuming we use gpt2 models but we are not saying explicitly. E.g. tokenize function
> 
> Would the tokenization change with another model? Any pointer to a place where that would break?

If the tokenizer is huggingface-like I think the tokenize function should work provided that we somehow expose: https://github.com/pytorch/rl/pull/1309/files#diff-a81d96807adcce02fb18ac959f3a6b327225882e56438b51ebc4f7da350b8493R193

torchrl/data/rlhf/dataset.py
```diff
@@ -0,0 +1,472 @@
...
+    """
+    if not _has_transformers:
+        raise ImportError("The transformers library is missing.")
+    from transformers import AutoTokenizer
+
+    if num_workers is None:
+        num_workers = max(os.cpu_count() // 2, 1)
+    if excluded_features is None:
+        excluded_features = {"text", "prompt", "label", "valid_sample"}
+    tokenizer = AutoTokenizer.from_pretrained("gpt2")
```
apbard, line 193: this

torchrl/data/rlhf/dataset.py
vmoens: Im fine with renaming it, though I think that users should be informed that this function caches the results and the name says that quite explicitly.
I'm not sure I see what the extra arg should do. If it does not exist we must load it, and if it does we should use the cache. What extra control should we give?

apbard: just a flag like exists_ok of mkdir that raises if it does not exists or something similar to `from_disk` but control whether we want to store cache or not.
I don't have strong opinion on this I am just reasoning if it's really necessary such a long name (even if definitely more explicit) or if it will be enough to say that we cache in the doc.

test/test_rlhf.py
tcbegley: Yes, that's the default, but you can change that behaviour

```python
@pytest.fixture(scope="session")
def data_dir():
    # unzip
    # yield path
```

torchrl/data/rlhf/dataset.py
vmoens: got it
Let's parametrize it then

torchrl/data/rlhf/dataset.py
```diff
@@ -0,0 +1,477 @@
...
+
+            tokenizer_class = AutoTokenizer
+        if tokenizer_model_name is None:
+            tokenizer_model_name = "gpt2"
+        self.make_tokenizer(
+            tokenizer_class=AutoTokenizer, tokenizer_model_name=tokenizer_model_name
+        )
+
+    def make_tokenizer(self, *, tokenizer_class, tokenizer_model_name):
+        tokenizer = tokenizer_class.from_pretrained("gpt2")
```
apbard, line 123:
```suggestion
        tokenizer = tokenizer_class.from_pretrained(tokenizer_model_name)
```

apbard (approved):
LGTM

test/test_rlhf.py
vmoens: Oh now i remember why i did not do that
we have 2 mini datasets but you can't pass a fixture to parametrize, so I couldn't write 2 different fixtures for the tests
The "ugly" solution would be

```
@pytest.mark.parametrize("dataset", ["comp", "prompt"])
def test_stuff(dataset, comp_fixture, prompt_fixture):
    if dataset == "comp":
        dataset = comp_fixture
    else:
        dataset = prompt_fixture
```

PR #1315, opened 2023-06-26
[Feature, NOMERGE] RLHF Reward Model

tcbegley:
**Do not merge before #1309**

This PR builds on top of #1309, adding the `GPT2RewardModel` class and tests. Changes from that PR are needed so that we can use the dataloaders in the tests.


torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,94 @@
...
+            model = GPT2LMHeadModel.from_pretrained(model_path, return_dict=False)
+        else:
+            model = GPT2LMHeadModel(GPT2LMHeadModel.config_class())
+
+        self.config = model.config
+        self.transformer = model.transformer
+
+        # replace last layer with the reward layer
+        self.lm_head = nn.Linear(self.config.n_embd, 1, bias=False)
+        self.PAD_ID = GPT2TokenizerFast.from_pretrained("gpt2").eos_token_id
```
apbard, line 29: shouldn't this be parametrized? the different gpt2 flavours have different tokenizer
```suggestion
        self.PAD_ID = GPT2TokenizerFast.from_pretrained(model.name_or_path).eos_token_id
```
Note: this means we should probably also store the tokenizer?

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def forward(self, input_ids=None, attention_mask=None):
+        outputs = self.transformer(input_ids=input_ids, attention_mask=attention_mask)
```
apbard, line 32:
```suggestion
        """ Returns a tuple (rewards, end_scores) where `rewards` contains all rewards computed at each timestep, `end_scores` contains the reward computed at the last-non-padding token
        """
        outputs = self.transformer(input_ids=input_ids, attention_mask=attention_mask)
```

apbard (commented):
shouldn't we rebase on #1309 to make sure tests are green?

torchrl/modules/models/rlhf.py
tcbegley: > the different gpt2 flavours have different tokenizer

They do? I was under the impression it was a single encoding scheme for all variants and it was the number of parameters that changes?

apbard: double checked. the comparison fails but the underlying data looks like to be the same.

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,96 @@
...
+        return rewards, torch.stack(end_scores)
+
+    @staticmethod
+    def compute_reward_loss(chosen_batch, rejected_batch, pad_token_id=50256):
+        """Compute the reward loss given a chosen and rejected batch.
+
+        The loss is computed as loss = -log_sigmoid(chosen_reward - rejected_reward).
+        This loss is small when the reward model favours the chosen data and large if
+        the model favours the rejected data.
+        """
```
apbard, line 56:
```suggestion
        Note: the loss is computed excluding the common "prefix" subsequence to effectively disregard contribution of the original prompt.
        """
```

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,96 @@
...
+        bs = chosen_rewards.shape[0]
+        loss = 0
+
+        for i in range(bs):
+            # Check if there is any padding otherwise take length of sequence
+            c_inds = (chosen_ids[i] == pad_token_id).nonzero()
+            c_ind = c_inds[0].item() if len(c_inds) > 0 else chosen_ids.shape[1]
+            r_inds = (rejected_ids[i] == pad_token_id).nonzero()
+            r_ind = r_inds[0].item() if len(r_inds) > 0 else rejected_ids.shape[1]
+            end_ind = max(c_ind, r_ind)
```
apbard, line 71: have we checked the reward is effectively carried over in presence of padding?

apbard (approved):
left couple of comments. but LGTM! thanks!

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,97 @@
...
+            model = GPT2LMHeadModel.from_pretrained(model_path, return_dict=False)
+        else:
+            model = GPT2LMHeadModel(GPT2LMHeadModel.config_class())
+
+        self.config = model.config
+        self.transformer = model.transformer
+
+        # replace last layer with the reward layer
+        self.lm_head = nn.Linear(self.config.n_embd, 1, bias=False)
+        self.PAD_ID = GPT2TokenizerFast.from_pretrained("gpt2").eos_token_id
```
vmoens, line 29: remove caps

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class GPT2RewardModel(nn.Module):
+    """Wrapper around GPT2-like models to enable their use as reward models.
+
+    This wrapper replaces the language modelling head of the GPT2 model with a new
+    linear layer with 1 output that can be used as a reward signal. It also exposes the
+    method ``compute_reward_loss`` which calculates the reward loss by comparing two
+    batches, one chosen, one rejected.
+    """
```
vmoens, line 15: Can we add an example of this class?

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,97 @@
...
+        rewards = self.lm_head(hidden_states).squeeze(-1)
+        end_scores = []
+        bs = input_ids.shape[0]
+
+        for i in range(bs):
+            pad_inds = (input_ids[i] == self.PAD_ID).nonzero()
+            first_pad_ind = (
+                pad_inds[0].item() if len(pad_inds) > 0 else input_ids.shape[1]
+            )
+            end_scores.append(rewards[i, first_pad_ind - 1])
```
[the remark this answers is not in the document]
vmoens: For instance, assuming that padding is only on the right:
```python
>>> import torch
>>> z = torch.arange(12).view(3, 4)
>>> z[0, 2:] = 100
>>> z[1, 3:] = 100
>>> z[2, 1:] = 100
>>> mask = z == 100
>>> mask = torch.cat([mask, torch.ones_like(mask[..., :1])], -1) # make sure that there is one True on each row
>>> first_pad = mask[..., :-1] ^ mask[..., 1:]
>>> first_pad = first_pad.nonzero()
>>> first_pad = first_pad[:, -1]
```

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,97 @@
...
+        rejected_ids = rejected_batch.input_ids
+        chosen_rewards = chosen_batch.rewards
+        rejected_rewards = rejected_batch.rewards
+
+        bs = chosen_rewards.shape[0]
+        loss = 0
+
+        for i in range(bs):
+            # Check if there is any padding otherwise take length of sequence
+            c_inds = (chosen_ids[i] == pad_token_id).nonzero()
```
vmoens, line 68: in general nonzero is expensive and shouldn't be called to often

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,97 @@
...
+            c_ind = c_inds[0].item() if len(c_inds) > 0 else chosen_ids.shape[1]
+            r_inds = (rejected_ids[i] == pad_token_id).nonzero()
+            r_ind = r_inds[0].item() if len(r_inds) > 0 else rejected_ids.shape[1]
+            end_ind = max(c_ind, r_ind)
+
+            # Retrieve first index where trajectories diverge
+            divergence_ind = (chosen_ids[i] != rejected_ids[i]).nonzero()[0]
```
vmoens, line 75: ditto, nonzero is expensive

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        chosen_ids = chosen_batch.input_ids
+        rejected_ids = rejected_batch.input_ids
+        chosen_rewards = chosen_batch.rewards
+        rejected_rewards = rejected_batch.rewards
+
+        bs = chosen_rewards.shape[0]
+        loss = 0
+
+        for i in range(bs):
```
vmoens, line 66: I'd be surprised that there is no more optimal solution than a loop with a bunch of calls to nonzero.
We don't need to optimize it now but let's make sure that we mark it as improvable

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,97 @@
...
+        return rewards, torch.stack(end_scores)
+
+    @staticmethod
+    def compute_reward_loss(chosen_batch, rejected_batch, pad_token_id=50256):
+        """Compute the reward loss given a chosen and rejected batch.
+
+        The loss is computed as loss = -log_sigmoid(chosen_reward - rejected_reward).
+        This loss is small when the reward model favours the chosen data and large if
+        the model favours the rejected data.
+        Note: the loss is computed excluding the common "prefix" subsequence to effectively disregard contribution of the original prompt.
```
vmoens, line 56: Perhaps an example of a call to this?

torchrl/modules/models/rlhf.py
```diff
@@ -0,0 +1,97 @@
...
+            pad_inds = (input_ids[i] == self.PAD_ID).nonzero()
+            first_pad_ind = (
+                pad_inds[0].item() if len(pad_inds) > 0 else input_ids.shape[1]
+            )
+            end_scores.append(rewards[i, first_pad_ind - 1])
+
+        return rewards, torch.stack(end_scores)
+
+    @staticmethod
+    def compute_reward_loss(chosen_batch, rejected_batch, pad_token_id=50256):
```
vmoens, line 50: I believe this isn't tested atm

vmoens (commented):
Great work thanks a lot for this.
A couple of comments regarding efficiency: I dont think we need to tackle this now but let's make sure we keep track somewhere that there could be room for improvement

torchrl/modules/models/rlhf.py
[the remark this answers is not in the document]
tcbegley: Thanks! I've factored the slow code out into a private method as you suggested. Did I understand right that we should land as is and follow up with speed improvements?

PR #1321, opened 2023-06-27
[Feature] CatFrames used as inverse 

BY571:
## Description

Adding the 'as_inverse' option to the CatFrames transform. With this transform we can use the CatFrames transform when adding data to the Replay Buffer. By that, the RB will consist of unfolded sequential samples. 

## Types of changes

## Checklist


test/test_transforms.py
```diff
@@ -466,6 +466,29 @@ def test_transform_rb(self, dim, N, padding, rbclass):
...
+        cat_frames = CatFrames(
+            N=N, in_keys=in_keys, dim=dim, padding=padding, as_inverse=True
+        )
+
+        env = TransformedEnv(
+            ContinuousActionVecMockEnv(),
+        )
+        obs_dim = env.specs["output_spec"]["_observation_spec"][
+            "observation_orig"
+        ].shape[0]
```
vmoens, line 485: why not `env.observation_spec["observation_orig].shape[0]`?

torchrl/envs/transforms/transforms.py
```diff
@@ -1960,6 +1960,7 @@ class CatFrames(ObservationTransform):
             has to be written. Defaults to the value of `in_keys`.
         padding (str, optional): the padding method. One of ``"same"`` or ``"zeros"``.
             Defaults to ``"same"``, ie. the first value is uesd for padding.
+        as_inverse (bool, optional): if ``True``, the transform is applied as an inverse transform.
```
vmoens, line 1963:
```suggestion
        as_inverse (bool, optional): if ``True``, the transform is applied as an inverse transform.
            Defaults to ``False``.
```

vmoens (commented):
We should also check what happens when we set `as_inverse=True` and append it to an environment. This should modify the observation in the input tensordict, which is pretty useless unless it is also part of the input keys. 
So either we raise an exception when `as_inverse=True` and the transform has a parent or we make it possible but then we need to define what behaviour is expected and what transformation the observation_spec / input_spec should have. I'm more in favour of the first option (raising an exception if it has a parent).

torchrl/envs/transforms/transforms.py
```diff
@@ -2150,6 +2163,12 @@ def transform_observation_spec(self, observation_spec: TensorSpec) -> TensorSpec
...
 
     def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
+        if self.as_inverse:
+            return tensordict
+        else:
+            return self.unfolding(tensordict)
+
+    def unfolding(self, tensordict: TensorDictBase) -> TensorDictBase:
         # it is assumed that the last dimension of the tensordict is the time dimension
         if not tensordict.ndim or tensordict.names[-1] != "time":
```
BY571, line 2173: Still unsure about this. I know its purpose but not sure if it is really needed. Also if we keep it we probably need a transform that adds the time dimension to a tensordict. Because when using CatFrames as an inverse applied to the replay buffer the data does not have the time dimension. At least for me, it errors out every time.

vmoens: What about a warning that says that we assume the last dim to be "time"? The thing we want to avoid it that time is on another dim and the catframe wrongly assumes that the last dim is time.
Another thing we could do: for now raise a warning, in the future change the behaviour and if the "time" dim is defined then we know which dim must be considered time, and if not we assume it's the last.
TLDR now you can just replace this by a warning

vmoens (commented):
Almost there, can you add a simple check in `test_transform_as_inverse` where you create the transform and append it to an env, and check that the error is raised appropriately?

vmoens (approved):
LGTM (final)

PR #1328, opened 2023-06-28
[Feature] RLHF Reward Model (reopened)

vmoens:
See https://github.com/pytorch/rl/pull/1315 for discussion

PR #1329, opened 2023-06-28
[Feature, NOMERGE] RLHF Rollouts (reopened)

vmoens:
See https://github.com/pytorch/rl/pull/1315 for discussion

### DIFF
--- a/torchrl/data/rlhf/utils.py
+++ b/torchrl/data/rlhf/utils.py
@@ -15,7 +15,8 @@ from transformers import GenerationConfig
 
 
 class RolloutFromModel:
-    """
+    """A class for performing rollouts with language models.
+
     Args:
         model (transformers.Transformer): the model to be used. Should have a
             :meth:`generate` method.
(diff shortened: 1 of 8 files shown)


vmoens:
@tcbegley I think we should merge/rebase main into this

tcbegley:
I've rebased and fixed up the tests. There's still the changes from #1328 visible in the history here because I need the reward model for the rollout. Once that's merged (I think it's ready), I'll do a final merge / rebase and we should be ready here too.

torchrl/data/rlhf/utils.py
```diff
@@ -0,0 +1,391 @@
...
+                sample_log_prob: Tensor(shape=torch.Size([4, 50, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
+            batch_size=torch.Size([4, 50]),
+            device=cpu,
+            is_shared=False)
+    """
+
+    EOS_TOKEN_ID = 50256
+
+    def __init__(
+        self, model, ref_model, reward_model, max_new_tokens=50, score_clip=10.0
```
vmoens, line 89: long term we may want to make this reward model optional and leave to the user the option of computing reward at a different time (eg when populating the replay buffer)

vmoens (commented):
I can't approve it bc it's my PR but LGTM anyway!

apbard (approved):
LGTM! thanks

PR #1331, opened 2023-06-28
[Refactor,Performance] Faster collectors (bis)

vmoens:
The perf of stack onto is better than the perf of stack (compared with call to contiguous(), otherwise no real stack occurs)
image


torchrl/collectors/collectors.py
```diff
@@ -783,8 +782,9 @@ def _step_and_maybe_reset(self) -> None:
         done_or_terminated = (
             (done | truncated) if truncated is not None else done.clone()
         )
-
         if done_or_terminated.any():
+            traj_ids = self._tensordict.get(("collector", "traj_ids"))
+            traj_ids = traj_ids.clone()
```
matteobettini, line 787: Why do we need this clone?

torchrl/collectors/collectors.py
```diff
@@ -795,11 +795,14 @@ def _step_and_maybe_reset(self) -> None:
                 dtype=torch.bool,
             )
             if td_reset.batch_dims:
+                # better cloning here than when passing the td for stacking
+                # cloning is necessary to avoid modifying dones in-place
+                self._tensordict = self._tensordict.clone()
```
matteobettini, line 800: Can you explain a bit more this? to avoid modifying the done we have to clone all the td?

matteobettini (approved):
LGTM, i just have a question about the 2 clones. These are very expensive so just wanna make sure there is absolutely no way to avoid them

torchrl/collectors/collectors.py
vmoens: these are integers so it's not expensive (not as expensive as others)
it's because down there we modify it in-place so it's either that or that.
If we don't clone, we will also modify the value we cache in rollout and all the values stacked on the output td will have the same content (which is bad)

torchrl/collectors/collectors.py
vmoens: First note that this clone and the other only occur when we reset. For pendulum for instance, that is 1% of the trials (more if you have a lot of envs in the collector obv)
The reason is the same as the one in traj_ids: after this, we index the `_tensordict` with the `_reset` mask and change values in place. If we don't clone, we will modify the td we keep in the cache in rollout, therefore modifying the data at t-1, which should be avoided at all cost

matteobettini: For the first note, I think we should not apply that reasoning.
I think batched environments will take over in RL since you can run them parallely on the GPU.
In those envs we will basically be resetting at every step bacause at least 1 out of 60_000 envs will be done.

So it is better to assume that reset is run at every step.

That said, yeah if we can't avoid it there is nothing to do.
But maybe we can only clone some keys of interest if the possible modification will not impact all keys

torchrl/collectors/collectors.py
```diff
@@ -795,11 +795,14 @@ def _step_and_maybe_reset(self) -> None:
                 dtype=torch.bool,
             )
```
matteobettini, line 796:
```suggestion
            td_reset = td_reset.exclude("_reset")
```

Does it make sense to exlude the _reset key here in order not to copy it in  self._tensordict

torchrl/collectors/collectors.py
```diff
@@ -809,7 +812,7 @@ def _step_and_maybe_reset(self) -> None:
             traj_ids[traj_done_or_terminated] = traj_ids.max() + torch.arange(
                 1, traj_done_or_terminated.sum() + 1, device=traj_ids.device
             )
-            self._tensordict.set_(
+            self._tensordict.set(
                 ("collector", "traj_ids"), traj_ids
             )  # no ops if they already match
```
matteobettini, line 817: Will there still be a case where this is np ops?

torchrl/collectors/collectors.py
vmoens: We must copy the data since reset resets the observation at the root for the selected tds. The alternative is to clone at each iteration in rollout, which is far worse
Looking at this we may have a solution:

image

I can try it if you want

torchrl/collectors/collectors.py
vmoens: no we can remove that comment

torchrl/collectors/collectors.py
vmoens: exclude is IMO far more expensive than cloning, as it's cpu-bound (clone is gpu-bound)
and for highly nested tds it will slow things down 
On my machine
image

matteobettini: right, so what the idea wast is that _reset is actually being copied into self._tensordict for no reason so I was just looking for a way to avoid that

torchrl/collectors/collectors.py
vmoens: Leaving this as a TODO as it breaks with lazy stacked tds

PR #1303, opened 2023-06-21
[BugFix] Parametric collectors

matteobettini:

PR #1332, opened 2023-06-28
[BugFix] Check env specs for nested envs

### LINKED ISSUE #1268: [BUG] `check_env_specs()` problems with nested keys
This line fails with nested specs https://github.com/pytorch/rl/blob/main/torchrl/envs/utils.py#L346

This is because it will call `_check_isin` with value `agents`, which is a key present in all specs.
So what happens is that `_check_isin` will always trigger `elif key in input_spec["_action_spec"].keys(True):` even when the agents key does not contain action subkeys but just observations.

matteobettini:
Fixes #1268
Depends on #1303

Here i mostly took code from @vmoens  in #1204

### DIFF
--- a/test/test_env.py
+++ b/test/test_env.py
@@ -1400,8 +1400,6 @@ class TestNestedSpecs:
         assert ("next", *env.done_key) in next_state.keys(True)
         assert ("next", *env.reward_key) in next_state.keys(True)
 
-        # check_env_specs(env)
-
     @pytest.mark.parametrize("batch_size", [(), (32,), (32, 1)])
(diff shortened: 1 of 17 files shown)


test/test_env.py
```diff
@@ -1402,8 +1402,6 @@ def test_nested_env(self, envclass):
         assert ("next", *env.done_key) in next_state.keys(True)
         assert ("next", *env.reward_key) in next_state.keys(True)
 
-        # check_env_specs(env)
```
matteobettini, line 1405: we can remove this since we test the same thing at line 1456

vmoens (approved):
LGTM
let's wait until tests pass

PR #1334, opened 2023-06-28
[BugFix] Fix `step_mdp()`

### LINKED ISSUE #1333: [BUG] `step_mdp` exclude arguments diverge from documentation
In the docstring of `step_mdp` (https://github.com/pytorch/rl/blob/main/torchrl/envs/utils.py#L53) examples are given using the following tensordict

```pyhton
td = TensorDict(
        {
            "done": torch.zeros((), dtype=torch.bool),
            "reward": torch.zeros(()),
            "extra": torch.zeros(()),
            "next": TensorDict(
                {
                    "done": torch.zeros((), dtype=torch.bool),
                    "reward": torch.zeros(()),
                    "obs": torch.zeros(()),
                },
                [],
            ),
            "obs": torch.zeros(()),
            "action": torch.zeros(()),
        },
        [],
    )
```

the results of the following snippets in main does not align with the ones in the docs and is quite counterintuitive
```python
res = step_mdp(td, exclude_done=True)
assert "done" not in res.keys() # Error
```

```python
res = step_mdp(td, exclude_reward=True)
assert "reward" not in res.keys() # Error
```

## Cause

This is caused by the fact that when `keep_other=True` (like in this cases since it is default) the root keys will be kept as is, so the flags `exclude_reward` and `exclude_done` will prevent the keys in next to be used, but `keep_other=True` will keep the root ones.

## Solutions

I see 2 possibilities

1. make the `exclude_reward` and `exclude_done` behave as in the docs and actually exclude both the ones in root and next
2. if for some some reason this is intended a refactoring of the arg names is needed

matteobettini:
Fixes #1333 with solution 1

[WIP] compatibility with nested keys

### DIFF
--- a/test/test_env.py
+++ b/test/test_env.py
@@ -1021,6 +1021,8 @@ def test_steptensordict(
     torch.manual_seed(0)
     tensordict = TensorDict(
         {
+            "reward": torch.randn(4, 1),
+            "done": torch.zeros(4, 1, dtype=torch.bool),
             "ledzep": torch.randn(4, 2),
             "next": {
                 "ledzep": torch.randn(4, 2),
--- a/torchrl/envs/utils.py
+++ b/torchrl/envs/utils.py
@@ -179,13 +179,11 @@ def step_mdp(
         return out
     out = tensordict.get("next").clone(False)
     excluded = set()
-    if exclude_done and exclude_reward:
-        excluded = {done_key, reward_key}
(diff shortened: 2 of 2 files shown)


vmoens (approved):
LGTM
I had a go at the docstrings, lmk if that fits what you had in mind

matteobettini:
Good to go for me!

PR #1337, opened 2023-06-29
[BugFix] valid_size not saved as attribute

tcbegley:
## Description

This PR fixes a bug that arose from valid_size not being used in the constructor of TokenizedDataLoader.

The bug was not caught by the tests because it was only relevant if we used a validation split of the data. To ensure that tests catch a similar regression in the future we parametrized `split` in the test.


vmoens (approved):
LGTM

PR #1339, opened 2023-06-30
[BugFix] `step_mdp` nested keys

matteobettini:


torchrl/envs/utils.py
```diff
@@ -155,6 +155,13 @@ def step_mdp(
             is_shared=False)
 
     """
+    if isinstance(done_key, tuple) and len(done_key) == 1:
+        done_key = done_key[0]
+    if isinstance(reward_key, tuple) and len(reward_key) == 1:
+        reward_key = reward_key[0]
+    if isinstance(action_key, tuple) and len(action_key) == 1:
+        action_key = action_key[0]
```
vmoens, line 163: this is expensive let's not do that if we can avoid it
We could blend this in unravel_keys in tensordict (which is coded in c++) if it's there for efficiency purposes

torchrl/envs/utils.py
```diff
@@ -181,28 +188,25 @@ def step_mdp(
...
     if exclude_done:
         excluded.add(done_key)
     if exclude_reward:
         excluded.add(reward_key)
     if len(excluded):
         out = out.exclude(*excluded, inplace=True)
-    td_keys = None
+
+    if not exclude_action:
+        out._set(action_key, tensordict.get(action_key))
```
vmoens, line 202: Here if action has to be kepts we're removing it and adding it back, which is expensive, let's avoid that

torchrl/envs/utils.py
```diff
@@ -181,28 +188,25 @@ def step_mdp(
...
         return out
+
     out = tensordict.get("next").clone(False)
-    excluded = set()
+    excluded = {action_key}
     if exclude_done:
         excluded.add(done_key)
     if exclude_reward:
         excluded.add(reward_key)
     if len(excluded):
```
vmoens, line 198: that will never happen anymore, so we'll always be calling exclude, which is expensive. Let's avoid that

torchrl/envs/utils.py
```diff
@@ -181,28 +188,25 @@ def step_mdp(
...
-            td_keys = td_keys - {action_key}
-    elif not exclude_action:
-        td_keys = {action_key}
-
-    if td_keys:
-        # update does some checks that we can spare
-        # out.update(tensordict.select(*td_keys))
-        for key in td_keys:
-            out._set(key, tensordict.get(key))
+        excluded = set.union(excluded, set(out.keys(True, True)))
```
vmoens, line 205: set on all the keys is super expensive, i'd avoid it if we can

vmoens (commented):
The tests look good, but I suspect the implementation of step_mdp will be drastically slower than the one we have now. 
Given the number of issues opened by users lately regarding environments overhead compared to gym, I'm reluctant to any change in step_mdp that makes it slower than what it already is.
Let's wait until we merge the 2 benchmark PRs (the one with step_mdp and the other to run benchmarks in PRs), then it'll be easier to iterate over this

torchrl/envs/utils.py
matteobettini: we are not removing it, the exclusion before is on "next"

torchrl/envs/utils.py
matteobettini: ok

torchrl/envs/utils.py
```diff
@@ -181,34 +187,46 @@ def step_mdp(
...
-        excluded.add(reward_key)
-    if len(excluded):
-        out = out.exclude(*excluded, inplace=True)
-    td_keys = None
+
+    td = tensordict.exclude("next")
+    td_next = tensordict.get("next")
+
+    td_keys = td.keys(True, True)
+    td_next_keys = td_next.keys(True, True)
```
matteobettini, line 195: These are is the only time we traverse.
Basically we just visit every key in tensordict

torchrl/envs/utils.py
```diff
@@ -181,34 +187,46 @@ def step_mdp(
...
+        done_key if exclude_done else None,
+        reward_key if exclude_reward else None,
+    }
+    for key in td_next_keys:
+        if key not in excluded:
+            _set_key(dest=out, source=td_next, key=key)
+
+    # Set the keys from root
+    if not exclude_action:
+        _set_key(dest=out, source=tensordict, key=action_key)
```
matteobettini, line 210: here we do the action separate just to not have another if later, perf is the same

torchrl/envs/utils.py
```diff
@@ -181,34 +187,46 @@ def step_mdp(
...
-    if exclude_done:
-        excluded.add(done_key)
-    if exclude_reward:
-        excluded.add(reward_key)
-    if len(excluded):
-        out = out.exclude(*excluded, inplace=True)
-    td_keys = None
+
+    td = tensordict.exclude("next")
+    td_next = tensordict.get("next")
```
vmoens, line 192: use pop

torchrl/envs/utils.py
```diff
@@ -181,34 +187,46 @@ def step_mdp(
...
+    td_keys = td.keys(True, True)
+    td_next_keys = td_next.keys(True, True)
+
+    out = TensorDict({}, td_next.batch_size)
+
+    # Set the keys from next
+    excluded = {
+        done_key if exclude_done else None,
+        reward_key if exclude_reward else None,
+    }
```
vmoens, line 203: Isn't weird to have None in the set?
Should we not build it iteratively?

torchrl/envs/utils.py
```diff
@@ -181,34 +187,46 @@ def step_mdp(
...
         return next_tensordict.update(out)
     else:
         return out
 
 
+def _set_key(dest, source, key):
+    dest._set(key, source.get(key))
+    if isinstance(key, tuple) and len(key) > 1:  # Setting the batch_sizes
+        for k in range(1, len(key)):
+            dest[key[:k]].batch_size = source[key[:k]].batch_size
```
vmoens, line 227: if two keys share the same root, we'll set the batch-time multiple times, which is time consuming

vmoens (commented):
I can see a version of that working but the way we set the batch size requires multiple access to nested tensordicts that will be time consuming 
If you merge main into this branch you'll get the time measure of your solution

torchrl/envs/utils.py
matteobettini: i can do it, i thought preformance was similar and it was more readable, but i ll change it

torchrl/envs/utils.py
matteobettini: yeah I know, i ll improve it

vmoens (commented):
LGTM

PR #1340, opened 2023-06-30
[Benchmark] Improve benchmark precision + step_mdp + fix GPU

vmoens:
cc @matteobettini

### DIFF
--- a/benchmarks/test_envs_benchmark.py
+++ b/benchmarks/test_envs_benchmark.py
@@ -8,8 +8,7 @@ import pytest
 import torch
 
 from tensordict import TensorDict
-from torchrl.envs import ParallelEnv, SerialEnv, StepCounter, TransformedEnv, \
-    step_mdp
+from torchrl.envs import ParallelEnv, SerialEnv, step_mdp, StepCounter, TransformedEnv
 from torchrl.envs.libs.dm_control import DMControlEnv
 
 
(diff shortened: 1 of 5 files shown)


benchmarks/test_envs_benchmark.py
```diff
@@ -36,21 +40,92 @@ def make_parallel_env():
     return ((env,), {})
 
 
+def make_nested_td():
+    return TensorDict(
+        {
+            ("agent", "action"): 0,
+            ("agent", "done"): 0,
+            ("agent", "obs"): 0,
```
matteobettini, line 48: we might want an "other" key here which is not in next

benchmarks/test_envs_benchmark.py
```diff
@@ -36,21 +40,92 @@ def make_parallel_env():
...
+        [],
+    )
+
+
+def make_flat_td():
+    return TensorDict(
+        {
+            ("action"): 0,
+            ("done"): 0,
+            ("obs"): 0,
```
matteobettini, line 63: ditto

benchmarks/test_envs_benchmark.py
```diff
@@ -36,21 +40,92 @@ def make_parallel_env():
     return ((env,), {})
 
 
+def make_nested_td():
```
matteobettini, line 43: If you want you can use some parts of test_nested in https://github.com/pytorch/rl/pull/1339/files for a more parametric creation of nested td with options to nest only some pats of it

vmoens: Seems too complex for a benchmark: the tensors are bigger and with random values, here we have just 0s and 1s and we just focus on the tensordict operations on keys etc

matteobettini: oh yes i was just saying that maybe we want `nested_done`, `nested_reward`, `nested_action` as arguments instead of a general `nested`

matteobettini (approved):
LGTM

PR #1341, opened 2023-06-30
[Benchmark,CI] Benchmarks in PR

vmoens:


benchmarks/conftest.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
apbard, line 1: move this file to a separate PR and merge it first. This should solve the error you are seing about missing --rank

vmoens: Done thanks

matteobettini (approved):
LGTM

PR #1347, opened 2023-07-03
[BugFix] Nested envs compatibility

matteobettini:


torchrl/envs/common.py
```diff
@@ -1252,8 +1249,12 @@ def policy(td):
                 break
             tensordict = step_mdp(
                 tensordict,
-                keep_other=True,
-                exclude_action=False,
+                keep_other=False,
+                exclude_action=True,
```
matteobettini, line 1253: If we set `keep_other=False`,`exclude_action=True` we may save time.
Was there a reason they were set differently? I might be missing something

torchrl/envs/common.py
```diff
@@ -1253,7 +1249,11 @@ def policy(td):
             tensordict = step_mdp(
                 tensordict,
                 keep_other=True,
-                exclude_action=False,
+                exclude_action=True,
```
matteobettini, line 1252: If we set `exclude_action=True` we may save time.
Was there a reason it was set differently?
 I might be missing something

vmoens: There was probably a reason why but I can't recall
If tests pass without it i'm open to the change

vmoens (approved):
LGTM
Can we merge main? A couple of tests should be fixed

matteobettini:
main is merged

PR #1348, opened 2023-07-03
[BugFix] Batched envs compatibility with custom keys

### LINKED ISSUE #1269: [BUG] VecEnv not compatible with parametric keys

matteobettini:
Depends on #1347 
Fixes #1269
Draws from #1270 which should be closed after

### DIFF
--- a/test/mocking_classes.py
+++ b/test/mocking_classes.py
@@ -1127,10 +1127,15 @@ class NestedCountingEnv(CountingEnv):
                 shape=self.batch_size,
             )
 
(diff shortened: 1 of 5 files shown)


vmoens (approved):
LGTM thanks for this
Let's wait for the tests to pass

PR #1352, opened 2023-07-03
[BugFix] SafeModule not safely handling specs

### LINKED ISSUE #1278: [BUG] ActorCriticWrapper overriding module specs
I am using ddpg with modules that have composite specs

I have an actor `AdditiveGaussianWrapper` with spec
```
CompositeSpec(
    agents: CompositeSpec(
        action: BoundedTensorSpec(
            shape=torch.Size([3, 2]),
            space=ContinuousBox(
                minimum=Tensor(shape=torch.Size([3, 2]), device=cpu, dtype=torch.float32, contiguous=True), 
                maximum=Tensor(shape=torch.Size([3, 2]), device=cpu, dtype=torch.float32, contiguous=True)),
            device=cpu,
            dtype=torch.float32,
            domain=continuous), device=cpu, shape=torch.Size([3])), device=cpu, shape=torch.Size([]))
```

after i pass it to the ddpg loss and ` actor_critic = ActorCriticWrapper(actor_network, value_network)` is run my actor spec immediaely is ovewritten with 

```
CompositeSpec(
    agents: CompositeSpec(
        action: None,
        state_action_value: None,
        param: None, device=cpu, shape=torch.Size([3])), device=cpu, shape=torch.Size([]))
```

matteobettini:
The safe module is very unsafe ;)
Fixes #1278

### DIFF
--- a/torchrl/data/tensor_specs.py
+++ b/torchrl/data/tensor_specs.py
@@ -2584,8 +2584,11 @@ class CompositeSpec(TensorSpec):
                     if self._device is None:
                         try:
                             self._device = item.device
-                        except RuntimeError:
-                            self._device = item._device
+                        except RuntimeError as err:
+                            if DEVICE_ERR_MSG in str(err):
+                                self._device = item._device
+                            else:
+                                raise err
                 self[k] = item
 
     @property
--- a/torchrl/data/tensor_specs.py
+++ b/torchrl/data/tensor_specs.py
@@ -2582,7 +2582,10 @@ class CompositeSpec(TensorSpec):
                     item = CompositeSpec(item, shape=shape)
                 if item is not None:
                     if self._device is None:
-                        self._device = item.device
+                        try:
+                            self._device = item.device
+                        except RuntimeError:
+                            self._device = item._device
                 self[k] = item
 
     @property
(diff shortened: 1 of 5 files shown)


matteobettini:
this might be the cause of #1278

matteobettini:
Actually fixes #1278

vmoens (approved):
LGTM thanks!

torchrl/data/tensor_specs.py
```diff
@@ -2582,7 +2582,10 @@ def __init__(self, *args, shape=None, device=None, **kwargs):
                     item = CompositeSpec(item, shape=shape)
                 if item is not None:
                     if self._device is None:
-                        self._device = item.device
+                        try:
+                            self._device = item.device
+                        except RuntimeError:
+                            self._device = item._device
```
matteobettini, line 2588: I added this because otherwise it would throw an error when trying to clone a CompositeSpec with no device

vmoens, line 2588: I'm ok with this but maybe let's keep the error and raise it if the item does not have `_device`? Or are we sure it has it?
```suggestion
                        try:
                            self._device = item.device
                        except RuntimeError as err:
                            if some_pattern in str(err):
                                self._device = item._device
                            else:
                                raise err
```

vmoens (approved):
LGTM minor comment

torchrl/data/tensor_specs.py
matteobettini: If the error is raised it means that item._device is None, in which case we can still clone with None device

matteobettini: we are sure that item._device is None so basically we set that

vmoens (approved):
Final

PR #1355, opened 2023-07-04
[BugFix] Nested keys in transforms

matteobettini:


torchrl/envs/transforms/transforms.py
```diff
@@ -2489,8 +2489,10 @@ def __init__(
                 )
         else:
             in_keys = sorted(in_keys, key=_sort_keys)
-        if type(out_key) != str:
-            raise Exception("CatTensors requires out_key to be of type string")
+        if not isinstance(out_key, str) and not (
+            isinstance(out_key, tuple) and set(map(type, out_key)) == {str}
+        ):
```
vmoens, line 2494: See https://github.com/pytorch-labs/tensordict/pull/479/files#r1254257836

I wonder if this is needed

vmoens (approved):
LGTM very thorough!

One small comment

torchrl/envs/transforms/transforms.py
matteobettini: should we just chek it is not a list?

vmoens:
examples CI seems to be broken, not sure it's related to this PR (maybe Double2Float?)

matteobettini:
oh thanks let me have a look

PR #1357, opened 2023-07-04
[Refactor] More unravel fixes

vmoens:
The problem this PR solves is that we made changes in the way unravel_key(s) work but we'll revert that in tenrordict.
In the meantime, torchrl has been adapted.
Rather than reverting the PR in torchrl which will break the CI, we will do the following:
  - Merge this PR in torchrl to emulate the behaviour of unravel_key in tensordict after https://github.com/pytorch-labs/tensordict/pull/473 is merged
  - Merge https://github.com/pytorch-labs/tensordict/pull/473 in tensordict
  - Revert imports in torchrl to tensordict instead of emulated functions

I expect this will temporarily impact negatively the benchmarks but they should recover afterwards.

cc @matteobettini

### DIFF
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -17,6 +17,8 @@ import warnings
 from copy import copy
 from distutils.util import strtobool
 from functools import wraps
+
+# from tensordict._tensordict import unravel_keys
 from importlib import import_module
 from typing import Any, Callable, cast, TypeVar, Union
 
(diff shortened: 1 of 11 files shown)


torchrl/_utils.py
```diff
@@ -529,3 +531,39 @@ def clone(self):
...
+    for key in key_list:
+        key = unravel_keys(key)
+        if isinstance(key, tuple) and len(key) == 1:
+            key_list_out.append(key[0])
+        else:
+            key_list_out.append(key)
+    return key_list_out
+
+
+def unravel_keys(key):
```
matteobettini, line 550: Why are we still keeping the name unravel_keys? Cant we do all this and also stop using that plural in all the 2 projects?
I thought after this whole thing we were going to have:

- unravel_key
- unravel_key_list
- _unravel_tuple_key

with unravel_keys dead

matteobettini (approved):
LGTM, just one comment

torchrl/_utils.py
vmoens: this is just for this PR, it'll be used for 1 hour max
I just don't want the tests to break at any time bc that will be a pain to handle internally on fbcode
I can rename it unravel_key but that's beyond the point IMO
We could name it `_tmp_fun_that_we_will_remove_` fwiw

matteobettini: ok got it go for it

vmoens: btw we won't kill anything for bc reasons, we'll just keep both names (perhaps wrapping the plural in a deprecation decorator)

PR #1366, opened 2023-07-06
[Refactor] Vmas nested

matteobettini:
According to all the changes in the lib for multiagent, here is the new version of vmas using nested keys.

Done is not nested as it is shared by all.
All the other specs are nested

### DIFF
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -1286,7 +1286,7 @@ class TestVmas:
     @pytest.mark.parametrize("n_envs", [1, 4])
     @pytest.mark.parametrize("n_workers", [1, 2])
     @pytest.mark.parametrize("n_agents", [1, 3])
-    def test_collector(self, n_envs, n_workers, n_agents, frames_per_batch=100):
+    def test_collector(self, n_envs, n_workers, n_agents, frames_per_batch=80):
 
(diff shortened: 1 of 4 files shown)


vmoens (approved):
LGTM
I see some keys have changed: does that mean that this is bc-breaking? If so, we should add the tag and perhaps think carefully about how impactful it is

matteobettini:
you are right, this is BC breaking.

it is unavoidable since we confirmed that the new multiagent api in torch rl will have none-shared agent keys in a nested tensordict called "agents".

This is the reflection of that change in vmas.

The main impacts for users are:
- the key "agents" has to be preappended to the "reward","observation","info", and "action" entries
- the default keys for modules are not valid anymore and the specific NestedKeys will have to be fed as parameters
- losses will have to be informed of the custom keys via using the appropriate `loss.set_keys()` method

To mitigate this I have written some examples on how to use the simulator in 
https://github.com/pytorch/rl/pull/1027 which show how all these things can be done for various losses

Do you want to wait to merge this for when we merge the examples to smooth the transition?

vmoens:
I added the label, I think we can consider it as not too annoying to be merged as is

matteobettini:
sounds good to me

PR #1368, opened 2023-07-07
Fix CI

matteobettini:


vmoens (approved):

vmoens:
Some tests are sporadically failing but we should be fine

PR #1374, opened 2023-07-07
Remove Distributed CCI job

osalpekar:
See Title. Also removes the definitions for the Linux CPU/GPU jobs since they aren't run on CCI anymore.

cc @vmoens


vmoens (approved):
LGTM Thanks

PR #1377, opened 2023-07-10
[WIP] Expose sub-losses

vmoens:
Exposes sub-objectives like loss_actor and loss_value across losses

cc @matteobettini @smorad

### DIFF
--- a/torchrl/objectives/td3.py
+++ b/torchrl/objectives/td3.py
@@ -200,6 +200,457 @@ class TD3Loss(LossModule):
     out_keys = [
         "loss_actor",
         "loss_qvalue",
+        "state_action_value_actor",
+        "pred_value",
+        "next_state_value",
+        "target_value",
+    ]
+
+    def __init__(
+        self,
+        actor_network: TensorDictModule,
+        qvalue_network: TensorDictModule,
+        *,
+        action_spec: TensorSpec = None,
+        bounds: Optional[Tuple[float]] = None,
+        num_qvalue_nets: int = 2,
+        policy_noise: float = 0.2,
+        noise_clip: float = 0.5,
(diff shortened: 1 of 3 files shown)


torchrl/objectives/td3.py
```diff
@@ -331,66 +330,484 @@ def in_keys(self, values):
     def _cached_detach_qvalue_network_params(self):
         return self.qvalue_network_params.detach()
 
-    @property
-    @_cache_values
-    def _cached_stack_actor_params(self):
-        return torch.stack(
-            [self.actor_network_params, self.target_actor_network_params], 0
+    def actor_loss(self, tensordict):
```
vmoens, line 333: Should we @dispatch?

matteobettini:
uhh amazing, do you think it makes sense to separate the target too as discussed in #1000?

torchrl/objectives/td3.py
matteobettini: mmh it would mean defining in and out keys for each function right.

It might make sense so that it allow understanding what keys each function should access, but it might clutter the code and if the forward is already dispatched it is only useful when mixing losses

vmoens:
> uhh amazing, do you think it makes sense to separate the target too as discussed in #1000?

We'd have 
```
def forward(self, td):
    target, metadata_target = self.compute_target(td)
    loss_actor, metadata_actor = self.actor_loss(td)
    loss_value, metadata_value = self.value_loss(td)
```

It could make sense yep, I just need to see that this has a limited overhead.

smorad:
Looks good. I'm questioning the separation of

gtnv:
hi @vmoens, do you plan to continue working on this? if not, should we close it?

PR #1378, opened 2023-07-11
[Algorithm] QMixer loss and multiagent models

matteobettini:


matteobettini:
@acciorocketships it would be nice to add a QGNN mixer in modules/models/multiagent.py.
implementing the mixer interface

Acciorocketships:
What about the QGNN model? If that's not already implemented, I would prioritise that. It has a much larger impact on performance.

matteobettini:
Yeah that is just a normal gnn right? I am thinking to make a MultiagentGNN as i made the MLP

matteobettini:
Ready for review @vmoens

matteobettini:
Thanks a lot! Will fix.

Most of the comments on the new loss are on parts that have been copy-pasted from DQN.
Should i modify that too when I fix those?

vmoens:
> Thanks a lot! Will fix.
> 
> Most of the comments on the new loss are on parts that have been copy-pasted from DQN. Should i modify that too when I fix those?

You mean the loss_function default? Yeah sure makes sense

matteobettini:
No not the default.
I am just saying that most of the code of qmixerloss comes from dqnloss.

So your review comments apply to dqnloss too.
I'll modify that too when i fix them

torchrl/objectives/multiagent/qmixer.py
```diff
@@ -0,0 +1,383 @@
...
+        priority_tensor = (pred_val_index - target_value).pow(2)
+        priority_tensor = priority_tensor.detach().unsqueeze(-1)
+        if tddevice.device is not None:
+            priority_tensor = priority_tensor.to(tddevice.device)
+
+        tensordict.set(
+            self.tensor_keys.priority,
+            priority_tensor,
+            inplace=True,
+        )
```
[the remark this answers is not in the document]
matteobettini: good question, i do not know the reason since i took this from dqn, should I change?

torchrl/objectives/multiagent/qmixer.py
```diff
@@ -0,0 +1,383 @@
...
+            "reward": self.tensor_keys.reward,
+            "done": self.tensor_keys.done,
+        }
+        self._value_estimator.set_keys(**tensor_keys)
+
+    @dispatch
+    def forward(self, tensordict: TensorDictBase) -> TensorDict:
+        device = self.device if self.device is not None else tensordict.device
+        tddevice = tensordict.to(device)
+
```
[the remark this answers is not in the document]
matteobettini: i removed the use of loss device in the dqn and qmix losses, but deprecation is left to future prs

torchrl/objectives/multiagent/qmixer.py
```diff
@@ -0,0 +1,383 @@
...
+    @property
+    def in_keys(self):
+        if self._in_keys is None:
+            self._set_in_keys()
+        return self._in_keys
+
+    @in_keys.setter
+    def in_keys(self, values):
+        self._in_keys = values
```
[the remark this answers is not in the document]
matteobettini: not sure, this was added by martin in dqn, we can remove

torchrl/objectives/dqn.py
```diff
@@ -304,13 +301,14 @@ def forward(self, tensordict: TensorDictBase) -> TensorDict:
             pred_val_index = (pred_val * action).sum(-1)
 
         target_value = self.value_estimator.value_estimate(
-            tddevice.clone(False), target_params=self.target_value_network_params
+            td_copy, target_params=self.target_value_network_params
```
matteobettini, line 304: I think this clone is not needed. 
Pleas double check

matteobettini: Also shouldn't this also be wrapped in no_grad()? 
Or the fact that we use target params already takes care of that?

vmoens: yep target are detached

torchrl/objectives/multiagent/qmixer.py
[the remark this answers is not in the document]
vmoens: Probably safer not to allow this, let's remove it and check that tests pass

torchrl/objectives/multiagent/qmixer.py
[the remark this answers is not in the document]
vmoens: I think originally the plan was not to have to call `update_priority` on the RB but it's unsafe, let's remove the inplace=True

torchrl/objectives/dqn.py
```diff
@@ -231,10 +231,6 @@ def in_keys(self):
             self._set_in_keys()
         return self._in_keys
 
-    @in_keys.setter
-    def in_keys(self, values):
-        self._in_keys = values
-
```
matteobettini, line 237: this change you requested makes it bc-breaking

matteobettini: do you want me to still allow it but warn?

vmoens: no if tests pass without needing it we're good without it

vmoens:
Tests are failing on GPU tests:
```
test/test_cost.py::TestQMixer::test_qmix_batcher[categorical-device0-True-2] FAILED [  4%]
___________ TestQMixer.test_qmix_batcher[categorical-device0-True-2] ___________
  File "/work/env/lib/python3.9/site-packages/tensordict/nn/common.py", line 1178, in forward
    raise err
  File "/work/env/lib/python3.9/site-packages/tensordict/nn/common.py", line 1164, in forward
    tensors = self._call_module(tensors, **kwargs)
  File "/work/env/lib/python3.9/site-packages/tensordict/nn/common.py", line 1121, in _call_module
    out = self.module(*tensors, **kwargs)
  File "/work/env/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1522, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/work/env/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1531, in _call_impl
    return forward_call(*args, **kwargs)
  File "/work/env/lib/python3.9/site-packages/tensordict/nn/functional_modules.py", line 567, in new_fun
    return getattr(type(self), fun_name)(self, *args, **kwargs)
  File "/work/env/lib/python3.9/site-packages/torch/nn/modules/linear.py", line 114, in forward
    return F.linear(input, self.weight, self.bias)
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument mat2 in method wrapper_CUDA_mm)
```

vmoens:
There is still this to be fixed
```
test/test_cost.py::TestQMixer::test_qmix_batcher[categorical-device0-False-3] - assert not True
```

matteobettini:
> There is still this to be fixed

Thanks a lot and sorry for not noticing that.

For the future, can you explain to me where do you see such failures?
I was opening the gpu tests pipelines but I did not see that

vmoens (approved):
Great work!  Woohoo!

PR #1386, opened 2023-07-13
[CI] Windows GPU Tests

osalpekar:
Windows GPU job is working (at least from an env perspective). This is the most expensive remaining job on CCI, so this PR migrated it to GHA. Will remove the GHA job in a follow-up.


vmoens (approved):
LGTM

Issue #1209, opened 2023-05-30
[Feature Request] Missing ActionScaling and FlattenAction 


duburcqa:
## Motivation

In general, the action space may be a dict instead of a plain vector, and is not normalized. Even for a classic `gym` environment, action wrappers performing action flattening and normalization according the min/max spec are not provided by the standard API, but are implemented by many libraries.

## Solution

First, add a normalization transform `ActionScaling` to `torchrl` that relies on the spec to compute `loc` and `scale` unlike `ObservationNorm` nor moving average like `VecNorm`. I would advice raising an exception for completely or partially unbounded action spaces. Second, add an action flattening transform `FlattenAction` foing very much the same as `FlattenObservation`.

## Alternatives

Keep relying on flattening and normalization wrappers at env-level provided by any other library.

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

vmoens:
Got it sounds about right!
We could also have a method in observation norm to compute loc and scale based on the spec now that you mention that.

duburcqa:
Yes, but it is much less common for observation space to be fully bounded. In any case, the observation distribution under the policy rarely span the whole space in practice, so it would probably be much less frequently used.

PSXBRosa:
Hello,

I would like to work on this Issue, is anyone already working on it?

Just for making sure I interpreted the issue correctly, the goal is implementing those two wrappers for the action space, is that right?

duburcqa:
Yes it is ! I don't think someone is working on it for now.

pxe3:
Hello, @vmoens. I apologize for the necro ping, is this issue still available?

I see that it has been some time since it was last worked on, but it was never resolved. If so, I would be happy to open a PR completing it. This would be my first open-source contribution (of many).

Let me know!

Drakit0:
I've split this into two small, reviewable PRs:

- #3765 - Add `ActionScaling` transform (part 1)
- #3766 - Add `FlattenAction` transform (part 2)

Both follow the existing transform conventions:

- `ActionScaling`: spec-driven affine normalization. Exposes a `[-1, 1]` (or `[0, 1]`) action space to the policy and rescales actions to the env's bounds on the inv direction. Raises `RuntimeError` for fully or partially unbounded specs (covering both `Unbounded` and bounded specs whose entries match the `torch.finfo` extremes used internally by `Unbounded`).
[...]

PR #1388, opened 2023-07-16
[WIP] Added new set of action transforms

PSXBRosa:
## Description

Added three new transforms for transforming actions:

* ActionTransform: a base class
* ActionScaling: for scaling actions according to loc and scale (similar to the ObservationNorm transform)
* ActionNorm: for automatically rescaling actions into a [-1,1] range, following what was proposed in the issue this PR tries to solve

## Motivation and Context

partially solves #1209 

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/__init__.py
+++ b/torchrl/envs/__init__.py
@@ -8,6 +8,8 @@ from .env_creator import EnvCreator, get_env_metadata
 from .gym_like import default_info_dict_reader, GymLikeEnv
 from .model_based import ModelBasedEnvBase
 from .transforms import (
+    ActionNorm,
+    ActionScaling,
     BinarizeReward,
     CatFrames,
     CatTensors,
--- a/torchrl/envs/transforms/__init__.py
+++ b/torchrl/envs/transforms/__init__.py
@@ -6,6 +6,8 @@
 from .r3m import R3MTransform
 from .rlhf import KLRewardTransform
 from .transforms import (
+    ActionNorm,
+    ActionScaling,
     BinarizeReward,
     CatFrames,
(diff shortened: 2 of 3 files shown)


PSXBRosa:
I'm unsure what's the standard for adding tests and of my implementation for the ActionNorm class. I would appreciate some orientation regarding that.

vmoens (commented):
Wonderful!
I like the idea of having a parent transform class for actions. We could make sure that some methods are overridden using `@abstractmethod` too.
Since transforms is getting pretty big, maybe this would be the right thing to put in a separate file?
Also we should have some tests (as you mentioned in the comment).
The process is the following:
In the `test/test_transforms.py` file, you will find a `TransformBase` parent class that lists all the tests we want to see. You can have a look at the other classes that are tested and do a copy paste of the one you think is the closest.
You can perfectly skip tests if they're irrelevant to you, but you must code them anyway (we do this to make sure we're not overlooking anything).
LMK if you have question!
Thanks again for this PR, it's really cool

PSXBRosa:
So I've been trying to expand on this and I've noticed a few things I hadn't noticed before. Apparently the "_apply_transform" method is only applied on the tensordict after passing it through the parent environment. The only Transform method that's called before passing the tensordict would be "_inv_apply_transform" but I think that using this method for adjusting the action tensors would be more of a hack. Which leaves me wondering whether I should adapt the way TransformedEnv calls _step to account for this.

vmoens:
The reason of existence of `_inv_apply_transform` is to apply transforms before the tensordict is exposed to `_step` so this is defo where the operation should happen!

PR #1391, opened 2023-07-18
[Test] Heterogeneous env

matteobettini:
@vmoens @hyerra 

This improves the test in #1204

Look at the spec of this env.

In my opinion it should cover all heterogenous use cases apart from composite actions, can you think of something I missed?


matteobettini:
> Thanks, this looks really good for the heterogeneous part! One suggestion though, is it possible to have a test for where agents have a different number of observations? This is possible in Unity, for instance. For example, maybe agent 1 has both lidar and camera, but agent 2 only has lidar.

are you referring to the fact that the obs tensordicts have the same number of keys? would it cover your case if I remove the sonar from agent 2?

hyerra:
Yep, exactly, that case should cover it! I'm not sure if this case was what caused the bug in the StrikersVsGoalie game to fail. I didn't fully debug it just yet, but that was my initial guess.

PR #1365, opened 2023-07-06
[BugFix] Fix lazy stacked composite specs

vmoens:
Fixes operations in lazy stack tds:
- [ ] unsqueeze
- [ ] squeeze
- [ ] print
- [ ] `__len__`
- [ ] `__eq__`
- [ ] `__ne__`
- [ ] `project`
- [ ] `type_check`
- [ ] `__delitem__`
- [ ] `__iter__`
- [ ] `keys`
- [ ] `__setitem__`

Issue #969, opened 2023-03-15
[Feature Request] Improvements to LazyStackedTensorSpec, LazyStackedCompositeSpec


tcbegley:
## Motivation

#892 introduced two new classes: `LazyStackedTensorSpec`, and `LazyStackedCompositeSpec`. These allow us to stack heterogeneous `TensorSpec` and `CompositeSpec` respectively.

These classes are missing a few features that still need to be added.

## TODO

- [ ] Add documentation - the classes have basic docstrings, but we should add examples etc.
- [ ] Add support for nestedtensors
- [ ] #960 changed the behaviour of `torch.stack` to only return lazy stacks in the heterogenous case. As a result a number of tests now only cover the homogeneous or heterogeneous case but not both. These tests need to be expanded to cover both cases.

**Unimplemented methods** - the following exist as unimplemented placeholder methods in `LazyStackedTensorSpec` and `LazyStackedCompositeSpec` respectively

`LazyStackedTensorSpec`
- [ ] `__eq__` (requires unbind?)
- `__len__`
- `project`
- `is_in`
- `encode` (not relevant, should be removed)
- `__iter__`
- `__setitem__`

LazyStackedCompositeSpec
- `update`
- `__eq__`
- `__len__`
- `project`
- `is_in`
- `type_check`
- `encode`
- `__delitem__`
- `__iter__`
- `__setitem__`

PR #1392, opened 2023-07-18
[BugFix] Fix `LazyStackedCompositeSpec` and introducing `consolidate_spec`

### LINKED ISSUE #1343: [BUG] Operations on Spec of Stacked CompositeSpec with different keys crashes
## Describe the bug

I'm trying to print a `CompositeSpec`, but the current implementation crashes with a `KeyError`. This is related to #1201.
cc: @matteobettini

## To Reproduce

Please try running the following code example:

```python
import torch
from torchrl.data.tensor_specs import (
    CompositeSpec,
)

specs = [CompositeSpec({
    "observation_0": UnboundedContinuousTensorSpec(
        shape=torch.Size([128, 128, 3]),
        device="cpu",
        dtype=torch.float32)}),
    CompositeSpec({
        "observation_1": UnboundedContinuousTensorSpec(
        shape=torch.Size([128, 128, 3]),
        device="cpu",
        dtype=torch.float32)}), CompositeSpec({
    "observation_2": UnboundedContinuousTensorSpec(
        shape=torch.Size([128, 128, 3]),
        device="cpu",
        dtype=torch.float32)}), CompositeSpec({
    "observation_3": UnboundedContinuousTensorSpec(
        shape=torch.Size([4]),
        device="cpu",
        dtype=torch.float32)})]

spec = torch.stack(specs, dim=0)
print(spec)
```

You will see the crash. Here is a traceback that I've seen:
```
  File "/Users/hyerra/Desktop/test.py", line 27, in <module>
    print(spec)
  File "/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/torchrl/data/tensor_specs.py", line 3165, in __repr__
    [indent(f"{k}: {repr(item)}", 4 * " ") for k, item in self.items()]
  File "/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/torchrl/data/tensor_specs.py", line 3165, in <listcomp>
    [indent(f"{k}: {repr(item)}", 4 * " ") for k, item in self.items()]
  File "/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/torchrl/data/tensor_specs.py", line 3139, in items
    yield key, self[key]
  File "/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/torchrl/data/tensor_specs.py", line 760, in __getitem__
    [composite_spec[item] for composite_spec in self._specs], dim=self.dim
  File "/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/torchrl/data/tensor_specs.py", line 760, in <listcomp>
    [composite_spec[item] for composite_spec in self._specs], dim=self.dim
  File "/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/torchrl/data/tensor_specs.py", line 2624, in __getitem__
    return self._specs[idx]
KeyError: 'observation_0'
```

## Expected behavior

I should be able to print the spec, especially if it was created successfully.

## System info

```python
import torchrl, numpy, sys
```

None 1.24.3 3.10.11 (main, Apr  7 2023, 07:24:53) [Clang 14.0.0 (clang-1400.0.29.202)] darwin

## Checklist

matteobettini:
This PR will be a superset of #1365, which should be closed.
It tackles tasks from #969
Together with #1391 it becomes a superset of #1204 which should be eventually closed

It fixes #1343

Methods missing are due to time constraint and we can implement them in the future on a need basis

Fixes operations in lazy stack composite specs:

- [X]  `unsqueeze`
- [x]  `squeeze`
- [x]  `print` (it behaves like LazyStackedTd)
- [x] ` __len__`
- [x]  `__eq__`
- [x]  `__ne__`
- [x]  `project`
- [x]  `type_check`
- [x]  `__delitem__`
- [x]  `__iter__`
- [x]  `keys`
- [x]  `__setitem__`
- [x] `update`
- [x] `is_in`
- [x] `unbind`

Missing:
- [ ] `encode` 

Bonus, some operations in lazy stacked specs:
- [X] `shape`
- [X] `stack only if same spec type, dtype and ndims`
- [X] `print`
- [X] `__len__`
- [x] `__eq__`
- [x] `squeeze`
- [x] `unsqueeze`
- [x] `unbind`

Missing:
- [ ] `is_in`
- [ ] `project`
- [ ]  `type_check`
- [ ]  `encode`


matteobettini:
@vmoens  for print of het dicts what about we first print all common, keys and then something like 
```
LazyStackedCompositeSpec(
    camera: BoundedTensorSpec(
        shape=torch.Size([3, 32, 32, 3]),
        space=ContinuousBox(
            minimum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True), 
            maximum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True)),
        device=cpu,
        dtype=torch.float32,
        domain=continuous),
    vector: LazyStackedTensorSpec(
         shape=torch.Size([3, -1]), space=None, device=cpu, dtype=torch.float32, domain=continuous), device=cpu, shape=torch.Size([3]),
    
   lazy_keys={
        0 -> lidar: BoundedTensorSpec(
                  shape=torch.Size([3, 32, 32, 3]),
                  space=ContinuousBox(
                      minimum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True), 
                      maximum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True)),
                  device=cpu,
                  dtype=torch.float32,
                  domain=continuous),
              obs_0: BoundedTensorSpec(
                  shape=torch.Size([3, 32, 32, 3]),
                  space=ContinuousBox(
                      minimum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True), 
                      maximum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True)),
                  device=cpu,
                  dtype=torch.float32,
                  domain=continuous),
        1 -> lidar: BoundedTensorSpec(
                  shape=torch.Size([3, 32, 32, 3]),
                  space=ContinuousBox(
                      minimum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True), 
                      maximum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True)),
                  device=cpu,
                  dtype=torch.float32,
                  domain=continuous),
              obs_1: BoundedTensorSpec(
                  shape=torch.Size([3, 32, 32, 3]),
                  space=ContinuousBox(
                      minimum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True), 
                      maximum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True)),
                  device=cpu,
                  dtype=torch.float32,
                  domain=continuous),
        2 -> sonar: BoundedTensorSpec(
                  shape=torch.Size([3, 32, 32, 3]),
                  space=ContinuousBox(
                      minimum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True), 
                      maximum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True)),
                  device=cpu,
                  dtype=torch.float32,
                  domain=continuous),
              obs_2: BoundedTensorSpec(
                  shape=torch.Size([3, 32, 32, 3]),
                  space=ContinuousBox(
                      minimum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True), 
                      maximum=Tensor(shape=torch.Size([3, 32, 32, 3]), device=cpu, dtype=torch.float32, contiguous=True)),
                  device=cpu,
                  dtype=torch.float32,
                  domain=continuous),
    }

 device=cpu, shape=torch.Size([3]))
```

vmoens:
Happy with the new print!

torchrl/data/tensor_specs.py
```diff
@@ -897,6 +863,71 @@ def rand(self, shape=None) -> TensorDictBase:
     def to(self, dest: Union[torch.dtype, DEVICE_TYPING]) -> T:
         return torch.stack([spec.to(dest) for spec in self._specs], self.dim)
 
+    def unbind(self, dim: int):
+        if dim == self.stack_dim:
+            return self._specs
+        shape = self.shape
+        if dim < 0 or dim > self.ndim - 1 or shape[dim] == -1:
+            raise ValueError(f"Provided dim is not valid for unbinding shape {shape}")
```
vmoens, line 871:
```suggestion
            raise ValueError(f"Provided dim {dim} is not valid for unbinding shape {shape}")
```

torchrl/data/tensor_specs.py
```diff
@@ -897,6 +863,71 @@ def rand(self, shape=None) -> TensorDictBase:
...
+        else:
+            raise ValueError("Can only unbind in stack dim")
```
vmoens, line 873:
```suggestion
            raise ValueError(f"A {type(self)} instance can only be unbound along its stack dimension. Expected {self.stack_dim}, received {dim} instead.")
```

torchrl/data/tensor_specs.py
```diff
@@ -2908,6 +2996,17 @@ def clone(self) -> CompositeSpec:
             shape=self.shape,
         )
 
+    def empty(self):
```
vmoens, line 2999: Don't we need a docstring here (or is it somewhere else?)

torchrl/data/tensor_specs.py
```diff
@@ -3135,62 +3251,146 @@ def to_numpy(self, val: TensorDict, safe: bool = None) -> dict:
...
     def encode(
         self, vals: Dict[str, Any], ignore_device: bool = False
     ) -> Dict[str, torch.Tensor]:
-        pass
+        raise NotImplementedError
 
-    def __delitem__(self, key):
-        pass
+    def __delitem__(self, key: NestedKey):
+        """Deletes a key only if present in all stacked specs."""
```
vmoens, line 3365:
```suggestion
        """Deletes a key from the stacked composite spec.
        
         This method will be executed if and only if the key is present in all stacked specs.
         """
```

torchrl/data/tensor_specs.py
```diff
@@ -3135,62 +3251,146 @@ def to_numpy(self, val: TensorDict, safe: bool = None) -> dict:
...
+        for spec in self._specs:
+            try:
+                del spec[key]
+                at_least_one_deletion = True
+            except KeyError:
+                continue
+        if not at_least_one_deletion:
+            raise KeyError(
+                f"Key {key} must be present in at least one of the stacked specs"
+            )
```
vmoens, line 3376: this is problematic, because you raise an exception but do not repopulate the spec.
If I do

```
try:
    del spec[some_partial_key]
except:
    pass
```
I will end up with a partially depopulated spec

torchrl/data/__init__.py
```diff
@@ -46,3 +46,4 @@
     UnboundedContinuousTensorSpec,
     UnboundedDiscreteTensorSpec,
 )
+from .utils import check_no_exclusive_keys, consolidate_spec
```
vmoens, line 49: avoid local imports

torchrl/data/utils.py
```diff
@@ -35,6 +42,98 @@
 INDEX_TYPING = Union[None, int, slice, str, Tensor, List[Any], Tuple[Any, ...]]
 
 
+def consolidate_spec(
+    spec: TensorSpec,
+    recurse_through_entries: bool = True,
+    recurse_through_stack: bool = True,
+):
+    """Given a TensorSpec, removes exclusive keys by adding 0 shaped specs."""
```
vmoens, line 50: Let's add a list of Args and kwargs

torchrl/data/utils.py
```diff
@@ -35,6 +42,98 @@
 INDEX_TYPING = Union[None, int, slice, str, Tensor, List[Any], Tuple[Any, ...]]
 
 
+def consolidate_spec(
+    spec: TensorSpec,
```
vmoens, line 46: Shouldn't be CompositeSpec?

torchrl/data/utils.py
```diff
@@ -35,6 +42,98 @@
...
+    elif isinstance(spec, LazyStackedTensorSpec):
+        shape = list(shape[: spec.stack_dim]) + list(shape[spec.stack_dim + 1 :])
+        return torch.stack(
+            [_empty_like_spec(sub_spec, shape) for sub_spec in spec._specs],
+            spec.stack_dim,
+        )
+    else:
+        spec_shape = spec.shape
+        shape = [dim if i < len(shape) else 0 for i, dim in enumerate(spec_shape)]
+        spec = spec[(0,) * len(spec_shape)]
```
vmoens, line 113: All dims are 0?
Why not just the first one, or the heterogeneous one?
Might be too much work or too opinionated but for a het tensor implementation having less het dimensions is always better

test/test_specs.py
```diff
@@ -2648,6 +3095,22 @@ def test_composite_contains():
     assert ("a", ("b", ("c",))) in spec.keys(True, True)
 
 
+def get_all_keys(spec: TensorSpec, include_exclusive: bool):
+    """Given a TensorSpec, returns all exclusive and non-exclusive keys as a set tuples."""
```
vmoens, line 3099: Let's add args and kwargs

vmoens (commented):
Great work!
Aside from the few comments I left, I noticed that the new classes are not part of the doc.

torchrl/data/__init__.py
matteobettini: Screenshot 2023-07-27 at 17 16 04
wait so this is also all wrong?

torchrl/data/tensor_specs.py
matteobettini: wait but I only raise the exception if no deletion has occurred.
` if not at_least_one_deletion:` will mean nothing was touched

torchrl/data/utils.py
```diff
@@ -35,6 +42,98 @@
...
+                spec._specs[spec_index] = sub_spec
+            for sub_spec_key in sub_spec.keys():
+                if sub_spec_key not in keys:  # lazy key
+                    lazy_keys_per_spec[spec_index].add(sub_spec_key)
+                    if sub_spec_key not in lazy_keys_examples:
+                        value = sub_spec[sub_spec_key]
+                        lazy_keys_examples.update({sub_spec_key: value})
+
+        for spec_index in range(len(spec._specs)):  # add missing exclusive entries
+            sub_spec = spec._specs[spec_index]
```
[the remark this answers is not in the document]
matteobettini: - it is not items because here i am cycling through the list, not the values. there are still missing keys which will not be present in keys(). the values are checked later when we have made all keys non-exclusive.
- we are indexing with integers because i want to `spec._specs[spec_index] = sub_spec`

test/test_specs.py
matteobettini: What do you mean precisely? do you want me to add the two *args and **kwargs arguments in the function signature? and what would they do?

torchrl/data/__init__.py
vmoens: Oh sorry in `__init__` it's ok nevermind
in other files I try to avoid them -- it crashes when running doctests

torchrl/data/tensor_specs.py
vmoens: Got it you're right, that should be ok
but the docstrings were misleading (from what I read it seemed to indicate that the key had to be everywhere)

torchrl/data/utils.py
vmoens: > if it was the first one: why the first one?

could be convention motivated by efficiency purposes

> if it is the het: what are the het here? this key does not exist, we are creating it. all the features are het

more like:
If all dims of the tensors that do exist are the same, take the first. If at least one tensor has a different shape along one dim, take that one instead. That heuristic minimises the number of heterogeneous dims and maximises efficiency. I think taking the first dim has little side effects (that I can see) since at the end of the day the `numel()` is still 0. Having all 0s will be cause major computational burden on the het tensors if we go down the road we're on...

torchrl/data/tensor_specs.py
matteobettini: oh you are right, the docstirng if very wrong, thanks!

torchrl/data/utils.py
matteobettini: when thinking of the semantics though, in the example above, you would give the user
```
 "obs": torch.tensor(2,4,-1,6, 7)
```
 imagine that we have a shape filter (the one we discussed) where we ask for shape (..., 7) for a neural network, then this filter will not filter out (2,4,0,6,7). 

 i think it is too opinionated to choose to keep 6 and 7 and not 5. and it will be ultra confusing for users.

Isn't it possible to work around the computational burden in het tensor?

vmoens: that makes things much more complex. I do really think that the lower the number of heterogeneous shapes the better. 
Wasn't the idea of the filter was to select the shapes such that they could be stacked: if a tensor in a list has 0 elements it can't be stacked. If the idea is to cat it can be cat. I don't think that this is a blocker.
Using the first dim is also convenient as the trailing dimensions are more restrictive for most usages (eg linear or conv layers).
To me making more dims than needed 0 has the risk of hiding things: if I have 3 tensordicts the first 2 with key "foo" and value shapes [1, 2], [1, 3] and no "foo" for the third, users will see a shape [-1, -1] for "foo" when (to me) [1, -1] would be more reasonable. If you call `filter(..., shape=[3])` you'll get the second value.
In the case of `[3, 1]`, `[2, 1]` and nothing, `filter(..., shape=[3, 1])` will also work seamlessly.

matteobettini: ok what about all 0s when there are no het dims (instead of arbitrarily choosing the first) and only 0s in the het dims if there are?

cc @acciorocketships can I also get what you think?

vmoens: Oh yeah that seems reasonable enough

matteobettini: on it

vmoens (approved):
LGTM
Do the boxes that are unticked raising errors when called?

vmoens (approved):
LGTM
Do the boxes that are unticked raising errors when called?

matteobettini:
> LGTM Do the boxes that are unticked raising errors when called?

i ll find out and fix if needed

vmoens (approved):
LGTM Thanks!

PR #1395, opened 2023-07-19
[Fix] Remove loss device

### LINKED ISSUE #1389: [BUG] Device check in DQN always takes first branch
in the DQN forward we have this code to check the device 
```
  if self.device is not None:
      warnings.warn(
          "The use of a device for the objective function will soon be deprecated",
          category=DeprecationWarning,
      )
      device = self.device
  else:
      device = tensordict.device
```

The problem is that `loss.device` is

```
  @property
  def device(self) -> torch.device:
      for p in self.parameters():
          return p.device
      return torch.device("cpu")
```

which will never return None, so the first branch is always taken and the warning always triggered.

So this logic has never made any sense.

How should we sort this out?

matteobettini:
closes #1389

### DIFF
--- a/torchrl/objectives/a2c.py
+++ b/torchrl/objectives/a2c.py
@@ -232,12 +232,14 @@ class A2CLoss(LossModule):
         self.convert_to_functional(critic, "critic", compare_against=policy_params)
         self.samples_mc_entropy = samples_mc_entropy
         self.entropy_bonus = entropy_bonus and entropy_coef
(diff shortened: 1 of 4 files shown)


vmoens (approved):
LGTM
Should we remove loss.device in a separate PR?

matteobettini:
i gave it a shot

vmoens (approved):
LGTM thanks for this!

PR #1398, opened 2023-07-19
[Features] Conv3dNet and PermuteTransform

xmaples:
## Description

Add features of `Conv3dNet`, similar to ConvNet but with 3d convolution layers, and `PermuteTransform`  being a utility tool to transform tensors to ones with shape required by conv3d.

## Types of changes

- [x] Example (in doc)
- [x] Tests.


torchrl/modules/models/__init__.py
```diff
@@ -8,6 +8,8 @@
 from .model_based import DreamerActor, ObsDecoder, ObsEncoder, RSSMPosterior, RSSMPrior
 from .models import (
     ConvNet,
+    Conv2dNet,
+    Conv3dNet,
```
vmoens, line 12: you will also want to add them to `modules/__init__.py`

torchrl/modules/models/models.py
```diff
@@ -481,6 +482,223 @@ def forward(self, inputs: torch.Tensor) -> torch.Tensor:
         return out
 
 
+Conv2dNet=ConvNet
+
+class Conv3dNet(nn.Sequential):
+    """A 3D-convolutional neural network.
+
+    Args:
+        in_features (int, optional): number of input features;
```
vmoens, line 491:
```suggestion
        in_features (int, optional): number of input features. A lazy implementation that automatically retrieves
            the input size will be used if none is provided.
```

torchrl/modules/models/models.py
```diff
@@ -481,6 +482,223 @@ def forward(self, inputs: torch.Tensor) -> torch.Tensor:
...
+        norm_class: Optional[Type[nn.Module]] = None,
+        norm_kwargs: Optional[dict] = None,
+        bias_last_layer: bool = True,
+        aggregator_class: Optional[Type[nn.Module]] = SquashDims,
+        aggregator_kwargs: Optional[dict] = None,
+        squeeze_output: bool = False,
+        device: Optional[DEVICE_TYPING] = None,
+    ):
+        if num_cells is None:
+            num_cells = [32, 32, 32]
```
vmoens, line 588: Perhaps
```suggestion
        if num_cells is None:
            if depth is None:
                num_cells = [32, 32, 32]
            else:
                num_cells = [32] * depth
```

torchrl/modules/models/models.py
```diff
@@ -481,6 +482,223 @@ def forward(self, inputs: torch.Tensor) -> torch.Tensor:
...
+        in_features (int, optional): number of input features;
+        depth (int, optional): depth of the network. A depth of 1 will produce a single linear layer network with the
+            desired input size, and with an output size equal to the last element of the num_cells argument.
+            If no depth is indicated, the depth information should be contained in the num_cells argument (see below).
+            If num_cells is an iterable and depth is indicated, both should match: len(num_cells) must be equal to
+            the depth.
+        num_cells (int or Sequence[int], optional): number of cells of every layer in between the input and output. If
+            an integer is provided, every layer will have the same number of cells. If an iterable is provided,
+            the linear layers out_features will match the content of num_cells.
+            default: [32, 32, 32];
```
vmoens, line 500:
```suggestion
            default: ``[32, 32, 32]`` or ``[32] * depth` is depth is not ``None``.
```

torchrl/modules/models/models.py
```diff
@@ -481,6 +482,223 @@ def forward(self, inputs: torch.Tensor) -> torch.Tensor:
...
+                    self.aggregator_class, device, **self.aggregator_kwargs
+                )
+            )
+
+        if self.squeeze_output:
+            layers.append(SqueezeLayer((-3,-2,-1)))
+        return layers
+
+    def forward(self, inputs: torch.Tensor) -> torch.Tensor:
+        *batch, C, D, L, W = inputs.shape
```
vmoens, line 694:
```suggestion
        try:
            *batch, C, D, L, W = inputs.shape
        except ValueError as err:
            raise ValueError(f"The input value of {self.__class__.__name__} must have at least 4 dimensions, got {inputs.ndim} instead.") from err
```

test/test_transforms.py
```diff
@@ -8027,6 +8028,14 @@ def test_transform_rb(self, rbclass):
...
         raise pytest.skip("No inverse for KLRewardTransform")
 
+class TestPermuteTransform:
+    @pytest.mark.parametrize("batch", [[], [2], [2, 4]])
+    def test_transform_no_env(self, batch):
+        D, W, H, C= 8, 32, 64, 3
+        trans=PermuteTransform(dims=(-1, -4,-2, -3), in_keys=["pixels"],) # DxWxHxC => CxDxHxW
+        td=TensorDict({"pixels": torch.randn((*batch, D, W, H, C))}, batch_size=batch)
+        td=trans(td)
+        assert td["pixels"].shape == torch.Size((*batch, C, D, H, W))
```
vmoens, line 8038: Better to use the TransformBase template to make sure that everything works properly. 
You will need an env to check that the env specs match what is expected.
We can use a conv mocking env perhaps (`DiscreteActionConvMockEnv`)?
Feel free to skip the tests that are not relevant to that class.

test/test_modules.py
```diff
@@ -174,6 +174,71 @@ def test_convnet(
...
+        aggregator_class=aggregator_class,
+        aggregator_kwargs=aggregator_kwargs,
+        squeeze_output=squeeze_output,
+        device=device,
+    )
+    if in_features is None:
+        in_features = 5
+    x = torch.randn(*batch, in_features, input_size, input_size, input_size, device=device)
+    y = convnet(x)
+    assert y.shape == torch.Size([*batch, expected_features])
```
vmoens, line 241: Let's also check that the error we want is raised if the input does not have enough dim?

test/test_modules.py
```diff
@@ -174,6 +174,71 @@ def test_convnet(
     y = convnet(x)
     assert y.shape == torch.Size([*batch, expected_features])
 
+@pytest.mark.parametrize("in_features", [3, 10, None])
+@pytest.mark.parametrize(
+    "input_size, depth, num_cells, kernel_sizes, strides, paddings, expected_features",
+    [(10, None, None, 3, 1, 0, 32 * 4 * 4 * 4), (10, 3, 32, 3, 1, 1, 32 * 10 * 10 * 10)],
```
vmoens, line 180: This seems pretty "big" and I wonder how much time it will take to allocate these tensors.
Do you thing we can reduce the size (32 -> 4 or something like it?)

vmoens (commented):
Great work! Really useful thanks a mil!
A few suggestions

test/test_modules.py
xmaples: it's ok if testing on `test_convnet` is acceptable, cos tensors in test_conv3dnet  are smaller than in test_convnet.

torchrl/modules/models/models.py
xmaples: nice

test/test_modules.py
xmaples: Good idea

test/test_transforms.py
xmaples: I've committed new testing codes with TransformBase template. And it's unable to run test since the initialization of torchrl (import torchrl) is failed for the sources on branch main, with error message:  "ModuleNotFoundError: No module named 'torchrl._torchrl'"

vmoens:
@xmaples I made [a PR](https://github.com/xmaples/pytorchrl/pull/3) against this branch. I think you can safely merge it and start from there. There should be very little to do once the patch is merged!

vmoens (approved):

PR #1409, opened 2023-07-24
[Feature] consolidate td

matteobettini:


torchrl/data/utils.py
```diff
@@ -35,6 +43,206 @@
...
+            else:
+                td.set(
+                    key,
+                    value,
+                )
+        except RuntimeError as err:
+            if re.match(r"Found more than one unique shape in the tensors", str(err)):
+                pass
+            else:
+                raise err
```
vmoens, line 153: capturing errors at runtime is costly, if this is expected to run once in a while we should find another way of doing it

torchrl/data/utils.py
```diff
@@ -35,6 +43,206 @@
...
+    return keys
+
+
+def all_eq(
+    td: Union[TensorDictBase, torch.Tensor],
+    other: Union[TensorDictBase, torch.Tensor],
+):
+    """Returns true if the two classes match all entries in the keys and stack dimensions."""
+    if td.__class__ != other.__class__:
+        return False
```
vmoens, line 219: why can't we compare a TensorDict and a LazyStackedTensorDict?

torchrl/data/utils.py
```diff
@@ -35,6 +43,206 @@
...
+
+    if td.shape != other.shape or td.device != other.device:
+        return False
+
+    if isinstance(td, LazyStackedTensorDict):
+        if td.stack_dim != other.stack_dim:
+            return False
+        for t, o in zip(td.tensordicts, other.tensordicts):
```
vmoens, line 227: let's use longer names

torchrl/data/utils.py
```diff
@@ -35,6 +43,206 @@
...
+            if not all_eq(t, o):
+                return False
+    elif isinstance(td, TensorDictBase):
+        td_keys = list(td.keys())
+        other_keys = list(other.keys())
+        if td_keys != other_keys:
```
vmoens, line 233: what if the keys have a different sorting?

torchrl/data/utils.py
```diff
@@ -35,6 +43,206 @@
...
+            return False
+        for k in td_keys:
+            if not all_eq(td[k], other[k]):
+                return False
+    elif isinstance(td, torch.Tensor):
+        return torch.equal(td, other)
+    else:
+        raise AssertionError()
```
vmoens, line 241: let's put an explicit message (also why do we have an exception here?)
Can we use something else than AssertionError?
It's basically the same as `assert` which we try to avoid in the code base

torchrl/data/utils.py
```diff
@@ -35,6 +43,206 @@
...
+                return False
+    elif isinstance(td, torch.Tensor):
+        return True
+    else:
+        return False
+
+    return True
+
+
+def get_all_keys(td, include_lazy: bool):
```
vmoens, line 194: is this something that we will keep in torchrl? My feeling is that it should not be exposed to the users

vmoens (commented):
Can we make all these methods private? It will block us later if we want to deprecate lazy stacked td

torchrl/data/utils.py
matteobettini: i didn't put it in tensordict because of the planned deprecation of lazy keys.
it was here cause i thought we might want to use it in a few places, but now i have moved it to the tests

torchrl/data/utils.py
matteobettini: the idea was that the 2 td we are comparing should have exactly identical structure.
happy to change it if you want

torchrl/data/utils.py
matteobettini: solved it, we might consider updating this too https://github.com/pytorch-labs/tensordict/blob/main/tensordict/tensordict.py#L8009 it is run at every print

torchrl/data/utils.py
matteobettini: i also made it an arg

test/_utils_internal.py
```diff
@@ -320,3 +319,85 @@ class MyClass:
     for key in td.keys():
         MyClass.__annotations__[key] = torch.Tensor
     return tensorclass(MyClass)
+
+
+class TestUtilsTd:
```
matteobettini, line 324: i move all the functions i use in the test here so they are hidden

torchrl/data/utils.py
```diff
@@ -35,6 +37,86 @@
 INDEX_TYPING = Union[None, int, slice, str, Tensor, List[Any], Tuple[Any, ...]]
 
 
+def _consolidate_entries_td(
```
matteobettini, line 40: i kept it for working with TensorDictBase for now because I think that coding this function now to work with lists of TensorDictBases would be have uneeded complexity to handle the various nesting possibbilities (interleaving of lists and LazyStacks).

I think it will be easier to code this with lists of tensordicts once we know there cannot be lazy stack entries.

PR #1411, opened 2023-07-25
[Feature] Heterogeneous Environments compatibility

matteobettini:
This introduces tests for het envs and key modifications to

- rollout
- step_mdp


torchrl/data/utils.py
```diff
@@ -42,6 +43,36 @@
 INDEX_TYPING = Union[None, int, slice, str, Tensor, List[Any], Tuple[Any, ...]]
 
 
+def dense_stack_tds(
+    td_list: Union[typing.Sequence[TensorDictBase], LazyStackedTensorDict],
+    stack_dim: int = None,
+) -> TensorDictBase:
+    """Densely stack a list of TensorDictBase objects (or a LazyStackedTensorDict) given that they have the same structure.
+
+    This must be used when some of the tds involved can have LazyTds among keys (or keys of keys, recursively).
```
vmoens, line 52: let's avoid abbreviations in docstrings.
Even better: we can link to tensordict classes by using 
```
:class:`tensordict.stuff`
```

torchrl/data/utils.py
```diff
@@ -42,6 +43,36 @@
...
+    In those cases, calling `torch.stack(td_list).to_tensordict()` is infeasible.
+    Thus, this function provides an alternative for densifying the list provided.
+
+    Args:
```
vmoens, line 56: I think an example for this method would be helpful.

vmoens (commented):
Thanks for this, really useful

torchrl/data/utils.py
```diff
@@ -42,6 +43,36 @@
 INDEX_TYPING = Union[None, int, slice, str, Tensor, List[Any], Tuple[Any, ...]]
 
 
+def dense_stack_tds(
```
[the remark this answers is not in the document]
matteobettini: I'll write some tests and examples for it.

I am down to move it to tensodict, I put it here because I thought we said we wanted the utis to be here for the deprecation move

matteobettini:
So i have updated and added all the requested changes.
my takes on the new function are

- it should be public because any user which has some nested lazy stacks somewhere should use it after calling `env.rollout(return_contiguous=False)`
- my personal preference is that it should be in tensordict utils but the choice is yours

vmoens:
> So i have updated and added all the requested changes. my takes on the new function are
> 
> * it should be public because any user which has some nested lazy stacks somewhere should use it after calling `env.rollout(return_contiguous=False)`
> * my personal preference is that it should be in tensordict utils but the choice is yours

let's put it in tensordict then!

torchrl/envs/utils.py
```diff
@@ -219,44 +219,63 @@ def _set_single_key(source, dest, key, clone=False):
...
     total_key = total_key + (key,)
     non_empty = False
     if unravel_key(total_key) not in excluded:
-        val = source.get(key)
-        if is_tensor_collection(val):
-            new_val = dest.get(key, None)
-            if new_val is None:
-                new_val = val.empty()
+        shape = source.get_item_shape(key)
+        if -1 not in shape:
```
vmoens, line 250: what's the overhead here?
Since this will likely be very frequent I'm a bit worried by increasing the control flow for every key

vmoens: For a regular td, this op means that we'll be accessing the tensor twice through get (once to get its shape, a second time to get it). Accessing it once should suffice.
Not I understand that we can't use `get` if the tensor is heterogeneous... A try/except may not be the right solution either since it'll be slow for heterogeneous envs. However regular envs won't be impacted

matteobettini: I am not sure what I can do here
- in the past i was using try blocks, but you made me move away from them
- an isinstance wont work since the key might be het even if the container is a regular td

so I can move back to try except if you prefer

matteobettini: lemme switch to try blocks

ideally tho this problem could be solved once and for all in tensordict, with an efficient way to understand if a key is het

vmoens: yep that's what I'm thinking too
If we have tensorstack we can even leave the block with try / except and it won't break (I guess)

vmoens: Just make sure you comment the try / excpet saying that it's temporary and for efficiency purpose

matteobettini:
updated

vmoens (approved):
_**Het**_ groene licht is gekomen

PR #1416, opened 2023-07-26
[CI] Fix tuto deps

matteobettini:
The vmas dep was missing in the docs requirement


matteobettini:
yep i downloaded the artifact and this fixed it

vmoens (approved):

PR #1419, opened 2023-07-27
[Feature] VMAS obs dict

matteobettini:
VMAS has recently introduced the possibility of agents having dicts of tensors as observations (reminds me of a certain class...) https://github.com/proroklab/VectorizedMultiAgentSimulator/pull/49

this pr is a simple change to allow them to be converted to that certain class


vmoens (approved):
LGTM

image

Issue #1404, opened 2023-07-21
[Feature Request] Action Masking


Kang-SungKu:
## Motivation
I recently started learning TorchRL, and creating a custom environment (using torchrl.envs.EnvBase) based on the documentation (https://pytorch.org/rl/reference/envs.html). For my environment, I would like to apply an action mask, such that the environment does not allow infeasible action based on the observation (for example, suppose the action is to choose a trump card, the number of `A` is limited, such that it cannot be chosen once all the `A`'s are drawn). So far, I could not find a way to implement action masking for the environment, so it would be helpful if there is a recommended practice to implement it.

## Solution
It would be convenient if I can include a mask as a part of the `observation_spec`, such that the environment can tell feasible/infeasible actions based on the observation (even when a random action is chosen). Currently, my environment cannot pass `torchrl.envs.utils.check_env_specs()` since infeasible actions are chosen.

If it is not reasonable, any alternative way to implement an action mask is appreciated.

## Checklist
I searched with the keyword 'Action Masking', but could not find relevant issues. Sorry if I missed something.

- [ X ] I have checked that there is no similar issue in the repo (**required**)

1030852813:
Please! I do need this feature so bad. Could anyone provide any RL library that supports action space?

smorad:
You could do this in a `TensorDictModule`.

```python
class ActionMask(TensorDictModule):
  ...
  def forward(self, td):
    if td['observation'] == 1:
      td['action'][0] = 0

policy = Sequential(EGreedyWrapper(Sequential(mlp, QValueModule())), ActionMask())
```

Kang-SungKu:
@1030852813 , TensorFlow Agents supports action masking. It allows to pass mask to their `policy` as the parameter named `observation_and_action_constraint_splitter`, and policies implement action masking in different ways, for example:
- `QPolicy` applies a mask such that the q-values of infeasible actions are assumed to be negative infinite
- `RandomTFPolicy` utilizes a mask to create a `MaskedCategorical` such that the action is sampled from it. It seems like torchrl also has `MaskedCategorical` (https://pytorch.org/rl/reference/generated/torchrl.modules.MaskedCategorical.html#torchrl.modules.MaskedCategorical) based on the same class.

The good thing about their implementation is that the policies are designed to use the same form of `observation_and_action_constraint_splitter`. I am considering several options based on what @smorad mentioned to see what is reasonable.

vmoens:
[...]

PR #1421, opened 2023-07-27
[Feature] Masking actions

vmoens:
## Description

Addresses #1404 

A quick example of the functionality
https://gist.github.com/vmoens/95d6427fcb5fa5714291b3dbfa7daa15#file-action_mask-ipynb

TODO:
- [ ] DQN compatibility
- [x] ActionMask tests
  - [x] focus on reset
  - [x] focus on exceptions
  - [x] focus on compose (do we change a clone or the actual spec?). what about nested compose?
- [x] test exceptions of masking specs


hyerra:
This is awesome, I think it might also be helpful for #1201 since Unity supports masked actions as well.

torchrl/envs/transforms/transforms.py
```diff
@@ -4353,3 +4394,84 @@ def _inv_apply_transform(
...
+        >>> env = TransformedEnv(base_env, ActionMask())
+        >>> r = env.rollout(10)
+        >>> env = TransformedEnv(base_env, ActionMask())
+        >>> r = env.rollout(10)
+        >>> r["mask"]
+
+    """
+
+    def __init__(self, action_key="action", mask_key="mask"):
+        if not isinstance(action_key, str):
```
matteobettini, line 4447: Let's not forget about nested keys

matteobettini:
I also think this is awsome. It will be needed for MLAgents, as @hyerra said, for PettingZoo #1441 and for SMAC #1466

matteobettini:
A few things to be aware of:

- this works well when you have to mask some of the actions but some are still available. how do we treat the case when all actions are unavailable 9eg an agent dies)
- check_env_specs needs also to be adapted, because it will try to check that the last action is in the current action spec, but the mask of the current action spec has already been updated after the last step.

vmoens:
> A few things to be aware of:
> 
> 
> 
> - this works well when you have to mask some of the actions but some are still available. how do we treat the case when all actions are unavailable (eg an agent dies)?

Shouldn't it be `done` then?
If we think about other situations, I'm not sure it makes sense to cover this explicitly. Example: if you're playing a game and you have N actions available but when your turn comes you don't have anything to do, there should be one more action N+1 "pass".

matteobettini:
> Example: if you're playing a game and you have N actions available but when your turn comes you don't have anything to do, there should be one more action N+1 "pass".

I m talking about a different scenario.
For example, you are playing a game, but only one player acts at a time. It is not your turn, but you are also not done.

What i am doing right now is that i am setting a  mask of all false for all actions of that agent, but not passing that to the spec (as it would crash), just having it in the env output. So the spec will still see all actions available for that agent, but in reality that whole transition for it will be masked out.

vmoens:
> Maybe passing a mask of all false should be allowed? And all specs should have mask compatibility?

Then what should `action_spec.rand()` return? It's gonna be confusing no?
IMO if your env has the option of having an action "do nothing" this is something that should be in the env otherwise it will be surprising.
In this case, we can do a transform:
```python
class SkipAction(Transform):
    """Adds an action to the env that skips the step if selected.
    """
```
For instance, if `action_spec` has 6 possible actions, an action with value `6` will instruct the transform to skip the step.
A bit of magic will be needed to make it work with composite actions...

matteobettini:
ok yeah I see.
we can keep this PR to the current scope then

matteobettini (approved):
LGTM

This does not work for multi discrete and multi one hot right?
I think it would be useful to support that

matteobettini:
Also we need to add this line to check_env_specs, because otherwise the check will fail when the action_mask gets updated
```
# Check specs
last_td = real_tensordict[..., -1]
# replace action with one sampled from last spec
# (since spec mask might be changed after last step)
last_td = env.rand_action(last_td) ## <------ this line
full_action_spec = env.input_spec["full_action_spec"]
full_state_spec = env.input_spec["full_state_spec"]
```

matteobettini (approved):
LGTM

are we leaving the multi spec for another PR?

vmoens:
> LGTM
> 
> are we leaving the multi spec for another PR?

you mean multi-one-hot (ie in a single tensor) or multiple masks?

matteobettini:
Multi one hot and multi discrete

vmoens:
@matteobettini now these are supported too

Issue #1407, opened 2023-07-23
 [BUG]  The actor_params attribute is lost when passing a Loss module to a multiprocessing process


albertbou92:
## Describe the bug

When passing a Loss module to another mp process, the actor_params attribute seems to be lost.

The problem is not present when deepcopying the module or when pickling and unpickling it.

## To Reproduce

Steps to reproduce the behavior.

```python
import sys
import pickle
from copy import deepcopy
from torch import nn
from torch import multiprocessing as mp
from tensordict.nn import TensorDictModule
from torchrl.objectives import PPOLoss
from torchrl.envs.libs.gym import GymEnv
from torchrl.collectors import SyncDataCollector
from torchrl.objectives.value.advantages import GAE
from torchrl.modules import OneHotCategorical, ProbabilisticActor


def run_collector(mp_collector, mp_advantage, mp_loss,  qout):
    for batch in mp_collector:
        try:
            batch = mp_advantage(batch)
            loss = mp_loss(batch)
            qout.put("succeeded")
            print('succeeded')
        except AttributeError:  #  AttributeError: 'PPOLoss' object has no attribute 'actor_params'
            qout.put("failed")
            print('failed')


if __name__ == "__main__":

    def env_maker():
        return GymEnv("CartPole-v1", device="cpu")

    test_env = env_maker()
    action_spec = test_env.action_spec
    obs_spec = test_env.observation_spec
    n = action_spec.space.n
    actor = ProbabilisticActor(
        module=TensorDictModule(nn.Linear(4, n), in_keys=["observation"], out_keys=["logits"]),
        in_keys=["logits"],
        out_keys=["action"],
        distribution_class=OneHotCategorical,
        return_log_prob=True,
        spec=action_spec)
    critic = TensorDictModule(nn.Linear(4, 1), in_keys=["observation"], out_keys=["state_value"])

    collector = SyncDataCollector(
        env_maker,
        actor,
        total_frames=200,
        frames_per_batch=200,
    )

    advantage_module = GAE(
        gamma=0.99,
        lmbda=0.95,
        value_network=critic,
        average_gae=True,
        advantage_key="advantage",
    )
    loss_module = PPOLoss(actor=actor, critic=critic)

    # The attribute actor_params is present here
    assert hasattr(loss_module, "actor_params")

    # 1. The attribute is still present after deepcopying the loss_module
    deepcopy_loss_module = deepcopy(loss_module)
    assert hasattr(deepcopy_loss_module, "actor_params")

[...]

PR #1424, opened 2023-07-28
[Refactor] Change objectives parameter/buffer/target logic

vmoens:
## Description

Aims at solving #1407


vmoens:
The last outstanding issues with this PR are
- the dtype and device casting
- the call to loss_module.requires_grad_(True) which will mess up the target params

We need target params that are not property (for within algo module repurposing).
We can take the data of the param and register it within a different nn.Parameter but then calling to(smth) will unbind the param and the target. Augmenting `module.to` will not solve the issue when the module is registered within another one, as the calls to `to` are not recursive. The previous solution with the pre-get hook did not work well either and was a bit brittle.

albertbou92 (approved):

PR #1427, opened 2023-07-28
[BugFix] Fix LSTM - VecEnv compatibility

### LINKED ISSUE #1406: [BUG] LSTMModule sets hidden state to 0.0 at every step when combined with ParallelEnv.
## Describe the bug

While trying to use an actor network that contains an LSTMModule together with a vectorised environment, I detected the following:
 - When the vectorised environment is created using the SerialEnv class everything works fine.
 - When the vectorised environment is created using the ParallelEnv, the LSTM hidden states are not passed forward from step to step and are always 0.0. 
I suspect somewhere in the ParallelEnv the hidden states are being reset at every step.

## To Reproduce

Here is the complete code to reproduce the issue. Note that if in the method create_env_fn(), ParallelEnv is replaced by SerialEnv, the code works.

```python

import torch
import tqdm
from torchrl.modules import ProbabilisticActor, LSTMModule, MLP
from torchrl.envs import (
    ParallelEnv,
    SerialEnv,
    TransformedEnv,
    InitTracker,
    StepCounter,
    RewardSum,
)


def create_model(input_size, output_size, hidden_size=256, num_layers=3, out_key="logits"):

    lstm_module = LSTMModule(
        input_size=input_size,
        hidden_size=hidden_size,
        num_layers=num_layers,
        in_key="observation",
        out_key="features",
    )
    mlp = TensorDictModule(
        MLP(
            in_features=hidden_size,
            out_features=output_size,
            num_cells=[],
        ),
        in_keys=["features"],
        out_keys=[out_key],
    )

    return TensorDictSequential(lstm_module, mlp)


def create_rhs_transform(input_size, hidden_size=256, num_layers=3):
    lstm_module = LSTMModule(
        input_size=input_size,
        hidden_size=hidden_size,
        num_layers=num_layers,
        in_key="observation",
        out_key="features",
    )
    return lstm_module.make_tensordict_primer()


def main():

    device = torch.device("cuda:0") if torch.cuda.is_available() else torch.device("cpu")
    test_env = GymEnv("CartPole-v1", device=device, categorical_action_encoding=True)
    action_spec = test_env.action_spec.space
    observation_spec = test_env.observation_spec["observation"]

    def create_transformed_env():
        env = GymEnv("CartPole-v1", device=device, categorical_action_encoding=True)
        env = TransformedEnv(env)
        env.append_transform(create_rhs_transform(input_size=observation_spec.shape[-1]))
        env.append_transform(StepCounter())
        env.append_transform(RewardSum())
        env.append_transform(InitTracker())
        return env

    def create_env_fn(num_workers=4):
        env = ParallelEnv(  # There is some bug here! When using ParallelEnv, hidden states are always zero
        # env = SerialEnv(  # This works!
            create_env_fn=create_transformed_env,
            num_workers=num_workers,
        )
        return env

    # Model
    ##################

    actor_model = create_model(input_size=observation_spec.shape[-1], output_size=action_spec.n)
    actor = ProbabilisticActor(
        module=actor_model,
        in_keys=["logits"],
        out_keys=["action"],
        distribution_class=torch.distributions.Categorical,
        return_log_prob=True,
    )
    actor = actor.to(device)

    # Collector
    ##################

            create_env_fn=create_env_fn,
            policy=actor,
            frames_per_batch=100,
            total_frames=1000,
            device=device,
            storing_device=device,
            max_frames_per_traj=-1,
            split_trajs=False,
        )

    # Collection loop
    ##################

    for data in tqdm.tqdm(collector):

        assert data["recurrent_state_c"].sum() != 0.0  # This is False when I use parallel envs
        assert data["next"]["recurrent_state_c"].sum() != 0.0  # This is False when I use parallel envs

    collector.shutdown()


if __name__ == "__main__":
    main()


```

## Checklist

vmoens:
## Description

Fixes #1406 by ensuring that env.step copies any preallocated "next" value to the next td.

cc @albertbou92


torchrl/envs/transforms/transforms.py
```diff
@@ -2962,6 +2962,27 @@ def transform_observation_spec(
...
+    #                 f"The leading shape of the primer specs ({self.__class__}) should match the one of the parent env. "
+    #                 f"Got state_spec.shape={state_spec.shape} but the '{key}' entry's shape is {spec.shape}."
+    #             )
+    #         try:
+    #             device = state_spec.device
+    #         except RuntimeError:
+    #             device = self.device
+    #         state_spec[key] = spec.to(device)
+    #     return input_spec
+
```
vmoens, line 2985: I'd like to keep this commented here
A bit unsure about what to do with it: per se the idea is that values set by the primer can be inputs but I don't thing it'll systematically be the case. 
An observation can be an input too if it is not cleared.
However one may want to consider these as inputs to the env...

albertbou92 (approved):

vmoens:
@albertbou92 tests are now passing but i'm still not 100% confident about TensorDictPrimer.

First let me explain what TensorDictPrimer is about:
- Collectors allocate buffers to communicate from process to process based on the environment specs. We basically do `env.fake_tensordict()` + some shape expansion and use that as a data container for communication.
- If there is some data that needs to be carried but is not produced by the env but by the policy, we need to pass it too. Therefore, we designed TensorDictPrimer, which adds the missing tensors to the env specs such that `env.fake_tensordict` is complete.
- In our case, we want to add the input and output hidden states to the specs.
- The question is whether we want these to be added to the `input` or `output` spec. By definition, `output` is what the env is producing so that would not make much sense to put it there. Therefore, we should probably put it in `input`. However, when calling `reset` in batched environments, we filter out the data that is not part of `output_spec`. This means that when calling `reset`, we won't see the preallocated tensors from `TensorDictPrimer`. One possible fix would be to also keep the tensors from `input_spec` if they can be found. 

My main concern is basically that this is all a bit hacky: we're telling the env that it must expect tensors as input when, in fact, they have nothing to do with the env. Perhaps we could add a spec to the policy instead and tell collectors to fetch the specs from the policy as well as the env, and combine the two. That way we'd keep things separated.

@smorad @matteobettini wdyt?

albertbou92:
Safe modules are a nice optional feature, but forcing practitioners to use them is not very user-friendly.

Regarding the LSTM hidden states, the TensorDictPrimer transform places them in the TensorDict and you already obtain them after calling reset. During the reset, the environment creates these hidden states, and in every subsequent step, it passes them forward without changing them but still returns them. So, in my view, it's not unreasonable to consider them as outputs.

In fact, same way you can think of output as what the environment produces, you can think of input as what the environment requires to take a step. However, not all the tensors we define in the input_spec fit this definition. For example, when using the RewardSum transform, the episode_reward specs are part of the input_specs, even though they are not something the environment uses to perform steps.

PR #1430, opened 2023-07-31
[Feature] D4rl direct download

MateuszGuzek:
## Description

Allow for a direct download of D4RL datasets, without installing D4RL.

This PR restores the previous PR: https://github.com/pytorch/rl/pull/1376 which was uintentionally closed. 

## Motivation and Context

TBD, raised during direct discussion with @vmoens


vmoens (approved):
I'm very happy with these changes @MateuszGuzek thanks a lot!
The main blocker right now is to document how this works in practice.
Being able to download data directly is great but we need to see what that means for environment execution:

The specific use case I would like to cover is:
```
$ pip install git+https://github.com/Farama-Foundation/d4rl.git
$ pip install gym -U
$ python -c """
# Works with latest gym installed:
from torchrl.data.datasets import D4RLExperienceReplay
data = D4RLExperienceReplay("halfcheetah-medium-v2", batch_size=10, from_env=False)
for batch in data:
    print(batch)
    break
 
# Not sure this makes sense if gym version is above 0.24
import d4rl
import gym
gym.make("halfcheetah-medium-v2")
"""
```

cc @BY571

vmoens:
Update on this:
I tested if updating gym would work but in short: no, it does not (as expected).
So this PR solves the problem of accessing the offline data but the env cannot be created after an upgrade to a more recent gym version. I guess this isn't something we can patch so I'm happy to merge the feature as is!

PR #1431, opened 2023-07-31
[Feature] Optional mapping of "state" in gym specs

### LINKED ISSUE #1418: [Feature Request] Gym spec conversion modifying key names
In the gym to torchrl spec transform https://github.com/pytorch/rl/blob/main/torchrl/envs/libs/gym.py#L181
the key "state" of dict spaces is overwritten with a new name. 

Is it necessary to do it in this location? users might want to transform their gym specs to torchrl without this side effect.

matteobettini:
fixes #1418

### DIFF
--- a/torchrl/envs/libs/gym.py
+++ b/torchrl/envs/libs/gym.py
@@ -179,7 +179,11 @@ __all__ = ["GymWrapper", "GymEnv"]
 
 
 def _gym_to_torchrl_spec_transform(
-    spec, dtype=None, device="cpu", categorical_action_encoding=False
+    spec,
+    dtype=None,
+    device="cpu",
+    categorical_action_encoding=False,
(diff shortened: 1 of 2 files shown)


torchrl/envs/libs/gym.py
```diff
@@ -179,7 +179,11 @@ def gym_backend(submodule=None):
 
 
 def _gym_to_torchrl_spec_transform(
-    spec, dtype=None, device="cpu", categorical_action_encoding=False
+    spec,
+    dtype=None,
+    device="cpu",
+    categorical_action_encoding=False,
+    remap_state_to_observation: bool = True,
```
vmoens, line 186: We need to edit the docstrings to match this

vmoens (approved):

PR #1442, opened 2023-08-02
[Feature] Auto-DoubleToFloat

### LINKED ISSUE #1437: [BUG] DoubleToFloat transform deletes keys
## Describe the bug

The `DoubleToFloat` transform appears to delete all input keys, and will break any environment it is attached to.

## To Reproduce

Steps to reproduce the behavior.

```python
from torchrl.envs import DoubleToFloat, TransformedEnv, Compose
GymEnv("CartPole-v0").reset()
```
```python
TensorDict(
    fields={
        done: Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.bool, is_shared=False),
        observation: Tensor(shape=torch.Size([4]), device=cpu, dtype=torch.float32, is_shared=False)},
    batch_size=torch.Size([]),
    device=cpu,
    is_shared=False)
```
```python
TransformedEnv(GymEnv("CartPole-v0"), Compose(DoubleToFloat())).reset()
```
```python
  File "<stdin>", line 1, in <module>
  File "/Users/smorad/code/from_src/torchrl/torchrl/envs/common.py", line 944, in reset
    tensordict_reset = self._reset(tensordict, **kwargs)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smorad/code/from_src/torchrl/torchrl/envs/transforms/transforms.py", line 655, in _reset
    out_tensordict = self.transform._call(out_tensordict)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smorad/code/from_src/torchrl/torchrl/envs/transforms/transforms.py", line 836, in _call
    tensordict = t._call(tensordict)
                 ^^^^^^^^^^^^^^^^^^^
  File "/Users/smorad/code/from_src/torchrl/torchrl/envs/transforms/transforms.py", line 201, in _call
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

## Expected behavior

Not crash

## System info

master, commit d500488f4e75d482792251725125568751360ccd

## Reason and Possible fixes

If you know or suspect the reason for this bug, paste the code lines and suggest modifications.

## Checklist

vmoens:
## Description

Closes #1437 

As per the discussion below, we also introduce a ``dTypeCastTransform``.
This class is tested through Double2Float: we do this as transform tests take a long time to run and double2float tests cover all functionalities of the parent class. This is not optimal but it saves about a minute of CI runtime.

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -2342,7 +2342,7 @@ class FiniteTensorDictCheck(Transform):
     forward = _call
 
 
-class dTypeCastTransform(Transform):
+class DTypeCastTransform(Transform):
     """Casts one dtype to another for selected keys.
 
     Depending on whether the ``in_keys`` or ``in_keys_inv`` are provided
@@ -2373,7 +2373,7 @@ class dTypeCastTransform(Transform):
         ...     {'obs': torch.ones(1, dtype=torch.double),
         ...     'not_transformed': torch.ones(1, dtype=torch.double),
         ... }, [])
-        >>> transform = dTypeCastTransform(torch.double, torch.float, in_keys=["obs"])
+        >>> transform = DTypeCastTransform(torch.double, torch.float, in_keys=["obs"])
         >>> _ = transform(td)
         >>> print(td.get("obs").dtype)
         torch.float32
@@ -2387,7 +2387,7 @@ class dTypeCastTransform(Transform):
         ...     {'obs': torch.ones(1, dtype=torch.double),
         ...     'not_transformed': torch.ones(1, dtype=torch.double),
         ... }, [])
-        >>> transform = dTypeCastTransform(torch.double, torch.float)
+        >>> transform = DTypeCastTransform(torch.double, torch.float)
         >>> _ = transform(td)
         >>> print(td.get("obs").dtype)
         torch.float32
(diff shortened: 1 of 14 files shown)


vmoens:
Cc @smorad 
Wdyt?

smorad:
LGTM. It might be worth allowing a user-specified `dtype` in case the user wants half precision or something, but I think it's fine either way. At that point maybe it shouldn't be called `DoubleToFloat` anymore 😛

vmoens:
Since this is expected to be more common we can either 
- set these dtypes as default and rename the class
- create a super class and instantiate this as a special case

smorad:
IDK if it's common, but I can imagine some people wanting to use half-precision on something like Atari. I think approach 1 makes more sense.

vmoens:
But it's bc breaking 😜
I can do 2 and put a warning

vmoens:
@smorad Can you stamp this?

torchrl/envs/transforms/transforms.py
```diff
@@ -2583,6 +2589,145 @@ def __repr__(self) -> str:
...
+            if self._keys_inv_unset:
+                in_keys_inv = []
+                state_spec = env_base.state_spec
+                if state_spec is not None:
+                    for key, spec in state_spec.items(True, True):
+                        if spec.dtype == torch.float64:
+                            in_keys_inv.append(unravel_key(key))
+                action_spec = env_base.action_spec
+                if action_spec.dtype == torch.float64:
+                    in_keys_inv.append(unravel_key(env_base.action_key))
```
smorad, line 2725: Is this a necessary override? It looks like `dTypeCastTransform` already implements `_set_in_keys` in a more general manner.

vmoens:
Addressed comments from @smorad

PR #1448, opened 2023-08-07
[PoC] Faster envs

### LINKED ISSUE #1325: [BUG] Throughput vs Gym AsyncVectorEnv

vmoens:
## Description

A PoC for faster envs execution. 

Partially fixes #1325 

This PR adds a `step_and_maybe_reset` public method for the environments that should speed up data collection in settings where reset are frequent.

Following this PR, `env._step` will not return a TD with `"next"` key anymore but only the `"next"` data. This is mildly bc-breaking as we're changing a private method, but it will have consequences across classes. It won't be possible to change data at the root anymore.

## Design

The idea of `step_and_maybe_reset` is to return the root and `"next"` tensordicts separately. The root may or may not be the result of a (partial) reset. This allows `rollout` and collectors to read the done state and decide how to organise the data. This allows us to code a remote function that automatically resets the envs that are done, without requiring the parallel env to send more signals asking for the spawn envs to reset.

## Speedup

For CartPole-v1, the speedup on CPU is about 1.4x (about 5900 fps former, 8400 fps with the new).
For Pendulum-v1, there is no speedup in the script I used (since the resets are virtually absent).

Based on https://github.com/pytorch-labs/tensordict/pull/507


torchrl/envs/common.py
```diff
@@ -801,6 +802,45 @@ def state_spec(self, value: CompositeSpec) -> None:
...
+        """
+        next_tensordict = self._step(tensordict)
+        next_tensordict = self._step_proc_data(next_tensordict)
+
+        done = next_tensordict.get(self.done_key)
+        truncated = next_tensordict.get("truncated", None)
+        if truncated is not None:
+            done = done | truncated
+
+        cur_td = _fuse_tensordicts(
```
matteobettini, line 830: Why do we need this?

can't we just ` cur_td = self.reset(next_tensordict.clone())`

vmoens:
Closed by #1457

PR #1450, opened 2023-08-07
Fix MacOS Mujoco Failure

osalpekar:
Despite runners running MacOS 12.6, we see wheels for mujoco being fetched for MacOS 10.6 despite MacOS 11.0 wheels being available. This is due to a strange pattern used for MacOS versioning - where essentially setting SYSTEM_VERSION_COMPAT determines what is the MacOS SDK version Python believes is being used. More info at https://github.com/deepmind/mujoco/issues/1004 and https://eclecticlight.co/2020/08/13/macos-version-numbering-isnt-so-simple/.


vmoens (approved):
Thanks for this!

PR #1455, opened 2023-08-09
[CI] Type Check Job

osalpekar:
Not running but just porting over since this job is defined in CCI (this doesn't run in CCI either).


vmoens (commented):
We're not planning on doing type checks for RL. We can create a dormant workflow but perhaps it's best not to do anything?

osalpekar:
> We're not planning on doing type checks for RL. We can create a dormant workflow but perhaps it's best not to do anything?

Sure thing. I can close for now. Feel free to reuse this in the future if needed.

PR #1456, opened 2023-08-09
[CI] Remove CCI Config

osalpekar:
Now that all jobs are ported over (PR's out for linux_stable job and mypy), we can remove the CCI config.


vmoens (approved):
LGTM

PR #1457, opened 2023-08-11
[Refactor] Faster envs (2)

vmoens:
A minimalist version of #1448


torchrl/envs/vec_env.py
```diff
@@ -359,42 +365,51 @@ def _create_td(self) -> None:
...
-        self._selected_reset_keys = self.env_obs_keys + self.done_keys
-        self._selected_step_keys = self.env_output_keys
+        # input keys
+        self._selected_input_keys = {
+            _unravel_key_to_tuple(key) for key in self._env_input_keys
+        }
+        # output keys after reset
+        self._selected_reset_keys = {
+            _unravel_key_to_tuple(key)
+            for key in self._env_obs_keys + self.done_keys + ["_reset"]
```
matteobettini, line 393: reset keys should not include "_reset" right? 
if so there will be multiple _reset keys (one per done)

torchrl/envs/vec_env.py
```diff
@@ -624,14 +634,17 @@ def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
...
 
         if self._single_task:
             # select + clone creates 2 tds, but we can create one only
-            out = TensorDict({}, batch_size=self.shared_tensordict_parent.shape)
+            out = TensorDict(
+                {}, batch_size=self.shared_tensordict_parent.shape, device=self.device
+            )
             for key in self._selected_reset_keys:
-                _set_single_key(self.shared_tensordict_parent, out, key, clone=True)
+                if _unravel_key_to_tuple(key)[-1] != "_reset":
```
matteobettini, line 646: again, no need to check here if you just exclude the reset key from selected reset keys

matteobettini: but if you wanna check, there could be multiple reset keys

torchrl/envs/vec_env.py
```diff
@@ -853,31 +867,37 @@ def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
...
-                self.shared_tensordicts[i].update_(data)
         if self._single_task:
             # select + clone creates 2 tds, but we can create one only
-            out = TensorDict({}, batch_size=self.shared_tensordict_parent.shape)
+            out = TensorDict(
+                {}, batch_size=self.shared_tensordict_parent.shape, device=self.device
+            )
             for key in self._selected_reset_keys:
-                _set_single_key(self.shared_tensordict_parent, out, key, clone=True)
+                if _unravel_key_to_tuple(key)[-1] != "_reset":
```
matteobettini, line 899: same as above

torchrl/envs/vec_env.py
vmoens: We need the "_reset" keys in the reset_keys as these need to be synced across processes

torchrl/collectors/collectors.py
```diff
@@ -848,10 +843,9 @@ def _step_and_maybe_reset(self) -> None:
             if td_reset.batch_dims:
                 # better cloning here than when passing the td for stacking
                 # cloning is necessary to avoid modifying entries in-place
-                self._tensordict = self._tensordict.clone()
-                self._tensordict[traj_done_or_terminated] = td_reset[
-                    traj_done_or_terminated
-                ]
+                self._tensordict = torch.where(
```
matteobettini, line 846: is this compatible with lazy stacks?

vmoens: [yep](https://github.com/pytorch-labs/tensordict/blob/87299f7069acc3c9a1cc2c53656c6883e7ed8f52/tensordict/tensordict.py#L7586)

torchrl/envs/utils.py
```diff
@@ -562,6 +568,101 @@ def make_composite_from_td(data):
     return composite
 
 
+def _fuse_tensordicts(*tds, excluded, selected=None, total=None):
```
matteobettini, line 571: Is this used anywhere?

matteobettini (approved):
LGTM

Let's remember about the multiple reset keys across vec_env.

And it seems like some` "next": {}` might still be there in some steps of mocking envs (for example MultiKeyConutngEnv)

vmoens:
> Let's remember about the multiple reset keys across vec_env.

What do you mean?

matteobettini:
> > Let's remember about the multiple reset keys across vec_env.
> 
> What do you mean?

i think you updated the pr, i was just saying: let’s remeber there are multiple reset keys

PR #1460, opened 2023-08-14
[Benchmark] full training benchmarks

albertbou92:
## Description

## Motivation and Context

## Types of changes

## Checklist


benchmarks/full_trainings/dqn_mnih15_atari.py
```diff
@@ -0,0 +1,378 @@
...
+        assert noops > 0
+        obs = None
+        for _ in range(noops):
+            obs, _, done, *other = self.env.step(self.noop_action)
+            if done:
+                obs = self.env.reset(**kwargs)
+        return obs
+
+
+class EpisodicLifeEnv(gym.Wrapper):
```
vmoens, line 74: Should be a transform

benchmarks/full_trainings/dqn_mnih15_atari.py
```diff
@@ -0,0 +1,378 @@
...
+    loss_module = DQNLoss(
+        value_network=model,
+        gamma=gamma,
+        loss_function="l2",
+        delay_value=True,
+        # action_space="categorical"
+    )
+    loss_module.make_value_estimator(gamma=gamma)
+
+    target_net_updater = HardUpdate(
```
vmoens, line 252: Use soft one for rainbow

vmoens:
As discussed, I would rather change the examples than put this in benchmark for the following reasons:
1. Benchmarks are to measure speed of low-level components of the lib. Maybe in the future we'll also test full training pipelines but this seems to be oriented toward performance benchmarks in terms of return (not speed)
2. We already have examples and it will be confusing for the users to have 2 examples of a single algo within the lib. 

At the end of the day, our examples have the same scope as these benchmarks: reproducing SOTA results. We can rename examples into "implementations" or anything else if that makes it clearer.

Also, IMO for DQN we should have 1 script for Rainbow (that we need) and 1 config for Rainbow + 1 config for DQN (original paper -- ablation study).

Images should not be hosted in torchrl but on our S3 bucket

vmoens:
Another point: I think each script should be an independent PR :) 
We should also clearly define what metrics we must log (in order for us to make sure that all algos log the same stuff).
Here's a few to get started:
- [ ] Returns
- [ ] Gradient norm
- [ ] Loss values (if they make sense?)
- [ ] Episode length (if it makes sense)

These are just ideas obv

Each script should be accompanied by a `benchmark_algoXX.yaml` that we run on the RC branch before the release to make sure that the algo runs fine. This yaml file is a SOTA implementation: it must be close to the paper results / public results on another lib. Runtime isn't really an issue for this file (we can have another one that is called `fast_blabla`).

@BY571 for context

benchmarks/full_trainings/dqn_mnih15_atari.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+def make_parallel_env(env_name, device, is_test=False):
+    env = ParallelEnv(1, EnvCreator(lambda: make_base_env(env_name, device)))
+    env = TransformedEnv(env)
+    env.append_transform(ToTensorImage())
+    env.append_transform(GrayScale())
+    env.append_transform(Resize(84, 84))
+    env.append_transform(CatFrames(N=4, dim=-3))
+    env.append_transform(RewardSum())
```
matteobettini, line 115: reward sum was broken for parallel envs, could have this been the problem?

Issue #1344, opened 2023-07-01
[Feature Request] Q Ensembles


smorad:
## Motivation
Twin Q/ensemble Q functions are used in many RL algorithms and mitigate Q overestimation. My understanding is that TorchRL only deals with ensembles in the loss functions. This is fine for actor/critic methods since we only use the critics to compute actor loss. But for critic-only methods (e.g. DQN), we need the Q ensemble at sample collection time. Doing so would also simplify the loss functions for DDPG/SAC/REDQ/etc.

## Solution
I would like to add ensemble Q function support to TorchRL, but I'm not sure on the best way to do this. I was thinking of creating a  `TensorDictModuleEnsemble` in `tensordict_module.py` that could be used for more than just Q functions. The issue is that we essentially need two `forward` functions: one at sample-time to compute some reduce operation like `min` over the ensemble outputs, and one at training time that does _not_ reduce, but rather does something like 
```python
self.ensemble(tensordict.expand(tensordict.shape, ensemble_size))
```
 so we can compute the loss for all Q functions. I'm not sure if there is a good way to tell a `TensorDictModule` whether it is in "sampling" or "training" mode.

I think it also makes sense to provide an option to keep separate datasets for each model/Q function, e.g. 
```python
tensordict.set("ensemble_idx", tensordict.get("ensemble_idx", torch.randint(ensemble_size, 1)))
for i in range(ensemble_size):
  sub_td = tensordict[tensordict["ensemble_idx"] == i]
  self.models[i](sub_td)
...
```
I'd like to avoid the for-loop here if possible, but I'm not sure how.
 
## Additional context

Related to https://github.com/pytorch/rl/issues/876 
## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

vmoens:
Thanks for raising this
Seems like an interesting problem to tackle.

Regarding basic implementation, I think we should keep a contiguous stack of params and use vmap over the Q-Value network, that should be faster than looping over the networks / params.

Next you're raising the question of how to select or not the min value. We do that in a very custom fashion in the losses, but I agree that we should find a more generic way of doing so. If we get a table of state-action values like
```python
[[v_00, v_01],
[v_10, v_11]]
```
for row-based actions 0, 1 and column-based networks 0, 1, how do you select an action using the `min` operator (knowing that you want the action with the maximum value)?
[...]

PR #1461, opened 2023-08-14
[BugFix] Discrete SAC rewrite

### LINKED ISSUE #1459: [BUG] Using DiscreteSAC in Multi-Agent environments
## Describe the bug

I am currently trying to use the `DiscreteSACLoss` in a multi-agent environment. I am currently following this [tutorial](https://pytorch.org/rl/tutorials/multiagent_ppo.html). When I try to run my example though, I get some dimension issues. Maybe we are handling the dimensions (unsqueezing/squeezing) incorrectly in the loss so that it doesn't generalize to multi-agent environments?

It is important to note that in my setup, I am only using a single-agent. However, the game environment I'm using (Unity) supports multiple agents so the tensordicts/specs are all setup as if it were a multi-agent setup. For the purposes of this bug though, we don't really need to consider Unity. I just provide an example that is independent of Unity below.

## To Reproduce

You can run this simple script using the latest `torchrl` and `tensordict` installed from `main`:

```python
import math

import torch
from tensordict.nn import InteractionType
from torch import nn
from torchrl.modules.distributions import OneHotCategorical
from torchrl.data.tensor_specs import (
    CompositeSpec,
    DiscreteTensorSpec,
)
from torchrl.modules import ProbabilisticActor, SafeModule, ValueOperator
from torchrl.objectives import DiscreteSACLoss
from torchrl.modules import MultiAgentMLP


device = "cpu"
actor_net = MultiAgentMLP(
    n_agent_inputs=128,
    num_cells=[256, 256],
    n_agent_outputs=5,
    centralised=False,
    n_agents=1,
    activation_class=nn.ReLU,
    share_params=False,
)

actor_module = SafeModule(
    actor_net,
    in_keys=[("agents", "encoder_vec")],
    out_keys=[
        ("agents", "logits"),
    ],
)
unbatched_action_spec = CompositeSpec(
    {
        "agents": CompositeSpec(
            {"action": DiscreteTensorSpec(n=5, shape=torch.Size([1, 1]), dtype=torch.int64)}
        )
    }
)
actor = ProbabilisticActor(
    spec=unbatched_action_spec,
    in_keys=[("agents", "logits")],
    out_keys=[("agents", "action")],
    module=actor_module,
    distribution_class=OneHotCategorical,
    default_interaction_type=InteractionType.RANDOM,
    return_log_prob=False,
)

qvalue_net = MultiAgentMLP(
    n_agent_inputs=128,
    num_cells=[256, 256],
    n_agent_outputs=5,
    centralised=False,
    n_agents=1,
    activation_class=nn.ReLU,
    share_params=False,
)

qvalue = ValueOperator(
    in_keys=[("agents", "encoder_vec")],
    out_keys=[("agents", "state_value")],
    module=qvalue_net,
)

model = torch.nn.ModuleList([actor, qvalue]).to(device)


loss_module = DiscreteSACLoss(
    actor_network=model[0],
    qvalue_network=model[1],
    num_actions=5,
    num_qvalue_nets=2,
    target_entropy_weight=0.2,
    loss_function="smooth_l1",
)
loss_module.set_keys(action=("agents", "action"), reward=("agents", "reward"), done=("agents", "done"), value=("agents", "state_value"), priority=("agents", "td_error"))
loss_module.make_value_estimator(gamma=0.99)

td = TensorDict(
    source={
        "agents": TensorDict(
            source={
                "action": torch.zeros((256, 1, 5), dtype=torch.float32),
                "encoder_vec": torch.zeros((256, 1, 128), dtype=torch.float32),
                "logits": torch.zeros((256, 1, 5), dtype=torch.float32),
            },
            batch_size=torch.Size([256, 1])
        ),
        "next": TensorDict(
            source={
                "agents": TensorDict(
                    source={
                        "reward": torch.zeros((256, 1, 1), dtype=torch.float32),
                        "done": torch.zeros((256, 1, 1), dtype=torch.bool),
                    },
                    batch_size=torch.Size([256, 1])
                )
            },
            batch_size=torch.Size([256])
        )
    },
    batch_size=torch.Size([256])
)
loss_module(td)
```

```bash
/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/torchrl/objectives/common.py:33: UserWarning: No target network updater has been associated with this loss module, but target parameters have been found. While this is supported, it is expected that the target network updates will be manually performed. You can deactivate this warning by turning the RL_WARNINGS env variable to False.
  warnings.warn(
/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/torchrl/objectives/common.py:343: UserWarning: No target network updater has been associated with this loss module, but target parameters have been found. While this is supported, it is expected that the target network updates will be manually performed. You can deactivate this warning by turning the RL_WARNINGS env variable to False.
  warnings.warn(
  File "/Users/hyerra/Desktop/crew-algorithms/.../harish_algorithm/test.py", line 107, in <module>
    loss_module(td)
  File "/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/torch/nn/modules/module.py", line 1538, in _call_impl
    result = forward_call(*args, **kwargs)
  File "/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/tensordict/_contextlib.py", line 126, in decorate_context
    return func(*args, **kwargs)
  File "/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/tensordict/nn/common.py", line 282, in wrapper
    return func(_self, tensordict, *args, **kwargs)
  File "/Users/hyerra/Library/Caches/pypoetry/virtualenvs/.../lib/python3.10/site-packages/torchrl/objectives/sac.py", line 1175, in forward
    td_error = (pred_val - target_value.expand_as(pred_val)).pow(2)
RuntimeError: The expanded size of the tensor (2) must match the existing size (256) at non-singleton dimension 0.  Target sizes: [2, 256].  Tensor sizes: [256, 1]
```

## Expected behavior

Ideally we should be handling dimensions so that it can work in these multi-agent environments as well.

## System info

Using the latest `torchrl` and `tensordict` installed directly from GitHub source.

```python
import torchrl, numpy, sys
```

Output:
```
None 1.25.0 3.10.12 (main, Jul 28 2023, 18:34:01) [Clang 14.0.3 (clang-1403.0.22.14.1)] darwin
```

## Checklist

cc: @matteobettini

matteobettini:
Fixes #1459.

This PR rewrites discrete SAC, enabling the following features:
- support for more discrete spaces
- support for data with batch size
- following the modular implementation of continuous sac with: `_compute_target` , `_actor_loss`, `_value_loss` modular subfunctions
- readability
- grouping utils that work with discrete spaces (in dqn, discsac, actors,..) into torchrl.data.utils 

@vmoens @smorad @hyerra


smorad:
It would also be very cool to utilize the `EnsembleModule` here instead of the previous approach of stacking network params. This will ensure the ensemble members are initialized correctly and give a bit more power to the user.

matteobettini:
I am not stacking params, I am using vmap like in the normal SAC so I guess it behaves exactly like the ensemble module?

matteobettini:
@smorad at a quick look it seems that the losses have their way of making ensembles in https://github.com/pytorch/rl/blob/main/torchrl/objectives/common.py#L174 (which also creates target params).
so I think that if we want to use EnsembleModule the best place would be to integrate it there in the future.
for now i think I am just going to stick to normal vmap as it is used in the other losses (like sac).

hyerra:
Hey @matteobettini, thanks for fixing this for me! It seems to work on my side as well

smorad:
> @smorad at a quick look it seems that the losses have their way of making ensembles in https://github.com/pytorch/rl/blob/main/torchrl/objectives/common.py#L174 (which also creates target params).

Yes they do, as the losses were implemented before the addition of `EnsembleModule`. @vmoens and I briefly discussed this [in this issue](https://github.com/pytorch/rl/issues/1344#issuecomment-1617581712)  as a more generic form of what is currently done in the losses. It is also vmapped so it should be very fast.

Adding it would be fairly straightforward:

```python
loss_fn = SACLoss(
  q_network=EnsembleModule(MyQNetwork(), num_copies=5)
  ...
)

class SACLoss:
...
def forward(...):
  ...
  q_values = self.q_network(td)
  if isinstance(self.q_network, EnsembleModule):
    # Min reduce over q values
    q_values['action_value'] = q_values['action_value'].min(0).values
    # EnsembleModule will expand at dim 0, so this just squeezes it back
    q_values = q_values[0]
 ...
```

It would open up a lot of interesting options. For example, one could consider using an Ensemble of actors too, and taking the combined mode action in the discrete SAC case.

matteobettini:
Yea ideally it would be cool to pass ensembles in place of each nn module passed to the losses.
Then losses that need ensembles by definition (such as SAC) will only create an ensemble if one was not provided.

I am not super convinced by the isinstance you called in the loss.

Wouldn't there be a seamless way of passing ensembles in place of current modules and manage the dimensionality outside of the loss? so that the loss will only see one more dim in the batch (which it is agnostic too anyway)

smorad:
Actually, my above code is indeed wrong, we don't want to reduce at all in the q function loss! Usually, we want to update all ensemble members at once:

You would instead just do
```python
def forward_q_loss(...):
  ...
  # Of shape [ensemble_size, batch, time, agent, ...]
  q_values = self.q_network(td)
  # Unsqueeze so this is [1, batch, time, agent, ...]
  target = td_estimator.estimate(td.unsqueeze(0))
  error = (q_values - target).pow(2).mean()
  return error

```

That code using `Reduce` would be used when training the actor and in the agent passed to the collector. There, we just want the "best" Q value estimate instead of all of them. You could even just pass a `Reduce` explicitly to the loss constructor:

```python
loss_fn = SACLoss(
  q_network=EnsembleModule(MyQFn(), num_copies=5),
  q_reduce=MinReduce(reduce_key='action_value')
  ...
)


def forward_actor_loss(...):
  # TD is originally of shape [ensemble, batch, ...] from the compute_q_loss
  # Now let's reduce it to [batch, ...]
  if reduce is not None:
    td = self.q_reduce(td)
  # Continue to compute actor loss as normal
  ...
```

You could easily do min/softmin/mean/median/whatever reduce you want. For the case you don't want an ensemble, just do this:
```python
loss_fn = SACLoss(
  q_network=MyQFn()
  q_reduce=None
  ...
)
```

torchrl/objectives/sac.py
```diff
@@ -1060,159 +1058,145 @@ def in_keys(self, values):
...
+        action_value = tensordict_expand.get(self.tensor_keys.action_value)
+        action = tensordict.get(self.tensor_keys.action)
+
+        # TODO this block comes from the dqn loss, we need to swap all thse with a proper
+        #  helper function which selects the value given the action for all discrete spaces
+        if self.action_space == "categorical":
+            if action.shape != action_value.shape:
+                # unsqueeze the action if it lacks on trailing singleton dim
+                action = action.unsqueeze(-1)
+            chosen_action_value = torch.gather(action_value, -1, index=action).squeeze(
```
hyerra, line 1153: Just noticed a bug here when I was testing in the categorical case. So in my setup, I have 2 qvalue nets. This causes my action_value to have a shape of: `torch.Size([2, 256, 1, 5])`. However, my action is just of the shape `torch.Size([256, 1, 1])`. Note that it has the extra dimension at the end because of the unsqueeze above. However, this causes the `gather` to fail as the `index` and `input` tensors have mismatching dimensions. Using the following code below seemed to work for me:

```python
action_idx = action.expand(torch.Size([action_value.shape[0], *action.shape]))
chosen_action_value = torch.gather(action_value, -1, index=action_idx).squeeze(-1)
```

matteobettini: thanks so much for spotting this, phew.
can you tell me if now it works in both one hot and categorical cases for you?

hyerra: Yep, it works for me!

torchrl/objectives/sac.py
```diff
@@ -1060,159 +1066,151 @@ def in_keys(self, values):
...
-            ],
-            0,
-        )
+        if shape:
+            tensordict.update(tensordict_reshape.view(shape))
+        out = {
+            "loss_actor": loss_actor.mean(),
+            "loss_qvalue": loss_value.mean(),
+            "loss_alpha": loss_alpha.mean(),
+            "alpha": self._alpha,
```
vmoens, line 1091: shouldn't we detach?

torchrl/objectives/sac.py
```diff
@@ -1060,159 +1066,151 @@ def in_keys(self, values):
...
+            "entropy": -tensordict_reshape.get(self.tensor_keys.log_prob)
+            .mean()
+            .detach(),
```
vmoens, line 1094: I'd rather not do this inside the dict construction

torchrl/objectives/sac.py
```diff
@@ -541,7 +539,9 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
             "loss_qvalue": loss_qvalue.mean(),
             "loss_alpha": loss_alpha.mean(),
             "alpha": self._alpha,
-            "entropy": -td_device.get(self.tensor_keys.log_prob).mean().detach(),
+            "entropy": -tensordict_reshape.get(self.tensor_keys.log_prob)
+            .mean()
+            .detach(),
```
vmoens, line 544: I'd rather not do this in the constructor

torchrl/objectives/sac.py
```diff
@@ -541,7 +539,9 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
             "loss_qvalue": loss_qvalue.mean(),
             "loss_alpha": loss_alpha.mean(),
             "alpha": self._alpha,
```
vmoens, line 541: detach?

vmoens (approved):
LGTM, some minor comments

torchrl/objectives/sac.py
matteobettini: this is a property, if you see it should be already detached (i follow what you wrote in the other sac)
```

   @property
    def _alpha(self):
        if self.min_log_alpha is not None:
            self.log_alpha.data.clamp_(self.min_log_alpha, self.max_log_alpha)
        with torch.no_grad():
            alpha = self.log_alpha.exp()
        return alpha
```

examples/multiagent/sac.py
matteobettini, line 1: I added this example as a bonus

matteobettini:
I have run the discrete sac example available on main to compare main vs this PR.

The reported metrics seems the same
Screenshot 2023-08-27 at 10 36 11

vmoens:
> It would also be very cool to utilize the `EnsembleModule` here instead of the previous approach of stacking network params. This will ensure the ensemble members are initialized correctly and give a bit more power to the user.

@smorad happy to consider this in the future. We should open an issue about it

PR #1462, opened 2023-08-15
[Feature] Allow multiple (nested) action, reward, done keys in `env`,`vec_env` and `collectors`

matteobettini:
Depends on [#512](https://github.com/pytorch-labs/tensordict/pull/512)

This PR is an important milestone in the journey to #1463.

It will also allow single-agent users to have more than one action, reward, and done keys.

This is very important when your agents has multiple action (e.g., some discrete, some continuous ).

## Main design problem: multiple dones

One major design choice here though is that if we allow multiple done keys, we need a "_reset" key for each (same for "truncated" , "is_init" and  all the other done-based keys). 

This is because we need to independently reset each done (e.g., in multiagent we want to reset only some done agents).

So the question is, what is the best way to do this? because we can imagine having a done_spec like

```python
done_spec = CompositeSpec{
     "nested_1": {
            "done": DiscreteSpec,
      }
      "nested_2": {
            "other_done": DiscreteSpec,
          
      }
       "done": DiscreteSpec,
}
```

With a spec of this type, we place a "_reset" key for each done present (and that has any trues).
Same will be valid for "truncated" and "is_init".

A _reset td will look like 
```python
reset_td = TensorDict{
     "nested_1": {
            "_reset": Tensor,
      }
      "nested_2": {
            "_reset": Tensor,
          
      }
       "_reset": Tensor,
}
```

This solution is already implemented in the PR

cc @hyerra

PR #1471, opened 2023-08-22
[Environment] Petting zoo

matteobettini:
Depends on #1421 and #1462

### DIFF
--- a/torchrl/envs/libs/pettingzoo.py
+++ b/torchrl/envs/libs/pettingzoo.py
@@ -1,0 +1,461 @@
+from typing import Dict, List, Optional, Union
+
+import torch
+from tensordict.tensordict import TensorDict, TensorDictBase
+
+from torchrl.data import (
+    CompositeSpec,
+    DEVICE_TYPING,
+    DiscreteTensorSpec,
+    LazyStackedCompositeSpec,
+    UnboundedContinuousTensorSpec,
+)
+from torchrl.envs.common import _EnvWrapper, EnvBase
+from torchrl.envs.libs.gym import _gym_to_torchrl_spec_transform, set_gym_backend
+from torchrl.envs.utils import _selective_unsqueeze
+
+IMPORT_ERR = None
+try:
+    import pettingzoo
+
+    _has_pettingzoo = True
+
+except ImportError as err:
+    _has_pettingzoo = False
+    IMPORT_ERR = err
+
+# __all__ = ["VmasWrapper", "VmasEnv"]
+
+
+def _get_envs() -> List[str]:
+    return [
+        "atari/basketball_pong_v3",
+        "atari/boxing_v2",
+        "atari/combat_tank_v2",
+        "atari/combat_plane_v2",
+        "atari/double_dunk_v3",
+        "atari/entombed_competitive_v3",
+        "atari/entombed_cooperative_v3",
+        "atari/flag_capture_v2",
+        "atari/foozpong_v3",
+        "atari/joust_v3",
+        "atari/ice_hockey_v2",
+        "atari/maze_craze_v3",
+        "atari/mario_bros_v3",
+        "atari/othello_v3",
+        "atari/pong_v3",
+        "atari/quadrapong_v4",
+        "atari/space_invaders_v2",
+        "atari/space_war_v2",
+        "atari/surround_v2",
+        "atari/tennis_v3",
+        "atari/video_checkers_v4",
+        "atari/volleyball_pong_v3",
+        "atari/wizard_of_wor_v3",
+        "atari/warlords_v3",
+        "classic/chess_v6",
+        "classic/rps_v2",
+        "classic/connect_four_v3",
+        "classic/tictactoe_v3",
+        "classic/leduc_holdem_v4",
+        "classic/texas_holdem_v4",
+        "classic/texas_holdem_no_limit_v6",
+        "classic/gin_rummy_v4",
+        "classic/go_v5",
+        "classic/hanabi_v5",
+        "butterfly/knights_archers_zombies_v10",
+        "butterfly/pistonball_v6",
+        "butterfly/cooperative_pong_v5",
+        "mpe/simple_adversary_v3",
+        "mpe/simple_crypto_v3",
+        "mpe/simple_push_v3",
+        "mpe/simple_reference_v3",
+        "mpe/simple_speaker_listener_v4",
+        "mpe/simple_spread_v3",
+        "mpe/simple_tag_v3",
+        "mpe/simple_world_comm_v3",
+        "mpe/simple_v3",
+        "sisl/multiwalker_v9",
+        "sisl/waterworld_v4",
+        "sisl/pursuit_v4",
+    ]
+
+
+class PettingZooWrapper(_EnvWrapper):
+
+    git_url = "https://github.com/Farama-Foundation/PettingZoo"
+    libname = "pettingzoo"
+    available_envs = _get_envs()
+
+    def __init__(
+        self,
+        env: Union[
+            "pettingzoo.utils.env.ParallelEnv", "pettingzoo.utils.env.AECEnv"
+        ] = None,
+        **kwargs,
+    ):
+        if env is not None:
+            kwargs["env"] = env
+        super().__init__(**kwargs)
+
+    @property
+    def lib(self):
+        return pettingzoo
+
+    def _build_env(
+        self,
+        env: "vmas.simulator.environment.environment.Environment",
+        from_pixels: bool = False,
+        pixels_only: bool = False,
+    ):
+        self.from_pixels = from_pixels
+        self.pixels_only = pixels_only
+
+        # TODO pixels
+        if self.from_pixels:
+            raise NotImplementedError("vmas rendering not yet implemented")
+
+        # Adjust batch size
+        if len(self.batch_size) == 0:
+            # Batch size not set
+            self.batch_size = torch.Size((env.num_envs,))
+        elif len(self.batch_size) == 1:
+            # Batch size is set
+            if not self.batch_size[0] == env.num_envs:
+                raise TypeError(
+                    "Batch size used in constructor does not match vmas batch size."
+                )
+        else:
+            raise TypeError(
+                "Batch size used in constructor is not compatible with vmas."
+            )
+
+        return env
+
+    @set_gym_backend("gym")
+    def _make_specs(
+        self, env: "vmas.simulator.environment.environment.Environment"
+    ) -> None:
+        # TODO heterogenous spaces
+
+        # Agent specs
+        action_specs = []
+        observation_specs = []
+        reward_specs = []
+        info_specs = []
+        for agent_index, agent in enumerate(self.agents):
+            action_specs.append(
+                CompositeSpec(
+                    {
+                        "action": _gym_to_torchrl_spec_transform(
+                            self.action_space[agent_index],
+                            categorical_action_encoding=True,
+                            device=self.device,
+                            remap_state_to_observation=False,
+                        )  # shape = (n_actions_per_agent,)
+                    },
+                )
+            )
+            observation_specs.append(
+                CompositeSpec(
+                    {
+                        "observation": _gym_to_torchrl_spec_transform(
+                            self.observation_space[agent_index],
+                            device=self.device,
+                            remap_state_to_observation=False,
+                        )  # shape = (n_obs_per_agent,)
+                    },
+                )
+            )
+            reward_specs.append(
+                CompositeSpec(
+                    {
+                        "reward": UnboundedContinuousTensorSpec(
+                            shape=torch.Size((1,)),
+                            device=self.device,
+                        )  # shape = (1,)
+                    }
+                )
+            )
+            agent_info = self.scenario.info(agent)
+            if len(agent_info):
+                info_specs.append(
+                    CompositeSpec(
+                        {
+                            key: UnboundedContinuousTensorSpec(
+                                shape=_selective_unsqueeze(
+                                    value, batch_size=self.batch_size
+                                ).shape[1:],
+                                device=self.device,
+                                dtype=torch.float32,
+                            )
+                            for key, value in agent_info.items()
+                        },
+                    ).to(self.device)
+                )
+
+        # Create multi-agent specs
+        multi_agent_action_spec = torch.stack(
+            action_specs, dim=0
+        )  # shape = (n_agents, n_actions_per_agent)
+        multi_agent_observation_spec = torch.stack(
+            observation_specs, dim=0
+        )  # shape = (n_agents, n_obs_per_agent)
+        multi_agent_reward_spec = torch.stack(
+            reward_specs, dim=0
+        )  # shape = (n_agents, 1)
+
+        self.het_specs = isinstance(
+            multi_agent_observation_spec, LazyStackedCompositeSpec
+        ) or isinstance(multi_agent_action_spec, LazyStackedCompositeSpec)
+
+        done_spec = DiscreteTensorSpec(
+            n=2,
+            shape=torch.Size((1,)),
+            dtype=torch.bool,
+            device=self.device,
+        )  # shape = (1,)
+
+        self.unbatched_action_spec = CompositeSpec({"agents": multi_agent_action_spec})
+        self.unbatched_observation_spec = CompositeSpec(
+            {"agents": multi_agent_observation_spec}
+        )
+        if len(info_specs):
+            multi_agent_info_spec = torch.stack(info_specs, dim=0)
+            self.unbatched_observation_spec[("agents", "info")] = multi_agent_info_spec
+
+        self.unbatched_reward_spec = CompositeSpec({"agents": multi_agent_reward_spec})
+        self.unbatched_done_spec = done_spec
+
+        self.action_spec = self.unbatched_action_spec.expand(
+            *self.batch_size, *self.unbatched_action_spec.shape
+        )
+        self.observation_spec = self.unbatched_observation_spec.expand(
+            *self.batch_size, *self.unbatched_observation_spec.shape
+        )
+        self.reward_spec = self.unbatched_reward_spec.expand(
+            *self.batch_size, *self.unbatched_reward_spec.shape
+        )
+        self.done_spec = self.unbatched_done_spec.expand(
+            *self.batch_size, *self.unbatched_done_spec.shape
+        )
+
+    def _check_kwargs(self, kwargs: Dict):
+        if "env" not in kwargs:
+            raise TypeError("Could not find environment key 'env' in kwargs.")
+        env = kwargs["env"]
+        if not isinstance(env, vmas.simulator.environment.Environment):
+            raise TypeError(
+                "env is not of type 'vmas.simulator.environment.Environment'."
+            )
+
+    def _init_env(self) -> Optional[int]:
+        pass
+
+    def _set_seed(self, seed: Optional[int]):
+        self._env.seed(seed)
+
+    def _reset(
+        self, tensordict: Optional[TensorDictBase] = None, **kwargs
+    ) -> TensorDictBase:
+        if tensordict is not None and "_reset" in tensordict.keys():
+            _reset = tensordict.get("_reset")
+            envs_to_reset = _reset.squeeze(-1)
+            if envs_to_reset.all():
+                self._env.reset(return_observations=False)
+            else:
+                for env_index, to_reset in enumerate(envs_to_reset):
+                    if to_reset:
+                        self._env.reset_at(env_index, return_observations=False)
+        else:
+            self._env.reset(return_observations=False)
+
+        obs, dones, infos = self._env.get_from_scenario(
+            get_observations=True,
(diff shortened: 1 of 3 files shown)


matteobettini:
@markhaoxiang if you want try it out and let me know if it works for you

torchrl/envs/libs/vmas.py
```diff
@@ -112,7 +112,7 @@ def __init__(
             if "device" in kwargs.keys() and kwargs["device"] != str(env.device):
                 raise TypeError("Env device is different from vmas device")
             kwargs["device"] = str(env.device)
-        super().__init__(**kwargs)
+        super().__init__(**kwargs, allow_done_after_reset=True)
```
vmoens, line 115: why is this in this PR?

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
+import copy
```
vmoens, line 1: missing headers

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+
+class PettingZooWrapper(_EnvWrapper):
+    """PettingZoo environment wrapper.
+
+    To install petting zoo follow the guide `here <https://github.com/Farama-Foundation/PettingZoo#installation>__`.
+
+    This class is a general torchrl wrapper for all PettingZoo environments.
+    It is able to wrap both ``pettingzoo.AECEnv`` and ``pettingzoo.ParallelEnv``.
+
+    Let's see how more in detail:
```
vmoens, line 157:
```suggestion
    Let's see how more in details:
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+
+    In wrapped ``pettingzoo.ParallelEnv`` all agents will step at each environment step.
+    If the number of agents during the task varies, please set ``use_action_mask=True``. "action_mask" will be provided
```
vmoens, line 160:
```suggestion
    If the number of agents during the task varies, please set ``use_action_mask=True``. ``"action_mask"`` will be provided
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+    as an output and should be used to mask out dead agents.
+    The environment will be reset as soon as one agent is done.
+
+    In wrapped ``pettingzoo.AECEnv``at each step only one agent will act.
```
vmoens, line 164:
```suggestion
    In wrapped ``pettingzoo.AECEnv``, at each step only one agent will act.
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+    For this reason, it is compulsory to set ``use_action_mask=True`` for this type of environment.
+    "action_mask" will be provided as an output and can be used to mask out non-acting agents and unavailable actions
```
vmoens, line 166:
```suggestion
    ``"action_mask"`` will be provided as an output and can be used to mask out non-acting agents and unavailable actions
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+    For example, you can provide ``MarlGroupMapType.ONE_GROUP_PER_AGENT``, telling that each agent should
+    have its own tensordict (similar to the pettingzoo parallel API).
+
+    Grouping is useful for leveraging vectorisation among agents whose data goes through the same
+    neural network.
+
+    Args:
+        env (``pettingzoo.utils.env.ParallelEnv`` or ``pettingzoo.utils.env.AECEnv``): the pettingzoo environment to wrap.
+        return_state (bool, optional): whether to return the global state from pettingzoo
+            (not available in all environments). Default False.
```
vmoens, line 211:
```suggestion
            (not available in all environments). Defaults to ``False``.
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+        group_map (MarlGroupMapType or Dict[str, List[str]]], optional): how to group agents in tensordicts for
+            input/output. By default, agents will be grouped by their name. Otherwise, a group map can be specified
+            or selected from some premade options. See :class:`torchrl.envs.utils.MarlGroupMapType` for more info.
+        use_action_mask (bool, optional): whether the environment should ouptut an "action_mask". This is compulsory in
```
vmoens, line 215:
```suggestion
        use_action_mask (bool, optional): whether the environment should output an ``"action_mask"``. This is compulsory in
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+            wrapped ``pettingzoo.AECEnv`` to mask out unavailable actions and non-acting agents and should be also used
+            for ``pettingzoo.ParallelEnv`` when the number of agents can vary. Default False.
```
vmoens, line 217:
```suggestion
            for ``pettingzoo.ParallelEnv`` when the number of agents can vary. Defaults to ``False``.
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+
+    Args:
+        env (``pettingzoo.utils.env.ParallelEnv`` or ``pettingzoo.utils.env.AECEnv``): the pettingzoo environment to wrap.
+        return_state (bool, optional): whether to return the global state from pettingzoo
+            (not available in all environments). Default False.
+        group_map (MarlGroupMapType or Dict[str, List[str]]], optional): how to group agents in tensordicts for
+            input/output. By default, agents will be grouped by their name. Otherwise, a group map can be specified
+            or selected from some premade options. See :class:`torchrl.envs.utils.MarlGroupMapType` for more info.
+        use_action_mask (bool, optional): whether the environment should ouptut an "action_mask". This is compulsory in
+            wrapped ``pettingzoo.AECEnv`` to mask out unavailable actions and non-acting agents and should be also used
```
vmoens, line 216:
```suggestion
            wrapped ``pettingzoo.AECEnv`` to mask out unavailable actions and non-acting agents and should be also used. Defaults to ``False``.
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+            for ``pettingzoo.ParallelEnv`` when the number of agents can vary. Default False.
+        seed (int, optional): the seed. Default None.
```
vmoens, line 218:
```suggestion
        seed (int, optional): the seed. Defaults to ``None``.
```

vmoens: We should move away from seeding in constructors IMO (seeding should be done via set_seed whenever we can)

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+IMPORT_ERR = None
+try:
+    import pettingzoo
+
+    _has_pettingzoo = True
+
+except ImportError as err:
+    _has_pettingzoo = False
+    IMPORT_ERR = err
+
```
vmoens, line 26: We should use local imports (if possible) and do `_has_pettingzoo = importlib.util.find_spec("pettingzoo") is not None` to reduce the time it takes to load the lib

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+__all__ = ["PettingZooWrapper", "PettingZooEnv"]
```
vmoens, line 27: we should move away from `__all__` as `from .pettingzoo import *` will not be used.

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,929 @@
...
+                    if agent not in agents_acting:
+                        group_mask[index] = False
+
+        return observation_dict, info_dict
+
+    def close(self) -> None:
+        self._env.close()
+
+
+class PettingZooEnv(PettingZooWrapper):
```
vmoens, line 768: edit docstring based on comments above

vmoens (commented):
Great stuff, some early comments while waiting for ActionMask to be integrated

torchrl/envs/libs/vmas.py
matteobettini: because I have introduced in this PR a flag to allow environments to be done after reset (this is necessary when you have multiple dones). 

vmas needed this flag because it has some scenarios which might happen to be done in some of the batch dimensions after a reset

torchrl/envs/libs/pettingzoo.py
matteobettini: but does it hurt to set it from the constructor? cause for envs like SMAC (you will see in the other PR) it can only be set at construction time

torchrl/envs/libs/pettingzoo.py
matteobettini: what do you mean local imports? 
i have updated to the spec finding code proposed

matteobettini:
Thanks for reviewing, I have updated the changes requested.

One thing that I wanted to discuss from this PR is the masking.
Here I am suing the "action_mask" with essentially 2 functions:

- reading the "action_mask" coming from pettingzoo which informs on available actions. In this case I also call `update_mask` on the spec (since it is just limiting the available actions but some actions will be available)
- setting it to all False for agents which are not supposed to act. In this case I am NOT feeding it to the spec (since an action mask of all False is not permitted)

so essentially when you want to use a turn-based (AEC) env from pettingzoo for training you will have a key "action_mask" which could be all Falses for agents not acting or partially Falses for agent acting with limited actions.

Now, my question is, how do we deal with this in training? cause if you feed it straight to a masked distribution it will cause problems for non acting agents.

to note  also is  that grouping is still available in AEC envs so you can have


```python
env = PettingZooEnv(
   task="tictactoe_v3",
   parallel=False,
  use_action_mask=True, # Must use it since one player plays at a time
)
env.rollout(10)
```

```
TensorDict(
    fields={
        next: TensorDict(
            fields={
                player: TensorDict(
                    fields={
                        action_mask: Tensor(shape=torch.Size([9, 2, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                        done: Tensor(shape=torch.Size([9, 2, 1]), device=cpu, dtype=torch.bool, is_shared=False),
                        observation: TensorDict(
                            fields={
                                observation: Tensor(shape=torch.Size([9, 2, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
                            batch_size=torch.Size([9, 2]),
                            device=cpu,
                            is_shared=False),
                        reward: Tensor(shape=torch.Size([9, 2, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([9, 2]),
                    device=cpu,
                    is_shared=False)},
            batch_size=torch.Size([9]),
            device=cpu,
            is_shared=False),
        player: TensorDict(
            fields={
                action: Tensor(shape=torch.Size([9, 2, 9]), device=cpu, dtype=torch.int64, is_shared=False),
                action_mask: Tensor(shape=torch.Size([9, 2, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                done: Tensor(shape=torch.Size([9, 2, 1]), device=cpu, dtype=torch.bool, is_shared=False),
                observation: TensorDict(
                    fields={
                        observation: Tensor(shape=torch.Size([9, 2, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
                    batch_size=torch.Size([9, 2]),
                    device=cpu,
                    is_shared=False)},
            batch_size=torch.Size([9, 2]),
            device=cpu,
            is_shared=False)},
    batch_size=torch.Size([9]),
    device=cpu,
    is_shared=False)

```

where by default player_0 and player_1 will be grouped toghether. here if you index along the dimension 2 of the action mask you will see that when it has some true for one agent, the other has all falses

or you can have

```python
env = PettingZooEnv(
   task="tictactoe_v3",
  parallel=False,
  use_action_mask=True, # Must use it since one player plays at a time
  group_map=MarlGroupMapType.ONE_GROUP_PER_AGENT,
   # This time let's split the players (even though since they are both named "player_i" the default would group them together)
 )
env.rollout(10)
```
```
Out[3]: 
TensorDict(
    fields={
        next: TensorDict(
            fields={
                player_1: TensorDict(
                    fields={
                        action_mask: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                        done: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.bool, is_shared=False),
                        observation: TensorDict(
                            fields={
                                observation: Tensor(shape=torch.Size([8, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
                            batch_size=torch.Size([8]),
                            device=cpu,
                            is_shared=False),
                        reward: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([8]),
                    device=cpu,
                    is_shared=False),
                player_2: TensorDict(
                    fields={
                        action_mask: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                        done: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.bool, is_shared=False),
                        observation: TensorDict(
                            fields={
                                observation: Tensor(shape=torch.Size([8, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
                            batch_size=torch.Size([8]),
                            device=cpu,
                            is_shared=False),
                        reward: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([8]),
                    device=cpu,
                    is_shared=False)},
            batch_size=torch.Size([8]),
            device=cpu,
            is_shared=False),
        player_1: TensorDict(
            fields={
                action: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.int64, is_shared=False),
                action_mask: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                done: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.bool, is_shared=False),
                observation: TensorDict(
                    fields={
                        observation: Tensor(shape=torch.Size([8, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
                    batch_size=torch.Size([8]),
                    device=cpu,
                    is_shared=False)},
            batch_size=torch.Size([8]),
            device=cpu,
            is_shared=False),
        player_2: TensorDict(
            fields={
                action: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.int64, is_shared=False),
                action_mask: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                done: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.bool, is_shared=False),
                observation: TensorDict(
                    fields={
                        observation: Tensor(shape=torch.Size([8, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
                    batch_size=torch.Size([8]),
                    device=cpu,
                    is_shared=False)},
            batch_size=torch.Size([8]),
            device=cpu,
            is_shared=False)},
    batch_size=torch.Size([8]),
    device=cpu,
    is_shared=False)
```

where now the action_mask in each group refers to a single agent


the question is, what does a training script look like for these envs? since the mask can be all falses for some agents

vmoens:
I'm not sure how to answer this without assuming anything about what the loss module does. For instance, with the last tensordict you displayed, what should we assume the model looks like? 

IIUC petting zoo can give you action masks that are all False, meaning no action should be taken, right?
From my point of view there is a confusion between masking actions and masking agents. The action_mask is used as a proxy for the second, but maybe we should explicitly register an entry that says whether an agent is valid or not:

```python
TensorDict(
    fields={
        next: TensorDict(
            fields={
                player_1: TensorDict(
                    fields={
                        action_mask: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                        done: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.bool, is_shared=False),
                        observation: TensorDict(
                            fields={
                                observation: Tensor(shape=torch.Size([8, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
                            batch_size=torch.Size([8]),
                            device=cpu,
                            is_shared=False),
                        reward: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([8]),
                    device=cpu,
                    is_shared=False),
                player_2: TensorDict(
                    fields={
                        action_mask: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                        done: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.bool, is_shared=False),
                        observation: TensorDict(
                            fields={
                                observation: Tensor(shape=torch.Size([8, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
                            batch_size=torch.Size([8]),
                            device=cpu,
                            is_shared=False),
                        reward: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([8]),
                    device=cpu,
                    is_shared=False)},
            batch_size=torch.Size([8]),
            device=cpu,
            is_shared=False),
        player_1: TensorDict(
            fields={
                action: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.int64, is_shared=False),
                mask: Tensor(shape=torch.Size([8]), device=cpu, dtype=torch.bool, is_shared=False), <-- Here, same shape as TD
                action_mask: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                done: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.bool, is_shared=False),
                observation: TensorDict(
                    fields={
                        observation: Tensor(shape=torch.Size([8, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
                    batch_size=torch.Size([8]),
                    device=cpu,
                    is_shared=False)},
            batch_size=torch.Size([8]),
            device=cpu,
            is_shared=False),
        player_2: TensorDict(
            fields={
                action: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.int64, is_shared=False),
                mask: Tensor(shape=torch.Size([8]), device=cpu, dtype=torch.bool, is_shared=False), <-- Here, same shape as TD
                action_mask: Tensor(shape=torch.Size([8, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                done: Tensor(shape=torch.Size([8, 1]), device=cpu, dtype=torch.bool, is_shared=False),
                observation: TensorDict(
                    fields={
                        observation: Tensor(shape=torch.Size([8, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
                    batch_size=torch.Size([8]),
                    device=cpu,
                    is_shared=False)},
            batch_size=torch.Size([8]),
            device=cpu,
            is_shared=False)},
    batch_size=torch.Size([8]),
    device=cpu,
    is_shared=False)
```


Another thing we could consider is that `env.step` does not always returns a full tensordict but only the leaves that are meaningful. `fake_tensordict` will still return the full data and can be used to build your replay buffer.

matteobettini:
Ok makes sense, we can separate agent masks from action masks to improve clarity and differenctiate semantics, I'll do that.

Regarding the step outputting partial data i would prefer to do not do that since there are a lot of compnents relying on data not changing structure over time.

Essentially we can do what I am doing now and build outputs from zeroed specs to make sure it is consistent

torchrl/envs/libs/pettingzoo.py
```diff
@@ -0,0 +1,872 @@
...
+)
+from torchrl.envs.common import _EnvWrapper
+from torchrl.envs.libs.gym import _gym_to_torchrl_spec_transform, set_gym_backend
+from torchrl.envs.utils import _replace_last, check_marl_grouping, MarlGroupMapType
+
+
+_has_pettingzoo = importlib.util.find_spec("pettingzoo") is not None
+
+if typing.TYPE_CHECKING and _has_pettingzoo:
+    import pettingzoo
```
vmoens, line 28: can this be a local import?

vmoens (approved):
LGTM
Let's remove global imports

torchrl/envs/libs/pettingzoo.py
matteobettini: this is used for type checking.
I think it cannot be local if we want the type hints

vmoens: personally I think we can just ditch all type hints but if you want them just do `sutff: "pettingzoo.something"`

vmoens: btw importing a lib for the sake of type hinting is a bit of an overkill no?

matteobettini:
done

vmoens (approved):

Issue #1198, opened 2023-05-26
TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.


Killpit:
I tried following the RL tutorial from [here](https://pytorch.org/tutorials/intermediate/reinforcement_ppo.html?utm_source=whats_new_tutorials&utm_medium=reinforcement_ppo). Even though it mentioned cuda, I use MPS and when I tried to apply the same code with MPS logic, I couldn't make it work, but it works perfectly without has_mps and using cuda.

```

device = "cpu" if not torch.has_mps else "mps:0"
num_cells = 256  #number of cells in each layer
lr = 3e-4
max_grad_norm = 1.0

frame_skip = 1
frames_per_batch = 1000 // frame_skip
#For a complete training, bring the number of frames up to 1M
total_frames = 50_000 // frame_skip

sub_batch_size = 64  #cardinality of the sub-samples gathered from the current
#data in the inner loop
num_epochs = 10  #optimization steps per batch of data collected
clip_epsilon = (0.2)  #clip value for PPO loss
gamma = 0.99
lmbda = 0.95
entropy_eps = 1e-4

base_env = GymEnv("InvertedDoublePendulum-v4", device=device, frame_skip=frame_skip)

env = TransformedEnv(
    base_env,
    Compose(
        #normalize observations
        ObservationNorm(in_keys=["observation"]),
        DoubleToFloat(in_keys=["observation"]),
        StepCounter(),
    ),
)

env.transform[0].init_stats(num_iter=1000, reduce_dim=0, cat_dim=0)

print("normalization constant shape:", env.transform[0].loc.shape)

```

vmoens:
We currently have limited coverage over MPS but it's a good point.
Let me see how we can make sure that we support that too!
Here it seems that the error occurs when we convert a float64 tensor from cpu to MPS when reading it from gym. Without the full error stack it's hard to really say what's going on but perhaps changing the corresponding env spec to have a type float32 instead of float64 could solve it (my guess is that this occurs during a call to `env.observation_spec.encode`?)

Killpit:
This is the full error

Traceback (most recent call last):
  File "/Users/atatekeli/PycharmProjects/PyTorchProjects/torch_rl.py", line 38, in <module>
    base_env = GymEnv("InvertedDoublePendulum-v4", device=device, frame_skip=frame_skip)
[...]

PR #1472, opened 2023-08-26
[Feature] Device transform

vmoens:
## Description

Adds a `DeviceCastTransform` transform to move environment data from one device to another.

As part of this PR, transforms now can transform the device of the parent env through `transform.transform_env_device`.


torchrl/envs/transforms/transforms.py
```diff
@@ -2708,6 +2738,87 @@ def __init__(
...
+        return tensordict.to(self.device, non_blocking=True)
+
+    def set_container(self, container: Union[Transform, EnvBase]) -> None:
+        if self.orig_device is None:
+            if isinstance(container, EnvBase):
+                self.orig_device = container.device
+                container._device = self.device
+            elif container.parent is not None:
+                self.orig_device = container.parent.device
+                raise RuntimeError("How can we change the device of the container?")
```
matteobettini, line 2788: Is this supposed to be here?

torchrl/envs/transforms/transforms.py
```diff
@@ -2708,6 +2738,87 @@ def __init__(
         super().__init__(torch.double, torch.float, in_keys, in_keys_inv)
 
 
+class DeviceCastTransform(Transform):
+    """Casts the env device.
```
matteobettini, line 2742: Is this really casting the env device?

Isn't it transforming the device of the data?

matteobettini (approved):
I have a question:

In environments like VMAS where there is an internal state that sits on a device, the way I thought that the environment is moved to another device is `env.to(device)` 

If i understand correctly here we are casting the env data and its specs to another device, but not its internal state (we do not call `env.to()`). 

does this make sense to use with a parent environment? when is this transform preferred to `env.to(device)`?

the only case i can imagine is when we need to keep an env of a specific device, apply some transforms there and then move the data to another device

vmoens:
> does this make sense to use with a parent environment? when is this transform preferred to `env.to(device)`?

This is to address https://github.com/pytorch/rl/issues/1198 where the issue is that if the env naturally sits on MPS we can't use float64. So first you must transform the data into float32 and then cast to device. Doing env.to(device) will not work but this transform will.

EkaterinaAbramova:
> > does this make sense to use with a parent environment? when is this transform preferred to `env.to(device)`?
> 
> This is to address #1198 where the issue is that if the env naturally sits on MPS we can't use float64. So first you must transform the data into float32 and then cast to device. Doing env.to(device) will not work but this transform will.

Could you please explain exactly how to do this? I am confused. I am flooding this tutorial https://pytorch.org/tutorials/intermediate/reinforcement_ppo.html and get that error about MPS flaot64 on the line: base_env = GymEnv("InvertedDoublePendulum-v4", device=device, frame_skip=frame_skip) What code exactly shall I write to correct this error please?

vmoens:
Can I ask what the value of device is in your case?

EkaterinaAbramova:
device="mps". 

I believe I solved it with this (after quite a few hours of trying different things!!!):

```
base_env = GymEnv("InvertedDoublePendulum-v4", device="cpu", frame_skip=frame_skip)
env = TransformedEnv(
    base_env,
    Compose(
        ObservationNorm(in_keys=["observation"]), # normalise observations (make it about Standard Normal)
        DoubleToFloat(),   
        StepCounter(),                            # count the number of steps before the environment is terminated
        DeviceCastTransform(device=device, orig_device="cpu"),
    ),
)
print(env.device) # gives mps now
```

Could you please kindly confirm if what Ive done is correct? I am on Apple M2 max trying to use MPS.

EkaterinaAbramova:
Everything was progressing smoothly through the tutorial: https://pytorch.org/tutorials/intermediate/reinforcement_ppo.html however at this code I get an error again about MPS. Please could you kindly advise syntax to solve this issue? I double checked everything seems to be on mps, so I don't understand where the error is coming from.

```
    env,
    policy_module,
    total_frames=total_frames,
    split_trajs=False,
    device=device,
)


  File ~/anaconda3/envs/gpu-torch-rl/lib/python3.10/site-packages/torch/nn/modules/module.py:1143 in convert
    return t.to(device, dtype if t.is_floating_point() or t.is_complex() else None, non_blocking)

TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.


env.device
Out[23]: device(type='mps')
```
Screenshot 2023-09-30 at 01 36 12

Screenshot 2023-09-30 at 01 38 39

vmoens:
@EkaterinaAbramova sorry you had this terrible experience, we should document things better for MPS. 
It's something we're actively looking at and any feedback is very welcome.
The way you implemented your env looks great to me!

Regarding the error I will have a look into it, shouldn't be too difficult to solve. Have you tried moving the `ObservationNorm` and `StepCounter` after the device casting transform? 
like:
```python
env = TransformedEnv(
    base_env,
    Compose(
        DoubleToFloat(),   
        DeviceCastTransform(device=device, orig_device="cpu"),
        ObservationNorm(in_keys=["observation"]), # normalise observations (make it about Standard Normal)
        StepCounter(),                            # count the number of steps before the environment is terminated
    ),
)
```
Like this the buffers in the `ObservationNorm` transform will sit on mps but with float32 and not float64.

EkaterinaAbramova:
@vmoens thank you for swiftly helping me, this issue is quite urgent, so very glad that you had the suggestion. It makes sense and I tried it, however this way around I get this error `AttributeError: 'DoubleToFloat' object has no attribute 'init_stats'` (it seems I need to pass some arguments, what shall I pass to be able to follow the tutorial please https://pytorch.org/tutorials/intermediate/reinforcement_ppo.html): 

Screenshot 2023-09-30 at 23 26 42

vmoens:
You need to call init_stats on the obs norm transform
```python
env.transform[-2].init_stats(...)
```
Because the transform has changed place

EkaterinaAbramova:
OK I get it. So now that Ive indexed the correct location, the MPS issue is back. 
The code you suggested doesn't work for me:
```
env = TransformedEnv(
    base_env,
    Compose(
        DoubleToFloat(),   
        DeviceCastTransform(device=device, orig_device="cpu"),
        ObservationNorm(in_keys=["observation"]), # normalise observations (make it about Standard Normal)
        StepCounter(),                            # count the number of steps before the environment is terminated
    ),
)
print(env.device)
env.transform[-2].init_stats(num_iter=1000, reduce_dim=0, cat_dim=0) 
I get the error: Cannot convert a MPS Tensor
```

TO RECAP: The original code was: 
```
env = TransformedEnv(
    base_env,
    Compose(
        ObservationNorm(in_keys=["observation"]), # normalise observations (make it about Standard Normal)
        DoubleToFloat(),   
        StepCounter(),                            # count the number of steps before the environment is terminated
        DeviceCastTransform(device=device, orig_device="cpu"),
    ),
)
env.transform[0].init_stats(num_iter=1000, reduce_dim=0, cat_dim=0) 
```
This code runs, however I have issues further below in the tutorial where I can't run:
```
    env,
    policy_module,
    total_frames=total_frames,
    split_trajs=False,
    device=device,
)
Cannot convert a MPS Tensor
```

Could you please provide a solution? I am really in a rush now, Ive been trying to solve this issue for days. The tutorial I am working on is online, so if it helps maybe you could try the suggestions to make sure they resolve the issue? Thank you very much for your help, I really need to move past this ASAP please.

vmoens:
Ok so that's an interesting bug, which basically boils down to some internal machinery within rollout, resets and transforms.
To quickly unblock you: can you compute the stats manually with your env?
```python
simple_env = TransformedEnv(
    base_env,
    Compose(
        DoubleToFloat(),   
        DeviceCastTransform(device=device, orig_device="cpu"),
   )
)
td0 = simple_env.rollout(100)
loc = td0["observation"].mean(dim=0)
scale = td0["observation"].std(dim=0)
env = TransformedEnv(
    base_env,
    Compose(
        DoubleToFloat(),   
        DeviceCastTransform(device=device, orig_device="cpu"),
        ObservationNorm(in_keys=["observation"], loc=loc, scale=scale),
        StepCounter(),
    ),
)
```
Hopefully that should help!
I should be able to put my hands on an apple silicon computer tomorrow morning if you're still stuck!

EkaterinaAbramova:
The suggestion didn't work Im afraid. Any other thing you could propose at this stage or only tomorrow? 

Screenshot 2023-10-01 at 21 27 05

So why in the version I provided above, everything is fine until I get down to the SyncDataCollector? Why is it failing there? What has not been yet converted to float32?

vmoens:
#1589 will solve your problem!

vmoens:
@EkaterinaAbramova you should be good to go now!

vmoens:
I can reproduce your issue, let me push a fix!

PR #1477, opened 2023-08-31
[BugFix] Fix NoopResetEnv behavior when trials exceeded.

skandermoalla:
## Description

When the `_MAX_NOOPS_TRIALS` were exceeded the transform was calling a final `rand_step()` without a `step_mdp()`.
This PR add the `step_mdp()` call and 
- Improves the error message.
- Adds tests to test for the edge case behaviors of the transform.
- Fixes the documentation of the tranform.

Some things could be discussed. I left a TODO in front, which we will remove before merging.

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -3114,7 +3114,7 @@ class NoopResetEnv(Transform):
         noops (int, optional): upper-bound on the number of actions
             performed after reset. Default is `30`.
             If noops is too high such that it results in the env being
-            done or truncated before the all noops are applied,
+            done or truncated before the all the noops are applied,
             in multiple trials, the transform raises a RuntimeError.
         random (bool, optional): if False, the number of random ops will
             always be equal to the noops value. If True, the number of
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -3111,8 +3111,12 @@ class NoopResetEnv(Transform):
         env (EnvBase): env on which the random actions have to be
             performed. Can be the same env as the one provided to the
             TransformedEnv class
-        noops (int, optional): number of actions performed after reset.
-            Default is `30`.
+        noops (int, optional): upper-bound on the number of actions
+            performed after reset. Default is `30`.
+            If noops is too high such that the env is done or truncated before
+            the end of the noops on multiple trials, the transform
+            will perform a single random step, and raises an error if
+            even a single random step leads to a done environment.
         random (bool, optional): if False, the number of random ops will
             always be equal to the noops value. If True, the number of
             random actions will be randomly selected between 0 and noops.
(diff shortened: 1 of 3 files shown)


torchrl/envs/transforms/transforms.py
```diff
@@ -3167,23 +3168,40 @@ def reset(self, tensordict: TensorDictBase) -> TensorDictBase:
...
             trial += 1
             if trial > _MAX_NOOPS_TRIALS:
+                trials_exceeded_message = (
+                    "Parent env was repeatedly done or truncated"
+                    " before the sampled number of noops could be applied. "
+                )
                 tensordict = parent.rand_step(tensordict)
-                if tensordict.get(("next", done_key)):
+                tensordict = step_mdp(tensordict, exclude_done=False)
+                if tensordict.get(done_key) or tensordict.get(
```
vmoens, line 3187: As of now this works only if done is a scalar. Should we contemplate making this transform compatible with batched envs? Not sure how it would look like tbh

torchrl/envs/transforms/transforms.py
```diff
@@ -3167,23 +3168,40 @@ def reset(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+                    )
+                else:
+                    warnings.warn(
+                        trials_exceeded_message + "A single noop step was applied."
                     )
+                    # TODO: up to discussion. Could also be reset instead of doing a single noop.
                 break
 
         if tensordict.get(done_key):
+            # TODO: This will never be reached. It should be removed.
```
vmoens, line 3202: I agree let's ditch it

vmoens (approved):
LGTM thanks for this. IMO we don't want to reset if we reach the max number of trials bc it could lead to a silent bug. WDYT?

skandermoalla:
Yes. I also think a single reset is arbitrary and would lead to issues. That was the original behavior, I just fixed a bug in it.

I will refactor the transform so that it becomes simpler to parse and has less edge cases.

torchrl/envs/transforms/transforms.py
skandermoalla: The transform assumes a single env as a parent at the moment:

\- \[Code\](https://github.com/pytorch/rl/blob/7ae614043b67bfccf849f661491587339e3262a2/torchrl/envs/transforms/transforms.py#L3148).

\- \[test\](https://github.com/pytorch/rl/blob/7ae614043b67bfccf849f661491587339e3262a2/test/test_transforms.py#L3072) 

This change will require a refactor of the whole transform, so I would keep it out of the scopre of this PR.

skandermoalla:
With this, the transform raises an error whenver its consistently not possible to ally the noops (1). We could also just raise a warning and not do any noop at reset (2).

Both options are valid in my opinion, and maybe the first one is better as it will nudge to user to be more careful when using the transform. E.g.

With (1) the user cannot get away by setting a high `noops` and \`random=True\` that may fail occasionally. With (2) they could get away with it at the env will just not receive `noops` occasinally.

PR #1479, opened 2023-09-01
CNN version of MultiAgentMLP

MarkHaoxiang:
## Description

Added a MultiAgentConvNet module, which wraps a ConvNet to work in the multiagent space.

## Motivation and Context

This is an utility written for PettingZoo Multiagent environments which have pixel based observations.

## Types of changes

## Checklist

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -19,6 +19,7 @@ from torchrl.modules import (
     DTActor,
     LSTMNet,
     MultiAgentMLP,
+    MultiAgentConvNet,
     OnlineDTActor,
     QMixer,
     SafeModule,
@@ -800,6 +801,78 @@ class TestMultiAgent:
             *batch, n_agents, n_agent_inputs
         )
         out = mlp(obs)
+        for i in range(n_agents):
+            if share_params:
+                # same input same output
+                assert torch.allclose(out[..., i, :], out[..., 0, :])
+            else:
+                for j in range(i + 1, n_agents):
+                    # same input different output
+                    assert not torch.allclose(out[..., i, :], out[..., j, :])
+
+    @pytest.mark.parametrize("n_agents", [1, 3])
+    @pytest.mark.parametrize("share_params", [True, False])
+    @pytest.mark.parametrize("centralised", [True, False])
+    @pytest.mark.parametrize(
+        "batch",
+        [
+            (10,),
+            (
+                10,
+                3,
+            ),
+            (),
+        ],
+    )
+    def test_cnn(
+        self,
+        n_agents,
+        centralised,
+        share_params,
+        batch,
+        x=50,
+        y=50,
+        channels=3
+    ):
+        torch.manual_seed(0)
+        cnn = MultiAgentConvNet(
+            n_agents=n_agents,
+            centralised=centralised,
+            share_params=share_params
+        )
+        td = TensorDict(
+            {
+                "agents": TensorDict(
+                    {"observation": torch.randn(*batch, n_agents, x, y, channels)},
+                    [*batch, n_agents],
+                )
+            },
+            batch_size=batch,
+        )
+        obs = td[("agents", "observation")]
+
+        out = cnn(obs)
+        assert out.shape[:-1] == (*batch, n_agents)
+        for i in range(n_agents):
+            if centralised and share_params:
+                assert torch.allclose(out[..., i, :], out[..., 0, :])
+            else:
+                for j in range(i + 1, n_agents):
+                    assert not torch.allclose(out[..., i, :], out[..., j, :])
+
+        obs[..., 0, 0, 0, 0] += 1
+        out2 = cnn(obs)
+        for i in range(n_agents):
+            if centralised:
+                # a modification to the input of agent 0 will impact all agents
+                assert not torch.allclose(out[..., i, :], out2[..., i, :])
+            elif i > 0:
+                assert torch.allclose(out[..., i, :], out2[..., i, :])
+
+        obs = torch.randn(*batch, 1, x, y, channels).expand(
+            *batch, n_agents, x, y, channels
+        )
+        out = cnn(obs)
         for i in range(n_agents):
             if share_params:
                 # same input same output
--- a/torchrl/modules/models/multiagent.py
+++ b/torchrl/modules/models/multiagent.py
@@ -326,8 +326,6 @@ class MultiAgentConvNet(nn.Module):
         else:
             inputs = inputs[0]
 
-        inputs = inputs.float()
-
         assert len(inputs.shape) >= 4, \
 f"""Multi-agent network expects" (*batch_size, agent_index, x, y, channels)"""
         assert inputs.shape[-4] == self.n_agents, \
--- a/torchrl/modules/models/multiagent.py
+++ b/torchrl/modules/models/multiagent.py
@@ -12,7 +12,7 @@ from torch import nn
 
 from ...data import DEVICE_TYPING
 
-from .models import MLP
+from .models import MLP, ConvNet
 
 
 class MultiAgentMLP(nn.Module):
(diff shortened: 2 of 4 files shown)


torchrl/modules/models/multiagent.py
```diff
@@ -229,6 +229,141 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+        assert len(inputs.shape) >= 4, \
+f"""Multi-agent network expects" (*batch_size, agent_index, x, y, channels)"""
+        assert inputs.shape[-4] == self.n_agents, \
+f"""Multi-agent network expects {self.n_agents} but got {inputs.shape[-4]}"""
+        # If the model is centralized, agents have full observability
+        if self.centralised:
+            shape = (*inputs.shape[:-4], inputs.shape[-3], inputs.shape[-2], self.n_agents * inputs.shape[-1])
+            inputs = torch.reshape(inputs, shape)
+        # Conv2D expects (c, H, W)
+        inputs = torch.transpose(inputs, -1, -3)
```
matteobettini, line 338: We want to keep the exact same behaviour of torchrl.models.ConvNet. Users that have (x,y,c) input should do this in a transform (ToTensorImage).

Could we refactor the class to take by default (c, H, W)?

torchrl/modules/models/multiagent.py
```diff
@@ -229,6 +229,141 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+                for _ in range(self.n_agents if not self.share_params else 1)
+            ]
+        )
+
+
+    def forward(self, *inputs: Tuple[torch.Tensor]):
+        if len(inputs) > 1:
+            inputs = torch.cat([*inputs], -1)
+        else:
+            inputs = inputs[0]
```
matteobettini, line 327: We should mimic the behavior of the ConvNet and expect only one input

torchrl/modules/models/multiagent.py
```diff
@@ -229,6 +229,141 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+            default: 5
+        strides (int or Sequence[int]): Stride(s) of the conv network. If iterable, the length must match the
+            depth, defined by the num_cells or depth arguments.
+            default: 2
+        activation_class (Type[nn.Module]): activation class to be used.
+            default: nn.ELU
+        **kwargs: for :class:`torchrl.modules.models.ConvNet` can be passed to customize the ConvNet.Args:
+
+    Examples:
+        >>> import torch
```
matteobettini, line 261: Could we add some examples for sharing and centralisation?
Cause they may not be intuitive features immediately.
somthing similar to the ones in the mlp class

torchrl/modules/models/multiagent.py
```diff
@@ -229,6 +229,141 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
 
         return output
 
+class MultiAgentConvNet(nn.Module):
```
matteobettini, line 232: we need to add it to the docs.
it will have to be added in one of the .rst files in the doc folder. search fro the mlp to see where

matteobettini (commented):
This is incredible work, thanks so much! I was actually in need of a model like this.

I left some comments, but after those are addressed we should be there!

matteobettini:
also make sure you format the files and lint them.
we have a pre-commit hook, instructiuons on how to run it should be in the CONTRIBUTING file

torchrl/modules/models/multiagent.py
MarkHaoxiang: Refactored class and doc to expect (c, H, W)

torchrl/modules/models/multiagent.py
MarkHaoxiang: Removed

torchrl/modules/models/multiagent.py
MarkHaoxiang: Added

MarkHaoxiang:
Thank you for the review. I have updated the changes requested.

matteobettini:
@vmoens I’m considering merging this as is and do the refactoring to vmap in a second time as in #1497, would that work for you? Unless @MarkHaoxiang wants to take care of it?

MarkHaoxiang:
@matteobettini  Have you started implementing the common multiagent nn module? If not I'll take a shot at it this week.

matteobettini:
Nope, yo can take a shot!

So what we want is to do the exact same things but using vmap like https://github.com/pytorch/rl/pull/1497 (see that pr for context)

matteobettini:
@MarkHaoxiang we are posticipating that refactoring a bit to wait for some features so we can go ahead with this as is.
@vmoens  could you have a look?

test/test_modules.py
```diff
@@ -809,6 +810,68 @@ def test_mlp(
...
+        "batch",
+        [
+            (10,),
+            (
+                10,
+                3,
+            ),
+            (),
+        ],
+    )
```
vmoens, line 826:
```suggestion
    @pytest.mark.parametrize("batch", [(10,), (10, 3), ()])
```

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
         return output
 
 
+class MultiAgentConvNet(nn.Module):
+    """Multi-agent CNN.
+
+    This is a CNN that can operate in multi-agent contexts.
+
+    It expects inputs with shape (*B, n_agents, channels, x, y)
```
vmoens, line 238:
```suggestion
    It expects inputs with shape ``(*B, n_agents, channels, x, y)``.
```

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+            an integer is provided, every layer will have the same number of cells. If an iterable is provided,
+            the linear layers out_features will match the content of num_cells.
+        kernel_sizes (int, Sequence[Union[int, Sequence[int]]]): Kernel size(s) of the conv network.
+            default: 5
+        strides (int or Sequence[int]): Stride(s) of the conv network. If iterable, the length must match the
+            depth, defined by the num_cells or depth arguments.
+            default: 2
+        activation_class (Type[nn.Module]): activation class to be used.
+            default: nn.ELU
+        **kwargs: for :class:`torchrl.modules.models.ConvNet` can be passed to customize the ConvNet.Args:
```
vmoens, line 258:
```suggestion
        centralised (bool): If ``True``, each agent will use the inputs of all agents to compute its output, resulting in input of shape ``(*B, n_agents * channels, x, y)``. Otherwise, each agent will only use its data as input.
        share_params (bool): If ``True``, the same :class:`~torchrl.modules.ConvNet` will be used to make the forward pass
            for all agents (homogeneous policies). Otherwise, each agent will use a different :class:`~torchrl.modules.ConvNet` to process
            its input (heterogeneous policies).
        device (str or torch.device, optional): device to create the module on.
        num_cells (int or Sequence[int], optional): number of cells of every layer in between the input and output. If
            an integer is provided, every layer will have the same number of cells. If an iterable is provided,
            the linear layers ``out_features`` will match the content of ``num_cells``.
        kernel_sizes (int, Sequence[Union[int, Sequence[int]]]): Kernel size(s) of the convolutional network.
            Defaults to ``5``.
        strides (int or Sequence[int]): Stride(s) of the convolutional network. If iterable, the length must match the
            depth, defined by the num_cells or depth arguments.
            Defaults to ``2``.
        activation_class (Type[nn.Module]): activation class to be used.
            Default to :class:`torch.nn.ELU`.
        **kwargs: for :class:`~torchrl.modules.models.ConvNet` can be passed to customize the ConvNet.

```

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+                (4): Conv2d(32, 32, kernel_size=(5, 5), stride=(2, 2))
+                (5): ELU(alpha=1.0)
+                (6): SquashDims()
+                )
+            )
+        )
+        torch.Size([3, 2, 7, 2592])
+        False
+
+        Or where inputs are shared but not parameters.
```
vmoens, line 343: The structure should be 
```
Examples:
    >>> # comment
    >>> code
    >>> multiline_code:
    ...     etc.
    Output
```

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+                (2): Conv2d(32, 32, kernel_size=(5, 5), stride=(2, 2))
+                (3): ELU(alpha=1.0)
+                (4): Conv2d(32, 32, kernel_size=(5, 5), stride=(2, 2))
+                (5): ELU(alpha=1.0)
+                (6): SquashDims()
+                )
+            )
+        )
+        torch.Size([3, 2, 7, 2592])
+        False
```
vmoens, line 366:
```suggestion
        >>> print(cnn)
        MultiAgentConvNet(
            (agent_networks): ModuleList(
                (0-6): 7 x ConvNet(
                (0): Conv2d(28, 32, kernel_size=(5, 5), stride=(2, 2))
                (1): ELU(alpha=1.0)
                (2): Conv2d(32, 32, kernel_size=(5, 5), stride=(2, 2))
                (3): ELU(alpha=1.0)
                (4): Conv2d(32, 32, kernel_size=(5, 5), stride=(2, 2))
                (5): ELU(alpha=1.0)
                (6): SquashDims()
                )
            )
        )
        >>> print(result.shape)
        torch.Size([3, 2, 7, 2592])
        >>> print(all(result[0,0,0] == result[0,0,1]))
        False
```

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
```
vmoens, line 341:
```suggestion
        >>> print(cnn)
        MultiAgentConvNet(
            (agent_networks): ModuleList(
                (0-6): 7 x ConvNet(
                (0): Conv2d(4, 32, kernel_size=(5, 5), stride=(2, 2))
                (1): ELU(alpha=1.0)
                (2): Conv2d(32, 32, kernel_size=(5, 5), stride=(2, 2))
                (3): ELU(alpha=1.0)
                (4): Conv2d(32, 32, kernel_size=(5, 5), stride=(2, 2))
                (5): ELU(alpha=1.0)
                (6): SquashDims()
                )
            )
        )
        >>> print(result.shape)
        torch.Size([3, 2, 7, 2592])
        >>> print(all(result[0,0,0] == result[0,0,1]))
        False
```

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
```
vmoens, line 316:
```suggestion
        >>> print(cnn)
        MultiAgentConvNet(
            (agent_networks): ModuleList(
                (0): ConvNet(
                (0): Conv2d(4, 32, kernel_size=(5, 5), stride=(2, 2))
                (1): ELU(alpha=1.0)
                (2): Conv2d(32, 32, kernel_size=(5, 5), stride=(2, 2))
                (3): ELU(alpha=1.0)
                (4): Conv2d(32, 32, kernel_size=(5, 5), stride=(2, 2))
                (5): ELU(alpha=1.0)
                (6): SquashDims()
                )
            )
        )
        >>> print(result.shape)
        torch.Size([3, 2, 7, 2592])
        >>> print(all(result[0,0,0] == result[0,0,1]))
        False
```

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+                (5): ELU(alpha=1.0)
+                (6): SquashDims()
+                )
+            )
+        )
+        >>> result = cnn(obs)
+        >>> print(result.shape)
+        torch.Size([3, 2, 7, 2592]) - where the final dimension is determined by the input x
+        >>> print(all(result[0,0,0] == result[0,0,1]))
+        True - since both shared observations and parameters leads to all agents having identical outputs (eg. for a value function)
```
vmoens, line 291: Put comments as comments, not as print

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+        )
+
+    def forward(self, inputs: torch.Tensor):
+        assert (
+            len(inputs.shape) >= 4
+        ), """Multi-agent network expects" (*batch_size, agent_index, x, y, channels)"""
+        assert (
+            inputs.shape[-4] == self.n_agents
+        ), f"""Multi-agent network expects {self.n_agents} but got {inputs.shape[-4]}"""
+        # If the model is centralized, agents have full observability
```
vmoens, line 410: We don't use assert in the main code base.
This should be a ValueError I guess?

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+            output = self.agent_networks[0](inputs)
+            if self.centralised:
+                # If the parameters are shared, and it is centralised all agents will have the same output.
+                # We expand it to maintain the agent dimension, but values will be the same for all agents
+                n_agent_outputs = output.shape[-1]
+                output = (
+                    output.view(*output.shape[:-1], n_agent_outputs)
+                    .unsqueeze(-2)
+                    .expand(*output.shape[:-1], self.n_agents, n_agent_outputs)
+                )
```
vmoens, line 444: we can maybe split that in mode lines if we need debugging?

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+                    [net(inputs) for net in self.agent_networks], dim=-2
+                )
+            else:
+                output = torch.stack(
+                    [
+                        net(inputs[..., i, :, :, :])
+                        for i, net in enumerate(self.agent_networks)
+                    ],
+                    dim=-2,
+                )
```
vmoens, line 433:
```suggestion
                output = torch.stack(
                    [
                        net(inp)
                        for i, (net, inp) in enumerate(zip(self.agent_networks, inputs.unbind(-4)))
                    ],
                    dim=-2,
                )
```

vmoens (commented):
Great work! A couple of comments regarding formatting

vmoens:
Tests are not passing, presumably because we moved the scripts from .circleci to .github for the CI

torchrl/modules/models/multiagent.py
MarkHaoxiang: Not sure how adjusting this would help - each function call is already on an individual line

MarkHaoxiang:
Thanks for the check - I've adjusted the formatting as requested

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+                (4): Conv2d(32, 32, kernel_size=(5, 5), stride=(2, 2))
+                (5): ELU(alpha=1.0)
+                (6): SquashDims()
+                )
+            )
+        )
+        >>> result = cnn(obs)
+        ... print(result.shape)
+        torch.Size([3, 2, 7, 2592])
+        >>> # where the final dimension is determined by the input x
```
vmoens, line 290: I don't get that comment

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+        >>> print(all(result[0,0,0] == result[0,0,1]))
+        True
+        >>> # Since both shared observations and parameters leads to all agents having identical outputs (eg. for a value function)
```
vmoens, line 293: same here
the flow seems a bit broken

If we go for comments in code, it should flow like comments in a script. Otherwise we do
```

    Some story about our features.
    Examples:
        >>> # a module can get smth as input
        >>> module(smth)

    Some other story about other features.
    Examples:
        >>> # we can reuse module here:
        >>> assert module.attribute is "awesome"
```

hope that helps

torchrl/modules/models/multiagent.py
```diff
@@ -230,6 +230,221 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+        torch.Size([3, 2, 7, 2592])
+        False
+
+        >>> # Or where inputs are shared but not parameters.
+        >>> cnn = MultiAgentConvNet(
+        ...     n_agents,
+        ...     centralised = True,
+        ...     share_params = False
+        ... )
+        ... print(cnn)
```
vmoens, line 351:
```suggestion
        >>> print(cnn)
```

torchrl/modules/models/multiagent.py
vmoens: We can't really put breakpoints on this I think.
And the error stack won't be super duper clear.

From a performance perspective I don't think python is considerably faster with multi-line code like this.

```suggestion
                output = output.view(*output.shape[:-1], n_agent_outputs)
                output = output.unsqueeze(-2)
                output = output.expand(*output.shape[:-1], self.n_agents, n_agent_outputs)
```

vmoens (commented):
Almost there!

torchrl/modules/models/multiagent.py
MarkHaoxiang: Clarified comments and adjusted order.

matteobettini:
CNN tests are failing

MarkHaoxiang:
Fixed a bug with reshaping the tensor step in the centralised/parameter sharing case.

```
output = (
                    output.view(*output.shape[:-1], self.n_agent_outputs)
                    .unsqueeze(-2)
                    .expand(*output.shape[:-1], self.n_agents, self.n_agent_outputs)
                )
````

@matteobettini whats the motivation behind expanding the tensor like this (for the MLP version) compared to 

```
output = output.unsqueeze(-2)
output = output.expand(*output.shape[:-2], self.n_agents, n_agent_outputs)
```

matteobettini:
They seem equivalent, but in the first case the view acts as an implicit assert on the shape, so it is more readable in my opinion.

matteobettini:
What was the problem with the previous way? Why was it a bug?

MarkHaoxiang:
Sure, I've changed both to use the multi-line version but with the view

vmoens (approved):
LGTM! Thanks for this

PR #1482, opened 2023-09-02
[BugFix] Fix Gym Categorical/One-hot issues

vmoens:
## Description

Makes sure that one-hot and categorical work ok when a gym env expects an integer.

### DIFF
--- a/torchrl/envs/libs/gym.py
+++ b/torchrl/envs/libs/gym.py
@@ -419,6 +419,16 @@ class GymWrapper(GymLikeEnv):
             env = self._build_gym_env(env, pixels_only)
         return env
 
+    def read_action(self, action):
+        action = super().read_action(action)
+        if (
+            isinstance(self.action_spec, (OneHotDiscreteTensorSpec, DiscreteTensorSpec))
+            and action.size == 1
+        ):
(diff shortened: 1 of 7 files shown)


torchrl/envs/libs/gym.py
```diff
@@ -419,6 +419,16 @@ def _build_env(
...
         return env
 
+    def read_action(self, action):
+        action = super().read_action(action)
+        if (
+            isinstance(self.action_spec, (OneHotDiscreteTensorSpec, DiscreteTensorSpec))
+            and action.size == 1
+        ):
+            # some envs require an integer for indexing
+            action = int(action)
```
mjlaali, line 429: tiny: You may want to use action.item() as it is more efficient 
scalar = np.array([1])
timeit.timeit(lambda: scalar.item())
> 0.05105495895259082
timeit.timeit(lambda: int(scalar))
> 0.34526545903645456

Moreover, it seems this functionality will be deprecated in future:
> DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)

mjlaali (approved):

PR #1484, opened 2023-09-04
[Feature] VMAS: choose between categorical or one-hot actions

matteobettini:
Ths PR introduces a flag when constructing vmas environments that allows users to choose whether they want one hot or categorical actions when the actions are discrete.

Defaults to categorical for bc-comaptibility


vmoens (approved):
LGTM thanks!

PR #1485, opened 2023-09-04
[BugFix] Nested key in replay buffer

matteobettini:
Added the flexibility to use a nested key in the replay buffer for priority


vmoens (approved):
LGTM Thanks!

PR #1487, opened 2023-09-05
[BugFix] Fix DDPG squeezing

### LINKED ISSUE #1486: [BUG] Unsqueezing the error in ddpg
https://github.com/pytorch/rl/blob/3595c45eeb0c3e45c41bfff99ecbc973a7d1704f/torchrl/objectives/ddpg.py#L272

this line is not compatible with multi-agent, as it unsuquezees a dimension before the agents one.

Is there a workaround? is a `unsqueeze(-1)` sufficient?

matteobettini:
There is a line in DDPG which unsqueezes according to the batch_size of a tensordict which is not compatible when the tensor has a multiagent dimension outside the tesnordict batch_size

Fixes #1486


matteobettini:
ok what kind of issue do you think I should try to look for?

vmoens (approved):

PR #1494, opened 2023-09-06
[BugFix] Multiagent "auto" entropy fix in SAC

matteobettini:
In SAC, when the entropy target is set to "auto", it is computed using the shape of the action.
In multi-agent settings this shape included the number of agents, which should not be the case.

This PR fixes that

### DIFF
--- a/torchrl/objectives/sac.py
+++ b/torchrl/objectives/sac.py
@@ -410,11 +410,9 @@ class SACLoss(LossModule):
                 else:
                     action_container_shape = action_spec.shape
                 target_entropy = -float(
-                    np.prod(
-                        action_spec[self.tensor_keys.action].shape[
+                    action_spec[self.tensor_keys.action].shape[
                             len(action_container_shape) :
-                        ]
-                    )
+                    ].numel()
                 )
             self.register_buffer(
                 "target_entropy_buffer", torch.tensor(target_entropy, device=device)
--- a/torchrl/objectives/sac.py
+++ b/torchrl/objectives/sac.py
@@ -10,15 +10,14 @@ from typing import Dict, Optional, Tuple, Union
 
 import numpy as np
 import torch
+
 from tensordict.nn import dispatch, make_functional, TensorDictModule
 from tensordict.tensordict import TensorDict, TensorDictBase
 from tensordict.utils import NestedKey
 from torch import Tensor
-
 from torchrl.data import CompositeSpec, TensorSpec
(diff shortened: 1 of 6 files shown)


torchrl/objectives/sac.py
```diff
@@ -401,8 +400,21 @@ def target_entropy(self):
...
+                    ].shape
+                else:
+                    action_container_shape = action_spec.shape
                 target_entropy = -float(
-                    np.prod(action_spec[self.tensor_keys.action].shape)
+                    np.prod(
+                        action_spec[self.tensor_keys.action].shape[
+                            len(action_container_shape) :
+                        ]
+                    )
```
vmoens, line 417: since we're at it we can do
```suggestion
                    action_spec[self.tensor_keys.action].shape[
                            len(action_container_shape) :
                    ].numel()
```

vmoens (commented):
Shall we also address this in other algos?
TD3, CQL, REDQ, DTs?

matteobettini:
> Shall we also address this in other algos? TD3, CQL, REDQ, DTs?

done

vmoens (commented):
we could run a quick check in the test by passing composite actions to the losses and look at the target entropy?
Also: what I understand is that the target entropy is the number of agents.
Why is it that for single agent the entropy is the size of the action (eg, an action of size 100 has a very low target entropy) but for batched actions (not always MA but sometimes just composite actions) the target entropy is the number of actions?
How do we compute the entropy? Sum of the entropies?

I don't see the math behind these changes

matteobettini:
> Also: what I understand is that the target entropy is the number of agents. Why is it that for single agent the entropy is the size of the action (eg, an action of size 100 has a very low target entropy) but for batched actions (not always MA but sometimes just composite actions) the target entropy is the number of actions? How do we compute the entropy? Sum of the entropies?
> 
> I don't see the math behind these changes

The purpose of this change is to exclude the agents from the entropy calculation.

there are no changes in the math here.
The target entropy is always computed using the number of actions.
If i have an action with shape [3] it is 3, if i have a multidimnesional action [3,4] it is 12.

Therefore, we need to remove the batch size before computing this multiplication, otherwise the batch size will influence it.
In this case i am removing the batch size of the composite spec containing the action from the action itself (which removes the number of agents).

vmoens:
Oh ok my mistake I understood something else!

vmoens (approved):

PR #1495, opened 2023-09-06
[Algorithm] Update PPO examples

albertbou92:
## Description

Updated PPO examples. Now the scripts reproduce the results from Atari and MuJoCo environments in the original PPO paper.
Some common improvements are added (like computing the advantage at every epoch)

## Motivation and Context

## Types of changes

## Checklist


albertbou92:
This is my suggestion to update the PPO examples. I have 1 script for MuJoCo and one for Atari since the architectures and the env transforms are different. I think it is easier to read, but we could also have a single script since the rest is almost the same. Maybe we could all review it as @matteobettini suggested and agree on a template for the other examples? what do you think? @vmoens @BY571

vmoens:
I think it's great. Having 2 separates files is fine, it's better to have clarity than highly engineered, poorly tested and unreadable single scripts :) .
We can (if you want) write a `.md` file in the directory that says that we have 2 files for clarity, even though it's self explanatory...

examples/ppo/ppo_mujoco.py
```diff
@@ -0,0 +1,179 @@
...
+                actor_optim.zero_grad()
+                critic_optim.zero_grad()
+
+        losses = losses.apply(lambda x: x.float().mean(), batch_size=[])
+        for key, value in losses.items():
+            logger.log_scalar(key, value.item(), collected_frames)
+
+        # Test logging
+        with torch.no_grad(), set_exploration_type(ExplorationType.MODE):
+            if (collected_frames - frames_in_batch) // cfg.logger.test_interval < (
```
matteobettini, line 152: can't you use `j` here?

examples/ppo/ppo_mujoco.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # Update the data buffer
+        data_buffer.extend(data_reshape)
+
+        losses = TensorDict({}, batch_size=[cfg.loss.ppo_epochs, num_mini_batches])
+        for j in range(cfg.loss.ppo_epochs):
+
+            for i, batch in enumerate(data_buffer):
+
+                # Linearly decrease the learning rate and clip epsilon
```
matteobettini, line 119: we are not annealing epsilon here

examples/ppo/ppo_mujoco.py
```diff
@@ -0,0 +1,179 @@
...
+
+        frames_in_batch = data.numel()
+        collected_frames += frames_in_batch
+        pbar.update(data.numel())
+
+        # Train loging
+        episode_rewards = data["next", "episode_reward"][data["next", "done"]]
+        if len(episode_rewards) > 0:
+            logger.log_scalar(
+                "reward_train", episode_rewards.mean().item(), collected_frames
```
matteobettini, line 103: i usually use names like "train/reward", this will make wandb automatically divide them into different panels.

here is the stuff i log for training
https://github.com/pytorch/rl/blob/147de71d090d5705182bfabd24a99f3b2ee4cec9/examples/multiagent/utils/logging.py#L76

and this some for eval
https://github.com/pytorch/rl/blob/147de71d090d5705182bfabd24a99f3b2ee4cec9/examples/multiagent/utils/logging.py#L122

matteobettini: i think we at least need to time the scripts and log times both for collection and training

examples/ppo/ppo_mujoco.py
albertbou92: doing it implicitly since collected_frames is the "number of batches collected" * frames_in_batch.
I can make it more explicit:

`if (i - 1) * frames_in_batch % test_interval < i * frames_in_batch % test_interval:`

examples/ppo/ppo_mujoco.py
albertbou92: It is not done for mujoco, but to be consistent I can add the option add set it to False in config

examples/ppo/ppo_mujoco.py
albertbou92: Makes sense to use this kind of naming. Probably in each example the metrics will vary a bit, but at least we can agree to make them always "train/..." and "eval/...".

Adding the timing also makes sense

examples/ppo/ppo_mujoco.py
matteobettini: why not `i % cfg.eval.evaluation_interval`?

matteobettini: ah because your test interval is in frames, then `if collected_frames_till_now % test_interval == 0`?

vmoens:
@albertbou92 we also need to update the examples CI (cc @BY571)

examples/ppo/ppo_mujoco.py
```diff
@@ -0,0 +1,209 @@
+"""
```
vmoens, line 1: missing headers

examples/ppo/ppo_atari.py
```diff
@@ -0,0 +1,227 @@
+"""
```
vmoens, line 1: missing headers

examples/ppo/utils_atari.py
```diff
@@ -0,0 +1,212 @@
+import gym
```
vmoens, line 1: missing headers

examples/ppo/utils_mujoco.py
```diff
@@ -0,0 +1,116 @@
+import gym
```
vmoens, line 1: missing headers

vmoens, line 1: Since you're using HalfCheeta-v4, we can use gymnasium no?
This version of halfcheetah with the gym from the CI makes it crash.
The CI uses gym 0.23 for D4RL compatibility, and gymnasium for newer stuff. 
(welcome to gym wonderland)

albertbou92: I used HalfCheetaf-v4 in case the CI env did not have MuJoCo. The default config has HalfCheetah-v3. 
Should we use gymnasium instead of gym then?

albertbou92:
for atari, we can now speedup training with the new vectorised envs right?

vmoens:
I think, I can check

albertbou92:
> @albertbou92 the examples test [are failing](https://github.com/pytorch/rl/actions/runs/6232631101/job/16916184230?pr=1495#step:11:5946)

solved!  also for A2C @vmoens

vmoens (approved):
LGTM!

PR #1497, opened 2023-09-06
[Refactor] Refactor multi-agent MLP

vmoens:
cc @matteobettini 

By the way, rather than doing a multi-agent MLP, a multi-agent ConvNet, a multi agent LSTM etc, should we not consider one way of batching ops in whichever nn.Module for multiagent and (maybe) just subclass that for every class?

Seems a bit repetitive no? Isn't there pare of the logic we can recycle?

TODO:
 - [ ] Use vmap and module ensemble for multiple nets


torchrl/modules/models/multiagent.py
```diff
@@ -158,22 +162,38 @@ def __init__(
...
+            self.params = TensorDictParams(TensorDict.from_module(net), no_convert=True)
+        make_functional(net)
+        self.net = net
+        if self.centralised:
+            self.net_call = torch.vmap(
+                self.net,
+                (None, 0)
+            ) if not self.share_params else self.net
+        else:
+            self.net_call = torch.vmap(self.net, (-2, 0)) if not self.share_params else self.net
```
matteobettini, line 196: Does this work and pass the test?

Cause it seems to me that we are not doing the resahping into (`n_agents` * `n_inputs_per_agent`) in the centralized case

vmoens: No it doesn't. In fact it's not ready for review :)

vmoens: But feel free to comment on the general idea from the PR description!

matteobettini:
Because rnns in torchrl are TensorDictModules no?
In fact we do not have them in the `models`.

Yeah i think a common multiagent nn module can makes sense, we need to be parametric to the multiagent dim but it should be possible

torchrl/modules/models/multiagent.py
```diff
@@ -158,22 +162,38 @@ def __init__(
...
+                **kwargs,
+            )
+
+        if not self.share_params:
+            agent_networks = [
+                make_net()
+                for _ in range(self.n_agents)
             ]
-        )
+            self.params = TensorDictParams(torch.stack([TensorDict.from_module(mod) for mod in agent_networks], 0).contiguous(), no_convert=True)
```
matteobettini, line 183:
```suggestion
            self.params = TensorDictParams(torch.stack([TensorDict.from_module(make_net()) for _ in range(self.n_agents)], 0).contiguous(), no_convert=True)
```

vmoens:
> Because rnns in torchrl are TensorDictModules no?
> In fact we do not have them in the models.

That can be easily worked around, it's up to us to adapt our design.

torchrl/modules/models/multiagent.py
```diff
@@ -158,22 +163,45 @@ def __init__(
...
+        if self.centralised:
+            if not self.share_params:
+                self.net_call = torch.vmap(self.net, in_dims=(None, 0), out_dims=(-2,))
+            else:
+                self.net_call = self.net
+        else:
+            if not self.share_params:
+                self.net_call = torch.vmap(self.net, in_dims=(-2, 0), out_dims=(-2,))
+            else:
+                self.net_call = torch.vmap(self.net, in_dims=(-2, None), out_dims=(-2,))
```
matteobettini, line 202: is this neded? wouldn't the mlp do vectorization by default?

matteobettini, line 202:
```suggestion
        if self.share_params:
            self.net_call = self.net
        else:
            if self.centralised:
                self.net_call = torch.vmap(self.net, in_dims=(None, 0), out_dims=(-2,))
           else:
                self.net_call = torch.vmap(self.net, in_dims=(-2, 0), out_dims=(-2,))
```

vmoens: Not sure all ops will support an arbitrary num of dims

matteobettini: oh ok let's revert to previous version then and we are good to go

vmoens:
The mappo test fails
I'm investigating that

vmoens:
I think this is a good use case for first class dimension, as the bug is a bit convoluted to solve. Let me work on this and put this on hold

kfu02:
Hi, I am just wondering if this issue is still being investigated, as I would like to try RNNs in the multi-agent setting (specifically for [MAPPO](https://arxiv.org/pdf/2103.01955.pdf), as that paper claims better performance for RNNs vs MLPs in many settings). Apologies if this is not the right place to ask.

vmoens:
This is definitely a good place to ask. I can refresh and merge this PR now, and make a CNN/RNN version too. Bear with me...

kfu02:
Great! I eagerly await the refresh/merge of this PR!

vmoens:
@kfu02 

Check out https://github.com/pytorch/rl/pull/1921
It should work faster and allow us to code RNNs in a successive PR.
The issue right now is just that we can't account for non initialized lazy params as they are gathered in a single TensorDict during construction.

The solution will be a bit convoluted but doable :)

PR #1500, opened 2023-09-06
[Feature] ClipTransform + rename `min/maximum` -> `low/high`

vmoens:
Clips observations, actions and rewrards.

- [x] tests to be implemented
- [x] docstrings

cc @albertbou92 

We also rename maximum and minimum in the specs to low and high and raise deprecation warnings when necessary.


albertbou92:
LGTM!

albertbou92 (approved):

PR #1504, opened 2023-09-07
[CI] Add macOS M1 binaries Wheels

DanilBaibak:
## Description

Add macOS M1 binaries Wheels


vmoens (commented):
Thanks for this @DanilBaibak 
This will be blocked as long as `pytorch-labs` does not have access to the M1 machines to build its own binaries I'm afraid. We can't have a pypi version of binaries for torchrl if the corresponding tensordict version isn't available.

vmoens:
@DanilBaibak What is the status of pytorch-labs and GHA runners? Any progress?

vmoens (approved):
LGTM

PR #1505, opened 2023-09-07
[Refactor] Refactor multi-sync collectors

vmoens:
## Description

Refactors collectors:
- MultiSyncCollector can (and should) concatenate data along time dimension but the old behaviour is kept for consistency

Because preemption now requires some reshaping, this could break with LazyStacked tensordicts. A warning should be raised when these two things are used jointly.


gtnv:
hey @vmoens, do you want to keep this open for more work? if not, should we close it?

PR #1517, opened 2023-09-12
[Feature] Log hparams from python dict

matteobettini:
The previous call to `var` made the `log_hparams` function not compatible with python dictionaries


vmoens:
I wonder why vecnorm is suddenly breaking on 3.11...

torchrl/record/loggers/csv.py
```diff
@@ -128,7 +128,7 @@ def log_hparams(self, cfg: "DictConfig") -> None:  # noqa: F821
         Args:
             cfg (DictConfig): The configuration of the experiment.
```
vmoens, line 129: We need to edit the dosctrings

vmoens (commented):
I know we currently don't test that feature but it should be integrated in the tests no?

matteobettini:
done

test/test_loggers.py
```diff
@@ -98,6 +110,13 @@ def test_log_video(self, steps, tb_logger):
                 step=steps[i] if steps else None,
             )
 
+    def test_log_hparams(self, tb_logger):
+        config = get_example_config()
+        del config["nested"]  # not supported in tensorboard
+        del config["list"]  # not supported in tensorboard
+        del config["tuple"]  # not supported in tensorboard
```
vmoens, line 117: what is the error message? Is it explicit on their end? Otherwsie we may want to capture it and test it (ie test that we get a good error message whenever a nested dict is passed)

test/test_loggers.py
```diff
@@ -177,6 +196,19 @@ def test_log_histogram(self):
                 data = torch.randn(10)
                 logger.log_histogram("hist", data, step=0, bins=2)
 
+    def test_log_config(self, tmpdir):
+        torch.manual_seed(0)
+
+        exp_name = "ramala"
```
vmoens, line 202: ding dong

test/test_loggers.py
```diff
@@ -36,6 +36,18 @@ def tb_logger(tmp_path_factory):
     del logger
 
 
+def get_example_config():
```
vmoens, line 39:
```suggestion
@fixture
def config():
```
oc we can change the name

torchrl/record/loggers/tensorboard.py
```diff
@@ -2,13 +2,18 @@
...
+from typing import Dict, Sequence, Union
 
 from torch import Tensor
 
 from .common import Logger
 
+_has_omgaconf = importlib.util.find_spec("omegaconf") is not None
+if _has_omgaconf:
+    from omegaconf import OmegaConf
+
```
vmoens, line 16: Can we have a local import instead?

test/test_loggers.py
```diff
@@ -98,6 +110,13 @@ def test_log_video(self, steps, tb_logger):
                 step=steps[i] if steps else None,
             )
 
+    def test_log_hparams(self, tb_logger):
+        config = get_example_config()
```
vmoens, line 114:
```suggestion
    def test_log_hparams(self, tb_logger, config):
```
if we use a fixture

test/test_loggers.py
```diff
@@ -177,6 +196,19 @@ def test_log_histogram(self):
                 data = torch.randn(10)
                 logger.log_histogram("hist", data, step=0, bins=2)
 
+    def test_log_config(self, tmpdir):
```
vmoens, line 199:
```suggestion
    def test_log_config(self, tb_logger, config):
```
if we use a fixture

test/test_loggers.py
```diff
@@ -177,6 +196,19 @@ def test_log_histogram(self):
...
+        torch.manual_seed(0)
+
+        exp_name = "ramala"
+        logger = CSVLogger(log_dir=tmpdir, exp_name=exp_name)
+        config = get_example_config()
```
vmoens, line 204: to remove (if fixture)
```suggestion
```

test/test_loggers.py
```diff
@@ -243,6 +275,14 @@ def test_log_video(self, wandb_logger):
                 video=video_wrong_format,
             )
 
+    def test_log_hparams(self, wandb_logger):
+        config = get_example_config()
```
vmoens, line 279:
```suggestion
    def test_log_hparams(self, wandb_logger, config):
```

test/test_loggers.py
```diff
@@ -326,6 +366,11 @@ def test_log_histogram(self, mlflow_fixture):
             data = torch.randn(10)
             logger.log_histogram("hist", data, step=0, bins=2)
 
+    def test_log_hparams(self, mlflow_fixture):
+        logger, client = mlflow_fixture
+        config = get_example_config()
```
vmoens, line 371:
```suggestion
    def test_log_hparams(self, mlflow_fixture, config):
        logger, client = mlflow_fixture
```

vmoens (approved):
LGTM, a few suggestions

test/test_loggers.py
matteobettini: it is quite good,
it will say that you tried to log a param of type {type} and that is not supported in {logger_name}

test/test_loggers.py
matteobettini: is this a reference to something?

matteobettini: i see, my culture grew today

vmoens (approved):

PR #1520, opened 2023-09-12
[Refactor] Dreamer v1 refactor

MateuszGuzek:
## Description
Draft PR to facilitate discussion:

The proposed change extracts training steps in wrappers to reduce length of the example, but as there few control logging parameters needed, it creates nasty large function signatures, which are split multi-line by µfmt and do not really achieve the goal IMHO. (If such change would be accepted, the wrappers can be clearly moved to utils or models file.)

@vmoens @BY571 

## Motivation and Context

Discuss ideas for refactoring Dreamer V1.

## Types of changes

Refactor

## Checklist


gtnv:
hi @MateuszGuzek, is this implementation still active? if not, can we close it for now?

PR #1523, opened 2023-09-14
[Algorithm] Update TD3 Example

BY571:
## Description

Updated TD3 script similar to [PPO Update](https://github.com/pytorch/rl/pull/1495), added time logging, more comments,  cleaner structure, and fixes here and there. Running some tests right now to verify performance.
What do you think @vmoens, @albertbou92 @matteobettini, how could we improve the example further?

## Motivation and Context

## Types of changes

## Checklist


examples/td3/config.yaml
```diff
@@ -1,47 +1,49 @@
-# Environment
+# task and env
 env:
   name: HalfCheetah-v3
   task: ""
-  exp_name: "HalfCheetah-TD3"
+  exp_name: "HalfCheetah-TD3-ICLR"
```
vmoens, line 5: Can we avoid any specific reference to a submission that we may forget to remove?

examples/td3/config.yaml
```diff
@@ -1,47 +1,49 @@
...
-# Collection
+# collector
 collector:
-  total_frames: 1000000
-  init_random_frames: 10000
+  total_frames: 3000000
+  init_random_frames: 25_000
   init_env_steps: 1000
-  frames_per_batch: 1000
+  frames_per_batch: 1
```
vmoens, line 15: this is going to be super inefficient, do we have to stick to that?
can we obtain the same perf with something faster?

examples/td3/td3.py
```diff
@@ -45,140 +47,156 @@ def main(cfg: "DictConfig"):  # noqa: F821
...
+                # Compute loss
                 loss_td = loss_module(sampled_tensordict)
 
                 actor_loss = loss_td["loss_actor"]
                 q_loss = loss_td["loss_qvalue"]
 
+                # Update critic
                 optimizer_critic.zero_grad()
-                update_actor = j % delayed_updates == 0
+                update_actor = update_counter % delayed_updates == 0
```
vmoens, line 131: let's move this upward, under update_counter (or move update_counter down)

examples/td3/td3.py
```diff
@@ -45,140 +47,156 @@ def main(cfg: "DictConfig"):  # noqa: F821
...
-            "train_reward": rewards[-1][1],
-            "collected_frames": collected_frames,
-        }
-        if q_loss is not None:
-            train_log.update(
-                {
-                    "actor_loss": np.mean(actor_losses),
-                    "q_loss": np.mean(q_losses),
-                }
+        training_time = time.time() - training_start
```
vmoens, line 151: We should not need that (if we're using wandb or any other logger) but happy to keep it if that makes logging the time spent clearer

vmoens (approved):
LGTM! Some minor comments

examples/td3/td3.py
BY571: I think it gives a better overview of how much time is spent at each big step (sample, update, and test) in the training loop. But I'm also okay with removing it as you said wandb and other logger keep track of timing themselves.

examples/td3/config.yaml
BY571: Performance might be a bit less stable, I'm testing it right now.

examples/td3/td3.py
vmoens: we can keep that no big deal

examples/td3/utils.py
vmoens, line 1: because we're using gym and not gymnasium, we should decorate calls to GymEnv with `with set_gym_backend("gym")`

Also, the headers are missing

torchrl/objectives/td3.py
vmoens, line 1: If we're refactoring td3, maybe we can use the changed from #1377?

BY571: I updated the loss like in #1377

BY571: Forgot to push, should be updated now. Also started some tests to check performance. CI test should be also up-to-date

vmoens: Got it
The TD3 tests are failing so we should investigate what is going on

test/test_cost.py
```diff
@@ -2357,8 +2357,10 @@ def test_td3_notensordict(self, observation_key, reward_key, done_key):
             loss_val_td = loss(td)
             torch.manual_seed(0)
             loss_val = loss(**kwargs)
-            for i, key in enumerate(loss_val_td.keys()):
-                torch.testing.assert_close(loss_val_td.get(key), loss_val[i])
+            for i in loss_val:
+                assert i in loss_val_td.values(), f"{i} not in {loss_val_td.values()}"
+            # for i, key in enumerate(loss_val_td.keys()):
+            # torch.testing.assert_close(loss_val_td.get(key), loss_val[i])
```
BY571, line 2363: This is dangerous as keys in the tensordict get ordered by name but output tuple loss_val doesn't. So for now im just checking if all values in the loss_val tuple are also in the loss_val_td.


```
# actor metadata
metadata = {
            "state_action_value_actor": state_action_value_actor.mean().detach(),
        }
# value metadata
        metadata = {
            "td_error": td_error,
            "next_state_value": next_target_qvalue.mean().detach(),
            "pred_value": current_qvalue.mean().detach(),
            "target_value": target_value.mean().detach(),
        }
# out tensordict
        td_out = TensorDict(
            source={
                "loss_actor": loss_actor,
                "loss_qvalue": loss_qval,
                **metadata_actor,
                **metadata_value,
            },
            batch_size=[],
        )
```
loss_vals will be in that order `(loss_actor, loss_qvalue, state_action_value_actor, next_state_value, pred_value, target_value)`
However, as the items are getting ordered in the TD by the keys the output tensordict has actually this order:
 `(loss_actor, loss_qvalue, next_state_value, pred_value, state_action_value_actor, target_value)`

vmoens: dispatch returns the keys in the order of out_keys.
So this is predictable, we can just do
```python
            for i, key in enumerate(loss.out_keys):
                torch.testing.assert_close(loss_val_td.get(key), loss_val[i])
```
does that solve the problem?

vmoens (approved):
LGTM
let's wait for the tests to pass!

PR #1526, opened 2023-09-14
[Refactor] Dreamer loss modules

MateuszGuzek:
## Description

Refactor of the dreamer loss. 
WIP, TODO: forward function.

## Motivation and Context

Requested by @vmoens per popular request 

## Types of changes

## Checklist


MateuszGuzek:
When you now run example, you encounter the error :
`
  File "./rl/examples/dreamer/dreamer.py", line 321, in main
    scaler_actor.scale(actor_loss_td["loss_actor"]).backward()
  File "~/anaconda3/envs/rl_dreamer_3_10/lib/python3.10/site-packages/torch/_tensor.py", line 487, in backward
    torch.autograd.backward(
  File "~/anaconda3/envs/rl_dreamer_3_10/lib/python3.10/site-packages/torch/autograd/__init__.py", line 200, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.FloatTensor [400, 1]], which is output 0 of AsStridedBackward0, is at version 2; expected version 1 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
`
I investigated and this happens always when backprop and optimization steps are applied to the results of the world model. Other losses or scaling does not cause the issue. torch.autograd.set_detect_anomaly(True) did not return anything interesting.

When each model loss is applied separately (like in the previous commit of this pull request) all works well.

Does it mean that there is some sharing of parameters that is triggered only in forward, but not in the standard call of the new loss functions? @vmoens

torchrl/objectives/dreamer.py
```diff
@@ -111,11 +140,59 @@ def __init__(
...
+        # result_tensordict.update(actor_tensordict) # Ultimately we could all lossed together everything,
+        # result_tensordict.update(value_tensordict) # but for the sake of simplicity of demonstration I pass the result as raw as possible
+        return (
+            world_tensordict,
+            actor_tensordict,
+            value_tensordict,
+            sampled_tensordict_world,
+            sampled_tensordict_actor,
+            sampled_tensordict_value,
+        )
```
vmoens, line 184: these should be nested in a single tensordict.
We usually put things together like this:
```python
TensorDict({
   "loss_actor": val,
   "loss_value": other_val,
    ....}, [])
```

vmoens (commented):
I think the error that you're getting is due to the fact that you're computing all losses all at once but then you're backpropagating once at a time. This means that you're changing the graph in between
In other words: when you build a graph, the tensors are stored as intermediate values. Not to overload the memory, it's a pointer to these tensors (eg, the parameters) that is stored.
Once you call backward on the first loss and change the value of the related params, you are changing values in the graph. Pytorch tracks that and tells you that this isn't expected and it can't compute the second backward pass because the intermediate values (the parameters) have changed. What you should do is either compute each loss independently or sum the losses and do a single backward.

torchrl/objectives/dreamer.py
```diff
@@ -111,11 +139,39 @@ def __init__(
...
+            self.gamma = gamma
+        if lmbda is not None:
+            warnings.warn(_GAMMA_LMBDA_DEPREC_WARNING, category=DeprecationWarning)
+            self.lmbda = lmbda
+        # value
+        self.value_loss_function = value_loss if value_loss is not None else "l2"
+        self.value_gamma = value_gamma
+        self.value_discount_loss = value_discount_loss
 
     def _forward_value_estimator_keys(self, **kwargs) -> None:
```
MateuszGuzek, line 159: Model and Value had pass here, and actor has this longer implementation. Do we need it here?

torchrl/objectives/dreamer.py
```diff
@@ -111,11 +139,39 @@ def __init__(
...
-        pass
+        if self._value_estimator is not None:
+            self._value_estimator.set_keys(
+                value=self._tensor_keys.value,
+            )
```
MateuszGuzek, line 163: Model and Value had pass here, and actor has this longer implementation. Do we need it here?

gtnv:
hi @MateuszGuzek, is this still being worked on? if not, should we close it for now?

PR #1529, opened 2023-09-14
[CI] Fix CI

vmoens:
Some CI fixes
- Windows does not find temp directory when closing it
- public mjkey is gone, now uploaded in pytorch S3

### DIFF
--- a/torchrl/envs/libs/pettingzoo.py
+++ b/torchrl/envs/libs/pettingzoo.py
@@ -138,7 +138,9 @@ class PettingZooWrapper(_EnvWrapper):
 
     git_url = "https://github.com/Farama-Foundation/PettingZoo"
     libname = "pettingzoo"
-    available_envs = _get_envs()
+    @property
+    def available_envs(self):
+        return _get_envs()
 
     def __init__(
         self,
--- a/torchrl/envs/libs/pettingzoo.py
+++ b/torchrl/envs/libs/pettingzoo.py
@@ -18,8 +18,12 @@ from torchrl.data import (
 )
 from torchrl.envs.common import _EnvWrapper
 from torchrl.envs.libs.gym import _gym_to_torchrl_spec_transform, set_gym_backend
-from torchrl.envs.utils import _replace_last, check_marl_grouping, \
-    MarlGroupMapType, _classproperty
+from torchrl.envs.utils import (
+    _classproperty,
+    _replace_last,
+    check_marl_grouping,
+    MarlGroupMapType,
+)
 
 _has_pettingzoo = importlib.util.find_spec("pettingzoo") is not None
 
--- a/torchrl/envs/libs/pettingzoo.py
+++ b/torchrl/envs/libs/pettingzoo.py
@@ -18,8 +18,8 @@ from torchrl.data import (
 )
 from torchrl.envs.common import _EnvWrapper
 from torchrl.envs.libs.gym import _gym_to_torchrl_spec_transform, set_gym_backend
-from torchrl.envs.utils import _replace_last, check_marl_grouping, MarlGroupMapType
-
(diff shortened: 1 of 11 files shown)


torchrl/envs/libs/pettingzoo.py
```diff
@@ -138,7 +138,10 @@ class PettingZooWrapper(_EnvWrapper):
 
     git_url = "https://github.com/Farama-Foundation/PettingZoo"
     libname = "pettingzoo"
-    available_envs = _get_envs()
+
+    @property
+    def available_envs(self):
+        return _get_envs()
```
matteobettini, line 144: for uniformity we could do this in all libs

matteobettini (approved):
LGTM

torchrl/envs/libs/pettingzoo.py
matteobettini: also this makes it an object property, but we might want it as a class property as users will have to know whats available before instantiating

vmoens: Yes that's right!

vmoens: I made it a generator in all cases such that one can do
```python
env_generator = MyEnvClass.available_envs()
```
but that won't call anything until you iterate over it. In some cases it can avoid importing the lib.
The only problem with that is when you do
```python
env_generator = MyEnvClass.available_envs()
for name in env_generator:
    print(name)
for name in env_generator:
    # never reached
    print(name)
```

Wdyt?

matteobettini: Personally I always just print it in one go, what is the need for iterating?
Also this would BC breaking.

What is the problem of a normal list? I think I am missing the reason why we are changing this

vmoens: the reason is just that I want to make extra sure that we don't import if we don't need it. For that a generator is good bc it's not executed unless you iterate over it.
Soon i'll make a PR where all envs are exposed in `torchrl.envs`. I want that to also mean that you can access the envs attributes without loading the related libraries. It is important because we may leave a `GymEnv.available_env` unattended somewhere in the code, and if we do that, every single subprocess launched will load gym. Same goes for Isaac, habitat, brax... imagine if we load them all when we run 64 workers! That also brings a high price to pay in the tests: eg
```
class TestGym:
    @pytest.mark.parametrize("stuff", GymEnv.available_envs)
    def mytest(self, stuff):
        ...
```
will load gym even if the test is ignored. I don't think it's optimal

vmoens: It is true it's bc-breaking but only mildly: you can still iterate. If the goal is to print you will soon realize you should put the result in a list. If you iterate (once) nothing changes. The issue is when you iterate twice and/or when you index (but I wonder who would index that).
By the way: generator makes more sense here to me also because indexing should be unimportant (GymEnv.available_envs[3] is meaningless)

vmoens: Yep
This is on my mcbook:
```python
>>> import time
>>> def import_gym():
...     t0 = time.time()
...     import gymnasium
...     print(time.time()-t0)
... 
>>> import_gym()
8.547016143798828
```

Now imagine we have a test run for collectors + gym. It spawns 2 workers, each having 2 envs in parallel.
So we wait at least 16 seconds just to launch the test, without counting execution. Imagine with the parametrization how fast it grows! We quickly end up with a single test taking > 5 minutes...

matteobettini: there might still be a few libs to update, like smac and vmas

vmoens: On it but it will be a separate PR

PR #1531, opened 2023-09-14
[NOMRG] Test skipping CI Smac

matteobettini:


matteobettini:
As you can see from the CI smac was skipped @vmoens

vmoens:
But it has the label so it shouldn't

PR #1532, opened 2023-09-15
[Refactor] Expose all wrappers in torchrl.envs

vmoens:
## Description

This PR ensures that:
- All env wrappers are available under `torchrl.envs`
- It also ensures that imports aren't executed unless explicitly asked for (eg during an env instantiation)


vmoens:
@matteobettini this should be ready for review when you have time

torchrl/envs/libs/__init__.py
```diff
@@ -2,3 +2,15 @@
...
+from .dm_control import DMControlEnv, DMControlWrapper
+from .envpool import MultiThreadedEnv, MultiThreadedEnvWrapper
+from .gym import GymEnv, GymWrapper
+from .habitat import HabitatEnv
+from .isaacgym import IsaacGymEnv, IsaacGymWrapper
+from .jumanji import JumanjiEnv, JumanjiWrapper
+from .openml import OpenMLEnv
+from .robohive import RoboHiveEnv
+from .smacv2 import SMACv2Env, SMACv2Wrapper
+from .vmas import VmasEnv, VmasWrapper
```
matteobettini, line 16: Pettingzoo?

torchrl/envs/libs/vmas.py
```diff
@@ -17,7 +17,7 @@
 )
 from torchrl.envs.common import _EnvWrapper, EnvBase
 from torchrl.envs.libs.gym import _gym_to_torchrl_spec_transform, set_gym_backend
-from torchrl.envs.utils import _selective_unsqueeze
+from torchrl.envs.utils import _classproperty, _selective_unsqueeze
 
 IMPORT_ERR = None
 try:
```
matteobettini, line 23: Global import.

Probably we can use the better module checking here and remove the global import.

Happy to do this, i think i can push here,

Let me know

matteobettini (approved):
LGTM a few comments

torchrl/envs/libs/__init__.py
vmoens: No! No petttingzoo!
I'm kidding good catch

torchrl/envs/libs/vmas.py
vmoens: Oh no that's ok i forgot that one

PR #1535, opened 2023-09-15
[Test] Skip tests in python 3.11

vmoens:
## Description

Skipping tests that involve nested multiprocessing in python 3.11

cc @albanD since you implemented the fix :)


matteobettini (approved):
LGTM

PR #1538, opened 2023-09-15
[BugFix] Fix SMAC-v2

vmoens:


matteobettini:
we need to set the flag and commit after that to run the ci

matteobettini (approved):
less gooo

PR #1539, opened 2023-09-15
[Feature] Gym compatibility: Terminal and truncated

vmoens:
Final list of changes part of this PR:

Add full_done_spec, full_observation_spec, full_action_spec, full_state_spec, full_reward_spec properties.
Make all envs have "done" and "terminated".
Add a _complete_done private method within envs to complete output data if done_keys are missing. Make this compatible with transforms to avoid duplicated calls to the method.
Full conversion to done_keys wherever done_key was used.
versioned reading helpers in GymWrapper to account for discrepancies between versions
terminated_or_truncated, a function that reads the data, writes the "_reset" keys and return a boolean indicating if anything is done.
Fix @implement_for, set_gym_backend and related + add tests
Fix envs tutorial and add info wrt new "done" attributes
Compatibility with batched environments
Fix transforms: VecGymEnv, Reward2GoTransform, RenameTransform, InitTracker, TimeMaxPool, SelectTransform, ExcludeTransform, StepCounter
Fix ModelBased envs, RoboHive, D4RL, RLHF
Adapt collectors to new logic
Better handling of workers refusing to close in collectors
Fix Gym tests across versions

cc @skandermoalla


vmoens:
@skandermoalla The current project is to keep `done` as it was and add `truncated`.
Most of our code base interprets `done` as `termination`, ie: in gym, you have `done = termination | truncation` but in torchrl you'd have `end_of_traj = done | truncation`.
This naming is a bit different than gym but the alternative would be to rename done in termination everywhere in the code.
With @matteobettini we thought that proceeding as we do now is less disruptive than renaming every single occurence of `done`. Happy to read your thoughts about this!

skandermoalla:
Thanks a lot for starting this and for pinging me!

IMO naming differences are fine as long as `done` and `truncation` consistently mean the Gymnasium `termination` and `truncation` everywhere else in TorchRL and that they're correctly used.

> Most of our code base interprets done as termination

So here "most" should be very carefully assessed. It would be terrible to have `done` mean `terminated` at some point and `end_of_traj` at some other point. I can see exceptions coming from the fact that TorchRL supports multiple environment libraries and some of them only return `end_of_traj`,  but I think there is no way to handle this underspecified signal and from the TorchRL perspective should be treated as `terminated`.

I think I sometimes confused the `done` in TorchRL for an `end_of_traj` signal. I don't recall exactly where, but that's to emphasize that it would be good to check this, especially in value estimators and transforms. (An `end_of_traj` signal would not allow to bootstrap correctly without further considering the `truncated` bit).

So maybe also making the docs more explicit about this definition would help a lot of people arriving from Gymnasium.

I'm also thinking it would be useful to allow the use `StepCounter(max_steps=N)` to mark a native time limit (with the `done` key) to adjust the finite horizon or an env, vs. to mark an external time limit (`truncation`). I think this is already possible by just changing the truncation key to done right? Will this be safe?
I don't know how Gymnasium differentiates between internal and external time limits to correctly set the truncation key.

skandermoalla (commented):
Adding an early review before disappearing until the ICLR deadline 🫠.

torchrl/data/utils.py
```diff
@@ -228,14 +228,15 @@ def _check_only_one_entry(
...
         else:
-            if not found_entry:
-                found_entry = True
-            else:
-                raise error
+            if key == "done":
+                if found_entry:
+                    raise error
+                else:
+                    found_entry = True
```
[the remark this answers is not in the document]
vmoens: @matteobettini I'm not even sure we need this function anymore. We wanted to avoid that people would have `"done_agent1"` and `"done_agent2"` and rather `("agent1", "done")` IIRC. 
Now we will allow `"done_spec"` to contain `"done"` and `"truncated"` at any level. Per se the `done_spec` could even be richer. So I don't think we need this check anymore: it's good practice to have only one done per level but I don't think that will break anywhere if you have more.

The only exception I see is computing the `"_reset"` signal which usually reflects `done | truncated`. In other words, `_reset` reflects somethings like
```
def _make_reset(data, done_spec):
    out = {}
    for key, item in done_spec.items():
        if isinstance(done_spec, CompositeSpec):
            out[key] = _make_reset(data.get(key), item)
        else:
            out["_reset"] = out.get("_reset", False) | data.get(key)
     return data.update(out)
```
This means tht if you put done_agent1 and done_agent2 at the same level you'll get a single reset for both. Not ideal but not really our business if the MARL API is well documented IMO.

In collectors we use `"done"` to compute the trajectory ids but this is an extra feature that could apply only in settings where done is at the root (or in other restricted settings). The trajectory id doesn't influence the collector behaviour, which only cares about the number of frames (number of elements of the root tensordict). 

For the rest, loss modules have parametric dones so I think we could rename `"done"` to `"merylstreep"` and they could still work. They will need refactoring to account for `"truncated"` anyway.

torchrl/envs/gym_like.py
```diff
@@ -133,18 +133,24 @@ def read_action(self, action):
...
+    def read_done(self, done, termination, truncation):
         """Done state reader.
 
-        Reads a done state and returns a tuple containing:
+        In TorchRL, done means that a trajectory is terminated (we do not support the
+        terminated signal). Truncated means the trajectory has been interrupted.
+
+        This method returns:
         - a done state to be set in the environment
         - a boolean value indicating whether the frame_skip loop should be broken
```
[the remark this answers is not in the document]
vmoens: yes I still need to go through the docstrings but thanks for reminding me!

torchrl/envs/gym_like.py
```diff
@@ -133,18 +133,24 @@ def read_action(self, action):
         """
         return self.action_spec.to_numpy(action, safe=False)
 
-    def read_done(self, done):
+    def read_done(self, done, termination, truncation):
         """Done state reader.
 
-        Reads a done state and returns a tuple containing:
+        In TorchRL, done means that a trajectory is terminated (we do not support the
+        terminated signal). Truncated means the trajectory has been interrupted.
```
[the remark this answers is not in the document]
vmoens: I agree it's confusing. Maybe
```
In torchrl, a `done` signal means that a trajectory is terminated (what is referred to as `termination` in gymnasium).
```

torchrl/envs/gym_like.py
```diff
@@ -133,18 +133,24 @@ def read_action(self, action):
...
         Args:
             done (np.ndarray, boolean or other format): done state obtained from the environment
-
+            termination (bool or None): termination signal.
+            truncation (bool or None): truncation signal.
         """
-        return done, done
+        if termination is not None:
+            return termination, truncation, done
+        return done, truncation, done
```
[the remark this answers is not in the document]
vmoens: Let me rephrase what you're saying here:
if older version of gym (<0.26) a truncation signal could be retrieved from the info. 
Currently, we don't use it so assuming that done=truncated will not work properly with value functions, is that right?

If so, I think that we should advise users of adopting a dedicated info reader to write the truncated signal, which will solve the problem.

torchrl/envs/gym_like.py
```diff
@@ -133,18 +133,24 @@ def read_action(self, action):
         """
         return self.action_spec.to_numpy(action, safe=False)
 
-    def read_done(self, done):
+    def read_done(self, done, termination, truncation):
```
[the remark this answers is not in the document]
vmoens: good point, it's getting a bit confusing!

torchrl/envs/gym_like.py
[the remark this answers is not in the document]
skandermoalla: Yep! That's clear!

torchrl/envs/gym_like.py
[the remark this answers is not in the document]
skandermoalla: > if older version of gym (<0.26) a truncation signal could be retrieved from the info.

Yes in the `TimeLimit.truncated` key if it happened. See https://gymnasium.farama.org/tutorials/gymnasium_basics/handling_time_limits/#solution.

> assuming that done=truncated will not work properly with value functions

Yes, and to be exact done here with the old gym versions would be `gym_done=terminated | truncated`. (It was an `end_of_traj`)

> I think that we should advise users of adopting a dedicated info reader to write the truncated signal, which will solve the problem.

Yes, that would be the correct way to record the `truncated` key and propagate it to the rest of TorchRL's components. However, the  `gym_done` key would still need to be corrected with `torchrl_done = gym_done and ~truncated` I don't how this could be done by the users.

torchrl/data/utils.py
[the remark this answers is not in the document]
matteobettini: So the situation is a little more complex, let me try to explain it here:

Since we decided that our naming convention is to have one "_reset" key per tensordict (due to the fact that if we had more we would have to change the name), then this key has to reflect the shape of  the "done"  in that tensordict.

If we allowed more then one leaf key in each td of the done_spec then the problem is what shape should the _reset reflect. We cannot do like vincent said cause the shapes might be arbitrary.

Therefore the convention established is that at least a key with the name "done" (and exactly done) should exist in each td of the done_spec and the reset will mimic its shape.

If now we want to change the convention and say that the constraint now is that all the values in the same tensordict of the done_spec should have the same shape, that is fine by me, but then we have to check and enforce that.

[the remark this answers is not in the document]
matteobettini: if we do like in this PR and assume to always have truncated in the done_spec, then we have to carefully adjust all the components of the codebase that create the _reset key.

Like here
https://github.com/pytorch/rl/blob/fe91c4fc0e7ec6bde969a4f25ea58b247ec4dc65/torchrl/collectors/collectors.py#L797

and here
https://github.com/pytorch/rl/blob/fe91c4fc0e7ec6bde969a4f25ea58b247ec4dc65/torchrl/envs/common.py#L1590

For example.

These line will currently create a `_reset` for each `done_spec.keys(True,True)`.
But if the `done_spec` now contains more than one leaf per td we have to change all this.

torchrl/envs/libs/gym.py
```diff
@@ -448,10 +452,33 @@ def __init__(self, env=None, categorical_action_encoding=False, **kwargs):
...
+        # once and for all. This is aimed at avoiding the need of decorating code
+        # with set_gym_backend + allowing for parallel execution (which would
+        # be troublesome when both an old version of gym and recent gymnasium
+        # are present within the same virtual env).
+        # These calls seemingly do nothing but they actually get rid of the @implement_for decorator.
+        # We execute them within the set_gym_backend context manager to make sure we get
+        # the right implementation.
+        with set_gym_backend(self.get_library_name(self._env)):
+            self._reset_output_transform = self._reset_output_transform
+            self._output_transform = self._output_transform
```
vmoens, line 481: @matteobettini 
giphy (10)

torchrl/data/utils.py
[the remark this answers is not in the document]
vmoens: > If now we want to change the convention and say that the constraint now is that all the values in the same tensordict of the done_spec should have the same shape, that is fine by me, but then we have to check and enforce that.

Per se, if someone writes an env that does not require `_reset` (which we use internally but is not exposed to users) it's fine. What I mean is that if a data structure has `"done_agent1"` of shape `X` and `"done_agent2"` of shape `Y`, computing a `"_reset"` key will fail but that only means that the env cannot be used within a collector.
Since we want to provide independent, modular components, I think it's a design choice that is ok not to enforce that all envs written by any user would fit in a collector. We need to make sure that the collector tells you specifically what is wrong with your env though.

What about letting envs have the `done_spec` they want, and let the collectors / parallel envs spot the potential problems during construction?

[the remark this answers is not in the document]
matteobettini: Yeah but also the rollout function uses the reset logic. So how would we treat that?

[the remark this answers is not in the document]
vmoens: Imagine someone wants to use TorchRL to write a multiagent env that does some job with the `"done_agentX"` structure.
I think it's cool if they do that and put that in their app. I don't think they should get an exception or a warning that tells them that they're going to be in a dead end when using a collector. Maybe they don't want to use the collectors.
Let's not prevent legitimate use of one class to prevent bug in another if the usage of the other is optional, wdyt?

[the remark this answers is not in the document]
matteobettini: Yes but vectorized envs will need to be fed a `_reset` flag upon reset in the `env.rollout` function.  so you are saying we will provide that with our best effort but if we cannot we just give a warning and say that in collector this will be an error?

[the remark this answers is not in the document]
vmoens: > Yeah but also the rollout function uses the reset logic. So how would we treat that?

Oh yeah good point.
This is a bit tricky: per se, we set a `_reset` in rollout and hope for the best but we don't enforce that it's being read anywhere.
The thing is that we don't really have any agency about what `reset(tensordict)` does under the hood. For ParallelEnv and similar we know what to do but in other cases I'm not 100% sure of what could be done generically.

But yeah, since we use `_reset` within the env class it makes sense that it's computed in a consistent way within the library. 
So what we would opt for is:
- during construction, walk through the specs and check that at each level, all done specs have a matching shape
- during execution, compute `_reset` as the combination of all `_reset` keys at each level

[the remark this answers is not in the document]
matteobettini: > during execution, compute _reset as the combination of all _reset keys at each level

->  as the combination of all **done_spec_keys** at each level

but yes that could work!

torchrl/envs/utils.py
```diff
@@ -728,3 +728,74 @@ def check_marl_grouping(group_map: Dict[str, List[str]], agent_names: List[str])
...
             raise ValueError(f"Agent {agent_name} not found in any group")
+
+
+def done_or_truncated(
+    data: TensorDictBase, full_done_spec=None, name="_reset"
+) -> TensorDictBase:
+    """Reads the done / truncated keys within a tensordict, and writes a new tensor where the values of both signals are aggregated.
+
+    The modification occurs in-place within the TensorDict instance provided.
+    This function can be used to compute the `"_reset"` signal in batched
```
matteobettini, line 739:
```suggestion
    This function can be used to compute the `"_reset"` signals in batched
```

torchrl/envs/utils.py
```diff
@@ -728,3 +728,74 @@ def check_marl_grouping(group_map: Dict[str, List[str]], agent_names: List[str])
...
+        tensor(True)
+    """
+    aggregate = None
+    has_entry = name in data.keys(isinstance(name, tuple))
+    if full_done_spec is None:
+        for key, item in data.items():
+            if not has_entry and key in ("done", "truncated"):
+                done = data.get(key, None)
+                if done is None:
+                    done = torch.zeros(
```
matteobettini, line 778: in case done is present and truncated not it seems like this will instantiate a full shaped truncated, can we avoid this?

matteobettini (approved):
LGTM thanks a lot fot this.

I left a few comment, i am interested in the new function, ping me when we adopt it in rollout and collector

torchrl/envs/gym_like.py
```diff
@@ -133,18 +133,40 @@ def read_action(self, action):
...
+
+        Returns: a tuple with 3 boolean values,
+            - a done state to be set in the environment.
+            - a truncated state, possibly None if no truncation is provided.
+            - a boolean value indicating whether the frame_skip loop should be broken.
 
         """
-        return done, done.any() if not isinstance(done, bool) else done
+        if termination is not None:
+            gym_done = termination
```
skandermoalla, line 160: This would be a torchrl `done` no?

torchrl/envs/gym_like.py
```diff
@@ -133,18 +133,40 @@ def read_action(self, action):
...
+            do_break = gym_done | truncation
+        else:
+            do_break = gym_done
+
+        return (
+            gym_done,
```
skandermoalla, line 166: torchrl `done`?

torchrl/envs/gym_like.py
```diff
@@ -133,18 +133,40 @@ def read_action(self, action):
...
+            - a boolean value indicating whether the frame_skip loop should be broken.
 
         """
-        return done, done.any() if not isinstance(done, bool) else done
+        if termination is not None:
+            gym_done = termination
+            do_break = gym_done | truncation
+        else:
+            do_break = gym_done
+
```
skandermoalla, line 164:
```suggestion
            done = gym_done
```

torchrl/envs/gym_like.py
skandermoalla, line 166:
```suggestion
            done,
```

torchrl/envs/gym_like.py
skandermoalla, line 160:
```suggestion
            done = termination
```

skandermoalla: I mean it's still correct, but just not to overload the variable if not returned directly (like in isaacgym for example)

vmoens: No you're right i did a replace all within the function
Thanks for the suggestions

torchrl/envs/batched_envs.py
```diff
@@ -586,6 +586,7 @@ def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
                         needs_resetting[i] += _reset[i].any()
                 else:
                     missing_reset = True
+                    # why do we break here?
```
vmoens, line 589: @matteobettini I don't undersetand why we break here.
We could have a reset that is missing somewhere (in which case we fall back on reset being an aggregate of the dones at this level) but other reset being present.
The risk is that someone partially populates the resets hoping that this will be passed to the sub-envs and then everything is silently reset.

torchrl/envs/common.py
```diff
@@ -1264,24 +1260,6 @@ def reset(
...
-                        -len(self.output_spec["full_done_spec"][done_key].shape) :
-                    ]
-                    != self.output_spec["full_done_spec"][done_key].shape
-                ):
-                    raise RuntimeError(
-                        "_reset flag in tensordict should follow env.done_spec"
-                    )
-                _reset_map.update({done_key: _reset})
-        else:
-            _reset_map = {}
```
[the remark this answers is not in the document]
matteobettini: this came from a previous version where we were doing a check on the _reset shape before calling the inner reset, and while we were at it here i was caching the reset flags for the check later.

also are we assuming that `tensordict` is not modified after the call to `_reset()` is this always the case?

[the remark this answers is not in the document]
vmoens: We're assuming that `_reset` leaves the tensordict untouched if that's the question. 
Can we remove this block?

torchrl/envs/common.py
```diff
@@ -1585,20 +1569,14 @@ def policy(td):
...
-                truncated = tensordict.get(
-                    ("next", _replace_last(done_key, "truncated")),
-                    default=torch.zeros((), device=done.device, dtype=torch.bool),
-                )
-                done = done | truncated
-                any_sub_done = done.any().item()
-                if any_sub_done and not break_when_any_done:
-                    # Add this done to the map, we will need it to reset
-                    _reset_map.update({_replace_last(done_key, "_reset"): done})
-                any_done += any_sub_done
```
matteobettini, line 1601: this was mostly for efficiency reasons. the reason behind it is that we write a _reset only if there are any dones.

torchrl/envs/common.py
```diff
@@ -1613,11 +1577,8 @@ def policy(td):
             )
 
             if not break_when_any_done and any_done:
-                for _reset_key, done in _reset_map.items():
-                    tensordict.set(_reset_key, done.clone())
                 self.reset(tensordict)
-                for _reset_key in _reset_map.keys():
-                    del tensordict[_reset_key]
```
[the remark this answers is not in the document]
matteobettini: we can take care of the removal in _reset yes.

torchrl/envs/common.py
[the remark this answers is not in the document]
matteobettini: if our assumption holds yes

torchrl/envs/libs/gym.py
```diff
@@ -965,3 +1107,12 @@ def __call__(self, info_dict, tensordict):
...
         return tensordict
+
+def _flip_info_tuple(info: Tuple[Dict]) -> Dict[str, tuple]:
+    # In Gym < 0.24, batched envs returned tuples of dict, and not dict of tuples.
+    # We patch this by flipping the tuple -> dict order.
+    info_example = info[0]
+    result = {}
+    for key in info_example:
+        result[key] = tuple(_info[key] for _info in info)
+    return result
```
vmoens, line 1118: @skandermoalla another reason to praise tensordict :p

test/test_env.py
```diff
@@ -955,6 +947,7 @@ def test_parallel_env_reset_flag(self, batch_size, n_workers, max_steps=3):
         while not _reset.any():
             _reset = env.done_spec.rand()
 
+        env._allow_done_after_reset = True
```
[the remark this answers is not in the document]
matteobettini: no, it can as long as it follows the _reset instructions

if we set `env._allow_done_after_reset = True` this test will have no use

test/test_env.py
```diff
@@ -1047,15 +1009,16 @@ def test_env_base_reset_flag(batch_size, max_steps=3):
     assert (td["next", "done"] == 1).all()
     assert (td["next", "observation"] == max_steps + 1).all()
 
-    _reset = env.done_spec.rand()
+    env._allow_done_after_reset = True
```
matteobettini, line 1012: again this flag should not be used in the tests, this is for envs that will be done on the _reset dimensions.

test/test_transforms.py
```diff
@@ -1406,18 +1411,14 @@ def test_trans_serial_env_check(self):
...
-
     @pytest.mark.skipif(not _has_gym, reason="Gym not found")
     def test_transform_env(self):
         env = TransformedEnv(GymEnv(PENDULUM_VERSIONED), StepCounter(10))
         td = env.rollout(100, break_when_any_done=False)
         assert td["step_count"].max() == 9
         assert td.shape[-1] == 100
 
-    @pytest.mark.parametrize("step_key", ["step_count", ("other", "key")])
+    @pytest.mark.parametrize("step_key", ["step_count", "other-key"])
```
matteobettini, line 1421: why are we not testing the nested key anymore?

torchrl/envs/common.py
```diff
@@ -986,13 +1069,6 @@ def done_spec(self, value: TensorSpec) -> None:
                         "An empty CompositeSpec was passed for the done spec. "
                         "This is currently not permitted."
                     )
-                _check_only_one_entry(
```
matteobettini, line 989: should we check that the shapes at each td level match?

torchrl/envs/batched_envs.py
matteobettini: also the convention was that missing resets are considered all true, not an aggregate of dones

test/test_transforms.py
vmoens: @matteobettini we now restrict step_count to follow the done. Since it must match in shape etc there's not reason to allow it to be at a nested level. The idea is that we want to be able to match 
reset <=> done <=> truncated <=> step_count <=> is_init
So we need them all to be always at the same level.
Otherwise you could have a step_count at any level and it's hard to find where it comes from.

test/test_env.py
vmoens: Removed

test/test_transforms.py
matteobettini: got it thanks

test/test_collector.py
```diff
@@ -1740,6 +1758,161 @@ def test_param_sync(self, give_weights, collector, policy_device, env_device):
             col.shutdown()
 
 
+class TestBringReset:
```
vmoens, line 1761: @matteobettini here is the place where we test the functionality of brining resets at the root.
In there, there's a check that if we have a partially populated list of resets we raise an exception

vmoens:
@skandermoalla 
To come back at your original point, I actually agree: naming done what is named terminated in gym isn't a great idea. It will only bring confusion, even for us.
I think the only way forward is this:
For now, any env that return termination, truncation gets a done spec made of `done, terminated, truncated`, ie, we add the done state. This has a low memory footprint and ensures that most things hold.
We adapt the collectors (as in this PR), the transforms, the value functions and the losses to read the proper end-of-traj signal: either truncated and terminated are provided, or only done. If only done, we assume that truncated = terminated = done.
@matteobettini wdyt? Feasible?

vmoens:
Good so replying to myself.

After a thorough discussion with @matteobettini we concluded that the best thing (least BC breaking) would be:
- Keep done in the lib as the default (not to break everything)
- If an env just returns a done, the most natural thing to us is to assume that this means terminated, ie we don't assume that this means truncated. This avoids ditching data in value functions for instance (a problem @BY571 just encoutered).
- If the env returns `termination, truncation` we interpret that as `done, truncated` in tochrl nomenclature. This isn't great from a UX perspective as done means 2 different things in different libs (gym doc says that `done = truncation | termination` but we say `done = termination`). Unfortunately it's that or do hugely bc-breaking changes...
- We add a `"stop"` entry along with `truncated` and `done` that means "end of trajectory (for whatever reason) => you should call reset"

We think that this is the least bc-breaking and the clearest we can do. Obviously this would be documented in bold characters in the doc.

A big thanks to @matteobettini for your patience during this discussion! I can be quite stubborn and opinionated :p 
Do you have anything to add?

@btx0424 for context

matteobettini:
I actually think that done should be named `panic`.

Joking, looks good to me!

skandermoalla:
> Keep done in the lib as the default (not to break everything)

Okay. And this done has the exact use of `terminated` in Gymnasium? I.e. it is when the MDP reaches a terminal state and should bypass bootstrapping.

> If an env just returns a done, the most natural thing to us is to assume that this means terminated, ie we don't assume that this means truncated.

Totally agree.

> This avoids ditching data in value functions for instance (a problem @BY571 just encountered).

This isn't relevant anymore, but can you elaborate on how truncated would ditch data in the value functions? My understanding is the opposite, truncated keeps the bootstrap, and terminated bypasses it. Truncation could instead lead to ditching episode returns. 

> If the env returns termination, truncation we interpret that as done, truncated in tochrl nomenclature.

Good.

> This isn't great from a UX perspective as done means 2 different things in different libs.

Yes, but it's functional and correct. As long as it's documented it's fine. Logic should be settled first. Some major refactoring could be done at some point if needed to change names but without changing any logic.

> We add a "stop" entry along with truncated and done that means "end of trajectory (for whatever reason) => you should call reset"

This is a very good thing. I would allow components that only need that bit to free themselves from the logic of termination and truncation. Making development on them also easier.

Thanks a lot for this guys!

torchrl/envs/libs/dm_control.py
```diff
@@ -240,9 +248,8 @@ def _output_transform(
             timestep_tuple = (timestep_tuple,)
         reward = timestep_tuple[0].reward
 
-        done = False  # dm_control envs are non-terminating
+        truncated = terminated = done = False  # dm_control envs are non-terminating
```
skandermoalla, line 251:
```suggestion
        truncated = terminated = gym_done = False  # dm_control envs are non-terminating
```

I spotted this here, but I'm probably missing it in some other places. Feel free to ignore it if it's not essential to change it everywhere.

torchrl/envs/libs/dm_control.py
```diff
@@ -240,9 +248,8 @@ def _output_transform(
...
         observation = timestep_tuple[0].observation
-        truncated = terminated = None
         info = {}
 
         return observation, reward, done, truncated, terminated, info
```
skandermoalla, line 255:
```suggestion
        return observation, reward, gym_done, truncated, terminated, info
```

torchrl/envs/libs/dm_control.py
skandermoalla: I think it would be a good thing to always call the old gym done `gym_done`

docs/source/reference/envs.rst
```diff
@@ -79,20 +79,33 @@ The following figure summarizes how a rollout is executed in torchrl.
...
 
   The Gym(nasium) API recently shifted to a splitting of the ``"done"`` state
-  into a ``terminated`` (the env is done and results should not be trusted)
-  and ``truncated`` (the maximum number of steps is reached) flags.
-  In TorchRL, ``"done"`` usually refers to ``"terminated"``. Truncation is
-  achieved via the :class:`~.StepCounter` transform class, and the output
-  key will be ``"truncated"`` if not chosen to be something else (e.g.
-  ``StepCounter(max_steps=100, truncated_key="done")``).
+  into a ``termination`` (the env is done and results should not be trusted)
+  and ``truncation`` (the maximum number of steps is reached) flags.
```
skandermoalla, line 91:
```suggestion
  and ``truncation`` (an external limit on the number of steps is reached) flags.
```

Not to confuse it with finite-horizon limits, which set the `termination` bit.

docs/source/reference/envs.rst
```diff
@@ -79,20 +79,33 @@ The following figure summarizes how a rollout is executed in torchrl.
...
+  the union of ``"done"`` (task completion) and ``"truncated"`` (task interruption).
+  If the environment carries a single value, it will interpreted as a completion
+  (``"done"``) signal by default (and not ``"stop"`` or ``"truncated"``). The
+  reason for this choice is that without any more information, one has to assume
+  task completion for follow-up tasks (eg, computing value estimations), so any
+  unspecified end-of-trajectory signal will eventually need to be interpreted
+  with a worst-case-scenario strategy. Another motivation is that a ``"stop"``
+  signal will be modified by a ``"truncation"`` implemented, for instance, via
+  a :class:`~.StepCounter` transform, and the original ``"stop"`` signal
+  will be lost.
```
skandermoalla, line 104: Not sure to understand this. Could you clarify?

vmoens:
> Okay. And this done has the exact use of `terminated` in Gymnasium? I.e. it is when the MDP reaches a terminal state and should bypass bootstrapping.
> 

Not now but I started a branch implementing these changes!
Value estimators will accept all the done/truncated/stop triad and make the best educated guess they can in case only a partial info is provided. Wish me luck because it's gonna be a nightmare with the vectorised versions!
> 
> 
> 
> This isn't relevant anymore, but can you elaborate on how truncated would ditch data in the value functions? My understanding is the opposite, truncated keeps the bootstrap, and terminated bypasses it. Truncation could instead lead to ditching episode returns. 
> 
Yes you're right. I realise I missed a step when writing my thought process. Anyhow: the point is that assuming unspecified eot is truncated is the least reasable option, and for value estimation we need to decide if it's terminated U truncated U terminated and truncated.
> 

> 
> 
> 
> Yes, but it's functional and correct. As long as it's documented it's fine. 

Glad you agree
> 
> 
> 
> This is a very good thing. I would allow components that only need that bit to free themselves from the logic of termination and truncation. Making development on them also easier.
> 
> 
Good! Do you think we should always have a stop when we have a truncated and a done (eg, in StepCounter)? Right now I disabled it by default there but enabled it in gymnasium => the lack of consistency isn't great...

torchrl/envs/libs/dm_control.py
vmoens: Or we could just stick to torchrl's names? Done is completed, stop is eot, truncated is interrupted, always. No terminated for instance.
"Gym_done" could also be twisted (if I'm a picky devil's advocate) into "done but the gym version which is terminated"

skandermoalla:
> Anyhow: the point is that assuming unspecified eot is truncated is the least reasable option

I agree and I don't think we should do anything more than assuming that an underdetermined done means termination.

> Value estimators will accept all the done/truncated/stop triad and make the best educated guess they can in case only a partial info is provided.

So here I'm not sure we want to do that educated guess.
I think the API unification should be solved at the environment level so that vectorized envs all carry the same semantics.
Then value estimators (as far as how we're currently doing RL) should only care about `done`. (`termination` in gymnasium).
They could optionally read the others if doing things that I don't see yet.

> Do you think we should always have a stop when we have a truncated and a done (eg, in StepCounter)? Right now I disabled it by default there but enabled it in gymnasium => the lack of consistency isn't great...

When is the `stop` present at the moment?
I would say it's fine (maybe better) to always have it. It would be a redundant bit in some cases, but if it allows for more consistency and for some components to just read that signal (instead of reading both done and truncated) then it'd be good.

torchrl/envs/libs/dm_control.py
skandermoalla: Ah would be even better.
So right after a a gym call have
```python
# new gym.
obs, reward, done, truncated, info = env.step(action) 
# old gym
obs, reward, stop, info = env.step(action)
```
and not have any `terminated` variable in the code?

matteobettini:
> So here I'm not sure we want to do that educated guess.
> I think the API unification should be solved at the environment level so that vectorized envs all carry the same semantics.
> Then value estimators (as far as how we're currently doing RL) should only care about `done`. (`termination` in gymnasium).
> They could optionally read the others if doing things that I don't see yet.

I think in estimators that use a trajectory like GAE and TDLambda, we do need to know also about truncations (if any) in order to avoid contamination when there are more consecutive not-done trajectories. 

this is currently an issue with MultiSyncCollectors where the environment has no batch size. this class will cat the frames of all workers on the time dimension without any mark (done or truncated) and thus this data will be leading to wrong outputs from the trajectory based estimators 

> I would say it's fine (maybe better) to always have it. It would be a redundant bit in some cases, but if it allows for more consistency and for some components to just read that signal (instead of reading both done and truncated) then it'd be good.

I think this choice requires us to be careful. I would prefer  to avoid  writing an additional stop key at every step. Imagine i am a non-gym user and i only use only the torchrl done (like vmas, smac, ...) then i would copy paste the only done key i have all the time into the stop key. 

Where is a stop key actually used? The only use i can think is for resetting, but that should only happen max once per step and it could be an or among all the done keys in the same td. Are there other usages for the stop key? 

- value functions should be happy with done and truncated
- environments should be happy with just done or the done-truncated pair

matteobettini:
> Value estimators will accept all the done/truncated/stop triad and make the best educated guess they can in case only a partial info is provided. Wish me luck because it's gonna be a nightmare with the vectorised versions!

To expand a bit on this it seems to me that the value estimators need 2 inputs regarding these flags and not 3:

1. one input is to indicate that trajectory ends and no bootstrapping
2. one input is to indicate that trajectorty ends but yes bootstrapping

now let's see how we use them:

- if i have only done this is input 1
- if i have done and truncated they are input 1 and 2
- if ihave done and truncated and stop then like above and stop is ignored
- if i have only stop (should not happen but who knows) this is input 1 cause that is the best i can do
- and so on..

so the user can just pass to input 1 the key that she prefers to indicate no bootstrapping and to input 2 the key she prefers to say traj is over but bootstrap.

and we do not have to juggle internally with these cases i enumerated

skandermoalla:
> To expand a bit on this it seems to me that the value estimators need 2 inputs regarding these flags and not 3:
1.one input is to indicate that trajectory ends and no bootstrapping
2.one input is to indicate that trajectorty ends but yes bootstrapping

Yes, my bad. I forgot about _consecutive trajectories from the same environment_ in the same sequence. The estimator would indeed need two bits (or to be exact, 1.5 bits). This is exactly the information you wrote above.

To me, the most intuitive way to encode this is with a stop key (indicating the end of a trajectory, to stop summing) and a done key (indicating to bypass or not bootstrapping).
This can always be present no matter whether the environment returns a truncation signal or not.

```python
# only done
done, stop = done, done
# terminated, truncated
done, stop = terminated, terminated | truncated
# only truncated?
done, stop = False, truncated
```

However, this is not how things have historically been done, and as you said, the information can also be expressed with a done and a truncated (at the expense of the value estimator computing its own stop).
Since, done and truncated are likely to stay, if not writing a stop saves some significant space and compute, then yes, I agree that the trouble of adding a stop is not worth it.

Now for _(non-consecutive) trajectories of different environments_ in the same sequence. We fall into the problem you mention here:

> this is currently an issue with MultiSyncCollectors where the environment has no batch size. this class will cat the frames of all workers on the time dimension without any mark (done or truncated) and thus this data will be leading to wrong outputs from the trajectory based estimators

This is quite bad indeed (I haven't used this collector before). We don't want to write an artificial truncated to split the trajectories as this will overload its semantics and potentially cause unexpected issues. We can't write a stop either for the same reason.
I don't see an obvious solution tbh, other than splitting the trajectories in the collector or adding another key to be read by the value estimator to effectively act like truncated but not be the truncated key.

vmoens:
> This is quite bad indeed (I haven't used this collector before). We don't want to write an artificial truncated to split the trajectories as this will overload its semantics and potentially cause unexpected issues. We can't write a stop either for the same reason.

I thought about this. What I would suggest as a refactoring of the multi-sync collector is to have it collect trajectories from the various sub-collectors and always stack along the time dimension. The reason is that we have the capability of imposing a `preemptive_threshold` that will stop some collectors from collecting once a certain amount of data is collected.
At this point: either we return only the valid data (which we can't stack because the output of the collectors will have different shapes) or we add a mask key and we can return padded data which, in this case can be stacked.
If we concatenate along the time dimension, we definitely need a truncated signal (maybe named otherwise) for the obvious reason we talked about previously.
I agree that we don't want for this truncated to interfere with the one coming from the env: overriding it would be loosing/corrupting info which is never appropriate.
We already have a `"collector"` entry in the data so we could add a `("collector", "truncated")` entry there too.
After that, we can adapt loss modules / value modules to check for that key and if present do a `truncated = data["next", "truncated"] | data["collector", "truncated"]`.

docs/source/reference/envs.rst
vmoens: I added a little something

torchrl/envs/libs/dm_control.py
vmoens: that's the idea!

skandermoalla:
I like the idea of a ("collector", "truncated") key. It seems to be the least disruptive.

But I also think that the masking feature would be nice to have as it would support other cases, e.g. episodic losses like REINFORCE: One can use the 
single SyncCollector with parallel environments to collect trajectories containing multiple episodes of different lengths, and stacked along the env batch dim, then the loss can get a mask to discard the trailing observations from unfinished episodes.
This is not a feature for long environments and modern estimators/losses, but for controlled experiments with old school REINFORCE loss.
I'm doing something like this myself for my current project.

vmoens:
@matteobettini @skandermoalla 
This PR is almost ready. 
One doubt I have is whether or not we should decorate the old gym envs with done, truncated and stop.
Pros: interchangeablility: we need to be able to write code that is somewhat oblivious to the gym version. For instance, if we do `final_reward = episode["next", "reward"][episode["next", "done"]]` this can result in an empty tensor after this PR for gymnasium ("done" must be swapped with "stop")
Cons: an env could be truncated, but not signaled as such, and our "truncated" won't track that. That could be misleading. We can always get out of this by saying it isn't our fault but it isn't idea.

An intermediate thing could be to always have "stop", but avoid "truncated" in order gym versions.

skandermoalla:
What's the current state of things regarding each key (done, truncated, stop)? Maybe a recap can be beneficial.

vmoens:
> What's the current state of things regarding each key (done, truncated, stop)? Maybe a recap can be beneficial.

I was thinking that following this PR:
- "done" would always mean completed
- "truncated" would always mean early interruption
- "stop" would be the intersection

If someone defines an env with a `"done"` key only, we automatically add a `"stop"` key in EnvBase `step` and `reset` methods to have this info accessible.
I'm just hesitant to add a `"truncated"` too

vmoens:
@skandermoalla 
Sorry to bother you again with this

I'm struggling a bit with `"stop"`
In the code, we seem to be assuming that `done` is `terminated | truncated` more often than `terminated` (for instance to look at the final reward or obs of a trajectory etc).
So My current thinking is:
- have `"completed"`, `"truncated"` and `"done"`. 
- Now `"done"` is the stop signal (yeah I know I change my mind often). The advantage is that it is minimally disruptive. Is is also aligned with gym nomenclature.
- If you have termination (completed) and truncation (terminated) signal we also register them, along with done.
- Otherwise, you just have done.
- When completed and truncated are needed (eg in losses / value) but not available, we set `completed = done; truncated=False` which is the best assumption we can have.

matteobettini:
If this is the road we are taking for me it should be "terminated", "truncated" and "done" to align with gym.
Introducing a new keyword seems a lot.

also will all the envs that just have a done in the codebase have that renamed to "completed" (or "terminated")?

vmoens:
> If this is the road we are taking for me it should be "terminated", "truncated" and "done" to align with gym.

Gym uses "done", "termination" and "truncation"
But you can say "an env is done" but not "an env is termination" so I don't think we should use these terms.
Open to terminated but while we're at it, why not use "completed"? For instance: if my job is terminated, it could mean that I'm fired (but haven't completed my work). If my work is completed, I have finished the task.

But I will fold to whatever the people say obviously :) 

> also will all the envs that just have a done in the codebase have that renamed to "completed" (or "terminated")?

No by default it will be done because we can't tell what you mean.
In older versions of gym it could mean both. Even in MARL codebase we use done as the intersection of both. So this is the minimally disruptive decision.

> if we do not do this and consider those dones as done then they will be modified by an external truncation (e.g., a transform)

Yes I added this to the doc.
(1) this is already the case so I would not worry too much about that.
(2) You can always rename your done via a `RenameTransform` if you want to keep the original value.
(3) I think it is what we want when we do [this](https://github.com/pytorch/rl/blob/434fe58b44cb46650e80ff18030d43a45eb532f2/examples/multiagent/utils/logging.py#L70-L75)

matteobettini:
> Gym uses "done", "termination" and "truncation"
> But you can say "an env is done" but not "an env is termination" so I don't think we should use these terms.
> Open to terminated but while we're at it, why not use "completed"? For instance: if my job is terminated, it could mean that I'm fired (but haven't completed my work). If my work is completed, I have finished the task.

you can say an env is terminated and an env is truncated. 

i can male the counter argument to yours that for example a video game can be game over when the task is not completed. to me when the env is terminated it just means that the mdp reached a terminal state and does not correlate with task completion. since the state is called "terminal", the think that makes most sense to me is terminated, which also aligns with gym

> No by default it will be done because we can't tell what you mean.

mmh this is not a conservative choice to me.
if we cannot tell what you mean it is safer to not overwrite it with an external truncated (instead of overwriting it by default) and assume it as the terminal state of the mdp.

vmoens:
> if we cannot tell what you mean it is safer to not overwrite it with an external truncated (instead of overwriting it by default) and assume it as the terminal state of the mdp.

I think the assumption here is that we cannot assume that we know so we should not override. But the library is very clear (and actually already behaves in this way) that we say that done is anything that ends a trajectory. To me it's even more surprising that a trajectory is truncated but not done. That means that the code snippet I linked above will fail whenever you truncate your trajectory. 
This is (1) minimally disruptive (not bc-breaking), (2) coherent with the rest of the library and (3) reversible and documented so I don't see why not adopting that. What other option do we have?

skandermoalla:
It seems like the thread would benefit more from a live discussion at this moment. Would you guys want to meet @vmoens @matteobettini?
We can summarise the result in a comment here afterward.

skandermoalla:
**Nomenclature**

> Gym uses "done", "termination" and "truncation"

terminated or termination, truncation or truncated, are not relevant. These are used interchangeably as adjectives or nouns. The variables in the code are usually the adjectives `terminated` and `truncated` (see code example on the gymnasium lading page https://gymnasium.farama.org/).

I think `terminated` would be better than `completed`. As @matteobettini argues, it refers to the MDP reaching a terminal state and is what people would probably appreciate more as it's the same as gymnasium.

In what follows I'll use the `terminated`, `truncated`, `done := terminated | truncated` nomenclature.

**What's necessary to have**

Let's make a recap of what components need what:
- Losses and value estimators need a done (to split trajectories) and a terminated (to know if to bypass bootstrapping or not). In particular, they don't necessarily need a truncated. (Value estimators will have to be fixed btw as `done` is not enough).
Done can always be provided (as it's a superset). Terminated is subject to assumptions. 
- Collectors for resetting etc. only need done.
- Computing _episodic_ stats needs terminated (done is missing information). Computing _trajectory_ stats can be settled with done. In particular, [this](https://github.com/pytorch/rl/blob/434fe58b44cb46650e80ff18030d43a45eb532f2/examples/multiagent/utils/logging.py#L70-L75) should be using `terminated` instead, and is already making the assumption that `done` in those envs means `terminated`.
- Maybe there are other components I'm missing.

What you can see from this is that ` truncated` doesn't seem to be necessary anywhere, as long as it's taken into account in `done`.

**Assumptions**

I think there are two conflicting facts that we can't rally with a unifying assumption. We need to treat different cases of single-done signals differently. 
1) In old gym, this meant `terminated | truncated`.
2) In other envs this can mean `terminated` as @matteobettini is saying. I forgot about those in my previous arguments.

I think we can make:
```Python
# As soon as you get the output from the native env do:

# terminated, truncated in gymnasium
terminated, truncated, done = terminated, truncated, terminated | truncated

# done only, in gym
terminated, truncated, done = done and 'TimeLimit.truncated' not in info, info.get("TimeLimit.truncated", False), done # https://gymnasium.farama.org/tutorials/gymnasium_basics/handling_time_limits/#solution

# done only, in other envs if not specified
terminated, done = done, done

# Otherwise handle the thing differently per library.
```
 
The two first snippets don't need justification. The last one is where the assumption is made.

matteobettini:
@skandermoalla 

me and @vmoens were thinking about a thing.

If the episode reward statistic should use terminated, but an environment is reset with just a truncation (and not a termination) the statistics will be invalid as it would cross over 2 episodes.

I think as long as we reset our environments using `done` we have to use that to compute episode statistics.

skandermoalla:
> If the episode reward statistic should use terminated, but an environment is reset with just a truncation (and not a termination) the statistics will be invalid as it would cross over 2 episodes.

Can you give more context on how those statistics are computed?

The way I think of it is that episode statistics are initialized whenever a done signal is received, keep being recorded, and are logged only when a terminated is received (i.e. at the end of the episode).
As opposed to trajectory statistics which are logged whenever a done is received (i.e. at the end of the trajectory).

In this sense, episode statistics are reset at every done (so they don't span 2 episodes) but only logged if a termination is reached.
trajectory statistics will always be logged.

vmoens:
@skandermoalla I added [this](https://github.com/pytorch/rl/blob/03001bcf53c86e0b8095ffbd9830607e93812b15/torchrl/envs/libs/gym.py#L1149) to read the truncated state in older versions of gym (it's a mess, the flag is only there when truncated id True, sight!)
Do you think it's robust enough?

skandermoalla:
LGTM. I'm just not aware of what the `_info_spec` of a `BaseDictInfoReader` is, so skipped [this line](https://github.com/pytorch/rl/blob/03001bcf53c86e0b8095ffbd9830607e93812b15/torchrl/envs/libs/gym.py#L1165).

With this, will the truncated key always be written in the output tensordict of `gym`/`gymnasium` environments?

skandermoalla:
After reading this `truncated` key, it should be combined with the `done` key of the older gym version to get the `terminated` key, where `terminated := done and not truncated`.

skandermoalla:
Btw, I don't know where this fits in the conversation at the moment, but time limits made to model the environment as a finite-horizon MDP should be written to the `terminated` key and not the `truncated` key.
This is not necessary for this PR, I'm just mentioning it in case we make design choices that make it harder to include later, but it would be a nice feature of the StepCounter wrapper. One would then need a signal representing the remaining time in the observation to maintain the Markov property.

vmoens:
> terminated := done and not truncated

Not sure about that one: what if the env is truly terminated but incidentally also truncated? Seems a bit dangerous to me

skandermoalla:
> Not sure about that one: what if the env is truly terminated but incidentally also truncated? Seems a bit dangerous to me

Ah, that's a good point. Somehow that's what Gymnasium [ says to do ](https://gymnasium.farama.org/tutorials/gymnasium_basics/handling_time_limits/#solution)
```
terminated = done and 'TimeLimit.truncated' not in info
```

> what if the env is truly terminated but incidentally also truncated? 

With the TorchRL step counter this can happen indeed, but can this happen with the TimeLimitWrapper of the old gym?

This is the code is v24.1, the last version prior to the API change if I'm correct: 
https://github.com/openai/gym/blob/66c431d4b3072a1db44d564dab812b9d23c06e14/gym/wrappers/time_limit.py#L13C4-L14C96

> The done-signal originates from the time limit (i.e. it signifies a *truncation*) if and only if
    the key `"TimeLimit.truncated"` exists in ``info`` and the corresponding value is ``True``.

And in the code they do 
https://github.com/openai/gym/blob/66c431d4b3072a1db44d564dab812b9d23c06e14/gym/wrappers/time_limit.py#L52
```
if self._elapsed_steps >= self._max_episode_steps:
            info["TimeLimit.truncated"] = not done
            done = True
```

So it seems like a correct implementation of a reader reading the `done` and `info` directly from the env would be:
```
terminated = done and not info.get('TimeLimit.truncated', False)
```

What do you think?

vmoens:
Now surprise, 
```python
info.get('TimeLimit.truncated', False)
```
can be a bool but also a list (with vec envs)
so it should be
```python
        truncated = info.get('TimeLimit.truncated', False)
        if not isinstance(truncated, bool):
            truncated = np.array(truncated)
        terminated = done and not truncated
```
but it's a detail...

I think my latest commit should do the trick

skandermoalla:
Haha yes yes, this
```python
terminated = done and not info.get('TimeLimit.truncated', False)
```
was only meant as pseudo-code. The details and getting around silent `__bool__(self)` will be harder.

vmoens:
I think we're reaching a good state for this PR.
I will do some cleanup tomorrow and maybe add a couple of tests. I will also test that on a separate branch with the SOTA implementations we have to check that everything's running smoothly!

vmoens:
Final list of changes part of this PR:
- Add full_done_spec, full_observation_spec, full_action_spec, full_state_spec, full_reward_spec properties.
- Make all envs have `"done"` and `"terminated"`.
- Add a `_complete_done` private method within envs to complete output data if done_keys are missing. Make this compatible with transforms to avoid duplicated calls to the method.
- Full conversion to `done_keys` wherever `done_key` was used.
- versioned reading helpers in GymWrapper to account for discrepancies between versions
- `terminated_or_truncated`, a function that reads the data, writes the `"_reset"` keys and return a boolean indicating if anything is done.
- Fix `@implement_for`, `set_gym_backend` and related + add tests
- Fix envs tutorial and add info wrt new "done" attributes
- Compatibility with batched environments
- Fix transforms: VecGymEnv, Reward2GoTransform, RenameTransform, InitTracker, TimeMaxPool, SelectTransform, ExcludeTransform, StepCounter
- Fix ModelBased envs, RoboHive, D4RL, RLHF
- Adapt collectors to new logic
- Better handling of workers refusing to close in collectors

matteobettini:
I ll try to make a full review before tonight if thats ok

torchrl/collectors/collectors.py
```diff
@@ -790,66 +791,38 @@ def iterator(self) -> Iterator[TensorDictBase]:
...
         if any_done:
             traj_ids = self._tensordict.get(("collector", "traj_ids"))
             traj_ids = traj_ids.clone()
             # collectors do not support passing other tensors than `"_reset"`
             # to `reset()`.
-            td_reset = self._tensordict.select(*done_map.keys())
-            for done_key, done in done_map.items():
-                td_reset.set(_replace_last(done_key, "_reset"), done)
-                del td_reset[done_key]
+            traj_sop = _bring_reset_to_root(td_reset)
```
matteobettini, line 814: we need to pass the reset keys here because it is very expensive to scan the whole td

torchrl/envs/batched_envs.py
```diff
@@ -635,14 +638,14 @@ def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
...
-
         if self._single_task:
             # select + clone creates 2 tds, but we can create one only
             out = TensorDict(
                 {}, batch_size=self.shared_tensordict_parent.shape, device=self.device
             )
             for key in self._selected_reset_keys:
-                if _unravel_key_to_tuple(key)[-1] != "_reset":
+                key = unravel_keys(key)
+                if key not in self.reset_keys:
```
matteobettini, line 648: the reset key should not be removed here

torchrl/envs/batched_envs.py
```diff
@@ -888,7 +889,8 @@ def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
...
```
matteobettini, line 893: same as above

vmoens:
> * pettingzoo has truncated and terminated in each agent group so we can adapt that

I don't get this comment, what do you mean? Is there some extra work needed? I think we cover multiple truncated and terminated within the data structure

torchrl/envs/batched_envs.py
```diff
@@ -591,20 +596,18 @@ def set_seed(
...
-                if _reset is not None:
-                    for i in range(self.num_workers):
-                        needs_resetting[i] += _reset[i].any()
-                else:
-                    missing_reset = True
-                    break
-
-        if tensordict is None or missing_reset:
-            needs_resetting = [True] * self.num_workers
+            needs_resetting = _bring_reset_to_root(tensordict)
```
matteobettini, line 600: we need to pass reset keys for efficiency

torchrl/envs/batched_envs.py
```diff
@@ -832,21 +835,19 @@ def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
...
-            needs_resetting = [False for _ in range(self.num_workers)]
-            for _reset_key in self.reset_keys:
-                _reset = tensordict.get(_reset_key, default=None)
-                if _reset is not None:
-                    for i in range(self.num_workers):
-                        needs_resetting[i] += _reset[i].any()
-                else:
-                    missing_reset = True
-                    break
+            needs_resetting = _bring_reset_to_root(tensordict)
```
matteobettini, line 839: same as above

torchrl/envs/utils.py
```diff
@@ -728,3 +727,175 @@ def check_marl_grouping(group_map: Dict[str, List[str]], agent_names: List[str])
...
+            if key == "_reset":
+                local_reset = td.get(key)
+                if local_reset.ndim > n:
+                    local_reset = local_reset.flatten(n, local_reset.ndim - 1)
+                    local_reset = local_reset.any(-1)
+                reset = reset | local_reset
+            # we need to check the entry class without getting the value,
+            # because some lazy tensordicts may prevent calls to items().
+            # This introduces some slight overhead as when we encounter a
+            # tensordict item, we'll need to get it twice.
```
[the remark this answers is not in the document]
matteobettini: maybe we could remove the skim_trough branch?

vmoens:
> * vmas, smac and similar simulators should output terminated instead of done so that we keep consistency

As discussed, they now output both without you (the user) needing to do anything:

docs/source/reference/envs.rst
```diff
@@ -79,22 +79,25 @@ The following figure summarizes how a rollout is executed in torchrl.
...
-  In TorchRL, ``"done"`` usually refers to ``"terminated"``. Truncation is
-  achieved via the :class:`~.StepCounter` transform class, and the output
-  key will be ``"truncated"`` if not chosen to be something else (e.g.
-  ``StepCounter(max_steps=100, truncated_key="done")``).
-  TorchRL's collectors and rollout methods will be looking for one of these
-  keys when assessing if the env should be reset.
+  In general, all TorchRL environment have a ``"done"`` and ``"terminated"``
+  entry in their output tensordict. If they are not present by design,
+  the :class:`~.EnvBase` metaclass will ensure that every done or truncated
+  is flanked with its dual.
```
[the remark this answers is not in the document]
vmoens: meant terminated, let me amend

test/mocking_classes.py
```diff
@@ -1700,6 +1748,7 @@ def _step(
         self.count += one_hot_action.to(torch.int)
         td["observation"] += expand_right(self.count, td["observation"].shape)
         done["done"] = self.count > self.max_steps
+        done["terminated"] = self.count > self.max_steps
```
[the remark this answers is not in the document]
vmoens: because we want done to be able to quickly know if we should reset. It avoids always going through all the possible keys and carries the summary of the end-of-traj in one tensor. 
So now all envs have at least one done and one terminated, possibly a truncated.
It's what we discussed with @skandermoalla IIRC.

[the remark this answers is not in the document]
matteobettini: but the done would be added automatically by EnvBase no?

torchrl/envs/utils.py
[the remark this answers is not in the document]
vmoens: Not necessarily, I'd like this to be usable without access to an env (I don't want to assume that we always have access to an env that is properly constructed).
Both use cases are properly tested so I think we're good with both options for now. 
Is there a strong argument to remove it?

torchrl/envs/batched_envs.py
vmoens: why? Can you explain? It used to be removed (see diff) and i'm pretty sure tests fail if you don't.

matteobettini:
> As discussed, they now output both without you (the user) needing to do anything:

yep got that but if we keep writing done explicitly instead of terminated in the wrappers then the users will be deceived on the meaning of what comes from the simulator.

I think terminated should be the one provided and done created through a clone

torchrl/envs/batched_envs.py
```diff
@@ -832,21 +835,19 @@ def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+            if needs_resetting.ndim > 2:
+                needs_resetting = needs_resetting.flatten(1, needs_resetting.ndim - 1)
+            if needs_resetting.ndim > 1:
+                needs_resetting = needs_resetting.any(-1)
+            elif not needs_resetting.ndim:
+                needs_resetting = needs_resetting.expand(self.batch_size)
+        else:
+            needs_resetting = torch.ones(
+                (self.num_workers,), device=self.device, dtype=torch.bool
+            )
```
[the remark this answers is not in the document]
vmoens: ?

matteobettini:
> I don't get this comment, what do you mean? Is there some extra work needed? I think we cover multiple truncated and terminated within the data structure

right now in pettingzoo there is an or between the 2 to create a done, we should just adapt that

torchrl/envs/utils.py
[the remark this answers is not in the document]
matteobettini: ok but at least let s make sure that within the codebase we call it with the keys for efficiency

torchrl/envs/batched_envs.py
matteobettini: there are many reasons we discussed

- they should alwayes be removed by `env.reset()`
- this removes them only when `self._single_task` and not otherwise

torchrl/envs/common.py
```diff
@@ -931,34 +1022,68 @@ def done_spec(self) -> TensorSpec:
...
+                shape=(*full_done_spec.shape, 1),
+                dtype=torch.bool,
+                device=self.device,
+            )
+            self.output_spec.unlock_()
+            self.output_spec["full_done_spec"] = full_done_spec
+            self.output_spec.lock_()
+            return
+
+        def check_local_done(spec):
```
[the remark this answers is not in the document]
vmoens: We can do that

torchrl/envs/batched_envs.py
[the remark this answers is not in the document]
matteobettini: why do we switch on the ndim, couldn't we simply squash it into 1 dim?
also expanding to self.batch_size seems wrong to me as that could be bigger than self.n_workers.

vmoens:
> > As discussed, they now output both without you (the user) needing to do anything:
> 
> yep got that but if we keep writing done explicitly instead of terminated in the wrappers then the users will be deceived on the meaning of what comes from the simulator.

ah got it, so your point is that as part of this PR all simulators that were writing done should write terminated instead?

> I think terminated should be the one provided and done created through a clone
Per se both are always provided
I think it's ok but it's not user-facing, mostly stuff that lives in private methods. We can check that the doc is accurate.

test/mocking_classes.py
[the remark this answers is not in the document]
skandermoalla: Not sure if my input is needed here, but here's how I understand this:
>  So now all envs have at least one done and one terminated, possibly a truncated.

Yes, this is what is expected.

> but the done would be added automatically by EnvBase no?
i am talking about this mocking class, it could just output terminated and done would be built

Yes, the done key can be inferred from the terminated (possibly truncated keys). I haven't looked at the implementation yet to know where this is done, but if the EnvBase takes care of this, then this mock can avoid writing the done key indeed to test the automatic done inference feature.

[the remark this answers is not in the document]
vmoens: I think it's best if we show both whenever we can. We can copy the done if you prefer but not showing that it's populated isn't great from a UX perspective IMO
The auto patch is supposedly just there to ensure that everything runs fine even if someone just writes a terminated/done key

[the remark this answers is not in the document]
vmoens: just did, what did I miss?

torchrl/envs/libs/pettingzoo.py
```diff
@@ -78,14 +78,18 @@ class PettingZooWrapper(_EnvWrapper):
...
+                        truncated: Tensor(shape=torch.Size([3, 1]), device=cpu, dtype=torch.bool, is_shared=False)},
                     batch_size=torch.Size([3]))},
                 adversary: TensorDict(
                     fields={
                         action: Tensor(shape=torch.Size([1, 9]), device=cpu, dtype=torch.int64, is_shared=False),
                         action_mask: Tensor(shape=torch.Size([1, 9]), device=cpu, dtype=torch.bool, is_shared=False),
                         done: Tensor(shape=torch.Size([1, 1]), device=cpu, dtype=torch.bool, is_shared=False),
-                        observation: Tensor(shape=torch.Size([1, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False)},
+                        observation: Tensor(shape=torch.Size([1, 3, 3, 2]), device=cpu, dtype=torch.int8, is_shared=False),
+                        done: Tensor(shape=torch.Size([1, 1]), device=cpu, dtype=torch.bool, is_shared=False),
```
matteobettini, line 91: this misses terminated

matteobettini:
> EDIT: if you look at the latest commit, what I did is just placing done in terminated and done in each env. I think it's clearer than placing just one and let the magic happen

so placing both directly in the env? we can do this but let s make sure to have it uniformly everywhere

torchrl/envs/batched_envs.py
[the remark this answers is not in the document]
vmoens: Sorry I'm still not following at all what the first comment is about, so let me explain what we're doing here:
If there is an input data structure, we gather the reset info.

We pass the reset_keys and obtain a reset of the size of the tensordict. This is the contract of _aggregate_resets.
The goal is to support arrays of envs that are more complex than just a list of envs.
In our case, if we have more than one dimension, we flatten and call any on the last dimension to obtain something that has the shape [n_workers]. And in case we don't have any element (which can happen) we expand to the appropriate shape. You can check that this last condition is necessary by running the tests but essentially the idea is just that tensordict can miss the `"_reset"` key so the funciton just returns a scalar.
Now sure we can expand to the num_workers that's safer!

torchrl/envs/libs/vmas.py
```diff
@@ -368,7 +378,7 @@ def _step(
         if not self.het_specs:
             agent_tds = agent_tds.to_tensordict()
         tensordict_out = TensorDict(
-            source={"agents": agent_tds, "done": dones},
+            source={"agents": agent_tds, "done": dones, "terminated": dones},
```
matteobettini, line 381: so not cloning? what if done is modified in place?

vmoens: Rule of thumb is not to consider in-place modif as a possibility because it's discouraged in pytorch.

In-place modifications in pytorch should never be the rule, and we abide by this rule for efficiency. 

IMO if someone does
```
tensordict.get("done").fill_(0)
```
they are looking for problems.

But we can clone if you think it's safer

torchrl/envs/batched_envs.py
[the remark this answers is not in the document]
matteobettini: my point is just that if `needs_resetting` has to have shape [n_workers]
`needs_resetting = needs_resetting.expand(self.batch_size)` will not do that (because the batch size can be something else)

torchrl/envs/batched_envs.py
matteobettini: ``> If the point is that it should also be excluded inf the case where there is more than 1 task I totally agree!

definitely let's uniform this.

my other point is just that i thought we decided that the convention was that all the _resets are removed only in `env.reset()` and in this particular instance they are removed bofore. this has 2 problems:

- `env.reset()` cannot perform the done check appropriately (it will see a missing _reset and assume it is all true but that is not the case)(this is a problem if the reset was partial and some dones are true after the reset as they should but they cause the check to fail)
-  and that this is not uniform as the batched env _reset() will be the only one to not provide _resets

vmoens: > env.reset() cannot perform the done check appropriately (it will see a missing _reset and assume it is all true but that is not the case)(this is a problem if the reset was partial and some dones are true after the reset as they should but they cause the check to fail)

no because env.reset gets its "_reset" from the input tensordict, not the output

> and that this is not uniform as the batched env _reset() will be the only one to not provide _resets

Same I don't think it is true: per se all _reset must return a new tensordict and it's written nowhere in the contract of that function that you must copy any part of the content of the input in the output. In other words: we never read "_reset" from the output of _reset. Check the `reset` method for more context if need reassurance.

matteobettini: ok got it, my bad.

let's just test that BatchedEnv._reset() does not return _resets in all branches

torchrl/collectors/collectors.py
```diff
@@ -790,66 +791,38 @@ def iterator(self) -> Iterator[TensorDictBase]:
...
         if any_done:
             traj_ids = self._tensordict.get(("collector", "traj_ids"))
             traj_ids = traj_ids.clone()
             # collectors do not support passing other tensors than `"_reset"`
             # to `reset()`.
-            td_reset = self._tensordict.select(*done_map.keys())
-            for done_key, done in done_map.items():
-                td_reset.set(_replace_last(done_key, "_reset"), done)
-                del td_reset[done_key]
+            traj_sop = _aggregate_resets(td_reset)
```
matteobettini, line 814: For efficiency we need to pass the _reset keys, cause scanning might take a long time an be avoidable

torchrl/envs/batched_envs.py
```diff
@@ -635,19 +638,20 @@ def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
             self.shared_tensordicts[i].update_(
                 _td.select(*self._selected_reset_keys, strict=False)
             )
-
+        selected_output_keys = {
+            unravel_keys(key) for key in self._selected_reset_keys
+        } - set(self.reset_keys)
```
matteobettini, line 643: Can we do this at init time?

torchrl/envs/batched_envs.py
```diff
@@ -882,18 +884,20 @@ def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
             event.wait()
             event.clear()
 
+        selected_output_keys = {
+            unravel_keys(key) for key in self._selected_reset_keys
+        } - set(self.reset_keys)
```
matteobettini, line 889: as above

matteobettini (approved):
LGTM, let's just addrress the remaining open comments and we are good to go for me

Thanks for this btw, it is an impressive work!

PR #1542, opened 2023-09-18
[Feature] Masked specs and masked distiribution uniformity

matteobettini:
Currently our masked distributions, when receiving a mask with all `False` entries, will sample with equal probability for all events.

https://github.com/pytorch/rl/blob/fe91c4fc0e7ec6bde969a4f25ea58b247ec4dc65/torchrl/modules/distributions/discrete.py#L145

This PR brings the same behaviour to specs.

When an `action_mask`  with all `False` entries is passed to `spec.update_mask(mask)`, this will be discarded and all actions will be considered available.

Another option would be to throw an error in both masked specs and distributions for this corner case, but this would come at the price of more runtime checks.

### DIFF
--- a/docs/source/reference/envs.rst
+++ b/docs/source/reference/envs.rst
@@ -543,6 +543,9 @@ to always know what the latest available actions are. You can do this like so:
         >>>     ActionMask(action_key="action", mask_key="action_mask"),
         >>> )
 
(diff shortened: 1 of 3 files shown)


vmoens (commented):
I would rather see an error in the distribution when the mask is void than having an unintuitive behaviour in both

PR #1544, opened 2023-09-19
[BugFIx] RewardSum transform for multiple reward keys

matteobettini:
This PR extends the `RewardSum` to work with multiple reward keys.


torchrl/envs/transforms/transforms.py
```diff
@@ -4067,12 +4067,11 @@ class RewardSum(Transform):
...
     This transform accepts a list of tensordict reward keys (i.e. ´in_keys´) and tracks their cumulative
     value along each episode. When called, the transform creates a new tensordict key for each in_key named
-    ´episode_{in_key}´ where  the cumulative values are written. All ´in_keys´ should be part of the env
+    ´episode_{in_key}´ where the cumulative values are written. All ´in_keys´ should be part of the env
     reward and be present in the env reward_spec.
 
     If no in_keys are specified, this transform assumes ´reward´ to be the input key. However, multiple rewards
-    (e.g. reward1 and reward2) can also be specified. If ´in_keys´ are not present in the provided tensordict,
-    this transform hos no effect.
+    (e.g. reward1 and reward2) can also be specified.
```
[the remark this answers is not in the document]
vmoens: Can we add an example too?

torchrl/envs/transforms/transforms.py
```diff
@@ -4083,8 +4082,8 @@ def __init__(
         """Initialises the transform. Filters out non-reward input keys and defines output keys."""
         if in_keys is None:
             in_keys = ["reward"]
-        if out_keys is None and in_keys == ["reward"]:
-            out_keys = ["episode_reward"]
+        if out_keys is None:
+            out_keys = [_replace_last(in_key, "episode_reward") for in_key in in_keys]
```
vmoens, line 4086: what if the reward is named "serendipity"?

torchrl/envs/transforms/transforms.py
```diff
@@ -4094,41 +4093,30 @@ def __init__(
...
         """Resets episode rewards."""
-        # Non-batched environments
-        _reset = tensordict.get("_reset", None)
-        if _reset is None:
-            _reset = torch.ones(
-                self.parent.done_spec.shape if self.parent else tensordict.batch_size,
-                dtype=torch.bool,
-                device=tensordict.device,
-            )
+        _reset = tensordict.get("_reset", None)  # Try to get reset from root
```
vmoens, line 4096: How is this supposed to work when you have one reward for agent1 and one for agent2, and their reset differ?
Seems a bit hard to do a generic logic.
Shouldn't we have as many reward transforms as agents, each pointing to a specific done signal?

torchrl/envs/transforms/transforms.py
```diff
@@ -4094,41 +4093,30 @@ def __init__(
...
-            for in_key, out_key in zip(self.in_keys, self.out_keys):
+        for in_key, out_key in zip(self.in_keys, self.out_keys):
+            in_key = _unravel_key_to_tuple(in_key)
+            # If the root reset is None, try to get it from the same td as the reward
+            if len(in_key) > 1 and _reset is None:
+                _reset = tensordict.get(_replace_last(in_key, "_reset"), None)
+
+            if _reset is None or _reset.any():
                 if out_key in tensordict.keys(True, True):
                     value = tensordict[out_key]
```
vmoens, line 4106: see below comment

torchrl/envs/transforms/transforms.py
```diff
@@ -4094,41 +4093,30 @@ def __init__(
...
-                        tensordict[out_key] = self.parent.observation_spec[
-                            in_key
-                        ].zero()
-                    except KeyError as err:
-                        raise KeyError(
-                            f"The key {in_key} was not found in the parent "
-                            f"observation_spec with keys "
-                            f"{list(self.parent.observation_spec.keys(True))}. "
-                        ) from err
+                    tensordict[out_key] = self.parent.output_spec["full_reward_spec"][
```
vmoens, line 4116: we'll soon have `env.full_reward_spec`

vmoens (commented):
How is this supposed to work with many done signals?
I think we should wait for #1539 to land first

torchrl/envs/transforms/transforms.py
matteobettini: by default the output will be named "episode_reward",
would you prefer "episode_serendipity"?

matteobettini:
This transform is supposed to work according to the MARL grouping api.

It will look for the _reset entry in the root and consider it the default reset
For each reward key then, if it finds a _reset in its tensordict it will use that instead.

so each reward key will be by default associated with the _reset in its td and, if that is not present, it will be associated with the _reset in root (if present)

the behavior aligns with the MARL api

vmoens:
> This transform is supposed to work according to the MARL grouping api.

> It will look for the _reset entry in the root and consider it the default reset

I thought our default was that there wasn't a "_reset" at the root (only if there is a `"done"` at the root)?

vmoens:
To be clear, the direction I understood we were moving towards:

```python
TensorDIct({"_reset": reset, "nested": {"done": done}), []) # no allowed! 
TensorDIct({"nested: {"_reset": reset, "done": done}), []) # allowed
TensorDIct({"_reset": reset, "done": done}, []) # allowed
```
The first is not allowed because:
- the "_reset" is not part of the `env.reset_keys` which MUST be an iterable that follows the grouped done
- it is unclear what this reset refers to and what to do with it in specific cases: does it prevail over sub-resets? do we allow any arbitrary shape? There are way too many decisions to be taken IMO

matteobettini:
i have updated with what discussed, now i ll just have to test it

vmoens (approved):
If @matteobettini you're happy with https://github.com/pytorch/rl/pull/1544/commits/8da298fe154b3834227d43bb8ba02740b863262b I'm good with merging this

torchrl/envs/transforms/transforms.py
```diff
@@ -4109,74 +4109,158 @@ class RewardSum(Transform):
...
-    .. note:: :class:`~RewardSum` currently only supports ``"done"`` signal at the root.
-        Nested ``"done"``, such as those found in MARL settings, are currently not supported.
-        If this feature is needed, please raise an issue on TorchRL repo.
+    Args:
+        in_keys (list of NestedKeys, optional): Input reward keys.
+            All ´in_keys´ should be part of the environment reward_spec.
+            If no ``in_keys`` are specified, this transform assumes ``"reward"`` to be the input key.
+            However, multiple rewards (e.g. ``"reward1"`` and ``"reward2""``) can also be specified.
+        out_keys (list of NestedKeys, optional): The output sum keys, should be one per each input key.
+        reset_keys (list of NestedKeys, optional): the list of reset_keys to be
```
matteobettini, line 4123: Here i preferred having done keys rather than reset keys, this is because users are familiar with what a done key is and could not know about reset keys. Plus there is a 1:1 matching between the 2

test/test_transforms.py
```diff
@@ -4530,6 +4533,36 @@ def test_trans_parallel_env_check(self):
...
+    def test_trans_multi_key(
+        self, has_in_keys, n_workers=2, batch_size=(3, 2), max_steps=5
+    ):
+        torch.manual_seed(0)
+        env_fun = lambda: MultiKeyCountingEnv(batch_size=batch_size)
+        base_env = SerialEnv(n_workers, env_fun)
+        if has_in_keys:
+            t = RewardSum(in_keys=base_env.reward_keys, reset_keys=base_env.reset_keys)
+        else:
+            t = RewardSum()
```
matteobettini, line 4546: Where is t used?

torchrl/envs/transforms/transforms.py
```diff
@@ -4109,74 +4109,158 @@ class RewardSum(Transform):
...
+
+    @property
+    def in_keys(self):
+        in_keys = self.__dict__.get("_in_keys", None)
+        if in_keys in (None, []):
+            # retrieve rewards from parent env
+            parent = self.parent
+            if parent is None:
+                # fall back on in_keys = ["reward"]
+                in_keys = ["reward"]
```
matteobettini, line 4185: I think "reward" is better

matteobettini (commented):
LGTM some comments

torchrl/envs/transforms/transforms.py
vmoens: sorry I'm not following, what is better?

torchrl/envs/transforms/transforms.py
vmoens: not the way it was done: if I pass `env.done_keys` there are some duplicates (eg, truncation / termination / done).
Having a "done_keys" list to me is more dangerous because of this, and eventually the only thing we're pointing is the tree structure where the reset_keys should be found. I personally prefer to pass `reset_keys`: it's what is needed, ie there is a lower risk that refactoring the done_keys mechanism in the future will break this transform. Per se asking users to pass a list X when we interpolate a list Y that is present within the env already as env.Y seems a convoluted way of doing things.

torchrl/envs/transforms/transforms.py
matteobettini: the current one, let's just remove the comments

torchrl/envs/transforms/transforms.py
```diff
@@ -4109,74 +4109,158 @@ class RewardSum(Transform):
...
+            parent = self.parent
+            if parent is None:
+                raise TypeError(
+                    "reset_keys not provided but parent env not found. "
+                    "Make sure that the reset_keys are provided during "
+                    "construction if the transform does not have a container env."
+                )
+            reset_keys = copy(parent.reset_keys)
+            self._reset_keys = reset_keys
+        return reset_keys
```
matteobettini, line 4233: when we are using the default env keys, are we sure that `env.reward_keys` and `env.reset_keys` are in the right order? (aka if i zip the 2 lists i will get the right corresponding keys)

torchrl/envs/transforms/transforms.py
vmoens: the comment explains the line under :)

torchrl/envs/transforms/transforms.py
vmoens: Yes they are! But it's a good point

torchrl/envs/transforms/transforms.py
matteobettini: i mean the ones below

matteobettini:
LGTM

PR #1547, opened 2023-09-19
[BugFix] Improve collector buffer initialisation when policy spec is unavailable

### LINKED ISSUE #1565: [BUG] `fake_tensordict()` in `EnvBase` is incompatible with `OneHotDiscreteTensorSpec`
## Describe the bug

When using collectors like `SyncDataCollector`, the constructor will try to call the `env.fake_tensordict()` (`env` is an argument of the constructor). Unfortunately, the implementation of `env.fake_tensordict()` will generate a fake observation tensordict using something like `observation_spec.zero()`. But for an observation spec like `OneHotDiscreteTensorSpec`, an all-zero tensor is illegal, which will lead to unexpected behaviors for the downstream models.

## To Reproduce

An example code to reproduce the behavior:

```python
from typing import Optional

import torch
from tensordict import TensorDict, TensorDictBase
from torchrl.envs import EnvBase
from torchrl.data import (
    CompositeSpec,
    OneHotDiscreteTensorSpec,
)


class DemoEnv(EnvBase):
    batch_locked = True

    def __init__(self):
        super().__init__()

        # For demo purpose, params are hard coded.
        self.node_num = 3
        self.batch_size = (2,)

        self.set_seed(torch.empty((), dtype=torch.int64).random_().item())
        self._make_spec()

    def _make_spec(self):
        """Set the observation, action, and reward specs."""
        self.observation_spec = CompositeSpec(
            current_node=OneHotDiscreteTensorSpec(
                n=self.node_num,
                shape=(*self.batch_size, self.node_num),
                dtype=torch.bool,
            ),
            shape=self.batch_size,
        )
        self.state_spec = self.observation_spec.clone()
        self.action_spec = UnboundedContinuousTensorSpec(shape=self.batch_size)
        self.reward_spec = UnboundedContinuousTensorSpec(shape=self.batch_size)

    def _set_seed(self, seed: Optional[int]):
        rng = torch.manual_seed(seed)
        self.rng = rng

    def _step(self, td: TensorDictBase) -> TensorDict:
        # Trivial step function for demo purpose.
        out = TensorDict(
            {
                "next": {
                    "current_node": td["current_node"],
                    "reward": td["action"],
                    "done": td["done"],
                }
            },
            td.shape,
        )
        return out

    def _reset(self, td) -> TensorDict:
        current_node = torch.zeros(*self.batch_size, self.node_num).bool()
        current_node[:, 0] = True

        out = TensorDict(
            {
                "current_node": current_node,
            },
            self.batch_size,
        )

        return out


def trival_policy(x):
    out = 1 / torch.sum(x, dim=-1)
    assert not torch.isinf(out).any(), "Oops! Inf value!"
    return out


env = DemoEnv()
policy = TensorDictModule(
    module=trival_policy,
    in_keys=["current_node"],
    out_keys=["action"],
)

    env,
    policy,
    frames_per_batch=10,
    total_frames=100,
)
```
You can always see the assertion exception raised here.

## Expected behavior

The example code should not raise an assertion exception, which means a one-hot style tensor should be delivered to the policy module!

## System info

Though I'm using the stable version of TorchRL, this bug may also exist in the nightly version. 

```python
import torchrl, numpy, sys
```
```
None 1.24.3 3.10.11 | packaged by conda-forge | (main, May 10 2023, 18:58:44) [GCC 11.3.0] linux
```

## Reason and Possible fixes

As I said in the description section, the implementation of `env.fake_tensordict()` will generate a fake observation tensordict using something like `observation_spec.zero()`, which leads to this bug.

current main branch code:
https://github.com/pytorch/rl/blob/95b1bfec3411b154870163a777537ee0657801c3/torchrl/envs/common.py#L1674

stable branch code:
https://github.com/pytorch/rl/blob/6d030c9ec1db72c252a23dff04dcad2a553678d2/torchrl/envs/common.py#L895

Note that in other cases, the same problem may also exist for the other spec (like action spec).

## Checklist

matteobettini:
Depends on #1539
Fixes #1565

This pr makes the collector use a sample policy forward on the reset data to generate their buffer when the policy spec is not available or partial.

The prior approach initialised the policy keys by passing the policy a `env.fake_tensordict()` full of zeros, this made certain policies that use action masks or other masks throw errors as these masks were all False.

This solution increases the generality of this initialization and makes sure that the policy is fed data that is suit to it


test/test_collector.py
```diff
@@ -1366,6 +1366,26 @@ def test_reset_heterogeneous_envs():
     ).all()
 
 
+def test_policy_with_mask():
```
matteobettini, line 1369: This test fails on main

torchrl/collectors/collectors.py
```diff
@@ -619,64 +619,45 @@ def __init__(
...
             with torch.no_grad():
-                self._tensordict_out = self._tensordict_out.to(self.device)
-                self._tensordict_out = self.policy(self._tensordict_out).unsqueeze(-1)
-            self._tensordict_out = (
-                self._tensordict_out.expand(*env.batch_size, self.frames_per_batch)
-                .clone()
-                .zero_()
-            )
-        # in addition to outputs of the policy, we add traj_ids and step_count to
+                policy_out = self.policy(self._tensordict.to(self.device).clone())
```
matteobettini, line 651: We use the data in `self._tensordict` (coming from a fresh reset) to initialise the policy keys.

Before i was using a rollout but this caused problems with seeding. (i.e., if the seeding was done in the env_fun, then the rollout output in the data returned by the collector might not have been consistent with what the user expected (given that it would have been a second rollout))

torchrl/collectors/collectors.py
```diff
@@ -619,64 +619,45 @@ def __init__(
...
+            self._tensordict_out.update(policy_out)
```
matteobettini, line 652: here we add the policy keys to the `_tensordict_out`

torchrl/collectors/collectors.py
```diff
@@ -619,64 +619,45 @@ def __init__(
...
             # See #505 for additional context.
             with torch.no_grad():
-                self._tensordict_out = self._tensordict_out.to(self.device)
-                self._tensordict_out = self.policy(self._tensordict_out).unsqueeze(-1)
-            self._tensordict_out = (
-                self._tensordict_out.expand(*env.batch_size, self.frames_per_batch)
-                .clone()
-                .zero_()
-            )
-        # in addition to outputs of the policy, we add traj_ids and step_count to
```
matteobettini, line 679: above here i just removed code repetitions but no logic changes

torchrl/collectors/collectors.py
```diff
@@ -619,64 +619,45 @@ def __init__(
...
+                # the policy could have more keys than the env:
+                policy_spec = self.policy.spec
+                if policy_spec.ndim < self._tensordict_out.ndim:
+                    policy_spec = policy_spec.expand(self._tensordict_out.shape)
+                for key, spec in policy_spec.items(True, True):
+                    if key in self._tensordict_out.keys(isinstance(key, tuple)):
+                        continue
+                    self._tensordict_out.set(key, spec.zero())
+
         else:
```
matteobettini, line 644: above here i just refactored

torchrl/collectors/collectors.py
```diff
@@ -619,64 +619,45 @@ def __init__(
...
             # otherwise, we perform a small number of steps with the policy to
             # determine the relevant keys with which to pre-populate _tensordict_out.
             # This is the safest thing to do if the spec has None fields or if there is
             # no spec at all.
             # See #505 for additional context.
+            self._tensordict_out.update(self._tensordict)
```
matteobettini, line 650: we update the ` self._tensordict_out` with the real data coming from `env.reset()`

torchrl/collectors/collectors.py
```diff
@@ -619,64 +619,45 @@ def __init__(
...
             with torch.no_grad():
-                self._tensordict_out = self._tensordict_out.to(self.device)
-                self._tensordict_out = self.policy(self._tensordict_out).unsqueeze(-1)
-            self._tensordict_out = (
-                self._tensordict_out.expand(*env.batch_size, self.frames_per_batch)
-                .clone()
-                .zero_()
-            )
-        # in addition to outputs of the policy, we add traj_ids and step_count to
+                self._tensordict_out = self.policy(self._tensordict_out.to(self.device))
```
matteobettini, line 652: and feed that to the policy

vmoens (approved):
Tests are passing, so LGTM :) Thanks a mil for this!

PR #1549, opened 2023-09-20
[Test] Remove import of test class

matteobettini:
This PR removes the import of `test_env` from the `test_collector` file

### DIFF
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -9,7 +9,12 @@ import sys
 import numpy as np
 import pytest
 import torch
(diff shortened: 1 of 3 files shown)


test/test_collector.py
```diff
@@ -1562,6 +1561,100 @@ def test_multi_collector_het_env_consistency(
 
 
 class TestMultiKeyEnvsCollector:
+    @staticmethod
```
vmoens, line 1564: What about avoiding copy paste and moving this to _utils_inernal?

vmoens (approved):
wonderful

PR #1550, opened 2023-09-20
[Feature] Pettingzoo: add multiagent dimension to single agent groups

matteobettini:
Prior to this PR, in PettingZoo, groups with one agent only had the multiagent dimension removed.

This PR adds the multiagent dimension to those groups for consistency purposes.


vmoens (approved):
less code is better code

PR #1551, opened 2023-09-20
[Refactor] Fix imports

vmoens:
## Description

- Fixes robohive import that occured at root level
- Fixes the mujoco device annoying print


matteobettini (approved):
LGTM

PR #1554, opened 2023-09-20
[BugFix] PettingZoo seeding

matteobettini:
Make the `set_seed()` function call reset as per gym convention


vmoens (approved):
I trust you on this one. No need for a check I reckon

PR #1557, opened 2023-09-21
[BugFix] Load collector frames and iter

matteobettini:
The collector state loading was previously not loading the collected frames.

This fixes the issue and introduces a test to check the non regression on all collectors


vmoens (approved):
LGTM thanks!

vmoens:
@matteobettini the lint tests didn't pass

PR #1559, opened 2023-09-21
[Feature] Threaded collection and parallel envs

vmoens:
## Description

This PR proposed to automatically adapt the number of threads of parallel envs and collectors based on the number of workers that are required. 
For each, the workers are set to `num_workers` on the parent env and to `1` (if not specified otherwise) on the children.

It solves the problem that all workeres were asking for as many threads as the number of CPUs on the machine, thereby maxing the cpu workload.

For a parallel env (N workers) within a collector (M workers), what will happen is that the main process will have `M` processes, and create workers that all have 1 thread. Each worker will then reset its number of threads to `N`, and create sub-processes that each have 1 thread only.

cc @skandermoalla


torchrl/collectors/collectors.py
```diff
@@ -1097,6 +1097,15 @@ class _MultiDataCollector(DataCollectorBase):
             Defaults to ``False``.
         preemptive_threshold (float, optional): a value between 0.0 and 1.0 that specifies the ratio of workers
             that will be allowed to finished collecting their rollout before the rest are forced to end early.
+        num_threads (int, optional): number of threads for this process.
+            Defaults to the number of workers.
+            This parameter has no effect for the :class:`~SerialEnv` class.
```
matteobettini, line 1102:
```suggestion
```

skandermoalla:
Quick check: what would be the default behavior of a `SyncDataCollector` with a `ParallelEnv` passed to its constructor?

vmoens:
> Quick check: what would be the default behavior of a `SyncDataCollector` with a `ParallelEnv` passed to its constructor?

Good question

Open your script, import torch -> process has X threads (X = num of cpus)
create parallel env -> process threads reduced to N (N = num workers of the parallel env)
sync data collector -> no effect, still N threads
launches N procs, each with 1 thread

vmoens:
@matteobettini @skandermoalla I don't really see a proper way of testing this, except with (pointless?) mocks 
Agree to leave it as it is?

matteobettini (approved):
LGTM

can't we launch parallel processes and assert on `torch.get_num_threads`?

vmoens:
Ok, I implemented some tests!

skandermoalla:
Thanks so much for this. Will test with my setup and give feedback!

vmoens:
@skandermoalla @matteobettini @Vittorio-Caggiano 
With my setup, I get roughly these numbers in terms of throughput (on #1539 + this)
Always 32 procs, a machine with 96 cpus and 1 A100

**Atari (Pong-v5):**
TorchRL ParallelEnv.rollout: 10281.7697
TorchRL ParallelEnv within SyncDataCollector (to test collector overhead): 8217.6762
TorchRL GymEnv + gym async env + env.rollout: 9554.7727
TorchRL ParallelEnv (N=4/worker) across 8 collectors, async: 22023.7621
TorchRL GymEnv + gym async env (N=4/worker) across 8 collectors, async: 17997.9470
TorchRL ParallelEnv (N=4/worker) across 8 collectors, async: 10806.8827
TorchRL GymEnv + gym async env (N=4/worker) across 8 collectors, async: 10141.7132

**Myo:**
TorchRL ParallelEnv.rollout: 12346.7140
TorchRL ParallelEnv within SyncDataCollector (to test collector overhead): 10127.3826
TorchRL GymEnv + gym async env + env.rollout: 7444.5473
TorchRL ParallelEnv (N=4/worker) across 8 collectors, async: 18115.6079
TorchRL GymEnv + gym async env (N=4/worker) across 8 collectors, async: 14534.9367
TorchRL ParallelEnv (N=4/worker) across 8 collectors, async: 12997.6912
TorchRL GymEnv + gym async env (N=4/worker) across 8 collectors, async: 11727.7253

I can (and will) run some more extensive benchmarks.

To me the most instructive take is that (beyond the fact that we're now way more competitive) the sync data collector is slower than plain `env.rollout` so there is some overhead there that needs to be taken care of. The other takeaway is that sync data collectors don't really help here

Anyhow, good progress, team!

skandermoalla:
Thanks a lot for this! It'd be great if the code of the benchmarks is also made available so that we can compare.

PR #1562, opened 2023-09-21
[Refactor] Use `wait` instead of `is_set` to get results in ParallelEnv

vmoens:
## Description

is_set is much slower in some cases, and the speed of `_reset` and `_step` is bound by the slowest worker anyway.


matteobettini (approved):
LGTM

PR #1563, opened 2023-09-21
[BugFix] DDPG select also critic input for actor loss

matteobettini:
Before we were only selecting the actor input and output keys when computing the actor loss (which involves a forward critic pass).

This is limiting as the critic can have extra inputs (for example a fully observable state or a centralized state in MARL)

This PR fixes that by extending the selected keys.

### DIFF
--- a/torchrl/objectives/ddpg.py
+++ b/torchrl/objectives/ddpg.py
@@ -236,7 +236,7 @@ class DDPGLoss(LossModule):
         self._set_in_keys()
 
     def _set_in_keys(self):
-        keys = {
+        in_keys = {
             unravel_key(("next", self.tensor_keys.reward)),
             unravel_key(("next", self.tensor_keys.done)),
             *self.actor_in_keys,
@@ -244,7 +244,7 @@ class DDPGLoss(LossModule):
             *self.value_network.in_keys,
             *[unravel_key(("next", key)) for key in self.value_network.in_keys],
         }
-        self._in_keys = sorted(keys, keys=str)
+        self._in_keys = sorted(in_keys, keys=str)
 
     @property
     def in_keys(self):
--- a/torchrl/objectives/ddpg.py
+++ b/torchrl/objectives/ddpg.py
@@ -11,10 +11,10 @@ from dataclasses import dataclass
 from typing import Tuple
 
 import torch
-
 from tensordict.nn import dispatch, make_functional, repopulate_module, TensorDictModule
 from tensordict.tensordict import TensorDict, TensorDictBase
-from tensordict.utils import NestedKey
+
+from tensordict.utils import NestedKey, unravel_key
 from torchrl.modules.tensordict_module.actors import ActorCriticWrapper
 from torchrl.objectives.common import LossModule
 from torchrl.objectives.utils import (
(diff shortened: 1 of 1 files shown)


torchrl/objectives/ddpg.py
```diff
@@ -293,7 +293,9 @@ def _loss_actor(
         self,
         tensordict: TensorDictBase,
     ) -> torch.Tensor:
-        td_copy = tensordict.select(*self.actor_in_keys).detach()
+        td_copy = tensordict.select(
+            *self.actor_in_keys, *self.value_network.in_keys
```
vmoens, line 297: The value net takes as input the output of the actor network, so it could be that the out_keys of the actor are not there yet. Either set `strict=False` (dangerous, because we could end up having a long trace stack in the error when calling actor_network because of a missing key) or do this

```python
        td_copy = tensordict.select(
            *self.actor_in_keys
        ).detach()
        td_copy = self.actor_network(
            td_copy,
            params=self.actor_network_params,
        )
        # check if the critic has some keys that can't be found in in or out_keys from actor
        critic_exclusive = (set(self.actor_in_keys) | set(self.actor_in_keys)) ^ set(self.value_network.in_keys)
        if critic_exclusive:
            td_copy = td_copy.update(tensordict.select(*critic_exclusive))
        td_copy = self.value_network(
            td_copy,
            params=self._cached_detached_value_params,
        )
```

vmoens (commented):
Makes sense, but not sure we cover all use cases. 
Can we add a test in TestDDPG?

Can we also change line 243:
```python
        self._in_keys = list(set(keys))
```
into
```python
        self._in_keys = sorted(set(keys), keys=str)
```
?

torchrl/objectives/ddpg.py
matteobettini: the actor is called in the next line of code, so its keys are assured to be there no?
i do not understand the problem and I would like to avoid adding further select and updates
can you explaint a bit more what you mean?

torchrl/objectives/ddpg.py
```diff
@@ -233,14 +236,15 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
         self._set_in_keys()
 
     def _set_in_keys(self):
-        keys = [
-            ("next", self.tensor_keys.reward),
-            ("next", self.tensor_keys.done),
+        in_keys = {
+            unravel_key(("next", self.tensor_keys.reward)),
+            unravel_key(("next", self.tensor_keys.done)),
```
vmoens, line 241: we should check but since it's a TensorDictModuleBase the in_keys are unravelled by default
https://github.com/pytorch-labs/tensordict/blob/accd8a4a31ec749f52e75a87a875424652069163/tensordict/nn/common.py#L474-L495

So I think we can spare the effort of doing that here

matteobettini: but even if they are already unraveled, we are creating a new tuple `("next", already_unraveled_key)` which could be 
`("next","done")` or `("next",("nested", "done"))`

matteobettini: that is why i am only unravleing the ones where we are putting next

vmoens: look at the link: they will be unravelled after you set them in theory

matteobettini: funky stuff! i ll make sure to test it

vmoens: Ah maybe not for properties, this one's harder actually
Let's keep it as it is

vmoens (approved):

PR #1570, opened 2023-09-23
[BugFix] Fix run_type_checks

vmoens:
Solves a bug when `run_type_checks=True` where a composite spec is queried with a non-existent `get` method.
Also adds tests for `env._run_type_checks`.

Note that this test may fail if the env is substituted with a GymEnv as GymEnv (and others) do the casting of the obs/reward/done internally during the step computation. This trades off security (ie, we can make sure that the dtype is always appropriate) to consistency between gym and torchrl (the dtype in torchrl can differ wrt gym at the expense of loosing information).


matteobettini (approved):
LGTM

PR #1571, opened 2023-09-24
[Test] Skip threading tests in OSX

vmoens:
## Description

OSX tests fail because the number of threads defined in sub-processes impact those from the main process.


matteobettini (approved):
LGTM

PR #1580, opened 2023-09-28
[Lint] Add TorchFix linter

kit1980:
Add TorchFix (https://github.com/pytorch/test-infra/tree/main/tools/torchfix) flake8 plugin.

To make it pass:

1. Update `from functorch import` to `from torch.func import vmap`.
2. Update deprecated `torch.norm` to `torch.linalg.norm`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
         additional_dependencies:
           - flake8-bugbear==22.10.27
           - flake8-comprehensions==3.10.1
-
+          - torchfix==0.0.2
 
   - repo: https://github.com/PyCQA/pydocstyle
     rev: 6.1.1
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,6 +15,15 @@ per-file-ignores =
     test/smoke_test_deps.py: F401
     test_*.py: F841, E731, E266
     test/opengl_rendering.py: F401
+    test/test_modules.py: F841, E731, E266, TOR101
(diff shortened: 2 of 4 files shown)


kit1980:
"Continuous Benchmark" fails unrelated to this PR:

> ImportError: cannot import name 'unravel_key' from 'tensordict'

vmoens (approved):
LGTM thanks!

vmoens:
The olddeps tests are failing because vmap cannot be found.
We try to be compatible with older versions of pytorch to support users that rely on these (for instance, libraries that permit deployment of pytorch model on hardware and that depend on a specific pytorch version).
Any idea how we can solve that issue?

kit1980:
@vmoens Thanks for looking.
I think I'll just exclude the files with `import functorch` from the linter and revert the changes to them.

kit1980:
@vmoens I've updated the PR to ignore functorch.
Please take another look.
Eventually I want to add a configuration to specify PyTorch version for TorchFix, but it's not there yet.

vmoens (approved):
LGTM

kit1980:
Hi @vmoens 
I wonder if you still want to support `import functorch`?
Functorch moved into torch.func long time ago.
RL readme says "install the latest (nightly) PyTorch release or the latest stable version of PyTorch", and both of them of course have torch.func.

vmoens:
Unfortunately functorch.dim is still a thing that we need and it hasn't moved to core.
But a bunch of people from core (including myself) are working on productionalizing it and porting it where it belongs!

PR #1581, opened 2023-09-28
[Feature] Make advantages compatible with Terminated, Truncated, Done

vmoens:
I manually tested that everything was working without truncated key for bc compatibility. 
Not sure if we should test that things work if terminated is missing in all losses as it's very redundant and the cost tests are already very long.
I didn't include the truncated -> done fallback in the docstrings, maybe we should but I don't want to overload them.

Depends on #1539


matteobettini:
Could you merge main?

vmoens:
You should be able to do it

matteobettini:
no, I can't
Screenshot 2023-10-02 at 08 37 25

torchrl/objectives/value/functional.py
```diff
@@ -259,23 +279,40 @@ def vec_generalized_advantage_estimate(
...
+        == done.shape
+        == terminated.shape
+    ):
+        print(
+            next_state_value.shape,
+            state_value.shape,
+            reward.shape,
+            done.shape,
+            terminated.shape,
+        )
```
matteobettini, line 306:
```suggestion
```

torchrl/objectives/value/functional.py
```diff
@@ -712,26 +832,26 @@ def td_lambda_return_estimate(
...
-            nd = not_done
-            _gamma = _gamma.unsqueeze(-2) * nd
             for i in reversed(range(k, T)):
-                g = reward[..., i, :] + _gamma[..., i, :] * (
-                    (1 - _lambda) * next_state_value[..., i, :] + _lambda * g
-                )
+                dn = done[..., i, :].int()
+                nv = next_state_value[..., i, :]
+                g = g * (1 - dn) + nv * dn
+                g = reward[..., i, :] + _gamma * ((1 - _lambda) * nv + _lambda * g)
```
matteobettini, line 854: it seems like here the logic changed quite a bit.
I suggest writing some in-line comments for this very crucial bits

torchrl/objectives/value/functional.py
```diff
@@ -948,46 +1098,47 @@ def _is_scalar(tensor):
...
+        v3 = (gammas * lambdas).squeeze(-1) * next_state_value * not_done
         v3[..., :-1] = 0
         out = _custom_conv1d(
-            reward + (gammas * (1 - lambdas)).squeeze(-1) * next_state_value + v3, dec
+            reward
+            + gammas.squeeze(-1)
+            * next_state_value
+            * (1 - lambdas.squeeze(-1) * not_done)
+            + v3,
+            dec,
```
matteobettini, line 1132: for bits like this I am not super able to check, so I am. trusting the tests

matteobettini (commented):
Thanks for this!

Since there are so many changes, I ll do another pass when main is merged,

but for now here are some comments

torchrl/objectives/value/advantages.py
```diff
@@ -483,10 +495,14 @@ def forward(
...
             tensordict (TensorDictBase): A TensorDict containing the data
-                (an observation key, "action", ("next", "reward"), ("next", "done") and "next" tensordict state
-                as returned by the environment) necessary to compute the value estimates and the TDEstimate.
-                The data passed to this module should be structured as :obj:`[*B, T, F]` where :obj:`B` are
-                the batch size, :obj:`T` the time dimension and :obj:`F` the feature dimension(s).
+                (an observation key, ``"action"``, ``("next", "reward")``,
+                ``("next", "done")``, ``("next", "terminated")``, and ``"next"``
+                tensordict state as returned by the environment) necessary to
+                compute the value estimates and the TDEstimate.
+                The data passed to this module should be structured as
```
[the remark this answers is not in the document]
vmoens: Right now the logic is that `time_dim = tensordict.ndim-1` (the time dimension is the last of the tensordict you pass as input -- but if the data is within a nested tensordict with more dims it's still going to work). I can adapt the docstrings to say that.

torchrl/objectives/value/functional.py
```diff
@@ -456,19 +522,29 @@ def td1_return_estimate(
...
         )
 
+    done_but_not_terminated = (done & ~terminated).int()
     if rolling_gamma:
-        gamma = gamma * not_done
+        gamma = gamma * not_terminated
         g = next_state_value[..., -1, :]
         for i in reversed(range(T)):
-            g = returns[..., i, :] = reward[..., i, :] + gamma[..., i, :] * g
+            # if not done and not terminated, get the bootstrapped value
```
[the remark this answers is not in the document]
vmoens: I don't get it sorry
Most values at not done and not terminated no?

torchrl/objectives/value/functional.py
vmoens: Only in apparence, before we were integrating the not_done in the gamma, now we do it on the gain.
The reason why this is needed is that we want for the next value to be equal to next_state_value when done=True. 
Before this was working because the next value was going to be always 0 when done=True as truncated was also True, but this does not hold anymore.

Happy to give a small comment on what we're doing oc!

torchrl/objectives/value/functional.py
vmoens: Yeah this is not super clear, but essentially as long as the tests show that the result is equal to the non vectorized version we should be fine IMO

torchrl/objectives/value/functional.py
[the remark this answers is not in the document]
matteobettini: i am just saying that not done implies not terminated

torchrl/objectives/value/advantages.py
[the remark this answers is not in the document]
matteobettini: yes that is what i was trying to say

write in the docs that we support tensors [*B,T,*F]  and by default we find T at  `time_dim = tensordict.ndim-1`

torchrl/objectives/value/functional.py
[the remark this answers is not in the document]
vmoens: Oh in this case yes, we could just say not done!

examples/multiagent/maddpg_iddpg.py
```diff
@@ -170,13 +174,6 @@ def train(cfg: "DictConfig"):  # noqa: F821
 
         sampling_time = time.time() - sampling_start
 
-        tensordict_data.set(
-            ("next", "done"),
-            tensordict_data.get(("next", "done"))
-            .unsqueeze(-1)
-            .expand(tensordict_data.get(("next", env.reward_key)).shape),
-        )  # We need to expand the done to match the reward shape
-
```
matteobettini, line 179: thanks for this

the same code was used in the multiagent tutorial, I don't know if we want to update that too.
I can do it later if you want

matteobettini (approved):
LGTM

examples/multiagent/maddpg_iddpg.py
vmoens: How should we do that? Hard to import from examples smth in the tutorials, and re-coding it seems a bit awkward.
To it'd make sense to have that transform in `transforms/multiagent.py` but it's up to you.
Maybe we can make all of this more uniform in a separate PR? I don't feel comfortable enough with this to make the docstrings and tests on my own tbh.

PR #1586, opened 2023-10-01
[Feature] Support for GRU

vmoens:


test/mocking_classes.py
```diff
@@ -444,7 +444,7 @@ def __new__(
                 action_spec = action_spec_cls(n=7, shape=batch_size)
             else:
                 action_spec_cls = OneHotDiscreteTensorSpec
-                action_spec = action_spec_cls(n=7, shape=(*batch_size, 7))
+                action_spec = action_spec_cls(n=7, shape=batch_size)
```
vmoens, line 447:
```suggestion
                action_spec = action_spec_cls(n=7, shape=(*batch_size, 7))
```

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+      compatible with single step inputs and the latter being compatible with
+      multi-step. This class enables both usages.
+
+
+    After construction, the module is *not* set in temporal mode, ie. it will
+    expect single steps inputs.
+
+    If in temporal mode, it is expected that the last dimension of the tensordict
+    marks the number of steps. There is no constrain on the dimensionality of the
+    tensordict (except that it must be greater than one for temporal inputs).
```
matteobettini, line 423: so, just to clarify, if I have 
- an input tensor of shape [*B,T,*N,F]
- in a tensordict with shape [*B,T]

my *N dimensions are treated as batch? 
Is this usecase supported?

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+        device (torch.device or compatible): the device of the module.
+        gru (torch.nn.GRU, optional): a GRU instance to be wrapped.
+            Exclusive with other nn.GRU arguments.
+
+    Attributes:
+        temporal_mode: Returns the temporal mode of the module.
+
+    Methods:
+        set_recurrent_mode: controls whether the module should be executed in
+            temporal mode.
```
matteobettini, line 465: the attribute is `temporal_mode` but the method is `set_recurrent_mode` is it possible to unify naming?

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+        in_keys=None,
+        out_key=None,
+        out_keys=None,
+        device=None,
+        gru=None,
+    ):
+        super().__init__()
+        if gru is not None:
+            if not gru.batch_first:
+                raise ValueError("The input lstm must have batch_first=True.")
```
matteobettini, line 555:
```suggestion
                raise ValueError("The input gru must have batch_first=True.")
```

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+            if gru.bidirectional:
+                raise ValueError("The input lstm cannot be bidirectional.")
```
matteobettini, line 557:
```suggestion
                raise ValueError("The input gru cannot be bidirectional.")
```

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+            if input_size is not None or hidden_size is not None:
+                raise ValueError(
+                    "An LSTM instance cannot be passed along with class argument."
```
matteobettini, line 560:
```suggestion
                    "A GRU instance cannot be passed along with class argument."
```

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+                )
+        else:
+            if not batch_first:
+                raise ValueError("The input lstm must have batch_first=True.")
```
matteobettini, line 564:
```suggestion
                raise ValueError("The input gru must have batch_first=True.")
```

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+            if bidirectional:
+                raise ValueError("The input lstm cannot be bidirectional.")
```
matteobettini, line 566:
```suggestion
                raise ValueError("The input gru cannot be bidirectional.")
```

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+        elif out_key:
+            out_keys = [out_key, *self.DEFAULT_OUT_KEYS]
+
+        in_keys = unravel_key_list(in_keys)
+        out_keys = unravel_key_list(out_keys)
+        if not isinstance(in_keys, (tuple, list)) or (
+            len(in_keys) != 2 and not (len(in_keys) == 3 and in_keys[-1] == "is_init")
+        ):
+            raise ValueError(
+                f"LSTMModule expects 3 inputs: a value, and two hidden states (and potentially an 'is_init' marker). Got in_keys {in_keys} instead."
```
matteobettini, line 597: maybe do a search on LSTM to do automatic renaming to GRU

matteobettini, line 597:
```suggestion
                f"GRUModule expects 2 inputs: a value, and a hidden state (and potentially an 'is_init' marker). Got in_keys {in_keys} instead."
```

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+            )
+        if not isinstance(out_keys, (tuple, list)) or len(out_keys) != 2:
+            raise ValueError(
+                f"LSTMModule expects 3 outputs: a value, and two hidden states. Got out_keys {out_keys} instead."
```
matteobettini, line 601: same

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+
+    @property
+    def temporal_mode(self):
+        return self._temporal_mode
+
+    @temporal_mode.setter
+    def temporal_mode(self, value):
+        raise RuntimeError("temporal_mode cannot be changed in-place. Call `module.set")
+
+    def set_recurrent_mode(self, mode: bool = True):
```
matteobettini, line 642: naming convention differences

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+                # we assume that the tensordict can be flattened like this
+                nelts = prod(tensordict_shaped.shape[: dims_to_flatten + 1])
+                tensordict_shaped = tensordict_shaped.apply(
+                    lambda value: value.flatten(0, dims_to_flatten),
+                    batch_size=[nelts, tensordict_shaped.shape[-1]],
+                )
+        else:
+            tensordict_shaped = tensordict.reshape(-1).unsqueeze(-1)
+
+        is_init = tensordict_shaped.get("is_init").squeeze(-1)
```
matteobettini, line 700: here we are assuming `is_init` has the tensordict shape [*B,T]?

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+            # let's recover our original shape
+            tensordict_shaped = _inv_pad_sequence(tensordict_shaped, splits).reshape(
+                tensordict_shaped_shape
+            )
+
+        if shape != tensordict_shaped.shape or tensordict_shaped is not tensordict:
+            tensordict.update(tensordict_shaped.reshape(shape))
+        return tensordict
+
+    def _lstm(
```
matteobettini, line 741: _gru?

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -399,3 +400,379 @@ def _lstm(
...
+        )
+        batch, steps = value.shape[:2]
+        device = value.device
+        dtype = value.dtype
+        # packed sequences do not help to get the accurate last hidden values
+        # if splits is not None:
+        #     value = torch.nn.utils.rnn.pack_padded_sequence(value, splits, batch_first=True)
+        if is_init.any() and hidden is not None:
+            hidden[is_init] = 0
+        val, hidden = self._lstm(value, batch, steps, device, dtype, hidden)
```
matteobettini, line 728: _gru?

matteobettini (approved):
LGTM, some comments

In particular, for MARL, i am interested in understanding if it could work  wit

- an input tensor of shape [*B,T,*N,F]
- in a tensordict with shape [*B,T]

torchrl/modules/tensordict_module/rnn.py
vmoens: Yes, see the `test_single_step_vs_multi` test

torchrl/modules/tensordict_module/rnn.py
vmoens: not sure what makes you think that? If so is it a problem? Can you elaborate?

torchrl/modules/tensordict_module/rnn.py
matteobettini: in that test the *B is part of the tensordict.

Here I am asking about the *N 
like the agent dimension in MARL for example, after the T, but before the F

torchrl/modules/tensordict_module/rnn.py
matteobettini: - What makes me think that is that we consider its last dimension as T in the lines after
- What might become a problem is that 1) we are hardcoding the name "is_init". 2) it might have a different shape as it follows a done no?

vmoens: I think this issue and the one above boil down to what we want to accept as input to this module.
We need to have a way to tell what the time dimension is.
- One option is to check the dim names but it's hard to get people to adopt that. I tried and got multiple issues of people saying that their td was lacking the "time" dim name and they did not know what to do.
- Another option is to pass the time dim in the constructor but it's brittle. If you say `time_dim=3` you won't be able to change the batch size, if you say `time_dim=-3` you need to say with respect to what (obs, init, tensordict?)

A regular usage is assumed to be that the time is the last dim of the tensordict, but it's true that we could cover others. Otherwise we assume that the TD provided is always the root and always has time dim at the end. Handling batch-dims after the time dim can also be considered.

Another issue is that we assume that the hidden state can more or less be indexed by is_init but i'm not sure what this looks like in cases where you have multiple done states and multiple observations. The batch size of the obs could differ from the done state (eg in VMAS IIUC you have one done state per group of agents but one obs per agent). 
In MARL settings you will also need to consider if the LSTM reads the aggregated observations of not, if you have one LSMT per agent or not etc. Overall it starts to look like a different class to me.

TL;DR 
Happy to have a follow-up work on GRU and LSTM to make them compatible with more convoluted data structures, but do you think we can integrate GRU as it is now since it applies to most single agents use cases already?
If you have a precise list of queries of what this class should also cover i'd be glad to look at it. 
What I would need is: what are the data structures that need to be considered, what behaviour we expect, and what we want to explicitly forbid because there is a solution. To give you an example, IMO the example of the done that has a shape that differs from the obs can be easily solved by forbidding that and telling users to append an rename + unsqueeze + expand transform before (which all work as regular nn.Modules).

vmoens: > i would assume that is_init, the input, and the recurrent state have all the same batch_size and the last dim respectively of 1, F, H. In the example above, is_init might have to be expanded into the group before passing it to the module

my view in this case is (1) raise an exception because we can't really tell what to do anticipatively with a is_init that has a mismatching shape and (2) provide info on how to do the shape modif

vmoens: If only first batch dims were working with rnns, sigh!

PR #1590, opened 2023-10-02
[Performance] Reduce key accessing in transforms

matteobettini:


torchrl/envs/transforms/transforms.py
```diff
@@ -208,8 +208,9 @@ def _call(self, tensordict: TensorDictBase) -> TensorDictBase:
         :meth:`TransformedEnv.reset`.
 
         """
+        tensordict_keys = tensordict.keys(include_nested=True)
         for in_key, out_key in zip(self.in_keys, self.out_keys):
-            if in_key in tensordict.keys(include_nested=True):
+            if in_key in tensordict_keys:
```
vmoens, line 213: not sure, what about making it a set then?
Another easy option is
```python
val = tensordict.get(in_key, default=None)
if val is not None:
    ...
```
that way you do the get and assertion in one go

matteobettini: but that would imply multiple gets

my option gets the keys once and for all instead of trying to get each key

matteobettini: you can imagine that in nested tensordicts with many levels, doing a get for each key when the `_apply_transform` is not implemented is quite expensive

matteobettini: why a set?

vmoens: a set to avoid scanning through nested td multiple times
by using a set we just scan it once and then query have a constant time

vmoens: > Author

that's not really true if you look at how `keys(True)` works: you get an iterable once, but every time you query it you reinstantiate it.
This solution checks the key presence twice. Using a set it's more like 1.5x. Using get it's just 1x: TensorDIct is designed such that the method I was pointing before (with `get(..., None)`) is the most effective (unless get makes a stack but if it does, the key is there and hence you'll be calling `_apply_transform` anyway.

matteobettini: ok i see for the set,

but i still do not get how multiple gets is better than scanning the td for keys before the loop when the _apply_tranform is not inplemented.

Say i have multimple keys `("a","b","c","d")` and `("a","b","c","e")` and many like this.
Then every time i run a get i will access  `("a","b","c")`.
If i call it at the beginning i will access `("a","b","c")` just once

vmoens: Quick example:
```python
import timeit

d = {}
for i in range(100):
    inner = d[str(i)] = {}
inner["a"] = 0
inner["b"] = 1
td = TensorDict(
    d, []
)
key_a = tuple(str(i) for _ in range(100)) + ("a",)
key_b = tuple(str(i) for _ in range(100)) + ("b",)

def query0(key_a=key_a, td=td):
    if key_a in td.keys(True):
        return td.get(key_a)
    return None
def query1(key_a=key_a, td=td):
    return td.get(key_a, None)
print(min(*timeit.repeat("query0()", globals={"query0": query0}, number=10_000)))
print(min(*timeit.repeat("query1()", globals={"query1": query1}, number=10_000)))
```

results in 
```
0.09300343899999941
0.04636183299999974
```

vmoens: Interestingly the set idea is really bad actually
```python
def query0(key_a=key_a, td=td):
    out = []
    keys = td.keys(True)
    if key_a in keys:
        out.append(td.get(key_a))
    if key_b in keys:
        out.append(td.get(key_b))
    return out
def query1(key_a=key_a, td=td):
    return [td.get(key_a, None), td.get(key_b, None)]
def query2(key_a=key_a, td=td):
    out = []
    keys = set(td.keys(True))
    if key_a in keys:
        out.append(td.get(key_a))
    if key_b in keys:
        out.append(td.get(key_b))
    return out

print(min(*timeit.repeat("query0()", globals={"query0": query0}, number=10_000)))
print(min(*timeit.repeat("query1()", globals={"query1": query1}, number=10_000)))
print(min(*timeit.repeat("query2()", globals={"query2": query2}, number=10_000)))

```
gives
```
0.18625211699999955
0.10194817
1.5792532490000006
```

matteobettini: but does this relate to my previous comment?
the case here is about accessing multiple nested keys when they are all None

vmoens: Like this?
```python
def query0(key_c=key_c, key_d=key_d, td=td):
    out = []
    keys = td.keys(True)
    if key_c in keys:
        out.append(td.get(key_c))
    if key_d in keys:
        out.append(td.get(key_d))
    return out
def query1(key_c=key_c, key_d=key_d, td=td):
    return [td.get(key_c, None), td.get(key_d, None)]
def query2(key_c=key_c, key_d=key_d, td=td):
    out = []
    keys = set(td.keys(True))
    if key_c in keys:
        out.append(td.get(key_c))
    if key_d in keys:
        out.append(td.get(key_d))
    return out

print(min(*timeit.repeat("query0()", globals={"query0": query0}, number=10_000)))
print(min(*timeit.repeat("query1()", globals={"query1": query1}, number=10_000)))
print(min(*timeit.repeat("query2()", globals={"query2": query2}, number=10_000)))
```
i get
```
0.19403007200000033
0.09681646599999993
1.4667162099999977
```
so get is still 2x as fast, as expected

matteobettini: for example

```python
    n_in_keys = 100
    n_nesting_levels = 100

    in_keys = []
    for i in range(n_in_keys):
        key = tuple([str(j) for j in range(n_nesting_levels)] + [str(i)])
        in_keys.append(key)

    td = TensorDict({key[:-1]: TensorDict({}, []) for key in in_keys}, [])

    def query0():
        td_keys = set(td.keys(True))
        for in_key in in_keys:
            if in_key in td_keys:
                return td.get(in_key)

    def query1():
        for in_key in in_keys:
            td.get(in_key, None)

    print(min(*timeit.repeat("query0()", globals={"query0": query0}, number=10_000)))
    print(min(*timeit.repeat("query1()", globals={"query1": query1}, number=10_000)))
```
gives

```
1.9879649999999998
39.815401540999986
```

now you can play with `n_in_keys` and `n_nesting_levels`, but this is just to make the point i was illustrating

vmoens: that doesn't count because you return early in one cand and not the other no?
if you do 
```
    def query1():
        for in_key in in_keys:
            out = td.get(in_key, None)
            if out is None: return None
```
 you should get the same runtime I think

matteobettini: i do not return early because the ` if in_key in td_keys` is never triggered

matteobettini: here is the updated version 
```python
    n_in_keys = 100
    n_nesting_levels = 100

    in_keys = []
    for i in range(n_in_keys):
        key = tuple([str(j) for j in range(n_nesting_levels)] + [str(i)])
        in_keys.append(key)

    td = TensorDict({key[:-1]: TensorDict({}, []) for key in in_keys}, [])

    def query0():
        td_keys = set(td.keys(True))
        for in_key in in_keys:
            if in_key in td_keys:
                td.get(in_key)

    def query1():
        for in_key in in_keys:
            td.get(in_key, None)

    print(min(*timeit.repeat("query0()", globals={"query0": query0}, number=10_000)))
    print(min(*timeit.repeat("query1()", globals={"query1": query1}, number=10_000)))
```
```
2.0131902089999993
40.72629070799999

```

vmoens: Actually I realize I don't get your example, why does it return in one case and not the other?

To come back at the transforms, we never return early (like in the example above). If we allow missing keys we just skip them.

I think this example is closer to what we want to do:
```python


n_in_keys = 100
n_nesting_levels = 100

in_keys = []
for i in range(n_in_keys):
    key = tuple(str(j) for j in range(n_nesting_levels)) + (str(i),)
    in_keys.append(key)

td = TensorDict({key[:-1]: TensorDict({}, []) for key in in_keys}, [])


def query0():
    td_keys = set(td.keys(True))
    #td_keys = td.keys(True)
    out = []
    for in_key in in_keys:
        if in_key in td_keys:
            out.append(td.get(in_key))
    return out

def query1():
    #td_keys = set(td.keys(True))
    td_keys = td.keys(True)
    out = []
    for in_key in in_keys:
        if in_key in td_keys:
            out.append(td.get(in_key))
    return out

def query2():
    out = []
    for in_key in in_keys:
        _out = td.get(in_key, None)
        if _out is not None:
            out.append(_out)
    return out

print("0", min(*timeit.repeat("query0()", globals={"query0": query0}, number=1_000)))
print("1", min(*timeit.repeat("query1()", globals={"query1": query1}, number=1_000)))
print("2", min(*timeit.repeat("query2()", globals={"query2": query2}, number=1_000)))
```


in this case using `set` is faster but this is only because 
(1) the structure is hugely nested and (2) we access at the leaves and (3) we access every time at the deepest level and many times.

Reasonably, a transform won't access more than 2 or 3 items in the tensordict, and return an error if it's absent unless calles at reset (which is not very frequent). So most isntances will be closer to my examples above.
We actually removes a lot of usage of `set` in torchrl and tensordict throughout development because it's usually an overkill. It's very rarely a benefit 

TL;DR: unless we can show that it's faster for regular use cases (couple of keys, couple of nesting layers) I don't think using set or caching the keys is wise.

vmoens: Another examples of why I think this won't work long term:
if we use set, we scan through all the keys.
But when we assess if a key is a member of keys, the way tensordict does it is that lower levels are scanned first and others later. 
So if an env has many levels (which often happens with infos, that are not part of most transforms) we will be reading all the keys of all levels to possibly gain a few ms when the key is absent, but in most cases the entry won't be very deep in the structure and we will have read everything for no reason. There is thsi thing that in general assessing if a key is there is slower if the key is absent than if it is present.
Since in most cases the key is present, the runtime of absent keys should not matter too much in how we design transforms, and we should certainly not scan all keys every time.

matteobettini: ok we can do it this way

I am not so sure that the keys are more present than they are absent though since if a transform does not implement `_call` all keys are guaranteed to be absent

vmoens: I'm not really following this comment 
All transforms implement `_call` (since it's in the base), if they override `_call` to something else this loop isn't executed.
You mean `_apply_transform`? 

Do you have examples of such transforms? How extensive is the problem?

matteobettini: yes sorry i meant `_apply_transform`

15 transforms implement it and the others don't
i don't know how many transforms we have but that - 15 is the number that will always get None

vmoens: This is the list of transforms we have
```
    ActionMask => overrides _call
    BinarizeReward => reward is missing during reset
    CatFrames => no impact
    CatTensors => no impact
    CenterCrop => no impact
    ClipTransform => no impact
    Compose => no impact
    DeviceCastTransform => no impact
    DiscreteActionProjection => no impact
    DoubleToFloat => no impact
    DTypeCastTransform => no impact
    ExcludeTransform => no impact
    FiniteTensorDictCheck => no impact
    FlattenObservation => no impact
    FrameSkipTransform => no impact
    GrayScale => no impact
    gSDENoise => no impact
    InitTracker => no impact
    KLRewardTransform => reward is missing during reset
    NoopResetEnv => no impact
    ObservationNorm => no impact
    ObservationTransform => no impact
    PermuteTransform => no impact
    PinMemoryTransform => no impact
    R3MTransform => no impact
    RandomCropTensorDict => no impact
    RenameTransform => no impact
    Resize => no impact
    RewardClipping => reward is missing during reset
    RewardScaling => reward is missing during reset
    RewardSum => reward is missing during reset
    Reward2GoTransform => no impact
    SelectTransform => no impact
    SqueezeTransform => no impact
    StepCounter => no impact
    TargetReturn => reward is missing during reset
    TensorDictPrimer => no impact
    TimeMaxPool => no impact
    ToTensorImage => no impact
    UnsqueezeTransform => no impact
    VecGymEnvTransform => no impact
    VecNorm => no impact
    VC1Transform => no impact
    VIPRewardTransform => reward is missing during reset, not sure it has an impact though
    VIPTransform => no impact
```
Which make 6 transforms where this is needed, and 40 where it isn't.

What I understand from your PR (the problem you're trying to solve, even though this isn't stated anywhere so I'm just guessing here) is that you want to reduce the number of accesses to the keys of the td.
What you want also is that if a key is absent things stay efficient, in which case a set could be a good idea.
This only works when we allow missing keys which is only turned on during reset to this is already just a tiny fraction of the calls to `_call`.
During these calls, a key can be absent because it can't be found at reset but only during step, which again is something rather peculiar. So we're at the edge case of the rare case. Even in cases where the key is missing, set is only more efficient if there are multiple keys to be checked, which reduces again the subspace where this change would be beneficial.
Roughly, a wild guess would put using set at being beneficial for < 2% of the envs (assuming 6/40 transforms * 1/2 calls to reset -- assuming one reset at each step -- * 1/5 where there are many keys to be checked in the transform).
In all other cases using `get(..., None)` is more efficient. 

Happy to be shown wrong though!

vmoens: > yes sorry i meant _apply_transform

> 15 transforms implement it and the others don't
> i don't know how many transforms we have but that - 17 is the number that will always get None

> i would say majority of transoforms does not implement it

I could be wrong but I think all that don't implement it do override _call so this PR won't affect them anyway

vmoens (approved):

Issue #1574, opened 2023-09-25
[Performance] Prioritised TensorDict replay buffers use for loops over the batch dimension


matteobettini:
In prioritised tensordict replay buffers, the `update_tensordict_priority`, method performs a for loop over the batch dimension

https://github.com/pytorch/rl/blob/434fe58b44cb46650e80ff18030d43a45eb532f2/torchrl/data/replay_buffers/replay_buffers.py#L760-L763

This causes significant slowdowns as this is the vectorised dimension used in the training pipelines and can get to really high sizes.

This method is called every time the buffer is extended or the priorities are updated.

vmoens:
Any update on this? Can I help?

matteobettini:
I am not super sure about all the cases that were envisioned when the component was created this way or why we were doing this. 
If you can give me some context i can try to fix it or if you already know an easy fix feel free to do it.

vmoens:
We just wanted to be able to handle lists of tensordicts i guess, and it should work ok with a list storage and stuff that you can't necessarily stack well, no more insight than this I'm afraid.
But in general I think that if all tests pass we should be covered.

matteobettini:
i don't think this is currently the case as the `update_tensordict_priority` expects `TensorDictBase` since in its code it calls methods such as `ndim`, `get` and so on.

while `extend` is compatible with lists, if `extends` is called with a list that is not stackable as a lazy stack, it will fail when the data is passed to `update_tensordict_priority`

furthermore, if `extends` is called with a list, the `_data` and  `index` fileds are not set, so it is not clear where the `update_tensordict_priority` should be able to find them.

i can try to do some patch work to make `update_tensordict_priority` and `_get_priority` work vectorized, but i am very much confused by how this class is supposed to work and its contracts

vmoens:
If the method is broken with list it won't be bc-breaking if we don't support lists anymore, and any fix that enables the passing of lists to extend will be a "new feature" (since now it isn't a feature).
In other words: i'm ok with considering that everything passed to that method is a td of some sort

matteobettini:
Ok I'll do this in #1598

PR #1598, opened 2023-10-03
[BugFix] Vectorized priority update in replay buffers

matteobettini:
This is a patch for #1574

It fixes the core problem highlighted in that issue but there are still points that will need attention in a future refactoring of this class as I am not sure it is compatible with all the cases it aims to support. I am happy to elucidate more about this if needed.

### DIFF
--- a/torchrl/data/replay_buffers/replay_buffers.py
+++ b/torchrl/data/replay_buffers/replay_buffers.py
@@ -662,7 +662,7 @@ class TensorDictReplayBuffer(ReplayBuffer):
         super().__init__(**kw)
         self.priority_key = priority_key
 
-    def _get_priority(self, tensordict: TensorDictBase) -> Optional[torch.Tensor]:
+    def _get_priority_item(self, tensordict: TensorDictBase) -> float:
         if "_data" in tensordict.keys():
             tensordict = tensordict.get("_data")
 
@@ -680,6 +680,23 @@ class TensorDictReplayBuffer(ReplayBuffer):
                 f" {tensordict.get(self.priority_key).shape} but expected "
                 f"scalar value"
             )
+        return priority
+
+    def _get_priority_vector(self, tensordict: TensorDictBase) -> torch.Tensor:
+        if "_data" in tensordict.keys():
+            tensordict = tensordict.get("_data")
+
+        priority = tensordict.get(self.priority_key, None)
+        if priority is None:
+            return torch.tensor(
+                [self._sampler.default_priority],
+                dtype=torch.float,
+                device=tensordict.device,
+            ).expand(tensordict.shape[0])
+
+        priority = priority.view(priority.shape[0], -1)
+        priority = _reduce(priority, self._sampler.reduction, dim=1)
+
         return priority
 
     def add(self, data: TensorDictBase) -> int:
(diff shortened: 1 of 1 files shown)


torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -682,6 +682,23 @@ def _get_priority(self, tensordict: TensorDictBase) -> Optional[torch.Tensor]:
...
         return priority
 
+    def _get_priority_vector(self, tensordict: TensorDictBase) -> torch.Tensor:
+        if "_data" in tensordict.keys():
+            tensordict = tensordict.get("_data")
+
+        priority = tensordict.get(self.priority_key, None)
+        if priority is None:
+            return torch.tensor(
+                [self._sampler.default_priority],
```
vmoens, line 692:
```suggestion
                self._sampler.default_priority,
```

torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -757,13 +774,16 @@ def update_tensordict_priority(self, data: TensorDictBase) -> None:
...
-                [self._get_priority(td) for td in data],
-                dtype=torch.float,
-                device=data.device,
-            )
+            if isinstance(data, LazyStackedTensorDict):
+                priority = torch.tensor(
+                    [self._get_priority_item(td) for td in data],
+                    dtype=torch.float,
+                    device=data.device,
+                )
```
vmoens, line 782: I guess we assume that the stack dim is 0 but it could not be (?)
I think we can consider the priority can be stacked, no? At the end of the day it's supposed to be one priority per item. Maybe I'm missing something

torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -1010,16 +1030,16 @@ def __call__(self, list_of_tds):
...
+        return tensor.max().item() if dim is None else tensor.max(dim=dim)[0]
     elif reduction == "min":
-        return tensor.min().item()
+        return tensor.min().item() if dim is None else tensor.min(dim=dim)[0]
     elif reduction == "mean":
-        return tensor.mean().item()
+        return tensor.mean().item() if dim is None else tensor.mean(dim=dim)
     elif reduction == "median":
-        return tensor.median().item()
+        return tensor.median().item() if dim is None else tensor.median(dim=dim)
```
vmoens, line 1042: what about 
```python
    if reduction == "max":
        result = tensor.max(dim=dim)
    elif reduction == "min":
        result = tensor.min(dim=dim)
    elif reduction == "mean":
        result = tensor.mean(dim=dim)
    elif reduction == "median":
        result = tensor.median(dim=dim)
    else:
        raise NotImplementedError(f"Unknown reduction method {reduction}")
    if isinstance(result, tuple):
        result = result[0]
    return result
```

vmoens (approved):
LGTM, see my few minor comments

torchrl/data/replay_buffers/replay_buffers.py
matteobettini: yes that works too, but we also need to call `.item()`

torchrl/data/replay_buffers/replay_buffers.py
matteobettini: this was the previous treatment so I am not super sure what was going on.

my guess is that it is assuming 0 as stack dim because in `expand` it stacks on 0.

are you suggesting to completely remove this and always go vectorized? I am down to try.

matteobettini: maybe it was this way because the priority can have different shapes along the stack dim?

That is the only explanation I can guess

vmoens (approved):
LGTM thanks!

PR #1599, opened 2023-10-03
[Docs] Fix multi-agent tutorial

matteobettini:
Fix the multi-agent tutorial after the recent "termianted" introduction

### DIFF
--- a/tutorials/sphinx-tutorials/multiagent_ppo.py
+++ b/tutorials/sphinx-tutorials/multiagent_ppo.py
@@ -253,11 +253,10 @@ env = VmasEnv(
 #
 #
 
-print("action_spec:", env.action_spec)
-print("reward_spec:", env.reward_spec)
-print("done_spec:", env.done_spec)
+print("action_spec:", env.full_action_spec)
+print("reward_spec:", env.full_reward_spec)
+print("done_spec:", env.full_done_spec)
 print("observation_spec:", env.observation_spec)
-
 
 ######################################################################
 # Using the commands just shown we can access the domain of each value.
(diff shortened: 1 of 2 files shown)


tutorials/sphinx-tutorials/multiagent_ppo.py
```diff
@@ -649,11 +636,18 @@
...
-    )  # We need to expand the done to match the reward shape (this is expected by the value estimator)
+        .expand(tensordict_data.get_item_shape(("next", env.reward_key))),
+    )
+    tensordict_data.set(
+        ("next", "agents", "terminated"),
+        tensordict_data.get(("next", "terminated"))
+        .unsqueeze(-1)
+        .expand(tensordict_data.get_item_shape(("next", env.reward_key))),
+    )
+    # We need to expand the done and terminated to match the reward shape (this is expected by the value estimator)
```
vmoens, line 650: what is the best way to approach this?
A common criticism of our tutorials is that they are too complex, I wonder what will be judged as the most complex thing here: (1) boilerplate code such as this, which IMO has little instructive value or (2) an ad-hoc transform that may be a bit too custom to be part of the core lib

vmoens (approved):
LGTM, see my comment and feel free to merge if you feel it's the right solution
I don't really have a preference

tutorials/sphinx-tutorials/multiagent_ppo.py
vmoens: 4. seems like a stretch, these classes are already very complex and hard to maintain

I think hese operations are actually 2fold: (1) rename without deleting and (2) unsqueeze + expand
We have a transform that does rename, one that does unsqueeze. We just need one that does expand (or expand_as).

The quickest way forward is this:
Merge as is, open an issue with a feature request for the expand_as transform. Wdyt?

PR #1607, opened 2023-10-05
[BugFix] Patch SAC to allow state_dict manipulation before exec

### LINKED ISSUE #1594: [BUG] `SACLoss` cannot be loaded from `state_dict`
The SACLoss has a `target_entropy_buffer` which is instantiated not at init time but lazily https://github.com/pytorch/rl/blob/106368ffe627aada70ebb33ecf545d0f193c7757/torchrl/objectives/sac.py#L424

This is a problem as it will prevent reloading this loss from `state_dict`.

What will happen is that `state_dict` will likely contain this buffer and the reloaded loss will not (as the lazy step has not happened yet)

## Solution
We instantiate all buffers at init time. That seems the only way to me to support loading from `state_dict`

vmoens:
## Description

Closes #1594

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -3260,6 +3260,33 @@ class TestSAC(LossModuleTestBase):
         assert loss_actor == loss_val_td["loss_actor"]
         assert loss_alpha == loss_val_td["loss_alpha"]
 
+    def test_state_dict(self, version):
+        if version == 1:
+            pytest.skip("Test not implemented for version 1.")
+        model = torch.nn.Linear(3, 4)
+        actor_module = TensorDictModule(model, in_keys=["obs"], out_keys=["logits"])
+        policy = ProbabilisticActor(
+            module=actor_module,
+            in_keys=["logits"],
+            out_keys=["action"],
+            distribution_class=TanhDelta,
+        )
+        value = ValueOperator(module=model, in_keys=["obs"], out_keys="value")
+
+        loss = SACLoss(
+            actor_network=policy,
+            qvalue_network=value,
+            action_spec=UnboundedContinuousTensorSpec(shape=(2,)),
+        )
+        state = loss.state_dict()
+
+        loss = SACLoss(
+            actor_network=policy,
+            qvalue_network=value,
+            action_spec=UnboundedContinuousTensorSpec(shape=(2,)),
+        )
+        loss.load_state_dict(state)
+
 
 @pytest.mark.skipif(
     not _has_functorch, reason=f"functorch not installed: {FUNCTORCH_ERR}"
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -3287,6 +3287,22 @@ class TestSAC(LossModuleTestBase):
         )
         loss.load_state_dict(state)
 
+        # with an access in between
+        loss = SACLoss(
+            actor_network=policy,
+            qvalue_network=value,
+            action_spec=UnboundedContinuousTensorSpec(shape=(2,)),
+        )
+        loss.target_entropy
+        state = loss.state_dict()
+
+        loss = SACLoss(
+            actor_network=policy,
+            qvalue_network=value,
+            action_spec=UnboundedContinuousTensorSpec(shape=(2,)),
+        )
+        loss.load_state_dict(state)
+
 
 @pytest.mark.skipif(
     not _has_functorch, reason=f"functorch not installed: {FUNCTORCH_ERR}"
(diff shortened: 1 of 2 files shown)


test/test_cost.py
```diff
@@ -3260,6 +3260,33 @@ def test_sac_notensordict(
...
+            distribution_class=TanhDelta,
+        )
+        value = ValueOperator(module=model, in_keys=["obs"], out_keys="value")
+
+        loss = SACLoss(
+            actor_network=policy,
+            qvalue_network=value,
+            action_spec=UnboundedContinuousTensorSpec(shape=(2,)),
+        )
+        state = loss.state_dict()
```
matteobettini, line 3281: maybe let's add a forward call or an access to the entropy before saving

torchrl/objectives/sac.py
```diff
@@ -384,48 +393,54 @@ def __init__(
...
+            action_spec = (
+                action_spec
+                if action_spec is not None
+                else getattr(actor_network, "spec", None)
+            )
+            if action_spec is None:
+                raise RuntimeError(
+                    "Cannot infer the dimensionality of the action. Consider providing "
+                    "the target entropy explicitely or provide the spec of the "
+                    "action tensor in the actor network."
```
matteobettini, line 419: there is a tangential interesting bug related to this.

To see it try
```
from torchrl.modules import QValueActor, ProbabilisticActor, TanhDelta, ValueOperator

from torchrl.objectives import SACLoss

if __name__ == "__main__":
    model = torch.nn.Linear(3, 4)  # does not work
    # model = torch.nn.Linear(1, 1) works
    actor_module = TensorDictModule(
        torch.nn.Linear(3, 4), in_keys=["obs"], out_keys=["logits"]
    )
    policy = ProbabilisticActor(
        module=actor_module,
        in_keys=["logits"],
        out_keys=["action"],
        distribution_class=TanhDelta,
    )
    value = ValueOperator(module=model, in_keys=["obs"], out_keys="value")

    loss = SACLoss(
        actor_network=policy,
        qvalue_network=value,
        # action_spec=UnboundedContinuousTensorSpec(shape=(2,)), not passing the spec
    )
    _ = loss.target_entropy
```
it will trigger this error but the user will see

```
File "/Users/matbet/PycharmProjects/rl/prova.py", line 29, in <module>
    loss.target_entropy
  File "/Users/matbet/PycharmProjects/rl/torchrl/objectives/common.py", line 348, in __getattr__
    return super().__getattr__(item)
  File "/Users/matbet/miniconda3/envs/torchrl/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1614, in __getattr__
    raise AttributeError("'{}' object has no attribute '{}'".format(
AttributeError: 'SACLoss' object has no attribute 'target_entropy'
```

so the message here is never shown

this is valid for all @property methods. isn't this curious?

matteobettini (approved):
LGTM thanks a lot

torchrl/objectives/sac.py
```diff
@@ -384,48 +393,54 @@ def __init__(
...
                 )
-            self.register_buffer(
-                "target_entropy_buffer", torch.tensor(target_entropy, device=device)
+            if not isinstance(action_spec, CompositeSpec):
+                action_spec = CompositeSpec({self.tensor_keys.action: action_spec})
+            if (
+                isinstance(self.tensor_keys.action, tuple)
+                and len(self.tensor_keys.action) > 1
+            ):
+                print(action_spec)
```
matteobettini, line 427:
```suggestion
             
```

test/test_cost.py
vmoens: then we're sure it is instantiated and the test loses its value no?

matteobettini:
the example in #1594 still fails, to test that you can either use that script or do a forward pass before saving the loss

torchrl/objectives/sac.py
vmoens: The `__getattr__` of torch.nn.Module isn't its greatest feature!
Let me think of a fix...

matteobettini: the same extends also to `EnvBase` and other components, so the scope of this might be outside this PR

test/test_cost.py
matteobettini: what we want to test is when the buffer is instantiated in the saved loss and it has to be loaded in a new loss which has just been init

vmoens:
Good catch i had forgotten an "_" in the delezify

matteobettini:
good to go for me, let's maybe add the example in https://github.com/pytorch/rl/issues/1594 in the tests

vmoens:
already done :)

PR #1611, opened 2023-10-06
[Feature] step_and_maybe_reset in env

vmoens:
## Contribution

This PR proposes the `step_and_maybe_reset` in EnvBase. 
This method executes a step followed by a reset, if necessary.

We also make `reset` more robust by ensuring that partial resets are handled uniformly. This is necessary since `step_and_maybe_reset` must take care of this functionality, and from our perspective handling partial resets is the responsibility of `reset` (the user should not have to worry about data not updated properly).

This has repercussions on the logic behind `TransformedEnv._reset` and `BatchedEnv._reset`.

I'm now considering having batched envs calling `reset` and not `_reset` to make sure that the data is well presented, since now the update of the input tensordict with the `tensordict_reset` occurs after `_reset` (hence, the output of `_reset` in SerialEnv is incomplete).
This could introduce some overhead but that's of limited impact since now `step_and_maybe_reset` is there to handle things faster.

### DIFF
--- a/docs/source/reference/envs.rst
+++ b/docs/source/reference/envs.rst
@@ -58,6 +58,23 @@ With these, the following methods are implemented:
 - :meth:`env.step`: a step method that takes a :class:`tensordict.TensorDict` input
   containing an input action as well as other inputs (for model-based or stateless
   environments, for instance).
+- :meth:`env.step_and_maybe_reset`: executes a step, and (partially) resets the
+  environments if it needs to. It returns the updated input with a ``"next"``
+  key containing the data of the next step, as well as a tensordict containing
+  the input data for the next step (ie, reset or result or
+  :func:`~torchrl.envs.utils.step_mdp`)
+  This is done by reading the ``done_keys`` and
+  assigning a ``"_reset"`` signal to each done state. This method allows
+  to code non-stopping rollout functions with little effort:
+
+    >>> data_ = env.reset()
+    >>> result = []
+    >>> for i in range(N):
+    ...     data, data_ = env.step_and_maybe_reset(data_)
+    ...     result.append(data)
+    ...
+    >>> result = torch.stack(result)
+
 - :meth:`env.set_seed`: a seeding method that will return the next seed
   to be used in a multi-env setting. This next seed is deterministically computed
   from the preceding one, such that one can seed multiple environments with a different
@@ -169,7 +186,37 @@ one can simply call:
         >>> print(a)
         9.81
 
-It is also possible to reset some but not all of the environments:
+TorchRL uses a private ``"_reset"`` key to indicate to the environment which
+sub-environments should be reset.
+This allows to reset some but not all of the environments.
+Keep in mind that this is a private key, and it should only be used when coding
+specific environment features that are internal facing.
+The way an environment deals with the ``"_reset"`` keys is proper to it and
+designing an environment that behaves according to ``"_reset"`` inputs is the
+developer's responsibility. The following assumptions are made:
+
+- Each ``"_reset"`` is paired with a ``"done"`` (and/or ``"terminated"``,
+  ``"truncated"`` etc.). This means that the following structure is not
+  allowed: ``TensorDict({"done": done, "nested": {"_reset": reset}}, [])``.
+- A reset at one level precludes the presence of a ``"_reset"`` at lower
+  levels below it. For instance, this is not allowed:
+  ``TensorDict({"_reset": reset_main, "nested": {"_reset": reset_leaf}}, [])``
+- When calling :meth:`env.reset(tensordict)` with a partial ``"_reset"`` entry
+  that will reset some but not all the done sub-environments, the input data
+  should contain the data of the sub-environemtns that are __not__ being reset.
+  The reason for this constrain is that the output of the ``env._reset(data)``
+  can only be predicted for the entries that are reset. For the others, we
+  cannot know in advance if they will be meaningful or not. For instance,
+  one could perfectly just pad the values of the non-reset environments, in
+  which case the data will be meaningless and should be discarded.
+  :meth:`~torchrl.envs.EnvBase.reset` will merge the input tensordict with
+  the one resulting from reset and pad missing values with 0 (except for
+  ``done`` signals).
+  The other reason why this constrain is enforced is that usually,
+  setting ``"_reset"`` is achieved by internal methods that have access to
+  the data at the previous step, and use that data to call ``reset`` (after
+  passing it through :func:`~torchrl.envs.utils.step_mdp`). We can safely
+  assume that the data presented to ``reset`` is therefore complete.
 
 .. code-block::
    :caption: Parallel environment reset
--- a/docs/source/reference/envs.rst
+++ b/docs/source/reference/envs.rst
@@ -187,36 +187,94 @@ one can simply call:
         9.81
 
 TorchRL uses a private ``"_reset"`` key to indicate to the environment which
-sub-environments should be reset.
-This allows to reset some but not all of the environments.
-Keep in mind that this is a private key, and it should only be used when coding
-specific environment features that are internal facing.
-The way an environment deals with the ``"_reset"`` keys is proper to it and
-designing an environment that behaves according to ``"_reset"`` inputs is the
-developer's responsibility. The following assumptions are made:
+component (sub-environments or agents) should be reset.
+This allows to reset some but not all of the components.
 
-- Each ``"_reset"`` is paired with a ``"done"`` (and/or ``"terminated"``,
-  ``"truncated"`` etc.). This means that the following structure is not
-  allowed: ``TensorDict({"done": done, "nested": {"_reset": reset}}, [])``.
-- A reset at one level precludes the presence of a ``"_reset"`` at lower
-  levels below it. For instance, this is not allowed:
-  ``TensorDict({"_reset": reset_main, "nested": {"_reset": reset_leaf}}, [])``
+The ``"_reset"`` key has two distinct functionalities:
+1. During a call to :meth:`~.EnvBase._reset`, the ``"_reset"`` key may or may
+   not be present in the input tensordict. TorchRL's convention is that the
+   absence of the ``"_reset"`` key at a given ``"done"`` level indicates
+   a total reset of that level (unless a ``"_reset"`` key was found at a level
+   above, see details below).
+   If it is present, it is expected that those entries and only those components
+   where the ``"_reset"`` entry is ``True`` (along key and shape dimension) will be reset.
+
+   The way an environment deals with the ``"_reset"`` keys in its :meth:`~.EnvBase._reset`
+   method is proper to its class.
+   Designing an environment that behaves according to ``"_reset"`` inputs is the
+   developer's responsibility, as TorchRL has no control over the inner logic
+   of :meth:`~.EnvBase._reset`. Nevertheless, the following point should be
+   kept in mind when desiging that method.
+
+2. After a call to :meth:`~.EnvBase._reset`, the output will be masked with the
+   ``"_reset"`` entries and the output of the previous :meth:`~.EnvBase.step`
+   will be written wherever the ``"_reset"`` was ``False``. In practice, this
+   means that if a ``"_reset"`` modifies data that isn't exposed by it, this
+   modification will be lost. After this masking operation, the ``"_reset"``
+   entries will be erased from the :meth:`~.EnvBase.reset` outputs.
+
+It must be pointed that ``"_reset"`` is a private key, and it should only be
+used when coding specific environment features that are internal facing.
+In other words, this should NOT be used outside of the library, and developers
+will keep the right to modify the logic of partial resets through ``"_reset"``
+setting without preliminary warranty, as long as they don't affect TorchRL
+internal tests.
+
+Finally, the following assumptions are made and should be kept in mind when
+designing reset functionalities:
+
+- Each ``"_reset"`` is paired with a ``"done"`` entry (+ ``"terminated"`` and,
+  possibly, ``"truncated"``). This means that the following structure is not
+  allowed: ``TensorDict({"done": done, "nested": {"_reset": reset}}, [])``, as
+  the ``"_reset"`` lives at a different nesting level than the ``"done"``.
+- A reset at one level does not preclude the presence of a ``"_reset"`` at lower
+  levels, but it annihilates its effects. The reason is simply that
+  whether the ``"_reset"`` at the root level corresponds to an ``all()``, ``any()``
+  or custom call to the nested ``"done"`` entries cannot be known in advance,
+  and it is explicitly assumed that the ``"_reset"`` at the root was placed
+  there to superseed the nested values (for an example, have a look at
+  :class:`~.PettingZooWrapper` implementation where each group has one or more
+  ``"done"`` entries associated which is aggregated at the root level with a
+  ``any`` or ``all`` logic depending on the task).
 - When calling :meth:`env.reset(tensordict)` with a partial ``"_reset"`` entry
   that will reset some but not all the done sub-environments, the input data
   should contain the data of the sub-environemtns that are __not__ being reset.
-  The reason for this constrain is that the output of the ``env._reset(data)``
-  can only be predicted for the entries that are reset. For the others, we
-  cannot know in advance if they will be meaningful or not. For instance,
-  one could perfectly just pad the values of the non-reset environments, in
-  which case the data will be meaningless and should be discarded.
-  :meth:`~torchrl.envs.EnvBase.reset` will merge the input tensordict with
-  the one resulting from reset and pad missing values with 0 (except for
-  ``done`` signals).
-  The other reason why this constrain is enforced is that usually,
-  setting ``"_reset"`` is achieved by internal methods that have access to
-  the data at the previous step, and use that data to call ``reset`` (after
-  passing it through :func:`~torchrl.envs.utils.step_mdp`). We can safely
-  assume that the data presented to ``reset`` is therefore complete.
+  The reason for this constrain lies in the fact that the output of the
+  ``env._reset(data)`` can only be predicted for the entries that are reset.
+  For the others, TorchRL cannot know in advance if they will be meaningful or
+  not. For instance, one could perfectly just pad the values of the non-reset
+  components, in which case the non-reset data will be meaningless and should
+  be discarded.
+
+Below, we give some examples of the expected effect that ``"_reset"`` keys will
+have on an environment returning zeros after reset:
+
+    >>> # single reset at the root
+    >>> data = TensorDict({"val": [1, 1], "_reset": [False, True]}, [])
+    >>> env.reset(data)
+    >>> print(data.get("val"))  # only the second value is 0
+    tensor([1, 0])
+    >>> # nested resets
+    >>> data = TensorDict({
+    ...     ("agent0", "val"): [1, 1], ("agent0", "_reset"): [False, True],
+    ...     ("agent1", "val"): [2, 2], ("agent1", "_reset"): [True, False],
+    ... }, [])
+    >>> env.reset(data)
+    >>> print(data.get(("agent0", "val")))  # only the second value is 0
+    tensor([1, 0])
+    >>> print(data.get(("agent1", "val")))  # only the second value is 0
+    tensor([0, 2])
+    >>> # nested resets are overridden by a "_reset" at the root
+    >>> data = TensorDict({
+    ...     "_reset": [True, True],
+    ...     ("agent0", "val"): [1, 1], ("agent0", "_reset"): [False, True],
+    ...     ("agent1", "val"): [2, 2], ("agent1", "_reset"): [True, False],
+    ... }, [])
+    >>> env.reset(data)
+    >>> print(data.get(("agent0", "val")))  # reset at the root overrides nested
+    tensor([0, 0])
+    >>> print(data.get(("agent1", "val")))  # reset at the root overrides nested
+    tensor([0, 0])
 
 .. code-block::
    :caption: Parallel environment reset
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -1590,7 +1590,7 @@ class TestVmas:
             [n_workers, list(env.num_envs)[0], n_rollout_samples]
         )
 
-    @pytest.mark.parametrize("num_envs", [1, 10])
+    @pytest.mark.parametrize("num_envs", [1, 2])
     @pytest.mark.parametrize("n_workers", [1, 3])
     @pytest.mark.parametrize(
         "scenario_name", ["simple_reference", "waterfall", "flocking", "discovery"]
@@ -1631,6 +1631,9 @@ class TestVmas:
         td_reset = TensorDict(
             rand_reset(env), batch_size=env.batch_size, device=env.device
         )
+        # it is good practice to have a "complete" input tensordict for reset
+        for done_key in env.done_keys:
+            td_reset.set(done_key, tensordict[..., -1].get(("next", done_key)))
         reset = td_reset["_reset"]
         tensordict = env.reset(td_reset)
         assert not tensordict["done"][reset].all().item()
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -1889,12 +1889,9 @@ class TestD4RL:
         env = GymWrapper(gym.make(task))
         rollout = env.rollout(2)
         for key in rollout.keys(True, True):
-            if "truncated" in key:
-                # truncated is missing from static datasets
-                continue
-            sim = rollout[key]
-            offline = sample[key]
-            assert sim.dtype == offline.dtype, key
+            sim = rollout.get(key)
+            offline = sample.get(key)
+            # assert sim.dtype == offline.dtype, key
             assert sim.shape[-1] == offline.shape[-1], key
         print(f"terminated test after {time.time()-t0}s")
 
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -1810,6 +1810,11 @@ class TestD4RL:
                 data_true._storage._storage.shape
                 == data_from_env._storage._storage.shape
             )
+            # for some reason, qlearning_dataset overwrites the next obs that is contained in the buffer,
+            # resulting in tiny changes in the value contained for that key. Over 99.99% of the values
+            # match, but the test still fails because of this.
+            # We exclude that entry from the comparison.
+            keys.discard(("_data", "next", "observation"))
             assert_allclose_td(
                 data_true._storage._storage.select(*keys),
                 data_from_env._storage._storage.select(*keys),
@@ -1889,6 +1894,9 @@ class TestD4RL:
         env = GymWrapper(gym.make(task))
         rollout = env.rollout(2)
         for key in rollout.keys(True, True):
+            if "truncated" in key:
+                # truncated is missing from static datasets
+                continue
             sim = rollout.get(key)
             offline = sample.get(key)
             # assert sim.dtype == offline.dtype, key
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -44,7 +44,7 @@ from torchrl.data.utils import CloudpickleWrapper, DEVICE_TYPING
 from torchrl.envs.common import EnvBase
 from torchrl.envs.transforms import StepCounter, TransformedEnv
 from torchrl.envs.utils import (
-    _aggregate_resets,
+    _aggregate_end_of_traj,
     _convert_exploration_type,
     ExplorationType,
     set_exploration_type,
@@ -786,7 +786,7 @@ class SyncDataCollector(DataCollectorBase):
 
     def _update_traj_ids(self, tensordict) -> None:
         # we can't use the reset keys because they're gone
-        traj_sop = _aggregate_resets(
+        traj_sop = _aggregate_end_of_traj(
             tensordict.get("next"), done_keys=self.env.done_keys
         )
         if traj_sop.any():
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -873,14 +873,14 @@ class SyncDataCollector(DataCollectorBase):
             # check that the env supports partial reset
             if prod(self.env.batch_size) == 0:
                 raise RuntimeError("resetting unique env with index is not permitted.")
-            _reset = torch.zeros(
-                self.env.done_spec.shape,
-                dtype=torch.bool,
-                device=self.env.device,
-            )
-            _reset[index] = 1
-            self._tensordict[index].zero_()
-            self._tensordict.set("_reset", _reset)
+            for reset_key, done_keys in zip(self.env.reset_keys, self.env.done_keys_groups):
+                _reset = torch.zeros(
+                    self.env.full_done_spec[done_keys[0]].shape,
+                    dtype=torch.bool,
+                    device=self.env.device,
+                )
+                _reset[index] = 1
+                self._tensordict.set(reset_key, _reset)
         else:
             _reset = None
             self._tensordict.zero_()
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -873,7 +873,9 @@ class SyncDataCollector(DataCollectorBase):
             # check that the env supports partial reset
             if prod(self.env.batch_size) == 0:
                 raise RuntimeError("resetting unique env with index is not permitted.")
-            for reset_key, done_keys in zip(self.env.reset_keys, self.env.done_keys_groups):
+            for reset_key, done_keys in zip(
+                self.env.reset_keys, self.env.done_keys_groups
+            ):
                 _reset = torch.zeros(
                     self.env.full_done_spec[done_keys[0]].shape,
                     dtype=torch.bool,
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -493,10 +493,14 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
 
         Keys are sorted by depth in the data tree.
         """
+        action_keys = self.__dict__.get("_action_keys", None)
+        if action_keys is not None:
+            return action_keys
         keys = self.input_spec["full_action_spec"].keys(True, True)
         if not len(keys):
             raise AttributeError("Could not find action spec")
         keys = sorted(keys, key=_repr_by_depth)
+        self.__dict__["_action_keys"] = keys
         return keys
 
     @property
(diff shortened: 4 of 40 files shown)


docs/source/reference/envs.rst
```diff
@@ -169,7 +186,37 @@ one can simply call:
...
+  one could perfectly just pad the values of the non-reset environments, in
+  which case the data will be meaningless and should be discarded.
+  :meth:`~torchrl.envs.EnvBase.reset` will merge the input tensordict with
+  the one resulting from reset and pad missing values with 0 (except for
+  ``done`` signals).
+  The other reason why this constrain is enforced is that usually,
+  setting ``"_reset"`` is achieved by internal methods that have access to
+  the data at the previous step, and use that data to call ``reset`` (after
+  passing it through :func:`~torchrl.envs.utils.step_mdp`). We can safely
+  assume that the data presented to ``reset`` is therefore complete.
```
matteobettini, line 219: Why do we have to enforce this? In my view we can:

- if no td is passed as input to `reset()` just return what you get from `_reset()`
- if a td is passed as input with partial keys (e.g. maybe just the _reset keys), I would not want to have the output padded by default, but i would want to update the keys from the input using the ones from `_reset()` and, when the input keys are missing, directly copy the ones from `_reset`

docs/source/reference/envs.rst
```diff
@@ -169,7 +186,37 @@ one can simply call:
...
+
+- Each ``"_reset"`` is paired with a ``"done"`` (and/or ``"terminated"``,
+  ``"truncated"`` etc.). This means that the following structure is not
+  allowed: ``TensorDict({"done": done, "nested": {"_reset": reset}}, [])``.
+- A reset at one level precludes the presence of a ``"_reset"`` at lower
+  levels below it. For instance, this is not allowed:
+  ``TensorDict({"_reset": reset_main, "nested": {"_reset": reset_leaf}}, [])``
+- When calling :meth:`env.reset(tensordict)` with a partial ``"_reset"`` entry
+  that will reset some but not all the done sub-environments, the input data
+  should contain the data of the sub-environemtns that are __not__ being reset.
```
matteobettini, line 206: How can the input contain only the data "of the sub-environments that are __not__ being reset." It has to contain all data no? otherwise it is sparse

docs/source/reference/envs.rst
```diff
@@ -169,7 +186,37 @@ one can simply call:
...
+  The reason for this constrain is that the output of the ``env._reset(data)``
+  can only be predicted for the entries that are reset. For the others, we
+  cannot know in advance if they will be meaningful or not. For instance,
+  one could perfectly just pad the values of the non-reset environments, in
+  which case the data will be meaningless and should be discarded.
+  :meth:`~torchrl.envs.EnvBase.reset` will merge the input tensordict with
+  the one resulting from reset and pad missing values with 0 (except for
+  ``done`` signals).
```
matteobettini, line 214: I don't see why if we cannot predict if they will be meaningful we automatically pad with 0s

It seems like the options are:
- pad with the values from `env._reset()` which could be meaningful or not
- pad with 0 which is certainly not meaningful

This is a very simple choice for me. What am I missing?
My suggestion is to remove any 0 padding and simlply use data from `env._reset()` whenever we would have padded

docs/source/reference/envs.rst
```diff
@@ -169,7 +186,37 @@ one can simply call:
         >>> print(a)
         9.81
 
-It is also possible to reset some but not all of the environments:
+TorchRL uses a private ``"_reset"`` key to indicate to the environment which
```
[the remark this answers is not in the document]
vmoens: yes, fair point. In the context of batched envs it seemed clearer to refer to it as envs
we could say "envs, tasks or agent"

docs/source/reference/envs.rst
vmoens: > if no td is passed as input to reset() just return what you get from _reset()

this is what happens

> if a td is passed as input with partial keys (e.g. maybe just the _reset keys), I would not want to have the output padded by default, but i would want to update the keys from the input using the ones from _reset() and, when the input keys are missing, directly copy the ones from _reset

that was the case so far but it's very tricky and brittle. Let me explain why:
- First I think we will all agree that we need a rule that applies to all envs, regardless of whether they are stateful, stateless, simple or nested / MARL. The rule must be transparent and always the same (in envs, collectors etc).
- Say we pass an input with just a `"_reset"` entry. We get an output from _reset. __Assuming__ the environment is stateful, it could make sense to say that all the data is valid. But that's only assuming the env is stateful __and__ that the users comply with the rule stated above (for which we have no guarantee). If it's stateless, we have no idea what the other data means. Also one could perfectly do
```python
def _reset(td):
    obs = torch.zeros(size)
    obs[td.get("_reset")] = some_obs_reset
    return TensorDict({"obs": obs}, [])
```
and IMO this should be valid (again, regardless of whether the env is stateful or stateless). However this will eventually break if we consider everything that comes out of _reset as valid.
- In batched envs, we have access to the buffers. We'll copy the entire buffer after a call to `_reset`, also for sub-envs that have not been reset. Then we make the wishful guess that the data of non-reset sub-envs is consistent with the data at the previous step, but we have no way to actually make sure that it's the case (eg, stateless envs and similar). These buffers should definitely __not__ be used to complete the data. 

In summary, choosing to "sometimes" use the data returned by the env has multiple ramifications and will cause a lot of headache long-term as it does not apply across all envs, and makes lots of unwated assumptions about internal machinery of the envs.  `EnvBase` responsibility is to provide a common working framework for environments given a set of rules. I feel that choosing to sometimes use or not the non-reset data coming from a reset is too opinionated for EnvBase to fullfill its purpose.

docs/source/reference/envs.rst
vmoens: who said only?

docs/source/reference/envs.rst
vmoens: That does not work unfortunately. 
We have no idea about what is going on within `_reset` in general, and per se the only thing we want to ask is that `"_reset"` points to what is being changed. For the rest we assume that the data is unchanged compared to the last call to `step`. We have access to that data so we can copy it. With stateless environment, not padding isn't really an option I think.

docs/source/reference/envs.rst
matteobettini: but if we specify that it should contain the data for those that are not being reset it might be read as implying that it contains only that. I think it is easier if we say it contains all data

docs/source/reference/envs.rst
vmoens: For instance, [here](https://github.com/pytorch/rl/blob/38dfc21758f57683071fc29114c338d12f4b8d31/torchrl/envs/batched_envs.py#L624-L640) we assume that the non-reset data is identical to the content of the buffer, but that's an implicit assumption and a hard one to enforce/test. 
It also introduces some unwanted overhead since we will be calling where in the collector (not anymore with this PR since we do the job just once).
What we're suggesting with this PR is to do the call of `torch.where` in a consistent way, just once after calling `reset`

docs/source/reference/envs.rst
vmoens: if we want to be precise, we don't really care that the data of the envs being reset is there actually. As long as `torch.where` will work.
I don't really see how it can contain one and not the other (without recurring to lazy stacks or such -- and if it does contain lazy stacks with just the non-reset data it should still work).

vmoens: Example:
```python
import torch
reset = torch.tensor([True, False])
td = torch.stack([
    TensorDict({}, []),
    TensorDict({"a": 1}, []),
])
td_reset = torch.stack([
    TensorDict({"a": 0}, []),
    TensorDict({}, []),
])
td_reset.where(~reset, other=td, pad=0)
```
which returns a non-sparse td
```
LazyStackedTensorDict(
    fields={
        a: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.int64, is_shared=False)},
    exclusive_fields={
    },
    batch_size=torch.Size([2]),
    device=None,
    is_shared=False,
    stack_dim=0)
```

docs/source/reference/envs.rst
vmoens: Another thought:
If we say "anything returned along the non-reset axis will be considered as valid data", we're not telling the truth since the collector currently discards that data in favour of the latest "next" data using `torch.where`. So either we say the opposite ("anything returned along the non-reset axis will be considered as **in**valid data") or we say "sometimes one, sometimes the other"

docs/source/reference/envs.rst
matteobettini: I think you misunderstood my point.

> and IMO this should be valid (again, regardless of whether the env is stateful or stateless). However this will eventually break if we consider everything that comes out of _reset as valid.

I agree that this should be valid and I am _not_ proposing to consider everything that comes out of _reset as valid.

My point is about corner cases where reset is called with a partially complete input.
From the docstring it seems like this corner case is treated with padding of 0s.

My proposition is to treat it with data from _reset() instead of 0s, which are always guaranteed to be invalid.

Another option is to throw an error in the cases where we would pad

matteobettini: in other words, the merge described in the docstring could happen without any padding, just creating a new tensordict from the 2 with the following rules

- when _reset is true use output of _reset()
- when _reset is false use data from  reset() input
- in case  _reset is false and there is no data from  reset() input, use data from _reset() (instead of 0s)

vmoens: I understood this point thanks.
That introduces inconsistencies between collectors (which call where and copy the previous "next" data) and calls to reset. We need a one fits all solution, and that doesn't make the cut.

docs/source/reference/envs.rst
matteobettini: > "anything returned along the non-reset axis will be considered as valid data"

I am not saying this.

I am saying that in the extreme case where `env.reset()` would pad, we fallback to it.

See my comments in the other thread.

docs/source/reference/envs.rst
vmoens: Let's imagine the following scenario with this: 
someone builds an app where they call `reset(data)` and use the non-reset data out of it.
Next they use the env in the collector and things break. Apparently there is not reason for it to fail since the collector should do the exact same thing.
We need to be super explicit regarding the non-safety of using data from reset that is not matched by a `"_reset"` entry. Allowing data to be passed is dangerous I think

matteobettini: > That introduces inconsistencies between collectors (which call where and copy the previous "next" data) and calls to reset. We need a one fits all solution, and that doesn't make the cut.

So collectors pad with zeros?

I don't think this can be compared with collectors as this edge case I am talking about cannot happen within them as they always present complete data.

My only point here is that I don't see a case in which padding with zeros in `env.reset()` is beneficial to the user.
I was just trying to propose a solution to this but any solution works for me.

What I want to avoid is some 0s silently added to data in core environment calls

vmoens: > So collectors pad with zeros?

No they use where but they override the previous values. The point of padding is to "invalidate" data.

> My only point here is that I don't see a case in which padding with zeros in env.reset() is beneficial to the user.

It avoids letting people thing that data can be used if it's not reset, that's all. It's important because otherwise you can have silent bugs when using that data + inconsistencies if you're using it between collector and pure env execution. It's just a way of making the env robust and have a one-fits-all solution. 

The option of not padding is kind of allowing the data to be used, but if you swap your gym with an brax env in a parallel / serial execution, you may have polluted data when calling reset in the brax case and not in gym. To prevent that, we treat everything the same way. 

What's the case against padding precisely?

matteobettini: I am not sure padding is a non-silent way to invalidate data.

In my opinion it will silently add arbitrary data and return it as if it was valid.

How can users know if padding happens?

> The option of not padding is kind of allowing the data to be used, but if you swap your gym with an brax env in a parallel / serial execution, you may have polluted data when calling reset in the brax case and not in gym. To prevent that, we treat everything the same way.

If treating everything the same way is padding, you are still polluting the data.

I would consider other alternatives, maybe error?

vmoens: > How can users know if padding happens?

They shouldn't care, the data is not valid. Why should they "know" if we pad, since sometimes the data will be rubbish anyway. The crucial point here is that we provide an API to code envs but we have no way of telling what happens within `_reset`, so we must be very transparent regarding what comes out of it for `"_reset"`s that are `False`. I think that "sometimes this work because my env is stateful and obbeys by gym standards" or anything like that is a no go: it makes the environments non-interchangeable.

This is what I'm trying to solve here: if we don't pad, we're treating different envs differently without consistency. Example:

```python
env = SerialEnv(2, lambda: GymEnv("Pendulum-v1"))
...
env.reset(TensorDict({"_reset": [True, False]}, [2])) # all data is valid
```

```python
env = BraxEnv(..., batch_size=[2])
...
env.reset(TensorDict({"_reset": [True, False]}, [2])) # we don't know anything about the second batch of data
```

Let me revert the question: how can the user know in which case the data can be used or not? From the outside, there is no way to tell, no attribute in env that can point you in that direction.

Add to that the inconsistency between this and collectors and we end up with a big stack of methods to call reset and return data that differ from each other and will be hard to track in the future. By having a clear pact (_reset = True means valid, _reset = False means 0) we are consistent.

I don't think we want an error either because one could have an env that takes care of returning the fully fleshed data after partial `_reset`.


Let's not forget a very important piece here: using `"_reset"` is reserved to TorchRL's components by definition. Anyone passing a tensordict with `"_reset"` in their code base (out of `pytorch/rl`) is doing it at their own risk. So the "user" shouldn't care about this. Developers should, but we document things in the doc so it's fine (they can know what is padded through the inverse of the `"_reset"` mask to answer the initial question).
This change is therefore not bc-breaking and safe. This is not public facing. And it will truly makes maintenance easier.

docs/source/reference/envs.rst
vmoens: again, i understand that, but that's the same as saying that we consider that data as "valid".
Let me give you an example: in this PR, i discovered many places where we do a partial reset and check that the non-reset data is the same as the data in the previous next. This is assuming that the env is stateful, and as such it's a bad decision (a bad thing to test). We should not assume that at any place in the code. Padding with 0s makes sure that if we assume that, we're wrong, and not "sometimes right", which is actually the key to prevent silent bugs.

vmoens: If reading data from non-reset entries is not supported, we should lock that feature. 
For instance: if you call `.backward()` on a value, you don't expect tensors that are out of the graph to have non-zero gradients assigned to them. If you are using a MaskedTensor, you don't expect any operation to produce meaningful values for masked entries. This is the same logic here, we lock this logic by padding (or better: masking) with 0s anything that is meaningless. it's a pretty straightforward contract I think

docs/source/reference/envs.rst
matteobettini: > I don't think we want an error either because one could have an env that takes care of returning the fully fleshed data after partial _reset.

this is surprising since I thought this PR was here to forbid this. How would one be able to do this if we discard all data from `_reset()` that is not `_reset`?

```python
env = BraxEnv(..., batch_size=[2])
...
env.reset(TensorDict({"_reset": [True, False]}, [2])) # we don't know anything about the second batch of data
```
I still do not super understand why, if we don't know anything about the second batch of data, do we want to force 0s there. What is a user that is served with that output supposed to do? other data might be tangentially zero so even checking what is 0 would not tell what is valid. Plus `reset()` will remove the `_reset` flags, so there is really no way to understand whats valid.

The risk here is that all that data will be passed to another component and the 0s interpreted as real data.

Do you see what I am referring to? Something like
```python
policy(env.reset(TensorDict({"_reset": [True, False]}, [2])))
```
will silently use the zeros

Is there no solution where, whatever i pass to reset(), all its output will be all valid? 
maybe we can say that _reset() should return all valid output even with partial resets as a convention and then we are sure brax will do that

docs/source/reference/envs.rst
vmoens: Here's another example:
@albertbou92 recently uncovered a bug where the following code fails
```python
torch.manual_seed(0)
env = ParallelEnv(2, lambda: TransformedEnv(GymEnv("CartPole-v1"), CatFrames(in_keys=["observation"], out_keys=["observation_cat"], N=4, dim=-1)))
env.set_seed(0)
r = env.rollout(1000, break_when_any_done=False)

torch.manual_seed(0)
env = TransformedEnv(ParallelEnv(2, lambda: GymEnv("CartPole-v1")), CatFrames(in_keys=["observation"], out_keys=["observation_cat"], N=4, dim=-1))
env.set_seed(0)
r2 = env.rollout(1000, break_when_any_done=False)

torch.testing.assert_close(r["observation_cat"], r2["observation_cat"])
```

This is caused by `CatFrames` having a buffer stored as `nn.Module._buffer` instance where we write data as it flows. But when we call a partial `reset`, we also call `transform._call` which applies the transform to all dimensions (not only the part of the data that is being reset). That means that we corrupt data for envs not being reset.
The solution I'm contemplating is to carry the buffer in the tensordict and not in the transform. 
That way, when we output the result of `CatFrames._reset`, we get some corrupted data that we can override with the data from the prev step. 
I think that presenting non-masked data that is corrupted is wrong n this case. It will make it harder to test that the class its doing its job (with the solution I propose we can just check that maskes data is 0 and that the rest respects the convention of the CatFrame transform).
It's a very specific example but I think there are many others where the call to `transform._call` within the transform class `_reset` method will corrupt data in a way that we want to get rid of

docs/source/reference/envs.rst
vmoens: > this is surprising since I thought this PR was here to forbid this. How would one be able to do this if we discard all data from _reset() that is not "_reset"?

fully fleshed doesn't mean valid, it just means contiguous (possibly masked or meaningless) data.
For internal usage it can be useful to call `env.reset` with a partial "_reset" without passing the full data as input, and since it's internal facing there's no need for an exception.

> What is a user that is served with that output supposed to do? 

the user never ever sees that, as stated earlier.

> other data might be tangentially zero so even checking what is 0 would not tell what is valid.

the developer has access to the `"_reset"` entry which masks the data, so one can very quickly know what is valid.

> Plus reset() will remove the _reset flags, so there is really no way to understand whats valid.

One still has the `"done"` from before and I think in all cases can backtrack what was being reset. 

```python
policy(env.reset(TensorDict({"_reset": [True, False]}, [2])))
```
that will never happen because `"_reset"` is just used within TorchRL. If  users are using it, they're not doing themselves a favour. `_` means private and should be considered as such. We don't have a single call to a policy like this in the code base I believe. And in the case of many envs it will still be very wrong to do it because the data of the second batch can or cannot be corrupted and there's not way to tell. 

 > Is there no solution where, whatever i pass to reset(), all its output will be all valid?

that can happen but since it cannot be always the case we cannot assume it and we must prevent that sort of usage at all cost!

docs/source/reference/envs.rst
matteobettini: > in this PR, i discovered many places where we do a partial reset and check that the non-reset data is the same as the data in the previous next. This is assuming that the env is stateful, and as such it's a bad decision (a bad thing to test)

This to me is the most important test regarding resets.
Sure, if we are using this in non-stateful enviornments, this test has to be adapted, but for stateful ones to me this is a core thing to check.

docs/source/reference/envs.rst
vmoens: To rephrase: even without this PR, 
```python
policy(env.reset(TensorDict({"_reset": [True, False]}, [2])))
```
is wrong and should not be used (or at least not expecting that the policy is doing anything meaningful with index `1`)

vmoens:
@matteobettini Just to make a TLDR and hopefully close this discussion:
1. `"_reset"` is a private feature of TorchRL and should not be used outside the library. We document its effects and test them, but not expose it. We explain what it does in the doc for the sole purpose of allowing external collaborators to implement new envs within the library.
2. The padding will only occur in tests, never in real execution. The reason is that a call with `"_reset"` will always follow a `step` because only the `step` can tell what to reset. Other usage of `"_reset"` is not supported and documented as such. To come back to padding, we use it to be able to witness the effect of a partial reset in isolation, and check that the rest of the data is 0, and that the non-masked data behaves as expected.
3. We cannot know in advance if an env is stateful or not, if it is transformed or not etc. So we must have a unified way of treating the output of reset with partial resets. Returning the data given by `_reset` is not an option (see the point about CatFrame above) because the data can be garbage and we have no way of enforcing that it is not. Again, this is not a problem since in practice any call to `reset` with `"_reset"` will be accompanied by the non-reset data.

I hope this clarifies things

docs/source/reference/envs.rst
vmoens: See my comment below: I think this _clarifies_ these tests as we're not assuming anything regarding the behaviour of non-reset entries.

matteobettini:
Thanks for the recap, I understand.

I would still like to highlight that this will come at some costs
(The following comments assume a user that wants to use a TorchRL env in isolation by accessing its `step` and `reset` functions):

1. making partial resets a feature recommended for maintainers only is quite constraining to me. Users should be able to freely call `step()` and `reset()` in a torchrl vectorized env and have a way to ask for which dimensions to reset. 
2. if I have a stateful vectorized env that returns all valid data after a partial reset, I will have no way to access the non-reset data after a partial reset in the future as it would be overwritten by 0s. This data was cool to have before as an additional check (for example in the tests on resetting). Before we had the possibility of it being garbage in some environments, now we have the uniformity of it being garbage in all environments. I am not saying that the library should rely on this garbage in any place. But at least it could not overwirte in with 0s so that people that use enviornments for which they know this is not garbage could still access it.

vmoens:
> making partial resets a feature recommended for maintainers only is quite constraining to me. Users should be able to freely call step() and reset() in a torchrl vectorized env and have a way to ask for which dimensions to reset.

I agree but right now it isn't a public feature and never has been.

> if I have a stateful vectorized env that returns all valid data after a partial reset, I will have no way to access the non-reset data after a partial reset in the future as it would be overwritten by 0s. This data was cool to have before as an additional check (for example in the tests on resetting).

I already commented on that but to the risk of repeating myself: it is inconsistent, unpredictable, sometimes broken and unclear. It should be avoided at all cost. 

> Before, we had the possibility of it being garbage in some environments, now we have the uniformity of it being garbage in all environments. 

I don't think it's the right way of putting it. It was inconsistently garbage, now it is invalid everywhere. It's important that things are consistent across envs, I'm struggling to understand why so much reluctance against having a consistent and predictable API. I'm probably missing something. 

Can I just ask you to comment on this scenario: 
Say we don't override with 0s and keep the tests we have where we check that the data is what was expected (basically remove the 0 padding and don't touch the tests). We're testing a non-feature, which is more a casual occurence of a behaviour that no-one should rely on, not even us. 
If anything happens in the env api that breaks that and tests don't pass, what should the contributor think? That they have to find a workaround? I don't want to lock a non-feature through tests, and this is currently what is happening. It's bad for us and for contributors IMO.
In a few words: why should we test (or allow to test) a non-feature in the tests? We should block it and make sure no one ever relies on it IMO, either in the tests or in algos.

torchrl/envs/common.py
```diff
@@ -1509,39 +1479,61 @@ def reset(
...
+                            f"A partial `'_reset'` key has been passed to `reset` ({reset_key}), "
+                            f"but the corresponding done_key ({done_key}) was not present in the input "
+                            f"tensordict. "
+                            f"This is discouraged, since the input tensordict should contain "
+                            f"all the data not being reset."
+                        )
+                        # we set the done val to tensordict, to make sure that
+                        # _update_during_reset does not pad the value
+                        tensordict.set(done_key, done_val)
+            elif self._allow_done_after_reset:
```
matteobettini, line 1531: this to me should be  `elif not self._allow_done_after_reset`
the `self._allow_done_after_reset` is there to allow dones to be true after they were asked to be reset

matteobettini (commented):
This is impressive! Lots of cool stuff, love the new rollout

Just a few questions

torchrl/envs/common.py
```diff
@@ -1816,25 +1838,109 @@ def policy(td):
...
+        callback,
+    ):
+        tensordicts = []
+        tensordict_ = tensordict
+        for i in range(max_steps):
+            if auto_cast_to_device:
+                tensordict_ = tensordict_.to(policy_device, non_blocking=True)
+            tensordict_ = policy(tensordict_)
+            if auto_cast_to_device:
+                tensordict_ = tensordict.to(env_device, non_blocking=True)
```
matteobettini, line 1871:
```suggestion
                tensordict_ = tensordict_.to(env_device, non_blocking=True)
```
I want a million candies.

Also how did this pass our tests? maybe let's test this

torchrl/envs/libs/pettingzoo.py
```diff
@@ -498,26 +514,8 @@ def _reset(
...
-            else:
-                for _reset in _resets:
-                    if not _reset.all():
-                        all_done = False
-                        break
-
-        if all_done:
-            self._env.reset(**kwargs)
+    def _reset_aec(self, **kwargs) -> Tuple[Dict, Dict]:
+        self._env.reset(**kwargs)
```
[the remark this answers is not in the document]
vmoens: You can check but I think so yes

torchrl/envs/libs/pettingzoo.py
```diff
@@ -608,8 +606,32 @@ def _step(
...
+                else:
+                    done = done & tensordict_out.get(done_key).all()
+                    if not done:
+                        break
+        else:
+            raise RuntimeError(
+                f"Could not find done values in tensordict {tensordict_out}."
+            )
+        return torch.tensor([done], device=self.device)
+
```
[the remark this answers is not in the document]
vmoens: I think they are, terminated is copied and truncated is set to 0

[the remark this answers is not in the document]
matteobettini: what i am saying is that these keys have to be given the same treatment as done. they are defined per agent like done.
and also the _aggregate_done is currently silently putting them in the done

docs/source/reference/envs.rst
```diff
@@ -169,7 +186,95 @@ one can simply call:
...
+internal tests.
+
+Finally, the following assumptions are made and should be kept in mind when
+designing reset functionalities:
+
+- Each ``"_reset"`` is paired with a ``"done"`` entry (+ ``"terminated"`` and,
+  possibly, ``"truncated"``). This means that the following structure is not
+  allowed: ``TensorDict({"done": done, "nested": {"_reset": reset}}, [])``, as
+  the ``"_reset"`` lives at a different nesting level than the ``"done"``.
+- A reset at one level does not preclude the presence of a ``"_reset"`` at lower
```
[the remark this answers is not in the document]
vmoens: Yes not having a reset associated with each done is very hard to come by

torchrl/envs/libs/pettingzoo.py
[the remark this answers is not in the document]
matteobettini: to me it seems that if we keep the lower level _resets than this will be called when any of those is true, even if the higher reset will be set with a `.all()` policy

test/test_libs.py
```diff
@@ -1590,7 +1590,7 @@ def make_vmas():
             [n_workers, list(env.num_envs)[0], n_rollout_samples]
         )
 
-    @pytest.mark.parametrize("num_envs", [1, 10])
+    @pytest.mark.parametrize("num_envs", [1, 2])
```
matteobettini, line 1593: vmas envs are quite cheap, maybe I would keep 10 in the vecotrized dimension to make sure we do not fall in a particular case

torchrl/envs/libs/pettingzoo.py
```diff
@@ -608,8 +606,50 @@ def _step(
...
+            if isinstance(key, tuple):
+                if use_any:
+                    if key[-1] == "done":
+                        done = done | tensordict_out.get(key).any()
+                    if key[-1] == "terminated":
+                        terminated = terminated | tensordict_out.get(key).any()
+                    if key[-1] == "truncated":
+                        truncated = truncated | tensordict_out.get(key).any()
+                    if done:
+                        break
```
matteobettini, line 633: with the new truncated and terminated this is not possible anymore

torchrl/envs/libs/pettingzoo.py
```diff
@@ -608,8 +606,50 @@ def _step(
...
+                else:
+                    if key[-1] == "done":
+                        done = done & tensordict_out.get(key).all()
+                    if key[-1] == "terminated":
+                        terminated = terminated & tensordict_out.get(key).all()
+                    if key[-1] == "truncated":
+                        truncated = truncated & tensordict_out.get(key).all()
+                    if not done:
+                        break
```
matteobettini, line 642: same

torchrl/envs/libs/pettingzoo.py
```diff
@@ -608,8 +606,50 @@ def _step(
...
+        else:
+            raise RuntimeError(
```
matteobettini, line 644: this else branch is currrently reached every time the for loop finishes

I would remove it

matteobettini (commented):
Some PettingZoo related stuff

torchrl/envs/libs/pettingzoo.py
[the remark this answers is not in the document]
vmoens: sorry I don't get that comment. Do you mean that `def _reset` should read `"_reset"` from the input tensordict, and call `_reset_aec` and `_reset_parallel` only if ``True`` / absent?

torchrl/envs/libs/pettingzoo.py
vmoens: I think the problem is that the loop is broken, but this should stay. If there's no done signal something defo went wrong

torchrl/envs/libs/pettingzoo.py
```diff
@@ -464,12 +480,18 @@ def _reset(
         self, tensordict: Optional[TensorDictBase] = None, **kwargs
     ) -> TensorDictBase:
 
+        _reset = tensordict.get("_reset", None)
```
matteobettini, line 483: This crashes when `tensordict is None`

vmoens: Yeah sorry about that, I can't test petting zoo locally so I'm always moving in the dark...

vmoens:
@matteobettini @albertbou92 @BY571 this should be (almost) mergeable.
Tests are passing (VMAS sporadically failing after tests complete and old deps broken because of #1622)

matteobettini:
Why would vmas sporadically fail?

vmoens:
> Why would vmas sporadically fail? 

Not sure you can have a look.
All tests pass but it fails when closing.
Seems like an issue with tensors not released on CUDA, it happens often with VMAS. It isn't even a flaky test, since they all pass, it's just a weird exit status

matteobettini:
> > Why would vmas sporadically fail?
> 
> Not sure you can have a look. All tests pass but it fails when closing. Seems like an issue with tensors not released on CUDA, it happens often with VMAS. It isn't even a flaky test, since they all pass, it's just a weird exit status

Was this happening before this PR?

vmoens:
> it happens often with vmas

I was referring to VMAS CI beyond this PR

torchrl/collectors/collectors.py
```diff
@@ -786,38 +784,14 @@ def iterator(self) -> Iterator[TensorDictBase]:
...
-            return
-        td_reset = self._tensordict.clone(False)
-        any_done = terminated_or_truncated(
-            td_reset,
-            full_done_spec=self.env.output_spec["full_done_spec"],
-            key="_reset",
+    def _update_traj_ids(self, tensordict) -> None:
+        # we can't use the reset keys because they're gone
+        traj_sop = _aggregate_resets(
+            tensordict.get("next"), done_keys=self.env.done_keys
```
matteobettini, line 790: I assume we have to do this becuse there are no reset keys anymore visible by the collector.
Since this is quite counterintitive, what about a `_aggregate_dones()`?
I also have other suggestions in the other comment relating to this function.

matteobettini (commented):
Some final questions/comments

torchrl/envs/utils.py
```diff
@@ -851,12 +974,15 @@ def inner_terminated_or_truncated(data, full_done_spec, key, curr_done_key=()):
...
+def _aggregate_resets(
+    data: TensorDictBase, reset_keys=None, done_keys=None
+) -> torch.Tensor:
     # goes through the tensordict and brings the _reset information to
     # a boolean tensor of the shape of the tensordict.
     batch_size = data.batch_size
     n = len(batch_size)
-
+    if done_keys is not None and reset_keys is None:
+        reset_keys = {_replace_last(key, "done") for key in done_keys}
```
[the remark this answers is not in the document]
vmoens: I don't see the problem, can you elaborate why `we should try to write this better`?
Is it a naming problem? We can rename the function `_aggregate_stop` or smth similar.

> always call the function on the root td and pass the keys with preappended "next" if you want to use that
That introduces some unwanted overhead when we can directly access "next" and read the done_keys. Recall that `td.get(("next", "key"))` is considerably slwoer than `next_td.get("key")` as we do here.


> do this key filtering and conversion outside of the function and let the function just operate on a set of keys to be cconsidered as reset keys

What's your suggestion for `_update_traj_ids` in `collectors.py` for instance? We don't have a "_reset" in the "next" tensordict, but I think this function does its job of aggregating the done signals to read what the trajectory ids are.
What I understand is that the confusion comes from the "reset" in the function name, but what this function really does is just aggregating end-of-trajectory signals (either reset or done) to the root.
Given this, I don't see why it should be changed. It's a private function, properly tested and I think it serves its purpose.

torchrl/collectors/collectors.py
vmoens: yep we can rename the function. I would rather say that it's more direct to aggregate the done rather than the "_reset" which come from the dones

torchrl/envs/common.py
```diff
@@ -1816,25 +1856,121 @@ def policy(td):
...
+        tensordict_ = step_mdp(
+            tensordict,
+            keep_other=True,
+            exclude_action=False,
+            exclude_reward=True,
+            reward_keys=self.reward_keys,
+            action_keys=self.action_keys,
+            done_keys=self.done_keys,
+        )
+        any_done = _terminated_or_truncated(
```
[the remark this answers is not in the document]
vmoens: The best thing it trying it out :)

```python
from torchrl.envs.utils import _terminated_or_truncated

data = TensorDict({"done": [False], ("agent", "done"): [True, False]}, [])
print(_terminated_or_truncated(data))
```
which returns `True`
So what you're saying is that it should be False since there's a False at the root?
I can correct that

torchrl/envs/utils.py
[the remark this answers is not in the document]
matteobettini: connecting to the comment below, we could have

- `_aggeregate_keys(keys=)` which can be called on anything

alternatively we could have both  `_aggeregate_dones` and `_aggregate_resets` where one calls the other

torchrl/envs/common.py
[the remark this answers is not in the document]
matteobettini: I think it should to follow the dominance rule we imposed right?
Or at least in this context definitely becuase we do not want to call reset. 
I don't know in what other contexts this function is used though, but if its primary use is to decide when to call reset then yes

matteobettini (approved):
LGTM

test/test_libs.py
vmoens: @matteobettini
sorry I missed that
they're not very cheap actually, compared to other libs running the vmas tests take a considerable amount of time and torchrl is already a very intensive CI for a small library. I'd rather keep small numbers unless we have a strong reason to have large ones

PR #1616, opened 2023-10-09
[Feature] Warning for `init_random_frames` rounding in collectors

matteobettini:
In collectors  `init_random_frames`  is rounded to the closest upper multiple of `frames_per_batch`.

This PR introduces a warning and docstring that make users aware of this previously silent behaviour.


vmoens (approved):
LGTM

PR #1617, opened 2023-10-09
[BugFix] Raise exception when zero is OOB

vmoens:
## Description

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/test/test_specs.py
+++ b/test/test_specs.py
@@ -3329,6 +3329,21 @@ class TestSpecMasking:
             assert (sp != s).all()
 
 
+class TestOOBound:
+    @pytest.mark.parametrize(
+        "low,high",
+        [
+            [-1, -0.5],
+            [0.5, 1],
+            [torch.tensor([-1, -1]), torch.tensor([-0.5, 1.0])],
+            [torch.tensor([1, -1.0]), torch.tensor([2.0, 2.0])],
+        ],
+    )
+    def test_zero_error(self, low, high):
+        with pytest.raises(ValueError, match="Calling zero()"):
+            BoundedTensorSpec(low=low, high=high, shape=(2,)).zero()
+
+
 if __name__ == "__main__":
     args, unknown = argparse.ArgumentParser().parse_known_args()
     pytest.main([__file__, "--capture", "no", "--exitfirst"] + unknown)
--- a/torchrl/data/tensor_specs.py
+++ b/torchrl/data/tensor_specs.py
@@ -1744,6 +1744,13 @@ class BoundedTensorSpec(TensorSpec):
             dtype=self.dtype,
         )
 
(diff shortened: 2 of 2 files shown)


matteobettini (approved):
LGTM

@smorad what do you think about the warning vs error solution?

I am more towards the warning side as there might be instances where we call `zero()` agnostically of the spec type (like `fake_tensordict()` which is used to initialise buffers in collectors end batched envs).

smorad:
IMO it might make more sense to just implement and use `sample` instead of `zero`. Then you can instantiate tensors that adhere to the spec. If needed, the user can call `.zero_()` on sample.

vmoens:
> IMO it might make more sense to just implement `sample` and remove `zero`. Then you can instantiate tensors that adhere to the spec. If needed, the user can call `.zero_()` on sample.

We use it at multiple places and removing it would be a big bc-breaking thing.
For instance, we can create a [reward of 0 value](https://github.com/pytorch/rl/blob/802f0e4a4695ca6485ebf06e339ef805973eb996/torchrl/envs/gym_like.py#L238) when it's None.
Initializing a td with zeros is much cheaper than sampling + zeroing (eg [here](https://github.com/pytorch/rl/blob/802f0e4a4695ca6485ebf06e339ef805973eb996/torchrl/envs/gym_like.py#L300)), especially if some day we have sampling strategies that are very expensive (in general, any sampling is more expensive than 0). In the case I just linked, one must consider that this operation can be called thousands of times per second, any overhead should be avoided.
Same thing for [info specs](https://github.com/pytorch/rl/blob/f12170179a77eddd53788468a877586c94e68d98/torchrl/envs/libs/gym.py#L1181) and missing keys in reset, and at multiple places in transforms.py.
In PettingZoo, we zero values in turn-based games to pad missing items. Not using 0 would be a problem I think, and here too overhead must be avoided.
In all of these cases (done, reward, infos), zeroing makes sense I think, so IMO there isn't a strong case in favour of dropping the feature.

matteobettini:
> In PettingZoo, we zero values in turn-based games to pad missing items. Not using 0 would be a problem I think, and here too overhead must be avoided.

Agreed.

Especially in light of this I think the error is not suited, maybe even the warning can be avoided.
As long as we use zero() for padding we need to allow it to work as we could have observations or info with a min greater than 0 but we still need to pad.

In light of this my opinion is to not introduce any change but document in the `zero()` of bounded spec that it could be out of bounds

vmoens:
Agreed

The risk is that we legitimately use zero as placeholder, padding or even default for reward (at the end of the day, a None reward is a zero reward) but then users get an exception (or warning) out of nowhere when they did nothing wrong. The source of the error will be buried deep in the modules which won't make it easy to debug.

PR #1624, opened 2023-10-10
[Deprecation] Deprecate ambiguous device for memmap replay buffer

vmoens:
## Description

We allow memmap storages to sit on a "device", which basically means that when accessed, the data will be sent to that device.
This is confusing and suboptimal, and will break once we have memory mapped tensors that are actually tensors (see https://github.com/pytorch/tensordict/pull/541)


matteobettini (approved):
LGTM

So before when they were loaded form disk they were firstly loaded to cpu and then automatically  put on device?

vmoens:
> LGTM
> 
> So before when they were loaded form disk they were firstly loaded to cpu and then automatically put on device?

Something like that yes
It was a very bad design choice.

PR #1640, opened 2023-10-23
[BugFix] Gracefully handle C++ import error in TorchRL

vmoens:
## Description

Allows to import torchrl when _torchrl.so isn't available.

Wdyt @matteobettini ?

### DIFF
--- a/torchrl/_extension.py
+++ b/torchrl/_extension.py
@@ -23,3 +23,10 @@ def _init_extension():
     if not is_module_available("torchrl._torchrl"):
         warnings.warn("torchrl C++ extension is not available.")
         return
+
+
+EXTENSION_WARNING = (
+    "Failed to import torchrl C++ binaries. Some modules (eg, prioritized replay buffers) may not work with your installation. "
+    "If you installed TorchRL from PyPI, please report the bug on TorchRL github. "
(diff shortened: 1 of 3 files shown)


matteobettini:
Why not an error? As I understand it, if that import fails, TorchRL is not installed as it should, so having a warning may just push errors down the line. 

In any case I agree that there should be a custom message printed with instructions.

vmoens:
There's already an error, the goal is to be able to use the other 99.9% of the lib if for some reason the built went wrong. It's a recurrent issue on gh and most of the time I think people would be happy to use the lib as is.
For instznce, you could install torchrl with torch 2.1 and then downgrade torch and still have most of it working. Now just importing the lib will fail in that case and you'll need a local built (which requires ninja, gcc etc)

It doesn't come for free as you say but at least people are informed that some features are not available.

matteobettini:
Sure then makes sense.

Hopefully once we identify the cause it will be less relevant

vmoens:
Somewhat, though this is orthogonal to the M1 issue (which I will solve separately).

The lib is made to be mostly python-based so the few C++ classes we have should not block users from using the lib. Same goes with pytorch version: we allow different versions of pytorch to be used but we don't make the life of users who want to use torchrl with these easy.

PR #1646, opened 2023-10-24
[Feature] Lazy imports for implement_for during torchrl import

vmoens:
## Description

Makes `implement_for` lazy: the import that checks the package version will only occur the first time the function is run.
This allows us to skip the gym imports during torchrl import, and speeds up the lib import itself.

@matteobettini


matteobettini (approved):
Amazing!

PR #1653, opened 2023-10-25
[BugFix] Fix random values in specs with infinite bounds

vmoens:
## Description

Replaces inf / -inf in bounded boxes with the max value given the dtype, to have finite values during rand calls.


vmoens:
Closing this for now but happy to reconsider

Issue #1018, opened 2023-04-03
[BUG] Safe policy modules that call _project on a smaller test env crash


matteobettini:
Let's say we have 2 environments:
-  one `env` for collection with action spec with shape `[500,3,2]`,  with [500] batch_size
-  one `test_env` for testing with action spec with shape `[1,3,2]`,  with [1] batch_size

We create a safe policy 
```python
policy_module = ProbabilisticActor(
        module=policy_module,
        spec=env.action_spec, # `[500,3,2]`
        in_keys=["loc", "scale"],
        distribution_class=IndependentNormal,
        safe=True,
    )
```
 and collect with it in the `env`.

Now, if we run `test_env.rollout(policy_module)` we get an error when the policy tries to project its action to the `test_env.action_spec`, because it was created using `env.action_spec`.

What is the cleanest way to solve this usecase?

##  Proposed solution

Make the `project` methods more flexible?
Or is the solution to keep a different policy for training and one for testing with the 2 separate specs?

vmoens:
I usually solve this by passing the spec of a non-batched env, would that work here?

matteobettini:
For the first issue yes, we could use the `test_env.input_spec` also for the batched one. But this is not super clear IMO.

For the second issue, that is not applicble i think

vmoens:
I'm not sure what your feature dimension is, but what I meant is not having a spec of shape `[1, 3, 2]` but more like `[3, 2]` or `[2]` depending on what the features are.
Happy to consider other solutions if that does not fit. We could implement a `__getitem__` in the specs to be able to reduce it easily:
```
policy = Actor(..., spec=my_spec[0, 0])
```

matteobettini:
```python
base_env = GymEnv("InvertedDoublePendulum-v4")
base_env = TransformedEnv(
    base_env,
    Compose(
        # normalize observations

        DoubleToFloat(
            in_keys=["observation"],
        ),

    ))
env = SerialEnv(3, lambda: base_env)
env.start()
env.action_spec.space.minimum[:] = -0.5

actor_net = nn.Sequential(
    nn.LazyLinear(2 * env.action_spec.shape[-1]),
    NormalParamExtractor(),
)
policy_module = TensorDictModule(
    actor_net, in_keys=["observation"], out_keys=["loc", "scale"]
)
policy_module = ProbabilisticActor(
    module=policy_module,
    spec=env.action_spec,
    in_keys=["loc", "scale"],
[...]

PR #1658, opened 2023-10-27
[Feature] Introduce grouping in VMAS

matteobettini:
This PR allows VMAS to use the MARL api with grouping.

Users will be able to provide a group_map to VMAS.
By default this will try to group agents named `name_n` into a group named "name".

In case at least one agent does not follow this convention and not group map is specified, VMAS will try to group all agents under the "agent" group

### DIFF
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -1548,7 +1548,11 @@ class TestVmas:
     @pytest.mark.parametrize("n_agents", [1, 5])
     @pytest.mark.parametrize("scenario_name", VmasWrapper.available_envs)
     def test_vmas_repr(self, scenario_name, num_envs, n_agents):
-        if n_agents == 1 and scenario_name == "balance":
+        if (
+            n_agents == 1
+            and scenario_name == "balance"
+            or scenario_name == "simple_adversary"
+        ):
             return
         env = VmasEnv(
             scenario=scenario_name,
@@ -1776,6 +1780,37 @@ class TestVmas:
 
         assert env.reward_key not in _td.keys(True, True)
         assert env.action_key not in _td["next"].keys(True, True)
+
+    @pytest.mark.parametrize("n_agents", [1, 5])
+    def test_grouping(self, n_agents, scenario_name="dispersion", n_envs=2):
+        env = VmasEnv(
+            scenario=scenario_name,
+            num_envs=n_envs,
+            n_agents=n_agents,
+        )
+        env = VmasEnv(
+            scenario=scenario_name,
+            num_envs=n_envs,
+            n_agents=n_agents,
+            # Put each agent in a group with its name
+            group_map={
+                agent_name: [agent_name] for agent_name in reversed(env.agent_names)
+            },
+        )
+
+        # CHeck that when setting the action for a specific group that is reflected to the right agent in the backend
+        for group in env.group_map.keys():
+            env.reset()
+            action = env.full_action_spec.zero()
+            action[group, "action"] += 1.0
+            prev_pos = {agent.name: agent.state.pos.clone() for agent in env.agents}
+            _ = env.step(action)
+            pos = {agent.name: agent.state.pos.clone() for agent in env.agents}
+            for agent_name in env.agent_names:
+                if agent_name == group:
+                    assert (pos[agent_name] > prev_pos[agent_name]).all()
+                else:
+                    assert (pos[agent_name] == prev_pos[agent_name]).all()
 
 
 @pytest.mark.skipif(not _has_d4rl, reason="D4RL not found")
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -1775,15 +1775,7 @@ class TestVmas:
         assert env.reward_key not in _td.keys(True, True)
         assert env.action_key not in _td["next"].keys(True, True)
 
-    @pytest.mark.parametrize("categorical", [True])
-    def test_multi_discrete(self, categorical):
-        env = VmasEnv(
-            scenario="simple_reference",
-            continuous_actions=False,
-            categorical_actions=categorical,
-            num_envs=2,
-        )
-        env.rollout(2)
+
 
 
 
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -1589,11 +1589,11 @@ class TestVmas:
 
         env = ParallelEnv(n_workers, make_vmas)
         tensordict = env.rollout(max_steps=n_rollout_samples)
-        env.close()
 
         assert tensordict.shape == torch.Size(
             [n_workers, list(env.num_envs)[0], n_rollout_samples]
         )
+        env.close()
 
     @pytest.mark.parametrize("num_envs", [1, 2])
     @pytest.mark.parametrize("n_workers", [1, 3])
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -1480,6 +1480,7 @@ class TestVmas:
             scenario=scenario_name,
             num_envs=num_envs,
             n_agents=n_agents,
+            group_map=MarlGroupMapType.ALL_IN_ONE_GROUP,
         )
         env.set_seed(0)
         tdreset = env.reset()
@@ -1743,6 +1744,7 @@ class TestVmas:
         env = VmasEnv(
             scenario="simple_tag",
             num_envs=n_envs,
+            group_map=MarlGroupMapType.ALL_IN_ONE_GROUP,
         )
         torch.manual_seed(1)
 
@@ -1775,8 +1777,15 @@ class TestVmas:
         assert env.reward_key not in _td.keys(True, True)
         assert env.action_key not in _td["next"].keys(True, True)
 
-
-
+    @pytest.mark.parametrize("categorical", [True])
+    def test_multi_discrete(self, categorical):
+        env = VmasEnv(
+            scenario="simple_reference",
+            continuous_actions=False,
+            categorical_actions=categorical,
+            num_envs=2,
+        )
+        env.rollout(2)
 
 
 @pytest.mark.skipif(not _has_d4rl, reason="D4RL not found")
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -1554,11 +1554,11 @@ class TestVmas:
             num_envs=num_envs,
             n_agents=n_agents,
         )
-        env.close()
         assert str(env) == (
             f"{VmasEnv.__name__}(num_envs={num_envs}, n_agents={env.n_agents},"
             f" batch_size={torch.Size((num_envs,))}, device={env.device}) (scenario={scenario_name})"
         )
+        env.close()
 
     @pytest.mark.parametrize("num_envs", [1, 10])
     @pytest.mark.parametrize("n_workers", [1, 3])
@@ -1665,7 +1665,7 @@ class TestVmas:
             )
             return env
 
-        env = ParallelEnv(2, make_vmas)
+        env = make_vmas()
 
         assert env.rollout(max_steps=3).device == devices[first]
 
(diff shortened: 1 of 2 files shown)


torchrl/envs/libs/vmas.py
```diff
@@ -50,9 +63,75 @@ def _get_envs():
     ]
 
 
+@set_gym_backend("gym")
```
matteobettini, line 66: Detached vmas from gym spec conversion to have custom conversion rules

torchrl/envs/libs/vmas.py
```diff
@@ -318,71 +467,93 @@ def _reset(
...
+            group_action_list = list(self.read_action(group_action, group=group))
+            agent_indices.update(
+                {
+                    self.agent_names_to_indices_map[agent_name]: i + n_agents
+                    for i, agent_name in enumerate(agent_names)
+                }
+            )
+            n_agents += len(agent_names)
+            action_list += group_action_list
+        action = [action_list[agent_indices[i]] for i in range(self.n_agents)]
```
matteobettini, line 519: this will convert actions from group dicts to the list (ordered as it should)

This bit of logic is the focus of the new test

matteobettini (commented):
@vmoens this is ready for review when u can

test/test_libs.py
```diff
@@ -1538,13 +1541,18 @@ def test_vmas_spec_rollout(
         for e in [env, wrapped]:
             e.set_seed(0)
             check_env_specs(e, return_contiguous=False if e.het_specs else True)
+            env.close()
```
vmoens, line 1544: we should avoid single character variables like `e`

test/test_libs.py
```diff
@@ -1636,10 +1646,10 @@ def make_vmas():
             td_reset.set(done_key, tensordict[..., -1].get(("next", done_key)))
         reset = td_reset["_reset"]
         tensordict = env.reset(td_reset)
-        assert not tensordict["done"][reset].all().item()
-        # vmas resets all the agent dimension if only one of the agents needs resetting
-        # thus, here we check that where we did not reset any agent, all agents are still done
-        assert tensordict["done"].all(dim=2)[~reset.any(dim=2)].all().item()
+
+        assert not tensordict["done"][reset].any().item()
```
vmoens, line 1650:
```suggestion
        assert not tensordict.get("done")[reset].any()
```

test/test_libs.py
```diff
@@ -1636,10 +1646,10 @@ def make_vmas():
...
+        assert tensordict["done"][~reset].all().item()
```
vmoens, line 1651:
```suggestion
        assert tensordict.get("done")[~reset].all()
```

test/test_libs.py
```diff
@@ -1769,6 +1781,37 @@ def test_collector_heterogeneous(self, n_envs=10, frames_per_batch=20):
...
+            scenario=scenario_name,
+            num_envs=n_envs,
+            n_agents=n_agents,
+            # Put each agent in a group with its name
+            group_map={
+                agent_name: [agent_name] for agent_name in reversed(env.agent_names)
+            },
+        )
+
+        # CHeck that when setting the action for a specific group that is reflected to the right agent in the backend
```
vmoens, line 1801:
```suggestion
        # Check that when setting the action for a specific group, it is reflected to the right agent in the backend
```

test/test_libs.py
```diff
@@ -1769,6 +1781,37 @@ def test_collector_heterogeneous(self, n_envs=10, frames_per_batch=20):
...
+        for group in env.group_map.keys():
+            env.reset()
+            action = env.full_action_spec.zero()
+            action[group, "action"] += 1.0
```
vmoens, line 1805:
```suggestion
            action.get((group, "action")) += 1.0
```
although we should avoid inplace modification in tensordicts like this (eg, if you're working with a lazy stack it won't do anything)

test/test_libs.py
```diff
@@ -1769,6 +1781,37 @@ def test_collector_heterogeneous(self, n_envs=10, frames_per_batch=20):
...
+            prev_pos = {agent.name: agent.state.pos.clone() for agent in env.agents}
+            _ = env.step(action)
```
vmoens, line 1807:
```suggestion
            env.step(action)
```

torchrl/envs/libs/vmas.py
```diff
@@ -50,9 +63,75 @@ def _get_envs():
...
+            f"spec of type {type(spec).__name__} is currently unaccounted for vmas"
+        )
+
+
 class VmasWrapper(_EnvWrapper):
     """Vmas environment wrapper.
 
+    Args:
+        env (``vmas.simulator.environment.environment.Environment``): the vmas environment to wrap.
+        categorical_actions (bool, optional): if the enviornments actions are discrete, whether to transform
```
vmoens, line 126:
```suggestion
        categorical_actions (bool, optional): if the environment actions are discrete, whether to transform
```

torchrl/envs/libs/vmas.py
```diff
@@ -50,9 +63,75 @@ def _get_envs():
...
+            them to categorical or one-hot.`
```
vmoens, line 127:
```suggestion
            them to categorical or one-hot.
```

torchrl/envs/libs/vmas.py
```diff
@@ -50,9 +63,75 @@ def _get_envs():
...
+        group_map (MarlGroupMapType or Dict[str, List[str]]], optional): how to group agents in tensordicts for
+            input/output. By default, if agent names follow the "{name}_{int}"
```
vmoens, line 129:
```suggestion
            input/output. By default, if the agent names follow the ``"<name>_<int>"``
```

torchrl/envs/libs/vmas.py
```diff
@@ -50,9 +63,75 @@ def _get_envs():
...
+            convention, they will be grouped by "{name}", if they do not follow this convention, they will be all put
```
vmoens, line 130:
```suggestion
            convention, they will be grouped by ``"<name>"``. If they do not follow this convention, they will be all put
```

torchrl/envs/libs/vmas.py
```diff
@@ -50,9 +63,75 @@ def _get_envs():
...
+            in one group named "agents".
```
vmoens, line 131:
```suggestion
            in one group named ``"agents"``.
```

torchrl/envs/libs/vmas.py
```diff
@@ -50,9 +63,75 @@ def _get_envs():
...
+            Otherwise, a group map can be specified or selected from some premade options.
+            See :class:`torchrl.envs.utils.MarlGroupMapType` for more info.
```
vmoens, line 133:
```suggestion
            See :class:`~torchrl.envs.utils.MarlGroupMapType` for more info.
```

torchrl/envs/libs/vmas.py
```diff
@@ -170,37 +251,132 @@ def _build_env(
 
         return env
 
-    @set_gym_backend("gym")
+    def _get_default_group_map(self, agent_names: List[str]):
+        # This function performs the default grouping in vmas
+        # Agents with names "str_int" will be grouped in group name "str"
+        # If any of the agents to not follow the naming convention, we fall back
+        # to all agents in one group named "agents"
```
vmoens, line 258:
```suggestion
        # This function performs the default grouping in vmas.
        # Agents with names "<name>_<int>" will be grouped in group name "<name>".
        # If any of the agents does not follow the naming convention, we fall back
        # back on having all agents in one group named "agents".
```

torchrl/envs/libs/vmas.py
```diff
@@ -170,37 +251,132 @@ def _build_env(
...
+        group_map = {}
+        follows_convention = True
+        for agent_name in agent_names:
+            # See if the agent follows the convention "name_int"
```
vmoens, line 262:
```suggestion
            # See if the agent follows the convention "<name>_<int>"
```

torchrl/envs/libs/vmas.py
```diff
@@ -170,37 +251,132 @@ def _build_env(
...
+            agent_name_split = agent_name.split("_")
+            if len(agent_name_split) == 1:
+                follows_convention = False
+            try:
+                int(agent_name_split[-1])
+            except ValueError:
+                follows_convention = False
```
vmoens, line 269: I'd rather not use `try/except` in this case. What about `str.isdigit()`?
```suggestion
        follows_convention = agent_name_split[-1].isdigit()
```

torchrl/envs/libs/vmas.py
```diff
@@ -170,37 +251,132 @@ def _build_env(
...
+        self.het_specs = False
+        self.het_specs_map = {}
+        for group in self.group_map.keys():
+            (
+                group_observation_spec,
+                group_action_spec,
+                group_reward_spec,
+                group_info_spec,
+            ) = self._make_unbatched_group_specs(group)
+            self.unbatched_action_spec[group] = group_action_spec
```
vmoens, line 318: is `self.unbatched_action_spec` public on purpose?
Can we review the public attributes of VMAS? I feel there are a lot of them and most are undocumented. This will hamper development in the future. I would make anything that is not user facing private to give us more dev freedom.

torchrl/envs/libs/vmas.py
```diff
@@ -170,37 +251,132 @@ def _build_env(
...
+            self.unbatched_observation_spec[group] = group_observation_spec
+            self.unbatched_reward_spec[group] = group_reward_spec
+            if group_info_spec is not None:
+                self.unbatched_observation_spec[(group, "info")] = group_info_spec
+            group_het_specs = isinstance(
+                group_observation_spec, LazyStackedCompositeSpec
+            ) or isinstance(group_action_spec, LazyStackedCompositeSpec)
+            self.het_specs_map[group] = group_het_specs
+            self.het_specs += group_het_specs
```
vmoens, line 327: using `+=` with a boolean isn't very transparent, can we use a plain `__or__`?

torchrl/envs/libs/vmas.py
```diff
@@ -419,14 +590,10 @@ def read_reward(self, rewards):
         rewards = _selective_unsqueeze(rewards, batch_size=self.batch_size)
         return rewards
 
-    def read_action(self, action):
+    def read_action(self, action, group: str):
```
vmoens, line 593: this is a public method and we're adding a new arg without default value, hence it's bc-breaking. Any chance we can make this non-bc breaking?

vmoens (approved):
> @vmoens this is ready for review when u can

I guess you mean "you", "u" is my father :p 

LGTM, I left some comments and suggestions

torchrl/envs/libs/vmas.py
matteobettini: Sure I will make private all I don't need to be public.
This will be bc-breaking though as the methods will not be available anymore.

Regarding the unbatched_spec, these are needed when creating modules. Giving the specs with the batch_size to the modules will lead to errors #1018. These attirbutes are used in BenchMARL for example

torchrl/envs/libs/vmas.py
matteobettini: I can, or I can make it private, as requested above, which will be more bc breaking though

matteobettini:
Ok I made all methods private.

I would like to keep the attributes public so that they can be accessed by libraries like BenchMARL.
I'll document them

torchrl/envs/libs/vmas.py
```diff
@@ -411,22 +590,18 @@ def read_info(self, infos: Dict[str, torch.Tensor]) -> torch.Tensor:
...
         done = _selective_unsqueeze(done, batch_size=self.batch_size)
         return done
 
-    def read_reward(self, rewards):
+    def _read_reward(self, rewards):
         rewards = _selective_unsqueeze(rewards, batch_size=self.batch_size)
         return rewards
 
-    def read_action(self, action):
+    def _read_action(self, action, group: str):
```
vmoens, line 601: I think this solution is worse, it used to be public and now it's private. Is there a solution that does not break previous behaviour?

vmoens (commented):
sorry for the spammy reviews

torchrl/envs/libs/vmas.py
matteobettini: yes but it will leave the method public.

Just so to understand, do you want these methods public or private? I am happy to leave them public and making thewm bc comptible, but I thoght that private is what you wanted

vmoens: The least bc-breaking the better.
I wasn't pointing at these methods in particular, more at all the attributes created when creating specs. It seems like there are loads of them.
`read_action` is public in other wrappers so I don't think it's a problem if it's public here too.

matteobettini:
ok now there shold be no bc changes and the public attributes should be all doxumented

PR #1663, opened 2023-10-30
[Feature] CatFrames constant padding

albertbou92:
## Description

Allows the CatFrames Transforms to use constant padding using any specified value, instead of only zeros.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -2552,8 +2552,9 @@ class CatFrames(ObservationTransform):
             to be concatenated. Defaults to ["pixels"].
         out_keys (sequence of NestedKey, optional): keys pointing to where the output
             has to be written. Defaults to the value of `in_keys`.
-        padding (str, optional): the padding method. One of ``"same"`` or ``"zeros"``.
-            Defaults to ``"same"``, ie. the first value is uesd for padding.
+        padding (str, optional): the padding method. One of ``"same"`` or ``"constant"``.
+            Defaults to ``"same"``, ie. the first value is used for padding.
+        padding_value (float, optional): the value to use for padding if ``padding="constant"``.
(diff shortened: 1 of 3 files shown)


torchrl/envs/transforms/transforms.py
```diff
@@ -2627,7 +2628,7 @@ class CatFrames(ObservationTransform):
         "dim must be < 0 to accomodate for tensordict of "
         "different batch-sizes (since negative dims are batch invariant)."
     )
-    ACCEPTED_PADDING = {"same", "zeros"}
+    ACCEPTED_PADDING = {"same", "constant"}
```
vmoens, line 2631: This is BC breaking, can we keep the "zeros" and maybe raise a deprecation warning if it's being used?

vmoens (commented):
Sounds reasonable but these changes are bc breaking. IMO we can live with both zeros and constant (zeros would be a shortcut to constant with 0 value). If you think it's best to keep only constant we should raise a depreciation warning.

torchrl/envs/transforms/transforms.py
albertbou92: ok! I just did that

torchrl/envs/transforms/transforms.py
```diff
@@ -2701,7 +2712,12 @@ def _make_missing_buffer(self, data, buffer_name):
         shape[self.dim] = d * self.N
         shape = torch.Size(shape)
         getattr(self, buffer_name).materialize(shape)
-        buffer = getattr(self, buffer_name).to(data.dtype).to(data.device).zero_()
+        buffer = (
+            getattr(self, buffer_name)
+            .to(data.dtype)
+            .to(data.device)
```
vmoens, line 2718:
```suggestion
            .to(dtype=data.dtype, device=data.device)
```

PR #1666, opened 2023-10-30
[Algorithm] Discrete CQL

BY571:
## Description

Adds discrete (DQN) CQL objective and example

## Motivation and Context

## Types of changes

## Checklist


BY571:
image


Converges on Cartpole as expected. Just needs some cleanup + tests

examples/cql/discrete_cql_online.py
```diff
@@ -0,0 +1,198 @@
...
+        cfg.collector.env_per_collector
+        * cfg.collector.frames_per_batch
+        * cfg.optim.utd_ratio
+    )
+    prb = cfg.replay_buffer.prb
+    eval_rollout_steps = cfg.env.max_episode_steps
+    eval_iter = cfg.logger.eval_iter
+    frames_per_batch = cfg.collector.frames_per_batch
+
+    sampling_start = time.time()
```
vmoens, line 95:
```suggestion
    collected_frames = 0
    pbar = tqdm.tqdm(total=cfg.collector.total_frames)

    init_random_frames = cfg.collector.init_random_frames
    num_updates = int(
        cfg.collector.env_per_collector
        * cfg.collector.frames_per_batch
        * cfg.optim.utd_ratio
    )
    prb = cfg.replay_buffer.prb
    eval_rollout_steps = cfg.env.max_episode_steps
    eval_iter = cfg.logger.eval_iter
    frames_per_batch = cfg.collector.frames_per_batch

    start_time = sampling_start = time.time()
```

test/test_cost.py
```diff
@@ -5164,6 +5165,367 @@ def test_cql_batcher(
...
+    ):
+        # Actor
+        if action_spec_type == "one_hot":
+            action_spec = OneHotDiscreteTensorSpec(action_dim)
+        elif action_spec_type == "categorical":
+            action_spec = DiscreteTensorSpec(action_dim)
+        # elif action_spec_type == "nd_bounded":
+        #     action_spec = BoundedTensorSpec(
+        #         -torch.ones(action_dim), torch.ones(action_dim), (action_dim,)
+        #     )
```
vmoens, line 5189:
```suggestion
```

test/test_cost.py
```diff
@@ -5164,6 +5165,367 @@ def test_cql_batcher(
...
+        else:
+            raise ValueError(f"Wrong {action_spec_type}")
```
vmoens, line 5191:
```suggestion
            raise ValueError(f"Wrong action spec type: {action_spec_type}")
```

torchrl/objectives/cql.py
```diff
@@ -793,3 +797,330 @@ def _alpha(self):
...
+            priority (NestedKey): The input tensordict key where the target priority is written to.
+                Defaults to ``"td_error"``.
+            reward (NestedKey): The input tensordict key where the reward is expected.
+                Will be used for the underlying value estimator. Defaults to ``"reward"``.
+            done (NestedKey): The key in the input TensorDict that indicates
+                whether a trajectory is done. Will be used for the underlying value estimator.
+                Defaults to ``"done"``.
+            terminated (NestedKey): The key in the input TensorDict that indicates
+                whether a trajectory is terminated. Will be used for the underlying value estimator.
+                Defaults to ``"terminated"``.
```
vmoens, line 904: this does not match the keys below

torchrl/objectives/cql.py
```diff
@@ -793,3 +797,330 @@ def _alpha(self):
...
+                # let's try with action_space then
+                try:
+                    action_space = value_network.action_space
+                except AttributeError:
+                    raise ValueError(self.ACTION_SPEC_ERROR)
+        if action_space is None:
+            warnings.warn(
+                "action_space was not specified. DiscreteCQLLoss will default to 'one-hot'."
+                "This behaviour will be deprecated soon and a space will have to be passed."
+                "Check the DiscreteCQLLoss documentation to see how to pass the action space. "
```
vmoens, line 966:
```suggestion
                "action_space was not specified. DiscreteCQLLoss will default to 'one-hot'. "
                "This behaviour will be deprecated soon and a space will have to be passed. "
                "Check the DiscreteCQLLoss documentation to see how to pass the action space."
```

torchrl/objectives/cql.py
```diff
@@ -793,3 +797,330 @@ def _alpha(self):
...
+    @_cache_values
+    def _cached_detached_target_value_params(self):
+        return self.target_value_network_params.detach()
+
+    def cql_loss(self, q_values, current_action):
+        """Computes the CQL loss for a batch of Q-values and one-hot encoded actions."""
+        logsumexp = torch.logsumexp(q_values, dim=-1, keepdim=True)
+        q_a = (q_values * current_action).sum(dim=-1, keepdim=True)
+
+        return (logsumexp - q_a).mean()
```
vmoens, line 1126: can we return metadata too, like we're hoping to do for all losses in the future?

torchrl/objectives/cql.py
```diff
@@ -793,3 +797,330 @@ def _alpha(self):
...
+        if self._in_keys is None:
+            self._set_in_keys()
+        return self._in_keys
+
+    @in_keys.setter
+    def in_keys(self, values):
+        self._in_keys = values
+
+    @dispatch
+    def forward(self, tensordict: TensorDictBase) -> TensorDict:
```
vmoens, line 1057: this should just be a couple of lines with dqn_loss and cql_loss IMO

vmoens (commented):
Great work! I left some high level comments, can you have a look?
Thanks for this

torchrl/objectives/cql.py
BY571: I actually tried to inherit from the DQN class and then do something like super.forward(tensordict) and only have the cql_loss calculation added but I got circular importing issues. Do you have any suggestions?

vmoens: Oh I wasn't suggesting to inherit from DQN, it's ok if they're separated. But the forward should just be a composition of `loss_actor` and `loss_critic` like we did in other losses (eg, TD3), where each sub-loss returns a tensor and a dict of metadata.

BY571: ah, got it! Should be adapted accordingly now.

vmoens (commented):
I don't understand it.
CQL loss is called in value loss, not in forward, why is that?
Why do we call item() on CQL loss value? Conventionally all losses in the output tensordict of a loss module should be differentials. 
Can you give me some context?

BY571:
> I don't understand it. CQL loss is called in value loss, not in forward, why is that? Why do we call item() on CQL loss value? Conventionally all losses in the output tensordict of a loss module should be differentials. Can you give me some context?

The CQL loss is more like an auxiliary term for the value loss not for a separate model like the actor. It just augments the value loss.  We could separate it but then we would need to forward pass through the model again to obtain the current q values, which would slow down the process and I think there is no need to obtain only the cql loss as in itself it's incomplete.

torchrl/objectives/cql.py
```diff
@@ -793,3 +797,355 @@ def _alpha(self):
...
+            inplace=True,
+        )
+        loss = distance_loss(pred_val_index, target_value, self.loss_function).mean()
+
+        metadata = {
+            "td_error": td_error.mean(0).detach(),
+            "loss_cql": cql_loss.item(),
+            "pred_value": pred_val.mean().detach(),
+            "target_value": target_value.mean().detach(),
+        }
```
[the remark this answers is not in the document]
BY571: you are right, I must have deleted it. Sorry for the confusion, I just updated and fixed it :)

vmoens:
What do you think of https://github.com/BY571/rl/pull/1? I think being able to run ablation studies has some value.

We need to fix the categorical case.

BY571:
> What do you think of [BY571#1](https://github.com/BY571/rl/pull/1)? I think being able to run ablation studies has some value.
> 
> We need to fix the categorical case.

I think yes, if someone wants to check how the cql loss term influences the agent performance and want to have simple "on/off" capability it makes sense. The changes you did look good, I also pushed some adaption for the categorical case to calculate the cql loss.

vmoens:
Cool LMK when you've merge the PR

torchrl/objectives/cql.py
```diff
@@ -793,3 +797,359 @@ def _alpha(self):
...
+        self._value_estimator.set_keys(**tensor_keys)
+
+    @property
+    def device(self) -> torch.device:
+        for p in self.parameters():
+            return p.device
+        raise RuntimeError(
+            "At least one of the networks of CQLLoss must have trainable " "parameters."
+        )
```
vmoens, line 1047: loss.device will be deprecated in the future to allow losses to be distributed on multiple devices. Let's not add it to new losses

torchrl/objectives/cql.py
```diff
@@ -793,3 +797,359 @@ def _alpha(self):
...
+
+    def __init__(
+        self,
+        value_network: Union[QValueActor, nn.Module],
+        *,
+        loss_function: Optional[str] = "l2",
+        delay_value: bool = True,
+        gamma: float = None,
+        action_space=None,
+        priority_key: str = None,
```
vmoens, line 934: passing keys to constructors will be deprecated for existing losses, let's not add it to new ones.

BY571:
Just merged and fixed the open issues. Let me know what you think.
Also, thank you for insisting on making the losses separate, I took advantage of it already and compared base DQN vs DQN+CQL loss :)
image

vmoens:
That looks great!
There are still 19 broken tests in the new test class and the example isn't running either.

vmoens (approved):
Cool let's merge this!

Issue #1659, opened 2023-10-27
[Feature Request] Add linear interpolation to FrameSkipTransform


maxweissenbacher:
## Motivation

I would suggest adding an option for linear interpolation to the `FrameSkipTransform` ([documentation here](https://pytorch.org/rl/reference/generated/torchrl.envs.transforms.FrameSkipTransform.html#torchrl.envs.transforms.FrameSkipTransform)). This means, instead of repeating the same action `frame_skip` number of times for each step in the environment, we could instead keep track of action from the previous step and then interpolate towards the new action over `frame_skip` steps.

This means that at the first step, the previous action would be executed, at the final step, the new action would be executed, and in between, we interpolate linearly between the two actions.

## Solution

Add a variant of the FrameSkipTransform, or otherwise add this as an option in the params of FrameSkipTransform.

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

Zekrom-7780:
@vmoens , can I work on this issue?

vmoens:
Of course!

vmoens:
Some pointers: we could want to make this stateless, if a buffer needs to be carried it could be put in the output tensordict under a private entry name, eg "_action_interp_buffer". We can also make it stateful within the transform. Both are viable options. A good example of the former approach is StepCounter or other stateless transforms like TargetReturn. CatFrames is a good example of stateful transform.

Another datapoint: be careful about reset, which will require some clever masking of the buffer with the reset key. Again have a look at the above mentioned transforms to get a sense of how it should work.

PR #1692, opened 2023-11-10
[Feature] Added linear interpolation to FrameSkipTransform

Zekrom-7780:
## Description

- Added an `action_interp` parameter to enable or disable the action interpolation feature.
- Added an `action_interp_buffer` attribute to store the previous action for interpolation.

## Motivation and Context

Solves Issue #1659 

## Types of changes

## Checklist

## Reviewers 

cc. @vmoens 

P.S. I don't know if I have to add unit-tests for this, OR if this is a breaking change, so I haven't added them here.

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -3894,11 +3894,22 @@ class FrameSkipTransform(Transform):
         frame_skip (int, optional): a positive integer representing the number
             of frames during which the same action must be applied.
         action_interp (bool, optional): whether to perform action interpolation over frame_skip steps.
+            Defaults to ``False``.
+        Examples:
+            >>> from torchrl.envs import GymEnv
+            >>> env = TransformedEnv(GymEnv("CartPole-v1"), FrameSkipTransform(3, action_interp=False))
+            >>> print(env.rollout)
+            # add print here
+            >>> env = TransformedEnv(GymEnv("CartPole-v1"), FrameSkipTransform(3, action_interp=True))
+            >>> print(env.rollout)
+            # add print here
+            
 
     """
+    
 
     def __init__(self, frame_skip: int = 1, action_interp: bool = False):
-        super().__init()
+        super().__init__()
         if frame_skip < 1:
             raise ValueError("frame_skip should have a value greater or equal to one.")
         self.frame_skip = frame_skip
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -3932,11 +3932,6 @@ class FrameSkipTransform(Transform):
 
         return next_tensordict.set(reward_key, reward)
 
-    def forward(self, tensordict):
-        raise RuntimeError(
-            "FrameSkipTransform can only be used when appended to a transformed env."
-        )
-
     def _linear_interpolation(
         self, start_action: Tensor, end_action: Tensor, num_steps: int
     ) -> List[Tensor]:
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -3883,6 +3883,7 @@ class DiscreteActionProjection(Transform):
             f"in_keys_inv={self.in_keys_inv})"
         )
 
+
 class FrameSkipTransform(Transform):
     """A frame-skip transform.
 
@@ -3905,7 +3906,8 @@ class FrameSkipTransform(Transform):
         self.action_interp_buffer = None
 
     def _step(
-        self, tensordict: TensorDictBase, next_tensordict: TensorDictBase) -> TensorDictBase:
+        self, tensordict: TensorDictBase, next_tensordict: TensorDictBase
+    ) -> TensorDictBase:
         parent = self.parent
         if parent is None:
             raise RuntimeError("parent not found for FrameSkipTransform")
@@ -3943,7 +3945,7 @@ class FrameSkipTransform(Transform):
             interpolated_action = start_action + alpha * (end_action - start_action)
             interpolation_steps.append(interpolated_action)
         return interpolation_steps
-    
+
     def forward(self, tensordict):
         raise RuntimeError(
             "FrameSkipTransform can only be used when appended to a transformed env."
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -3908,12 +3908,13 @@ class FrameSkipTransform(Transform):
     """
     
 
-    def __init__(self, frame_skip: int = 1, action_interp: bool = False):
-        super().__init__()
+    def __init__(self, frame_skip: int = 1, action_interp: bool = False, action_key: str = "_action"):
+        super().__init()
         if frame_skip < 1:
             raise ValueError("frame_skip should have a value greater or equal to one.")
         self.frame_skip = frame_skip
         self.action_interp = action_interp
+        self.action_key = action_key
         self.action_interp_buffer = None
 
     def _step(
@@ -3924,9 +3925,8 @@ class FrameSkipTransform(Transform):
             raise RuntimeError("parent not found for FrameSkipTransform")
 
         if self.action_interp:
-            action_key = "_action"
-            current_action = tensordict.get(action_key)
-            next_action = next_tensordict.get(action_key)
+            current_action = tensordict.get(self.action_key)
+            next_action = next_tensordict.get(self.action_key)
             if self.action_interp_buffer is not None:
                 interpolated_actions = self._linear_interpolation(
                     self.action_interp_buffer, next_action, self.frame_skip
@@ -3935,7 +3935,7 @@ class FrameSkipTransform(Transform):
             else:
                 interpolated_actions = [current_action] * (self.frame_skip - 1)
                 self.action_interp_buffer = next_action
-            next_tensordict.set(action_key, interpolated_actions)
+            next_tensordict.set(self.action_key, interpolated_actions)
 
         reward_key = parent.reward_key
         reward = next_tensordict.get(reward_key)
@@ -3956,6 +3956,7 @@ class FrameSkipTransform(Transform):
             interpolated_action = start_action + alpha * (end_action - start_action)
             interpolation_steps.append(interpolated_action)
         return interpolation_steps
+
 
     def forward(self, tensordict):
         raise RuntimeError(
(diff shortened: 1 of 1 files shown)


torchrl/envs/transforms/transforms.py
```diff
@@ -3893,28 +3893,59 @@ class FrameSkipTransform(Transform):
...
         frame_skip (int, optional): a positive integer representing the number
             of frames during which the same action must be applied.
+        action_interp (bool, optional): whether to perform action interpolation over frame_skip steps.
 
     """
 
-    def __init__(self, frame_skip: int = 1):
-        super().__init__()
+    def __init__(self, frame_skip: int = 1, action_interp: bool = False):
+        super().__init()
```
vmoens, line 3901: why this change?

torchrl/envs/transforms/transforms.py
```diff
@@ -3893,28 +3893,59 @@ class FrameSkipTransform(Transform):
     Args:
         frame_skip (int, optional): a positive integer representing the number
             of frames during which the same action must be applied.
+        action_interp (bool, optional): whether to perform action interpolation over frame_skip steps.
```
vmoens, line 3896:
```suggestion
        action_interp (bool, optional): whether to perform action interpolation over frame_skip steps.
            Defaults to ``False``.
```

vmoens: Let's add an example that in the docstring
```suggestion
Suggested change
        action_interp (bool, optional): whether to perform action interpolation over frame_skip steps.
            Defaults to ``False``.

        Examples:
            >>> from torchrl.envs import GymEnv
            >>> env = TransformedEnv(GymEnv("CartPole-v1"), FrameSkipTransform(3, action_interp=False))
            >>> print(env.rollout)
            # add print here
            >>> env = TransformedEnv(GymEnv("CartPole-v1"), FrameSkipTransform(3, action_interp=True))
            >>> print(env.rollout)
            # add print here
            
```

torchrl/envs/transforms/transforms.py
```diff
@@ -3893,28 +3893,59 @@ class FrameSkipTransform(Transform):
...
 
     def _step(
         self, tensordict: TensorDictBase, next_tensordict: TensorDictBase
     ) -> TensorDictBase:
         parent = self.parent
         if parent is None:
             raise RuntimeError("parent not found for FrameSkipTransform")
+
+        if self.action_interp:
+            action_key = "_action"
```
vmoens, line 3916: this should be a parameter in the class constructor

torchrl/envs/transforms/transforms.py
```diff
@@ -3893,28 +3893,59 @@ class FrameSkipTransform(Transform):
...
+            current_action = tensordict.get(action_key)
+            next_action = next_tensordict.get(action_key)
+            if self.action_interp_buffer is not None:
+                interpolated_actions = self._linear_interpolation(
+                    self.action_interp_buffer, next_action, self.frame_skip
+                )
+                self.action_interp_buffer = next_action
+            else:
+                interpolated_actions = [current_action] * (self.frame_skip - 1)
```
vmoens, line 3925: would it make sense to store this contiguously (with a torch.stack around it)?

torchrl/envs/transforms/transforms.py
```diff
@@ -3893,28 +3893,59 @@ class FrameSkipTransform(Transform):
...
         for _ in range(self.frame_skip - 1):
             next_tensordict = parent._step(tensordict)
             reward = reward + next_tensordict.get(reward_key)
+
         return next_tensordict.set(reward_key, reward)
 
+    def _linear_interpolation(
+        self, start_action: Tensor, end_action: Tensor, num_steps: int
+    ) -> List[Tensor]:
+        if num_steps <= 0:
```
vmoens, line 3940: can num_steps be smaller than 0?

torchrl/envs/transforms/transforms.py
```diff
@@ -3893,28 +3893,59 @@ class FrameSkipTransform(Transform):
...
+            return [start_action]
+        interpolation_steps = [start_action]
+        for step in range(1, num_steps):
+            alpha = step / num_steps
+            interpolated_action = start_action + alpha * (end_action - start_action)
+            interpolation_steps.append(interpolated_action)
```
vmoens, line 3946: I'm not sure I see what this block is doing, can you give an example?
It'd be cool to see (eg in the tests) what you would expect to get from it

vmoens (commented):
Thanks so much for this!
We need to add a couple of tests for this feature.
What should happen at reset time, if there is more than one env being run?

torchrl/envs/transforms/transforms.py
Zekrom-7780: Ohh, that is a mistake from my side, sorry!

torchrl/envs/transforms/transforms.py
Zekrom-7780: Suree, will add it in the next commit

torchrl/envs/transforms/transforms.py
Zekrom-7780: Got it

torchrl/envs/transforms/transforms.py
Zekrom-7780: I don't understand what you mean by this, could you like elaborate?

torchrl/envs/transforms/transforms.py
Zekrom-7780: They cannot be, but still, I felt the need to add them, just in case the people using them don't run into any issues (like trying to enter a negative number just for fun and all)

torchrl/envs/transforms/transforms.py
Zekrom-7780: Ok, so my reasoning here was for let's say `num_steps = 5`, we have:

At `step = 1`, `alpha = 1/5` and `interpolated_action` is calculated as follows:
`interpolated_action=start_action+(1/5)*(end_action - start_action)`

Similarly for each step (like for the 4th step, `alpha=4/5` ) and `interpolated_action` is calculated as follows:
`interpolated_action=start_action+(4/5)*(end_action - start_action)`

And finally, at `step = 5`, `alpha = 5/5` (which is 1), and `interpolated_action` becomes :
`interpolated_action=end_action`

So like the`interpolation_steps` list contains the sequence of actions that go from from `start_action` to `end_action` over the specified number of steps, which in this example is 5. 
I felt that this made sense for moving from `start_action` to `end_action` , for smooth action transitions over multiple steps

Zekrom-7780: I don't know if this makes sense, but I felt that this is how i have to do it

vmoens: Ok got it, do you mind if I try to do a vectorized version of this once you have written tests? I think we can make it faster

Zekrom-7780: Suree! please do it!
I'll send another round of commits by the next 3ish hours
Also please could you answer what 

> would it make sense to store this contiguously (with a torch.stack around it)?

This means?

Zekrom-7780:
@vmoens, added the changes that you told me to.

## Ready for another round of comments

> would it make sense to store this contiguously (with a torch.stack around it)?

I wasn't sure what to do here, so I just wrapped `interpolated_actions` around a torch.stack and returned it.

Added `action_key = "_action"` as a parameter in the constructor, and now, will have to change the docstrings as well
Is the action correct, or you wanted a different word, or a different way of doing it?

> Thanks so much for this!
We need to add a couple of tests for this feature.
What should happen at reset time, if there is more than one env being run?

Here, Should I create a `action_interp_buffer` (let's call it that) for each environment, and like reset when that environment is called?
But then how do I get the envirnoment like ID so as to know which one to reset?

vmoens:
> @vmoens, added the changes that you told me to.
> 
> ## Ready for another round of comments
> > would it make sense to store this contiguously (with a torch.stack around it)?
> 
> I wasn't sure what to do here, so I just wrapped `interpolated_actions` around a torch.stack and returned it.
> 
> Added `action_key = "_action"` as a parameter in the constructor, and now, will have to change the docstrings as well Is the action correct, or you wanted a different word, or a different way of doing it?
> 
> > Thanks so much for this!
> > We need to add a couple of tests for this feature.
> > What should happen at reset time, if there is more than one env being run?
> 
> Here, Should I create a `action_interp_buffer` (let's call it that) for each environment, and like reset when that environment is called? But then how do I get the envirnoment like ID so as to know which one to reset?

I'd make a step back and define precisely what we expect this to do.
What I understand we want from that feature is to have a stack of interpolated actions, for instance:
- if action at `t=0` is `act = [1]` and action at `t=1` with `frame_skip=4` is `act=[5]` we want the interpolated action to be `[[1], [2], [3], [4], [5]]`, is that right?
 - Maybe people will want to have the interpolated action _and_ the action, so we should be able to pass the interpolated action as a separate key. Something like `action = [5]` and `interpolated_action = [[1], [2], [3], [4], [5]]`? I think overwriting the `"action"` entry by default is dangerous because the action you will see out of the transformed environment will be incompatible with the base environment, so the default should be a separate entry for the interpolated action. If people want, they can use the `"action"` key name in the constructor but that will be subject to bugs (from which the user is responsible since this is not the default). 
 - The way we do the interpolation (in a buffer or in the tensordict) seems accessory for now, I would first code the feature and all the test, then we can change the inner mechanism (I'll be happy to help do not worry).
 - Do we allow any action type? Eg, categorical or one-hot? If not we should check that the `action_spec` of the parent env (`transform.parent.full_action_spec[transform.action_key]`) has a compatible type (ie, if of dtype float or smth like that).
 - To me the top-1 priority here is to write a series of tests. To do that, go in `test_transforms.py` and add one or more tests in `TestFrameSkipTransform`. Check that the resulting action has the right values by inputting two actions at two consecutive steps and check that the values stricly match what you're expecting. That way we will know that any change we make works within the bounds of what we want to do. Also test what happens at reset time. Let's work with a single environment for now (no batched envs) and we'll adapt that to batched envs later.

Zekrom-7780:
@vmoens , I read your comments, and you mention that my first priority should be adding tests

 So I looked at the `test_transforms.py` , and it has like 9000ish lines of code!!
Where should I like write my tests?
This file made me feel extremely overwhelmed!!

vmoens:
Ahah I can imagine!
There's a TestFrameSkipTransform with all the frame-skip related stuff. You just need a couple of functions that test that if you create a frame stkip transform the results are what is expected

Zekrom-7780:
Thanks a lot, will push some tests later, most probably in 24ish hours

gtnv:
hi @Zekrom-7780, do you still plan to work on this? if not, can we close it?

PR #1700, opened 2023-11-15
[BugFix] make sure the params of exploration-wrapper is float

FrankTianTT:
## Description

we should make sure the params of exploration-wrapper is float, or the params decay maybe abnormal.

## Motivation and Context

If the user passes an int into the wrapper (e.g. "1" instead of "1.0", like stupid me), the params will always be `0` and don't change throughout the `step` is called:
```python
 policy = TensorDictModule(
      module=torch.nn.Linear(4, 4, bias=False),
      in_keys="observation",
      out_keys="action",
  )
  p = OrnsteinUhlenbeckProcessWrapper(
      policy,
      eps_init=1,
  )
  for i in range(10):
      p.step()
      print(p.eps.item())
```

result is 
```
0
0
0
0
0
0
0
0
0
0
```

if we pass a float, the function of wrapper is normal:
 
```python
 policy = TensorDictModule(
      module=torch.nn.Linear(4, 4, bias=False),
      in_keys="observation",
      out_keys="action",
  )
  p = OrnsteinUhlenbeckProcessWrapper(
      policy,
      eps_init=1.0,
  )
  for i in range(10):
      p.step()
      print(p.eps.item())
```

result is 
```
0.9991000294685364
0.9982000589370728
0.9973000884056091
0.9964001178741455
0.9955001473426819
0.9946001768112183
0.9937002062797546
0.992800235748291
0.9919002652168274
0.9910002946853638
```

the reason of this bug is we use the type of `eps_init` as the type of `self.eps`.

## Types of changes

## Checklist

### DIFF
--- a/test/test_exploration.py
+++ b/test/test_exploration.py
@@ -51,7 +51,7 @@ from torchrl.modules.tensordict_module.exploration import (
 
 
 class TestEGreedy:
-    @pytest.mark.parametrize("eps_init", [0.0, 0.5, 1.0])
+    @pytest.mark.parametrize("eps_init", [0.0, 0.5, 1])
     @pytest.mark.parametrize("module", [True, False])
     def test_egreedy(self, eps_init, module):
         torch.manual_seed(0)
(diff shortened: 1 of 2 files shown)


vmoens (approved):
Fantastic thanks a million

PR #1710, opened 2023-11-23
[Docs] Pointers to BenchMARL

matteobettini:


vmoens (commented):
I think we should avoid *We*, because we don't want to suggest that we're building an ecosystem (like other foundations...)
We should also avoid giving sense of temporality like "we have just done this" because we'll forget about that and 2y from now it will still look like we have "just" done something.

For the Ecosystem part, I would wait until github comes back to us and shows how we can display the dependency graph. 
Gymnasium used to have a list of useful resources and such, and they took it off. IMO it was because everyone wanted to be in it, and eventually there wasn't a clear-cut rule on who should or should not have the right to be in it. 
We already have a bunch of users, and they could all legitimately ask to be in the list. 
What rule of thumb would you suggest?

matteobettini:
Got it.

I have adapted it with this feedback

vmoens (approved):
LGTM

PR #1718, opened 2023-11-28
[BugFix] `RewardSum` key check

matteobettini:
In the `RewardSum` transform, reset keys where checked to be valid only when not directly provided by the user.

https://github.com/pytorch/rl/blob/07fcfb1cff2e897fff406ef45c2e5ff7ea57a14f/torchrl/envs/transforms/transforms.py#L4773

This led to silent errors when the users provides reset keys manually but there are not as many as the in_keys gathered automatically from the reward_spec.

This PR runs the key check in both the case of user provided or autoimatically gathered reset keys, making the function more resilient.

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -9,6 +9,7 @@ import importlib.util
 
 import itertools
 import pickle
+import re
 import sys
 from copy import copy
 from functools import partial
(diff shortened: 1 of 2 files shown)


vmoens:
Is this needed?

If we look at the message in the error, we specifically tell the user that we cannot retrieve them automatically and hence the user should pass them.
If we pass them manually, the check should be different, e.g., we could just test that there are as many in_keys as reset_keys.

matteobettini:
I think having both checks is useful and at worst it cannot hurt.
The second part of the check will help preventing the user to pass "("smth","_reset") when the in_key is "reward"

The checks are run just once

vmoens:
The current check makes sure that the root of in_keys has a similar root for resets. If resets are passed manually, they could be located anywhere IMO, not at the sample place as in_keys. I guess we want manual resets to be as flexible as possible. If we restrict them as we do for automatically gathered reset entries, what do we gain with the manual version?

matteobettini:
I thought the manual version was just for when the auto does not work, but i see the point. 
I ll update to just check the length of the key list in both cases and what to i want is the zip silently failing due to diff lenghts (which just happened to me)

vmoens:
At this point we need a better set of tests for these error messages! It isn't easy to make sure we get them right just by looking at the code

matteobettini:
I have added a test that would fail on main that checks all the errors related to the lenght of keys in rewardsum

vmoens:
What do you mean "that would fail on main"? In the sense that they test the new feature or that the feature is bc breaking?

vmoens (approved):
LGTM thanks

matteobettini:
> What do you mean "that would fail on main"? In the sense that they test the new feature or that the feature is bc breaking?

They test that where before there was a silent error, now there an explicit exeption.

PR #1740, opened 2023-12-07
[BugFix, Feature] Vmap randomness in losses

BY571:
## Description

RuntimeError running off-policy examples with dropout.  

```
RuntimeError: vmap: called random operation while in randomness error mode. Please either use the 'same' or 'different' randomness flags on vmap or perform the randomness operation out of vmap
```

Propose setting randomness flag to `randomness="different"` and not `same` as this would be the natural case running over the networks sequentially. Also, we want to have different randomness values in for example the ensemble Q-Networks.  

## Motivation and Context

## Types of changes

## Checklist


BY571:
> Maybe we could make this optional? I'm assuming that if this isn't the default for vmap it is because it comes with a drawback, we should investigate


I found this on functorch:
```
The flag can only be passed to vmap and can take on 3 values, “error,” “different,” or “same,” defaulting to error. Under “error” mode, any call to a random function will produce an error asking the user to use one of the other two flags based on their use case.
```
We could make it an additional flag for the objective being error as default like `vmap_randomness="error"`. But not sure, it's only important when using the objective with networks that utilize dropout.

BY571:
Looks good to me, I will adapt one objective class and then we can recheck if its what we expect. 

I wonder if at some point we can put together all those general functions to a higher level objective class lets say OffpolicyObjective from which the specific objectives inherit. Such a higher level class would handle the vmap and also the make_value_estimator etc. Because right now the objective classes get bigger and bigger probably hard to read and understand for people who just want to see the td3/sac loss calculation.

BY571:
Turns out that self.modules does not include modules like Dropout. I updated the function accordingly using actor and qvalue modules directly like:

```
    @property
    def vmap_randomness(self):
        if self._vmap_randomness is None:
            # look for nn.Dropout modules
            dropouts = (torch.nn.Dropout, torch.nn.Dropout2d, torch.nn.Dropout3d)
            for a, q in zip(
                self.actor_network.modules(), self.qvalue_network.modules()
            ):
                if isinstance(a, dropouts) or isinstance(q, dropouts):
                    self._vmap_randomness = "different"
                    break
            else:
                self._vmap_randomness = "error"

        return self._vmap_randomness
```

let me know what you think :)

vmoens:
> OffpolicyObjective

In general we can think about it but if there's a clear use case for it, not just some vmap compatibility. Plus I'm working on consistent dropout for RL, which works also on the online setting...

vmoens:
> ```
>     @property
>     def vmap_randomness(self):
>         if self._vmap_randomness is None:
>             # look for nn.Dropout modules
>             dropouts = (torch.nn.Dropout, torch.nn.Dropout2d, torch.nn.Dropout3d)
>             for a, q in zip(
>                 self.actor_network.modules(), self.qvalue_network.modules()
>             ):
>                 if isinstance(a, dropouts) or isinstance(q, dropouts):
>                     self._vmap_randomness = "different"
>                     break
>             else:
>                 self._vmap_randomness = "error"
> 
>         return self._vmap_randomness
> ```

Ah right bc we don't want the modules to appear in the `modules` list actually! 
The problem with this is that it breaks as soon as you name your `actor_network` something like `actor`.
Also why do we consider that actor and qvalue networks have the same number of modules? I don't think we should use zip, we should iterate over one and then the other. Something like
```python
do_break = False
for val in self.__dict__.values():
    if isinstance(val, nn.Module):
        for module in val.modules():
             if isinstance(module, RANDOM_MODULE_LIST): # not only nn.Dropout is random, could be something else
                 self._vmap_randomness = "different"
                 do_break = True
                 break
    if do_break:
        # double break
        break
else:
    self._vmap_randomness = "error"

```

([one guy](https://stackoverflow.com/a/4553525) claims that we should break nested loops with exception but I'd rather be dead in a ditch than doing that)

BY571:
> Also why do we consider that actor and qvalue networks have the same number of modules? I don't think we should use zip, we should iterate over one and then the other.

You are right, we shouldn't expect them to be the same. I'll update it!

torchrl/objectives/td3.py
```diff
@@ -26,6 +26,25 @@
 )
 from torchrl.objectives.value import TD0Estimator, TD1Estimator, TDLambdaEstimator
 
+RANDOM_MODULE_LIST = (
```
BY571, line 29: I expanded the list quite a bit to cover more modules. What do you think about how we should handle it? Importing the list from torchrl/objectives/utils.py?

torchrl/objectives/cql.py
```diff
@@ -335,8 +335,12 @@ def __init__(
...
             )
 
-        self._vmap_qvalue_networkN0 = _vmap_func(self.qvalue_network, (None, 0))
-        self._vmap_qvalue_network00 = _vmap_func(self.qvalue_network)
+        self._vmap_qvalue_networkN0 = _vmap_func(
+            self.qvalue_network, (None, 0), randomness="different"
+        )
+        self._vmap_qvalue_network00 = _vmap_func(
+            self.qvalue_network, randomness="different"
+        )
```
vmoens, line 343: why not using the class' attribute?

torchrl/objectives/iql.py
```diff
@@ -286,7 +288,9 @@ def __init__(
         if gamma is not None:
             warnings.warn(_GAMMA_LMBDA_DEPREC_WARNING, category=DeprecationWarning)
             self.gamma = gamma
-        self._vmap_qvalue_networkN0 = _vmap_func(self.qvalue_network, (None, 0))
+        self._vmap_qvalue_networkN0 = _vmap_func(
+            self.qvalue_network, (None, 0), randomness="different"
+        )
```
vmoens, line 293: ditto

torchrl/objectives/redq.py
```diff
@@ -234,6 +235,7 @@ class _AcceptedKeys:
         "next.state_value",
         "target_value",
     ]
+    _vmap_randomness = None
```
vmoens, line 238: ditto

torchrl/objectives/redq.py
```diff
@@ -404,6 +409,28 @@ def in_keys(self):
...
+                    # double break
+                    break
+            else:
+                self._vmap_randomness = "error"
+
+        return self._vmap_randomness
+
+    def set_vmap_randomness(self, value):
+        self._vmap_randomness = value
+
```
vmoens, line 433: why do we need this in multiple places and not in the parent class?

torchrl/objectives/sac.py
```diff
@@ -530,6 +536,28 @@ def out_keys(self):
...
+                    # double break
+                    break
+            else:
+                self._vmap_randomness = "error"
+
+        return self._vmap_randomness
+
+    def set_vmap_randomness(self, value):
+        self._vmap_randomness = value
+
```
vmoens, line 560: ditto

torchrl/objectives/sac.py
```diff
@@ -1083,6 +1114,28 @@ def in_keys(self):
...
```
vmoens, line 1138: ditto

torchrl/objectives/td3.py
```diff
@@ -343,6 +348,28 @@ def _cached_stack_actor_params(self):
...
+                    # double break
+                    break
+            else:
+                self._vmap_randomness = "error"
+
+        return self._vmap_randomness
+
+    def set_vmap_randomness(self, value):
+        self._vmap_randomness = value
+
```
vmoens, line 372: ditto

vmoens (commented):
Thanks for this, I think we can drastically reduce the amount of code in this PR by reusing the same methods across classes.
Some tests would be helpful!

BY571:
> Some tests would be helpful!

Any Idea regarding how we could test it effectively? Adding dropout as an option to the mock actors/critics in the tests for the objectives is an option but this would mean adding it for all of the objectives independently again. Not sure if ideal...

vmoens:
@BY571 
You can adapt this:
```python
from torchrl.objectives import LossModule
import torch
from torchrl.objectives.utils import _vmap_func
from tensordict.nn import TensorDictModule as Mod
from torch import nn

def test_loss_vmap_random():
    class MyLoss(LossModule):
        def __init__(self):
            super().__init__()
            mod = Mod(nn.Dropout(0.1), in_keys=["obs"], out_keys=["action"])
            self.convert_to_functional(mod, "mod", expand_dim=4)
            self.vmap_mod = _vmap_func(self.mod, (None, 0))

        def forward(self, td):
            out = self.vmap_mod(td, self.mod_params)
            return {"loss": out["action"].mean()}
    loss_mod = MyLoss()
    td = TensorDict({"obs": torch.randn(3, 4)}, [3])
    loss_mod(td)
test_loss_vmap_random()
```
Please clean up the code before putting it in ;)

test/test_cost.py
```diff
@@ -233,6 +234,38 @@ def set_advantage_keys_through_loss_test(
             )
 
 
+@pytest.mark.parametrize("device", get_default_devices())
```
BY571, line 237: Let me know what you think! :)

test/test_cost.py
```diff
@@ -233,6 +234,38 @@ def set_advantage_keys_through_loss_test(
...
+            return {"loss": out["action"].mean()}
+
+    loss_module = VmapTestLoss()
+    td = TensorDict({"obs": torch.randn(3, 4).to(device)}, [3])
+
+    # If user sets vmap randomness to a specific value
+    if vmap_randomness in ("different", "same") and dropout > 0.0:
+        loss_module.set_vmap_randomness(vmap_randomness)
+
+    loss_module(td)["loss"]
```
vmoens, line 266: maybe let's test that things actually fail if we don't call the `set_vmap_randomness` before?

torchrl/objectives/utils.py
```diff
@@ -29,6 +30,14 @@
     "run `loss_module.make_value_estimator(ValueEstimators.<value_fun>, gamma=val)`."
 )
 
+RANDOM_MODULE_LIST = (
+    nn.Dropout,
+    nn.Dropout2d,
+    nn.Dropout3d,
```
vmoens, line 36: Do these guys have a parent, common class?

vmoens (commented):
Almost there, just a couple of minor edits!
Thanks so much

test/test_cost.py
BY571: If we don't do `loss_module.set_vmap_randomness(vmap_randomness) `and have a Module that uses randomness vmap_randomness sets default to "different". So its only if the user wants a specific vmap_randomness. I think there is no case in which we should expect an error, only if the user sets vmap_randomness manually to "error" and uses dropout for example. 

I can add a test for that but not sure if thats what you meant.

vmoens: > only if the user sets vmap_randomness manually to "error" and uses dropout for example

Yes that is what I meant. Here we only test that the code runs, but we're not really checking that it would have been broken had we done things differently.

torchrl/objectives/utils.py
BY571: Indeed, all the Dropouts have a common parent _DropoutNd. Ill update it!

torchrl/objectives/utils.py
```diff
@@ -29,6 +31,8 @@
     "run `loss_module.make_value_estimator(ValueEstimators.<value_fun>, gamma=val)`."
 )
 
+RANDOM_MODULE_LIST = (dropout._DropoutNd,)
```
BY571, line 34: Should we keep it a list in case we are going to extend it in the future?

test/test_cost.py
BY571: added it!

test/test_cost.py
```diff
@@ -262,8 +266,14 @@ def forward(self, td):
     # If user sets vmap randomness to a specific value
     if vmap_randomness in ("different", "same") and dropout > 0.0:
         loss_module.set_vmap_randomness(vmap_randomness)
+        loss_module(td)["loss"]
+    # Fail case
+    elif vmap_randomness == "error" and dropout > 0.0:
+        with pytest.raises(RuntimeError):
```
vmoens, line 272: let's check that the error message is what is expected with `match=`

test/test_cost.py
```diff
@@ -262,8 +266,14 @@ def forward(self, td):
...
+            loss_module(td)["loss"]
 
-    loss_module(td)["loss"]
+    else:
+        loss_module(td)["loss"]
```
vmoens, line 276:
```python
    if vmap_randomness in ("different", "same") and dropout > 0.0:
        loss_module.set_vmap_randomness(vmap_randomness)
    # Fail case
    elif vmap_randomness == "error" and dropout > 0.0:
        with pytest.raises(RuntimeError):
            loss_module(td)["loss"]
        return
    loss_module(td)["loss"]
```

vmoens (approved):
LGTM

PR #1746, opened 2023-12-14
[CI] Fix windows CI

vmoens:


atalman:
@vmoens the torch pip install on windows is fixed. Please take a look at the next issue: https://github.com/pytorch/rl/actions/runs/7211403586/job/19652490509?pr=1746#step:9:578

Issue #1671, opened 2023-11-03
[BUG] Saving episodes with variable steps in the replay buffer


truncs:
In an env when the episode terminates early, I would like to still save in the buffer and fixed size episodes from it. This currently fails for me giving the error

```
RuntimeError: indexed destination TensorDict batch size is torch.Size([1001]) (batch_size = torch.Size([1000, 1001]), index=5), which differs from the source batch size torch.Size([40])
```

## To Reproduce
The actual trace for my env is below
```
Traceback (most recent call last):
  File "/env/lib/python3.11/site-packages/tensordict/tensordict.py", line 3754, in __setitem__
    value = value.expand(indexed_bs)
            ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/env/lib/python3.11/site-packages/tensordict/tensordict.py", line 4430, in expand
    raise RuntimeError(
RuntimeError: Incompatible expanded shape: The expanded shape length at non-singleton dimension should be same as the original length. target_shape = (1001,), existing_shape = torch.Size([40])

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "train.py", line 56, in train
    trainer.train()
  File "trainer/online_trainer.py", line 93, in train
    self._ep_idx = self.buffer.add(torch.cat(self._tds))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "buffer.py", line 99, in add
    self._buffer.add(tds)
  File "/env/lib/python3.11/site-packages/torchrl/data/replay_buffers/replay_buffers.py", line 255, in add
    return self._add(data)
           ^^^^^^^^^^^^^^^
  File "/env/lib/python3.11/site-packages/torchrl/data/replay_buffers/replay_buffers.py", line 259, in _add
    index = self._writer.add(data)
            ^^^^^^^^^^^^^^^^^^^^^^
  File "/env/lib/python3.11/site-packages/torchrl/data/replay_buffers/writers.py", line 55, in add
    self._storage[self._cursor] = data
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/zfs/aditya/workspace/tdmpc2/env/lib/python3.11/site-packages/torchrl/data/replay_buffers/storages.py", line 72, in __setitem__
    ret = self.set(index, value)
          ^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #1079, opened 2023-04-24
[Feature Request] Ability to skip transitions based on masks


Mayankm96:
## Motivation

Certain simulators (such as Isaac Gym and Isaac Sim) create several environment instances as part of the same stage. Unlike CPU-based vectorization, where each environment instance, has its own stepping, in GPU-based parallelized environments, the stepping is common. This leads to an issue when certain environments want to have dummy steps for the scene to come to rest. For instance, we often spawn deformable bodies (such as cloths) from a height and let them fall onto the ground and come to rest before we start learning.

There isn't a clear mechanism to have these dummy steps in the simulators and probably this functionality won't be added since it ties deeply into the whole parallelization philosophy (i.e. all environments are just a huge single GPU data). Thus, this issue would be best handled by the learning framework itself.

## Solution

A possible way to deal with this issue could be having the ability to add masking of transition tuples. For instance, if out of N environments, (2, 5, 10) indices were last reset, we enable their masking (additional quantity to "infos" dictionary) for certain time steps. If the masks evaluate to true, then the trajectory or replay buffer doesn't add transitions for those indices into the memory. 

## Alternatives

I couldn't really think of any other alternatives here but would be interested to hear your thoughts on this.

## Additional context

* Proposed in [Orbit #64](https://github.com/NVIDIA-Omniverse/Orbit/discussions/64)

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

vmoens:
Thanks for posting this issue, certainly worth having a discussion.
@matteobettini @btx0424 thoughts?

So if I understand correctly, the main issue is that some steps during the execution of simulations with multiple envs may not be worth storing in the replay buffer. Is there anything more general than that?
In the sense: one solution here could be to design a transform for the replay buffer that simply filters out data based on a boolean mask stored in the output tensordict.

If it's more general than that we can also look for a more consistent change in the API.

matteobettini:
Yeah it seems like the env should output this mask key and the buffer could have a transform that masks out data given a str key.

PS @vmoens another example of a util key output by the env which will go in obs_spec.

btx0424:
I agree that if the number of dummy steps is not significant, filtering them out afterwards is a reasonable solution.

But more generally I think this is related to the structure of the buffers in torchrl.

The first type is the rollout buffer allocated by a `Collector`, it has fixed batch (e.g., num of envs) and time dimensions. No saving is possible here, with or without the mask.

[...]

PR #1748, opened 2023-12-15
[Feature] SliceSampler

vmoens:
In this PR I propose a new sampler that samples slices of data in a storage.

## Setting

Let RB be a replay buffer of capacity `M` (for simplicity the replay buffer is full). 
We have `N` trajectories stored in this buffer, with length `T_{min} < T_i < T_{max}` for all `T_i for i in 1:N`.
We want to sample batches of data of some arbitrary length such that the batch size `B = S * U` (potentially with a tensordict of shape `[S, U]` or `[S * U]`, TBD) where `S` is the number of slices and `U` a fixed sub-trajectory length.
Assume `U > T_{min}`.
The slices sampled from the RB can start anywhere from `t_start=0` to `t_start = T_i - U` and end at `t_start + U`.

## Usage

In many cases, users want to sample sub-trajectories out of complete trajectories. Those sub-trajectories should generally not start at the beginning of the trajectory or end at the end (hence, we say that users want _slices_ of trajectories). Typical usages are:
- RNN/Transformer-based policies
- Offline RL datasets

## Proposed class

The `SliceSampler` we propose here does exactly what was suggested above.

To keep the batch-size unchanged, the user has to specify how many slices she wants out of the batch-size (`S`). Alternatively, we could ask for a trajectory length (`U`) (or one of these but not both).
Eg,
```
class SliceSampler(Sampler):
    def __init__(self, ..., num_slices=None, traj_len=None): # either traj_len or num_slices must be non-none, but not both
       ...
```

We can sample trajectories when we know the start and stop signals. To gather these, users can pass either the key pointing to the trajectory indicators `traj_key` or a done state `done_key`. 

**Note**: passing `"done"` can be dangerous since if one gets an incomplete trajectory, it will not be marked as `"done"` at the last step! Using the `"truncated"` signal from the collector could be an alternative, but in general it should be safer to use the trajectory indicator. The user is responsible of making sure that two consecutive trajectories do not share the same indicator.

Computing the start and stop signals can be expensive, and for static datasets one could cache the start and stop signals. This speeds up sampling by an important factor (0.3 ms with caching vs 1ms on my computer)

Example code:
```python
import timeit
import torch
import tqdm

from torchrl.data.replay_buffers import LazyMemmapStorage, \
    TensorDictReplayBuffer
from torchrl.data.replay_buffers.samplers import SliceSampler

prefetch = 0
for cache in [True, False]:
    rb = TensorDictReplayBuffer(
        storage=LazyMemmapStorage(1_000_000),
        sampler=SliceSampler(cache_values=cache, num_slices=10),
        batch_size=320,
        prefetch=prefetch
        )
    episode = torch.zeros(1000, dtype=torch.int)
    episode[:300] = 1
    episode[300:550] = 2
    episode[550:700] = 3
    episode[700:] = 4
    data = TensorDict(
        {
            "episode": episode,
            "obs": torch.randn((3, 4, 5)).expand(1000, 3, 4, 5),
            "act": torch.randn((20,)).expand(1000, 20),
            "other": torch.randn((20, 50)).expand(1000, 20, 50),
        }, [1000]
    )
    for _ in tqdm.tqdm(range(1000)):
        rb.extend(data)

    print(prefetch, min(*timeit.repeat("rb.sample()", globals={"rb": rb}, number=1000)))

```

Results in `0.3468737329999989` ms with caching, `1.150314997999999` without

## Why is this PR important

This PR paves the way for a new API to store trajectories, where one can flatten the data coming from the collector and extend the RB with it. Storing trajectories of different lengths will be easy, and sampling trajectories should be cheaper. This will also participate in making the dataset and dynamic RB APIs uniform.
Pseudocode:
```python
for data in collector: # data is of shape [C, D]
    replay_buffer.extend(data.view(-1)) # adds the C consecutive trajectories of length D
    samples = replay_buffer.sample(B) # B elements divided in S slices of length U
    samples = samples.view(S, U)
```
Ultimately, we will be able to deprecate the hacky `"_data"` key in `TensorDictReplayBuffers` that is bothering us so much!

cc: @matteobettini @albertbou92 @BY571 @skandermoalla @nicklashansen @truncs @btx0424 @smorad @vikashplus

## Question

- Should `batch_size` indicate the number of slices (ie, trajectories) sampled? In other words: Should users ask for `B` or `S`. Pros for `B`: clearer API, I ask for `B` transitions I get `B` transitions, whatever the sampler. Pros for `S`: users may ask for a certain "number of items" referring items to the trajectories. I'm still in favour of `B` because otherwise we assume that the sampler impacts the shape of the samples, which isn't great for composability (swapping samples will change the amount of data consumed, or UTD of the algo).
- Related: Should we return batches of size `[S, U]` or `[S * U]`? I like the second best for the same reason as before (composability).
- Last one: what do we do if some trajectories have an insufficient length (T_min > U)? My take: there should be an arg `pad` in the constructor, `None` by default. If `pad=True`, we pad with 0s. If `pad=smth` and `smth` is not `None`, we pas with `smth`. If `pad=None` (default) an exception is raised if `T_min < U` (or cheaper option: if `T_i < U` is encountered at some point during training.
- Should we give a device to the sampler to speed up the trajectory recovering (with cuda)?

Related issues
https://github.com/pytorch/rl/issues/1671
https://github.com/pytorch/rl/issues/1079

TODO:
- [x] Doc
- [x] Tests
- [x] Allow padding => @matteobettini Samplers cannot do padding (they just return indices) so we just return fewer indices if trajectories don't comply with minimum length whenever `strict_length=False`.
- [x] allow sampling without replacement (what does that look like?)
- [ ] sampling each transition with equal probability? => NOT PLANNED
- [ ] prepare ordered dataset iterations => follow-up PR

### DIFF
--- a/docs/source/reference/data.rst
+++ b/docs/source/reference/data.rst
@@ -34,8 +34,9 @@ We also give users the ability to compose a replay buffer using the following co
     Sampler
     PrioritizedSampler
     RandomSampler
-    SliceSampler
     SamplerWithoutReplacement
+    SliceSampler
+    SliceSamplerWithoutReplacement
     Storage
     ListStorage
     LazyTensorStorage
--- a/docs/source/reference/data.rst
+++ b/docs/source/reference/data.rst
@@ -34,6 +34,7 @@ We also give users the ability to compose a replay buffer using the following co
     Sampler
     PrioritizedSampler
     RandomSampler
+    SliceSampler
     SamplerWithoutReplacement
     Storage
     ListStorage
@@ -109,41 +110,59 @@ It is not too difficult to store trajectories in the replay buffer.
 One element to pay attention to is that the size of the replay buffer is always
 the size of the leading dimension of the storage: in other words, creating a
 replay buffer with a storage of size 1M when storing multidimensional data
-does not mean storing 1M frames but 1M trajectories.
+does not mean storing 1M frames but 1M trajectories. However, if trajectories
+(or episodes/rollouts) are flattened before being stored, the capacity will still
+be 1M steps.
 
 When sampling trajectories, it may be desirable to sample sub-trajectories
 to diversify learning or make the sampling more efficient.
-To do this, we provide a custom :class:`~torchrl.envs.Transform` class named
-:class:`~torchrl.envs.RandomCropTensorDict`. Here is an example of how this class
-can be used:
+TorchRL offers two distinctive ways of accomplishing this:
+- The :class:`~torchrl.data.replay_buffers.samplers.SliceSampler` allows to
+  sample a given number of slices of trajectories stored one after another
+  along the leading dimension of the :class:`~torchrl.data.replay_buffers.samplers.TensorStorage`.
+  This is the recommended way of sampling sub-trajectories in TorchRL __especially__
+  when using offline datasets (which are stored using that convention).
+  This strategy requires to flatten the trajectories before extending the replay
+  buffer and reshaping them after sampling. The :class:`~torchrl.data.replay_buffers.samplers.SliceSampler`
+  gives extensive details about this storage and sampling strategy.
 
-.. code-block::Python
+- Trajectories can also be stored independently, with the each element of the
+  leading dimension pointing to a different trajectory. This requires
+  for the trajectories to have a congruent shape (or to be padded).
+  We provide a custom :class:`~torchrl.envs.Transform` class named
+  :class:`~torchrl.envs.RandomCropTensorDict` that allows to sample
+  sub-trajectories in the buffer. Note that, unlike the :class:`~torchrl.data.replay_buffers.samplers.SliceSampler`-based
+  strategy, here having an ``"episode"`` or ``"done"`` key pointing at the
+  start and stop signals isn't required.
+  Here is an example of how this class can be used:
 
-    >>> import torch
-    >>> from tensordict import TensorDict
-    >>> from torchrl.data import LazyMemmapStorage, TensorDictReplayBuffer
-    >>> from torchrl.envs import RandomCropTensorDict
-    >>>
-    >>> obs = torch.randn(100, 50, 1)
-    >>> data = TensorDict({"obs": obs[:-1], "next": {"obs": obs[1:]}}, [99])
-    >>> rb = TensorDictReplayBuffer(storage=LazyMemmapStorage(1000))
-    >>> rb.extend(data)
-    >>> # subsample trajectories of length 10
-    >>> rb.append_transform(RandomCropTensorDict(sub_seq_len=10))
-    >>> print(rb.sample(128))
-    TensorDict(
-        fields={
-            index: Tensor(shape=torch.Size([10]), device=cpu, dtype=torch.int32, is_shared=False),
-            next: TensorDict(
-                fields={
-                    obs: Tensor(shape=torch.Size([10, 50, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
-                batch_size=torch.Size([10]),
-                device=None,
-                is_shared=False),
-            obs: Tensor(shape=torch.Size([10, 50, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
-        batch_size=torch.Size([10]),
-        device=None,
-        is_shared=False)
+  .. code-block::Python
+
+      >>> import torch
+      >>> from tensordict import TensorDict
+      >>> from torchrl.data import LazyMemmapStorage, TensorDictReplayBuffer
+      >>> from torchrl.envs import RandomCropTensorDict
+      >>>
+      >>> obs = torch.randn(100, 50, 1)
+      >>> data = TensorDict({"obs": obs[:-1], "next": {"obs": obs[1:]}}, [99])
+      >>> rb = TensorDictReplayBuffer(storage=LazyMemmapStorage(1000))
+      >>> rb.extend(data)
+      >>> # subsample trajectories of length 10
+      >>> rb.append_transform(RandomCropTensorDict(sub_seq_len=10))
+      >>> print(rb.sample(128))
+      TensorDict(
+          fields={
+              index: Tensor(shape=torch.Size([10]), device=cpu, dtype=torch.int32, is_shared=False),
+              next: TensorDict(
+                  fields={
+                      obs: Tensor(shape=torch.Size([10, 50, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
+                  batch_size=torch.Size([10]),
+                  device=None,
+                  is_shared=False),
+              obs: Tensor(shape=torch.Size([10, 50, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
+          batch_size=torch.Size([10]),
+          device=None,
+          is_shared=False)
 
 Checkpointing Replay Buffers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -1646,10 +1646,7 @@ class TestSamplers:
             assert too_short
 
         assert len(trajs_unique_id) == 4
-        if _data_prefix:
-            truncated = info[("_data", ("next", "truncated"))]
-        else:
-            truncated = info[("next", "truncated")]
+        truncated = info[("next", "truncated")]
         assert truncated.view(num_slices, -1)[:, -1].all()
 
     def test_slice_sampler_errors(self):
@@ -1773,10 +1770,7 @@ class TestSamplers:
             trajs_unique_id = trajs_unique_id.union(
                 cur_episodes,
             )
-        if _data_prefix:
-            truncated = info[("_data", ("next", "truncated"))]
-        else:
-            truncated = info[("next", "truncated")]
+        truncated = info[("next", "truncated")]
         assert truncated.view(num_slices, -1)[:, -1].all()
 
 
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -1596,13 +1596,14 @@ class TestSamplers:
         sampler = SliceSampler(
             num_slices=num_slices, traj_key=episode_key, end_key=done_key
         )
-        index, _ = sampler.sample(storage, batch_size=batch_size)
-        if _data_prefix:
-            samples = storage._storage["_data"][index]
-        else:
-            samples = storage._storage[index]
         trajs_unique_id = set()
         for _ in range(5):
+            index, info = sampler.sample(storage, batch_size=batch_size)
+            if _data_prefix:
+                samples = storage._storage["_data"][index]
+            else:
+                samples = storage._storage[index]
+
             # check that trajs are ok
             samples = samples.view(num_slices, -1)
             assert samples["another_episode"].unique(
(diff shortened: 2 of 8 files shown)


truncs:
This is great! Some thoughts - 

1. I think sampling in just trajectories would be great but I can see how some API users would like to have more control of how to convert the chunks to trajectories.
2. Trajectories of insufficient chunk length should be padded up to the chunk size and saved IMHO. Another option could be to let API users pass a function to handle this case where they can decide if they want to throw an exception or just pad the insufficient chunks.

skandermoalla:
I'm not sure to understand the use case of this. Could you put in more context?

It seems useful to methods that use full episodes.


> Should batch_size indicate the number of slices (ie, trajectories) sampled?

Yes, it seems a bit odd and error prone to have hald the information in the slice variable and the other half in the batch size.

> Related: Should we return batches of size [M, N] where M is the number of slices and N the trajectory lengths?

This would allow to have a determinitic batch size, but would be lots of zeros. I would prefer a determinitic batch size by default, and if there's a way to save memory on the tail it'd be great.


> Last one: what do we do if some trajectories have an insufficient length? Should they be considered (and possibly padded)? IMO we should just return an exception or a warning but i'm open to other suggestions.

What does insuffient lenght mean here?
It seems like sampling and giving a mask to user would be the easiest.


> Should we give a device to the transform to speed up the trajectory recovering (with cuda)?

The transform?

vmoens:
> I'm not sure to understand the use case of this. Could you put in more context?

Sure!
Currently, when people need to sample trajectories, we advise them to build RBs storages with shape [N, T], but that's clunky because it means that if I tell you "build a RB of 1M elements" you have to divide 1M by the traj length.
Also it means that you have to pad or use trajs that all have the same length. On top of that, getting sub-trajectories is hard and inefficient... Finally, because in general trajectory (episode) lengths cannot be assumed to be identical so we store them as [N x T], but people will eventually want to sample some trajectories or some parts of trajectories...

> It seems useful to methods that use full episodes.
> 
> > Should batch_size indicate the number of slices (ie, trajectories) sampled?
> 
> Yes, it seems a bit odd and error prone to have hald the information in the slice variable and the other half in the batch size.

On the other hand, you will get more elements than you asked for! Could be surprising too.

> > Related: Should we return batches of size [M, N] where M is the number of slices and N the trajectory lengths?
> 
> This would allow to have a determinitic batch size, but would be lots of zeros. I would prefer a determinitic batch size by default, and if there's a way to save memory on the tail it'd be great.

Why lots of zeroes? What I meant is: should this reshape happen within the transform?
https://github.com/pytorch/rl/blob/0d7631a1c6a86625d0d95d0dc1b18c7f9245a712/torchrl/data/replay_buffers/samplers.py#L539-L541

> > Last one: what do we do if some trajectories have an insufficient length? Should they be considered (and possibly padded)? IMO we should just return an exception or a warning but i'm open to other suggestions.
> 
> What does insuffient lenght mean here? It seems like sampling and giving a mask to user would be the easiest.

Say some trajectories in your buffer have length 10, and you ask for a batch of total 320, across 10 splits: you will want have traj length of 32, which is greater than 10 -> should we pad or jusy throw an exception?

> > Should we give a device to the transform to speed up the trajectory recovering (with cuda)?
> 
> The transform?

Sorry it's Friday, I meant the sampler!

matteobettini:
This is very cool and it is much needed to have an easy way to sample trajectories!

I am a bit confused from the decrtiption though, it seems to me that `N` and `M` are overloaded in different parts of the text

> ```python
> for data in collector: # data is of shape [N, T]
>     replay_buffer.extend(data.view(-1)) # adds the N consecutive trajectories of length T
>     samples = replay_buffer.sample(M) # M elements divided in N slices
>     samples = samples.view(N, -1)
> ```

- Why is `T` trajectory length? in normal collectors T is just the number of collected frames divided by the batch size of collection

I think we might benefit from rewriting the description with clearer symbols and a recap of the current state of affairs prior to this change

## My opinion

1. I think users should request a desired trajectory length rather than a number of slices (at construction of the sampler). this is more intuive as most people care how long the traj fragment is (leaving this as a byproduct of number_of_slices is counterintuitive to me ). Therefore I would prefer something like `TrajectorySampler(cache_values=cache,traj_length=10)`. For this reason is also suggest naming the sampler `TrajectorySampler`
2. Regarding the batch size, when users request `buffer.sample(n_samples)` I think they should get a td with batch_size `[n_slices,traj_length]` where n_slices is computed as `n_samples / traj_length` (`n_samples` and `traj_length` are the user-defined quantities and `n_slices` is the byproduct)
3. If the actual episode length is different from the `traj_length` length provided at construction (longer for some episodes, shorter for some other). We should: 
     - if shorter: pad and give a warning that some are padded
     - if longer: just cut it at  `traj_length` and put the overflowing bit in the next slice

vmoens:
Good points
I think we should raise an error if the length is shorter unless users explicitly set a `pad` arg to True, because literally everyone in RL disable warnings, and no one ever looks at the log.

vmoens:
> if longer: just cut it at traj_length and put the overflowing bit in the next slice

I don't get that

matteobettini:
> > if longer: just cut it at traj_length and put the overflowing bit in the next slice
> 
> I don't get that

If the user requests 100 samples and my buffer has one epsiode which is 100 steps long and the user asked to sample trajectory fragments of length 70, we should return 2 fragments in a td of batch size [2,70]:
 - the first fragment is the first 70 steps
 - the second fragment is the last 30 steps plus 40 padded steps (supposing pagging is reuqested)

vmoens:
That isn't the contract of this class, which is to sample randomly subsequences (this is why it isn't a trajectory sampler but a slice sampler - it samples incomplete episodes). This is also why we add a truncated signal at the end.
The implementation you're suggesting seems more constraining to me, and surprising: consecutive samples of a batch should not be correlated.
I don't see why we should complete the trajectory in the next batch, what's the advantage? If users want full trajectories of length 100 why not asking for that?
In this implementation, I could also expect a slice of length 70 without padding because all my trajectories are longer than 70, and here I'll get some trajectories that are padded. Pretty sure that will lead to issues on the repo!

vmoens:
Another issue with this is that it seems to be that it assumes that all trajectories have the same length. Say this is not the case (eg, all our datasets) and you want sub trajs of length 100. One trajectory may be 101 transitions long so you will get 2 batches, one of 100 and another padded of 1. Users will never see the second batch contiguously with the first (it will appear like this one element alone isn't part of the same trajectory). Also it's not clear what's the batch size here: 200? 101? What if users wanted only 100 transitions? What if they wanted 200, should they get samples from another trajectory too, thereby having a varying output size from the buffer?

vmoens:
I updated the description!

matteobettini:
Thanks for the update! The new descriptions clears a lot of things and the choice of leaving to the users if they want to give S or U is great imo!

I guess the point I got confused about is if sampling is handled with or without replacement.
Ideally, both on policy algos (like PPO) and off policy algos may want to use this, so maybe the replacement strategy of this sampler might be important.

What I understand from your answers is that this sampler samples with replacement (so the data of 2 slices sampled in the same call could be the same)  

I still think though I am not getting some API details of this sampler, so here are some examples that would be helpful for me to clarify:

1. from what i understand, data from different episodes will never be present in the same slice. I.e. a slice will always contain a fragment from the same episode
2. Are we sampling from trajectories independently? For example, if i have a buffer containing 2 episodes: one of length 30 and the other of length 70, and the users asks for 100 samples in 2 slices of length 50 each, can you explain what would be given to the user?
    - Are we going to sample 2 slices of 50 from the episode with length 70? 
    - Are we going to sample one slice from the episode with length 70 and one padded one from the episode with length 30?
    - Are we going to sample in a random way and pad if there are not 50 steps after the selected starting indices?

vmoens:
> What I understand from your answers is that this sampler samples with replacement (so the data of 2 slices sampled in the same call could be the same)

yes but that can be edited

> from what i understand, data from different episodes will never be present in the same slice. I.e. a slice will always contain a fragment from the same episode

yes that is the point, but we could also loosen that (RNNs and value estimators allow mixed trajectories in torchrl so that wouldn't be an issue for them)

> Are we sampling from trajectories independently? For example, if i have a buffer containing 2 episodes: one of length 30 and the other of length 70, and the users asks for 100 samples in 2 slices of length 50 each, can you explain what would be given to the user?

We sample first the episodes, then within the episodes. So each element does not have the same probabilty of ending up in the batch. Not sure how to implement the other option mathematically: perhaps sample items from the buffer and then sample a slice around these samples. That will mean that some trajectories have a higher priority than others, which is as opinionated as the first.

> if i have a buffer containing 2 episodes: one of length 30 and the other of length 70, and the users asks for 100 samples in 2 slices of length 50 each

Currently, you would have one sample of 50 from trajectory 1 and a sample of length 30 + 20 pads from trajectory 0.

vmoens:
@matteobettini What does sampling without replacement mean?
If I have a trajectory of length 101 and ask for length 100, the second time I sample from it I will have only 1 sample. This means that the samples at the end of the epoch will be degenerated. If we samples another slice that contains that element, longer trajectories will have less repetitions across samples (and it won't be truly without replacement). The only option I see would be so sample each trajectory only once (that makes more sense to me) but then we would end up with a different likelihood per transition in the buffer (which I don't think is an issue).

Remember that in some settings people allocate themselves a number of "episodes" rather than a number of "transitions", which kind of make sense: for cartpole for instance, if you solve it in 100 episodes and not 200 it's way more impressive than 1000 samples and not 2000 (since success = longer episode)

matteobettini:
> Are we sampling from trajectories independently? For example, if i have a buffer containing 2 episodes: one of length 30 and the other of length 70, and the users asks for 100 samples in 2 slices of length 50 each, can you explain what would be given to the user?
> 
> We sample first the episodes, then within the episodes. So each element does not have the same probabilty of ending up in the batch. Not sure how to implement the other option mathematically: perhaps sample items from the buffer and then sample a slice around these samples. That will mean that some trajectories have a higher priority than others, which is as opinionated as the first.
> 
> if i have a buffer containing 2 episodes: one of length 30 and the other of length 70, and the users asks for 100 samples in 2 slices of length 50 each
> 
> Currently, you would have one sample of 50 from trajectory 1 and a sample of length 30 + 20 pads from trajectory 0.
> 

Now I got it! Thanks for clarifying!

To me it was not clear that we first sample episode indeces and then slices.
So the sampling with replacement happens at 2 levels: for episodes (first) and trajecotries (second).
Or is the first sampling of episodes with replacement?

Yes I guess this has some implications. Basically each episode will be weighted equally (so if you have an episode of length 1 where the agent committed suicide in one step) this will have a dedicated slice of length U (potentially polluting the data)

The other implementation I was referring to is that you sample S indeces at random in the buffer flattened data and then you look U steps in the future from that index (if you do not have U steps in the future, you pad the remaining ones).

This implementation removes the hierarchical sampling (of episodes first and then trajs), but (as you said) gives less weight to shorter trajectories.

vmoens:
@matteobettini added `SliceSamplerWithoutReplacement`, which simply inherits from both parent classes in a rather simple way.

skandermoalla:
Sorry for being late to the party. This is really cool and much clearer now!

Raising a broader question: the description of the PR seems outdated again, so for recent features do we want to rebuild the documentation and serve it somewhere? or rely on the PR descriptions for the temporary documentation?

vmoens:
Good point I can update it! But the doc should be accurate

PR #1750, opened 2023-12-16
[BugFix] Fix discrete SAC log-prob

vmoens:
cc @matteobettini

### DIFF
--- a/torchrl/objectives/sac.py
+++ b/torchrl/objectives/sac.py
@@ -1201,6 +1201,7 @@ class DiscreteSACLoss(LossModule):
         # get probs and log probs for actions
         with self.actor_network_params.to_module(self.actor_network):
             dist = self.actor_network.get_dist(tensordict.clone(False))
+        prob = dist.probs
         log_prob = dist.logits
 
         td_q = tensordict.select(*self.qvalue_network.in_keys)
--- a/torchrl/objectives/sac.py
+++ b/torchrl/objectives/sac.py
@@ -1202,7 +1202,7 @@ class DiscreteSACLoss(LossModule):
         with self.actor_network_params.to_module(self.actor_network):
             dist = self.actor_network.get_dist(tensordict.clone(False))
         prob = dist.probs
-        log_prob = prob.clamp_min(torch.finfo(prob.dtype).resolution)
+        log_prob = prob.clamp_min(torch.finfo(prob.dtype).resolution).log()
 
         td_q = tensordict.select(*self.qvalue_network.in_keys)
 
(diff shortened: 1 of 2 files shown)


matteobettini:
I don't get the fix.

We need both prob and log_prob

We also need to rerun the example training run for discrete SAC and confirm the returns curve match

torchrl/objectives/sac.py
```diff
@@ -1201,8 +1201,7 @@ def _actor_loss(
         # get probs and log probs for actions
         with self.actor_network_params.to_module(self.actor_network):
             dist = self.actor_network.get_dist(tensordict.clone(False))
-        prob = dist.probs
-        log_prob = prob.clamp_min(torch.finfo(prob.dtype).resolution)
+        log_prob = dist.logits
```
matteobettini, line 1204:
```suggestion
        prob = dist.probs
        log_prob = dist.logits
```

matteobettini (approved):
LGTM

If we want to be extra careful we should run the discrete_sac example to check

matteobettini:
This fix actually changed the behavior and is breaking the CI of BenchMARL for discrete SAC with MaskedCategorical.

What happens is that if `dist.probs` returns  `[[1.0000, 0.0000, 0.0000, 0.0000, 0.0000, 0.0000, 0.0000, 0.0000,
          0.0000, 0.0000, 0.0000]`

`dist.logits` will give `[ 0.0000,    -inf,    -inf,    -inf,    -inf,    -inf,    -inf,
             -inf,    -inf,    -inf,    -inf]`

while `torch.log(torch.where(dist.probs == 0, 1e-8, dist.probs))` gives ` [[  0.0000, -18.4207, -18.4207, -18.4207, -18.4207, -18.4207, -18.4207,
          -18.4207, -18.4207, -18.4207, -18.4207]`

Can we return to the previous behavior?

vmoens:
Not sure returning to the previous behaviour is appropriate. Why is this a problem?
-18 log prob is roughly equivalent to -inf and is more numerically stable IMO

matteobettini:
The -inf causes a but in MaskedCategorical https://github.com/facebookresearch/BenchMARL/actions/runs/7278697272/job/19833445677

If you prefer to adapt how the logits are computed in `MaskedCategorical`, that's should be a viable solution too

matteobettini:
I wonder what was the problem of the previous behavior tho.
We were just clamping the probs and computing the log, without calling `dist.logits`

vmoens:
Efficiency: we were recomputing something already computed by the dist
Stability: the log-probs should be normalised. The dist does that, we didn't.

And the code previously was seriously ugly

PR #1751, opened 2023-12-18
[Feature] Open-X Embodiement datasets

vmoens:
The data can be either downloaded or streamed. 

Example usage:
```python
import tempfile

from torchrl.data.datasets.openx import OpenXExperienceReplay
# Download the data, and sample 128 elements in each batch out of two trajectories
num_slices = 2
with tempfile.TemporaryDirectory() as root:
    dataset = OpenXExperienceReplay("cmu_stretch", batch_size=128,
        num_slices=num_slices, download=True, streaming=False, root=root)
    for batch in dataset:
        print(batch.reshape(num_slices, -1))
        break
    # Read data from a stream. Deliver entire trajectories when iterating
    dataset = OpenXExperienceReplay("cmu_stretch",
        num_slices=num_slices, download=True, streaming=False, root=root)
    for data in dataset: # data does not have a consistent shape
        break
    # Define batch-size dynamically
    data = dataset.sample(128)  # delivers 2 sub-trajectories of length 64
```

cc @jxu124 @quanvuong @nicklashansen

- [x] Should we return an exception when shuffle=False and calling sample?


vmoens:
Cc @nicklashansen I think this is more or less ready, if you want to give it a look!

nicklashansen:
Thanks! Seems to run quite smoothly for me. As one might expect, streaming is extremely slow but practical for debugging; download works like a charm!

One question I have: would there be a computational advantage to only sampling specific keys (provided by the user) rather than sampling the full TensorDict and then only using some of them? E.g. I might want to use only a single image per step, or no images, or leave out the `next` data field. Is that (a) feasible to implement, and (b) even necessary?

vmoens:
This should do the trick if you want to exclude `("next", "reward")` and `"observation"` entries:

```python
from torchrl.envs import SelectTransform, ExcludeTransform 
dataset = OpenXExperienceReplay(..., transform=ExcludeTransform(("next", "reward"), "observation"))
```

`SelectTransform` has the opposite behaviour.

Would that help?

nicklashansen:
Yea! I guess my question is whether there is any *computational* advantage to excluding / selecting data as a transform vs. sampling the full TensorDict and just ignoring that the extra fields are there?

nicklashansen:
Thanks! This works wonders.

quanvuong:
do you still need my input : ) ?

nicklashansen:
@quanvuong We have a working dataloader now, so no need I think! Thanks though :-)

PR #1753, opened 2023-12-19
[Minor] Fix runnability of RLHF example in examples/rlhf

ianbarber:
## Description

Small amount of bit rot on the RLHF example.

- Correct passing of configuration in train transformer and reward steps
- Fix typo on prompt logger
- Update log_scalar to match class spec.

## Motivation and Context

There was one larger issue which I didn't address, which was that tensordictmodule wasn't playing well with dynamo, due to the optimizedmodule not forwarding getattr calls to the underlying module: I didn't get round to verifying that was fixed in nightlies (though I think it might be), but I did think about about updating the train/train_reward configs to default compile to be off as it will fail in the stable release version with `TypeError: _forward_unimplemented() got an unexpected keyword argument 'input_ids'` or similar. Its easy enough to disable compile in config or on command line, but it might be worth disabling in the base config as well.

## Types of changes

## Checklist

### DIFF
--- a/examples/rlhf/train.py
+++ b/examples/rlhf/train.py
@@ -62,7 +62,7 @@ def main(cfg):
     dtype = cfg.sys.dtype
     compile_ = cfg.sys.compile
 
-    ctx = setup(device=device, dtype=dtype)
+    ctx = setup(cfg.sys)
 
     train_loader = get_dataloader(
         data_cfg.batch_size,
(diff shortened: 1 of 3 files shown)


vmoens:
> which was that tensordictmodule wasn't playing well with dynamo, due to the optimizedmodule not forwarding getattr calls to the underlying module

Yep! `tensordict` (and related modules) compatibility with dynamo and the rest of the compile stack is a big TODO. Happy to chap about what the roadmap for this looks like offline.

vmoens (approved):
LGTM
Our `examples` CI should have caught these...

PR #1760, opened 2023-12-22
[Benchmark] Fix RB benchmarks

vmoens:
cc @harnvo

I changed the names so that we can have a fresh look at the benchmarks on the benchmark tool. 
I will try to remove the old traces that are wrong (need to look at how I can modify the xml files on gh-pages)

### DIFF
--- a/benchmarks/test_replaybuffer_benchmark.py
+++ b/benchmarks/test_replaybuffer_benchmark.py
@@ -3,6 +3,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 import argparse
+import functools
 
 import pytest
 import torch
@@ -58,8 +59,7 @@ def sample(rb):
 
 
 def iterate(rb):
-    for _ in rb:
-        break
+    next(rb)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


harnvo:
Hi, just a suggestion. The `partial` fix proposed in my [issue](../../rl/issues/1759) is just for easy illustration but might not be appropriate for a merge.

Somehow pytest do not recognize the class name anymore after `partial(TensorDictPrioritizedReplayBuffer, alpha=1, beta=0.9)`. (In practice you would see `pb6`, `pb7`, ... or something as their class name. ) This adds confusion when benchmarking (especially when you would like to add more tests with partial!!).

This is because `partical` change the `__name__` of a class, and I would like to suggest solving this issue by doing the following:

```python
# preserve the name of the class even after partial
_prb_name = TensorDictPrioritizedReplayBuffer.__name__
_TensorDictPrioritizedReplayBuffer = partial(TensorDictPrioritizedReplayBuffer, alpha=0.6, beta=0.4)
_TensorDictPrioritizedReplayBuffer.__name__ = _prb_name
```

and then change the `partial(TensorDictPrioritizedReplayBuffer, alpha=1, beta=0.9)` within `parameterize` into `_TensorDictPrioritizedReplayBuffer`

vmoens:
@harnvo edited following your suggestion, would mean a lot if you could submit your review!

benchmarks/test_replaybuffer_benchmark.py
```diff
@@ -17,6 +18,12 @@
 )
 from torchrl.data.replay_buffers import RandomSampler, SamplerWithoutReplacement
 
+_TensorDictPrioritizedReplayBuffer = partial(
```
harnvo, line 21: typo: functools.partial

vmoens, line 21:
```suggestion
_TensorDictPrioritizedReplayBuffer = functools.partial(
```

benchmarks/test_replaybuffer_benchmark.py
```diff
@@ -71,22 +77,28 @@ def iterate(rb):
...
+        rb=rb,
+        storage=storage,
+        sampler=sampler,
         populated=True,
         size=size,
     )()
     benchmark(sample, rb)
 
 
+def infinite_iter(obj):
```
harnvo, line 96: It seems there exists a infinite loop here and the `test_rb_iterate` could not end the 4th test. (Possibly due to the circular sampling in `SamplerWithoutReplacement`)

The old `iterate` works fine and I think you should change this back.

harnvo (changes requested):

benchmarks/test_replaybuffer_benchmark.py
vmoens: Yep, I wanted to avoid the cost of entering `__iter__` but it's still better than an infinite loop!

harnvo: In this case, the following would work:

```python 
def iterate(rb_iter):
    try:
        next(rb_iter)
    except StopIteration:
        pass
```

and inside `test_rb_iterate`:

```python
benchmark(iterate, iter(rb))
```

vmoens: The problem is that I was call iter on iter, not solving the problem that the first iter was empty. I will push a fix in a sec

harnvo (approved):
Works in my PC

vmoens:
@harnvo it should run now!

PR #1762, opened 2023-12-25
[BugFix] Fix slice sampler

vmoens:
cc @nicklashansen

### DIFF
--- a/torchrl/data/replay_buffers/samplers.py
+++ b/torchrl/data/replay_buffers/samplers.py
@@ -675,7 +675,9 @@ class SliceSampler(Sampler):
                     and self._used_traj_key[0] == "_data"
                 )
                 vals = self._find_start_stop_traj(trajectory=trajectory[: len(storage)])
-                return self._cache.setdefault("stop-and-length", vals)
+                if self.cache_values:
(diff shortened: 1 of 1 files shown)


nicklashansen:
@vmoens It appears that the last time step of an episode never gets sampled. Unclear whether this is unique to SliceBuffer or also present in other samplers? Discovered by @dasGringuen using the TD-MPC2 repo and the new SliceBuffer, and reproduced by me.

**How to reproduce:**

Appending a single time step with all `nan`s does not break the code in any way (they never get sampled during model updates). Appending **two** time steps with `nan`s raises a CUDA error.

```
self._tds.append(self.to_td()) # Add NaNs to end of episode
self._tds.append(self.to_td()) # Add NaNs to end of episode <- this line breaks the code
self._ep_idx = self.buffer.add(torch.cat(self._tds))
```

where `self.to_td()` constructs a TensorDict that obeys the shape of transitions but is filled with `nan`s.

Added here in the tdmpc2 repo: https://github.com/nicklashansen/tdmpc2/blob/33876d124f19e6996c3a09b16b43183da2ef6cf3/tdmpc2/trainer/online_trainer.py#L91

Issue #1676, opened 2023-11-06
[Feature Request] Add a Transform that remembers the previous action taken


maxweissenbacher:
## Motivation

I would propose implementing a simple Transform that remembers the previous action taken. This can be very useful when considering environments with time delay, or when training agents with recurrent policies (for instance, the policy could be a function not only of the state but also the action taken in the time step prior).

## Solution

I'd be happy to implement this if you think this could be useful.

The Transform could simply add a new key "previous_action" (or otherwise chosen by user) to the "next" tensordict which copies the "action" from the tensordict and initialises to zero (or another default value specified by the user).

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

vmoens:
Sure that wouldn't be too difficult. The _step function of your transform will take the action from the `tensordict` input and copy in the `next_tensordict` and that is all!

maxweissenbacher:
I'll get to work on it :)

maxweissenbacher:
note to self: the CatFrames transform does this (more generally, it implements a buffer) with observations instead of actions.

vmoens:
Do we need a time lag for this? Otherwise RenameTransform could take care of it (if we just need the last action)

YeonwooSung:
Taking this as a LastAction transform (CatFrames-for-actions): NestedKey in_keys/out_keys, copy action into next on _step, register the spec, reset default. Comment if RenameTransform is still preferred instead.

PR #1763, opened 2023-12-28
[Feature] Transform for Recording History Observations

btx0424:
## Description

This draft PR proposes a `Transform` for recording histories of observations, which is a common practice in robotic tasks. It should also address #1676: to record observation-action history, append the previous action to observation, and then transform the environment with `History`.

The naming, description (docstring), and tests are for the initial commit. Opinions and suggestions are wanted.

## Types of changes

## Checklist


test/test_transforms.py
```diff
@@ -9366,6 +9367,74 @@ def test_transform_inverse(self):
...
+        env = CountingBatchedEnv(max_steps=max_steps, batch_size=[n])
+        transform = Compose(StepCounter(), History(["observation"], steps=16, include_last=include_last))
+        env = TransformedEnv(env, transform)
+        rollout = env.rollout(32, break_when_any_done=False)
+        for td in rollout.unbind(0):
+            obs = td["observation"]
+            if include_last:
+                obs_h = td["observation_h"]
+            else:
+                obs_h = td[("next", "observation_h")]
```
vmoens, line 9426:
```suggestion
                obs_h = td["next", "observation_h"]
```

test/test_transforms.py
```diff
@@ -9366,6 +9367,74 @@ def test_transform_inverse(self):
...
+            steps = td["step_count"]
+            dones = td[("next", "done")].squeeze().nonzero().squeeze(-1)
```
vmoens, line 9428:
```suggestion
            dones = td["next", "done"].squeeze().nonzero().squeeze(-1)
```

torchrl/envs/transforms/transforms.py
```diff
@@ -6552,3 +6552,103 @@ def transform_observation_spec(self, observation_spec: TensorSpec) -> TensorSpec
...
+        >>> env = TransformedEnv(GymEnv("CartPole-v1"), History(["observation"]))
+        >>> td = env.reset()
+        >>> print(td["observation_h"].shape)
+        torch.Size([4, 16])
+    """
+    def __init__(
+        self,
+        in_keys: Sequence[NestedKey],
+        out_keys: Sequence[NestedKey] = None,
+        steps: int = 16,
```
vmoens, line 6586: Is there a reason why we use 16?

vmoens (commented):
Cool stuff!
Is this the `stack` version of `CatFrames`? Do we want to have two separate transforms for that? If what we want is for `CatFrames` to have the option to discard the last observation or not, or use stack instead of cat, these could be passed as arguments (though I can see why one would rather have two separate classes). I can give some advice if we want to go down that route.

If we have two separate transforms, maybe we could name this one `StackFrames` for consistency? Is there any piece of code from CatFrames that we should recycle?

We should add the transform to the doc too.

btx0424:
Thanks for replying! 

Yes, I think renaming it to `StackFrames` is a good idea. However, do you prefer it to be stateful (like `CatFrames`) or stateless (the current implementation)? I will make it reuse `CatFrames` more if we go for stateful.

I will complete the tests and doc soon.

vmoens:
It's good to have it stateless I agree. Let's keep it as it is!
Let's align the doc of this class and CatFrames and refer one to the other, it's likely that users will jump from one to the other.
Thanks again!

vmoens:
@btx0424 Anything I can help you with here? I can find some time to help if needed

btx0424:
> @btx0424 Anything I can help you with here? I can find some time to help if needed

Hi there. I had some trouble figuring out the no-env usage while trying to align it with `CatFrames`. I noticed that `CatFrames`, when used for RBs, do not split between trajectories. In other words, some observations may contain information from the last episode. Is that the desired behavior?

vmoens:
Just seeing this now sorry
No definitely not!
We should fix this if it's the case

gtnv:
hi @btx0424, is there more work planned here? if not, can we close it?

PR #1768, opened 2024-01-03
[Feature] Composite replay buffers

vmoens:
## Description

In this PR, I propose a composite Replay Buffer API where one can put together multiple replay buffers and sample from each of them either in a structured or random way.

The composite replay buffer can be built in two different ways:
```python
rb = ReplayBufferEnsemble(*list_of_replay_buffers)
```
or
```python
rb = ReplayBufferEnsemble(storages=StorageEnsemble(*storages), ...)
```

That way, users have either full control over storage, sampler etc (provided they have the appropriate signatures) or they can just combine them naively. 

The RBEnsemble class accepts a `p` argument that tells how much importance is to be given to each replay buffer during sampling. A `collate_fn` can be passed too, it will be executed after the `collate_fn` of each independent replay buffer has been called.  A `sample_from_all` argument allows users to choose whether the sampled batches should contain a sub-batch from each buffer or not. If not, `num_buffer_sampled` will indicate how many buffers have to be sampled from every time (using the probabilities `p` if provided). 

Samples are stacked along the first dimension, giving samples of shape `[B, S]` where `B=num_buffer_sampled` and `S=batch_size//num_buffer_sampled`.

Thanks to `LazyStackedTensorDict`, we can stack any kind of data but one may wish to make the data uniform first, for this, independent transforms have to be plugged onto each buffer (eg, key names differ in the buffers or the image sizes are incongruent).

Questions:
- [x] What should we do with the writers? I think it will be clunky and hard to maintain to have the possibility to write in that buffer when we could simply just write in the buffers independently (ie, the writer should just raise `NotImplementedError` or `RuntimeError` when called)
- [x] when calling `rb.__getitem__`, we get the item in the storage and massage it a bit (using transforms etc) with a regular replay buffer. Do we want to allow that method in the ensemble too? Maybe with `rb[buffer_id, index]`?

cc @nicklashansen


nicklashansen:
> What should we do with the writers? I think it will be clunky and hard to maintain to have the possibility to write in that buffer when we could simply just write in the buffers independently (ie, the writer should just raise NotImplementedError or RuntimeError when called)

I'm not sure I fully understand the problem here. Are you suggesting that writing to the ensemble buffer would raise an error, but that users can still access and write to each of the buffers in the ensemble? One use case for writing would be multi-task online RL, where the user may wish to maintain separate buffers for each task and perhaps sample data from each task at different rates.

> when calling rb.__getitem__, we get the item in the storage and massage it a bit (using transforms etc) with a regular replay buffer. Do we want to allow that method in the ensemble too? Maybe with rb[buffer_id, index]?

Sounds reasonable to me!

vmoens:
> users can still access and write to each of the buffers in the ensemble

Exactly, that seems to be the most intuitive thing to do. I don't see how `extend` or `add` could be routed to the appropriate replay buffer.

vmoens:
@nicklashansen not too far from done, check the doc here https://docs-preview.pytorch.org/pytorch/rl/1768/reference/data.html#composing-datasets

nicklashansen:
@vmoens Thanks! It looks pretty straightforward based on the documentation.

I did try it out on the OpenXExperienceReplay dataset, but am getting an error during sampling. Seems like there is some dimension mismatch in the batch size? This is a slightly different use case than the one in the documentation, since I'm working with datasets rather than replay buffers.

Example:
```
self._dataset = OpenXExperienceReplay(
			dataset_id=cfg.get('dataset_id', 'ucsd_pick_and_place_dataset_converted_externally_to_rlds'),
			shuffle=True,
			slice_len=cfg.horizon+1,
			streaming=False,
			root=cfg.data_dir,
			download=True,
		)
self._datasets = ReplayBufferEnsemble(
			self._dataset,
			self._dataset,
			p=[0.5, 0.5],
		)
self._datasets.sample(32)
```

results in the error
```
  File "/data2/nihansen/code/openx/tdmpc2/train.py", line 94, in launch
    train(0, cfg)
  File "/data2/nihansen/code/openx/tdmpc2/train.py", line 70, in train
    buffer=OpenX(cfg),
  File "/data2/nihansen/code/openx/tdmpc2/common/openx.py", line 36, in __init__
    data = self._datasets.sample(32)
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/torchrl/data/replay_buffers/replay_buffers.py", line 444, in sample
    ret = self._sample(batch_size)
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/torchrl/data/replay_buffers/replay_buffers.py", line 1320, in _sample
    sample, info = super()._sample(*args, **kwargs)
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/torchrl/data/replay_buffers/utils.py", line 47, in decorated_fun
    output = fun(self, *args, **kwargs)
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/torchrl/data/replay_buffers/replay_buffers.py", line 380, in _sample
    info["index"] = index
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/tensordict/_td.py", line 535, in __setitem__
    self._set_tuple(
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/tensordict/_td.py", line 1351, in _set_tuple
    return self._set_str(key[0], value, inplace=inplace, validated=validated)
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/tensordict/_td.py", line 1321, in _set_str
    value = self._validate_value(value, check_shape=True)
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/tensordict/base.py", line 3817, in _validate_value
    value.batch_size = self.batch_size
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/tensordict/_td.py", line 1271, in batch_size
    self._batch_size_setter(new_size)
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/tensordict/base.py", line 477, in _batch_size_setter
    self._check_new_batch_size(new_batch_size)
  File "/data/nihansen/miniconda3/envs/openx/lib/python3.9/site-packages/tensordict/base.py", line 3774, in _check_new_batch_size
    raise RuntimeError(
RuntimeError: the tensor index has shape torch.Size([2, 16]) which is incompatible with the batch-size torch.Size([2, 16, 1]).
```

Any suggestions for how to resolve this?

PR #1773, opened 2024-01-06
[Feature] GenDGRL

vmoens:
Gen-DGRL dataset.

To comply with other datasets, we use a `dataset_id` that contains the procgen env as well as the category, eg `bigfish-1M_E`.

Tests just process the first 100 samples of a bunch of datasets.

Cc @ishitamed19 @SharathRaparthy @braj29


torchrl/data/datasets/gen_dgrl.py
```diff
@@ -0,0 +1,369 @@
...
+_has_tqdm = importlib.util.find_spec("tqdm", None) is not None
+_has_requests = importlib.util.find_spec("requests", None) is not None
+
+
+class GenDGRLExperienceReplay(TensorDictReplayBuffer):
+    """Gen-DGRL Experience Replay dataset.
+
+    This dataset accompanies the paper "The Generalization Gap in Offline Reinforcement Learning".
+    Arxiv: https://arxiv.org/abs/2312.05742
+    GitHub: https://arxiv.org/abs/2312.05742
```
YhgzXxfz, line 32: tiny: this should be https://github.com/facebookresearch/gen_dgrl?

torchrl/data/datasets/gen_dgrl.py
```diff
@@ -0,0 +1,369 @@
...
+
+    This class gives you access to the ProcGen dataset. Each `dataset_id` registered
+    in `GenDGRLExperienceReplay.available_datasets` consists in a particular task
+    (`"bigfish"`, `"bossfight"`, ...) separated from a category (`"1M_E1"`, `"1M_S"`, ...)
```
YhgzXxfz, line 36: typo: "1M_E"

torchrl/data/datasets/gen_dgrl.py
```diff
@@ -0,0 +1,367 @@
...
+    def _build_urls_with_category_name(
+        cls, dataset, category_name: str
+    ) -> tp.List[str]:
+        path = [cls.BASE_URL, cls._convert_category_name(category_name)]
+        path += [f"{dataset}.tar.xz"]
+        return os.path.join(*path)
+
+    @staticmethod
+    def _convert_category_name(category_name: str) -> str:
+        if category_name == "1M_E":
```
YhgzXxfz, line 196: This is good for now. In the future if we have more datasets, we would want to modify the `category_name` to reflect the parameters, like "1M_200_E", "100k_40_S", etc, so that we can easily map them separately and combine together for paths:
- "1M" -> "1M", "100k" -> "100k", ...
- "200" -> "level_200", "40" -> "level_40", ...
- "E" -> "expert", "S" -> "suboptimal"

vmoens: Sure! If you add more of these just ping us and we'll do the fixes

PR #1775, opened 2024-01-08
[Feature] batched trajectories - SliceSampler compatibility

vmoens:
## What does this PR do?

We introduce a new way of extending replay buffers that is more faithful to the way trajectories are organised in torchrl.
Improvements:
- Replay buffers can handle an arbitrary number of dimensions and users can choose which dimension to use for extension.
- Replay buffer size can reflect the total number of transitions even for multi-dimensional storages. 
- The samplers (incl. `SliceSampler` and subclasses) can handle different batch dimensions. For multidim storages, samples are always flattened (as if the storage was indexed by a mask)

## Why is this change needed

We currently have 2 ways of extending a buffer which I summarize in this ugly drawings

image
image

The **first attempt** we had at storing trajectories was to build the buffer according to 
```python
T = ... # length of a trajectory
B = ... # number of envs simulated at once in the collector or in ParallelEnv
rb = ReplayBuffer(storage=LazyTensorStorage(max_size=total_num_transitions // T), transfrom=RandomCropTensorDict(...))
for data in collector:
   # data has shape [B, T]
   rb.extend(data)
```
Sampling consists in getting trajectories returned by the collector. This isn't amazing because trajectories could be interrupted in the middle with a truncated/terminated signal, or simply be longer than `T`.
The other issue is that your buffer size isn't the number of transitions but the number of traj stored. One can immediately see why this isn't as flexible as one could hope for!
Finaly drawback: subsampling slices needed to be done via https://github.com/pytorch/rl/blob/31bea14293b4dcb2c41636c5673035c16bade867/torchrl/envs/transforms/transforms.py#L5712 which is slow.

The second attempt was simply to flatten the tensordict and use [`SliceSampler`](https://github.com/pytorch/rl/blob/31bea14293b4dcb2c41636c5673035c16bade867/torchrl/data/replay_buffers/samplers.py#L560) to get sub-trajectories
```python
T = ... # length of a trajectory
B = ... # number of envs simulated at once in the collector or in ParallelEnv
rb = ReplayBuffer(storage=LazyTensorStorage(max_size=total_num_transitions), sampler=SliceSampler(...))
for data in collector:
   # data has shape [B, T]
   rb.extend(data.reshape(-1))
```
Here the drawings above come in handy: we still have the problem that trajectories are not contiguous and can be truncated in multiple places if a ParallelEnv is used.
But it's closer to what people want and the buffer size is accurate.

## New API
The new API is quite simple: one just needs to pass the number of dimensions to the buffer and all the other components will know what to do:
```python
T = ... # length of a trajectory
B = ... # number of envs simulated at once in the collector or in ParallelEnv
rb = ReplayBuffer(storage=LazyTensorStorage(max_size=total_num_transitions, ndim=2), sampler=SliceSampler(...))
for data in collector:
   # data has shape [B, T]
   rb.extend(data) # rb is `data.numel()` longer!
```
This is the second picture I pasted earlier. The big advantage with SyncCollector + ParallelEnv is that you now have trajectories that are well structured and as long as you could wish.

## Other changes

- Using tensorclasses within `TensorDictReplayBuffer` is a bit buggy because we can't store the indices so I made this not possible anymore. I believe it is the only bc-breaking change of this PR, but all the features that used to be supported still are via a `ReplayBuffer` usage.
- I removed the `"_data"` prefix in storages placed within `TensorDictReplayBuffer` instances.

Gist: https://gist.github.com/vmoens/b928af1a3a9567b1f0862a58ec592f85

cc @btx0424 @albertbou92 @Cadene


vmoens:
I just pushed a completely broken commit but here's the status:
For slice samplers we now need to be able to get indices of the type 
```python
tensor([3, 1],
[4, 1],
[5, 1])
```
where the first column is the index along the time dimension and the second the index along the batch dimension (since dims are flipped in the buffer).
That means that indexing the storage with `storage[idx]` won't work since we need to use the first and second columns separately. We will need to do `storage[idx.unbind(-1)]` which will work just fine.

test/test_rb.py
```diff
@@ -1898,8 +1885,6 @@ def test_slice_sampler(
             [100],
             device=device,
         )
-        if _data_prefix:
-            data = TensorDict({"_data": data}, [100])
```
Cadene, line 1902: Happy that you found a way to simplify by removing the data_prefix :)

test/test_transforms.py
```diff
@@ -2381,10 +2381,7 @@ def test_transform_rb(self, include_forward, rbclass):
         )
         rb.extend(td)
 
-        if rbclass is ReplayBuffer:
-            storage = rb._storage._storage
-        else:
-            storage = rb._storage._storage.get("_data")
+        storage = rb._storage._storage[:]
```
Cadene, line 2384: I would suggest to use copy() or clone() to be more explicit than [:] 

```suggestion
        storage = rb._storage._storage.copy()
```

torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -858,67 +953,48 @@ def add(self, data: TensorDictBase) -> int:
...
+                    break
+            else:
+                raise RuntimeError(
+                    f"could not find how to reshape index with shape {index.shape} to fit in tensordict with shape {tensordict.shape}"
+                )
+            tensordict.set("index", index)
+            return
+        tensordict.set("index", expand_as_right(index, tensordict))
+
     def update_tensordict_priority(self, data: TensorDictBase) -> None:
```
Cadene, line 998: The signature of the method has changed since we added `return self.update_priority(index, priority)`

torchrl/data/replay_buffers/samplers.py
```diff
@@ -818,28 +849,12 @@ def _get_stop_and_length(self, storage, fallback=True):
...
-                        done = storage._storage.get(self._used_end_key)
-                    except KeyError:
-                        done = storage._storage.get(("_data", self.end_key))
-                        # cache that value for future use
-                        self._used_end_key = ("_data", self.end_key)
-                    self._uses_data_prefix = (
-                        isinstance(self._used_end_key, tuple)
-                        and self._used_end_key[0] == "_data"
+                try:
+                    done = storage[:].get(self.end_key)
```
Cadene, line 853: Same here, I would suggest to use copy() or clone() instead of [:] to be more explicit.

Cadene (approved):

Cadene:
Didnt run the code, but overall looks good to me!

I already expressed my small concern about this feature. Intuitively I would prefer to keep everything flattened for the sake of consistency, even tho it adds some complexity when rolling out multiple environments. But if you think this way makes more sense, I trust you!

Thanks for the solid work on this feature and all the additional improvements.

vmoens:
@Cadene 

> I would prefer to keep everything flattened for the sake of consistency

This will still be the default behaviour. I agree about your point regarding using multiple envs, but one might also argue that using a ParallelEnv should not change the training performance, just speed things up. In the current state of the buffer, if you are using `ParallelEnv` or `MultiSyncCollector` you get a buffer that is organised differently than when using a single env, and if you're using a SliceSampler the training with a single env will give you to data that is better organised, hence a better training performance. This PR allows you to solve that.

test/test_transforms.py
vmoens: The `[:]` does not copy but it indexes all valid elements of the storage (ie the storage up to its maximum length).
Note also that this can return anything (eg a tensordict, a tensorclass or any pytree), but not a Storage object.

torchrl/data/replay_buffers/samplers.py
vmoens: See comment above

PR #1782, opened 2024-01-09
[WIP, CI] Pre-release submitit scripts

vmoens:
## Description

In this PR, I propose a script to run all our benchmarks before the release.

cc @matteobettini @albertbou92 @BY571 @giadefa 

TODO:
- [ ] make sure Wandb logging is uniform across scripts
- [ ] Add wandb flag with release name and git commit
- [ ] find a way to report these results and compare them across releases (public wandb channel?)

### DIFF
--- a/benchmarks/sota-check/README.md
+++ b/benchmarks/sota-check/README.md
@@ -1,0 +1,8 @@
+# SOTA Performance checks
+
+This folder contains a `submitit-release-check.sh` file that executed all
+the training scripts using `sbatch` with the default configuration and a wandb
+logger.
+
+This script is to be executed before every release to assess the performance of
+the various algorithms available in torchrl.
--- a/benchmarks/sota-check/run_a2c_atari.sh
+++ b/benchmarks/sota-check/run_a2c_atari.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+#SBATCH --job-name=a2c_atari
+#SBATCH --ntasks=32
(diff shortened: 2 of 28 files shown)


vmoens:
@albertbou92 Do you plan on working on this or should I keep on doing it?
No worry if you don't have time but we need it to be wrapped by end of next week :)

albertbou92:
yes, I think I will have some time, but maybe I need a bit of guidance. 
So the idea is:
1) Make sure that all training scripts log to wandb
2) Review the examples to unify the metrics logged right?

vmoens:
> So the idea is:

First check if we can get all scripts to run ok with something as simple as what I drafted here :)

> Make sure that all training scripts log to wandb

Yes and logging should have some uniform format with little addition (all under the same project with just one arg in the command line for instance)

> Review the examples to unify the metrics logged

yep

> Centralise all generated data in a specified dir I guess?

What do you mean?
For now I think we can do things in house and share the results using wandb API between us, in the future a public display would be great!

> right? and then we manually run the script whenever we want to check all examples work as expected and verify that by visual inspection in wandb.

Yes all good

Also we have to check that wandb has the latest commit registered, that would be super useful for later (we could even put that commit in the name of the project??)

albertbou92:
https://github.com/pytorch/rl/pull/1822

gtnv:
hey @vmoens, is this something you still want to finish? if not, can we close it?

PR #1786, opened 2024-01-11
[BugFix] Ensure that infos and samples have the same batch-size in SamplerEnsemble

vmoens:
Solves a bug where `from_dict` was allocating the max batch-size to the info tensordict in the SamplerEnsemble.

cc @nicklashansen

### DIFF
--- a/torchrl/data/datasets/openx.py
+++ b/torchrl/data/datasets/openx.py
@@ -572,10 +572,12 @@ class _StreamingStorage(Storage):
             else:
                 yield data
 
-    def get(self, index: int) -> Any:
+    def get(self, index: range) -> Any:
         if not isinstance(index, range):
-            # we use a range to indicate how much data we want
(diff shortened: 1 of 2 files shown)


torchrl/data/datasets/openx.py
```diff
@@ -572,10 +572,12 @@ def __iter__(self):
...
 
-    def get(self, index: int) -> Any:
+    def get(self, index: range) -> Any:
         if not isinstance(index, range):
-            # we use a range to indicate how much data we want
-            raise RuntimeError("iterable datasets do not support indexing.")
+            if (index[1:] != index[:-1] + 1).any():
+                # we use a range to indicate how much data we want
+                raise RuntimeError("iterable datasets do not support indexing.")
+            index = range(index.shape[0])
```
vmoens, line 580: This is a bit expensive but it's necessary since we can't be sure of the type of the input

PR #1787, opened 2024-01-12
[Doc] Revamp envs doc

vmoens:
TODO:
- [x] Fix habitat install and check doc
- [x] envpool doc

### DIFF
--- a/torchrl/envs/libs/vmas.py
+++ b/torchrl/envs/libs/vmas.py
@@ -141,9 +141,10 @@ class VmasWrapper(_EnvWrapper):
         available_envs (List[str]): the list of the scenarios available to build.
 
     .. warning::
-        VMAS does not distinguish between truncation (when the env reached ``max_steps``) and termination, and all
-        end-of-trajectories are marked  as terminated. If you deem the ``truncation`` signal necessary, set
-        ``max_steps`` to ``None`` and use a :class:`~torchrl.envs.transforms.StepCounter` transform.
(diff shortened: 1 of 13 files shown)


torchrl/envs/libs/vmas.py
```diff
@@ -128,6 +138,12 @@ class VmasWrapper(_EnvWrapper):
         unbatched_reward_spec (TensorSpec): version of the spec without the vectorized dimension
         het_specs (bool): whether the enviornment has any lazy spec
         het_specs_map (Dict[str, bool]): dictionary mapping each group to a flag representing of the group has lazy specs
+        available_envs (List[str]): the list of the scenarios available to build.
+
+    .. warning::
+        VMAS does not distinguish between truncation (when the env reached ``max_steps``) and termination, and all
+        end-of-trajectories are marked  as terminated. If you deem the ``truncation`` signal necessary, set
+        ``max_steps`` to ``None`` and use a :class:`~torchrl.envs.transforms.StepCounter` transform.
```
vmoens, line 146: cc @matteobettini

matteobettini: What about 
```
VMAS returns a single ``done`` flag which does not distinguish between when the env reached ``max_steps`` and termination. If you deem the ``truncation`` signal necessary, set ``max_steps`` to ``None`` and use a :class:`~torchrl.envs.transforms.StepCounter` transform.
```

PR #1788, opened 2024-01-12
[Feature] `SamplerWithoutReplacement` state dictionary

matteobettini:
TODO:
- [x] Tests


vmoens (approved):
LGTM thanks so much for this

vmoens:
There is some torchsnapshot import that needs to be fixed

matteobettini:
Could you link me the problem in the CI?

I am not able to find it.
(I only see another problem with a vec env which is unrelated to this)

.github/unittest/linux_olddeps/scripts_gym_0_13/environment.yml
```diff
@@ -25,3 +25,4 @@ dependencies:
     - patchelf
     - pyopengl==3.1.4
     - ray<2.8.0
+    - torchsnapshot
```
vmoens, line 28: Nope let's avoid installing it everywhere, let's just skip tests where it is not installed already

vmoens (commented):
We already have pipelines where snapshot is installed and tested. Let's avoid installing it anywhere else!

test/test_rb.py
```diff
@@ -1553,6 +1559,53 @@ def test_error_noninit(self):
 
 
 class TestSamplers:
+    @pytest.mark.parametrize("backend", ["torch", "torchsnapshot"])
+    def test_sampler_without_rep_state_dict(self, backend):
+        if backend == "torchsnapshot" and not _has_snapshot:
+            pytest.skip()
```
vmoens, line 1565: Otherwise you could make the backend list dependent of whether _has_torchsnapshot is true of false

matteobettini: if you prefer that...

In that case it doesn't show that there were tests skipped and the number of tests will vary depending on the pipeline

vmoens: Yes but that's better, for instance fbcode does not like skipping tests 
Better not have a test listed than skipped systematically

vmoens (approved):
LGTM! Thanks
Can you edit the description?

PR #1795, opened 2024-01-13
[Feature] TorchRL2Gym conversion

### LINKED ISSUE #1200: [Feature Request] CatTensors could delete now empty nested tensordicts
## Motivation

The current implementation of `CatTensor` can process nested keys and delete them afterward. It is great and works just fine. However, It may be convenient to better handle keys mapping to `TensorDict` objects. Currently, only leaves can be provided. So, I'm just gathering all of them in a list. If I do so, it does what it is supposed to do, but do not delete the highest-level key which is now mapping to an empty `TensorDict`.

## Solution

One option would be to support providing keys mapping to `TensorDict` objects out-of-the-box, so that it knows it should process of the leaves internally and delete the top-level key. 

## Alternatives

Another option would just be to check for all any empty `tensorDict` after processing of all leave-keys and delete them (maybe only checking the common roots of the provided keys). In the later case, it would be great to add an assert to check that the provided keys are truly associated to leaves.

## Checklist

vmoens:
## Description

Proposes an API to register a torchrl env in gym(nasium).

This is aimed at being a universal converter from any simulator (incl torchrl itself) to gym.

For instance, you could use dm_control, brax or anything else with gym without any boilerplate code.

Check the [docstrings](https://docs-preview.pytorch.org/pytorch/rl/1795/reference/generated/torchrl.envs.EnvBase.html#torchrl.envs.EnvBase.register_gym) and [tests](https://github.com/pytorch/rl/blob/f1dae58891642d3a49905809d2c31363cd993c1e/test/test_libs.py#L237) to know more!

TODO: 
- [x] stateless envs: we should save the input data in the env during step for the next call to step.

Closes #1200 as well

cc @skandermoalla @BY571 @albertbou92 @duburcqa @vikashplus


vmoens:
I had to add a new transform (RemoveEmptySpecs) to remove empty specs and tensordicts from an env output

This might be relevant to MARL folks (@matteobettini) since they have highly nested structures.

duburcqa:
If I remember correctly it was a feature I requested at some point !

duburcqa:
Here we are: https://github.com/pytorch/rl/issues/1200

vmoens:
Yes I remembered that too but I can't find it haha

vmoens:
Yep thanks!
Interestingly I encountered the same problem with SelectTransform so I thought that having a single solution for all was preferrable

PR #1800, opened 2024-01-14
[Feature] UTD batch-size correction

vmoens:
Isn't is the case that the UTD should correct for the batch-size? 

cc @BY571

### DIFF
--- a/examples/cql/cql_online.py
+++ b/examples/cql/cql_online.py
@@ -12,6 +12,7 @@ The helper functions are coded in the utils.py associated with this script.
 
 """
 import logging
+import math
 import time
 
 import hydra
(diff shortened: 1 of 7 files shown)


vmoens:
For instance [in cleanrl](https://github.com/vwxyzjn/cleanrl/blob/fc4bd19a6ff5314301ab7a186ad8078267187bd5/cleanrl/dqn.py#L60-L71) they use a UTD of 10 following that convention (though they don't name UTD)

They do 1 update / sample every 10 iterations over 128 samples when collecting one frame at a time 
So UTD = 10

Maybe we should adapts our UTDs to match cleanrl's?

vmoens:
Perhaps, one option could be that utd is equal to the batch-size by default (which would be conservative, the "real" utd is unchanged in practice). That won't improve runtime but that will ensure that if you change the batch-size to a bigger one your training will get shorter.

cc @cadene for context

vmoens:
I'm not sure, because per se if you have a batch-size 10x bigger you're still consuming data 10x faster so in my view UTD should be bigger... 
I couldn't find any ref even though it seems that most think of UTD like we currently do, but I find it counterintuitive

PR #1802, opened 2024-01-15
[Doc] More depth in VMAS docs

matteobettini:


torchrl/envs/libs/vmas.py
```diff
@@ -627,22 +634,35 @@ class VmasEnv(VmasWrapper):
     Paper: https://arxiv.org/abs/2207.03530
 
     Args:
-        scenario (str or vmas.simulator.scenario.BaseScenario): the vmas environment to build.
+        scenario (str or vmas.simulator.scenario.BaseScenario): the vmas scenario to build.
+            Must be one of :attr:`.available_envs`. For a description and rendering of available scenarios see
+            https://github.com/proroklab/VectorizedMultiAgentSimulator?tab=readme-ov-file#main-scenarios
```
vmoens, line 639: Is there a way we can link that to a specific commit? I'm afraid this title changes one day and the link gets silently broken.
Also I would use the \`test \<link\>\`_ syntax

torchrl/envs/libs/vmas.py
```diff
@@ -115,13 +115,20 @@ class VmasWrapper(_EnvWrapper):
         env (``vmas.simulator.environment.environment.Environment``): the vmas environment to wrap.
 
     Keyword Args:
-        num_envs (int): Number of vectorized simulation environments.
-        device (torch.device, optional): Device for simulation.
-        continuous_actions (bool, optional): Weather to use continuous actions. Defaults to ``True``.
-        max_steps (int, optional): Maximum number of steps in each vectorized environment after which done is returned.
-            Defaults to ``None`` (no truncation).
+        num_envs (int): Number of vectorized simulation environments. VMAS perfroms vectorized simulation using PyTorch.
+            This argument decides the number of vectorized envuironments that should be simulated in a batch. It will also
```
vmoens, line 119:
```suggestion
            This argument decides the number of vectorized environments that should be simulated in a batch. It will also
```

torchrl/envs/libs/vmas.py
```diff
@@ -115,13 +115,20 @@ class VmasWrapper(_EnvWrapper):
...
+            determine the batch size of the environment.
+        device (torch.device, optional): Device for simulation. Defaults to ``"cpu"``. All the tensors created by VMAS
```
vmoens, line 121: Defaults to `"cpu"` or default device?
ie what if I do
```python
with torch.device("cuda"):
   env = VMASEnv(...)
```
?

torchrl/envs/libs/vmas.py
```diff
@@ -115,13 +115,20 @@ class VmasWrapper(_EnvWrapper):
...
+            will be placed on this device.
+        continuous_actions (bool, optional): Whether to use continuous actions. Defaults to ``True``. If ``False``, actions
+            will be discrete. The number of actions and their size will depend on the scenario chosen.
+            See the VMAS repositiory for more info.
+        max_steps (int, optional): Horizon of the task. Defaults to ``None`` (infinite horizon). Each VMAS scenario can
```
vmoens, line 126: just to make sure: even when rewards can be consumed there is no max_steps predefined right? All envs are not truncated?

torchrl/envs/libs/vmas.py
```diff
@@ -115,13 +115,20 @@ class VmasWrapper(_EnvWrapper):
...
+            implement a ``done`` function that will define when the scenario is terminated. If ``max_steps`` is specified,
+            the scenario will also be terminated after this horizon has been reached. If instead of terminating the scenario
```
vmoens, line 128:
```suggestion
            the scenario can also be terminated whenever this horizon is reached. If instead of terminating the scenario
```

torchrl/envs/libs/vmas.py
```diff
@@ -115,13 +115,20 @@ class VmasWrapper(_EnvWrapper):
...
+            you wish to truncate it, please use a :class:`~torchrl.envs.transforms.StepCounter` transform.
```
vmoens, line 129: I don't get the distinction except that StepCounter sets a truncation. It's not clear to which extend they differ from each other.

torchrl/envs/libs/vmas.py
```diff
@@ -627,22 +634,35 @@ class VmasEnv(VmasWrapper):
...
+            Must be one of :attr:`.available_envs`. For a description and rendering of available scenarios see
+            https://github.com/proroklab/VectorizedMultiAgentSimulator?tab=readme-ov-file#main-scenarios
 
     Keyword Args:
-        num_envs (int): Number of vectorized simulation environments.
-        device (torch.device, optional): Device for simulation.
-        continuous_actions (bool, optional): Weather to use continuous actions. Defaults to ``True``.
-        max_steps (int, optional): Maximum number of steps in each vectorized environment after which done is returned.
-            Defaults to ``None`` (no truncation).
+        num_envs (int): Number of vectorized simulation environments. VMAS perfroms vectorized simulation using PyTorch.
```
vmoens, line 642: see above for edits

torchrl/envs/libs/vmas.py
```diff
@@ -627,22 +634,35 @@ class VmasEnv(VmasWrapper):
...
-            them to categorical or one-hot.
+            them to categorical or one-hot. Defaults to ``True``.
         group_map (MarlGroupMapType or Dict[str, List[str]], optional): how to group agents in tensordicts for
             input/output. By default, if the agent names follow the ``"<name>_<int>"``
             convention, they will be grouped by ``"<name>"``. If they do not follow this convention, they will be all put
             in one group named ``"agents"``.
             Otherwise, a group map can be specified or selected from some premade options.
             See :class:`~torchrl.envs.utils.MarlGroupMapType` for more info.
+        **kwargs (Dict, optional): These are additional argument that can be passed to the VMAS scenario constructor.
+            (e.g., number of agents, reward sparsity). The available arguments will vary based on the scenario chosen.
```
vmoens, line 663:
```suggestion
            (e.g., number of agents, reward sparsity). The available arguments will vary based on the chosen scenario.
```

torchrl/envs/libs/vmas.py
```diff
@@ -627,22 +634,35 @@ class VmasEnv(VmasWrapper):
...
+            To see the available arguments for a specific scneario, see the constructor in its file from
```
vmoens, line 664:
```suggestion
            To see the available arguments for a specific scenario, see the constructor in its file from
```

torchrl/envs/libs/vmas.py
```diff
@@ -627,22 +634,35 @@ class VmasEnv(VmasWrapper):
...
+             https://github.com/proroklab/VectorizedMultiAgentSimulator/tree/main/vmas/scenarios
```
vmoens, line 665: same, let's use a link to a specific commit if possible + \`text \<link\>\`_

torchrl/envs/libs/vmas.py
```diff
@@ -115,13 +115,20 @@ class VmasWrapper(_EnvWrapper):
...
-        continuous_actions (bool, optional): Weather to use continuous actions. Defaults to ``True``.
-        max_steps (int, optional): Maximum number of steps in each vectorized environment after which done is returned.
-            Defaults to ``None`` (no truncation).
+        num_envs (int): Number of vectorized simulation environments. VMAS perfroms vectorized simulation using PyTorch.
+            This argument decides the number of vectorized envuironments that should be simulated in a batch. It will also
+            determine the batch size of the environment.
+        device (torch.device, optional): Device for simulation. Defaults to ``"cpu"``. All the tensors created by VMAS
+            will be placed on this device.
+        continuous_actions (bool, optional): Whether to use continuous actions. Defaults to ``True``. If ``False``, actions
+            will be discrete. The number of actions and their size will depend on the scenario chosen.
```
vmoens, line 124:
```suggestion
            will be discrete. The number of actions and their size will depend on the chosen scenario.
```

torchrl/envs/libs/vmas.py
```diff
@@ -115,13 +115,20 @@ class VmasWrapper(_EnvWrapper):
...
+            See the VMAS repositiory for more info.
+        max_steps (int, optional): Horizon of the task. Defaults to ``None`` (infinite horizon). Each VMAS scenario can
+            implement a ``done`` function that will define when the scenario is terminated. If ``max_steps`` is specified,
```
vmoens, line 127: what does it mean that an env can define a done funtion? Some envs are never done? Can I choose if they have a done function?

vmoens (commented):
Thanks so much for this! I left some comments

torchrl/envs/libs/vmas.py
matteobettini: see the answer on the other question on this line

torchrl/envs/libs/vmas.py
matteobettini: This would be wrong. Can expresses possibility, here I am communicating certanty

torchrl/envs/libs/vmas.py
matteobettini: max_steps stets the internal vmas done (terminated in torchrl). StepCounter would set truncation

torchrl/envs/libs/vmas.py
matteobettini: Yes, each scenario has a optional "done" function. Some scenarios do not present that yes. Those scenarios would be eventually terminated if you use max_steps. You can choose if you implement a scenario yourself, otherwise it is a property of the scenario

torchrl/envs/libs/vmas.py
matteobettini: If I link to a specific commit, users won't see new senarios

torchrl/envs/libs/vmas.py
matteobettini: If I link to a specific commit, users won't see new senarios

torchrl/envs/libs/vmas.py
vmoens: got it
What is the `done` function in practice?
How does it relate to the `"done"` key?
I'm asking this to clarify if it's required to mention a function that we do not see appearing anywhere or if we can just say that env can be non-terminating unless max_steps is provided.

matteobettini: I am also pro to not mention it. I just did cause I did not know how much depth is required here.

we can definitely just say what you proposed. Seems better to me to.

We can say something like: some scnearios are terminating and some not. In all cases max_steps provides an additional termination condition

torchrl/envs/libs/vmas.py
vmoens: sure but if the link breaks we won't know about it either. A fixed link seems more conservative to me, wdyt?

vmoens: If there was a doc we could link to the doc which I would assume would not change, but given that there isn't the best seems to be a link to some commit we know won't change

torchrl/envs/libs/vmas.py
vmoens: I don't think users need to know the inner machinery but they should know precisely how changing a kwarg will impact what they get from the environment.

torchrl/envs/libs/vmas.py
vmoens: see my comment below

matteobettini: done

torchrl/envs/libs/vmas.py
matteobettini: ok done

torchrl/envs/libs/vmas.py
matteobettini: Changed

torchrl/envs/libs/vmas.py
```diff
@@ -627,22 +634,37 @@ class VmasEnv(VmasWrapper):
...
         categorical_actions (bool, optional): if the environment actions are discrete, whether to transform
-            them to categorical or one-hot.
+            them to categorical or one-hot. Defaults to ``True``.
         group_map (MarlGroupMapType or Dict[str, List[str]], optional): how to group agents in tensordicts for
             input/output. By default, if the agent names follow the ``"<name>_<int>"``
             convention, they will be grouped by ``"<name>"``. If they do not follow this convention, they will be all put
             in one group named ``"agents"``.
             Otherwise, a group map can be specified or selected from some premade options.
             See :class:`~torchrl.envs.utils.MarlGroupMapType` for more info.
+        **kwargs (Dict, optional): These are additional argument that can be passed to the VMAS scenario constructor.
```
vmoens, line 663:
```suggestion
        **kwargs (Dict, optional): These are additional arguments that can be passed to the VMAS scenario constructor.
```

vmoens (approved):
LGTM

PR #1803, opened 2024-01-15
[Performance] faster gym-like class

vmoens:
## Description

In this PR, I propose a change in the `_step` from `GymLikeEnv` where we validate the value provided to the tensordict once and for all if possible. The behaviour is as follows:

At first, `validated` is `None` by default. `None` means that we need to validate whether or not the data must be validated by the tensordict.

Then, we check if building the tensordict a second time changes the identity of its leaves. If so, it means that some casting has occurred. If not, we are confident that the value passed to the TD are the same as originally and no check should be executed.

We then save this `validated` attribute and never check it again.

On my old macbook, `TensorDict.__init__` was raking 5th in the list of the most expensive functions with 15% of runtime associated, i now ranks way below (about 30th?) with 3% only.

I will report some more detailed throughput measures soon.

@duburcqa

### DIFF
--- a/torchrl/envs/gym_like.py
+++ b/torchrl/envs/gym_like.py
@@ -307,6 +307,7 @@ class GymLikeEnv(_EnvWrapper):
         tensordict_out = TensorDict(
             source=source,
             batch_size=self.batch_size,
+            _run_checks=not self.validated,
         )
         if self.info_dict_reader and info is not None:
             for info_dict_reader in self.info_dict_reader:
--- a/torchrl/envs/gym_like.py
+++ b/torchrl/envs/gym_like.py
@@ -180,9 +180,9 @@ class GymLikeEnv(_EnvWrapper):
             if truncated is not None:
                 truncated = [truncated]
         return (
-            terminated,
-            truncated,
-            done,
+            torch.as_tensor(terminated),
+            torch.as_tensor(truncated),
+            torch.as_tensor(done),
             do_break.any() if not isinstance(do_break, bool) else do_break,
         )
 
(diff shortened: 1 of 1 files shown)


duburcqa:
> On my old macbook, TensorDict.__init__ was raking 5th in the list of the most expensive functions

On my new MacBook, it was also ranking 5th but 1st if I ignore non-torchrl methods. The improvement sounds amazing ! I can run my benchmark on this PR once you feel it is ready.

vmoens:
> On my new MacBook, it was also ranking 5th but 1st if I ignore non-torchrl methods. The improvement sounds amazing ! I can run my benchmark on this PR once you feel it is ready.

Feel free to try it out on the nightlies.

I can maybe improve some other bits (related to parallel env in particular). Not sure how much room for improvement there is on this side though

duburcqa:
> Feel free to try it out on the nightlies.

Ok i will try to do it today !

> I can maybe improve some other bits (related to parallel env in particular). Not sure how much room for improvement there is on this side though

From the old timings I provided, I'm not sure it is the main bottleneck at the time being:
```
    %Own   %Total  OwnTime  TotalTime  Function (filename)                                                                                                                                                                                             
    70.00%  82.00%   172.6s    222.0s   step (jiminy_py/simulator.py)
->  1.00%   2.00%    3.93s     5.29s   inner_terminated_or_truncated (torchrl/envs/utils.py)
->  1.00%   1.00%    2.75s     2.81s   encode (torchrl/data/tensor_specs.py)
->  0.00%   0.00%    2.56s     2.56s   __new__ (tensordict/tensordict.py)
->  1.00%   2.00%    2.02s    10.65s   _set (torchrl/envs/utils.py)
->  3.00%   7.00%    1.90s     6.39s   _set_str (tensordict/tensordict.py)
->  2.00%   4.00%    1.84s     4.08s   _validate_value (tensordict/tensordict.py)
->  0.00%   0.00%    1.79s     3.33s   get (tensordict/tensordict.py)
->  2.00%   2.00%    1.61s     1.68s   _convert_to_tensor (tensordict/tensordict.py)
->  0.00%   0.00%    1.44s     1.44s   __getitem__ (torchrl/data/tensor_specs.py)
```

duburcqa:
I did the update, things are not going well ! using the same benchmark script as before:
```bash
  File "/opt/homebrew/Cellar/python@3.11/3.11.7/Frameworks/Python.framework/Versions/3.11/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/opt/homebrew/lib/python3.11/site-packages/torchrl/_utils.py", line 621, in run
    return mp.Process.run(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.7/Frameworks/Python.framework/Versions/3.11/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/homebrew/lib/python3.11/site-packages/torchrl/collectors/collectors.py", line 2197, in _main_async_collector
    raise RuntimeError(
```

What I did to install the latest pre-release:
```bash
pip install --upgrade --pre torch --index-url https://download.pytorch.org/whl/nightly/cpu
pip install --upgrade tensordict-nightly torchrl-nightly
```

duburcqa:
After a quick benchmark with `SyncDataCollector `, sadly it runs slower on nightly than the current latest official release. The slowdown is quite significant, something ranging from 5 to 10% I would say.

PR #1810, opened 2024-01-17
[BugFix] Make `TransformedEnv` mirror `allow_done_after_reset` property of base env

### LINKED ISSUE #1806: [BUG] `TransformedEnv` does not copy env properties
When wrapping an env in `TransofrmedEnv`, its properties are changed

```python
from torchrl.envs import VmasEnv, RewardSum, TransformedEnv

if __name__ == "__main__":
    env = VmasEnv("balance", num_envs=3)
    print(env._allow_done_after_reset)
    t_env = TransformedEnv(env, RewardSum())
    print(t_env._allow_done_after_reset)

```
```
True
False
```

matteobettini:
Fixes #1806

The context is:

Some enviornments can be done immediately after a reset (e.g., some agents can be still dead after a reset in multiagent settings).
TorchRL by default does not allow this and will throw an error.
If you, however, would like to allow your environment to be done after reset, you can set `allow_done_after_reset` during construction of your environment.
`PettingZoo` and `VMAS` need this.

However, in #1806, it is shown that when a transform is added to the environment, this attribute changes.

In this PR, we fix this, by making the attribute of transformend enviornments mirror the one of the base environment.

### DIFF
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -335,6 +335,14 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         self._run_type_checks = run_type_checks
 
     @property
+    def allow_done_after_reset(self) -> bool:
+        return self._allow_done_after_reset
+
+    @allow_done_after_reset.setter
+    def allow_done_after_reset(self, allow_done_after_reset: bool):
+        self._allow_done_after_reset = allow_done_after_reset
+
+    @property
     def batch_size(self) -> torch.Size:
         _batch_size = self.__dict__["_batch_size"]
         if _batch_size is None:
@@ -1522,7 +1530,7 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
             if reset_value is not None:
                 for done_key in done_key_group:
                     done_val = tensordict_reset.get(done_key)
-                    if done_val[reset_value].any() and not self._allow_done_after_reset:
+                    if done_val[reset_value].any() and not self.allow_done_after_reset:
                         raise RuntimeError(
                             f"Env done entry '{done_key}' was (partially) True after reset on specified '_reset' dimensions. This is not allowed."
                         )
@@ -1540,7 +1548,7 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
                         # we set the done val to tensordict, to make sure that
                         # _update_during_reset does not pad the value
                         tensordict.set(done_key, done_val)
-            elif not self._allow_done_after_reset:
+            elif not self.allow_done_after_reset:
                 for done_key in done_key_group:
                     if tensordict_reset.get(done_key).any():
                         raise RuntimeError(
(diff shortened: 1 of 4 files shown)


vmoens (commented):
This is untested and I'm not sure where it would fail in practice, can you add a test (and some more context?)
As of now I think we should rather have a property that falls back onto the env_base attribute, such that changing the env_base changes the transformed env. That will avoid silent bugs down the line if things are modified dynamically (ie, you won't be able to modify the transformed env and if we end up in that situation we will deal with it accordingly)

matteobettini:
Yes I can do that.

matteobettini:
What I did is I now treat it exactly like `env._run_type_checks`

I made a public setter and getter for it. The getter of the transform calls the one of the base env and the setter throws an error.

torchrl/envs/common.py
```diff
@@ -334,6 +334,14 @@ def run_type_checks(self) -> bool:
     def run_type_checks(self, run_type_checks: bool) -> None:
         self._run_type_checks = run_type_checks
 
+    @property
+    def allow_done_after_reset(self) -> bool:
+        return self._allow_done_after_reset
+
+    @allow_done_after_reset.setter
+    def allow_done_after_reset(self, allow_done_after_reset: bool):
+        self._allow_done_after_reset = allow_done_after_reset
```
vmoens, line 343: why do we want this to be public? We should be super conservative with public attributes because once it's there we can't take it away!

matteobettini: So basically if I did not make it public, implementing the getter and setter solution would not be possible.

It is quite convoluted, but let me explain.

imagin i implement private getter and setters for `TransformedEnv` with the setter trhowing the error. Now, when in the constructor we call the `super().__init__()` https://github.com/pytorch/rl/blob/baea10b804f0836c6d95ce0d23e1e0906f49068f/torchrl/envs/transforms/transforms.py#L575
that will try to set `env._allow_done_on_reset` that however now calls the setter of the `TransofrmedEnv`, throwing an error.

matteobettini: This is the reason behind my first prroposed solution.

matteobettini: A possible solution is that the setter of the transformed env does not trhow an error but sets the variable in the base env. It seems to me that you wanted to avoid this though

vmoens: Why not: set it to `None`, `get` it, if the value is not `None` throw an error, otherwise set the private attribute.
Setting the private attribute will be ok as long as it's `None` (the `setter` of this property will be a no-op that passes silently if the value is `None` and throws an exception otherwise).
The `getter` on the other hand gets it from the env.

Like this we don't change anything, have the error where appropriate and keep the attribute private.

matteobettini: What i understand from the second sentence  might work.

i am not super able to understand what you mean in the first sentence.

but i can make the setter a no-op for None values if that is what you are suggesting, that might work. (Maybe just a bit confusing imo)

matteobettini: Implemented

vmoens (approved):
LGTM

PR #1813, opened 2024-01-18
[Doc] Update StepCounter doc

skandermoalla:
Corrected done update from intersection/conjunction (incorrect) to union/disjunction (correct)


vmoens (approved):
Thanks for this!

PR #1815, opened 2024-01-18
[Feature] Atari DQN dataset

vmoens:
Docstrings can be read [here](https://docs-preview.pytorch.org/pytorch/rl/1815/reference/generated/torchrl.data.datasets.AtariDQNExperienceReplay.html#torchrl.data.datasets.AtariDQNExperienceReplay)

cc @agarwl: do you think we could add this in https://github.com/mila-iqia/SGI/blob/master/src/offline_dataset.py after the release (next week)?


agarwl:
I think that's the repository of @MaxASchwarzer and you need to ask him about this. 

I can put a link to the pytorch dataset in the original offline RL repo.

PR #1817, opened 2024-01-19
[Feature] More flexibility in loading PettingZoo

matteobettini:
This change is a PettingZoo change which is BC-compatible to all the pettingzoo versions we already supported.

In addition, in case pettingzoo cannot load all modules, it leverages the new PettingZoo 1.24.3 functionalities to load only group of environments

This ameliorates the issue https://github.com/Farama-Foundation/PettingZoo/issues/1072

And fixes https://github.com/facebookresearch/BenchMARL/issues/41

cc @elliottower


vmoens (approved):
LGTM thanks

PR #1819, opened 2024-01-19
[WIP] Hindsight Experience Replay Transform

BY571:
## Description

Adds Hindsight Experience Replay ([HER](https://arxiv.org/pdf/1707.01495.pdf)) Transform 

## Motivation and Context

The first draft for the HER transform. However, I am not sure if it should be a `Transform` or if we create an extra `Augmentation` class as we are not transforming a single element in the tensordict but augmenting existing collection data. Could be interesting for future "data augmentation strategies", which I think we do not have until now. 

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -46,6 +46,7 @@ from torchrl.envs.transforms.utils import (
 )
 from torchrl.envs.utils import _sort_keys, _update_during_reset, step_mdp
 from torchrl.objectives.value.functional import reward2go
+from torchrl.objectives.value.utils import _get_num_per_traj, _split_and_pad_sequence
 
 try:
     from torchvision.transforms.functional import center_crop
@@ -5600,7 +5601,6 @@ class TimeMaxPool(Transform):
     def _reset(
         self, tensordict: TensorDictBase, tensordict_reset: TensorDictBase
     ) -> TensorDictBase:
-
         _reset = _get_reset(self.reset_key, tensordict)
         for in_key in self.in_keys:
             buffer_name = self._buffer_name(in_key)
@@ -6686,7 +6686,6 @@ class BurnInTransform(Transform):
         raise RuntimeError("BurnInTransform can only be appended to a ReplayBuffer.")
 
     def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
-
         if self.burn_in == 0:
             return tensordict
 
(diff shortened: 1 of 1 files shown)


torchrl/envs/transforms/transforms.py
```diff
@@ -6796,3 +6795,211 @@ def _reset(
...
+        rewards = []
+        for ach, des in zip(splitted_achieved_goals, new_goals):
+            rewards.append(vmap_rewards(ach[: des.shape[0], :], des))
+
+        cat_rewards = torch.cat(rewards).reshape(b, t, self.samples, -1).squeeze(-1)
+        cat_new_goals = torch.cat(new_goals).reshape(b, t, self.samples, -1)
+
+        augmentation_td = TensorDict(
+            {
+                "observation": sampled_td.get("observation").repeat_interleave(
```
BY571, line 6902: If we keep it a transform we probably need to specify all those tensordict keys ... Not sure what a better alternative would be. Any idea?

vmoens (commented):
@ahmed-touati suggested we use a sampler for this rather than a transform. I'm not strongly opinionated on the matter, mostly because I need more context on what we're trying to achieve here. 
Can you elaborate a bit more on what this transform does, maybe with a bunch of examples?

torchrl/envs/transforms/transforms.py
```diff
@@ -6796,3 +6795,211 @@ def _reset(
...
+            self.reward_function = reward_function
+
+        if not isinstance(samples, torch.Tensor):
+            samples = torch.tensor(samples)
+
+        self.register_buffer("samples", samples)
+
+    def _inv_call(self, tensordict: TensorDictBase) -> TensorDictBase:
+        augmentation_td = self.her_augmentation(tensordict)
+        return torch.cat([tensordict, augmentation_td], dim=0)
```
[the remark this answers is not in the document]
vmoens: Why not a transform? I think it's pretty neat to use a transform. Who said a transform had to change things in-place?
Our API to modify samples at writing time is to use either a transform or a different writer. If you think this can be achieved with a writer I'm on board. But I don't think there's anything wrong with the transform.

An advantage of using a writer instead os that it feels more natural (transforms can be used with envs unless specified otherwise, writers are dedicated to RBs)

BY571:
> @ahmed-touati suggested we use a sampler for this rather than a transform. I'm not strongly opinionated on the matter, mostly because I need more context on what we're trying to achieve here. Can you elaborate a bit more on what this transform does, maybe with a bunch of examples?

So HER is mainly used in goal-conditioned RL with sparse reward signals where the agent has to reach/achieve a goal state and only gets a reward (+1) when the goal state is achieved, otherwise no reward. The observation consists of three elements: the observation the agent sees, the state the agent had (could be x,y,z position), and the goal state the agent should reach (x,y,z). A typical task could be a robot that has to reach a goal position. The observation will include the agent position but its mostly added as additional information also helps here for understanding. 

Now as we have a sparse reward function most of the trajectories will have no learning signal for the agent as it might not be possible for the agent to reach the goal position randomly or by pure luck. 
What HER now does is that for each step in a trajectory that you want to add to the buffer, you sample a new goal state and pretend that this was the actual goal the agent had to reach.  

So lets say you have a real transition (obs, action, reward, done, next obs, achieved_position, goal_position) for this tuple you now want to sample a new goal_position and then calculate the reward based on this new goal_position and the real achieved_position. So you then add the real transition  (obs, action, reward, done, next obs, achieved_position, goal_position) but also the HER augmented transition  (obs, action, new_reward, done, new next obs, achieved_position, new_goal_position).  The sampling can happen in different ways but is not important for now. However, I think important will be that we need the reward function, Im not sure if we can pass it to the writer/sampler for the buffer, that's why my first thought was a transform. Most of the time the reward function might just be Euclidean distance but maybe for other tasks the user needs to provide a more sophisticated reward function.

vmoens:
> However, I think important will be that we need the reward function, Im not sure if we can pass it to the writer/sampler for the buffer, that's why my first thought was a transform.

Why not? I would guess that even if it's a complex nn.Module you can still do pretty much everything with a well tailored function (at least nothing less than with a transform).

vmoens:
Thanks for the context btw!

BY571:
> Why not? I would guess that even if it's a complex nn.Module you can still do pretty much everything with a well tailored function (at least nothing less than with a transform).

Revisiting this I think it would make much more sense to do it with a writer. We want to augment current incoming data with new sampled goal states and store them all together in the buffer. I think this would be generally a good way to add other data augmentation strategies with writer instead of transforms. Having a closer look right now on the writer classes and will update the code here

vmoens:
You could still transform your data before passing it to the writer, but not after

torchrl/envs/transforms/transforms.py
```diff
@@ -6796,3 +6795,211 @@ def _reset(
...
+        # get indices for each trajectory
+        idxs = self.generate_sample_idxs(trajectories)
+
+        # create new goals based idxs
+        new_goals = []
+        for i, ids in enumerate(idxs):
+            new_goals.append(splitted_achieved_goals[i][ids])
+
+        # calculate rewards given new desired goals and old achieved goals
+        vmap_rewards = torch.vmap(distance_reward_function)
```
dtsaras, line 6892: I think you wanna call ```self.reward_function``` instead of ```distance_reward_function```. Also maybe the ```reward_function``` should be a TensorDictModule such it can be more easily customized for a given environment. There is a ```torchrl.modules.VmapModule``` for wrapping TensorDictModules with with vmap.

torchrl/envs/transforms/transforms.py
```diff
@@ -6796,3 +6795,211 @@ def _reset(
...
+        rewards = []
+        for ach, des in zip(splitted_achieved_goals, new_goals):
+            rewards.append(vmap_rewards(ach[: des.shape[0], :], des))
+
+        cat_rewards = torch.cat(rewards).reshape(b, t, self.samples, -1).squeeze(-1)
+        cat_new_goals = torch.cat(new_goals).reshape(b, t, self.samples, -1)
+
+        augmentation_td = TensorDict(
```
dtsaras, line 6900: I think the ```augmentation_td``` should still maintain other metadata that are related to the state rather than selecting only the keys: observation, action, terminated, truncated, ...

dtsaras (commented):
While hindsight experience replay is pretty useful, I think it falls under the category of specialized algorithm rather than a building block @vmoens

vmoens:
Not sure about that one. 
We've had that request already 3 times but for 3 different purposes so if there's a way to make it a modular component of the lib I'd like to consider that above a script that is harder to reuse (and more error prone on the user side)

dtsaras:
> Are you planning to continue working on this PR? If not, I'd be happy to help out and look into finishing it myself.
> 
> 
> 
> Additionally, @vmoens, do you now have a clearer sense of whether this should be implemented as a transform, writer, or sampler? It would be great to make sure we're aligned on the best approach before moving forward.
> 
> 
> 
> Looking forward to your thoughts!

I have it an already complete version of it in a gist but I have not made a PR for it yet. Maybe you can have a look and provide feedback.

dtsaras:
> I really like the modular approach you've taken! That said, I have a few questions about some details of the implementation.
> 
> From what I understand, the reward function would be implemented as a `HERRewardTransform`. I wonder if this might be more complex than necessary compared to an alternative approach, such as passing a `compute_reward_fn: Callable[[torch.Tensor, torch.Tensor], torch.Tensor]` directly to the `HERTransform` class.
> 
> Could you share your thoughts on the reasoning behind including the `HERSubGoalSampler`? If we implement the common sampling methods (`random`, `final`, `future`), it seems like the additional complexity might not be strictly necessary – but I'd love to hear your perspective!

You are correct, the HERRewardTransform would be responsible to assign the rewards for all the intermediate states. While it's not necessary that it has to be its own special transform, it has to be "something" that reassigns the rewards to intermediate states of a trajectory. The reason I choose for it to be a Transform rather than a callable is to utilize the torchRL API. For example, the user does not need to reimplement the discounted reward function as Reward2GoTransform exists and multiple transforms can also be nicely composed into one.

The HERSubgoalSampler I have implemented only contains `[final (marked as last in the code), future]`. I did not implement random since it did not produce good results in the original paper if I recall correctly. The HERSubgoalSampler is only responsible for sampling states i.e. identifying the indices to be used for HERSubgoalAssigner.  Do you mind elaborating more on your concerns about the complexity?

Maybe I should create a PR and work on it together where you can make changes.

gtnv:
hi @BY571, is this still active? if not, would you like to close it for now?

PR #1820, opened 2024-01-19
[Deprecation] Deprecate in prep for release

vmoens:


torchrl/objectives/ppo.py
```diff
@@ -363,7 +361,7 @@ def critic(self):
 
     @property
     def actor_params(self):
-        logging.warning(
+        warnings.warning(
```
matteobettini, line 364:
```suggestion
        warnings.warn(
```

I think the syntax is this

PR #1822, opened 2024-01-20
[Feature] Submitit run script

albertbou92:
## Description

A script to run all examples and log results to a unified wandb project


albertbou92:
Some examples need to be updated:
1) redq does not  seem to log
2) dt is broken
3) dt online is broken too
4) rlhf: I get  RuntimeError: Current CUDA Device does not support bfloat16. Please switch dtype to float16. but I guess its fine.
5) bandits does not log
6) dreamer does not use the same config format and I believe @BY571 is working on a updated version.

I will look into 1-4, maybe the changes needed are minor. Maybe 5 and 6 required another PR.
Also some work will be needed in to better unify the logged values.

vmoens:
Let's put RLHF and bandits on the side but put a note in their respective folders under a README.md to say that these are not part of the benchmarked results for this release, wdyt?

albertbou92:
I think we should also leave out of the benchmark Dreamer and REDQ, these examples need an update, wdyt?

vmoens:
> I think we should also leave out of the benchmark Dreamer and REDQ, these examples need an update, wdyt?

Sounds good, for now let's focus on getting things to work for what we report in the paper.
You can leave out anything you want as long as we make some soft disclaimer in folders we don't cover. The message should be a more elaborate version of this: this script should work but probably needs a bit of manual tuning. Contributions are welcome, as always!

examples/dreamer/README.md
```diff
@@ -0,0 +1,7 @@
+# Dreamer example
+
+## Note:
+This example is not included in the benchmarked results of the current release. The intention is to include it in the
```
vmoens, line 4: can you add 
```suggestion
This example is not included in the benchmarked results of the current release (v0.3). The intention is to include it in the
```
I intend to look for all the v0.3 in the next release and address them. It's easier for us to keep track if we pinpoint precisely every part of the code that needs to be revised before the release

examples/redq/README.md
```diff
@@ -0,0 +1,7 @@
+# REDQ example
+
+## Note:
+This example is not included in the benchmarked results of the current release. The intention is to include it in the
```
vmoens, line 4: ditto

examples/rlhf/README.md
```diff
@@ -3,6 +3,12 @@
 This example uses RLHF (Reinforcement Learning with Human Feedback) to train a 
 language model to summarize Reddit posts.
 
+## Note:
+This example is not included in the benchmarked results of the current release. The intention is to include it in the
```
vmoens, line 7: ditto

vmoens (approved):
LGTM, amazing stuff

Let's see what the logs look like and we can merge!

benchmarks/sota-check/run_multiagent_sac.sh
```diff
@@ -0,0 +1,27 @@
...
+
+current_commit=$(git rev-parse --short HEAD)
+project_name="torchrl-example-check-$current_commit"
+group_name="marl_sac"
+
+export PYTHONPATH=$(dirname $(dirname $PWD))
+python $PYTHONPATH/examples/multiagent/sac.py \
+  logger.backend=wandb \
+  logger.project_name="$project_name" \
+  logger.group_name="marl_sac"
```
albertbou92, line 18:
```suggestion
  logger.group_name="$group_name"
```

benchmarks/sota-check/run_iql_online.sh
```diff
@@ -0,0 +1,26 @@
...
+#SBATCH --error=slurm_errors/iql_online_%j.txt
+
+current_commit=$(git rev-parse --short HEAD)
+project_name="torchrl-example-check-$current_commit"
+group_name="iql_online"
+export PYTHONPATH=$(dirname $(dirname $PWD))
+python $PYTHONPATH/examples/iql/iql_online.py \
+  logger.backend=wandb \
+  logger.project_name="$project_name" \
+  logger.group_name="iql_online"
```
albertbou92, line 17:
```suggestion
  logger.group_name="$group_name"
```

vmoens (approved):
LGTM

PR #1824, opened 2024-01-21
[Docs] Fix doc of ToTensorImage transforms.py

skandermoalla:
https://github.com/pytorch/rl/blob/a10cdbfc3643fd10c14432ab0dfb89fc2326a755/torchrl/envs/transforms/transforms.py#L1224


vmoens (commented):
Thanks for this!

PR #1828, opened 2024-01-22
pyproject.toml: remove unknown properties

GaetanLepage:
## Description

Removes irrelevant (and invalid) fields in the `pyproject.toml` file (`excludes`, `target-version` and `first_party_detection`).

## Motivation and Context

When installing this package from the source, I get the following error:
```
ERROR Failed to validate `build-system` in pyproject.toml: Unknown properties: excludes, target-version, first_party_detection
```
From my understanding, the following three tags are not meant to be under the `build-system` category.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,12 +3,3 @@ first_party_detection = false
 
 [build-system]
 requires = ["setuptools", "wheel", "torch", "ninja"]
-
-first_party_detection = false
-
-target-version = ["py38"]
-
-excludes = [
-    "gallery",
-    "tutorials",
-]


0001-pyproject.toml-remove-unknown-properties.patch
vmoens, line 1: Is this intentional?

vmoens (commented):
I think you added a file unintentionally but otherwise LGTM!

pyproject.toml
```diff
@@ -3,12 +3,3 @@ first_party_detection = false
 
 [build-system]
 requires = ["setuptools", "wheel", "torch", "ninja"]
-
-first_party_detection = false
-
-target-version = ["py38"]
-
-excludes = [
```
vmoens, line 11: I think those excludes make sense

vmoens (commented):
Upon reflection I'm not sure about the exclude

0001-pyproject.toml-remove-unknown-properties.patch
GaetanLepage: No, it was a mistake.
I just removed it !

pyproject.toml
GaetanLepage: Hmm probably but it does not belong to the `[build-system]` table.
After looking at the [PEP 518](https://peps.python.org/pep-0518) and [PEP 517](https://peps.python.org/pep-0517) specifications, it looks like only three keys are accepted in `[build-system]`: `requires`, `backend` and `backend-path`.

vmoens (approved):
LGTM

PR #1835, opened 2024-01-25
[Feature] Fine control over devices in collectors

vmoens:
- init
- amend

## Description

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -192,22 +192,15 @@ class DataCollectorBase(IterableDataset, metaclass=abc.ABCMeta):
                 Callable[[TensorDictBase], TensorDictBase],
             ]
         ] = None,
-        device: Optional[DEVICE_TYPING] = None,
         observation_spec: TensorSpec = None,
-    ) -> Tuple[TensorDictModule, torch.device, Union[None, Callable[[], dict]]]:
+    ) -> Tuple[TensorDictModule, Union[None, Callable[[], dict]]]:
         """Util method to get a policy and its device given the collector __init__ inputs.
-
-        From a policy and a device, assigns the self.device attribute to
-        the desired device and maps the policy onto it or (if the device is
-        ommitted) assigns the self.device attribute to the policy device.
 
         Args:
             create_env_fn (Callable or list of callables): an env creator
                 function (or a list of creators)
             create_env_kwargs (dictionary): kwargs for the env creator
             policy (TensorDictModule, optional): a policy to be used
-            device (int, str or torch.device, optional): device where to place
-                the policy
             observation_spec (TensorSpec, optional): spec of the observations
 
         """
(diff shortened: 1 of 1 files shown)


torchrl/collectors/collectors.py
```diff
@@ -363,35 +371,56 @@ class SyncDataCollector(DataCollectorBase):
...
-            :meth:`~.update_policy_weights_` method should be queried
-            at appropriate times during the training loop to accommodate for
-            the lag between parameter configuration at various times.
-            Defaults to ``None`` (i.e. policy is kept on its original device).
+             Defaults to ``-1`` (endless collector).
+        device (int, str or torch.device, optional): The generic device of the
+            collector. The ``device`` args fills any non-specified device: if
+            ``device`` is not ``None`` and any of ``storing_device``, ``policy_device`` or
+            ``env_device`` is not specified, its value will be set to ``device``.
+            Defaults to ``None`` (No default device).
```
matteobettini, line 388: Just to understand:

If my env is on cuda and i create a collector with `device=cpu` then my env is cast to cpu.
If, on the other hand, I want to preserve my env device, but have all other devices be cpu, I do `stroing_device=cpu, policy_device=cpu`

vmoens: the other option was that device would override all other devices but that's more messy and it makes all device kwargs useless as soon as you use device

torchrl/collectors/collectors.py
```diff
@@ -820,69 +940,98 @@ def rollout(self) -> TensorDictBase:
...
+                        self._shuttle.clone().to(self.storing_device, non_blocking=True)
+                    )
+                else:
+                    tensordicts.append(self._shuttle.clone())
+
+                self._shuttle.update(
+                    env_next_output.set(
+                        "collector", env_output.get("collector").copy()
+                    ),
+                    inplace=True,
```
vmoens, line 993: So this piece of code is pretty annoying but I don't see how to do otherwise.

Here's the pitch:

We want to support outputs that are on more than one device. Imagine your env is on cpu and your policy on cuda. You build your collector with
```python
collector = SyncCollector(..., policy_device="cuda", env_device="cpu")
```
You will have the `"action"`s on cuda and the rest on cpu.
Obviously the tensordict you will get in output won't have a device.

The way I make this happen is that I build a `_shuttle` that has preallocated tensors on the right devices and I update it inplace after each call to policy or env.step.
I do that because before you call the policy you need to cast your data to cuda using `shuttle.to(policy_device)` and conversely when you call env.step you need to cast it to cpu using `shuttle.to(env_device)` (note that in this case we could just cast the input data since the env tells us what input keys it needs -- which isn't always the case with the policy).

Because we need to cast stuff back and forth from devices, we need a copy of `shuttle` that has the proper devices allocated.

This also means that we need to clone the shuttle to save it in the stack of outputs (since modifications of shuttle must happen in-place). 


The problems are:
1. cloning induces an overhead that we'd like to avoid
2. modifying tensors in-place will make it hard to compile the collection loop when the time comes.

One way out would be to eagerly switch to a more efficient behaviour whenever policy_device and env_device match (no casting to device, no updating inplace). I think we'll still need the cloning but maybe not.

cc @matteobettini @skandermoalla

matteobettini: I might need to look more in the whole shuttle thing, but a first point that i think is important is:

In this more common case, which is that policy and env are on the same device, we need to avoid any cloning and casting at all costs.

I thought that this is what currently happens in collectors, so for me a hard requirement is that these PR changes do not slowdown any already existing use-case.

We can then discuss the efficiency of the new use-cases but I just want to see if we are at least aligned on this point.

vmoens: Alright! I got rid of all the clones and inplace modifications.
The throughput seems not to be affected by this PR.

PR #1848, opened 2024-01-29
[Feature, BugFix] Better thread control in penv and collectors

### LINKED ISSUE #1847: [BUG] Parallel envs override each others' num threads
Not sure this is the desired behavior. The parallel envs will run with the last defined num_threads. Not sure what is best to do by default in this case though.

## To Reproduce

```python
import torch
from torchrl.envs import EnvCreator, ParallelEnv


def run():
    env1 = ParallelEnv(8, EnvCreator(lambda: GymEnv("CartPole-v1")))
    print(torch.get_num_threads())

    env1.reset()
    print(torch.get_num_threads())

    env2 = ParallelEnv(3, EnvCreator(lambda: GymEnv("CartPole-v1")))
    print(torch.get_num_threads())

    env2.reset()
    print(torch.get_num_threads())

    env1.reset()
    print(torch.get_num_threads())


if __name__ == "__main__":
    run()
```

```bash
/usr/local/lib/python3.10/dist-packages/gymnasium/core.py:311: UserWarning: WARN: env.num_envs to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do `env.unwrapped.num_envs` for environment variables or `env.get_wrapper_attr('num_envs')` that will search the reminding wrappers.
  logger.warn(
/usr/local/lib/python3.10/dist-packages/gymnasium/core.py:311: UserWarning: WARN: env.reward_space to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do `env.unwrapped.reward_space` for environment variables or `env.get_wrapper_attr('reward_space')` that will search the reminding wrappers.
  logger.warn(
64
9
9
/usr/local/lib/python3.10/dist-packages/gymnasium/core.py:311: UserWarning: WARN: env.num_envs to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do `env.unwrapped.num_envs` for environment variables or `env.get_wrapper_attr('num_envs')` that will search the reminding wrappers.
  logger.warn(
/usr/local/lib/python3.10/dist-packages/gymnasium/core.py:311: UserWarning: WARN: env.reward_space to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do `env.unwrapped.reward_space` for environment variables or `env.get_wrapper_attr('reward_space')` that will search the reminding wrappers.
  logger.warn(
4
4
```

vmoens:
Closes #1847 

To test:
```python
import torch
from torchrl.envs import ParallelEnv, GymEnv

if __name__ == "__main__":
    print(torch.get_num_threads()) # 8
    
    env3 = ParallelEnv(3, lambda: GymEnv("Pendulum-v1"))
    env3.rollout(2) 
    
    print(torch.get_num_threads()) # 5
    
    env2 = ParallelEnv(2, lambda: GymEnv("Pendulum-v1"))
    env2.rollout(2)
    
    print(torch.get_num_threads()) # 3
    
    env2.close()
    
    print(torch.get_num_threads()) # 5
    
    env3.close()

    print(torch.get_num_threads()) # 8
```

cc @skandermoalla

### DIFF
--- a/test/test_env.py
+++ b/test/test_env.py
@@ -2329,30 +2329,60 @@ class TestTerminatedOrTruncated:
         assert not data["nested", "_reset"].any()
 
 
-@pytest.mark.skipif(
-    IS_OSX, reason="setting different threads across workeres can randomly fail on OSX."
-)
-def test_num_threads():
-    from torchrl.envs import batched_envs
-
-    _run_worker_pipe_shared_mem_save = batched_envs._run_worker_pipe_shared_mem
-    batched_envs._run_worker_pipe_shared_mem = decorate_thread_sub_func(
-        batched_envs._run_worker_pipe_shared_mem, num_threads=3
+class TestLibThreading:
+    @pytest.mark.skipif(
+        IS_OSX,
+        reason="setting different threads across workeres can randomly fail on OSX.",
     )
-    num_threads = torch.get_num_threads()
-    try:
-        env = ParallelEnv(
-            2, ContinuousActionVecMockEnv, num_sub_threads=3, num_threads=7
+    def test_num_threads(self):
+        from torchrl.envs import batched_envs
+
+        _run_worker_pipe_shared_mem_save = batched_envs._run_worker_pipe_shared_mem
+        batched_envs._run_worker_pipe_shared_mem = decorate_thread_sub_func(
+            batched_envs._run_worker_pipe_shared_mem, num_threads=3
         )
-        # We could test that the number of threads isn't changed until we start the procs.
-        # Even though it's unlikely that we have 7 threads, we still disable this for safety
-        # assert torch.get_num_threads() != 7
-        env.rollout(3)
-        assert torch.get_num_threads() == 7
-    finally:
-        # reset vals
-        batched_envs._run_worker_pipe_shared_mem = _run_worker_pipe_shared_mem_save
-        torch.set_num_threads(num_threads)
+        num_threads = torch.get_num_threads()
+        try:
+            env = ParallelEnv(
+                2, ContinuousActionVecMockEnv, num_sub_threads=3, num_threads=7
+            )
+            # We could test that the number of threads isn't changed until we start the procs.
+            # Even though it's unlikely that we have 7 threads, we still disable this for safety
+            # assert torch.get_num_threads() != 7
+            env.rollout(3)
+            assert torch.get_num_threads() == 7
+        finally:
+            # reset vals
+            batched_envs._run_worker_pipe_shared_mem = _run_worker_pipe_shared_mem_save
+            torch.set_num_threads(num_threads)
+
+    @pytest.mark.skipif(
+        IS_OSX,
+        reason="setting different threads across workeres can randomly fail on OSX.",
+    )
+    def test_auto_num_threads(self):
+        init_threads = torch.get_num_threads()
+
+        try:
+            env3 = ParallelEnv(3, lambda: GymEnv("Pendulum-v1"))
+            env3.rollout(2)
+
+            assert torch.get_num_threads() == max(1, init_threads - 3)
+
+            env2 = ParallelEnv(2, lambda: GymEnv("Pendulum-v1"))
+            env2.rollout(2)
+
+            assert torch.get_num_threads() == max(1, init_threads - 5)
+
+            env2.close()
+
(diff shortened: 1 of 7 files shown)


skandermoalla:
Replying here about the issue too.

Looks like a valid idea to me. But I don't think I have enough breadth to predict if it will impact other things.
Does it play nicely with `OMP_NUM_THREADS=K` if I want PyTorch to use K threads by default in a process?

Also, why does the main process need N+1 threads when running N parallel environments? Aren't these processes and not threads? So independent from the number of threads of the main process?

vmoens:
> Does it play nicely with OMP_NUM_THREADS=K if I want PyTorch to use K threads by default in a process?

`OMP_NUM_THREADS=K` is equal to torch.get_num_threads, so this represents the number of threads you can consume with this approach. 

> Also, why does the main process need N+1 threads when running N parallel environments?

That won't be the case anymore: this approach gives 1 thread by default to each node and allocates the rest to main.

In other words, the plan is that if you have one multi-async collector with 8 collectors, each having 4 workers, you will have 32 less threads in the main process.

test/test_env.py
```diff
@@ -2329,30 +2329,60 @@ def test_terminated_or_truncated_spec(self):
...
-        env.rollout(3)
-        assert torch.get_num_threads() == 7
-    finally:
-        # reset vals
-        batched_envs._run_worker_pipe_shared_mem = _run_worker_pipe_shared_mem_save
-        torch.set_num_threads(num_threads)
+        num_threads = torch.get_num_threads()
+        try:
+            env = ParallelEnv(
+                2, ContinuousActionVecMockEnv, num_sub_threads=3, num_threads=7
```
skandermoalla, line 2347: Why would this need more than one thread if the envs are running on different processes?
And why would that set the num_threads of the current process?

torchrl/envs/libs/gym.py
```diff
@@ -306,16 +306,16 @@ def _gym_to_torchrl_spec_transform(
             shape = torch.Size([1])
         if dtype is None:
             dtype = numpy_to_torch_dtype_dict[spec.dtype]
-        low = torch.tensor(spec.low, device=device, dtype=dtype)
-        high = torch.tensor(spec.high, device=device, dtype=dtype)
+        low = torch.as_tensor(spec.low, device=device, dtype=dtype)
+        high = torch.as_tensor(spec.high, device=device, dtype=dtype)
```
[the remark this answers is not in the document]
vmoens: they were giving me warnings, but you're right i will make another PR with them

test/test_env.py
vmoens: That's an old test just to check that if you manually assign the number of threads things go smoothly.
In practice if you don't do anything your parallel env will take 2 threads total (get_num_threads will decrease by 2).
The logic is that we automatically take away threads from the main process if we assign them to sub-process unless you tell me how many threads you want in this process (in this example, 7)

skandermoalla: Yeah, I felt that `num_threads` shouldn't exist as an option. It seems useless/misleading to use it.

torchrl/collectors/collectors.py
```diff
@@ -1595,7 +1606,19 @@ def _queue_len(self) -> int:
...
+        if self.num_threads is None:
+            import torchrl
+
+            total_workers = self._total_workers_from_env(self.create_env_fn)
+            self.num_threads = max(
+                1, torchrl._THREAD_POOL - total_workers
+            )  # 1 more thread for this proc
+
         torch.set_num_threads(self.num_threads)
+        assert torch.get_num_threads() == self.num_threads
```
skandermoalla, line 1618: @vmoens Not sure this assert is wanted here. It will probably break with on OSX where just doing this breaks:

```python
print(torch.get_num_threads())
    torch.set_num_threads(5)
    print(torch.get_num_threads())
    torch.set_num_threads(1)
    print(torch.get_num_threads())

```
gives
```bash
8
5
/Users/moalla/projects/claire/po-dynamics/tests/issue_num_threads_collector.py:27: UserWarning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (Triggered internally at /Users/runner/work/_temp/anaconda/conda-bld/pytorch_1702400234613/work/aten/src/ATen/ParallelNative.cpp:230.)
  torch.set_num_threads(1)
5
```

PR #1857, opened 2024-01-31
[BugFix] Fix warning call

matteobettini:
Deprecation warnings are usually called with
https://github.com/pytorch/rl/blob/b0653c44fb6e3776653973861a4732c89801173e/torchrl/objectives/ppo.py#L357

This PR modifies the warning in https://github.com/pytorch/rl/blob/b0653c44fb6e3776653973861a4732c89801173e/torchrl/objectives/ppo.py#L366 to the right syntax, as the current call will not expect the `category` attribute

### DIFF
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -5,7 +5,6 @@
 from __future__ import annotations
 
 import contextlib
-import logging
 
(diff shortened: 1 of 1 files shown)


vmoens:
part of #1820

PR #1866, opened 2024-02-01
[BugFix] Fix update in serial / parallel env

vmoens:


vmoens:
@skandermoalla I updated a test in collectors that will check that everything's working for parallel env. It fails on main

PR #1875, opened 2024-02-06
[Feature] Add PrioritizedSliceSampler

Cadene:
## Description

Add PrioritizedSliceSampler which inherits from PrioritizedSampler and SliceSampler.

Contrary to PrioritizedSampler which samples steps with a priority weight for each, PrioritizedSliceSampler samples slices with a priority weight for each. Implementation-wise, it only samples the start index of each slice, while discarding the indices at the end of each episode since they can't form long enough slices.

Contrary to SliceSampler which first samples trajectories, then randomly samples slices inside each trajectory, PrioritizedSliceSampler directly samples slices. It is easier to implement that way, but as a consequence PrioritizedSliceSampler more often samples slices from longer trajectories than shorter trajectories when priority weights are uniform. Nevertheless, since the priority weights are updated along training, slices from longer trajectories shouldn't be oversampled after a while.

## Motivation and Context

This type of sampling is used in the literature:
https://github.com/fyhMer/fowm/blob/main/src/algorithm/helper.py#L504-L510
https://github.com/fyhMer/fowm/blob/main/src/algorithm/tdmpc.py#L334-L337

Addresses this feature request: https://github.com/pytorch/rl/issues/1876

## Types of changes

## Checklist

### DIFF
--- a/torchrl/data/replay_buffers/samplers.py
+++ b/torchrl/data/replay_buffers/samplers.py
@@ -1082,7 +1082,56 @@ class SliceSamplerWithoutReplacement(SliceSampler, SamplerWithoutReplacement):
 class PrioritizedSliceSampler(SliceSampler, PrioritizedSampler):
     """Samples slices of data along the first dimension, given start and stop signals, using prioritized sampling.
 
-    For more info see: SliceSampler and PrioritizedSampler
+    This class samples sub-trajectories with replacement following a priority weighting presented in "Schaul, T.; Quan, J.; Antonoglou, I.; and Silver, D. 2015.
+        Prioritized experience replay."
+        (https://arxiv.org/abs/1511.05952)
+
+    For more info see :class:`~torchrl.data.replay_buffers.samplers.SliceSampler` and :class:`~torchrl.data.replay_buffers.samplers.PrioritizedSampler`.
+
+    Args:
+        alpha (float): exponent α determines how much prioritization is used,
+            with α = 0 corresponding to the uniform case.
+        beta (float): importance sampling negative exponent.
+        eps (float, optional): delta added to the priorities to ensure that the buffer
+            does not contain null priorities. Defaults to 1e-8.
+        reduction (str, optional): the reduction method for multidimensional
+            tensordicts (ie stored trajectory). Can be one of "max", "min",
+            "median" or "mean".
+
+    Keyword Args:
+        num_slices (int): the number of slices to be sampled. The batch-size
+            must be greater or equal to the ``num_slices`` argument. Exclusive
+            with ``slice_len``.
+        slice_len (int): the length of the slices to be sampled. The batch-size
+            must be greater or equal to the ``slice_len`` argument and divisible
+            by it. Exclusive with ``num_slices``.
+        end_key (NestedKey, optional): the key indicating the end of a
+            trajectory (or episode). Defaults to ``("next", "done")``.
+        traj_key (NestedKey, optional): the key indicating the trajectories.
+            Defaults to ``"episode"`` (commonly used across datasets in TorchRL).
+        ends (torch.Tensor, optional): a 1d boolean tensor containing the end of run signals.
+            To be used whenever the ``end_key`` or ``traj_key`` is expensive to get,
+            or when this signal is readily available. Must be used with ``cache_values=True``
+            and cannot be used in conjunction with ``end_key`` or ``traj_key``.
+        trajectories (torch.Tensor, optional): a 1d integer tensor containing the run ids.
+            To be used whenever the ``end_key`` or ``traj_key`` is expensive to get,
+            or when this signal is readily available. Must be used with ``cache_values=True``
+            and cannot be used in conjunction with ``end_key`` or ``traj_key``.
+        cache_values (bool, optional): to be used with static datasets.
+            Will cache the start and end signal of the trajectory.
+        truncated_key (NestedKey, optional): If not ``None``, this argument
+            indicates where a truncated signal should be written in the output
+            data. This is used to indicate to value estimators where the provided
+            trajectory breaks. Defaults to ``("next", "truncated")``.
+            This feature only works with :class:`~torchrl.data.replay_buffers.TensorDictReplayBuffer`
+            instances (otherwise the truncated key is returned in the info dictionary
+            returned by the :meth:`~torchrl.data.replay_buffers.ReplayBuffer.sample` method).
+        strict_length (bool, optional): if ``False``, trajectories of length
+            shorter than `slice_len` (or `batch_size // num_slices`) will be
+            allowed to appear in the batch.
+            Be mindful that this can result in effective `batch_size`  shorter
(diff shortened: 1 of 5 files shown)


torchrl/data/replay_buffers/samplers.py
```diff
@@ -867,7 +867,7 @@ def _sample_slices(
                 truncated[seq_length.cumsum(0) - 1] = 1
             traj_terminated = stop_idx[traj_idx] == start_idx[traj_idx] + seq_length - 1
             terminated = torch.zeros_like(truncated)
-            if terminated.any():
+            if traj_terminated.any():
```
Cadene, line 870: Note: Bugfix discussed with @vmoens

torchrl/data/replay_buffers/samplers.py
```diff
@@ -1079,6 +1079,197 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
         return SamplerWithoutReplacement.load_state_dict(self, state_dict)
 
 
+class PrioritizedSliceSampler(SliceSampler, PrioritizedSampler):
```
Cadene, line 1082: Note: Ideally I would like to avoid multi-inheritance, since it is prone to generate spaghetti code, but this pattern was used elsewhere in the code so I kept the same logic: `SliceSamplerWithoutReplacement(SliceSampler, SamplerWithoutReplacement)`

vmoens:
Amazing! Will do a proper review shortly. Any chance we can get the bugfix in a separate PR to put it in the minor release?

torchrl/data/replay_buffers/samplers.py
vmoens: I agree I don't like it either. The other option is to copy paste methods, or port them out of the classes and make the classes rely on external functions. 
To me if it seems organic it's ok, even if it isn't the best feature in python

torchrl/data/replay_buffers/samplers.py
```diff
@@ -1079,6 +1079,197 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+    For more info see :class:`~torchrl.data.replay_buffers.samplers.SliceSampler` and :class:`~torchrl.data.replay_buffers.samplers.PrioritizedSampler`.
+
+    Args:
+        alpha (float): exponent α determines how much prioritization is used,
+            with α = 0 corresponding to the uniform case.
+        beta (float): importance sampling negative exponent.
+        eps (float, optional): delta added to the priorities to ensure that the buffer
+            does not contain null priorities. Defaults to 1e-8.
+        reduction (str, optional): the reduction method for multidimensional
+            tensordicts (ie stored trajectory). Can be one of "max", "min",
```
vmoens, line 1098:
```suggestion
            tensordicts (i.e., stored trajectory). Can be one of "max", "min",
```

torchrl/data/replay_buffers/samplers.py
```diff
@@ -1079,6 +1079,197 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+
+        # force to not sample index at the end of a trajectory.
+        # no need to update self._min_tree.
+        self._sum_tree[preceding_stop_idx] = 0.0
+
+        starts, info = PrioritizedSampler.sample(
+            self, storage=storage, batch_size=batch_size // seq_length
+        )
+        starts = torch.from_numpy(starts).to(device=lengths.device)
+        index = self._tensor_slices_from_startend(seq_length, starts)
```
vmoens, line 1221: note to self: we should avoid returning numpy arrays and stick to torch

cc @albertbou92

torchrl/data/replay_buffers/samplers.py
```diff
@@ -1079,6 +1079,197 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+        assert index.shape[0] == batch_size
```
vmoens, line 1222: Let's remove this or make it a RuntimeError
We try to avoid asserts in torchrl except in tests, as if things break they don't provide a lot of insight to the users

torchrl/data/replay_buffers/samplers.py
```diff
@@ -1079,6 +1079,197 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+            done = terminated | truncated
+
+            info.update(
+                {
+                    truncated_key: truncated,
+                    done_key: done,
+                    terminated_key: terminated,
+                }
+            )
+        return index.to(torch.long), info
```
vmoens, line 1254: we should check if there isn't a way to enforce this dtype earlier and avoid casting

torchrl/data/replay_buffers/samplers.py
```diff
@@ -1079,6 +1079,197 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+            trajectory breaks. Defaults to ``("next", "truncated")``.
+            This feature only works with :class:`~torchrl.data.replay_buffers.TensorDictReplayBuffer`
+            instances (otherwise the truncated key is returned in the info dictionary
+            returned by the :meth:`~torchrl.data.replay_buffers.ReplayBuffer.sample` method).
+        strict_length (bool, optional): if ``False``, trajectories of length
+            shorter than `slice_len` (or `batch_size // num_slices`) will be
+            allowed to appear in the batch.
+            Be mindful that this can result in effective `batch_size`  shorter
+            than the one asked for! Trajectories can be split using
+            :func:`torchrl.collectors.split_trajectories`. Defaults to ``True``.
```
vmoens, line 1134: It's always nice to have an example for people to play with. I usually do an edulcorated version of a test I wrote

vmoens (approved):
Wonderful! Thanks it's a super useful feature
I left a couple of comments, most are just suggestions so don't feel like you need to address them all.

If we can just move the bugfix to a separate PR, I'll merge that one straight away and then we can move to this.

torchrl/data/replay_buffers/samplers.py
Cadene: FYI I borrowed this logic from SliceSampler

https://github.com/pytorch/rl/blob/144f5470e9b0358d33cc8ac346f0fb1a07839067/torchrl/data/replay_buffers/samplers.py#L932

Cadene: I checked on my side, and index is already a torch.long, so the cast is a no-op

torchrl/data/replay_buffers/samplers.py
Cadene: Done

torchrl/data/replay_buffers/samplers.py
Cadene: Done. Also added the example as a test

Cadene:
> Amazing! Will do a proper review shortly. Any chance we can get the bugfix in a separate PR to put it in the minor release?

@vmoens Done ;) https://github.com/pytorch/rl/pull/1884

PR #1890, opened 2024-02-09
[Feature] Add reduction parameter to On-Policy losses.

albertbou92:
## Description

This PR introduces a reduction option to the on-policy losses, similar to how Torch does it (e.g. https://pytorch.org/docs/stable/generated/torch.nn.MSELoss.html).

The ideas is to validate the approach for on-policy losses, the then move on to replicate it in the other losses.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -5960,6 +5960,7 @@ class TestPPO(LossModuleTestBase):
             value,
             loss_critic_type="l2",
             separate_losses=True,
+            reduction=reduction,
         )
 
         if advantage is not None:
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -6872,8 +6872,7 @@ class TestA2C(LossModuleTestBase):
     )
     @pytest.mark.parametrize("advantage", ("gae", "vtrace", None))
     @pytest.mark.parametrize("device", get_default_devices())
-    @pytest.mark.parametrize("reduction", [None, "mean", "sum"])
-    def test_a2c_tensordict_keys_run(self, device, advantage, td_est, reduction):
+    def test_a2c_tensordict_keys_run(self, device, advantage, td_est):
         """Test A2C loss module with non-default tensordict keys."""
         torch.manual_seed(self.seed)
         gradient_mode = True
@@ -6943,9 +6942,6 @@ class TestA2C(LossModuleTestBase):
                 loss_fn.make_value_estimator(td_est)
 
         loss = loss_fn(td)
-        if reduction is None:
-            assert loss.batch_size == td.batch_size
-            loss = loss.apply(lambda x: x.float().mean(), batch_size=[])
         loss_critic = loss["loss_critic"]
         loss_objective = loss["loss_objective"] + loss.get("loss_entropy", 0.0)
         loss_critic.backward(retain_graph=True)
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -5873,7 +5873,7 @@ class TestPPO(LossModuleTestBase):
         loss = loss_fn(td)
         if reduction is None:
             assert loss.batch_size == td.batch_size
-            loss.apply(lambda x: x.float().mean(), batch_size=[])
+            loss = loss.apply(lambda x: x.float().mean(), batch_size=[])
 
         loss_critic = loss["loss_critic"]
         loss_objective = loss["loss_objective"] + loss.get("loss_entropy", 0.0)
@@ -5967,7 +5967,7 @@ class TestPPO(LossModuleTestBase):
         loss = loss_fn(td)
         if reduction is None:
             assert loss.batch_size == td.batch_size
-            loss.apply(lambda x: x.float().mean(), batch_size=[])
+            loss = loss.apply(lambda x: x.float().mean(), batch_size=[])
 
         loss_critic = loss["loss_critic"]
         loss_objective = loss["loss_objective"] + loss.get("loss_entropy", 0.0)
@@ -6068,7 +6068,7 @@ class TestPPO(LossModuleTestBase):
         loss = loss_fn(td).exclude("entropy")
         if reduction is None:
             assert loss.batch_size == td.batch_size
-            loss.apply(lambda x: x.float().mean(), batch_size=[])
+            loss = loss.apply(lambda x: x.float().mean(), batch_size=[])
 
         sum(val for key, val in loss.items() if key.startswith("loss_")).backward()
         grad = TensorDict(dict(model.named_parameters()), []).apply(
@@ -6144,7 +6144,7 @@ class TestPPO(LossModuleTestBase):
             loss = loss_fn(td)
             if reduction is None:
                 assert loss.batch_size == td.batch_size
-                loss.apply(lambda x: x.float().mean(), batch_size=[])
+                loss = loss.apply(lambda x: x.float().mean(), batch_size=[])
 
         loss_critic = loss["loss_critic"]
         loss_objective = loss["loss_objective"] + loss.get("loss_entropy", 0.0)
@@ -6304,7 +6304,7 @@ class TestPPO(LossModuleTestBase):
         loss = loss_fn(td)
         if reduction is None:
             assert loss.batch_size == td.batch_size
-            loss.apply(lambda x: x.float().mean(), batch_size=[])
+            loss = loss.apply(lambda x: x.float().mean(), batch_size=[])
 
         loss_critic = loss["loss_critic"]
         loss_objective = loss["loss_objective"] + loss.get("loss_entropy", 0.0)
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -6080,7 +6080,7 @@ class TestPPO(LossModuleTestBase):
         loss2 = loss_fn2(td).exclude("entropy")
         if reduction is None:
             assert loss2.batch_size == td.batch_size
-            loss2.apply(lambda x: x.float().mean(), batch_size=[])
+            loss2 = loss2.apply(lambda x: x.float().mean(), batch_size=[])
 
         model.zero_grad()
         sum(val for key, val in loss2.items() if key.startswith("loss_")).backward()
@@ -6582,7 +6582,8 @@ class TestA2C(LossModuleTestBase):
     @pytest.mark.parametrize("device", get_default_devices())
     @pytest.mark.parametrize("td_est", list(ValueEstimators) + [None])
     @pytest.mark.parametrize("functional", (True, False))
-    def test_a2c(self, device, gradient_mode, advantage, td_est, functional):
+    @pytest.mark.parametrize("reduction", [None, "mean", "sum"])
+    def test_a2c(self, device, gradient_mode, advantage, td_est, functional, reduction):
         torch.manual_seed(self.seed)
         td = self._create_seq_mock_data_a2c(device=device)
 
@@ -6612,7 +6613,13 @@ class TestA2C(LossModuleTestBase):
         else:
             raise NotImplementedError
 
-        loss_fn = A2CLoss(actor, value, loss_critic_type="l2", functional=functional)
+        loss_fn = A2CLoss(
+            actor,
+            value,
+            loss_critic_type="l2",
+            functional=functional,
+            reduction=reduction,
+        )
 
         # Check error is raised when actions require grads
         td["action"].requires_grad = True
@@ -6629,6 +6636,9 @@ class TestA2C(LossModuleTestBase):
         elif td_est is not None:
             loss_fn.make_value_estimator(td_est)
         loss = loss_fn(td)
+        if reduction is None:
+            assert loss.batch_size == td.batch_size
+            loss = loss.apply(lambda x: x.float().mean(), batch_size=[])
         loss_critic = loss["loss_critic"]
         loss_objective = loss["loss_objective"] + loss.get("loss_entropy", 0.0)
         loss_critic.backward(retain_graph=True)
@@ -6671,11 +6681,15 @@ class TestA2C(LossModuleTestBase):
         loss_fn2.load_state_dict(sd)
 
     @pytest.mark.parametrize("separate_losses", [False, True])
-    def test_a2c_separate_losses(self, separate_losses):
+    @pytest.mark.parametrize("reduction", [None, "mean", "sum"])
+    def test_a2c_separate_losses(self, separate_losses, reduction):
         torch.manual_seed(self.seed)
         actor, critic, common, td = self._create_mock_common_layer_setup()
         loss_fn = A2CLoss(
-            actor_network=actor, critic_network=critic, separate_losses=separate_losses
+            actor_network=actor,
+            critic_network=critic,
+            separate_losses=separate_losses,
+            reduction=reduction,
         )
 
         # Check error is raised when actions require grads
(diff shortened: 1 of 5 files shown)


test/test_cost.py
```diff
@@ -155,6 +155,10 @@
 )
 
 
+# Increase recursion limit, some objectives have more than 1000 tests
+sys.setrecursionlimit(2000)
```
albertbou92, line 159: I needed to increase this. Otherwise classes with more than 1000 tests fail after hitting this number

torchrl/objectives/a2c.py
```diff
@@ -234,6 +240,7 @@ def __init__(
         functional: bool = True,
         actor: ProbabilisticTensorDictSequential = None,
         critic: ProbabilisticTensorDictSequential = None,
+        reduction: str = "mean",
```
vmoens, line 243: I would use `None` as a default, it can come in handy if some day we change the default
smth like `if reduction is None: reduction="mean"` later in the constructor.

torchrl/objectives/a2c.py
```diff
@@ -388,7 +396,7 @@ def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
             entropy = dist.entropy()
         except NotImplementedError:
             x = dist.rsample((self.samples_mc_entropy,))
-            entropy = -dist.log_prob(x)
+            entropy = -dist.log_prob(x).mean(0)
```
vmoens, line 399: why 0? What if the input tensordict has 2 or 3 dims? Don't we want to mean across the number of dims of the input tensordict?

torchrl/objectives/a2c.py
```diff
@@ -457,14 +465,16 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
-            td_out.set("loss_entropy", -self.entropy_coef * entropy.mean())
+            td_out.set("entropy", entropy.detach())  # for logging
+            td_out.set("loss_entropy", -self.entropy_coef * entropy)
         if self.critic_coef:
-            loss_critic = self.loss_critic(tensordict).mean()
-            td_out.set("loss_critic", loss_critic.mean())
+            loss_critic = self.loss_critic(tensordict)
+            td_out.set("loss_critic", loss_critic)
+        if self.reduction is not None:
+            td_out = td_out.apply(lambda x: _reduce(x, self.reduction), batch_size=[])
```
vmoens, line 477:
```suggestion
            td_out = td_out.apply(functools.partial(_reduce, reduction=self.reduction), batch_size=[])
```

torchrl/objectives/ppo.py
```diff
@@ -85,6 +86,10 @@ class PPOLoss(LossModule):
             Functionalizing permits features like meta-RL, but makes it
             impossible to use distributed models (DDP, FSDP, ...) and comes
             with a little cost. Defaults to ``True``.
+        reduction (str, optional): Specifies the reduction to apply to the output:
+            ``"none"`` | ``"mean"`` | ``"sum"``. ``"none"``: no reduction will be applied,
+            ``"mean"``: the sum of the output will be divided by the number of
+            elements in the output, ``"sum"``: the output will be summed. Default: ``"mean"``.
```
vmoens, line 92: ditto

torchrl/objectives/ppo.py
```diff
@@ -278,6 +283,7 @@ def __init__(
         functional: bool = True,
         actor: ProbabilisticTensorDictSequential = None,
         critic: ProbabilisticTensorDictSequential = None,
+        reduction: str = "mean",
```
vmoens, line 286: ditto

torchrl/objectives/ppo.py
```diff
@@ -444,7 +451,7 @@ def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
             entropy = dist.entropy()
         except NotImplementedError:
             x = dist.rsample((self.samples_mc_entropy,))
-            entropy = -dist.log_prob(x)
+            entropy = -dist.log_prob(x).mean(0)
```
vmoens, line 454: ditto

torchrl/objectives/ppo.py
```diff
@@ -530,15 +537,20 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
-            td_out.set("loss_entropy", -self.entropy_coef * entropy.mean())
+            td_out.set("entropy", entropy.detach())  # for logging
+            td_out.set("loss_entropy", -self.entropy_coef * entropy)
         if self.critic_coef:
-            loss_critic = self.loss_critic(tensordict).mean()
-            td_out.set("loss_critic", loss_critic.mean())
+            loss_critic = self.loss_critic(tensordict)
+            td_out.set("loss_critic", loss_critic)
+        if self.reduction is not None:
+            td_out = td_out.apply(lambda x: _reduce(x, self.reduction), batch_size=[])
```
vmoens, line 552:
```suggestion
            td_out = td_out.apply(functools.partial(_reduce, reduction=self.reduction), batch_size=[])
```

torchrl/objectives/ppo.py
```diff
@@ -692,6 +708,7 @@ def __init__(
         normalize_advantage: bool = True,
         gamma: float = None,
         separate_losses: bool = False,
+        reduction: str = "mean",
```
vmoens, line 711: ditto

torchrl/objectives/ppo.py
```diff
@@ -634,6 +646,10 @@ class ClipPPOLoss(PPOLoss):
             Functionalizing permits features like meta-RL, but makes it
             impossible to use distributed models (DDP, FSDP, ...) and comes
             with a little cost. Defaults to ``True``.
+        reduction (str, optional): Specifies the reduction to apply to the output:
+            ``"none"`` | ``"mean"`` | ``"sum"``. ``"none"``: no reduction will be applied,
+            ``"mean"``: the sum of the output will be divided by the number of
+            elements in the output, ``"sum"``: the output will be summed. Default: ``"mean"``.
```
vmoens, line 652: ditto

torchrl/objectives/ppo.py
```diff
@@ -764,16 +782,19 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+            td_out.set("entropy", entropy.detach())  # for logging
+            td_out.set("loss_entropy", -self.entropy_coef * entropy)
         if self.critic_coef:
             loss_critic = self.loss_critic(tensordict)
-            td_out.set("loss_critic", loss_critic.mean())
-        td_out.set("ESS", ess.mean() / batch)
+            td_out.set("loss_critic", loss_critic)
+
+        if self.reduction is not None:
+            td_out = td_out.apply(lambda x: _reduce(x, self.reduction), batch_size=[])
```
vmoens, line 796:
```suggestion
            td_out = td_out.apply(functools.partial(_reduce, reduction=self.reduction), batch_size=[])
```

torchrl/objectives/ppo.py
```diff
@@ -764,16 +782,19 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+            td_out.set("ESS", _reduce(ess, self.reduction) / batch)
```
vmoens, line 797: it's a bit problematic that the output depends on the reduction... 
I'm not sure that having the batch-size set is going to work across all losses anyway (some losses subsample a part of data to compute part of the loss).
I agree it isn't super clear though but maybe the best we can do to keep things generic.

torchrl/objectives/ppo.py
```diff
@@ -832,7 +853,10 @@ class KLPENPPOLoss(PPOLoss):
             Functionalizing permits features like meta-RL, but makes it
             impossible to use distributed models (DDP, FSDP, ...) and comes
             with a little cost. Defaults to ``True``.
-
+        reduction (str, optional): Specifies the reduction to apply to the output:
+            ``"none"`` | ``"mean"`` | ``"sum"``. ``"none"``: no reduction will be applied,
+            ``"mean"``: the sum of the output will be divided by the number of
+            elements in the output, ``"sum"``: the output will be summed. Default: ``"mean"``.
```
vmoens, line 859: ditto

torchrl/objectives/ppo.py
```diff
@@ -895,6 +919,7 @@ def __init__(
         normalize_advantage: bool = True,
         gamma: float = None,
         separate_losses: bool = False,
+        reduction: str = "mean",
```
vmoens, line 922: ditto

torchrl/objectives/ppo.py
```diff
@@ -978,20 +1004,21 @@ def forward(self, tensordict: TensorDictBase) -> TensorDict:
...
-            td_out.set("loss_entropy", -self.entropy_coef * entropy.mean())
-
+            td_out.set("entropy", entropy.detach())  # for logging
+            td_out.set("loss_entropy", -self.entropy_coef * entropy)
         if self.critic_coef:
             loss_critic = self.loss_critic(tensordict)
-            td_out.set("loss_critic", loss_critic.mean())
+            td_out.set("loss_critic", loss_critic)
+        if self.reduction is not None:
+            td_out = td_out.apply(lambda x: _reduce(x, self.reduction), batch_size=[])
```
vmoens, line 1021:
```suggestion
            td_out = td_out.apply(functools.partial(_reduce, reduction=self.reduction), batch_size=[])
```

torchrl/objectives/reinforce.py
```diff
@@ -60,6 +62,10 @@ class ReinforceLoss(LossModule):
             Functionalizing permits features like meta-RL, but makes it
             impossible to use distributed models (DDP, FSDP, ...) and comes
             with a little cost. Defaults to ``True``.
+        reduction (str, optional): Specifies the reduction to apply to the output:
+            ``"none"`` | ``"mean"`` | ``"sum"``. ``"none"``: no reduction will be applied,
+            ``"mean"``: the sum of the output will be divided by the number of
+            elements in the output, ``"sum"``: the output will be summed. Default: ``"mean"``.
```
vmoens, line 68: ditto

torchrl/objectives/reinforce.py
```diff
@@ -388,10 +396,13 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
-        loss_actor = loss_actor.mean()
-        td_out = TensorDict({"loss_actor": loss_actor}, [])
+        td_out = TensorDict(
+            {"loss_actor": loss_actor}, batch_size=tensordict.batch_size
+        )
 
-        td_out.set("loss_value", self.loss_critic(tensordict).mean())
+        td_out.set("loss_value", self.loss_critic(tensordict))
+        if self.reduction is not None:
+            td_out = td_out.apply(lambda x: _reduce(x, self.reduction), batch_size=[])
```
vmoens, line 405:
```suggestion
            td_out = td_out.apply(functools.partial(_reduce, reduction=self.reduction), batch_size=[])
```

torchrl/objectives/reinforce.py
```diff
@@ -223,6 +229,7 @@ def __init__(
         functional: bool = True,
         actor: ProbabilisticTensorDictSequential = None,
         critic: ProbabilisticTensorDictSequential = None,
+        reduction: str = "mean",
```
vmoens, line 232: ditto

vmoens (commented):
Great stuff!
I wonder if long term we should not structure the loss better, to avoid the batch-size issue pointed here.

```python
TensorDict({"loss": {"actor": tensor, ...}, "metadata": {...}}, [])
```
which could let us set a different batch-size at diferent levels.
It isn't going to be easy to move to that format though! So for now I think the best would be to keep the output without batch-size until we figure out how to account for it long-term.
```

File: torchrl/objectives/a2c.py (hunk 3 above)
Comment (albertbou92) [reply torchrl/objectives/a2c.py:399]: Because dist.rsample generates `self.samples_mc_entropy` samples along dimension 0. So we average then also in dim 0 to get an estimation of the entropy for each position of the input tensordict.

File: torchrl/objectives/a2c.py (hunk 2 above)
Comment (albertbou92) [reply torchrl/objectives/a2c.py:243]: I am not sure if I follow. Default `mean` ensures BC and seems to me a a reasonable choice. Also Torch losses have `mean` as default, which I think is consistent.

File: torchrl/objectives/a2c.py [hunk 19]
```diff
@@ -68,6 +70,10 @@ class A2CLoss(LossModule):
             Functionalizing permits features like meta-RL, but makes it
             impossible to use distributed models (DDP, FSDP, ...) and comes
             with a little cost. Defaults to ``True``.
+        reduction (str, optional): Specifies the reduction to apply to the output:
+            ``"none"`` | ``"mean"`` | ``"sum"``. ``"none"``: no reduction will be applied,
+            ``"mean"``: the sum of the output will be divided by the number of
+            elements in the output, ``"sum"``: the output will be summed. Default: ``"mean"``.
```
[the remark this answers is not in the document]
albertbou92: I used `str`'s because I was trying to be consistent with torch losses, which don't use Enum. I am fine using an Enum if preferred.

torchrl/objectives/ppo.py
albertbou92: What about adding all losses with the correct shape in the first level and add a nested TensorDict with all the other tensors. We can call it `metadata` as suggested in the first comment.

albertbou92:
I incorporated a part of the suggestions and left comments in the points that might need further discussion!

torchrl/objectives/a2c.py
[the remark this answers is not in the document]
vmoens: Let's use strings for now. I like enum but it's also always the one more import that doens't really make things that much easier...

torchrl/objectives/a2c.py
vmoens: It will still be "mean". But not having "mean" is good practice because it allows us to detect whenever the user is passing or not the reduction. 
Might come in handy at some point: for instance, if we want to change the default for one loss someday, we will just have to check whether the reduction is None or not.

torchrl/objectives/a2c.py
vmoens: Oh right!

torchrl/objectives/ppo.py
vmoens: that won't work because the nested tensordict must have the same leading dims as the root (+ other optional dims) as it is the case for tensors

torchrl/objectives/a2c.py
albertbou92: and how do you define a no-reduction option then?

torchrl/objectives/ppo.py
albertbou92: aah true...

torchrl/objectives/a2c.py
vmoens: `"none"` (not `None`)

test/test_cost.py
```diff
@@ -7144,6 +7184,9 @@ def test_reinforce_value_net(
             elif td_est is not None:
                 loss_fn.make_value_estimator(td_est)
             loss_td = loss_fn(td)
+            if reduction == "none":
+                assert loss_td.batch_size == td.batch_size
+                loss_td = loss_td.apply(lambda x: x.float().mean(), batch_size=[])
```
vmoens, line 7189: maybe here instead of converting to float we can do
```python
def func(x):
    if x.dtype != torch.float:
        return
    return x.mean()
loss_td = loss_td.apply(func, batch_size=[])
```
because I suspect we convert to float things that we may not want to convert to float (eg metadata)

vmoens (approved):
LGTM just smth minor in the tests

PR #1893, opened 2024-02-09
[NOMERG] Add @overload to forward in losses

vmoens:
Loss functions accept non-tensordict data thanks to the tensordict.dispatch decorator.

However, we do not provide an overloaded forward, which could be useful to let users know about the typical signature using default settings.

In this PR, I propose that feature for A2C. The list input keys can be found in `A2CLoss.in_keys` property and also extrapolated from `test/test_cost.py:TestA2C:test_a2c_notensordict`.

All losses that are tested with `test_<smth>_notensordict` can benefit from this overload.

### DIFF
--- a/torchrl/objectives/a2c.py
+++ b/torchrl/objectives/a2c.py
@@ -6,10 +6,10 @@ import contextlib
 import warnings
 from copy import deepcopy
 from dataclasses import dataclass
-from typing import Tuple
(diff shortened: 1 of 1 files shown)


gtnv:
hi @vmoens, is this change still being worked on? if not, would you like to close it?

PR #1910, opened 2024-02-15
[BugFix] Use setdefault in _cache_values

### LINKED ISSUE #1907: [BUG] I had to patch these 2 methods in order to run my script
## Describe the bug

(1)
`DoubleToFloat` transform cannot transform to `float` if env to be transformed is already a `float` (e.g. it is a `float` and the input dtype was declared as `double`). 

I mean, all it has to do is call `.to(torch.float)` or something, which can be called regardless of the `dtype` of the input tensor, so why not?

(2)
The `torchrl.objectives.utils._cache_values` function is also bugged I think. If there is no `_cache` attribute in `self.__dict__`, it raises an error, but the function seems to be implemented in a way that it can work anyway. In fact, when I manually do `self.__dict__["_cache"] = {}`, all is good.

This happens with the `torchrl.objectives.ClipPPOLoss`.

## To Reproduce

I do not know how to reproduce, although my project is fully available at: https://gitlab.com/svnv-svsv-jm/chess-rl/-/tree/develop?ref_type=heads

For more context, I am trying to combine `torchrl` with `pytorch-lightning`.

## Reason and Possible fixes

For the cache issue with `torchrl.objectives.ClipPPOLoss`, I patched this:

```python
# in torchrl.objectives.utils
def _cache_values(fun):
    """Caches the tensordict returned by a property."""
    name = fun.__name__

    def new_fun(self, netname=None):
        __dict__ = self.__dict__
       # MY PATCH START
        try:
            _cache = __dict__["_cache"]
        except KeyError as ex:
            __dict__["_cache"] = {}
            _cache = __dict__["_cache"]
        # MY PATCH END
        attr_name = name
        if netname is not None:
            attr_name += "_" + netname
        if attr_name in _cache:
            out = _cache[attr_name]
            return out
        if netname is not None:
            out = fun(self, netname)
        else:
            out = fun(self)
        # TODO: decide what to do with locked tds in functional calls
        # if is_tensor_collection(out):
        #     out.lock_()
        _cache[attr_name] = out
        return out

    return new_fun
```

For the `torch.envs.DoubleToFloat`, I patched `DTypeCastTransform`:

```python
# class DTypeCastTransform
def transform_observation_spec(self, observation_spec):
        full_observation_spec = observation_spec
        for observation_key, observation_spec in list(
            full_observation_spec.items(True, True)
        ):
            # find out_key that match the in_key
                if observation_key == in_key:
                    if observation_spec.dtype != self.dtype_in:
                        # PATCH START
                        self.dtype_in = observation_spec.dtype
                        # raise TypeError(
                        #     f"observation_spec.dtype is not {self.dtype_in}"
                        # )
                        # PATCH END
                    full_observation_spec[out_key] = self._transform_spec(
                        observation_spec
                    )
        return full_observation_spec
```

After these patches, all runs smoothly again.

Sorry for not providing a minimal example, but I really wouldn't know how to create one from my code. These are the tests where I encountered the errors: https://gitlab.com/svnv-svsv-jm/chess-rl/-/tree/develop/tests/shark/models/ppo?ref_type=heads

vmoens:
Closes #1907

### DIFF
--- a/torchrl/objectives/utils.py
+++ b/torchrl/objectives/utils.py
@@ -459,7 +459,7 @@ def _cache_values(fun):
 
     def new_fun(self, netname=None):
         __dict__ = self.__dict__
-        _cache = __dict__["_cache"]
+        _cache = __dict__.setdefault("_cache", {})
         attr_name = name
         if netname is not None:
             attr_name += "_" + netname


svnv-svsv-jm:
Thanks!

PR #1911, opened 2024-02-15
[BugFix] Non exclusive terminated and truncated

### LINKED ISSUE #1899: [Feature Request] Why have mutually exclusive terminated and truncated?
truncated and terminated seem to be mutually exclusive in the design of TorchRL. As in this line
https://github.com/pytorch/rl/blob/1bd5ec640a4a37cbb95c402df91b199c9a8b1736/torchrl/envs/transforms/transforms.py#L5183
Is that necessary? IMO this should not be enforced and the truncated signal can still safely be set while the terminated signal is true, which will allow a more faithful status of the trajectory (it was supposed to be truncated but it was incidentally terminated).
This should be safe as components that read terminated (e.g., values estimators) don't need truncated and should ignore it (truncated sets done and then becomes useless in that case.)

This would allow for example to set the truncation key to terminated when one wants to consider finite-horizon environments.
`StepCounter(truncated_key="terminated")`.
 An alternative would be to have something like a `FiniteHorizonTransform` that sets termination based on a step limit.

Happy to hear your thoughts.

vmoens:
Closes #1899

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -5179,8 +5179,11 @@ class StepCounter(Transform):
                 if self.update_done:
                     done = next_tensordict.get(done_key, None)
                     terminated = next_tensordict.get(terminated_key, None)
-                    if terminated is not None:
-                        truncated = truncated & ~terminated
+
+                    # we can have terminated and truncated
+                    # if terminated is not None:
+                    #     truncated = truncated & ~terminated
+
                     done = truncated | done  # we assume no done after reset
                     next_tensordict.set(done_key, done)
                 next_tensordict.set(truncated_key, truncated)
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -5163,11 +5163,10 @@ class StepCounter(Transform):
     def _step(
         self, tensordict: TensorDictBase, next_tensordict: TensorDictBase
     ) -> TensorDictBase:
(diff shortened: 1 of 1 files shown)


torchrl/envs/transforms/transforms.py
```diff
@@ -5178,9 +5177,12 @@ def _step(
...
                 if self.update_done:
                     done = next_tensordict.get(done_key, None)
-                    terminated = next_tensordict.get(terminated_key, None)
-                    if terminated is not None:
-                        truncated = truncated & ~terminated
+
+                    # we can have terminated and truncated
+                    # terminated = next_tensordict.get(terminated_key, None)
+                    # if terminated is not None:
+                    #     truncated = truncated & ~terminated
```
skandermoalla, line 5184: I don't really get the comment left here.

skandermoalla:
Line 5081+ can be removed I think
```python
@property
    def terminated_keys(self):
        terminated_keys = self.__dict__.get("_terminated_keys", None)
        if terminated_keys is None:
            # make the default terminated keys
            terminated_keys = []
            for reset_key in self.parent._filtered_reset_keys:
                if isinstance(reset_key, str):
                    key = "terminated"
                else:
                    key = (*reset_key[:-1], "terminated")
                terminated_keys.append(key)
        self.__dict__["_terminated_keys"] = terminated_keys
return terminated_keys
```

skandermoalla:
A test would be nice? To check if when truncated_key="terminated" everything still works fine.

PR #1918, opened 2024-02-16
[BugFix,Refactor] Dreamer refactor

BY571:
## Description

Updates the dreamer example to be aligned with other example scripts and prove paper performance.
Adds the option to run dreamer with non-image obs environments and fixes small issues in the objective and other elements.

  

## Motivation and Context

## Types of changes

## Checklist


BY571:
image

Current performance for cheetah run dm_control task. Not yet paper performance.

torchrl/envs/common.py
```diff
@@ -2421,6 +2421,10 @@ def _rollout_stop_early(
                     tensordict = tensordict.to(policy_device, non_blocking=True)
                 else:
                     tensordict.clear_device_()
+
```
BY571, line 2424: This is not ideal and just a current solution, Id be interested to hear what you think @vmoens. We need to detach those values similar to [this](https://github.com/danijar/dreamer/blob/master/dreamer.py#L276) maybe via a detach transform?

vmoens: do we need to backprop anything in this loop?

vmoens: I added a transform to the mb env,
 can you confirm that we always want to detach these values in `mbenv.step`?

torchrl/objectives/dreamer.py
```diff
@@ -260,24 +265,27 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
...
+        # TODO: do we need exploration here?
+        with timeit("actor_loss/time-rollout"), hold_out_net(self.model_based_env), set_exploration_type(
+            ExplorationType.MEAN
         ):
-            tensordict = self.model_based_env.reset(tensordict.clone(recurse=False))
+            # action_td = self.actor_model(td)
+
+            # TODO: we are not using the actual batch beliefs as starting ones - should be solved! took of the primer for the mb_env
+            tensordict = self.model_based_env.reset(tensordict.copy())
+            # TODO: do we detach state gradients when passing again for new actions: action = self.actor(state.detach())
```
vmoens, line 280: @BY571 I need context for these TODOs

torchrl/objectives/dreamer.py
```diff
@@ -342,6 +344,7 @@ def make_value_estimator(self, value_type: ValueEstimators = None, **hyperparams
             self._value_estimator = TDLambdaEstimator(
                 **hp,
                 value_network=value_net,
+                vectorized=False,  # TODO: vectorized version seems not to be similar to the non vectoried
```
vmoens, line 347: @BY571 I'm curious about this, it's supposed to be extensively tested in the CI

torchrl/objectives/dreamer.py
```diff
@@ -148,6 +152,41 @@ def forward(self, tensordict: TensorDict) -> torch.Tensor:
...
+        #     # compute model loss
+        #     kl_loss = self.kl_loss(
+        #         tensordict.get(("next", self.tensor_keys.prior_mean)),
+        #         tensordict.get(("next", self.tensor_keys.prior_std)),
+        #         tensordict.get(("next", self.tensor_keys.posterior_mean)),
+        #         tensordict.get(("next", self.tensor_keys.posterior_std)),
+        #     )
+        #
+        #     dist: IndependentNormal = self.decoder.get_dist(tensordict)
+        #     reco_loss = -dist.log_prob(
```
BY571, line 173: Going back to distance loss caused the improvements?
I dont mind keeping the distance, just changed it as thats how its done in the original [implementation](https://github.com/danijar/dreamer/blob/master/dreamer.py#L165) but if we keep distance loss we should test both versions or make it selectable?

torchrl/objectives/dreamer.py
BY571: I remember that I tested it like vec_value == value and there were some False mentioned. However, checking the values directly they were the same. This TODO can be taken off

torchrl/objectives/dreamer.py
vmoens: log-prob of a normal is a distance loss if you check the equation. The only difference is (1) we don't add n times log(2*pi) to it and (2) the summing is a bit different here. But in the end they boil down to the same result

torchrl/objectives/dreamer.py
```diff
@@ -160,6 +199,12 @@ def forward(self, tensordict: TensorDict) -> torch.Tensor:
             tensordict.detach(),
         )
 
+    @staticmethod
+    def normal_log_probability(x, mean, std):
+        return (
+            -0.5 * ((x.to(mean.dtype) - mean) / std).pow(2) - std.log()
+        )  # - 0.5 * math.log(2 * math.pi)
+
```
vmoens, line 207: here's the equation of the normal log prob if you're interested. It's a distance loss as you can see

torchrl/objectives/dreamer.py
```diff
@@ -259,25 +304,22 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
...
-            tensordict = tensordict.reshape(-1)
-
-        with hold_out_net(self.model_based_env), set_exploration_type(
-            ExplorationType.RANDOM
-        ):
-            tensordict = self.model_based_env.reset(tensordict.clone(recurse=False))
+        tensordict = tensordict.select("state", self.tensor_keys.belief).detach()
+        tensordict = tensordict.reshape(-1)
+
+        # TODO: do we need exploration here?
```
vmoens, line 310:
```suggestion
```

torchrl/objectives/dreamer.py
```diff
@@ -404,6 +447,22 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
...
+        # dist = self.value_model.get_dist(tensordict_select)
+        # if self.discount_loss:
+        #     discount = self.gamma * torch.ones_like(
+        #         lambda_target, device=lambda_target.device
+        #     )
+        #     discount[..., 0, :] = 1
+        #     discount = discount.cumprod(dim=-2)
+        #     value_loss = -(discount * dist.log_prob(lambda_target).unsqueeze(-1)).mean()
+        # else:
+        #     value_loss = -dist.log_prob(lambda_target).mean()
```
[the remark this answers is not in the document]
vmoens: same comment as above, both approaches are roughly equivalent, this is just more efficient and scales better

torchrl/objectives/dreamer.py
```diff
@@ -148,6 +152,41 @@ def forward(self, tensordict: TensorDict) -> torch.Tensor:
...
+        #     # reco_loss = -self.normal_log_probability(x, loc, scale).mean()
+        #
+        #     dist: IndependentNormal = self.reward_model.get_dist(tensordict)
+        #     reward_loss = -dist.log_prob(
+        #         tensordict.get(("next", self.tensor_keys.true_reward))
+        #     ).mean()
+        #     # x = tensordict.get(("next", self.tensor_keys.true_reward))
+        #     # loc = dist.base_dist.loc
+        #     # scale = dist.base_dist.scale
+        #     # reward_loss = -self.normal_log_probability(x, loc, scale).mean()
```
vmoens, line 188:
```suggestion
```

torchrl/objectives/dreamer.py
```diff
@@ -404,6 +446,22 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
...
+        # dist = self.value_model.get_dist(tensordict_select)
+        # if self.discount_loss:
+        #     discount = self.gamma * torch.ones_like(
+        #         lambda_target, device=lambda_target.device
+        #     )
+        #     discount[..., 0, :] = 1
+        #     discount = discount.cumprod(dim=-2)
+        #     value_loss = -(discount * dist.log_prob(lambda_target).unsqueeze(-1)).mean()
+        # else:
+        #     value_loss = -dist.log_prob(lambda_target).mean()
```
vmoens, line 464:
```suggestion
```

PR #1921, opened 2024-02-16
[Refactor] Faster and more generic multi-agent nets

vmoens:
cc @matteobettini  @kfu02

TODO:
- [x] Account for non-initialized params
- [x] Have TD.from_modules work with lazy params (the issue here being that the list of parameters will change between before and after the first call is made).

### DIFF
--- a/torchrl/modules/models/multiagent.py
+++ b/torchrl/modules/models/multiagent.py
@@ -2,20 +2,113 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+from __future__ import annotations
 
+import abc
 from typing import Optional, Sequence, Tuple, Type, Union
 
 import numpy as np
 
 import torch
-from torch import nn
 
+from tensordict import TensorDict
+from torch import nn
 from torchrl.data.utils import DEVICE_TYPING
 
 from torchrl.modules.models import ConvNet, MLP
 
 
-class MultiAgentMLP(nn.Module):
+class MultiAgentNetBase(nn.Module):
+    """A base class for multi-agent networks."""
+    _empty_net: nn.Module
+
+    def __init__(
+        self,
+        *,
+        n_agents: int,
+        centralised: bool,
+        share_params: bool,
+        agent_dim: int,
+        **kwargs,
+    ):
+        super().__init__()
+
+        self.n_agents = n_agents
+        self.share_params = share_params
+        self.centralised = centralised
+        self.agent_dim = agent_dim
+
+        agent_networks = [
+            self._build_single_net(**kwargs)
+            for _ in range(self.n_agents if not self.share_params else 1)
+        ]
+        kwargs["device"] = "meta"
+        self.__dict__["_empty_net"] = self._build_single_net(**kwargs)
+        if self.share_params:
+            self.params = TensorDict.from_module(agent_networks[0], as_module=True)
+        else:
+            self.params = TensorDict.from_modules(*agent_networks, as_module=True)
+
+    @abc.abstractmethod
+    def _build_single_net(self, *, device, **kwargs):
+        ...
+
+    @abc.abstractmethod
+    def _pre_forward_check(self, inputs):
+        ...
+
+    @staticmethod
+    def vmap_func_module(module, *args, **kwargs):
+        def exec_module(params, *input):
+            with params.to_module(module):
+                return module(*input)
+
+        return torch.vmap(exec_module, *args, **kwargs)
+
+    def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
+        if len(inputs) > 1:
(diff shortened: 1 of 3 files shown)


torchrl/modules/models/multiagent.py
```diff
@@ -150,87 +243,52 @@ def __init__(
...
-                f" but got {output.shape}"
-            )
-
-        return output
+            inputs = inputs.flatten(-2, -1)
+        return inputs
+
+    def _build_single_net(self, *, device, **kwargs):
+        n_agent_inputs = self.n_agent_inputs
+        if self.centralised and n_agent_inputs is not None:
```
matteobettini, line 278: Super cool that we are allowing `n_agent_inputs` to be `None` and thus have lazy modules.

We need to document this in the docstring and in the type hint

torchrl/modules/models/multiagent.py
```diff
@@ -2,20 +2,113 @@
...
+        self.share_params = share_params
+        self.centralised = centralised
+        self.agent_dim = agent_dim
+
+        agent_networks = [
+            self._build_single_net(**kwargs)
+            for _ in range(self.n_agents if not self.share_params else 1)
+        ]
+        kwargs["device"] = "meta"
+        self.__dict__["_empty_net"] = self._build_single_net(**kwargs)
```
matteobettini, line 46: Just curious the resason for these 2 lines. I don't understand them

matteobettini (approved):
This is amazing. Thanks so much for this! Can't wait to see the perf gains.

Before merging, we should run a comparision on main and on this PR on one of the multiagent example scripts (e.g., mappo_ippo.py) to test that the reward is the same and gather data for the performance differences.

vmoens:
It is faster, we ran multiple benchmarks on this.
It is faster in low regimes if you consider the backward pass (which is much slower when you build multiple graphs) and when you populate your optimizers with many more params (since you'll be calling `step` and `zero` grad on many more tensors with ops executed in python loops).

torchrl/modules/models/multiagent.py
vmoens: To execute the functional module we need an "empty" copy (with meta device - ie no device - ie no data). We also want that copy not to be registered within the list of modules to avoid having those empty parameters pop up when we cann `module.parameters()`.

kfu02 (approved):
This is cool! Thank you so much! 

To follow up on your comment [here](https://github.com/pytorch/rl/pull/1497#issuecomment-1949335593), what do you anticipate adding an RNN will require in addition to this? Is the issue that the hidden state must be initialized first? I don't quite understand what "non initialized lazy params" means or how that creates an issue with RNNs.

vmoens:
@kfu02 @matteobettini for context: the problem with unitilialized params is that if you have modules with lazy, non initialized params it is usually assumed that you can pass them to an optimizer and the optmizer will know that it must wait until they are initialized to do smth with them.

Currently this isn't supported by `from_modules` which will only create dense params. We could create lazy params but I'm not super duper sure I see how that will work with vmap...

matteobettini:
> It is faster, we ran multiple benchmarks on this.

That is cool! I would still run the full multiagent training script to check those 2 things



> @kfu02 @matteobettini for context: the problem with unitilialized params is that if you have modules with lazy, non initialized params it is usually assumed that you can pass them to an optimizer and the optmizer will know that it must wait until they are initialized to do smth with them.


Since lazy modules was not a feature of these classes in the first place, why do we just not leave it that way?

i personally would prefer this to a complex solution to support it

vmoens:
It was a feature of CNN so I just made it uniform

torchrl/modules/models/multiagent.py
```diff
@@ -2,20 +2,146 @@
...
+
+    def _make_params(self, agent_networks):
+        if self.share_params:
+            self.params = TensorDict.from_module(agent_networks[0], as_module=True)
+        else:
+            self.params = TensorDict.from_modules(*agent_networks, as_module=True)
+
+    @abc.abstractmethod
+    def _build_single_net(self, *, device, **kwargs):
```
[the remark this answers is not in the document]
vmoens: yep that is the idea!

kfu02:
> @kfu02 wanna give a shot at MA-RNNs with this once it's merged? Or should I draft a PR?

Yes! I will put up a draft by the end of the week!

matteobettini:
I ran some benchmarks on the mappo_ippo example for the MLP. With MAPPO and non-sharinmg params for 3 agents.

It seems to work! Do not see any regression in this case

W B Chart 21_02_2024, 11_18_37
W B Chart 21_02_2024, 11_18_50

matteobettini:
Further to #1957 i ran some tests with MASAC and non-shared parameters on the same task for 4 agents and the results look good!
W B Chart 26_02_2024, 09_16_20
W B Chart 26_02_2024, 09_16_32

Almost half training time!

PR #1922, opened 2024-02-16
[BugFix] Fix KLPENPPOLoss KL computation

### LINKED ISSUE #1920: [BUG] kl divergence calculation in KLPENPPOLoss is always zero
## Describe the bug
KL divergence calculation in KLPENPPOLoss is always zero, causing the contribution to the loss to be 0.

It seems that the way the KLPENPPOLoss calculates the previous and current distributions end up using the same values/parameters (?), causing the KL divergence to always be zero.

## To Reproduce

o Steps to reproduce the behavior.
Replace ClipPPOLoss with KLENPPOLoss in the ppo mujoco example code and remove the clip_epsilon argument:
```python
    loss_module = KLPENPPOLoss(
        actor_network=actor,
        critic_network=critic,
        #clip_epsilon=cfg.loss.clip_epsilon,
        loss_critic_type=cfg.loss.loss_critic_type,
        entropy_coef=cfg.loss.entropy_coef,
        critic_coef=cfg.loss.critic_coef,
        normalize_advantage=True,
    )
```
Run the example. The code path used is 
```python
            kl = torch.distributions.kl.kl_divergence(previous_dist, current_dist)
```
(line 969).

Note with debugger that previous_dist and current_dist are always the same. This means that on line 973, kl is always 0.

## Expected behavior
This should be log prob calculated with old and new parameters. 

## Screenshots

N/A

## System info
Python 3.11.7, torchrl-nightly.

## Additional context

N/A

## Reason and Possible fixes
The idea is to use cached logits, I think, but the call to
```python
log_weight, dist = self._log_weight(tensordict)
```
runs the fresh params on the tensordict and overwrites the old logits, which causes the calls to return the same values. A hack that works for me is to save the incoming tensordict (in KLPENPPOLoss.forward) before it is cloned, and use it explicitly in the call to build_dist_from_params, but I'm not sure what the ideal solution is.

## Checklist

- [x ] I have checked that there is no similar issue in the repo (**required**)
- [x ] I have read the [documentation](https://github.com/pytorch/rl/tree/main/docs/) (**required**)
- [ x] I have provided a minimal working example to reproduce the bug (**required**)

vmoens:
Fixes #1920

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -6485,7 +6485,9 @@ class TestPPO(LossModuleTestBase):
         loss_val_td = loss(td)
 
         for i, out_key in enumerate(loss.out_keys):
-            torch.testing.assert_close(loss_val_td.get(out_key), loss_val[i], msg=out_key)
+            torch.testing.assert_close(
+                loss_val_td.get(out_key), loss_val[i], msg=out_key
+            )
 
         # test select
         torch.manual_seed(self.seed)
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -6481,7 +6481,7 @@ class TestPPO(LossModuleTestBase):
         loss_val = loss(**kwargs)
         torch.manual_seed(self.seed)
         if beta is not None:
-            setattr(loss, "beta", beta.clone())
+            loss.beta = beta.clone()
         loss_val_td = loss(td)
 
         for i, out_key in enumerate(loss.out_keys):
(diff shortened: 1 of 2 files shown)


skandermoalla:
@vmoens @albertbou92 you may very likely need to redesign the beta update. Updating every minibatch (every loss run) and on a minibatch kl, which may be noisy, is very unstable. I would recommend taking it out of the forward, making a method called something like `update_adaptive parameters` that also takes an estimate kl (that is typically the mean kl over all the minibatches of the last epoch) and call it outside at the algorithm level at every batch.
This is what is done in the original implementation of KLPPO
https://github.com/joschu/modular_rl/blob/6970cde3da265cf2a98537250fea5e0c0d9a7639/modular_rl/ppo.py#L208C19-

albertbou92:
In my experience it is also useful sometimes to be able to use the ClipPPOLoss with the KL term (for example against a prior). I think we should enable that. Maybe the KL term should be a loss module by itself so it can be combined with any other loss?

PR #1927, opened 2024-02-19
[BugFix] No grad on collector reset

matteobettini:
If an environment has gradients coming out of the reset function, these will be present in the shuttle


vmoens (approved):
LGTM thanks!

Issue #1929, opened 2024-02-19
[BUG] Rollout vs collector in training loop


matteobettini:
Hello,

I was recently trying out substituting a `rollout` to my Sync collector in the sampling stage of multiagent training.

This is due to the fact that I need to compute gradients of the reward function (which the collector does not allow)

However, even with no gradients, replacing a rollout batch to the collector batch completely impedes training in my case.

Both the rollout and the collection are happening on a single vectorized environement with everything on the same device.

I have made an example script of my problem in #1928 where basically just by substituting `tensordict_data` from a rollout to the `tensordict_data` from the collector compeltely impedes training performance in a very simple task.

I would like more insights on why this is the case or if there is something wrong I am doing that i missed.

Thanks

P.S. all shapes of data from collector and rollout are the same

vmoens:
Not sure I understand the issue, can you try to pinpoint a little more what's happening (maybe with a gist?)

Collectors do not do the same thing as rollout, mainly from a reset perspective. There are other things that might cause the issue... We do test that rollout and iterating over the collector spit out the same data. 
Perhaps you can try to do the same, iterating over the collector and using rollout and compare the results key-by-key?

matteobettini:
> Not sure I understand the issue, can you try to pinpoint a little more what's happening (maybe with a gist?)

I'll try. What is happening is that everything works but the training rewards collapse in the rollout case, so I suspect that it has something to do with an interaction with the policy.

> Collectors do not do the same thing as rollout, mainly from a reset perspective

I tested it locally in an environemnt that just spits done at the end of max_steps and it occurs also in this case (no resets)

> We do test that rollout and iterating over the collector spit out the same data.
Perhaps you can try to do the same, iterating over the collector and using rollout and compare the results key-by-key?

I'll try this! i suspect that it has something to do with the policy interaction

vmoens:
To mimic the dataloader you should pass a tensordict to use instead of reset to rollout, have you tried that?

matteobettini:
You mean change

```python
          tensordict_data = env.rollout(
              max_steps=cfg.env.max_steps,
              policy=policy,
              break_when_any_done=False,
              auto_reset=True,
          )
```
to

```python
          td_reset = env.reset()
          tensordict_data = env.rollout(
              max_steps=cfg.env.max_steps,
              policy=policy,
              break_when_any_done=False,
              auto_reset=False,
              tensordict=td_reset,
          )
```
this does not change the outcome

[...]

PR #1936, opened 2024-02-20
[BugFix] Remove reset on last step of a rollout

matteobettini:
Discussion in #1929

### DIFF
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -2540,11 +2540,6 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
                 is_shared=False)
         """
         tensordict = self.step(tensordict)
-        tensordict_ = self.maybe_reset(tensordict)
-
-        return tensordict, tensordict_
-
-    def maybe_reset(self, tensordict: TensorDictBase) -> TensorDictBase:
         # done and truncated are in done_keys
         # We read if any key is done.
         tensordict_ = step_mdp(
@@ -2556,6 +2551,10 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
             action_keys=self.action_keys,
             done_keys=self.done_keys,
         )
+        tensordict_ = self.maybe_reset(tensordict_)
+        return tensordict, tensordict_
+
+    def maybe_reset(self, tensordict_: TensorDictBase) -> TensorDictBase:
         any_done = _terminated_or_truncated(
             tensordict_,
             full_done_spec=self.output_spec["full_done_spec"],
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -2273,7 +2273,9 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
                 called on the sub-envs that are done. Default is True.
             return_contiguous (bool): if False, a LazyStackedTensorDict will be returned. Default is True.
             tensordict (TensorDict, optional): if auto_reset is False, an initial
-                tensordict must be provided.
+                tensordict must be provided. Rollout will check if this tensordict has done flags and reset the
+                environment in those dimensions (if needed). This normally should not occur if ``tensordict`` is the
+                output of a reset, but can occur if ``tensordict`` is the last step of a previous rollout.
 
         Returns:
             TensorDict object containing the resulting trajectory.
@@ -2368,6 +2370,26 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
             is_shared=False)
             >>> print(rollout.names)
             [None, 'time']
+
+        Rollouts can be used in a loop to emulate data collection.
+        To do so, you need to pass as input the last tensordict coming from the previous rollout after calling
+        :func:`~torchrl.envs.utils.step_mdp` on it.
+
+        Examples:
+            >>> from torchrl.envs import GymEnv, step_mdp
+            >>> env = GymEnv("CartPole-v1")
+            >>> epochs = 10
+            >>> input_td = env.reset()
+            >>> for i in range(epochs):
+            ...     rollout_td = env.rollout(
+            ...         max_steps=100,
+            ...         break_when_any_done=False,
+            ...         auto_reset=False,
+            ...         tensordict=input_td,
+            ...     )
+            ...     input_td = step_mdp(
+            ...         rollout_td[..., -1],
+            ...     )
 
         """
         if auto_cast_to_device:
@@ -2566,15 +2588,24 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         tensordict_ = self.maybe_reset(tensordict_)
         return tensordict, tensordict_
 
-    def maybe_reset(self, tensordict_: TensorDictBase) -> TensorDictBase:
+    def maybe_reset(self, tensordict: TensorDictBase) -> TensorDictBase:
+        """Checks the done keys of the input tensordict and, if needed, resets the environment where it is done.
+
+        Args:
+            tensordict (TensorDictBase): a tensordict coming from the output of :func:`~torchrl.envs.utils.step_mdp`.
+        Returns:
+            A tensordict that is identical to the input where the environment was
+            not reset and contains the new reset data where the environment was reset.
+
+        """
         any_done = _terminated_or_truncated(
-            tensordict_,
+            tensordict,
             full_done_spec=self.output_spec["full_done_spec"],
             key="_reset",
         )
         if any_done:
-            tensordict_ = self.reset(tensordict_)
-        return tensordict_
+            tensordict = self.reset(tensordict)
+        return tensordict
 
     def empty_cache(self):
         """Erases all the cached values.
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -2593,6 +2593,7 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
 
         Args:
             tensordict (TensorDictBase): a tensordict coming from the output of :func:`~torchrl.envs.utils.step_mdp`.
+
         Returns:
             A tensordict that is identical to the input where the environment was
             not reset and contains the new reset data where the environment was reset.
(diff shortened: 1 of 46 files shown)


vmoens (commented):
Thanks for this! Ideally we'd like a non regression test.
Also now that there is a reset missing, if the user takes the tensordict passed as output and feeds it back to rollout, rollout will need to call reset on it. I don't think this was part of rollout contract before: either a tensordict is passed or reset is called but not both (as this PR will dictate)

matteobettini:
> Ideally we'd like a non regression test.

I'll work on the test.

> Also now that there is a reset missing, if the user takes the tensordict passed as output and feeds it back to rollout, rollout will need to call reset on it. I don't think this was part of rollout contract before: either a tensordict is passed or reset is called but not both (as this PR will dictate)

So this is a design choice, let me explain my view.

## Before the PR

Before the PR, a user could not take the last tensordict from a rollout and safely pass it back in.

This is because if the env was reset on the last step, the data from that reset would not be part of the rollout results.

Thus, if the user passed the output of a rollout back into a new rollout, It would be providing old data from a previous trajectory to a freshly reset env which is in a different state.

# After this PR

In the current state of this PR, the user still cannot just take the last tensordict from a rollout and safely pass it back in.

Doing so would have the same effect as prior to this PR, with the difference that now at least the env is not in a new reset state.

In my opinion this should remain like this (and like it was).

If the users chooses `auto_reset=False` it will be the duty of the user to pass non-done data as the input to a rollout (just as before).


The difference that this PR allows is that now users that want to run rollouts in a row are able to by adding the reset logic outside the rollout function (as detailed in the snippet in https://github.com/pytorch/rl/issues/1929#issuecomment-1954286298). Before they could not do this.

vmoens:
If we patch things let's patch them correctly no?
Why not having something like
```python
        if auto_reset:
            if tensordict is not None:
                raise RuntimeError(
                    "tensordict cannot be provided when auto_reset is True"
                )
            tensordict = self.reset()
        elif tensordict is None:
            raise RuntimeError("tensordict must be provided when auto_reset is False")
		else:
			aggregate_reset = _aggregate_end_of_traj(tensordict)
            if aggregate_reset.any():
				self.reset(tensordict)
```
I don't understand the reluctance against this, does this break anything?

matteobettini:
> I don't understand the reluctance against this, does this break anything?

Oh no this works too! I'm not against it.

It will just be a new feature.

Aka rollout will check that the passed tensordict is done and it will reset the env if so.

We can do it, a few considerations:
- this feature should be activated if also  `auto_reset`  is set. This would change the meaning of auto_reset as it could be set at the same time as an input tensordict, generating the following cases:
      - no tensordict passed -> auto_reset will either call reset or not (like before)
      - tensordict passed and auto_reset=True -> rollout will check if the passed td is done and if so reset (this is the new param combination that was not allowed before and it will trigger the new feature)
      - tensordict passed and auto_reset=False -> rollout will not check the input td for reset (like before)
    
     keeping the last case is good for bc compatibility and there might be cases where you do not want the rollout to call a reset

vmoens:
> It will just be a new feature.

I think we're at the edge between new feature and bug fix. If you can't pass the tensordict you got from the last step instead of calling reset safely, this would be a bugfix.

matteobettini:
> If you can't pass the tensordict you got from the last step instead of calling reset safely, this would be a bugfix.

Yeah if this was supposed to be a feature before, I guess this is a bug fix

vmoens:
> keeping the last case is good for bc compatibility and there might be cases where you do not want the rollout to call a reset

If the env does not allow a step after done and you pass a tensordict that is done, it will result in an error. IMO we should capture that and make sure it does not happen for ease of use. I can't think of anyone using rollout reasonably right now who would be annoyed by this change and consider it bc-breaking, but I could be overlooking things.

To me `auto_reset` just means reset when starting. I don't think we should change the meaning. The consideration I was putting forward is orthogonal to `auto_reset` IMO.

matteobettini:
> If the env does not allow a step after done and you pass a tensordict that is done, it will result in an error. IMO we should capture that and make sure it does not happen for ease of use. I can't think of anyone using rollout reasonably right now who would be annoyed by this change and consider it bc-breaking, but I could be overlooking things.
> 
> To me `auto_reset` just means reset when starting. I don't think we should change the meaning. The consideration I was putting forward is orthogonal to `auto_reset` IMO.

This consideration makes an assumption about the env.
I can also make an example about users that might want to take steps in done environments. (I personally have cases for this)

Forcing a reset on these users would narrow the flexibility of rollout with respect to prior versions.
I think if we introduce this now auto resetting when a td is given, it is important to make it deactivatable.

If we want to have another param to deactivate it other than auto_reset that is fine too, I just though that auto_reset fits nicely as it was not allowed before and it is true by default.

vmoens:
Not sure I'm following here

Previously we had a broken version of rollout where the reset data of the last step was potentially lost. Anyone wanting to do anything with this was doomed to fail. I don't think that in this scenario working with env that do not reset when done was an option. So this scenario can be excluded entirely and considered as a separate issue.

If we want to land this, to me having it work with tensordict that are in a done state is a pre-requirement. The change I'm proposing isn't bc-breaking in any way since it build upon a bug-fix: anyone who was doing anything with rollout before can only have fewer bugs now. 

The runtime won't be affected since the check I'm suggesting is already performed by `step_and_maybe_reset`, which we now replace by `step`.

matteobettini:
There is something that we still did not discuss.

If you really want to chain calls of rollout, then the function should call `sefl.maybe_reset` on the input dict (like in the current version of the PR).
Which includes `step_mdp`.

Or are we assuming that users call `step_mdp` beteween rollout calls?

vmoens:
It's safe to assume that users call `step_mdp` because the contract is that the tensordict you pass is similar to the one you'd get out of a call to `reset`

matteobettini:
Ok if you wanna have a look now it should do what we want

vmoens (commented):
LGTM, just missing a docstring and a test
(maybe also writing a note in `rollout` doctring about all of this? Like "how to use rollout as a data collector" or similar?)

matteobettini:
Further question:

if `break_when_any_done` is true, we still do the reset and ignore it?

matteobettini:
> LGTM, just missing a docstring and a test (maybe also writing a note in `rollout` doctring about all of this? Like "how to use rollout as a data collector" or similar?)

We should be gucci!

Also, I still haven't come around the problem in #1929, so currently rollout is still not working for me as a data collector.
Have to see if it is a bug of my brain or a a very subtle bug of the library.

Nevertheless, this can be merged.

torchrl/envs/common.py
```diff
@@ -2369,6 +2371,33 @@ def rollout(
             >>> print(rollout.names)
             [None, 'time']
 
+        Rollouts can be used in a loop to emulate data collection.
+        To do so, you need to pass as input the last tensordict coming from the previous rollout after calling
+        :meth:`step_mdp` on it.
```
vmoens, line 2376:
```suggestion
        :func:`~torchrl.envs.utils.step_mdp` on it.
```

torchrl/envs/common.py
```diff
@@ -2369,6 +2371,33 @@ def rollout(
...
+            ...     )
+            ...     input_td = step_mdp(
+            ...         rollout_td[..., -1],
+            ...         keep_other=True,
+            ...         exclude_action=False,
+            ...         exclude_reward=True,
+            ...         reward_keys=env.reward_keys,
+            ...         action_keys=env.action_keys,
+            ...         done_keys=env.done_keys,
+            ...     )
```
vmoens, line 2399:
```suggestion
            ...     input_td = step_mdp(
            ...         rollout_td[..., -1],
            ...     )
```

torchrl/envs/common.py
```diff
@@ -2369,6 +2371,33 @@ def rollout(
...
+            >>> epochs = 10
+            >>> input_td = env.reset()
+            >>> for i in range(epochs):
+            ...     rollout_td = env.rollout(
+            ...         max_steps=100,
+            ...         policy=None,
+            ...         break_when_any_done=False,
+            ...         auto_reset=False,
+            ...         tensordict=input_td,
+            ...     )
```
vmoens, line 2390:
```suggestion
            ...     rollout_td = env.rollout(
            ...         max_steps=100,
            ...         break_when_any_done=False,
            ...         auto_reset=False,
            ...         tensordict=input_td,
            ...     )
```

torchrl/envs/common.py
```diff
@@ -2557,14 +2592,30 @@ def step_and_maybe_reset(
...
             done_keys=self.done_keys,
         )
+        tensordict_ = self.maybe_reset(tensordict_)
+        return tensordict, tensordict_
+
+    def maybe_reset(self, tensordict: TensorDictBase) -> TensorDictBase:
+        """Checks the done keys of the input tensordict and, if needed, resets the environment where it is done.
+
+        Args:
+            tensordict (TensorDictBase): a tensordict coming from the output of :meth:`step_mdp`
```
vmoens, line 2602:
```suggestion
            tensordict (TensorDictBase): a tensordict coming from the output of :func:`~torchrl.envs.utils.step_mdp`.
```

torchrl/envs/common.py
```diff
@@ -2557,14 +2592,30 @@ def step_and_maybe_reset(
...
+        Returns:
+            TensorDictBase: a tensordict that is identical to the input one where the environment was
```
vmoens, line 2604:
```suggestion
            A tensordict that is identical to the input where the environment was
```

torchrl/envs/common.py
```diff
@@ -2557,14 +2592,30 @@ def step_and_maybe_reset(
...
+            not reset and contains the new reset data where the environment was reset
```
vmoens, line 2605:
```suggestion
            not reset and contains the new reset data where the environment was reset.
```

torchrl/envs/common.py
```diff
@@ -2557,14 +2592,30 @@ def step_and_maybe_reset(
...
+
+        This method is part of :meth:`~.step_and_maybe_reset` and should be called on the output of a :meth:`~.step`
+        on which :meth:`step_mdp` has been called.
+
```
vmoens, line 2609:
```suggestion
```

PR #1943, opened 2024-02-21
[BugFix] Fix multiple context syntax in multiagent examples

matteobettini:
Fix the syntax of multiple contexts in the multiagent examples


vmoens (approved):
LGTM

PR #1944, opened 2024-02-21
[BugFix,Feature] Allow non-tensor data in envs

vmoens:
Depends on https://github.com/pytorch/tensordict/pull/685


vmoens:
@maximilianigl @Michael-C-Strobel I'm going back to this
For a single env it is easy to patch.
For batched envs there are two ways of doing this:
- We used to rely entirely on buffers in batched envs to share data across processes. We can't do that generically with non-tensor data atm. So I was thinking to share the non-tensor data separately through the pipes and the rest on the buffers. This is awfully complex but it could work
- The simpler idea could be to prevent users from using buffers altogether. Envs with non-tensor data will be slow to be executed in parallel envs, but that's going to be the price to pay to get them running. 

I guess my question is this: how much do you guys care about parallel envs runtime when it comes to envs with non tensor data?

vmoens:
I fixed both options (with and without buffers) and all are incorporated in the tests.
There may still be some incompatibility with collectors, but I think this should belong to a separate PR.

maximilianigl:
For my project I've hacked together an "encoder"/"decoder" step that encodes non-tensor data into ints and stores the values in a dictionary. That allowed me to use buffers and just "decode" the data in places I need it. 
Regarding your question: If I still had the problem, I'd care about runtime. But not sure how much complexity that justifies.

Michael-C-Strobel:
I think parallel envs will end up being pretty important to us. In our case our environment initialization instantiates an object from another package but neither the environment nor agent modifies it (the agent just needs access to some of its methods/attributed). 

For us, a constant NonTensorSpec would probably work and I suspect wouldn’t cause any issues for the collector or replay buffer.

vmoens:
Ok so I think that now runtime should be ok-ish (at least not drastically slower)
Next steps after this PR are collectors and buffers.

PR #1946, opened 2024-02-22
[Feature] Add time_dim arg in value modules

vmoens:


wbinventor:
It looks like this didn’t make it in the new 0.3.1 release. Will this be included in the following release?

wbinventor:
That makes sense! Just wondered about the release plan since the new parameter is rendered in the online documentation even though it’s not yet available.

vmoens:
You're pointing at a longstanding issue here which is that the doc of the main branch is automatically synced and we don't have yet a separate doc for the release and for main.
We should fix this! It's one of these "one maintainer cannot do it all" sort of things :/

PR #1952, opened 2024-02-22
[Feature] Checking for dtype to all tensorspec is_in funcs

### LINKED ISSUE #793: [Feature Request] TensorSpec is_in methods should check the dtype of val
## Motivation

As mentioned in #783, we should check that `is_in` always checks the dtype.

## Solution

Add self.dtype == val.dtype condition in the `is_in` methods. Should be very simple to fix, but we need to make sure that all tests pass.

## Checklist

namanxkumar:
## Description

I've added a check to all `is_in` functions in TensorSpec for the datatype, as specified in #793.

## Motivation and Context

Close #793 initially proposed in #783

## Types of changes

## Checklist

### DIFF
--- a/torchrl/data/tensor_specs.py
+++ b/torchrl/data/tensor_specs.py
@@ -1456,7 +1456,7 @@ class OneHotDiscreteTensorSpec(TensorSpec):
         shape = torch.broadcast_shapes(shape, val.shape)
         mask_expand = self.mask.expand(shape)
         gathered = mask_expand & val
-        return gathered.any(-1).all()
+        return gathered.any(-1).all() and val.dtype == self.dtype
 
     def __eq__(self, other):
         if not hasattr(other, "mask"):
@@ -1765,7 +1765,7 @@ class BoundedTensorSpec(TensorSpec):
         try:
             return (val >= self.space.low.to(val.device)).all() and (
                 val <= self.space.high.to(val.device)
-            ).all()
+            ).all() and (val.dtype == self.dtype)
         except RuntimeError as err:
             if "The size of tensor a" in str(err):
                 warnings.warn(f"Got a shape mismatch: {str(err)}")
@@ -1894,7 +1894,7 @@ class UnboundedContinuousTensorSpec(TensorSpec):
         return torch.empty(shape, device=self.device, dtype=self.dtype).random_()
 
     def is_in(self, val: torch.Tensor) -> bool:
-        return True
+        return val.dtype == self.dtype
 
     def expand(self, *shape):
         if len(shape) == 1 and isinstance(shape[0], (tuple, list, torch.Size)):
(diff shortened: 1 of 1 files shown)


torchrl/data/tensor_specs.py
```diff
@@ -1456,7 +1456,7 @@ def is_in(self, val: torch.Tensor) -> bool:
         shape = torch.broadcast_shapes(shape, val.shape)
         mask_expand = self.mask.expand(shape)
         gathered = mask_expand & val
-        return gathered.any(-1).all()
+        return gathered.any(-1).all() and val.dtype == self.dtype
```
vmoens, line 1459: this isn't related to this PR but I think the proper check should be

```suggestion
        return (gathered.sum(-1) == 1).all() and val.dtype == self.dtype
```

torchrl/data/tensor_specs.py
```diff
@@ -2333,7 +2333,7 @@ def is_in(self, val: torch.Tensor) -> bool:
         vals = self._split(val)
         if vals is None:
             return False
-        return all(spec.is_in(val) for val, spec in zip(vals, self._split_self()))
+        return all(spec.is_in(val) for val, spec in zip(vals, self._split_self())) and (val.dtype == self.dtype)
```
vmoens, line 2336: I guess the sub-calls to `is_in` will take care of the check in this case

torchrl/data/tensor_specs.py
```diff
@@ -3585,7 +3585,7 @@ def is_in(self, val: Union[dict, TensorDictBase]) -> bool:
             val_item = val.get(key)
             if not item.is_in(val_item):
                 return False
-        return True
+        return val.dtype == self.dtype
```
vmoens, line 3588: composite specs don't have a meaningful dtype, we can revert this

torchrl/data/tensor_specs.py
```diff
@@ -4107,7 +4107,7 @@ def is_in(self, val) -> bool:
         for spec, subval in zip(self._specs, val.unbind(self.dim)):
             if not spec.is_in(subval):
                 return False
-        return True
+        return val.dtype == self.dtype
```
vmoens, line 4110: ditto here
we did the checks so this can be skipped

vmoens (commented):
Great stuff!
While we're at it I think we should also test that the shape matches when it makes sense:
```python
value.shape[-len(self.shape):] == self.shape and value.dtype == self.dtype
```

torchrl/data/tensor_specs.py
namanxkumar: Should I add this to the current PR?

namanxkumar:
I've fixed most of these issues, and added checks for shape where applicable.

torchrl/data/tensor_specs.py
vmoens: Yes I think those two are related and can be part of the same PR

namanxkumar: Okay, done

vmoens (commented):
I linted the code. The tests should be fixed, some are broken now

Thanks a lot for this!

In the future make sure to use another branch than main on your fork, that really helps!

gtnv:
hey @namanxkumar, is this still moving forward? if not, would you like to close it?

PR #1968, opened 2024-02-27
[BugFix] Refactor reductions

vmoens:
cc @albertbou92 

The tests are failing because when collecting `loss.action_loss(...)` the tensor is not reduced.

I think if the sub-loss is a public method the reduction should happen there.

Also we should only reduce the losses, not the metadata.

The PR solves both these issues


albertbou92:
Great thanks! The current errors don't seem related to the reduction parameter.

albertbou92:
I have added the on-policy tests here: https://github.com/pytorch/rl/pull/1972

PR #1970, opened 2024-02-27
[Feature] `reset_parameters` for multiagent nets

### LINKED ISSUE #1967: [Feature] Reset parameters of multiagent networks
Hello!

So I usually used a function like this to reset parameters of the multiagent networks

```python
def reset_child_params(module):
    for layer in module.children():
        if hasattr(layer, "reset_parameters"):
            layer.reset_parameters()
        reset_child_params(layer)

```

After #1921 this seems to have no effect.

Is there a suggested way to reset the parameters?

Thanks!

matteobettini:
Fixes #1967

Todo:
- [x] Tests

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -1049,10 +1049,10 @@ class TestMultiAgent:
                     assert not torch.allclose(out[..., i, :], out[..., j, :])
 
     def test_multiagent_cnn_lazy(self):
-        n_channels = 5
+        n_agents = 5
         n_channels = 3
         cnn = MultiAgentConvNet(
-            n_agents=5,
+            n_agents=n_agents,
             centralised=False,
             share_params=False,
             in_features=None,


torchrl/modules/models/multiagent.py
```diff
@@ -125,6 +125,21 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+        def vmap_reset_module(module, *args, **kwargs):
+            def reset_module(params):
+                with params.to_module(module):
+                    module.reset_parameters()
+                    return params
+
+            return torch.vmap(reset_module, *args, **kwargs)
+
+        if not self.share_params:
+            vmap_reset_module(self._empty_net, randomness="different")(self.params)
```
matteobettini, line 138: I would like to discuss the vmap randomness.

```python
    @property
    def _vmap_randomness(self):
        if self.initialized:
            return "error"
        return "same"
```

Why would this be a class property and why are we having those values?

For me this should be

```python
    @property
    def _vmap_randomness(self):
        return "different"
```

matteobettini: In this case it needs to be different to have a different reset values for each agent.

But also in the forward pass I feel like it should be "different"

vmoens: IMO it should be
```python
    @property
    def _vmap_randomness(self):
        if self.initialized:
            return self.vmap_randomness
        return "different"
```

and users are in charge of telling the module what randomness they want.

matteobettini: Could you explain this a bit? Why do we have a switch on initialization?

vmoens: For init you need "different" because you must have different weights for each net.
But in other settings you can't tell, and the best is to let the user choose. 
They may as well want the same random number for each element of the batch

torchrl/modules/models/utils.py
```diff
@@ -123,3 +124,26 @@ def create_on_device(
     else:
         return module_class(*args, **kwargs).to(device)
         # .to() is always available for nn.Module, and does nothing if the Module contains no parameters or buffers
+
+
+def reset_parameters_recursive(module, warn_if_no_op: bool = True) -> bool:
+    """Recursively resets the parameters of a torch.nn.Module in-place.
```
vmoens, line 130:
```suggestion
    """Recursively resets the parameters of a :class:`~torch.nn.Module` in-place.
```

torchrl/modules/models/utils.py
```diff
@@ -123,3 +124,26 @@ def create_on_device(
     else:
         return module_class(*args, **kwargs).to(device)
         # .to() is always available for nn.Module, and does nothing if the Module contains no parameters or buffers
+
+
+def reset_parameters_recursive(module, warn_if_no_op: bool = True) -> bool:
```
vmoens, line 129:
```suggestion
def _reset_parameters_recursive(module, warn_if_no_op: bool = True) -> bool:
```

vmoens: Make private or place in doc

torchrl/modules/models/utils.py
```diff
@@ -123,3 +124,26 @@ def create_on_device(
...
+    """Recursively resets the parameters of a torch.nn.Module in-place.
+
+    Args:
+        module (torch.nn.Module): the module to reset
```
vmoens, line 133:
```suggestion
        module (torch.nn.Module): the module to reset.
```

torchrl/modules/models/utils.py
```diff
@@ -123,3 +124,26 @@ def create_on_device(
...
+        warn_if_no_op (bool): whether to raise a warning in case this is a no-op
```
vmoens, line 134:
```suggestion
        warn_if_no_op (bool, optional): whether to raise a warning in case this is a no-op.
            Defaults to ``True``.
```

torchrl/modules/models/utils.py
```diff
@@ -123,3 +124,26 @@ def create_on_device(
...
+
+    Returns: whether any parameters where reset
```
vmoens, line 136:
```suggestion
    Returns: whether any parameter has been reset.
```

torchrl/modules/models/multiagent.py
```diff
@@ -125,6 +126,21 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
 
         return output
 
+    def reset_parameters(self):
+        def vmap_reset_module(module, *args, **kwargs):
+            def reset_module(params):
+                with params.to_module(module):
+                    reset_parameters_recursive(module)
```
vmoens, line 133:
```suggestion
                    _reset_parameters_recursive(module)
```

torchrl/modules/models/multiagent.py
```diff
@@ -125,6 +126,21 @@ def forward(self, *inputs: Tuple[torch.Tensor]) -> torch.Tensor:
...
+                    return params
+
+            return torch.vmap(reset_module, *args, **kwargs)
+
+        if not self.share_params:
+            vmap_reset_module(self._empty_net, randomness="different")(self.params)
+        else:
+            with self.params.to_module(self._empty_net):
+                reset_parameters_recursive(self._empty_net)
```
vmoens, line 142:
```suggestion
                _reset_parameters_recursive(self._empty_net)
```

test/test_modules.py
```diff
@@ -1051,6 +1075,28 @@ def test_multiagent_cnn_lazy(self):
...
+            num_cells=[5, 5],
+            n_agents=n_agents,
+            centralised=centralised,
+            share_params=share_params,
+        )
+        params_before = deepcopy(list(actor_net.parameters()))
+        actor_net.reset_parameters()
+        params_after = list(actor_net.parameters())
+        for p1, p2 in zip(params_before, params_after):
+            assert (p1 != p2).all()
```
vmoens, line 1098: Better to check that they are not close than different

test/test_modules.py
```diff
@@ -947,6 +948,29 @@ def test_multiagent_mlp_lazy(self):
...
```
vmoens, line 972: ditto

test/test_modules.py
```diff
@@ -947,6 +948,29 @@ def test_multiagent_mlp_lazy(self):
...
+    ):
+        actor_net = MultiAgentMLP(
+            n_agent_inputs=4,
+            n_agent_outputs=6,
+            num_cells=[5, 5],
+            n_agents=n_agents,
+            centralised=centralised,
+            share_params=share_params,
+        )
+        params_before = deepcopy(list(actor_net.parameters()))
```
vmoens, line 968:
```suggestion
        params_before = actor_net.params.clone()
```

test/test_modules.py
```diff
@@ -947,6 +948,29 @@ def test_multiagent_mlp_lazy(self):
...
+        actor_net.reset_parameters()
+        params_after = list(actor_net.parameters())
```
vmoens, line 970:
```suggestion
        params_after = actor_net.params()
```

test/test_modules.py
```diff
@@ -947,6 +948,29 @@ def test_multiagent_mlp_lazy(self):
...
+        for p1, p2 in zip(params_before, params_after):
```
vmoens, line 971:
```suggestion
        for p1, p2 in zip(params_before.values(True, True), params_after.values(True, True)):
```

test/test_modules.py
```diff
@@ -1051,6 +1075,28 @@ def test_multiagent_cnn_lazy(self):
...
+        share_params,
+    ):
+        actor_net = MultiAgentConvNet(
+            in_features=4,
+            num_cells=[5, 5],
+            n_agents=n_agents,
+            centralised=centralised,
+            share_params=share_params,
+        )
+        params_before = deepcopy(list(actor_net.parameters()))
```
vmoens, line 1094:
```suggestion
        params_before = actor_net.params.clone()
```

test/test_modules.py
```diff
@@ -1051,6 +1075,28 @@ def test_multiagent_cnn_lazy(self):
...
+        actor_net.reset_parameters()
+        params_after = list(actor_net.parameters())
```
vmoens, line 1096:
```suggestion
        params_after = actor_net.params
```

test/test_modules.py
```diff
@@ -1051,6 +1075,28 @@ def test_multiagent_cnn_lazy(self):
...
+        for p1, p2 in zip(params_before, params_after):
```
vmoens, line 1097:
```suggestion
        for p1, p2 in zip(params_before.values(True, True), params_after.values(True, True)):
```

vmoens (commented):
Thanks!

torchrl/modules/models/multiagent.py
matteobettini: if I try to change this to 
```python
    @property
    def _vmap_randomness(self):
        if self.initialized:
            return self.vmap_randomness
        return "different"
```
the lazy layers will crash

vmoens: > the lazy layers will crash

this is a statement that is hard to reproduce, can you share more?

For instance this works fine on my end:
```python
from torchrl.modules import MLP
import torch
from functorch import dim

d0 = dim.dims(1)
modules = [torch.nn.Linear(2, 3) for _ in range(3)]

td = TensorDict.from_modules(*modules, as_module=True)

def reset(td):
    with td.to_module(modules[0]):
        modules[0].reset_parameters()
    return td

td = torch.vmap(reset, randomness="same")(td)
print(td["weight"])
td = torch.vmap(reset, randomness="different")(td)
print(td["weight"])
```
the first produces a stack of identical tensors, the second different

torchrl/modules/models/multiagent.py
```diff
@@ -54,9 +57,11 @@ def __init__(
         self.__dict__["_empty_net"] = self._build_single_net(**kwargs)
 
     @property
-    def _vmap_randomness(self):
+    def vmap_randomness(self):
         if self.initialized:
-            return "error"
+            return self._vmap_randomness
+        # Matteo: There seems to be a problem with lazy layers when using "different" here
```
vmoens, line 63: taking care of this.
We should avoid leaving comments with names like this, it's very easy to miss during review and can stay there forever!

vmoens: This is the culprit:
https://github.com/pytorch/tensordict/blob/f09f20dd586ee1749d244e5ae42ee5cdc992bd77/tensordict/_td.py#L771-L776

When we call vmap we actually don't batch the lazy tensor (we can't do it) so it must be `"same"`. `"same"` isn't a problem though because the tensor isn't batched so it's not going to copy the same config for every element. TLDR we can stick to same

matteobettini: Ok then can you write  a comment with this in the code? and then we are good to go!

PR #1984, opened 2024-03-01
[Feature] Offline objectives reduction parameter

albertbou92:
## Description

Same as https://github.com/pytorch/rl/pull/1890, but for Offline losses.


vmoens (commented):
LGTM thanks!

albertbou92:
should be fixed now!

PR #1989, opened 2024-03-04
[Feature] Preproc for datasets

vmoens:
This PR allows to preprocess datasets once and for all.

One can use transforms of any kind (function or regular transforms).

TODO:
- [x] Complete tests for other datasets
- [x] Document the feature
- [x] Consider allowing non-pickable functions
- [x] download=True must erase the previous dataset

cc @nicklashansen: I have a couple of questions for you:
- Do you think it makes sense to replace the dataset, or should we leave the option of not replacing it by default? For instance we could have a `dest` argument in `preprocess` that would default to `dataset_path`. 
- If the dataset is replaced, you can get the original by doing `MyDataSet(..., download="force")`. Does that seem intuitive?
- Does this feature satisfy your needs? Have a look at [`TensorDict.map`](https://pytorch.org/tensordict/reference/generated/tensordict.TensorDictBase.html#tensordict.TensorDictBase.map) to see the available options, which is essentially what we're using here.

Example usage with OpenX:

```python
from torchrl.envs import Compose, RenameTransform, Resize

dataset = OpenXExperienceReplay(
    "cmu_stretch",
    download="force",
    streaming=False,
    batch_size=64,
    shuffle=True,
    num_slices=8,
    slice_len=None,
)

t = Compose(
    Resize(
        64,
        64,
        in_keys=[("observation", "image"), ("next", "observation", "image")],
    ),
    RenameTransform(
        in_keys=[
            ("observation", "image"),
            ("next", "observation", "image"),
            ("observation", "state"),
            ("next", "observation", "state"),
        ],
        out_keys=["pixels", ("next", "pixels"), "state", ("next", "state")],
    ),
)

def fn(data: TensorDict):
    data.unlock_()
    data = data.select(
        "action",
        "done",
        "episode",
        ("next", "done"),
        ("next", "observation"),
        ("next", "reward"),
        ("next", "terminated"),
        ("next", "truncated"),
        "observation",
        "terminated",
        "truncated",
    )
    data = t(data)
    data = data.select(*data.keys(True, True))
    return data

dataset.preprocess(
    CloudpickleWrapper(fn),
    num_workers=max(1, os.cpu_count() - 2),
    num_chunks=500,
    mp_start_method="fork",
)
sample = dataset.sample(32)
assert "observation" not in sample.keys()
assert "pixels" in sample.keys()
assert ("next", "pixels") in sample.keys(True)
assert "state" in sample.keys()
assert ("next", "state") in sample.keys(True)
assert sample["pixels"].shape == torch.Size([32, 3, 64, 64])
dataset = OpenXExperienceReplay(
    "cmu_stretch",
    download=True,
    streaming=False,
    batch_size=64,
    shuffle=True,
    num_slices=8,
    slice_len=None,
)
sample = dataset.sample(32)
assert "observation" not in sample.keys()
assert "pixels" in sample.keys()
assert ("next", "pixels") in sample.keys(True)
assert "state" in sample.keys()
assert ("next", "state") in sample.keys(True)

```

After the transform, the data looks like this
```
 OpenXExperienceReplay(
    storage=TensorStorage(
        data=TensorDict(
            fields={
                action: MemoryMappedTensor(shape=torch.Size([25016, 8]), device=cpu, dtype=torch.float64, is_shared=True),
                done: MemoryMappedTensor(shape=torch.Size([25016, 1]), device=cpu, dtype=torch.bool, is_shared=True),
                episode: MemoryMappedTensor(shape=torch.Size([25016]), device=cpu, dtype=torch.int32, is_shared=True),
                image: MemoryMappedTensor(shape=torch.Size([25016, 3, 64, 64]), device=cpu, dtype=torch.uint8, is_shared=True),
                next: TensorDict(
                    fields={
                        done: MemoryMappedTensor(shape=torch.Size([25016, 1]), device=cpu, dtype=torch.bool, is_shared=True),
                        image: MemoryMappedTensor(shape=torch.Size([25016, 3, 64, 64]), device=cpu, dtype=torch.uint8, is_shared=True),
                        reward: MemoryMappedTensor(shape=torch.Size([25016, 1]), device=cpu, dtype=torch.float32, is_shared=True),
                        state: MemoryMappedTensor(shape=torch.Size([25016, 4]), device=cpu, dtype=torch.float64, is_shared=True),
                        terminated: MemoryMappedTensor(shape=torch.Size([25016, 1]), device=cpu, dtype=torch.bool, is_shared=True),
                        truncated: MemoryMappedTensor(shape=torch.Size([25016, 1]), device=cpu, dtype=torch.bool, is_shared=True)},
                    batch_size=torch.Size([25016]),
                    device=cpu,
                    is_shared=False),
                state: MemoryMappedTensor(shape=torch.Size([25016, 4]), device=cpu, dtype=torch.float64, is_shared=True),
                terminated: MemoryMappedTensor(shape=torch.Size([25016, 1]), device=cpu, dtype=torch.bool, is_shared=True),
                truncated: MemoryMappedTensor(shape=torch.Size([25016, 1]), device=cpu, dtype=torch.bool, is_shared=True)},
            batch_size=torch.Size([25016]),
            device=cpu,
            is_shared=False), 
        shape=torch.Size([25016]), 
        len=25016, 
        max_size=25016), 
    sampler=SliceSampler(num_slices=8, slice_len=None, end_key=('next', 'done'), traj_key=episode, truncated_key=('next', 'truncated'), strict_length=True), 
    writer=ImmutableDatasetWriter(), 
    batch_size=64, 
    collate_fn=<function _collate_id at 0x123f3f4c0>)
```

### DIFF
--- a/.github/unittest/linux/scripts/run_all.sh
+++ b/.github/unittest/linux/scripts/run_all.sh
@@ -124,9 +124,9 @@ git submodule sync && git submodule update --init --recursive
 printf "Installing PyTorch with %s\n" "${CU_VERSION}"
 if [[ "$TORCH_VERSION" == "nightly" ]]; then
   if [ "${CU_VERSION:-}" == cpu ] ; then
-      pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu -U
+      pip3 install 'mpmath<1.4' --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu -U
   else
-      pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/$CU_VERSION
+      pip3 install 'mpmath<1.4' --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/$CU_VERSION
   fi
 elif [[ "$TORCH_VERSION" == "stable" ]]; then
     if [ "${CU_VERSION:-}" == cpu ] ; then
--- a/.github/unittest/linux/scripts/run_all.sh
+++ b/.github/unittest/linux/scripts/run_all.sh
@@ -194,11 +194,11 @@ pytest test/smoke_test_deps.py -v --durations 200 -k 'test_gym or test_dm_contro
 if [ "${CU_VERSION:-}" != cpu ] ; then
   python .github/unittest/helpers/coverage_run_parallel.py -m pytest test \
     --instafail --durations 200 -vv --capture no --ignore test/test_rlhf.py \
-    --timeout=120
+    --timeout=120 --mp_fork_if_no_cuda
 else
   python .github/unittest/helpers/coverage_run_parallel.py -m pytest test \
     --instafail --durations 200 -vv --capture no --ignore test/test_rlhf.py --ignore test/test_distributed.py \
-    --timeout=120
+    --timeout=120 --mp_fork_if_no_cuda
 fi
 
 coverage combine
--- a/.github/unittest/linux/scripts/run_all.sh
+++ b/.github/unittest/linux/scripts/run_all.sh
@@ -124,9 +124,9 @@ git submodule sync && git submodule update --init --recursive
 printf "Installing PyTorch with %s\n" "${CU_VERSION}"
 if [[ "$TORCH_VERSION" == "nightly" ]]; then
   if [ "${CU_VERSION:-}" == cpu ] ; then
-      pip3 install 'mpmath<1.4' --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu -U
+      pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu -U
   else
-      pip3 install 'mpmath<1.4' --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/$CU_VERSION
+      pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/$CU_VERSION
   fi
 elif [[ "$TORCH_VERSION" == "stable" ]]; then
     if [ "${CU_VERSION:-}" == cpu ] ; then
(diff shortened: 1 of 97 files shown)


dennismalmgren:
When I last did this, I didn't get the transforms right the first time (or second, or third...), and it took some iterations - sometimes jointly with algorithm development. Hence, I'd prefer an option not to replace the downloaded dataset, to speed up that iterative process.

vmoens:
Makes sense!
We could:
- Provide a `dest` keyword arg for the dest. Either it defaults to the source and replaces it -- or we only do it if a `"same"` dest is provided. 
- Provide the option to do only X frames (e.g. max_frames). By default, the `preproces` will do the entire dataset, but if you can ask for less. This would be mainly useful for debugging

vmoens:
Upon reflection it seems it makes more sense to encourage users to copy the data somewhere else.

What should happen to:
- Sampler? I guess we can generally keep the same sampler as the one set in the dataset
- Writer? I would keep the immutable writer by default
- transforms? Here I would vote for discarding any transform that has been set in the dataset. Transforms are presumably used to do the same job as this preproc

vmoens:
@nicklashansen @dennismalmgren I need some input here.

The issue I'm facing is that there is no way to tell what the storage will look like after the transform, and hence the sampler / writer / transforms will in many cases be obsolete.

If we recycle them by default there's a big chunk of cases where the transformed dataset will not be usable anymore.

Therefore, since we need at least the space for the dataset and the transformed dataset to use this functionality (you can't delete the first before you get the second), I think 
(1) we should never replace the dataset with its transformed version: it's a bit chaotic, hard to implement because the dataset can come with many other attributes and on top of that we can't be sure that the signature of the dataset will work with the new one
(2) the transform should not return a replay buffer but just a storage and the user should build the replay buffer from scratch:

```python
dataset = OpenXExperienceReplay(...)
new_storage = dataset.preprocess(func, ...)
# we give users the option to delete the dataset
dataset.clear()
new_dataset = ReplayBuffer(storage=new_storage, sampler=whatever, ...)
```

A bit more code is needed but at least people know what they're doing. Note that `new_dataset = ReplayBuffer(storage=new_storage)` will always work with a default writer and sampler.

Wdyt?

nicklashansen:
@vmoens thank you for initiating this! i agree with @dennismalmgren comment, and in general i see this as being used mostly as a separate step as opposed to being part of the training process / data loading. and i agree that it the most convenient solution would be to store the transformed dataset somewhere else by default, but perhaps giving users the option of deleting the original data after preprocessing. im not overly concerned about the (potentially) 2x disk space requirements.

ideally one would be able to download all of the torchrl datasets to disk, preprocess and save them in the desired format (would probably have to be specified manually but so be it), and then whichever training process and data loader that users build downstream (most likely in a different process) will use that new, common (in case of multi-dataset training) format without the added overhead of preprocessing on the fly. that said, i can imagine scenarios in which a user might still want to apply *some* transforms when sampling, e.g., data augmentation or random slicing, but that seems like something that would be supported as is?

PR #1994, opened 2024-03-06
[Refactor,Feature] Refactor collector shapes and stack_result in sync collector

vmoens:
## Description

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/docs/source/reference/collectors.rst
+++ b/docs/source/reference/collectors.rst
@@ -59,6 +59,49 @@ Besides those compute parameters, users may choose to configure the following pa
 - exploration_type: the exploration strategy to be used with the policy.
 - reset_when_done: whether environments should be reset when reaching a done state.
 
+Collectors and batch size
+-------------------------
+
+Because each collector has its own way of organizing the environments that are
+run within, the data will come with different batch-size depending on how
+the specificities of the collector. The following table summarizes what is to
+be expected when collecting data:
+
+
++--------------------+---------------------+-----------------------------+------------------------------+
+|                    | SyncDataCollector   |MultiSyncDataCollector (n=B) |MultiaSyncDataCollector (n=B) |
++====================+=====================+=============+===============+==============================+
+|   `stack_result`   |          NA         |    `True`   |    `False`    |             NA               |
++--------------------+---------------------+-------------+---------------+------------------------------+
+|     Single env     |         [T]         |   `[B, T]`  |  `[B*(T//B)`  |              [T]             |
++--------------------+---------------------+-------------+---------------+------------------------------+
+| Batched env (n=P)  |       [P, T]        | `[B, P, T]` |  `[B * P, T]` |            [P, T]            |
++--------------------+---------------------+-------------+---------------+------------------------------+
+
+In each of these cases, the last dimension (``T`` for ``time``) is adapted such
+that the batch size equals the ``frames_per_batch`` argument passed to the
+collector.
+
+:class:`~torchrl.collectors.collectors.MultiSyncDataCollector` should only be
+used with ``stack_result=False`` when preemption is required, as the
+data will be stacked along the batch dimension or the time dimension, which
+can introduce some confusion when swapping regular and batched environments.
+In all other cases, using :func:`~torch.stack` is a better thing to do as it
+will keep each dimension separate and provide better interchangeability
+between configurations and collector classes.
+
+Another thing to note is that whereas :class:`~torchrl.collectors.collectors.MultiSyncDataCollector`
+has a dimension corresponding to the number of sub-collectors being run (``B``),
+:class:`~torchrl.collectors.collectors.MultiaSyncDataCollector` doesn't. This
+is easily understood when considering that :class:`~torchrl.collectors.collectors.MultiaSyncDataCollector`
+delivers batches of data on a first come first serve basis, whereas
+:class:`~torchrl.collectors.collectors.MultiSyncDataCollector` gathers data from
+each sub-collector before delivering it.
+
+Collectors and replay buffers interoperability
+----------------------------------------------
+
+
 
 Single node data collectors
 ---------------------------
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -52,7 +52,12 @@ from torchrl.collectors.collectors import (
     MultiSyncDataCollector,
 )
 from torchrl.collectors.utils import split_trajectories
-from torchrl.data import CompositeSpec, UnboundedContinuousTensorSpec
+from torchrl.data import (
+    CompositeSpec,
+    LazyTensorStorage,
+    ReplayBuffer,
+    UnboundedContinuousTensorSpec,
+)
 from torchrl.envs import (
     EnvBase,
     EnvCreator,
@@ -1626,8 +1631,8 @@ class TestPreemptiveThreshold:
 
         interruptor.stop_collection()
         for batch in collector:
-            assert batch["collector"]["traj_ids"][0] != -1
-            assert batch["collector"]["traj_ids"][1] == -1
+            assert batch["collector", "traj_ids"][0] != -1
+            assert batch["collector", "traj_ids"][1] == -1
         collector.shutdown()
         del collector
 
@@ -1660,7 +1665,7 @@ class TestPreemptiveThreshold:
         )
 
         for batch in collector:
-            trajectory_ids = batch["collector"]["traj_ids"]
+            trajectory_ids = batch["collector", "traj_ids"]
             trajectory_ids_mask = trajectory_ids != -1  # valid frames mask
             assert trajectory_ids[trajectory_ids_mask].numel() < frames_per_batch
         collector.shutdown()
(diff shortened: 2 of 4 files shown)


vmoens:
@albertbou92 I also took this opportunity to refactor preemption. Now you can stack (with padding) or cat (with masking) the values coming from the collectors.
I noticed we preempt based on the proportion of collectors that have reached the end, not on the proportion of frames collected. IMO that would be more sensible.
We could use a `mp.Value()` shared between processes that tracks how many frames have been collected, and as soon as this goes above the threshold we break. Would you like to give it a shot once this PR lands? No worry if you don't have the time!
I suspect it can bring some decent speed-up over what we're doing.

In a nutshell:
```python
# in SyncDataCollector.rollout
def rollout(self):
    ...
    if self.interruptor is not None:
        self._collected_frames += frame_count

# in MultiSyncDataCollector.iterator
def iterator(self):
    if self.preemptive_threshold and ...:
        if self._collected_frames > self.frames_per_batch * self.preemptive_threshold:
            break # or similar
```

albertbou92:
> @albertbou92 I also took this opportunity to refactor preemption. Now you can stack (with padding) or cat (with masking) the values coming from the collectors. I noticed we preempt based on the proportion of collectors that have reached the end, not on the proportion of frames collected. IMO that would be more sensible. We could use a `mp.Value()` shared between processes that tracks how many frames have been collected, and as soon as this goes above the threshold we break. Would you like to give it a shot once this PR lands? No worry if you don't have the time! I suspect it can bring some decent speed-up over what we're doing.
> 
> In a nutshell:
> 
> ```python
> # in SyncDataCollector.rollout
> def rollout(self):
>     ...
>     if self.interruptor is not None:
>         self._collected_frames += frame_count
> 
> # in MultiSyncDataCollector.iterator
> def iterator(self):
>     if self.preemptive_threshold and ...:
>         if self._collected_frames > self.frames_per_batch * self.preemptive_threshold:
>             break # or similar
> ```

Yes the preemption was based on the idea that all workers collect a fixed number of frames and can not communicate during collection, but if we can have a shared global value that tracks how many frames that would be much better. Yes, I can give it shot!

vmoens:
@matteobettini This PR addresses an issue that you pointed a long time ago RE the stacking of results in MultiSync
Do you think it solves it in an adequate way?

The idea is to build the collector with `cat_results="stack"` if you want to stack or `cat_results=-1` if you want to cat along the time dimension. `cat_result=None` results in `cat_results=0` + warning since this is the previous behaviour but we want to discourage it. 
We recommend "stack" > -1 > 0

Sorry it took so long to solve this!

matteobettini:
> @matteobettini This PR addresses an issue that you pointed a long time ago RE the stacking of results in MultiSync Do you think it solves it in an adequate way?
> 
> The idea is to build the collector with `cat_results="stack"` if you want to stack or `cat_results=-1` if you want to cat along the time dimension. `cat_result=None` results in `cat_results=0` + warning since this is the previous behaviour but we want to discourage it. We recommend "stack" > -1 > 0
> 
> Sorry it took so long to solve this!

Amamzing!

So `stack_results=True` now is exactly like a single collector with the number of collectors B preappended to the batch size which feels so smooth!

This is what i wanted yes!

docs/source/reference/collectors.rst
```diff
@@ -59,6 +59,49 @@ Besides those compute parameters, users may choose to configure the following pa
...
+the specificities of the collector. The following table summarizes what is to
+be expected when collecting data:
+
+
++--------------------+---------------------+-----------------------------+------------------------------+
+|                    | SyncDataCollector   |MultiSyncDataCollector (n=B) |MultiaSyncDataCollector (n=B) |
++====================+=====================+=============+===============+==============================+
+|   `stack_result`   |          NA         |    `True`   |    `False`    |             NA               |
++--------------------+---------------------+-------------+---------------+------------------------------+
+|     Single env     |         [T]         |   `[B, T]`  |  `[B*(T//B)`  |              [T]             |
```
matteobettini, line 76:
```suggestion
|     Single env     |         [T]         |   `[B, T]`  |  `[B*(T//B)]`  |              [T]             |
```

docs/source/reference/collectors.rst
```diff
@@ -59,6 +59,49 @@ Besides those compute parameters, users may choose to configure the following pa
...
++--------------------+---------------------+-------------+---------------+------------------------------+
+| Batched env (n=P)  |       [P, T]        | `[B, P, T]` |  `[B * P, T]` |            [P, T]            |
```
matteobettini, line 78: `[B * P, T]`, this is given that ` cat_results=0`

my personal preference, to avoid having `cat_results` be both int and str is to have 2 args
- `cat_results` true or false
- `collectors_dim` dimension where to cat or stack

matteobettini (approved):

docs/source/reference/collectors.rst
vmoens: but then what is `collector_dim=0` and `cat_results=False`? That should not be allowed. So we need a complicated doc that lists the available configs
If we allow them we also need to test every single combination...

docs/source/reference/collectors.rst
```diff
@@ -59,6 +59,112 @@ Besides those compute parameters, users may choose to configure the following pa
...
+
+
++--------------------+---------------------+--------------------------------------------+------------------------------+
+|                    | SyncDataCollector   |       MultiSyncDataCollector (n=B)         |MultiaSyncDataCollector (n=B) |
++====================+=====================+=============+==============+===============+==============================+
+|   `cat_results`    |          NA         |  `"stack"`  |      `0`     |      `-1`     |             NA               |
++--------------------+---------------------+-------------+--------------+---------------+------------------------------+
+|     Single env     |         [T]         |   `[B, T]`  |  `[B*(T//B)` |  `[B*(T//B)`  |              [T]             |
++--------------------+---------------------+-------------+--------------+---------------+------------------------------+
+| Batched env (n=P)  |       [P, T]        | `[B, P, T]` |  `[B * P, T]`|  `[B, T * P]` |            [P, T]            |
```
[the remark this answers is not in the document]
vmoens: I don't think so, if you cat along -1 you still have B at the beginning

docs/source/reference/collectors.rst
vmoens: Note that the goal is to get rid of this arg in v0.6 or v0.7 so adding 2 args is twice as bad as adding 1 given that we want to break things on the long run

docs/source/reference/collectors.rst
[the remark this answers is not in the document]
matteobettini: oh i thought B is the dimension that doens't show up when you cat, this is counter intuitive to me

docs/source/reference/collectors.rst
matteobettini: > collector_dim=0 and cat_results=False

isn't this what happens currently when `cat_results="stack"`?

Anyway yes how you prefer

docs/source/reference/collectors.rst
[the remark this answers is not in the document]
matteobettini: in my view of this

- stack -> B will show up (forcefully at 0)
- cat -> B will be cat to the shape index provided

docs/source/reference/collectors.rst
vmoens: Yes sorry, I meant collector_dim=-1 or 1 or anything that isn't 0

you can't stack along any dim (and should not be able to do so), that's what I mean. Stack should be along dim 0

docs/source/reference/collectors.rst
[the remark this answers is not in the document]
vmoens: Ah I think you're right lol 
P is the dim of the env!

docs/source/reference/collectors.rst
matteobettini: i don't super see this as in the example table we are referring to, in the row for stacking in multisync i can see both

- B, P, T (collector_dim=0 or -3, cat_results=False) (currently the only available option)
- P, B, T (collector_dim=1 or -2, cat_results=False)
- P, T, B (collector_dim=2 or -1, cat_results=False)

vmoens: There is no `cat_results=False`, it can only be a string ("stack") or an int
cat_results="stack" => stack B collectors along 0 => [B, P, T]
cat_results=0 => cat B collectors along 0 => [B * P, T]
cat_results=-1 => cat B collectors along -1 => [B, P * T]

vmoens: I think I didn't get what you were trying to say though

matteobettini: I know, that is the point of this thread, I am discussing why for me it makes sense to separate them.

vmoens: > P, B, T (collector_dim=1 or -2, cat_results=False)
> P, T, B (collector_dim=2 or -1, cat_results=False)

I don't think we should support these. If you want that you can transpose your resulting tensordict. Implementing it would require a lot of expensive tests for a feature we will deprecate in 2 releases.

PR #1999, opened 2024-03-07
[BugFix] Fix sampling without replacement with ndim storages

vmoens:


torchrl/data/replay_buffers/samplers.py
```diff
@@ -1307,14 +1309,24 @@ def sample(
         seq_length, num_slices = self._adjusted_batch_size(batch_size)
         indices, _ = SamplerWithoutReplacement.sample(self, storage, num_slices)
         storage_length = storage.shape[0]
+
+        # traj_idx will either be a single tensor or a tuple that can be reorganized
+        # like a non-zero through stacking.
+        def tuple_to_tensor(traj_idx, lengths=lengths):
+            if isinstance(traj_idx, tuple):
+                traj_idx = torch.arange(len(storage), device=lengths.device).view(
```
vmoens, line 1317: I compared this with `np.ravel_multi_index` using

```python
torch.as_tensor(np.ravel_multi_index(tuple(idx.numpy() for idx in unravelled), shape))
```

Rumtimes are roughly equivalent, with a slight advantage for the numpy version

vmoens: This is way slower, about 2.5x the numpy solution
```python
def ravel_multi_index(x, shape):
    out = 0
    shape_modif = np.cumprod(list(reversed((*shape, 1))))
    for i, idx in enumerate(reversed(x)):
        out += idx * shape_modif[i]
    return out

```

vmoens: A more vectorized version still underperforms arange and numpy

```python
def ravel_multi_index(x, shape):
    out = 0
    shape_modif = torch.flipud(
        torch.cumprod(torch.tensor(list(reversed((*shape[1:], 1)))), 0)
    ).unsqueeze(0)
    return (torch.stack(x, -1) * shape_modif).sum(-1)

```

PR #2004, opened 2024-03-09
[BugFix] Set strict=False in tensordict.select() calls for objective classes

albertbou92:
## Description

This PR proposes to set `strict=False` for the `tensordict.select()` calls in the objective classes.

The logic behind this is that in some situations the Actors and Value networks should still be able to run through the losses without some of the in_keys. 

Examples are: 
1. RNN without the `recurrent_state` key can still be called when the `is_init` is True.
2. Transformer networks may need a mask during inference but not during training.
3. More generally, users could make custom TensorDict modules that have optional parameters. 

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/torchrl/objectives/a2c.py
+++ b/torchrl/objectives/a2c.py
@@ -411,7 +411,9 @@ class A2CLoss(LossModule):
             raise RuntimeError(
                 f"tensordict stored {self.tensor_keys.action} require grad."
             )
-        tensordict_clone = tensordict.select(*self.actor_network.in_keys).clone()
+        tensordict_clone = tensordict.select(
+            *self.actor_network.in_keys, strict=False
+        ).clone()
         with self.actor_network_params.to_module(
             self.actor_network
         ) if self.functional else contextlib.nullcontext():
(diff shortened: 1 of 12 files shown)


torchrl/objectives/a2c.py
```diff
@@ -411,7 +411,9 @@ def _log_probs(
             raise RuntimeError(
                 f"tensordict stored {self.tensor_keys.action} require grad."
             )
-        tensordict_clone = tensordict.select(*self.actor_network.in_keys).clone()
```
vmoens, line 414: is it required to change it in the tests too?

torchrl/objectives/a2c.py
```diff
@@ -425,7 +427,9 @@ def loss_critic(self, tensordict: TensorDictBase) -> torch.Tensor:
             # TODO: if the advantage is gathered by forward, this introduces an
             # overhead that we could easily reduce.
             target_return = tensordict.get(self.tensor_keys.value_target)
-            tensordict_select = tensordict.select(*self.critic_network.in_keys)
```
vmoens, line 428: ditto

vmoens (approved):
LGTM just curious why we needed to adapt the tests too

Issue #1977, opened 2024-02-27
Value clipping for PPO loss


wbinventor:
## Motivation

It would nice to add a new option to compute the clipped value loss as used in OpenAI's Baselines for PPO:

https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/ppo2/model.py#L66-L75

Currently, the `PPOLoss().loss_critic` method calls the `torchrl.objectives.utils.distance_loss` function which supports the `"l1"`, `"l2"` and `"smooth_l1"` loss functions. Perhaps this new clipped value loss function could be implemented as a new loss type within `distance_loss`.

The clipping fraction is also commonly reported as a metric by OpenAI's Baselines and this could be useful to report from `PPOLoss` for the clipped value loss, as well as for `ClipPPOLoss` for the `loss_objective` loss term.

vmoens:
That makes sense!

@albertbou92 Thoughts? Do you want to take care of this one?

albertbou92:
Yes I also thought in the past about adding the PPO value clipping. 
Not sure yet whether the logic should go in PPO or in `distance_loss`. My first thought is in PPO, since I have only seen it used in PPO, but I guess could potentially be used in other methods.
In any case,  I can definitely start a PR to implement it.

vmoens:
@matteobettini we could raise a warning if the value requires grad in that case? 

In a way what's great about a good lib is that it allows you to do loads of stuff but also tells you when you're off the road

albertbou92:
Here is the code generally used:

```
# Compute normal values loss
value_loss = (new_v - return).pow(2)

if self.use_clipped_value_loss:
    value_pred_clipped = old_v + (new_v - old_v).clamp(-self.clip_param, self.clip_param)
    value_loss_clipped = (value_pred_clipped - return).pow(2)
    
    # Chose the most pessimistic value prediction between clipped and non-clipped
    # so the updates are smaller
    value_loss = torch.max(value_loss, value_loss_clipped)

value_loss = 0.5 * value_loss.mean()
```
where, 
`old_v` : value predicted during collection, without grads
`new_v`: values predicted during loss computation, with grads

So thinking in terms of PPO, the `value_loss` and `value_pred_clipped` will be the same in the first batch update.
In the next iterations, we are simply mitigate excessively large updates to the value function.

[...]

PR #2005, opened 2024-03-09
[Feature] Adds value clipping in ClipPPOLoss loss

albertbou92:
## Description

This PR add a value clipping option in ClipPPOLoss loss.

This PR is related to https://github.com/pytorch/rl/issues/1977.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/torchrl/objectives/a2c.py
+++ b/torchrl/objectives/a2c.py
@@ -247,7 +247,7 @@ class A2CLoss(LossModule):
         actor: ProbabilisticTensorDictSequential = None,
         critic: ProbabilisticTensorDictSequential = None,
         reduction: str = None,
-        clip_value: float = None,
+        clip_value: float | None = None,
     ):
         if actor is not None:
             actor_network = actor
--- a/torchrl/objectives/a2c.py
+++ b/torchrl/objectives/a2c.py
@@ -475,6 +475,7 @@ class A2CLoss(LossModule):
                 f"TDLambdaEstimate and TDEstimate all return a 'value_target' entry that "
                 f"can be used for the value loss."
             )
+        clip_fraction = None
         if self.clip_value:
             clip_value = self.clip_value.to(state_value.device)
             state_value_clipped = old_state_value + (
@@ -487,7 +488,13 @@ class A2CLoss(LossModule):
             )
             # Chose the most pessimistic value prediction between clipped and non-clipped
             loss_value = torch.max(loss_value, loss_value_clipped)
-        return self.critic_coef * loss_value
+            clip_fraction = (
+                (state_value / old_state_value)
+                .clamp(1 - clip_value, 1 + clip_value)
+                .abs()
+                .detach()
+            )
+        return self.critic_coef * loss_value, clip_fraction
 
     @property
     @_cache_values
@@ -516,8 +523,10 @@ class A2CLoss(LossModule):
             td_out.set("entropy", entropy.detach().mean())  # for logging
             td_out.set("loss_entropy", -self.entropy_coef * entropy)
         if self.critic_coef:
-            loss_critic = self.loss_critic(tensordict)
+            loss_critic, value_clip_fraction = self.loss_critic(tensordict)
             td_out.set("loss_critic", loss_critic)
+            if value_clip_fraction is not None:
+                td_out.set("value_clip_fraction", value_clip_fraction)
         td_out = td_out.named_apply(
             lambda name, value: _reduce(value, reduction=self.reduction).squeeze(-1)
             if name.startswith("loss_")
--- a/torchrl/objectives/a2c.py
+++ b/torchrl/objectives/a2c.py
@@ -75,7 +75,7 @@ class A2CLoss(LossModule):
             ``"mean"``: the sum of the output will be divided by the number of
             elements in the output, ``"sum"``: the output will be summed. Default: ``"mean"``.
         clip_value_loss (float, optional): If provided, it will be used to compute a clipped version of the value
-            prediction with respect to the input tensordict value estimate and use it to calculate the value loss.
+            prediction with respect to the input value estimate and use it to calculate the value loss.
             The purpose of clipping is to limit the impact of extreme value predictions, helping stabilize training
             and preventing large updates. However, it will have no impact if the value estimate was done by the current
             version of the value estimator. Defaults to ``None``.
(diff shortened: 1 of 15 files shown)


torchrl/objectives/ppo.py
```diff
@@ -664,6 +664,8 @@ class ClipPPOLoss(PPOLoss):
             ``"none"`` | ``"mean"`` | ``"sum"``. ``"none"``: no reduction will be applied,
             ``"mean"``: the sum of the output will be divided by the number of
             elements in the output, ``"sum"``: the output will be summed. Default: ``"mean"``.
+        clip_value_loss (bool, optional): if ``True``, the value loss will be clipped with respect to the
```
vmoens, line 667: I would have this
- not provided / False = no clipping
- True = same clipping as log ratio
- number = clip with that number 

That way you can reuse the same semantic for other losses

albertbou92:
We can potentially add it to more losses (PPOLoss, KLPENPPOLoss, A2C, Reinforce) with the only difference that we would require the user to provide a float and not a bool since we can not default to the log ratio clipping value. 

But I am not sure if we should see if the feature is asked first for those other losses, since I have never seen in used in those.

vmoens:
I don't thinks it's a problem to add it, we don't want to copy existing repos but enable people to experiment and swap configurations quickly.

torchrl/objectives/reinforce.py
```diff
@@ -433,6 +456,20 @@ def loss_critic(self, tensordict: TensorDictBase) -> torch.Tensor:
                 f"TDLambdaEstimate and TDEstimate all return a 'value_target' entry that "
                 f"can be used for the value loss."
             )
+
+        if self.clip_value_loss:
+            self.clip_value_loss = self.clip_value_loss.to(state_value.device)
```
vmoens, line 461: ditto, let's avoid the casting of the attribute but cast the variable only

torchrl/objectives/ppo.py
```diff
@@ -524,6 +547,20 @@ def loss_critic(self, tensordict: TensorDictBase) -> torch.Tensor:
             state_value,
             loss_function=self.loss_critic_type,
         )
+
+        if self.clip_value_loss:
+            self.clip_value_loss = self.clip_value_loss.to(state_value.device)
```
vmoens, line 552: ditto, let's avoid the casting of the attribute but cast the variable only

vmoens (commented):
Thanks for this! Cool feature! 
A couple of comments along the way but otherwise LGTM

albertbou92:
I added the clip_fraction for both the PPO loss and the Value loss

torchrl/objectives/a2c.py
```diff
@@ -474,8 +523,10 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
             td_out.set("entropy", entropy.detach().mean())  # for logging
             td_out.set("loss_entropy", -self.entropy_coef * entropy)
         if self.critic_coef:
-            loss_critic = self.loss_critic(tensordict)
+            loss_critic, value_clip_fraction = self.loss_critic(tensordict)
             td_out.set("loss_critic", loss_critic)
+            if value_clip_fraction is not None:
+                td_out.set("value_clip_fraction", value_clip_fraction)
```
vmoens, line 529: we should include that key in the out_keys

torchrl/objectives/ppo.py
```diff
@@ -506,6 +527,16 @@ def loss_critic(self, tensordict: TensorDictBase) -> torch.Tensor:
                 f"can be used for the value loss."
             )
 
+        if self.clip_value:
+            try:
+                old_state_value = tensordict.get(self.tensor_keys.value).clone()
```
vmoens, line 532: why do we clone here?

torchrl/objectives/a2c.py
```diff
@@ -445,7 +475,26 @@ def loss_critic(self, tensordict: TensorDictBase) -> torch.Tensor:
...
+            )
+            # Chose the most pessimistic value prediction between clipped and non-clipped
+            loss_value = torch.max(loss_value, loss_value_clipped)
+            clip_fraction = (
+                (state_value / old_state_value)
+                .clamp(1 - clip_value, 1 + clip_value)
+                .abs()
+                .detach()
+            )
+        return self.critic_coef * loss_value, clip_fraction
```
vmoens, line 497: That seems to be a lot of duplicated code, can we move that to objectives.utils within a private function?

torchrl/objectives/a2c.py
```diff
@@ -445,7 +475,26 @@ def loss_critic(self, tensordict: TensorDictBase) -> torch.Tensor:
...
+                state_value - old_state_value
+            ).clamp(-clip_value, clip_value)
+            loss_value_clipped = distance_loss(
+                target_return,
+                state_value_clipped,
+                loss_function=self.loss_critic_type,
+            )
+            # Chose the most pessimistic value prediction between clipped and non-clipped
+            loss_value = torch.max(loss_value, loss_value_clipped)
+            clip_fraction = (
```
vmoens, line 491: I don't get what clip_fraction represents

One thing I like to look at is what proportion of data is being clipped

Also we should be careful about this since it doesn't come for free. The overhead introduced by logging this metric could impact performance...

torchrl/objectives/a2c.py
```diff
@@ -445,7 +475,26 @@ def loss_critic(self, tensordict: TensorDictBase) -> torch.Tensor:
...
+                (state_value / old_state_value)
+                .clamp(1 - clip_value, 1 + clip_value)
+                .abs()
+                .detach()
```
vmoens, line 495: Instead of detach() I think we should put the whole thing within a torch.no_grad()

torchrl/objectives/ppo.py
albertbou92: Because the `self.tensor_keys.value` prediction will be recomputed once we pass tensordict through the value network and we will lose this tensor.

torchrl/objectives/a2c.py
albertbou92: Yes, as I understand it, it is a measure of how much the old model that collected the data differs from the current model. If the fraction values stay close to 1.0 the model is changing slowly.  If the ratio consistently deviates from 1.0 and approaches or hits the clipping value, it indicates that the new model is substantially different from the old policy, indicating a faster rate of policy change.

torchrl/objectives/a2c.py
albertbou92: Makes sense. Otherwise the oprations up the detach are still tracked by autograd right?

vmoens (approved):
LGTM

PR #2008, opened 2024-03-11
[Feature] RB MultiStep transform

vmoens:
- [x] Allow rbs to handle extend with no data (will/may happen during the first iterations when calling extend / add)
- [x] Design class
    - [x] Handle change of horizon correctly
- [x] Test equivalence with MultiStep in collectors
- [x] Document feature + compare with collector version

Test script

```python
import torch

from torchrl.envs import GymEnv, TransformedEnv, StepCounter, SerialEnv
from torchrl.envs.transforms.rb_transforms import MultiStepTransform
from tensordict.utils import assert_allclose_td

# env = TransformedEnv(SerialEnv(2, lambda:GymEnv("CartPole-v1")), StepCounter())
env = TransformedEnv(GymEnv("CartPole-v1"), StepCounter())

env.set_seed(0)
torch.manual_seed(0)

t = MultiStepTransform(3, 0.98)

outs_2 = []
td = env.reset()
for _ in range(1):
    rollout = env.rollout(250, auto_reset=False, tensordict=td, break_when_any_done=False)
    out = t._inv_call(rollout)
    td = rollout[..., -1]
    outs_2.append(out)

outs_2 = torch.cat(outs_2, -1).split([47, 50, 50, 50, 50], -1)

t = MultiStepTransform(3, 0.98)

env.set_seed(0)
torch.manual_seed(0)

outs = []
td = env.reset()
for i in range(5):
    rollout = env.rollout(50, auto_reset=False, tensordict=td, break_when_any_done=False)
    out = t._inv_call(rollout)
    assert_allclose_td(out, outs_2[i])
    td = rollout[..., -1]["next"]
    outs.append(out)

outs = torch.cat(outs, -1)

```
cc @AechPro

### DIFF
--- a/docs/source/reference/data.rst
+++ b/docs/source/reference/data.rst
@@ -823,3 +823,11 @@ Utils
     consolidate_spec
     check_no_exclusive_keys
     contains_lazy_spec
+
+.. currentmodule:: torchrl.envs.transforms.rb_transforms
+
+.. autosummary::
+    :toctree: generated/
+    :template: rl_template.rst
+
+    MultiStepTransform
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -547,7 +547,7 @@ class TestDQN(LossModuleTestBase):
         loss_fn2 = DQNLoss(actor, loss_function="l2", delay_value=delay_value)
         loss_fn2.load_state_dict(sd)
 
-    @pytest.mark.parametrize("n", range(4))
+    @pytest.mark.parametrize("n", range(1, 4))
     @pytest.mark.parametrize("delay_value", (False, True))
     @pytest.mark.parametrize("device", get_default_devices())
     @pytest.mark.parametrize("action_spec_type", ("one_hot", "categorical"))
@@ -579,7 +579,7 @@ class TestDQN(LossModuleTestBase):
 
         with torch.no_grad():
             loss = loss_fn(td)
-        if n == 0:
+        if n == 1:
             assert_allclose_td(td, ms_td.select(*td.keys(True, True)))
             _loss = sum(
                 [item for name, item in loss.items() if name.startswith("loss")]
@@ -1125,7 +1125,7 @@ class TestQMixer(LossModuleTestBase):
         loss_fn2 = QMixerLoss(actor, mixer, loss_function="l2", delay_value=delay_value)
         loss_fn2.load_state_dict(sd)
 
-    @pytest.mark.parametrize("n", range(4))
+    @pytest.mark.parametrize("n", range(1, 4))
     @pytest.mark.parametrize("delay_value", (False, True))
     @pytest.mark.parametrize("device", get_default_devices())
     @pytest.mark.parametrize("action_spec_type", ("one_hot", "categorical"))
@@ -1158,7 +1158,7 @@ class TestQMixer(LossModuleTestBase):
 
         with torch.no_grad():
             loss = loss_fn(td)
-        if n == 0:
+        if n == 1:
             assert_allclose_td(td, ms_td.select(*td.keys(True, True)))
             _loss = sum(
                 [item for name, item in loss.items() if name.startswith("loss")]
@@ -1801,7 +1801,7 @@ class TestDDPG(LossModuleTestBase):
                 raise NotImplementedError(k)
             loss_fn.zero_grad()
 
-    @pytest.mark.parametrize("n", list(range(4)))
+    @pytest.mark.parametrize("n", range(1, 4))
     @pytest.mark.parametrize("device", get_default_devices())
     @pytest.mark.parametrize("delay_actor,delay_value", [(False, False), (True, True)])
     def test_ddpg_batcher(self, n, delay_actor, delay_value, device, gamma=0.9):
@@ -1832,7 +1832,7 @@ class TestDDPG(LossModuleTestBase):
 
         with torch.no_grad():
             loss = loss_fn(td)
-        if n == 0:
+        if n == 1:
             assert_allclose_td(td, ms_td.select(*list(td.keys(True, True))))
             _loss = sum(
                 [item for name, item in loss.items() if name.startswith("loss_")]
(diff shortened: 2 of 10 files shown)


vmoens:
@AechPro I did not change the n_steps argument as you suggested for consistency within torchrl.
I'm open to change both but I'm not sure how to proceed to make things non-bc breaking

I'll ping you when the preview of the doc is built for you to check if things make sense, in the meantime you can look at the diff

AechPro:
@vmoens I understand that backwards compatibility is important, but I think if we expect torchrl to be largely used by a research community it seems a bit strange for the n_steps argument to behave in a slightly unexpected way. 

Consider the case where a user of torchrl is re-implementing an algorithm like RAINBOW in good faith, and they naturally set the n_step argument to 3 as suggested by the paper. Using the current behavior this would actually produce a multi-step return estimate with n=4 according to the algorithm implemented in the paper, which would meaningfully change the performance of the algorithm. I can imagine our hypothetical torchrl user becoming quite confused when they cannot replicate the results of RAINBOW despite having (seemingly) implemented everything correctly here. 

Further, if we expect new algorithms to be written using torchrl, this sort of discrepancy in the meaning of the multi-step parameter could introduce a point of confusion in the opposite direction: if a person **not** using torchrl is attempting to implement a paper which describes results from an algorithm implemented with torchrl using the multi-step return object, that person may find it difficult to replicate the results of the torchrl algorithm for the same reason as our hypothetical RAINBOW user.

In my opinion preserving the meaning of the parameter in the multi-step return algorithm is more important than backwards compatibility because of the two examples I presented above, but I understand others may feel differently. At the very least I would strongly advocate for some easy to see documentation about this behavior whenever there is a tutorial using the `MultiStep` object and in the docstring for the object itself so we can mitigate these two potential issues.

vmoens:
Makes sense! Then let's make this `n_step` congruent with what's expected by the community

vmoens:
Here's the doc https://docs-preview.pytorch.org/pytorch/rl/2008/reference/generated/torchrl.envs.transforms.rb_transforms.MultiStepTransform.html#torchrl.envs.transforms.rb_transforms.MultiStepTransform

AechPro:
I'm having a little trouble understanding the output of the example. We have the `MultiStepTransform` with `n_steps=3`, then we sample some timesteps and slice the first 5 entries from our replay buffer. The first entry in that slice is reporting a step count of `9`, which I assume is supposed to be equal to `5+n`, which would be `5 + 3 = 8` but that would still be an incorrect behavior. The first timestep emitted by the transform should be timestep 1 (or zero if you start counting at zero), because that contains the state from which we're computing the n-step return, so I would expect `rb[:]["step_count"][:, 0]` to be 1 (or zero) and the final entry at `rb[:]["step_count"][:, -1]` to be `T-n+1` because the internal buffers should only need to contain `n-1` waiting timesteps unless there is a terminal state at the end of the internal buffer, in which case it should just compute all of the remaining possible returns at each of those timesteps. 

I'm sure I must be misunderstanding what's happening in the example. Could you clarify?

vmoens:
Let me clarify:
Here we look at the step count at the root:
```python
>>> print("step_count", rb[:]["step_count"][:, :5])
step_count tensor([[[ 9],
         [10],
         [11],
         [12],
         [13]],

        [[12],
         [13],
         [14],
         [15],
         [16]]])
```
Env 0 has steps [9, 10, 11...] and env 1 [12, 13, 14,...]

Then we look at the "next" entry to see the shift.
Without MultiStep, we would have [10, 11, 12, ...] for env 0 and [13, 14, 15, ...] for env 1
Because we use multi-step with a shift of 3 we have [13, 14, 15, ...] and [16, 17, 18, ...] resp.
```python
>>> print("next step_count", rb[:]["next", "step_count"][:, :5])
next step_count tensor([[[13],
         [14],
         [15],
         [16],
         [17]],

        [[16],
         [17],
         [18],
         [19],
         [20]]])
```

Note that we're looking at the replay buffer content so it doesn't really matter what those values are (ie it's expected that it doesn't start at 0).

For a single env and n=3, you would have this data structure accessible in the buffer
```python
done:                             [0, 0, 0, 0, 0, 0, 0, 0, 0, 1]
step_count:                   [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
next, step_count_orig: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
next, step_count:          [4, 5, 6, 7, 8, 9, 10, 10, 10, 10]
```

If you think this isn't the desired behaviour I'd be happy to read what you think would be expected, but to me it looks like what multi-step requires.

AechPro:
Ah-hah, my apologies for misunderstanding. Thanks for the clarification, this looks good!

AechPro:
Although if you are looking to implement the change to the `n_step` parameter we spoke about earlier right now then the example with 1 env should be as follows:
```py
done:        [0, 0, 0, 0, 0, 0, 0, 0, 0, 1]
step_count:        [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
next, step_count_orig:        [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
next, step_count:        [3, 4, 5, 6, 7, 8, 9, 10, 10, 10]
```

vmoens:
So given what you're saying, what it should be is: 
n_steps=0 => error
n_steps=1 => no transform
n_steps=2 => 1 shift in the future
etc

I will change that thx

AechPro:
I'm not sure what the right outcome for `n_steps=0` is. It seems reasonable to think we should set all of the rewards to zero and then set all the next states to the current states (i.e. `step_count = t` and `next, step_count = t`) because this would turn the equation `return_estimates = batch["rewards"] + batch["next]["gammas"] * value_estimator(batch["next"]["observation"])` into `return_estimates = 0 + 1 * value_estimator(batch["next"]["observation"]) = value_estimator(batch["observation"])` provided we also set the gamma values to 1. This makes some amount of sense because the value of `n_steps` can be taken to mean the number of reward terms we need to incorporate before using an estimator of the value function in our return estimate, so zero there would just lead us to use our value estimator completely with no reward terms at all

With that said I'm not aware of anything in the literature using n=0, and I wonder if it would be confusing or if it's even necessary.

vmoens:
The way I see it n=0 is equivalent to "doing nothing" which can be achieved by... doing nothing haha

AechPro:
LOL 😅 yeah it makes sense to do it that way. I was just imagining a scenario where maybe a user is doing some sort of hyper-parameter investigation and they wanted to vary the value of n from 0 to some number without changing anything in the underlying learning algorithms, which is maybe a realistic thing to want if someone is interested in measuring the impact of rewards on the return estimate.

AechPro:
Yeah, fair enough. Let's keep it as you suggested then!

vmoens:
cc @agarwl this is an implementation of multi-step that allows to dynamically change the horizon during training

Issue #1862, opened 2024-02-01
[BUG] Ray's actor construction is broken in docstring


sergiovalmac:
## To Reproduce

Steps to reproduce the behavior.

This is a minimal example to reproduce the bug - it's been taken from the docstrings in `collectors/distributed/ray.py` :
```python

from torch import nn
from tensordict.nn import TensorDictModule
from torchrl.envs.libs.gym import GymEnv
from torchrl.collectors.collectors import SyncDataCollector
from torchrl.collectors.distributed import RayCollector


def env_maker():
    return GymEnv("Pendulum-v1", device="cpu")


policy = TensorDictModule(nn.Linear(3, 1), in_keys=["observation"], out_keys=["action"])

distributed_collector = RayCollector(
    create_env_fn=[env_maker],
    policy=policy,
    collector_class=SyncDataCollector,
    max_frames_per_traj=50,
    init_random_frames=-1,
    reset_at_each_iter=-False,
    collector_kwargs={
        "device": "cpu",
        "storing_device": "cpu",
    },
    num_collectors=1,
    total_frames=10000,
    frames_per_batch=200,
)
for i, data in enumerate(distributed_collector):
    if i == 2:
        print(data)
        break
```
Error messages and stack trace:
```bash
C:\Users\me\code\torchrl> python .\sample_envs.py
C:\Users\me\venvs\3_10\lib\site-packages\gymnasium\core.py:311: UserWarning: WARN: env.num_envs to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do `env.unwrapped.num_envs` for environment variables or `env.get_wrapper_attr('num_envs')` that will search the reminding wrappers.
  logger.warn(
C:\Users\me\venvs\3_10\lib\site-packages\gymnasium\core.py:311: UserWarning: WARN: env.reward_space to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do `env.unwrapped.reward_space` for environment variables or `env.get_wrapper_attr('reward_space')` that will search the reminding wrappers.
  logger.warn(
2024-02-01 00:11:48,502 INFO worker.py:1621 -- Started a local Ray instance.
Traceback (most recent call last):.\ 
File "C:\Users\me\code\torchrl\sample_envs.py", line 53, in <module>
    distributed_collector = RayCollector(
  File "C:\Users\me\venvs\3_10\lib\site-packages\torchrl\collectors\distributed\ray.py", line 463, in __init__
    self.add_collectors(
[...]

PR #2013, opened 2024-03-18
[CI] Upgrade 3.8 to 3.10 GPU jobs

vmoens:


vmoens:
cc @albertbou92 upgrading the GPU jobs to 3.10 makes the Ray jobs fail, but the 3.8 is starting to fail on its own due to Hydra issues (see #2014).
Any clue on what I should do here?

albertbou92:
This solution solves the problem in https://github.com/pytorch/rl/issues/1862: https://github.com/pytorch/rl/pull/2015

Although I now see an error that seems unrelated:
`TypeError: TensorDict.__init__() got an unexpected keyword argument 'filter_empty'`

PR #2017, opened 2024-03-18
[Feature] Enable parameter reset in loss module

BY571:
## Description

Allows to reset the parameters in the loss module.

### DIFF
--- a/torchrl/objectives/common.py
+++ b/torchrl/objectives/common.py
@@ -9,9 +9,10 @@ import abc
 import warnings
 from copy import deepcopy
 from dataclasses import dataclass
-from typing import Iterator, List, Optional, Tuple
+from typing import Callable, Iterator, List, Optional, Tuple
 
 import torch
+import torch.nn.init as init
 from tensordict import is_tensor_collection, TensorDict, TensorDictBase
 
 from tensordict.nn import TensorDictModule, TensorDictModuleBase, TensorDictParams
@@ -362,6 +363,35 @@ class LossModule(TensorDictModuleBase, metaclass=_LossMeta):
     def reset(self) -> None:
         # mainly used for PPO with KL target
         pass
+
+    def reset_parameters(
+        self,
+        module_names: Optional[List[Parameter]] = None,
+        init_func: Optional[Callable] = None,
+    ) -> None:
+        """Reset the parameters of the specified modules.
+
+        Args:
+            module_names (Optional[List[Parameter]]): A list of module names to reset the parameters for.
+                If None, all modules with names ending in "_params" will be reset.
+            init_func (Optional[Callable]): A function to initialize the parameters.
+                If None, the parameters will be initialized with uniform random values between -1 and 1.
+        """
+        if module_names is None:
+            params_2_reset = [
+                name for name in self._modules.keys() if name.endswith("_params")
+            ]
+        else:
+            params_2_reset = [name + "_params" for name in module_names]
+
+        def _reset_params(param):
+            if init_func is not None:
+                init_func(param.data)
+            else:
+                init.uniform_(param.data, -1, 1)
+
+        for name in params_2_reset:
+            getattr(self, name).apply(_reset_params)
 
     @property
     def value_estimator(self) -> ValueEstimatorBase:
--- a/examples/td3/td3.py
+++ b/examples/td3/td3.py
@@ -123,7 +123,8 @@ def main(cfg: "DictConfig"):  # noqa: F821
                 q_losses,
             ) = ([], [])
             for _ in range(num_updates):
-
+                if update_counter % 5 == 0:
+                    loss_module.reset_parameters()
                 # Update actor every delayed_updates
                 update_counter += 1
                 update_actor = update_counter % delayed_updates == 0


torchrl/objectives/common.py
```diff
@@ -363,6 +364,35 @@ def reset(self) -> None:
...
+        module_names: Optional[List[Parameter]] = None,
+        init_func: Optional[Callable] = None,
+    ) -> None:
+        """Reset the parameters of the specified modules.
+
+        Args:
+            module_names (Optional[List[Parameter]]): A list of module names to reset the parameters for.
+                If None, all modules with names ending in "_params" will be reset.
+            init_func (Optional[Callable]): A function to initialize the parameters.
+                If None, the parameters will be initialized with uniform random values between -1 and 1.
```
vmoens, line 378: Seems very unlikely that anyone would want to use that init IMO. Shouldn't we use the init method from the corresponding nn.Module if there is?

torchrl/objectives/common.py
```diff
@@ -363,6 +364,35 @@ def reset(self) -> None:
         # mainly used for PPO with KL target
         pass
 
+    def reset_parameters(
+        self,
+        module_names: Optional[List[Parameter]] = None,
+        init_func: Optional[Callable] = None,
```
vmoens, line 370:
```suggestion
        init_func: Callable[[torch.Tensor], None] | None = None,
```

torchrl/objectives/common.py
```diff
@@ -363,6 +364,35 @@ def reset(self) -> None:
         # mainly used for PPO with KL target
         pass
 
+    def reset_parameters(
+        self,
+        module_names: Optional[List[Parameter]] = None,
```
vmoens, line 369:
```suggestion
        module_names: List[Parameter] | None = None,
```

torchrl/objectives/common.py
```diff
@@ -363,6 +364,35 @@ def reset(self) -> None:
...
+        init_func: Optional[Callable] = None,
+    ) -> None:
+        """Reset the parameters of the specified modules.
+
+        Args:
+            module_names (Optional[List[Parameter]]): A list of module names to reset the parameters for.
```
vmoens, line 375:
```suggestion
            module_names (list of nn.Parameter, optional): A list of module names to reset the parameters for.
```

torchrl/objectives/common.py
```diff
@@ -363,6 +364,35 @@ def reset(self) -> None:
...
+                If None, all modules with names ending in "_params" will be reset.
+            init_func (Optional[Callable]): A function to initialize the parameters.
```
vmoens, line 377:
```suggestion
            init_func (Callable[[torch.Tensor], None]): A function to initialize the parameters.
```

torchrl/objectives/common.py
```diff
@@ -363,6 +364,35 @@ def reset(self) -> None:
...
+                If None, the parameters will be initialized with uniform random values between -1 and 1.
+        """
+        if module_names is None:
+            params_2_reset = [
+                name for name in self._modules.keys() if name.endswith("_params")
+            ]
+        else:
+            params_2_reset = [name + "_params" for name in module_names]
+
+        def _reset_params(param):
```
vmoens, line 387: Having one single reset function will be hard to handle, we need a way to tie the reset function and the module.

vmoens (commented):
Thanks for this!

We'll need tests for the feature.

How do we handle the target parameters?

Wouldn't something like this be a bit more robust?

```python
from torchrl.objectives import DQNLoss
from torchrl.modules import QValueActor
from torch import nn

module = nn.Sequential(nn.Linear(1, 64), nn.ReLU(), nn.Linear(64, 64))

value_net = QValueActor(module=module, action_space="categorical")
loss = DQNLoss(value_network=value_net, action_space="categorical")

with loss.value_network_params.to_module(loss.value_network):
    loss.apply(lambda module: module.reset_parameters() if hasattr(module, "reset_parameters") else None)
```

vmoens (commented):
Thanks for this!

We'll need tests for the feature.

How do we handle the target parameters?

Wouldn't something like this be a bit more robust?

```python
from torchrl.objectives import DQNLoss
from torchrl.modules import QValueActor
from torch import nn

module = nn.Sequential(nn.Linear(1, 64), nn.ReLU(), nn.Linear(64, 64))

value_net = QValueActor(module=module, action_space="categorical")
loss = DQNLoss(value_network=value_net, action_space="categorical")

with loss.value_network_params.to_module(loss.value_network):
    loss.apply(lambda module: module.reset_parameters() if hasattr(module, "reset_parameters") else None)
```

BY571:
> Thanks for this!
> 
> We'll need tests for the feature.
> 
> How do we handle the target parameters?
> 
> Wouldn't something like this be a bit more robust?
> 
> ```python
> from torchrl.objectives import DQNLoss
> from torchrl.modules import QValueActor
> from torch import nn
> 
> module = nn.Sequential(nn.Linear(1, 64), nn.ReLU(), nn.Linear(64, 64))
> 
> value_net = QValueActor(module=module, action_space="categorical")
> loss = DQNLoss(value_network=value_net, action_space="categorical")
> 
> with loss.value_network_params.to_module(loss.value_network):
>     loss.apply(lambda module: module.reset_parameters() if hasattr(module, "reset_parameters") else None)
> ```

I like the solution! But we are accessing the parameters directly in your example so we would need to define a reset function manually, which I think is perfectly fine because then the user has to decide the way how to reset weights and biases:



```python
def reset_parameters(params):
    """ User specified resetting function depending on their needs for initialization """
    if len(params.shape) > 1:
        # weights
        nn.init.xavier_uniform_(params)
    elif len(params.shape) == 1:
        # biases
        nn.init.zeros_(params)
    else:
        raise ValueError("Unknown parameter shape: {}".format(params.shape))
  
with loss.value_network_params.to_module(loss.value_network):
    loss.apply(lambda x: reset_parameters(x.data) if hasattr(x, "data") else None)
```

And for handling the target_network_params I think we could simply do something like:
```python
loss.target_value_network_params.update(loss.value_network_params)
```

What do you think? I think we can close the draft. But we might want to mention the way to reset parameters somewhere in the docs.

vmoens:
> loss.target_value_network_params.update(loss.value_network_params)

This won't work because the target params are locked (you can't update them). They're locked because we want to avoid this kind of operation :) 
You should update the data inplace:
```python
loss.target_value_network_params.apply(lambda dest, src: dest.data.copy_(src), loss.value_network_params)
```

vmoens:
```python
def reset_parameters(params):
    """ User specified resetting function depending on their needs for initialization """
    if len(params.shape) > 1:
        # weights
        nn.init.xavier_uniform_(params)
    elif len(params.shape) == 1:
        # biases
        nn.init.zeros_(params)
    else:
        raise ValueError("Unknown parameter shape: {}".format(params.shape))
  
with loss.value_network_params.to_module(loss.value_network):
    loss.apply(lambda x: reset_parameters(x.data) if hasattr(x, "data") else None)
```
Unfortunately this isn't very generic
(1) all tensors have a data attribute, even buffers. By doing this you will also use Xavier init on batch-norm buffers if they're 2d
(2) If the model has a mixture of linear, conv and other layers it's going to be hard to have a fine grained control over the params being updated.

Not all modules are "weights" and "biases" and "biases" can be 2d (my point is: the dimension is a very indirect determinator of the tensor role in a model)

The way I usually see this work is to use the module `reset_parameters` if there is one, which provides a better control over difference in initialization methods. 

Maybe we could allow the user to pass a reset function, but in that case we don't even need to re-populate the module (we can just do `tensordict.apply(reset)`). Note that you could also do
```python
def reset(name, tensor):
    if name == "bias":
        tensor.data.zero_()
    if name == "weight":
        nn.init.xavier_uniform_(tensor)
tensordict.apply(reset, named=True)
```
which is more straightforward IMO

gtnv:
hey @BY571, is this still something you're working on? if not, should we close it?

PR #2021, opened 2024-03-18
[BugFix] Fix clip_fraction in PO losses

vmoens:
cc @albertbou92 @wbinventor


albertbou92:
This one also need to be corrected
https://github.com/pytorch/rl/blob/main/torchrl/objectives/ppo.py#L864

PR #2024, opened 2024-03-19
[Feature] "fork" start method for mutli-collectors

vmoens:


gtnv:
hi @vmoens, is this still being worked on? if not, should we close it for now?

PR #2031, opened 2024-03-21
[Performance] Faster slice sampler

vmoens:
TODO:
 - Make sure that this is working OK with shared buffers (extend on one worker won't erase the buffer on the other!)

cc @ahmed-touati @Cadene

### DIFF
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -1835,7 +1835,6 @@ class TestMultiProc:
     def test_error_prb(self):
         # PrioritizedSampler cannot be shared
         with pytest.raises(RuntimeError, match="cannot be shared between processes"):
-            # with pytest.warns():
             self.exec_multiproc_rb(
                 sampler_type=lambda: PrioritizedSampler(21, alpha=1.1, beta=0.5)
             )
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -1835,6 +1835,7 @@ class TestMultiProc:
     def test_error_prb(self):
         # PrioritizedSampler cannot be shared
         with pytest.raises(RuntimeError, match="cannot be shared between processes"):
+            # with pytest.warns():
             self.exec_multiproc_rb(
                 sampler_type=lambda: PrioritizedSampler(21, alpha=1.1, beta=0.5)
             )
--- a/torchrl/data/replay_buffers/samplers.py
+++ b/torchrl/data/replay_buffers/samplers.py
@@ -637,7 +637,18 @@ class SliceSampler(Sampler):
             if the last element of the trajectory tensor is identical to the first,
             the same trajectory spans across end and beginning.
         cache_values (bool, optional): to be used with static datasets.
-            Will cache the start and end signal of the trajectory.
+            Will cache the start and end signal of the trajectory. This can be safely used even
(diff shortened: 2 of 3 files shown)


vmoens:
Some benchmarks:
```python
import time

import torch
import tqdm

from torchrl.data import ReplayBuffer, SliceSampler, LazyTensorStorage

for compile in [True, False]:
    for cached in [True, False]:

        rb = ReplayBuffer(storage=LazyTensorStorage(1_000_000),
                          sampler=SliceSampler(num_slices=16, traj_key="traj_idx", compile=compile, cache_values=cached),
                          batch_size=256)

        tds = TensorDict({
            "traj_idx": torch.arange(1_000_000) // 100,
            "x": torch.randn(1_000_000),
            ("next", "y"): torch.randn(1_000_000),
        }, [1_000_000]).split(1000)


        def iter_over_tds():
            while True:
                yield from tds


        iterator = iter_over_tds()
        rb.extend(next(iterator))
        rb.sample()

        n_samples = 20
        t0 = time.time()
        for i, data in tqdm.tqdm(enumerate(iterator), total=5000, desc=f"compile={compile}, cache={cached}"):
            rb.extend(data)
            for j in range(n_samples):
                rb.sample()
            if i == 5000:
                break
        print(f"compile={compile}, cache={cached}, time={time.time() - t0: 4.4f}")

```


Results:
```
compile=True, cache=True: 100%|██████████| 5000/5000 [00:16<00:00, 307.65it/s]
compile=True, cache=True, time= 16.2532
compile=True, cache=False: 100%|██████████| 5000/5000 [01:39<00:00, 50.17it/s]
compile=True, cache=False, time= 99.6700
compile=False, cache=True: 100%|██████████| 5000/5000 [00:18<00:00, 265.16it/s]
compile=False, cache=True, time= 18.8568
compile=False, cache=False: 100%|██████████| 5000/5000 [06:26<00:00, 12.95it/s]
compile=False, cache=False, time= 386.0987

```

So we have a clear very impressive gain of cache (compile=True => 625%, compile = False => 2000%) and some gain thanks to compile too (cache = True => 110%, cache=False => 388%)
The total improvement compared to the previous state for this example is **2383%** (24x faster)!

test/test_rb.py
```diff
@@ -1835,6 +1835,7 @@ def test_error_maxwriter(self):
     def test_error_prb(self):
         # PrioritizedSampler cannot be shared
         with pytest.raises(RuntimeError, match="cannot be shared between processes"):
+            # with pytest.warns():
```
vmoens, line 1838:
```suggestion
```

torchrl/data/replay_buffers/samplers.py
```diff
@@ -360,6 +360,14 @@ def __getstate__(self):
...
                 f"Samplers of type {type(self)} cannot be shared between processes."
             )
+            # TODO: Use a warning instead
+            # logger.warning(
+            #     f"It seems you are sharing a {type(self).__name__} across processes. "
+            #     f"If the sampler is queried on another process than the one populating "
+            #     f"the storage, this will fail as the sampler is updated locally and "
+            #     f"not globally. If this feature is required, please file an issue on "
+            #     f"torchrl GitHub."
+            # )
```
vmoens, line 370:
```suggestion
```

Issue #2036, opened 2024-03-22
[BUG] VecNorm.to_observation_norm broken for multiple keys


maxweissenbacher:
## To Reproduce

```
gym_env = GymEnv("MountainCarContinuous-v0", device='cpu')
transformed_env = TransformedEnv(gym_env, VecNorm(in_keys=["observation", "reward"]))
transformed_env.rollout(100)
transformed_env.transform[0].to_observation_norm()
```

## Expected behavior

Should return ObservationNorm but gives following error:

```
...
out += ObservationNorm
TypeError: 'type' object is not iterable
```

## Reason and Possible fixes

Using the VecNorm transform with only one key works - so to do it for multiple keys one can just use it repeatedly with single keys.

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)
- [x] I have read the [documentation](https://github.com/pytorch/rl/tree/main/docs/) (**required**)
- [x] I have provided a minimal working example to reproduce the bug (**required**)

maxweissenbacher:
Another bug here:

When using the VecNorm transform with only one in_key, but a key that isn't a pure string, to_observation_norm also fails. There is an easy fix for this: the to_observation_norm method is missing a call to `key = self._key_str(key)`. I will create a pull request to add this if that's okay with you.

An example of this situation might occur for instance in the multiagent setting, where the observations might be stored under the key `("agents", "observation")`.

PR #2045, opened 2024-03-28
[BugFix] to_observation_norm now works with keys which are not strings

maxweissenbacher:
## Description

Added the _key_str method in the to_observation_norm method of VecNorm. A call to to_observation_norm would have previously given an error, since in the update method, a call is made to _key_str to store the sum argument etc. However, this call was not made in to_observation_norm, resulting in the key not being able to be found in _td.

Are there any tests that need to be updated to reflect this?

I've raised this in issue #2036.

## Types of changes

## Checklist


vmoens:
I took the liberty of making a couple of edits. I paved the way for this with some commented changes, I will leave them here as a suggestion for anyone who'd be interested in implementing that.

The next step is to store the tracking values in the tensordict rather than in the transform (we could add an underscore before the key string to indicate that these values are private and should not be returned by collectors for instance)

PR #2047, opened 2024-03-29
[Tutorial] PettingZoo Parallel competitive environment

matteobettini:


tutorials/sphinx-tutorials/simple_tag.gif
matteobettini, line 1: @vmoens can you upload on s3 and give the link pls?

vmoens:
I'd rather have them as a separate PR if possible

matteobettini:
> This would however be a synchronous environment so the tutorial would **NOT SHOW**:

Moved the changes to https://github.com/pytorch/rl/pull/2048

matteobettini:
I have filled in the python code of the tutorial, which now runs.

Now I will focus on structuring it and rewriting the text

tutorials/sphinx-tutorials/off_policy_training_pettingzoo_vmas.png
matteobettini, line 1: also this needs s3

matteobettini:
@vmoens  this is ready for review

matteobettini:
@elliottower here is the first draft of the tuto in case you are curious and want to have a look.

We still need to upload a few things to s3 and to run the ci for you to see a rendered version

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+    )
+
+    critics[group] = TensorDictSequential(
+        cat_module, critic_module
+    )  # Run them in sequence
+
+
+######################################################################
+# Let us try our policy and critic modules. As pointed earlier, the usage of
+# :class:`TensorDictModule` makes it possible to directly read the output
```
vmoens, line 626:
```suggestion
# :class:`~tensordict.nn.TensorDictModule` makes it possible to directly read the output
```

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+replay_buffers = {}
+for group, _agents in env.group_map.items():
+    replay_buffer = ReplayBuffer(
+        storage=LazyTensorStorage(
+            memory_size, device=device
+        ),  # We will store up to memory_size multi-agent transitions
+        sampler=RandomSampler(),
+        batch_size=train_batch_size,  # We will sample batches of this size
+    )
+    replay_buffers[group] = replay_buffer
```
vmoens, line 689: what is preventing us from building a single RB?

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+#
+# Replay buffers are a common building piece of off-policy RL sota-implementations.
+# There are many types of buffers, in this tutorial we use a basic buffer to store and sample tensordict
+# data randomly.
+#
+
+replay_buffers = {}
+for group, _agents in env.group_map.items():
+    replay_buffer = ReplayBuffer(
+        storage=LazyTensorStorage(
```
vmoens, line 683: By default I would advise people to use Memmap

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+    group: {
+        "loss_actor": torch.optim.Adam(
+            loss.actor_network_params.flatten_keys().values(), lr=lr
+        ),
+        "loss_value": torch.optim.Adam(
+            loss.value_network_params.flatten_keys().values(), lr=lr
+        ),
+    }
+    for group, loss in losses.items()
+}
```
vmoens, line 734: when I see that we make a different loss, a different rb and a different optimizer for each group I wonder if we aren't doing things wrong. This looks like smth that could be simplified. 
After reading this code and knowing the library, I'm pretty sure I couldn't get the whole script right the first time!

vmoens: In other words, I'd like to understand why we can't just have one RB, one loss and one optimizer.

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+    excluded_keys = []
+    for other_group in env.group_map.keys():
+        if other_group != group:
+            excluded_keys += [other_group, ("next", other_group)]
+    excluded_keys += ["info", (group, "info"), ("next", group, "info")]
+    return excluded_keys
+
+
+def process_batch(
+    batch: TensorDictBase, group_map: Dict[str, List[str]]
```
vmoens, line 761: if group_map is constant, maybe we can remove it from the args?

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+# Training utils
+# --------------
+#
+# We do have to define two helper functions that we will use in the training loop.
+# They are very simple and do not contain any important logic.
+#
+#
+
+
+def get_excluded_keys(group: str):
```
vmoens, line 746: this looks like a piece of code that will be executed many times but could be cached once and for all

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+#    env.frames[0].save(
+#        f"{scenario_name}.gif",
+#        save_all=True,
+#        append_images=env.frames[1:],
+#       duration=3,
+#       loop=0,
+#    )
+#
+#    from IPython.display import Image
+#    Image(open(f"{scenario_name}.gif", "rb").read())
```
vmoens, line 959: For consistency with the rest of the lib I'd rather use VIdeoRecorder
https://github.com/pytorch/rl/pull/2070

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+#
+# In this tutorial, we have seen:
+#
+# - How to create a competitive multi-group multi-agent environment in TorchRL, how its specs work, and how it integrates with the library;
+# - How to create multi-agent network architectures in TorchRL for multiple groups;
+# - How we can use :class:`tensordict.TensorDict` to carry multi-agent multi-group data;
+# - How we can tie all the library components (collectors, modules, replay buffers, and losses) in a multi-agent multi-group MADDPG/IDDPG training loop.
+#
+# Now that you are proficient with multi-agent DDPG, you can check out all
+# `TorchRL multi-agent examples <https://github.com/pytorch/rl/tree/main/sota-implementations/multiagent>`__.
```
vmoens, line 975: let's avoid all links to code if possible

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+# These are code-only scripts of many popular MARL sota-implementations such as the ones seen in this tutorial,
+# QMIX, MADDPG, IQL, and many more!
+#
+# Also do remember to check out our
+# `multi-agent PPO tutorial <https://pytorch.org/rl/tutorials/multiagent_ppo.html>`__ in VMAS.
```
vmoens, line 980: ditto + this one can be handled by sphinx

tutorials/sphinx-tutorials/simple_tag.gif
vmoens: unfortunately no, not atm :(

vmoens (commented):
Thanks for this, very complete!

I'm wondering if it isn't a bit too long. The code is about 1k lines of code long, there's a lot of theory and code. When someone goes through this tuto, what should they get out of it exactly (a theory about MARL, a way to code DDPG in MARL, how to use the env?)
Can't this be split in more tutos? Or can we remove some extra bits that aren't 100% required to train the algo?

vmoens (commented):
Thanks for this, very complete!

I'm wondering if it isn't a bit too long. The code is about 1k lines of code long, there's a lot of theory and code. When someone goes through this tuto, what should they get out of it exactly (a theory about MARL, a way to code DDPG in MARL, how to use the env?)
Can't this be split in more tutos? Or can we remove some extra bits that aren't 100% required to train the algo?

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+# was called.
+# The transformed environment will inherit
+# the device and meta-data of the wrapped environment, and transform these depending on the sequence
+# of transforms it contains.
+#
+
+env = TransformedEnv(
+    env,
+    RewardSum(
+        in_keys=env.reward_keys, reset_keys=["_reset"] * len(env.group_map.keys())
```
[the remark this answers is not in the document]
matteobettini: There is a reward key per agent group but just one `_reset` key.  The constructor wants one reset key per reward key

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
matteobettini: we do not want to mix data from different groups. We may decide to collect more for one group or smth like that

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
matteobettini: Plus we want to allow different groups to have different buffers losses and lrs

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
matteobettini: This is for interactive rendering, not for video recording. These envs have a GUI rendering function and do not return pixels in the data

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+  able to simulate multiple environments in a GPU batch to speed up computation.
+
+In the *simple_tag* environment,
+there are two teams of agents: the chasers (or "adversaries") and the evaders (or "agents").
+Chasers are rewarded for touching evaders. Upon a contact the team of chasers is collectively rewarded and the
+evader touched is penalized with the same value. Evaders have higher speed and acceleration than chasers.
+
+
+
+.. figure:: https://pytorch.s3.amazonaws.com/torchrl/github-artifacts/img/simple_tag.gif
```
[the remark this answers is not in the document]
matteobettini: You mean I should use a private host from my side or you are finding a torchrl owned host?

matteobettini:
> Thanks for this, very complete!
> 
> I'm wondering if it isn't a bit too long. The code is about 1k lines of code long, there's a lot of theory and code. When someone goes through this tuto, what should they get out of it exactly (a theory about MARL, a way to code DDPG in MARL, how to use the env?) Can't this be split in more tutos? Or can we remove some extra bits that aren't 100% required to train the algo?

What the tuto is doing is showing the env and the grouping mechanism and training it.

- Regarding the code: All the python code is necessary for training, except the section where i print the env specs and rollout which i think is needed to show readers the agent groups
- Regarding the text: this follows the PPO and multi-agent PPO format. I can remove some theory, this was there as it motivates some choices on how to centralize the critic and so on, but if you think too verbous we can cut. I can also cut the DDPG explaination

docs/requirements.txt
```diff
@@ -25,4 +25,5 @@ imageio[ffmpeg,pyav]
 memory_profiler
 pyrender
 pytest
-vmas==1.2.11
+vmas
```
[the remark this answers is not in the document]
matteobettini: We are making sure that all runs with the latest VMAS version and I will personally make sure vmas stays BC-compatible

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -0,0 +1,992 @@
...
+# where a deterministic policy is optimized using the gradients from the critic network.
+# For more information, see the `Deep Deterministic Policy Gradients <https://arxiv.org/abs/1509.02971>`_ paper.
+#
+# This type of algorithm is usually trained *off-policy*. At every learning iteration, we have a
+# **sampling** and a **training** phase. In the **sampling** phase of iteration :math:`t`, rollouts are collected
+# form agents' interactions in the environment using the policies :math:`\mathbf{\pi}_t` and stored in the replay buffer.
+# In the **training** phase, rollouts from any time prior and including :math:`t` are randomly sampled from the replay buffer and fed
+# to the training process to perform backpropagation. This leads to updated policies which are then used again for sampling.
+# It is important to note that, unlike on-policy methods, any policy can be used to collect the data fed to the training phase
+# as these methods learn the optimal value function. In fact many users like to prefill their buffer with data from a random policy.
```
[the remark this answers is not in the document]
matteobettini: This whole pararaphs were removed entirely

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
matteobettini: it is extremely cheap as it just loops among 2 entries of a python dictionary.

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
matteobettini: I want to reference the sota-implementations folder. How else can I do it?

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
vmoens: Sure! Just looks awkward to define that as a separate function in a turorial IMO
When I see a function in a tuto like this I assume it's some complex stuff that must be done on-the-fly

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
vmoens: Not sure but it's dangerous because we don't test the links so if we move things from a tiny bit we'll have to manually go through the entire repo looking for this link.

vmoens:
> * Regarding the text: this follows the PPO and multi-agent PPO format. I can remove some theory, this was there as it motivates some choices on how to centralize the critic and so on, but if you think too verbous we can cut. I can also cut the DDPG anf off-policy explainations

I guessed so, but most feedback I got even from very close, top collaborators such as yourself is that our tutorials are very long and hard to follow, if not worse than that.
If we can make sure that tutorials that are aimed at a large audience (which I suspect is the case here since it starts with the very basics like off-policy vs on-policy) are also completed in 10 mins top I think they will be more impactful.

Concretely, I would either make this tutorial advanced and presuppose that users have an extensive knowledge of MARL and RL in general, and just point them to refs in case context is needed, or split the tutorial in 2-5 sub-tutorials

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
matteobettini: i'll simplify this

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
matteobettini: Still if there is no alternative I would prefer to keep the reference as I think the users miss out if they are not aware of these code examples

matteobettini:
> Can we unify that? If not what's the blocker?

I answered that in the thread. We want to keep those separate to allow different choices across groups. Plus losses cannot physicall work with multiple groups

matteobettini:
> Got it!
> Not super clean that we ask people to pass _reset, which is private
> We should ask for done instead

The problem is that the dones are still one per group plus the shared one. The _reset on the other hand is just one

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
matteobettini: It is not trivial as the env is batched. To use the recorder I would have to call the render in the callback of the rollout and write it to the pixels key of the td. But i can't since the render is from a single env and the td has the batch_size containing data from multiple environments.

These are just a few lines to display the gif in colab and they are also commented out. 

To use a recorder here i would have to create a bunch of infrastructure and I think it would just get out of scope.

vmoens:
> > Got it!
> > Not super clean that we ask people to pass _reset, which is private
> > We should ask for done instead
> 
> The problem is that the dones are still one per group plus the shared one. The _reset on the other hand is just one

IIRC the global reset/done precedes over the local ones. 

Am I right that what you're trying to do is override the local resets with the global one then?
Couldn't we use "done" like you're using reset?

vmoens:
> > Can we unify that? If not what's the blocker?
> 
> I answered that in the thread. We want to keep those separate to allow different choices across groups. Plus losses cannot physicall work with multiple groups

Sorry I missed this, the thread must have been closed. 

I don't understand this statement though, can you elaborate?
In particular

> losses cannot physicall work with multiple groups

what does "physicall" mean here?
What is blocking us to use a single loss across groups?

> We want to keep those separate to allow different choices across groups

What "choices"? Hyperparameters?
What I see is that they are all the same so unless we use different ones it isn't immediate to understand.

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
vmoens: I'll patch things for you it should be simple

matteobettini:
> Am I right that what you're trying to do is override the local resets with the global one then? Couldn't we use "done" like you're using reset?

So in pettingzoo we have the global done and the local dones but only the global _reset. So the local resets just do not exist.

since the class needs a reset for each agent key, we are telling all reward keys to track the global reset.

matteobettini:
> > > Can we unify that? If not what's the blocker?
> > 
> > 
> > I answered that in the thread. We want to keep those separate to allow different choices across groups. Plus losses cannot physicall work with multiple groups
> 
> Sorry I missed this, the thread must have been closed.
> 
> I don't understand this statement though, can you elaborate? In particular
> 
> > losses cannot physicall work with multiple groups
> 
> what does "physicall" mean here? What is blocking us to use a single loss across groups?
> 
> > We want to keep those separate to allow different choices across groups
> 
> What "choices"? Hyperparameters? What I see is that they are all the same so unless we use different ones it isn't immediate to understand.

Ok let me elaborate more, because this is quite a fundamental concept about the whole MARL api and BenchMARL.

- losses are stateful components that only take one set of keys (reward,state,done, and so on). Therefore they cannot operate across multiple groups
- each group can have different global values, like global critic output, that sit at root. Therefore if you kept multiple groups in the same loss it is unclear how to handle these
- in the example of this tutorial, we stop training one group after x iterations. This would not be possible with a shared buffer/loss/optimizer
- 

The core of the MARL API is that if you want a shared loss/buffer/optimizer you can keep agents in the same group. But when agents are in different groups they can be decoupled into individual components so that you can

- use different losses/buffers(/optimizers for each group
- Train them asynchronously
- not pollute the data of other groups when writing global keys for your group

vmoens:
Thanks for the clarification
I can see why it isn't easy to have one loss for multiple groups atm, but don't we want that to be a feature, rather than asking users to repeat the tedious construction of N different buffers, N different optimizers and N different losses?
It feels like there's a lot of boilerplate that we could get rid of without too much trouble from within the library.

Since we can use param_groups in the optimizers, we could obtain the same behaviour as having many optimizers by just splitting the param groups carefully. Note that this will come with significant speedups on cuda too since you'll be able to update your params asynchronously.

> Train them asynchronously

Isn't the training synchronous here? They're trained sequentially for a given number of steps no?

> in the example of this tutorial, we stop training one group after x iterations. This would not be possible with a shared buffer/loss/optimizer

Can you point me to the part of the tutorial that does that? What I see is that there's a constant number of steps for each component and the learning rates as well as other hyperparameters are all shared. From a non MARL trained eye it seems like there's a bunch of stuff that could be simplified (for instance why not having one replay buffer from which you sample and then get the relevant keys with indexing / select?)

matteobettini:
> Can you point me to the part of the tutorial that does that? What I see is that there's a constant number of steps for each component and the learning rates as well as other hyperparameters are all shared. From a non MARL trained eye it seems like there's a bunch of stuff that could be simplified (for instance why not having one replay buffer from which you sample and then get the relevant keys with indexing / select?)

```python
  # Stop training a certain group when a condition is met (e.g., number of training iterations)
    if iteration == iteration_when_stop_training_evaders:
        del train_group_map["agent"]
```

Again, at some point we may sample more from the buffer of one group or stop completely sampling from one group like in this tuto.

> Isn't the training synchronous here? They're trained sequentially for a given number of steps no?

I mean train them at different times. In this tuto for the seond half of the iterations we are trianing just one

> I can see why it isn't easy to have one loss for multiple groups atm, but don't we want that to be a feature, rather than asking users to repeat the tedious construction of N different buffers, N different optimizers and N different losses?
> It feels like there's a lot of boilerplate that we could get rid of without too much trouble from within the library.

I doi not agree. I think that this is not a feature we want as it is good to have the clear separation and have losses being single group.

Repeating the construction in a for loop is peanuts compared to the complexity that we would introduce by moving the for loop inside the losses (because at the end of the day the for loop will be somewhere since it is all heterogeneous networks). 

I do not think that the current way is boiler plate as users can just experiment with setting different values for groups.

- What if they want a different learning rate for the evaders?
- What if they want MADDPG for chasers and MAPPO for evaders? (at this point one will have a buffer and the other no)
- What if they want to use a curriculum of evaders to train the chasers?

All these cases benefit from the flexibility of training groups independently and I currently see no benefits in making groups share training components (because the for loops will be just nested inside the loss)

But since this is maybe the most important MARL choice of the lib and kind of the core of my internship let's maybe discuss it more also offline

PR #2048, opened 2024-03-29
[Features] PettingZoo possibility to choose reset strategy

matteobettini:
Here I have a small set of modifications to PettingZoo wrapper that have been requested by users.

The modifications are:
- possibility to override if an environment is done whaen any or all agents are done
- setting the done flags for dead agents properly

cc @renos

### DIFF
--- a/.github/unittest/linux_libs/scripts_pettingzoo/environment.yml
+++ b/.github/unittest/linux_libs/scripts_pettingzoo/environment.yml
@@ -20,4 +20,4 @@ dependencies:
     - expecttest
     - pyyaml
     - autorom[accept-rom-license]
-    - pettingzoo[all]==1.24.1
+    - pettingzoo[all]==1.24.3
--- a/.github/unittest/linux_libs/scripts_pettingzoo/environment.yml
+++ b/.github/unittest/linux_libs/scripts_pettingzoo/environment.yml
@@ -8,7 +8,7 @@ dependencies:
     - cloudpickle
     - gym
     - gym-notices
-    - importlib-metadata
+    - importlib_metadata
     - six
     - zipp
     - pytest
--- a/.github/unittest/linux_libs/scripts_pettingzoo/environment.yml
+++ b/.github/unittest/linux_libs/scripts_pettingzoo/environment.yml
@@ -8,7 +8,6 @@ dependencies:
     - cloudpickle
     - gym
     - gym-notices
-    - importlib_metadata
     - six
     - zipp
     - pytest
--- a/.github/unittest/linux_libs/scripts_smacv2/environment.yml
+++ b/.github/unittest/linux_libs/scripts_smacv2/environment.yml
@@ -7,7 +7,7 @@ dependencies:
     - cloudpickle
     - gym
     - gym-notices
-    - importlib-metadata
+    - importlib_metadata
     - zipp
     - pytest
     - pytest-cov
--- a/.github/unittest/linux_libs/scripts_smacv2/environment.yml
+++ b/.github/unittest/linux_libs/scripts_smacv2/environment.yml
@@ -7,7 +7,6 @@ dependencies:
     - cloudpickle
     - gym
     - gym-notices
-    - importlib_metadata
     - zipp
     - pytest
     - pytest-cov
--- a/.github/unittest/linux_libs/scripts_vmas/environment.yml
+++ b/.github/unittest/linux_libs/scripts_vmas/environment.yml
@@ -7,7 +7,7 @@ dependencies:
     - cloudpickle
     - gym
     - gym-notices
-    - importlib-metadata
+    - importlib_metadata
     - numpy
     - pyglet==1.5.27
     - six
--- a/.github/unittest/linux_libs/scripts_vmas/environment.yml
+++ b/.github/unittest/linux_libs/scripts_vmas/environment.yml
@@ -7,7 +7,6 @@ dependencies:
     - cloudpickle
     - gym
     - gym-notices
-    - importlib_metadata
     - numpy
     - pyglet==1.5.27
     - six
--- a/torchrl/envs/libs/pettingzoo.py
+++ b/torchrl/envs/libs/pettingzoo.py
@@ -90,12 +90,12 @@ class PettingZooWrapper(_EnvWrapper):
     If the number of agents during the task varies, please set ``use_mask=True``.
     ``"mask"`` will be provided
     as an output in each group and should be used to mask out dead agents.
-    The environment will be reset as soon as one agent is done.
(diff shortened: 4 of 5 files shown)


vmoens (commented):
LGTM, I would just add a test for this

matteobettini:
> LGTM, I would just add a test for this

Added!

torchrl/envs/libs/pettingzoo.py
```diff
@@ -156,6 +156,9 @@ class PettingZooWrapper(_EnvWrapper):
         categorical_actions (bool, optional): if the enviornments actions are discrete, whether to transform
             them to categorical or one-hot.
         seed (int, optional): the seed. Defaults to ``None``.
+        done_on_any (bool, optional): whether the environment's done keys are set by aggregating the agent keys
+            using ``any()`` (when True) or ``all()`` (when False). Default (``None``) is to use ``any()`` for
```
vmoens, line 160:
```suggestion
            using ``any()`` (when ``True``) or ``all()`` (when ``False``). Default (``None``) is to use ``any()`` for
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -892,6 +914,9 @@ class PettingZooEnv(PettingZooWrapper):
         categorical_actions (bool, optional): if the enviornments actions are discrete, whether to transform
             them to categorical or one-hot.
         seed (int, optional): the seed.  Defaults to ``None``.
+        done_on_any (bool, optional): whether the environment's done keys are set by aggregating the agent keys
+            using ``any()`` (when True) or ``all()`` (when False). Default (``None``) is to use ``any()`` for
```
vmoens, line 918:
```suggestion
            using ``any()`` (when ``True``) or ``all()`` (when ``False``). Default (``None``) is to use ``any()`` for
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -204,6 +207,7 @@ def __init__(
         use_mask: bool = False,
         categorical_actions: bool = True,
         seed: int | None = None,
+        done_on_any: bool = None,
```
vmoens, line 210:
```suggestion
        done_on_any: bool | None = None,
```

torchrl/envs/libs/pettingzoo.py
```diff
@@ -930,6 +955,7 @@ def __init__(
...
```
vmoens, line 958:
```suggestion
        done_on_any: bool | None = None,
```

vmoens (approved):
Minor edits

.github/unittest/linux_libs/scripts_pettingzoo/environment.yml
```diff
@@ -20,4 +20,4 @@ dependencies:
     - expecttest
     - pyyaml
     - autorom[accept-rom-license]
-    - pettingzoo[all]==1.24.1
+    - pettingzoo[all]==1.24.3
```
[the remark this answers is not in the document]
matteobettini: We are bounding the version in the tests since we do not want future releases of the lib to cause failure in our tests (or to inject code).

However, I would not prevent using the wrapper with future versions of the lib by default as they may be compatible.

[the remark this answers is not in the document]
vmoens: > However, I would not prevent using the wrapper with future versions of the lib by default as they may be compatible

either we guarantee forward compatibility or we don't but either way we need to relax the tests or enforce the version. Otherwise we risk having errors being reported because of bc-breaking changes on their side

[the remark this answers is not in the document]
matteobettini: Relaxing the version will open them to bc-breaking changes on their side and will obfuscate if the source of the failing tests is us or the lib.

Furthermore, I cannot guarantee I will be available to fix any failing CI error due to bc-compatibility on their side in the future.

Preventing to use the wrapper with some version is too constraining. 

My preferred solution is to test a specific version and say that using other versions is allowed but not tested.

Otherwise, I would remove the version in the tests, but I cannot guarantee I will be the one long term updating our side of PettingZoo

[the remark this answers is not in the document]
vmoens: > Furthermore, I cannot guarantee I will be available to fix any failing CI error due to bc-compatibility on their side in the future.

Same as above, if we can't commit to solve upcoming issues we should not allow more recent versions. 
This is like pinning the version of tensordict and torch in the releases, it's good practice. 
If we don't we need to remove the version check in the tests otherwise anyone who does a pip install pettingzoo faces issues when using it with torchrl and we're not helping neither ourselves nor users.

A warning is fine but we should make sure users are made aware of when they're off charts

torchrl/envs/libs/pettingzoo.py
```diff
@@ -265,6 +270,13 @@ def _build_env(
     ):
         import pettingzoo
 
+        if importlib.metadata.version("pettingzoo") > "1.24.3":
```
vmoens, line 273: I would check if the version differs from 1.24.3 (above or below) and adapt the message accordingly

vmoens:
Tests are failing with 
```
  File "/pytorch/rl/torchrl/envs/libs/pettingzoo.py", line 273, in _build_env
    if importlib.metadata.version("pettingzoo") != "1.24.3":
AttributeError: module 'importlib' has no attribute 'metadata'
```

matteobettini:
this is weird, it seems to throw that error when using vec_env

vmoens:
Can't we just use `pettingzoo.__version__` along with `packaging.version.parse`?

.github/unittest/linux_libs/scripts_pettingzoo/environment.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
     - cloudpickle
     - gym
     - gym-notices
-    - importlib-metadata
```
vmoens, line 11: Do we still need this?

.github/unittest/linux_libs/scripts_smacv2/environment.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
     - cloudpickle
     - gym
     - gym-notices
-    - importlib-metadata
```
vmoens, line 10: Ditto

.github/unittest/linux_libs/scripts_vmas/environment.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
     - cloudpickle
     - gym
     - gym-notices
-    - importlib-metadata
```
vmoens, line 10: Ditto

PR #2058, opened 2024-04-06
[Feature] Return depth from RoboHiveEnv

### LINKED ISSUE #2051: [Feature Request] Return depth from RoboHiveEnv
## Motivation

Simulated environments are capable of rendering depth images which can be very useful in downstream training as they enable 3D understanding of the task being performed.  RoboHive environments, simulated with mujoco, can render and return depth images along with RGB. However, this functionality is currently not exposed in the torchrl wrapper `RoboHiveEnv`.

## Solution

A flag enabling depth images to  be returned from `RoboHiveEnv`. The solution has been implemented [here](https://github.com/sriramsk1999/rl/commit/f8bd510741963b305eadd72bbd6b6f6ac6df6a1a)

## Checklist

sriramsk1999:
## Description

This PR adds a flag to enable returning of depth images from `RoboHiveEnv`.

## Motivation and Context

Described in #2051 
`close #2051`

- [X] I have raised an issue to propose this change (required for new features and bug fixes)

## Types of changes

## Checklist

(I don't see any tests specific to `RoboHiveEnv` so I have not added any tests.)

### DIFF
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -3396,10 +3396,11 @@ class TestRoboHive:
     # In the CI, robohive should not coexist with other libs so that's fine.
     # Robohive logging behaviour can be controlled via ROBOHIVE_VERBOSITY=ALL/INFO/(WARN)/ERROR/ONCE/ALWAYS/SILENT
     @pytest.mark.parametrize("from_pixels", [False, True])
+    @pytest.mark.parametrize("from_depths", [False, True])
     @pytest.mark.parametrize("envname", RoboHiveEnv.available_envs)
-    def test_robohive(self, envname, from_pixels):
+    def test_robohive(self, envname, from_pixels, from_depths):
         with set_gym_backend("gymnasium"):
-            torchrl_logger.info(f"{envname}-{from_pixels}")
+            torchrl_logger.info(f"{envname}-{from_pixels}-{from_depths}")
             if any(
                 substr in envname for substr in ("_vr3m", "_vrrl", "_vflat", "_vvc1s")
             ):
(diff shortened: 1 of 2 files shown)


vmoens (commented):
Cool! Can you also add a test in test_libs.py?

sriramsk1999:
While adding the test, I noticed that the current test code was actually broken. I've raised PR #2062 with the changes, but some tests are failing. Will rebase and add new tests after that is merged.

vmoens:
robohive tests raise an error bc the key can't be found, do you want me to investigate @sriramsk1999 ?

sriramsk1999:
I got probably the same error, only for the `rpFrankaRobotiqData` environment. 

```
AssertionError: The keys of the specs and data do not match:
    - List of keys present in real but not in fake: set(),
    - List of keys present in fake but not in real: {('next', 'depths'), 'depths'}.
```

I haven't gotten around to checking out the error in detail yet, please feel free to have a look :) Will update if I find a fix.

sriramsk1999:
All RoboHive tests are now passing:

-  `rpFrankaRobotiqData-v0` is the only environment that has pre-registered visual_keys, so `from_pixels` is overridden to True even if the user sets it to False. I don't think this is expected behaviour, so I've removed this snippet in [this commit](https://github.com/pytorch/rl/pull/2058/commits/7d8c693850654931a168d9f92d246c972fb08332)
- I've added `from_depths` as well to the tests but there are now 2k tests just for RoboHive and it takes a while to complete (20 min on my PC)
- Unit tests were failing because the environment was already registered. So I've enabled overriding the already registered env [here](https://github.com/pytorch/rl/pull/2058/commits/9c394120a9cc07737fe02ff059d2ab1d7adf8f98#diff-ccbe2efde180bfb5fdb4bf24c919c7712ef48af50f1989bbaa3705e8c8b07074L240)

vmoens (approved):
LGTM

PR #2071, opened 2024-04-09
[Feature] Extend TensorDictPrimer default_value options

albertbou92:
## Description

This PR aims to extend the possible values taken by the tensors added by the TensorDictPrimer transform, allowing to use callable to create them.

## Motivation and Context

## Types of changes

## Checklist


torchrl/envs/transforms/transforms.py
```diff
@@ -4505,8 +4512,23 @@ def __init__(
         if not isinstance(kwargs, CompositeSpec):
             kwargs = CompositeSpec(kwargs)
         self.primers = kwargs
+        if (random is not None) and isinstance(default_value, (dict, Callable)):
```
vmoens, line 4515: We could set the default to None and just check if it's not None

torchrl/envs/transforms/transforms.py
```diff
@@ -4505,8 +4512,23 @@ def __init__(
...
+            raise ValueError(
+                "Setting random to True and providing a default_value are incompatible."
+            )
         self.random = random
+        if isinstance(default_value, dict):
+            if len(default_value) != len(self.primers) and set(dict.keys()) != set(
```
vmoens, line 4521: Before that we can call unravel_key on all the keys of the duct

torchrl/envs/transforms/transforms.py
```diff
@@ -4505,8 +4512,23 @@ def __init__(
...
+                self.primers.keys(True, True)
+            ):
+                raise ValueError(
+                    "If a default_value dictionary is provided, it must match the primers keys."
+                )
+            default_value = {
+                key: default_value[key] for key in self.primers.keys(True, True)
+            }
         self.default_value = default_value
```
vmoens, line 4530: I think we should always turn default_value to a dict

torchrl/envs/transforms/transforms.py
```diff
@@ -4568,15 +4593,20 @@ def transform_observation_spec(
...
         for key, spec in self.primers.items():
             if spec.shape[: len(observation_spec.shape)] != observation_spec.shape:
-                raise RuntimeError(
-                    f"The leading shape of the primer specs ({self.__class__}) should match the one of the parent env. "
-                    f"Got observation_spec.shape={observation_spec.shape} but the '{key}' entry's shape is {spec.shape}."
-                )
+                try:
+                    expanded_spec = self._try_expand_shape(spec)
+                except AttributeError:
+                    raise RuntimeError(
```
vmoens, line 4599: When will this be reached?

torchrl/envs/transforms/transforms.py
```diff
@@ -4589,8 +4619,25 @@ def transform_input_spec(self, input_spec: TensorSpec) -> TensorSpec:
     def _batch_size(self):
         return self.parent.batch_size
 
+    def _validate_value_tensor(self, value, spec):
+        if value.shape != spec.shape:
```
vmoens, line 4623: You could just do spec.is_in(value)

torchrl/envs/transforms/transforms.py
```diff
@@ -4589,8 +4619,25 @@ def transform_input_spec(self, input_spec: TensorSpec) -> TensorSpec:
...
+            )
+        if value.dtype != spec.dtype:
+            raise RuntimeError(
+                f"Value dtype ({value.dtype}) does not match the spec dtype ({spec.dtype})."
+            )
+        if value.device != spec.device:
+            raise RuntimeError(
+                f"Value device ({value.device}) does not match the spec device ({spec.device})."
+            )
+        if not spec.is_in(value):
```
vmoens, line 4635: Here it is, I don't think this can be false after the previous checks but it does the same job

torchrl/envs/transforms/transforms.py
```diff
@@ -4601,11 +4648,22 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
             if self.random:
                 value = spec.rand()
             else:
-                value = torch.full_like(
-                    spec.zero(),
-                    self.default_value,
-                )
+                if isinstance(self.default_value, dict):
```
vmoens, line 4651: Can be simplified if we make it a dict in all cases

torchrl/envs/transforms/transforms.py
```diff
@@ -4601,11 +4648,22 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+                    value = self.default_value[key]
+                else:
+                    value = self.default_value
+                if callable(value):
+                    value = value()
+                    if not self._validated:
+                        self._validate_value_tensor(value, spec)
+                else:
+                    value = torch.full_like(
```
vmoens, line 4660: Could be full(spec.shape, value), which is faster

torchrl/envs/transforms/transforms.py
```diff
@@ -4634,22 +4692,39 @@ def _reset(
...
         return tensordict_reset
 
     def __repr__(self) -> str:
         class_name = self.__class__.__name__
-        return f"{class_name}(primers={self.primers}, default_value={self.default_value}, random={self.random})"
+        default_value = (
+            self.default_value
+            if isinstance(self.default_value, float)
+            else self.default_value.__class__.__name__
+        )
```
vmoens, line 4726: Maybe always a dict like {key: val}
And if val is a callable we just write `Callable` instead

vmoens (commented):
Great work, left a bunch of comments.
Thanks a million!

torchrl/envs/transforms/transforms.py
albertbou92: if for any reason parent is None

vmoens: when would `transform_observation_spec` be called when `parent` is `None`?

torchrl/envs/transforms/transforms.py
```diff
@@ -4505,8 +4512,30 @@ def __init__(
...
+            raise ValueError(
+                "Setting random to True and providing a default_value are incompatible."
+            )
+        default_value = (
+            default_value or 0.0
+        )  # if not random and no default value, use 0.0
         self.random = random
+        if isinstance(default_value, dict):
+            primer_keys = {unravel_key(key) for key in self.primers.keys(True, True)}
+            default_value_keys = {unravel_key(key) for key in default_value.keys()}
```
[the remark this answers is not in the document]
albertbou92: Ah very cool option @vmoens !

atm I need to transform it back to dict, because I get the following behaviour for non-tensor data:

```
import torch
default_value = {
    "mykey1": lambda: torch.ones(3),
    "mykey2": lambda: torch.tensor(1, dtype=torch.int64),
}
default_value = TensorDict(default_value, [])
keys = default_value.keys(True, True)
print(keys)
```

output:
```
_TensorDictKeysView([],
    include_nested=True,
    leaves_only=True)

```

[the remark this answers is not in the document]
vmoens: Are you using the latest nightly?
You can always define your own is_leaf for keys:
```python
default_value.keys(True, True, is_leaf=lambda x: isinstance(x, (torch.Tensor, NonTensorData)))
```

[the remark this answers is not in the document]
albertbou92: I was using the latest commit, yes. But this solution works fine :)

vmoens (approved):
LGTM!

vmoens:
TestgSDE is failing because we patched the behaviour for wrong primers, can you fix that?

albertbou92:
done!

vmoens (approved):
LGTM

PR #2073, opened 2024-04-09
[Feature] Auto-resetting envs

vmoens:
Allows the creation of auto-resetting envs through `env = SomeEnvClass(..., auto_reset=True)`

This should be doable with all envs without requiring users to put auto_reset as an arg in the constructor.

cc @luisenp


torchrl/envs/batched_envs.py
```diff
@@ -503,6 +503,15 @@ def _get_in_keys_to_exclude(self, tensordict):
...
 
+    @property
+    def auto_reset(self) -> bool:
+        return False
+
+    @auto_reset.setter
+    def auto_reset(self, value) -> None:
+        if value not in (False, None):
+            raise ValueError("auto_reset must be None of False")
+
```
vmoens, line 514:
```suggestion
```

torchrl/envs/common.py
```diff
@@ -2909,7 +2927,12 @@ def fake_tensordict(self) -> TensorDictBase:
 
         fake_state = state_spec.zero()
         fake_action = action_spec.zero()
-        fake_input = fake_state.update(fake_action)
+        if any(
+            isinstance(spec, LazyStackedTensorDict) for spec in fake_action.values(True)
```
vmoens, line 2931:
```suggestion
            isinstance(val, LazyStackedTensorDict) for val in fake_action.values(True)
```

torchrl/envs/libs/gym.py
```diff
@@ -759,6 +759,14 @@ def _is_batched(self):
...
         )
 
+    @property
+    def auto_reset(self) -> bool:
+        return self._is_batched or self._auto_reset
+
+    @auto_reset.setter
+    def auto_reset(self, value: bool) -> None:
+        self._auto_reset = value
+
```
vmoens, line 769:
```suggestion
```

PR #2099, opened 2024-04-23
[Feature] A PixelRenderTransform

vmoens:
Adds a transform to record rgb_arrays coming from env.render()

cc @matteobettini @vikashplus

@Miffyli: I also added a bit of doc RE recording videos in torchrl, LMK if you see anything missing.


Miffyli (commented):
Thanks for adding this stuff! I don't dare to give LGTM, being still relatively fresh to TorchRL, but just left some generic comments :)

PR #2126, opened 2024-04-26
[Doc] Indicate necessary context to run multiprocessed collectors in doc

GJBoth:
This PR updates the docstrings for the `MultiSyncCollector` and `MultiaSyncCollector. It does several things:
- Add a note that code should be run inside a `if __name__ == "__main__":` block.
- Updated the examples and fixes some errors, including:
  - The gymenv already has a stepcounter, so removed it.
  - `devices` and `storing_devices` kwargs don't exist - it needs to be `device` and `storing_device`.
  - Added `cat_results="stack"` to follow future behaviour.
- The collector wasn't actually imported in the example, so I added that too - it's now self-contained.
- The current `main` branch also does not pass pre-commit, which is why `tensor_specs.py` is changed too.

I'm currently building the docs to see if everything looks good, but it's taking *very* long. Any advice there?

### DIFF
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -535,7 +535,10 @@ class SyncDataCollector(DataCollectorBase):
         self.reset_when_done = reset_when_done
         self.n_env = self.env.batch_size.numel()
 
-        (self.policy, self.get_weights_fn,) = self._get_policy_and_device(
+        (
+            self.policy,
+            self.get_weights_fn,
+        ) = self._get_policy_and_device(
             policy=policy,
             observation_spec=self.env.observation_spec,
         )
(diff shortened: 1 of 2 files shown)


GJBoth:
@vmoens Could you help out here with the failed tests? Some of them failed due to main failing, and I didn't even touch code - just added docs -, so not sure why the CI on benchmarks is failing?

vmoens:
Yeah those are broken tests on CUDA, I will handle that separately :)

torchrl/collectors/collectors.py
```diff
@@ -1884,28 +1887,35 @@ class MultiSyncDataCollector(_MultiDataCollector):
     trajectory and the start of the next collection.
     This class can be safely used with online RL sota-implementations.
 
+    note:: Python requires multiprocessed code to be instantiated within a
+    `if __name__ == "__main__":` block. See https://docs.python.org/3/library/multiprocessing.html
+    for more info.
```
vmoens, line 1892: This doesn't render well
image


```suggestion
    .. note:: Python requires multiprocessed code to be instantiated within a main guard:

            >>> if __name__ == "__main__":
            ...     # Create your collector here

        See https://docs.python.org/3/library/multiprocessing.html for more info.

```

torchrl/collectors/collectors.py
```diff
@@ -1884,28 +1887,35 @@ class MultiSyncDataCollector(_MultiDataCollector):
...
+            ...     init_random_frames=-1,
+            ...     reset_at_each_iter=False,
+            ...     device="cpu",
+            ...     storing_device="cpu",
+            ...     cat_results="stack",
+            ... )
+        >>>     for i, data in enumerate(collector):
+        ...         if i == 2:
+        ...             print(data)
+        ...             break
```
vmoens, line 1918:
```suggestion
        >>> from torchrl.collectors import MultiSyncDataCollector
        >>> if __name__ == "__main__":
        ...     env_maker = lambda: GymEnv("Pendulum-v1", device="cpu")
        ...     policy = TensorDictModule(nn.Linear(3, 1), in_keys=["observation"], out_keys=["action"])
        ...     collector = MultiSyncDataCollector(
        ...         create_env_fn=[env_maker, env_maker],
        ...         policy=policy,
        ...         total_frames=2000,
        ...         max_frames_per_traj=50,
        ...         frames_per_batch=200,
        ...         init_random_frames=-1,
        ...         reset_at_each_iter=False,
        ...         device="cpu",
        ...         storing_device="cpu",
        ...         cat_results="stack",
        ...     )
        ...     for i, data in enumerate(collector):
        ...         if i == 2:
        ...             print(data)
        ...             break
```

torchrl/collectors/collectors.py
```diff
@@ -2232,27 +2241,34 @@ class MultiaSyncDataCollector(_MultiDataCollector):
...
```
vmoens, line 2271: See comments above

vmoens (commented):
Thanks! A couple of formatting issues but we should be good to go after that!

GJBoth:
Just resolved the issues!

vmoens (approved):
LGTM

PR #2143, opened 2024-04-30
[Feature] Dynamic specs

vmoens:
Implements specs with negative shape dimension, allowing these dimensions to account for tensors of variable sizes.

This PR provides basic functionality for specs, environments, batched envs (serial and parallel) and collectors (Sync, MultiSync, MultiaSync).

Performance without buffers in multiprocessed settings isn't great. Funnily, using "cuda" devices to pass data without shared mem buffers is faster than serializing and deserializing tensors (@albanD @mikaylagawarecki please don't make fun of me):
```python
if __name__ == "__main__":
    # this executes a step on cpu on the worker, puts the data on cuda, sends the data through the pipe, and maps it back to cpu afterwards 
    env = ParallelEnv(16, lambda: GymEnv("CartPole-v1", device="cuda"), use_buffers=False, device="cpu")
    timer = torch.utils.benchmark.Timer("env.rollout(200, break_when_any_done=False, return_contiguous=False)")
    print(timer.adaptive_autorange())
```

I also played a bit in rewriting the reducing methods to tensordict to see if there was any opportunity to make the unpickling faster through some hacky multithreading, without success.

cc @dtsaras @matteobettini


dtsaras:
I will start testing this today and provide you with some feedback!

PR #2159, opened 2024-05-12
[Docs] Reintroduce BenchMARL pointers in MARL tutos

matteobettini:
I think this might be helpful to the reader.

I am not fixed on any phrasing so we can iterate on it

### DIFF
--- a/tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
+++ b/tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
@@ -3,6 +3,12 @@ Competitive Multi-Agent Reinforcement Learning (DDPG) with TorchRL Tutorial
 ===========================================================================
 **Author**: `Matteo Bettini <https://github.com/matteobettini>`_
 
+.. note::
+   If you are interested in benchmarking Multi-Agent Reinforcement Learning (MARL) in
+   TorchRL, there exists an external facebookresearch project
+   `BenchMARL <https://github.com/facebookresearch/BenchMARL>`__ that can help you
+   train and compare MARL sota-implementations, tasks, and models using TorchRL!
+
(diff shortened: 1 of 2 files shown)


tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -3,6 +3,12 @@
 ===========================================================================
 **Author**: `Matteo Bettini <https://github.com/matteobettini>`_
 
+.. note::
+   If you are interested in benchmarking Multi-Agent Reinforcement Learning (MARL) in
+   TorchRL, there exists an external facebookresearch project
+   `BenchMARL <https://github.com/facebookresearch/BenchMARL>`__ that can help you
+   train and compare MARL sota-implementations, tasks, and models using TorchRL!
```
vmoens, line 10: What do we want from these headers exactly? When I read this it feels promotional and not a valuable info for anyone who wants to get proficient with the MARL API, which is the goal of the tuto (ie. it's unclear how this can be a next step for users who are reading this tuto).

I'd stick to a very concise, simple sentence like
```suggestion
   The `BenchMARL <https://github.com/facebookresearch/BenchMARL>`__ library provides state-of-the-art implementations of MARL algorithms.
```

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -3,6 +3,12 @@
 ===========================================================================
 **Author**: `Matteo Bettini <https://github.com/matteobettini>`_
 
+.. note::
```
vmoens, line 6:
```suggestion
.. seealso::
```

matteobettini:
Got it.

I have integrated the suggestions, just added "using TorchRL" at the end

PR #2178, opened 2024-05-29
[BugFix] buffer `__iter__` for samplers without replacement + prefetch

JulianKu:
## Description

Adds loop termination condition in replay buffer `__iter__` method for yielding samples in prefetch queue and another condition in its `sample` method for only adding new jobs to the prefetch queue if the sampler has not yet run out.

## Motivation and Context

For replay buffers with prefetch enabled while having a sampler without replacement (that can "run out"), the buffer's `__iter__` method will not yield all samples. This is because its while loop terminates on sampler `ran_out` which is set as soon as a prefetch job is executed that triggers running out. However, there might still be prefetch jobs in the queue left that could yield samples, therefore their samples are simply skipped.

## Types of changes

## Checklist

### DIFF
--- a/torchrl/data/replay_buffers/replay_buffers.py
+++ b/torchrl/data/replay_buffers/replay_buffers.py
@@ -633,7 +633,9 @@ class ReplayBuffer:
             ret = self._sample(batch_size)
         else:
             with self._futures_lock:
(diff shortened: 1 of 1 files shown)


vmoens (approved):
LGTM thanks for the fix!

PR #2181, opened 2024-05-30
[BugFix] Fix "_reset" assignment in batched envs

vmoens:
Solves a bug reported [here](https://discuss.pytorch.org/t/pettingzoo-trouble-running-multiple-marl-environments-in-parallel/203706/4?u=matteobettini)

```python
from torchrl.envs import PettingZooEnv, ParallelEnv, SerialEnv

if __name__ == "__main__":

    def base_env_fn():
        return PettingZooEnv(
            task="multiwalker_v9",
            parallel=True,
            seed=42,
            n_walkers=3,
            shared_reward=False,
            max_cycles=1000,
            render_mode=None,
            device="cpu",
        )

        lambda: ParallelEnv(
            num_workers=4,  # noqa: E731
            create_env_fn=base_env_fn,
            device="cpu",
        ),
        policy=None,  # the explora
        frames_per_batch=100,
        max_frames_per_traj=50,
        total_frames=200,
        device="cpu",
        reset_at_each_iter=False,
    )
    try:
            print("Reached")
    finally:
        collector.shutdown()
```

### DIFF
--- a/docs/source/reference/envs.rst
+++ b/docs/source/reference/envs.rst
@@ -248,6 +248,29 @@ designing reset functionalities:
   components, in which case the non-reset data will be meaningless and should
   be discarded.
 
+.. warning:: The ``"_reset"`` entries may have a different shape than the ``done``
+  states, but they should usually always have the shape of their parent tensordict unless
+  in specific limited cases.
+  The reason for this different is the following:
+  For single agent, single task scripts, the ``"done"`` value is usually configured with
+  a trailing singleton dimension such that these code snipped can work without unwanted
+  broadcasting:
+
+    >>> obs = data["next", "observation"] # shape [B, O]
+    >>> reward = data["next", "reward"] # shape [B, 1]
+    >>> done = data["next", "done"] # shape [B, 1]
+    >>> next_value = pred(obs) # shape [B, 1]
+    >>> pred_value = reward + (1 - done.int()) * next_value # shape [B, 1]
+
+  This code snippet would require multiple unsqueeze operations by the user to avoid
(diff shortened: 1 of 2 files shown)


matteobettini:
I do not think the bug is in the changed code. 

"_reset" in torchrl has always the same shape as "done" (https://github.com/pytorch/rl/blob/d59b8105cd7437cccb7c480f15ce6caf67f587c6/torchrl/envs/common.py#L2784). It keeps the singleton when there is the singleton.

You can check this in any env. In vmas for example, "_reset" has shape `(n_envs,1)`, just like done.

The strategy in `metadata_from_env` was correct I think. The bug regarding the different `_reset` shapes must be stemmng from elswhere

matteobettini:
The bug stems from here https://github.com/pytorch/rl/blob/d59b8105cd7437cccb7c480f15ce6caf67f587c6/torchrl/envs/common.py#L2779-L2798

Where if the done is simple you create the `_reset` with the right shape (the one of the done) and if the done is complex you create a `_reset` at root with no matching done key.

Therefore, this new `_reset` is not present in the fake_td as no corresponding key for it was created in `metadata_from_env`

## The fix

`metadata_from_env` strategy was good. aka currently torchrl creates a `_reset` key for each done key with the same shape

the thing that `metadata_from_env` is missing is the creation of a further `_reset` key at root when the done is not simple

### Further suggestion

It might be worth considering to align the if branches of the script linked in this comment to have the trailing dimension added to the `_reset` by `_terminated_or_truncated ()`

vmoens:
Sounds better than my solution :) 
Do you want to make a PR with this?

PR #2185, opened 2024-05-31
[BugFix] buffer __iter__ for samplers without replacement + prefetch

JulianKu:
Reopen #2178 (see #2182)

## Description

Adds loop termination condition in replay buffer `__iter__` method for yielding samples in prefetch queue and another condition in its `sample` method for only adding new jobs to the prefetch queue if the sampler has not yet run out.

## Motivation and Context

For replay buffers with prefetch enabled while having a sampler without replacement (that can "run out"), the buffer's `__iter__` method will not yield all samples. This is because its while loop terminates on sampler `ran_out` which is set as soon as a prefetch job is executed that triggers running out. However, there might still be prefetch jobs in the queue left that could yield samples, therefore their samples are simply skipped.

## Types of changes

## Checklist

### DIFF
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -1923,6 +1923,25 @@ class TestSamplers:
         s = new_replay_buffer.sample(batch_size=1)
         assert (s.exclude("index") == 0).all()
 
+    def test_sampler_without_replacement_cap_prefetch(self):
+        torch.manual_seed(0)
+        data = TensorDict({"a": torch.arange(10)}, batch_size=[10])
+        rb = ReplayBuffer(
+            storage=LazyTensorStorage(10),
+            sampler=SamplerWithoutReplacement(),
+            batch_size=2,
+            prefetch=3,
+        )
+        rb.extend(data)
+
+        for _ in range(100):
+            s = set()
+            for i, d in enumerate(rb):
+                assert i <= 4
+                s = s.union(set(d["a"].tolist()))
+            assert i == 4
+            assert s == set(range(10))
+
     @pytest.mark.parametrize(
         "batch_size,num_slices,slice_len,prioritized",
         [
--- a/torchrl/data/replay_buffers/replay_buffers.py
+++ b/torchrl/data/replay_buffers/replay_buffers.py
@@ -718,9 +718,7 @@ class ReplayBuffer:
                 "Cannot iterate over the replay buffer. "
                 "Batch_size was not specified during construction of the replay buffer."
             )
-        while not self._sampler.ran_out or (
-            self._prefetch and len(self._prefetch_queue)
-        ):
+        while not self._sampler.ran_out:
             yield self.sample()
 
     def __getstate__(self) -> Dict[str, Any]:
--- a/torchrl/data/replay_buffers/replay_buffers.py
+++ b/torchrl/data/replay_buffers/replay_buffers.py
@@ -634,7 +634,8 @@ class ReplayBuffer:
         else:
             with self._futures_lock:
                 while (
-                    len(self._prefetch_queue) < self._prefetch_cap
+                    len(self._prefetch_queue)
+                    < min(self._sampler._remaining_batches, self._prefetch_cap)
                 ) and not self._sampler.ran_out:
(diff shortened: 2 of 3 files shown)


JulianKu:
Somehow, for me the example you provided seems to run fine (at least no errors apparent).
I set-up a simple Github codespace for that... 

```shell-session
@JulianKu ➜ /workspaces/torchrl (fix/iter-replacement-prefetch) $ python sota-implementations/iql/iql_offline.py optim.gradient_steps=55 logger.backend=

/workspaces/torchrl/sota-implementations/iql/iql_offline.py:35: UserWarning: 
The version_base parameter is not specified.
Please specify a compatability version level, or None.
Will assume defaults for version 1.1
  @hydra.main(config_path="", config_name="offline_config")
/usr/local/python/3.10.13/lib/python3.10/site-packages/hydra/_internal/hydra.py:119: UserWarning: Future Hydra versions will no longer change working directory at job runtime by default.
See https://hydra.cc/docs/1.2/upgrades/1.1_to_1.2/changes_to_job_working_dir/ for more information.
  ret = run_job(
/workspaces/torchrl/sota-implementations/iql/iql_offline.py:35: UserWarning: 
The version_base parameter is not specified.
Please specify a compatability version level, or None.
Will assume defaults for version 1.1
  @hydra.main(config_path="", config_name="offline_config")
/workspaces/torchrl/sota-implementations/iql/iql_offline.py:35: UserWarning: 
The version_base parameter is not specified.
Please specify a compatability version level, or None.
Will assume defaults for version 1.1
  @hydra.main(config_path="", config_name="offline_config")
/workspaces/torchrl/sota-implementations/iql/iql_offline.py:35: UserWarning: 
The version_base parameter is not specified.
Please specify a compatability version level, or None.
Will assume defaults for version 1.1
  @hydra.main(config_path="", config_name="offline_config")
/workspaces/torchrl/sota-implementations/iql/iql_offline.py:35: UserWarning: 
The version_base parameter is not specified.
Please specify a compatability version level, or None.
Will assume defaults for version 1.1
  @hydra.main(config_path="", config_name="offline_config")
/workspaces/torchrl/sota-implementations/iql/iql_offline.py:35: UserWarning: 
The version_base parameter is not specified.
Please specify a compatability version level, or None.
Will assume defaults for version 1.1
  @hydra.main(config_path="", config_name="offline_config")
2024-05-31 08:05:06,428 [torchrl][INFO] Downloading dataset: http://rail.eecs.berkeley.edu/datasets/offline_rl/gym_mujoco_v2/halfcheetah_medium-v2.hdf5 to /tmp/tmp9rza7qdi/halfcheetah_medium-v2.hdf5
/home/codespace/.local/lib/python3.10/site-packages/torch/nn/modules/lazy.py:181: UserWarning: Lazy modules are a new feature under heavy development so changes to the API or functionality can happen at any moment.
  warnings.warn('Lazy modules are a new feature under heavy development '
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 55/55 [00:02<00:00, 18.35it/s]
```

vmoens:
I still get the error.

Besides, I think that this won't really solve it because `self._sampler.ran_out` will only be `True` when the sample is gathered, which in the case of prefetch>0 will only occur once the sample is accessible (so we still can have samples coming in after the sampler has ran out).

Making a step back, what we want to avoid is that `_sample` is reached when the sampler has ran out, not because it raises an exception but because it will just reset the `ran_out` attribute and restart sampling.
So, what I would suggest is this: 
- We create an `auto_reset` attribute in the sampler that is checked whenever a sample is drawn. If `auto_reset=False` and `ran_out=True`, the sampler returns a special value (e.g. `StopIteration`)
- Whenever we get a StopIteration we tell the `__iter__` to stop
- Next time we call `__iter__` we reset the sampler

I gave an attempt [here](https://github.com/pytorch/rl/tree/fix/iter-replacement-prefetch-bis), we still have the issue that there can be run conditions between the workers so we need to guard the accessed to ran_out and auto_reset with a lock I think.

vmoens:
I believe this PR now solves this problem

PR #2188, opened 2024-05-31
[BugFix] `LazyTensorStorage` only allocates data on the given device

### LINKED ISSUE #2187: [BUG] `LazyTensorStorage` allocates storage on wrong device
https://github.com/pytorch/rl/blob/d59b8105cd7437cccb7c480f15ce6caf67f587c6/torchrl/data/replay_buffers/storages.py#L870-L873

As can be seen in the snippet, the storage first allocates everything on the data device and only later moves it to the proper device.

This is problematic as, if i request

```python
return TensorDictReplayBuffer(
            storage=LazyTensorStorage(
                100_000_000, # Say this is 30GB
                device="cpu",
            ),
            sampler=sampler,
            batch_size=100,
        )
```

it allocates the 30GB first on the data device (gpu) and then later moves it to cpu

matteobettini:
Fixes #2187

### DIFF
--- a/torchrl/data/replay_buffers/storages.py
+++ b/torchrl/data/replay_buffers/storages.py
@@ -867,10 +867,10 @@ class LazyTensorStorage(TensorStorage):
             return (self.max_size, *data_shape)
 
         if is_tensor_collection(data):
-            out = data.expand(max_size_along_dim0(data.shape))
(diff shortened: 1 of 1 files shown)


vmoens (approved):
LGTM thanks

Issue #2186, opened 2024-05-31
[BUG] Incorrect Calculation of Mode for TanhNormal Distribution


Emile-Aquila:
## Describe the bug

I'd like to express my gratitude for the swift and diligent maintenance of the torchrl library. 

I have identified a potential issue with the implementation of the mode method in the TanhNormal distribution within TorchRL.
The calculation of the mode appears to be incorrect due to the nature of the tanh function applied to a normal distribution.

- current implementation
  - https://github.com/pytorch/rl/blob/main/torchrl/modules/distributions/continuous.py#L405

I think the mode of the TanhNormal distribution should accurately reflect the peak of the probability density function after applying the tanh transformation to the underlying normal distribution. Given the non-linearity of the tanh function, the mode calculation should account for this complexity.

<br>

## To Reproduce

The current implementation of the mode method does not correctly compute the mode, resulting in inaccurate values.
For example, in the following scenario, the mode is expected to be around 1, but the method returns approximately 0.197.

```python
import torch
from torchrl.modules import TanhNormal
import matplotlib.pyplot as plt

torch.random.manual_seed(0)

loc = torch.tensor([0.2], dtype=torch.float32)
scale = torch.tensor([1.0], dtype=torch.float32)

dist = TanhNormal(loc, scale, min=-1, max=1)
print("mode: ", dist.mode.item())  # mode:  0.1973753273487091

sample = dist.sample_n(10000)
plt.hist(sample.numpy(), bins=500, range=(-1, 1))
plt.show()

```

![image](https://github.com/pytorch/rl/assets/35189007/12c52b23-64b1-4f49-a29e-200e87bc672d)

<br>

<br>

Thank you again for your continuous support and hard work on maintaining the torchrl library.

vmoens:
Hello
Thanks for reporting this
#2198 should fix it and improve the API.
Note that properly finding the mode of the distribution requires to find its maximum (well, the mode is the maximum haha) but there is not analytical expression. For the mean I implemented it using a regular stochastic expectation, but for the maximum i had to rely on Newton-Raphson which is considerably slower than what we had before, and non-differentiale.

Basically, before it was fast, now it's accurate

![image](https://github.com/pytorch/rl/assets/25529882/003febe8-6ec1-40a8-a4fb-6095199964f2)

Emile-Aquila:
Hello 

Thank you for your prompt response and for fixing the implementation error in the mode of TanhNormal distribution calculations. 
[...]

PR #2198, opened 2024-06-03
[BugFix] Fix tanh normal mode

vmoens:
Solves #2186, please post comments about the feature in that discussion.

Related PR: https://github.com/pytorch/tensordict/pull/824

cc @matteobettini

### DIFF
--- a/torchrl/modules/distributions/continuous.py
+++ b/torchrl/modules/distributions/continuous.py
@@ -420,7 +420,8 @@ class TanhNormal(FasterTransformedDistribution):
         self.low = low
         self.high = high
 
-        t = SafeTanhTransform()
+        # t = SafeTanhTransform()
+        t = D.TanhTransform()
         if self.non_trivial_max or self.non_trivial_min:
             t = D.ComposeTransform(
                 [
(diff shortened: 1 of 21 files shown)


torchrl/modules/distributions/continuous.py
```diff
@@ -392,22 +426,71 @@ def update(self, loc: torch.Tensor, scale: torch.Tensor) -> None:
...
+                break
+        else:
+            torchrl_logger.info(
+                f"Failed to find the mode of the TanhNormal distribution. Max step={abs(step).max()}"
+            )
         return m
 
+    @property
+    def mean(self):
+        return self.rsample((1000,)).mean(0)
```
matteobettini, line 492: This is a bit arbitrary no? Isn't it better to not provide this feature and tell users that they can compute their empirical estimate themselves?

torchrl/modules/distributions/continuous.py
```diff
@@ -322,36 +324,58 @@ class TanhNormal(FasterTransformedDistribution):
...
+
     def __init__(
         self,
         loc: torch.Tensor,
         scale: torch.Tensor,
         upscale: Union[torch.Tensor, Number] = 5.0,
-        min: Union[torch.Tensor, Number] = -1.0,
-        max: Union[torch.Tensor, Number] = 1.0,
-        event_dims: int = 1,
+        low: Union[torch.Tensor, Number] = -1.0,
```
matteobettini, line 339: If we change this we should change it in all bounded distributions. I would do it in a separate PR

PR #2202, opened 2024-06-05
[BugFix] Fix strict length in PRB+SliceSampler

vmoens:
I wrote dedicated tests under ``test_slice_sampler_prioritized``

TODO:
- [x] Check caching
- [x] Add left-right span options
- [ ] ~Aggregate/reduce priorities~


vmoens:
@wertyuilife2

Should we reduce the priorities of each traj while we're at it?
I don't think it'd require much compute and it would make sure that all items are equally weighted within a traj

Take the following 2 trajs with associated priorities

Item:      [0, 1, 2, 3, 4, 5, 6, 7]
Traj:       [0, 0, 0, 1, 1, 1, 1, 1]
Priority: [10, 1, 1, 10, 1, 2, 1, 1]

Currently, item 0 and 3 having a higher priority they have more chances of being sampled as start points and hence you will get more trajs starting with these. If we reduce, we will have (10 + 1 + 1)/3=4 for the first and (10 + 1 + 2 + 1 +1)/5=3 for the second.

Priority: [4, 4, 4, 3, 3, 3, 3, 3]

At this point, the start point is equally likely within a traj but some trajs have a higher prob of being sampled (which seems to make more sense to me?)

I guess any solution will make someone unhappy...

wertyuilife2:
@vmoens I believe that when discussing `PrioritizedSampler`, there is only **one correct approach**: we should **not** reduce the priorities of each trajectory while we are at it. 

The core idea of PER is that certain **samples** (not trajectories) are important (such as a critical action) and need to be learned frequently. Reducing the priorities of each trajectory would make it difficult for PER to focus on updating specific important samples.

When discussing `PrioritizedSliceSampler`, we face the choice of whether to reduce the priorities of each **slice** while we are at it. My suggestion is to leave this choice to the user, as the calculation of priorities and the calling of update_priority() are both handled by the user. In other words, we still should not reduce the priorities of each slice.

vmoens:
> The core idea of PER is that certain samples (not trajectories) are important (such as a critical action) and need to be learned frequently. Reducing the priorities of each trajectory would make it difficult for PER to focus on updating specific important samples.

Got it thanks for that, indeed that's how I edited the docstring (users should be in charge of setting the proper priority).
But when we say prioritized, slice sampler I can imagine someone imagining: I have a transition with high priority therefore there is a chance to find it anywhere (not just at the beginning) in my sample -- whereas now there is a higher chance to find it at the beginning of a slice than at the end

torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -1086,6 +1087,15 @@ def add(self, data: TensorDictBase) -> int:
...
         if index is not None:
+            # if self._storage.ndim > 1:
+            #     if isinstance(index, torch.Tensor) and index.ndim == 2:
+            #         index = index.unbind(-1)
+            #         index = (index[-1], *index[:-1])
+            #         index = torch.stack(list(index), dim=-1)
+            #     elif isinstance(index, tuple):
+            #         index = (index[-1], *index[:-1])
+            #         index = torch.stack(list(index), dim=-1)
+
```
vmoens, line 1098:
```suggestion
```

torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -1105,8 +1115,22 @@ def extend(self, tensordicts: TensorDictBase) -> torch.Tensor:
...
+        #  so the actualy indices will have to be used on the _storage directly (not on the buffer)
+        # if self._storage.ndim > 1:
+        #     if isinstance(index, torch.Tensor) and index.ndim == 2:
+        #         index = index.unbind(-1)
+        #         index = (index[-1], *index[:-1])
+        #         index = torch.stack(list(index), dim=-1)
+        #     elif isinstance(index, tuple):
+        #         index = (index[-1], *index[:-1])
+        #         index = torch.stack(list(index), dim=-1)
+
```
vmoens, line 1130:
```suggestion
```

torchrl/data/replay_buffers/samplers.py
```diff
@@ -414,58 +417,57 @@ def sample(self, storage: Storage, batch_size: int) -> torch.Tensor:
...
+        #     #   w_i = (p_i / sum(p) * N) ^ (-beta)
+        #     #   weight_i = w_i / max(w)
+        #     #   weight_i = (p_i / sum(p) * N) ^ (-beta) /
+        #     #       ((min(p) / sum(p) * N) ^ (-beta))
+        #     #   weight_i = ((p_i / sum(p) * N) / (min(p) / sum(p) * N)) ^ (-beta)
+        #     #   weight_i = (p_i / min(p)) ^ (-beta)
+        #     # weight = np.power(weight / (p_min + self._eps), -self._beta)
+        #     weight = torch.pow(weight / p_min, -self._beta)
+        #     if storage.ndim > 1:
+        #         index = torch.unravel_index(index, storage._storage.shape)
```
vmoens, line 454: @wertyuilife2 Happy to hear your thoughts about this

Basically the idea would be to sample uniformly if a priority hasn't been passed. I won't make it part of this PR but we could consider this in the future.

torchrl/data/replay_buffers/samplers.py
```diff
@@ -1731,6 +1762,33 @@ def sample(self, storage: Storage, batch_size: int) -> Tuple[torch.Tensor, dict]
...
+            diff[diff < 0] = diff.max() + 1
+            # Take the arg min along dim 0 (thereby reducing dim M)
+            idx = diff.argmin(dim=0)
+            stops = stop_idx[idx, 0]
+            # TODO: here things may not work bc we could have spanning trajs,
+            #  though I cannot show that it breaks in the tests
+            if starts_tensor.ndim > 1:
+                starts_tensor = starts_tensor[:, 0]
+            seq_length = (stops - starts_tensor + 1).clamp_max(seq_length)
+            assert (seq_length > 0).all()
```
vmoens, line 1790:
```suggestion
            if (seq_length <= 0).any():
                raise RuntimeError("failed to compute seq_length, please report this bug")
```

torchrl/data/replay_buffers/samplers.py
wertyuilife2: @vmoens  I believe we should ensure that **priorities are initialized at all times** and throw an error when `p_sum <= 0` or `p_min <= 0`. We should design around `default_priority` to handle potential issues rather than ignoring them.

Even considering buffer silence bug detection, I don't recommend such an implementation. 

For example, if some user's incorrect behavior or a bug causes `len(storage)` to be larger than it should be, this implementation could result in undetected silent bugs.

vmoens, line 454:
```suggestion
```

PR #2215, opened 2024-06-07
[BugFix] Fix max-priority update

vmoens:


vmoens:
@wertyuilife2 can you confirm that this makes sense?

wertyuilife2:
@vmoens I did some further search, found that many open-source libraries implementing PER maintain the historical maximum priority, such as [dopamine](https://github.com/google/dopamine/blob/master/dopamine/replay_memory/prioritized_replay_buffer.py). 

But the original PER paper and some other source codes in my domain use the buffer's maximum priority approach, 
including [EfficientZero](https://github.com/YeWR/EfficientZero/blob/main/core/replay_buffer.py), which is the SOTA data-efficient method in RL.

So, overall, both approaches make sense, and this is not a bug(my bad), I believe the buffer's maximum priority approach is more robust to the priority value.

wertyuilife2:
I am raising the issue because I found in practice that during the early stages of training, when a transition is first time being sampled, its PER weight is typically `1e-8` (which is the value of epsilon) which is weird. This is because `max_priority=1` and the bellman error I get in the early training stage is 0 (cause network output are init to 0 for stablility).

So, basically, this is more like an additional feature for better priority adaptation or similar to priority normalization. I think it is not an essential feature (if you find implementing it to be complicated), but it can be available as an option.

vmoens:
So should I make the erasing of the max_priority during extend optional?

wertyuilife2:
yep, I think so. Maybe in some task the historical max will be better.

PR #2217, opened 2024-06-08
[BugFix] Global import of optional library

matteobettini:
Can we also have a test pipeline of torchrl that does not install any optional dependency?


vmoens (approved):
LGTM

PR #2221, opened 2024-06-10
[Performance, Refactor] Faster loading of uninitialized storages

vmoens:
cc @teopir 

cc @shagunsodhani this is a good example of prealloc with tensordict. We were using a lot of lazy stacks and stacking at the last minute. Using a preallocated TD instead (create an empty td -> get a bunch of views of that td -> write on the first view, and all views get instantiated instantaneously) made the whole thing 20 - 1000x faster!

### DIFF
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -93,6 +93,7 @@ from torchrl.envs.transforms.transforms import (
     RewardClipping,
     RewardScaling,
     SqueezeTransform,
+    StepCounter,
     ToTensorImage,
     UnsqueezeTransform,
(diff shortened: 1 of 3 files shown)


shagunsodhani:
> cc @teopir
> 
> cc @shagunsodhani this is a good example of prealloc with tensordict. We were using a lot of lazy stacks and stacking at the last minute. Using a preallocated TD instead (create an empty td -> get a bunch of views of that td -> write on the first view, and all views get instantiated instantaneously) made the whole thing 20 - 1000x faster!

Awesome <3

PR #2239, opened 2024-06-20
[BugFix] Fix OOB sampling in PrioritizedSliceSampler

### LINKED ISSUE #2230: [BUG] Unexpected behavior of SumSegmentTree Resulting in Invalid Slices in PrioritizedSliceSampler.sample()
In `PrioritizedSampler.sample()`, `_sum_tree.scan_lower_bound()` sometimes generates index greater than len(storage).

The test code is as follows:

```python
import torch
from torchrl._torchrl import SumSegmentTreeFp32
import numpy as np

def test_sum_tree():
    torch.manual_seed(0)
    np.random.seed(0)
    sum_tree = SumSegmentTreeFp32(500)
    
    # repeat to ensure the bug happens
    for _ in range(1000):
        # update priority
        index = torch.arange(0,100, dtype=torch.long, device=torch.device("cpu"))
        priority = torch.rand(100, device=torch.device("cpu"))
        sum_tree[index] = priority+1e-8 # 1e-8 are not essential, w/o 1e-8, bug still happens
        
        # sample
        p_sum = sum_tree.query(0, 100)
        mass = np.random.uniform(0.0, p_sum, size=1000000) # sample a lot to ensure the bug happens.
        scaned_index = sum_tree.scan_lower_bound(mass)
        if scaned_index.max()>=100:
            print("Unexpected index! p_sum:{}, mass.max():{}, scaned_index.max():{}".format(p_sum, mass.max(),scaned_index.max()))
        
test_sum_tree()

```

In `PrioritizedSampler.sample()`, this unexpected behavior is handled by `index.clamp_max_(len(storage) - 1)`, but it still causes unexpected behavior in `PrioritizedSliceSampler.sample()`.

This occurs because, although the sample at `index = len(storage) - 1` is covered by `preceding_stop_idx`, it can still be sampled as start index, leading to cross-trajectory sampling.

## Checklist

vmoens:
Closes #2230

Unfortunately, I'm not sure how to test this. I guess we should save somewhere a tree structure and mass to replicate the issue (and possibly fix it directly in the c++ code @xiaomengy if you can help with this)

### DIFF
--- a/torchrl/data/replay_buffers/samplers.py
+++ b/torchrl/data/replay_buffers/samplers.py
@@ -475,6 +475,15 @@ class PrioritizedSampler(Sampler):
             index = index.unsqueeze(0)
         index.clamp_max_(len(storage) - 1)
         weight = torch.as_tensor(self._sum_tree[index])
+        # get indices where weight is 0
+        zero_weight = weight == 0
+        index = index
+        while zero_weight.any():
+            index = torch.where(zero_weight, index - 1, index)
+            if (index < 0).any():
+                raise RuntimeError("Failed to find a suitable index")
+            zero_weight = torch.as_tensor(self._sum_tree[index])
+            zero_weight = weight == 0
 
         # Importance sampling weight formula:
         #   w_i = (p_i / sum(p) * N) ^ (-beta)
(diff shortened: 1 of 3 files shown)


torchrl/data/replay_buffers/samplers.py
```diff
@@ -475,6 +475,15 @@ def sample(self, storage: Storage, batch_size: int) -> torch.Tensor:
             index = index.unsqueeze(0)
         index.clamp_max_(len(storage) - 1)
         weight = torch.as_tensor(self._sum_tree[index])
```
wertyuilife2, line 477: @vmoens it seems the `weight` here is not updated, I think we should do `weight = torch.as_tensor(self._sum_tree[index])` after changing the `index`?

Issue #2269, opened 2024-07-05
[Feature Request] Support for distributional-DQNalgorithms (C51, Rainbow)


roger-creus:
## Motivation
Is the Distributional Q-Value Actor currently fully-supported? If so, are there any plans to integrate C51 and more importantly, Rainbow, to the list of sota-implementations?

vmoens:
We have a version of this here
https://pytorch.org/rl/stable/reference/generated/torchrl.objectives.DistributionalDQNLoss.html#torchrl.objectives.DistributionalDQNLoss
but I don't think we have an official version of Rainbow yet (although this is the first thing we had in the lib - for some reason we never made a script that was high-quality enough to be made public!)
LMK if you need further help with it!

roger-creus:
I have implemented a first version of Rainbow containing all tricks! (Dueling DQN, Distributional, Prioritized Experience, etc.) and I am now running some preliminary experiments to debug its performance and make sure it works well.

However, I had to change [this line](https://github.com/pytorch/rl/blob/55f0a52ac7025c459dd71ec180c9c7d1c7576f9c/torchrl/objectives/dqn.py#L563C18-L563C81) to  `Tz = reward + (1 - terminated.to(reward.dtype)) * discount.unsqueeze(-1) * support.repeat(batch_size, 1)`.

Otherwise I would get shape errors. Let me know if this makes sense!

PR #2270, opened 2024-07-05
[BugFix] Fixed shape for MultiStep returns + Distributional loss

roger-creus:
## Description

I encountered an error when combining MultiStep returns + Distributional DQN Loss. Because [this line](https://github.com/pytorch/rl/blob/55f0a52ac7025c459dd71ec180c9c7d1c7576f9c/torchrl/objectives/dqn.py#L563) assumes discount is an integer but if using MultiStep returns it is a tensor of size `(batch_size,)`.

This PR only changes that line to:

`Tz = reward + (1 - terminated.to(reward.dtype)) * discount.unsqueeze(-1) * support.repeat(batch_size, 1)`

which gives the same results if the discount is an integer but fixes the shape error if it is a batch.

Solves issue #2269

## Types of changes

## Checklist


torchrl/objectives/dqn.py
```diff
@@ -560,7 +560,7 @@ def forward(self, input_tensordict: TensorDictBase) -> TensorDict:
             support = support.to("cpu")
             pns_a = pns_a.to("cpu")
 
-            Tz = reward + (1 - terminated.to(reward.dtype)) * discount * support
+            Tz = reward + (1 - terminated.to(reward.dtype)) * discount.unsqueeze(-1) * support.repeat(batch_size, 1)
```
vmoens, line 563: Looks good
I think we're making a couple of assumptions about discount and support here.
I would feel safer if we had a block before this where we check explicitly the shapes:
```python
if not check(discount.shape):
    raise RuntimeError(f"Expected `discount` to have either shape X or Y, got {discount.shape}")
if not other_check(support.shape):
    raise RuntimeError(f"Expected `support` to have either shape X or Y, got {support.shape}")
```
(`check` and `other_check` are just placeholders)
If possible I would also make an instance of this loss class with the support and discount shape that you are using to test that it works ok (see `test_cost.py:TestDQN:test_distributional_dqn`)

vmoens: We should also make sure that discount is a tensor:
```python
discount = torch.as_tensor(discount)
```
should do it (it's a no-op if it's already a tensor)

vmoens (commented):
Thanks for this contribution! I left a couple of comments, would you have time to address them?

PR #2283, opened 2024-07-09
[Feature] TensorDictMap

vmoens:
This code is copy-pasted from https://github.com/pytorch/tensordict/pull/826

Since we have the opportunity of building composite storages or storages of any type, I would be keen to remove the key_to_storage dict and just have one storage per map (a storage can have anything in it)

Credits to @mjlaali

Missing:
- [ ] Tests
- [ ] Docs
- [ ] contains is broken currently
- [ ] Dynamic storage with a default value was pretty cool, we should have something similar


torchrl/data/map/hash.py
```diff
@@ -0,0 +1,191 @@
...
+        binary_features = (
+            torch.heaviside(features, self.zero_tensor)
+            if self.convert_to_binary
+            else features
+        )
+        feature_parts = binary_features.reshape(shape=(-1, self.num_bits))
+        digits = torch.vmap(torch.dot, (None, 0))(
+            self.bases, feature_parts.to(self.bases.dtype)
+        )
+        digits = digits.reshape(shape=(-1, features.shape[-1] // self.num_bits))
```
mjlaali, line 78: nit: features.shape[-1] -> num_features

torchrl/data/map/tdstorage.py
```diff
@@ -0,0 +1,278 @@
...
+from tensordict import NestedKey, TensorDictBase
+from tensordict.nn.common import TensorDictModuleBase
+from torchrl.data.map import QueryModule, RandomProjectionHash
+from torchrl.data.replay_buffers.storages import _get_default_collate, LazyTensorStorage
+
+K = TypeVar("K")
+V = TypeVar("V")
+
+
+class TensorMap(abc.ABC, Generic[K, V]):
```
mjlaali, line 21: nit: shall we update the name of file to align with the class name?

mjlaali:
Pretty neat. I was wondering how to update the tree. In particular, can we store a count per child and update it after each rollout?

test/test_storage_map.py
```diff
@@ -0,0 +1,243 @@
...
+    "gym", None
+)
+
+
+class TestHash:
+    def test_binary_to_decimal(self):
+        binary_to_decimal = BinaryToDecimal(
+            num_bits=4, device="cpu", dtype=torch.int32, convert_to_binary=True
+        )
+        binary = torch.Tensor([[0, 0, 1, 0, 0, 0, 0, 1], [0, 0, 0, 0, 0, 0, 10, 0]])
```
dtsaras, line 34: I think the 10 should be a 1 in this case?

vmoens: I think it's on purpose, to test that we convert non binary to binary (see the class constructor)

vmoens:
Closing in favour of https://github.com/pytorch/rl/pull/2304

PR #2287, opened 2024-07-10
[Tests] Fix VMAS tests

matteobettini:


vmoens (approved):
Is this bc-breaking? Since we had tests that worked and they don't, should the users be informed of anything?

matteobettini:
> Is this bc-breaking? Since we had tests that worked and they don't, should the users be informed of anything?

nono, not bc-breaking, the tests had some narrow assumpions that have been extended.

PR #2294, opened 2024-07-18
[BugFix] `EnvBase._complete_done` to complete "terminated" key properly

### LINKED ISSUE #2291: [BUG] `_complete_done` always sets missing terminated to `False`
## Describe the bug

For a custom environment that only outputs "done" keys in `_step`, as in the [example](https://pytorch.org/rl/stable/tutorials/pendulum.html#coding-the-effect-of-an-action-step), the automatically added "terminated" key will always be False across all dimensions and not mirror the done-tensor.

This happens regardless of whether the user defines the done spec or it automatically added in `EnvBase. _create_done_specs`.

## To Reproduce

I tried to keep the example to a minimum.
The important part is in `_step` where we add the "done" key as `torch.tensor([True, False)]`.
```python
from typing import Optional

import torch
from tensordict import TensorDictBase, TensorDict
from torchrl.data import (
    CompositeSpec,
    BinaryDiscreteTensorSpec,
    OneHotDiscreteTensorSpec,
)
from torchrl.envs import EnvBase

class CustomEnv(EnvBase):

    def __init__(
        self,
        *,
        device=None,
        batch_size: Optional[torch.Size] = torch.Size([2]),
        run_type_checks: bool = False,
        allow_done_after_reset: bool = False,
    ):
        assert batch_size == (2,)  # hardcoded for minimal example
        super().__init__(
            device=device,
            batch_size=batch_size,
            run_type_checks=run_type_checks,
            allow_done_after_reset=allow_done_after_reset,
        )

        self.observation_spec = CompositeSpec(
            observation=UnboundedContinuousTensorSpec(
                shape=torch.Size(batch_size + (1,))
            ),
            shape=batch_size,
        )
        self.action_spec = OneHotDiscreteTensorSpec(n=2, shape=batch_size)
        self.reward_spec: BinaryDiscreteTensorSpec = BinaryDiscreteTensorSpec(
            n=1, dtype=torch.int8, shape=torch.Size([2, 1])
        )

        done = torch.tensor([True, False], dtype=torch.bool)
        next_observation = torch.randn(self.observation_spec["observation"].shape)
        return TensorDict(
            {"observation": next_observation, "done": done, "reward": torch.ones((2,))},
            batch_size=(2,),
        )

    def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
        return TensorDict(
            {"observation": torch.randn(self.observation_spec["observation"].shape)},
            batch_size=(2,),
        )

    def _set_seed(self, seed: Optional[int]):
        pass



  env = CustomEnv()
  td = env.reset()
  env.rand_action(td)
  env.step(td)
  assert env.done_keys == ["done", "terminated"]
  assert torch.equal(td[("next", "done")],torch.tensor([[True], [False]]))
  assert torch.equal(td[("next", "terminated")], torch.tensor([[True], [False]]))
```

```bash
terminated_env_bug.py:62 (test_)
def test_():
        env = CustomEnv()
        td = env.reset()
        env.rand_action(td)
        env.step(td)
        assert env.done_keys == ["done", "terminated"]
        assert torch.equal(td[("next", "done")], torch.tensor([[True], [False]]))
>       assert torch.equal(td[("next", "terminated")], torch.tensor([[True], [False]]))
E       assert False
E        +  where False = <built-in method equal of type object at 0x7fabbdc64800>(tensor([[False],\n        [False]]), tensor([[ True],\n        [False]]))
E        +    where <built-in method equal of type object at 0x7fabbdc64800> = torch.equal
E        +    and   tensor([[ True],\n        [False]]) = <built-in method tensor of type object at 0x7fabbdc64800>([[True], [False]])
E        +      where <built-in method tensor of type object at 0x7fabbdc64800> = torch.tensor
```

## Expected behavior

The ("next", "terminated") entry is equal to the ("next", "done") entry as documented for `EnvBase._complete_done`.

## System info

Using torchrl-nightly installed with pip. But should also apply for the main-branch as relevant code is the same.

```python
import torchrl, numpy, sys
2024.7.3 2.0.0 3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0] linux

```

## Reason and Possible fixes

The problem seems to come from [`EnvBase._complete_done` ](https://github.com/pytorch/rl/blob/f764c0253c88a9faeecfa406c0bc29a5031e70a4/torchrl/envs/common.py#L1496).
The`for key, item in done_spec.items(False, True):`  loop in Line 1509 iterates both over the "done" and the "terminated" key, however only for "done" a value is present in `data`.
For the "done" key (happening first) `data.set("terminated", val)` is set correctly to the values of `data["done"]` in line 1537.
But then for the "terminated" key the `elif val is None:` is triggered and `data["terminated"]` is overridden again.

```python
...
for key, item in done_spec.items(False, True):  # goes over done and terminated (order is important)
                val = vals.get(key, None)  # will be [[True], [False]] for "done" but None for "terminated"
                if (
                    key == "done"
                    and val is not None
                    and "terminated" in done_spec_keys
                    and "terminated" not in data_keys
                ):
                    if "truncated" in data_keys:
                        raise RuntimeError(
                            "Cannot infer the value of terminated when only done and truncated are present."
                        )
                    data.set("terminated", val)
                elif (
                    key == "terminated"
                    and val is not None
                    and "done" in done_spec_keys
                    and "done" not in data_keys
                ):
                    if "truncated" in data_keys:
                        done = val | data.get("truncated")
                        data.set("done", done)
                    else:
                        data.set("done", val)
                elif val is None:
                    # we must keep this here: we only want to fill with 0s if we're sure
                    # done should not be copied to terminated or terminated to done
                    # in this case, just fill with 0s
                    data.set(key, item.zero(leading_dim))  # overrides the "terminated" key with False
        return data
```

cc @kurtamohler

kurtamohler:
## Description

Fix the logic in `EnvBase._complete_done` to properly complete the "terminated" key.

## Motivation and Context

close #2291

## Types of changes

## Checklist

### DIFF
--- a/test/test_env.py
+++ b/test/test_env.py
@@ -311,6 +311,62 @@ def test_rollout_predictability(device):
     ).all()
 
 
+# Check that the "terminated" key is filled in automatically if only the "done"
+# key is provided in `_step`.
+def test_done_key_completion_done():
+    class DoneEnv(CountingEnv):
+        def _step(
+            self,
+            tensordict: TensorDictBase,
+        ) -> TensorDictBase:
+            self.count += 1
+            tensordict = TensorDict(
+                source={
+                    "observation": self.count.clone(),
+                    "done": self.count > self.max_steps,
+                    "reward": torch.zeros_like(self.count, dtype=torch.float),
+                },
+                batch_size=self.batch_size,
+                device=self.device,
+            )
+            return tensordict
+
+    env = DoneEnv(max_steps=torch.tensor([[0], [1]]), batch_size=(2,))
+    td = env.reset()
+    env.rand_action(td)
+    td = env.step(td)
+    assert torch.equal(td[("next", "done")], torch.tensor([[True], [False]]))
+    assert torch.equal(td[("next", "terminated")], torch.tensor([[True], [False]]))
+
+
+# Check that the "done" key is filled in automatically if only the "terminated"
+# key is provided in `_step`.
+def test_done_key_completion_terminated():
+    class TerminatedEnv(CountingEnv):
+        def _step(
+            self,
+            tensordict: TensorDictBase,
+        ) -> TensorDictBase:
+            self.count += 1
+            tensordict = TensorDict(
+                source={
+                    "observation": self.count.clone(),
+                    "terminated": self.count > self.max_steps,
+                    "reward": torch.zeros_like(self.count, dtype=torch.float),
+                },
+                batch_size=self.batch_size,
+                device=self.device,
+            )
+            return tensordict
+
+    env = TerminatedEnv(max_steps=torch.tensor([[0], [1]]), batch_size=(2,))
+    td = env.reset()
+    env.rand_action(td)
+    td = env.step(td)
+    assert torch.equal(td[("next", "done")], torch.tensor([[True], [False]]))
+    assert torch.equal(td[("next", "terminated")], torch.tensor([[True], [False]]))
+
+
 @pytest.mark.skipif(not _has_gym, reason="no gym")
 @pytest.mark.parametrize("env_name", [PENDULUM_VERSIONED])
 @pytest.mark.parametrize("frame_skip", [1])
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM thanks for this!

Issue #2295, opened 2024-07-18
Convert `TensorDictModuleWrapper`s to `TensorDictModule`s


kurtamohler:
## Motivation

[`EGreedyWrapper`](https://github.com/pytorch/rl/blob/c4b2eb0e0856da812fb9de13170e96fa185ec42b/torchrl/modules/tensordict_module/exploration.py#L184) was converted to [`EGreedyModule`](https://github.com/pytorch/rl/blob/c4b2eb0e0856da812fb9de13170e96fa185ec42b/torchrl/modules/tensordict_module/exploration.py#L31) a while ago because `TensorDictModule`s are more user-friendly than `TensorDictModuleWrapper`s.

There are still some `TensorDictModuleWrapper`s that have not been converted to a `TensorDictModule`:

* [`AdditiveGuassianWrapper`](https://github.com/pytorch/rl/blob/c4b2eb0e0856da812fb9de13170e96fa185ec42b/torchrl/modules/tensordict_module/exploration.py#L252)
* [`OrnsteinUhlenbeckProcessWrapper`](https://github.com/pytorch/rl/blob/c4b2eb0e0856da812fb9de13170e96fa185ec42b/torchrl/modules/tensordict_module/exploration.py#L385)

## Solution

We should convert these remaining `TensorDictModuleWrapper`s to `TensorDictModule`s.

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

PR #2297, opened 2024-07-19
[Feature] Add `modules.OrnsteinUhlenbeckProcessModule`

kurtamohler:
## Description

Add `OrnsteinUhlenbeckProcessModule`, which serves the same purpose as `OrnsteinUhlenbeckProcessWrapper`, but derives from `TensorDictModuleBase` rather than `TensorDictModuleWrapper`

## Motivation and Context

part of #2295

## Types of changes

## Checklist

### DIFF
--- a/docs/source/reference/modules.rst
+++ b/docs/source/reference/modules.rst
@@ -75,6 +75,7 @@ other cases, the action written in the tensordict is simply the network output.
     AdditiveGaussianWrapper
     EGreedyModule
     EGreedyWrapper
+    OrnsteinUhlenbeckProcessModule
     OrnsteinUhlenbeckProcessWrapper
 
 Probabilistic actors
--- a/test/test_exploration.py
+++ b/test/test_exploration.py
@@ -47,6 +47,7 @@ from torchrl.modules.tensordict_module.exploration import (
     AdditiveGaussianWrapper,
     EGreedyModule,
     EGreedyWrapper,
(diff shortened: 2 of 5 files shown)


vmoens (approved):
LGTM
nit: Since we want to deprecate the OU wrapper, can you add a deprecation warning? 2 releases from 0.5, so it will be in 0.7!
Thanks!

docs/source/reference/modules.rst
```diff
@@ -75,6 +75,7 @@ other cases, the action written in the tensordict is simply the network output.
     AdditiveGaussianWrapper
     EGreedyModule
     EGreedyWrapper
+    OrnsteinUhlenbeckProcessModule
     OrnsteinUhlenbeckProcessWrapper
```
vmoens, line 79:
```suggestion
    OrnsteinUhlenbeckProcessModule
```
since we want to deprecate it

vmoens: hmm upon reflection, let's keep this one, people may still see it in their code

kurtamohler:
I've added the deprecation warning

docs/source/reference/modules.rst
kurtamohler: Yeah I'd think it's typically a good idea to keep the old documentation and only remove it at the point when the deprecated feature is actually removed

PR #2330, opened 2024-07-29
[Docs] InitTracker cleanup

### LINKED ISSUE #2326: [BUG] `InitTracker` key flexibility
Two questions about the following code 

https://github.com/pytorch/rl/blob/0063741839a3e5e1a527947945494d54f91bc629/torchrl/envs/transforms/transforms.py#L6437-L6440

- is there a reason why we are preventing the `init_key` to be a nested key
- is the setting of `self.reset_key` done properly/ever used? it seems not

matteobettini:
fixes #2326


vmoens (approved):
LGTM thanks

Issue #2327, opened 2024-07-28
[BUG] Issues with `TensorDictPrimer`


matteobettini:
## Without the primer, the collector does not feed any hidden state to the policy

in the [RNN tutorial ](https://github.com/pytorch/rl/blob/main/tutorials/sphinx-tutorials/dqn_with_rnn.py) it is stated that the primer is optional and it is used just to store the hidden states in the buffer.

This is not true in practice. Not adding the primer will result in the collector not feeding the hidden states to the policy during execution. Which will silently cause the rnn to loose any recurrency.

To reproduce, comment out this line 

https://github.com/pytorch/rl/blob/0063741839a3e5e1a527947945494d54f91bc629/tutorials/sphinx-tutorials/dqn_with_rnn.py#L269

and print the policy input at this line

https://github.com/pytorch/rl/blob/0063741839a3e5e1a527947945494d54f91bc629/torchrl/collectors/collectors.py#L733

you will see that no hidden state is fed to the rnn during execution and no  errors or warnings are thrown

## The primer overwrites any nested spec

Consider an env with nested specs

```python
 env = VmasEnv(
        scenario="balance,
        num_envs=5,
    )
```

add to it a primer for a nested hidden state

```python
    env = TransformedEnv(
        env,
        TensorDictPrimer(
            {
                "agents": CompositeSpec(
                    {
                        "h": UnboundedContinuousTensorSpec(
                            shape=(*env.shape, env.n_agents, 2, 128)
                        )
                    },
                    shape=(*env.shape, env.n_agents),
                )
            }
        ),
    )
```

the primer code in https://github.com/pytorch/rl/blob/0063741839a3e5e1a527947945494d54f91bc629/torchrl/envs/transforms/transforms.py#L4649 will overwirite the observation spec instead of updating it, resulting in the loss of all the spec keys that previoulsy were in the "agents" spec

The same result is obtained with

```python
    env = TransformedEnv(
        env,
        TensorDictPrimer(
            {
               ("agents","h"): UnboundedContinuousTensorSpec(
                            shape=(*env.shape, env.n_agents, 2, 128)
                )
            }
        ),
    )
```

here, updating the spec instead of overwriting it should do the job

## The order of the primer in the transforms seems to have an impact

In the same vmas environemnt as above, if i put the primer and then the reward sum

```python
 env = TransformedEnv(
        env,
        Compose(
            TensorDictPrimer(
                {
                    "agents": CompositeSpec(
                        {
[...]

PR #2332, opened 2024-07-29
[BugFix] TensorDictPrimer updates spec instead of overwriting

### LINKED ISSUE #2331: [BUG] `TensorDictPrimer` overwrites nested specs in the environemnt
part of #2327

## The primer overwrites any nested spec

Consider an env with nested specs

```python
 env = VmasEnv(
        scenario="balance,
        num_envs=5,
    )
```

add to it a primer for a nested hidden state

```python
    env = TransformedEnv(
        env,
        TensorDictPrimer(
            {
                "agents": CompositeSpec(
                    {
                        "h": UnboundedContinuousTensorSpec(
                            shape=(*env.shape, env.n_agents, 2, 128)
                        )
                    },
                    shape=(*env.shape, env.n_agents),
                )
            }
        ),
    )
```

the primer code in https://github.com/pytorch/rl/blob/0063741839a3e5e1a527947945494d54f91bc629/torchrl/envs/transforms/transforms.py#L4649 will overwirite the observation spec instead of updating it, resulting in the loss of all the spec keys that previoulsy were in the "agents" spec

The same result is obtained with

```python
    env = TransformedEnv(
        env,
        TensorDictPrimer(
            {
               ("agents","h"): UnboundedContinuousTensorSpec(
                            shape=(*env.shape, env.n_agents, 2, 128)
                )
            }
        ),
    )
```

here, updating the spec instead of overwriting it should do the job

matteobettini:
fixes #2331

it also solves the last issue in #2327


torchrl/envs/transforms/transforms.py
```diff
@@ -4638,15 +4638,20 @@ def transform_observation_spec(
             raise ValueError(
                 f"observation_spec was expected to be of type CompositeSpec. Got {type(observation_spec)} instead."
             )
-        for key, spec in self.primers.items():
-            if spec.shape[: len(observation_spec.shape)] != observation_spec.shape:
-                expanded_spec = self._expand_shape(spec)
-                spec = expanded_spec
+        try:
+            device = observation_spec.device
```
vmoens, line 4642: when is observation_spec.device returning an `RuntimeError`?
IIRC this should not happen anymore

vmoens (commented):
Thanks for this, can you add a small test in `TestTensorDictPrimer`?
Just create a TDPrimer and an observation spec with nested specs, and call `transform_observation_spec` to check that the entries are updated accordingly.

matteobettini:
I added the test

vmoens (approved):
LGTM

Issue #2133, opened 2024-04-29
[Feature Request] Integration for OpenSpiel environments


Brunozml:
## Motivation
It would be great to be able to use OpenSpiel's environments with TorchRL. 

## Solution

Its `rl_environment` interface is basically identical to that of gym, so the integration should relatively straightforward. I'm more than willing to help achieve this.

## Alternatives

 I'm not sure if it's already possible to do this using `GymLikeEnv`, but so far my experimentation has failed. 

## Checklist

- [X] I have checked that there is no similar issue in the repo (**required**)

vmoens:
That would be wonderful!

Never tried it to be completely open about that.

Looking at it the API seems quite different from the one from gym.

IIUC, the steps to achieve this would be:
- map actions to tensors, map states to tensors
- find a generic way to generate specs, check that actions and states mapped above are within the specs, and that specs generate valid states and action at random
- I guess the env will be stateless so we will need a copy of the state specs in `state_spec` and `observation_spec`

Concretely, I'd love to see a dummy rollout where we get the state (represented as a single array or multiple arrays or tensors) and the action (idem).

If the games have anything remotely related to "specs" i'd like to see them too!

kurtamohler:
I'm looking into the possibility of making the OpenSpiel wrapper stateless. I'm not sure yet if it's possible. I've submitted an issue in OpenSpiel asking about it: https://github.com/google-deepmind/open_spiel/issues/1254

kurtamohler:
I would like to reopen this, since I still have to add support for chance nodes

PR #2345, opened 2024-08-01
[Feature] Add `OpenSpielWrapper` and `OpenSpielEnv`

kurtamohler:
## Description

Adds environment wrapper classes for OpenSpiel.

`OpenSpielWrapper.reset` supports resetting to a specified state.

## Motivation and Context

Part of #2133

- [x] Someone has raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/libs/openspiel.py
+++ b/torchrl/envs/libs/openspiel.py
@@ -1,0 +1,655 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import annotations
+
+import importlib.util
+from typing import Dict, List
+
+import torch
+from tensordict import TensorDict, TensorDictBase
+
+from torchrl.data.tensor_specs import (
+    Categorical,
+    Composite,
+    NonTensor,
+    OneHot,
+    Unbounded,
+)
+from torchrl.envs.common import _EnvWrapper
+from torchrl.envs.utils import _classproperty, check_marl_grouping, MarlGroupMapType
+
+_has_pyspiel = importlib.util.find_spec("pyspiel") is not None
+
+
+def _get_envs():
+    if not _has_pyspiel:
+        raise ImportError(
+            "open_spiel not found. Consider downloading and installing "
+            f"open_spiel from {OpenSpielWrapper.git_url}."
+        )
+
+    import pyspiel
+
+    return [game.short_name for game in pyspiel.registered_games()]
+
+
+class OpenSpielWrapper(_EnvWrapper):
+    """Google DeepMind OpenSpiel environment wrapper.
+
+    GitHub: https://github.com/google-deepmind/open_spiel
+
+    Documentation: https://openspiel.readthedocs.io/en/latest/index.html
+
+    Args:
+        env (pyspiel.State): the game to wrap.
+
+    Keyword Args:
+        device (torch.device, optional): if provided, the device on which the data
+            is to be cast. Defaults to ``None``.
+        batch_size (torch.Size, optional): the batch size of the environment.
+            Defaults to ``torch.Size([])``.
+        allow_done_after_reset (bool, optional): if ``True``, it is tolerated
+            for envs to be ``done`` just after :meth:`~.reset` is called.
+            Defaults to ``False``.
+        group_map (MarlGroupMapType or Dict[str, List[str]]], optional): how to
+            group agents in tensordicts for input/output. See
+            :class:`~torchrl.envs.utils.MarlGroupMapType` for more info.
+            Defaults to
+            :class:`~torchrl.envs.utils.MarlGroupMapType.ALL_IN_ONE_GROUP`.
+        categorical_actions (bool, optional): if ``True``, categorical specs
+            will be converted to the TorchRL equivalent
+            (:class:`torchrl.data.Categorical`), otherwise a one-hot encoding
+            will be used (:class:`torchrl.data.OneHot`).  Defaults to ``False``.
+        return_state (bool, optional): if ``True``, "state" is included in the
+            output of :meth:`~.reset` and :meth:`~step`. The state can be given
+            to :meth:`~.reset` to reset to that state, rather than resetting to
+            the initial state.
+            Defaults to ``False``.
+
+    Attributes:
+        available_envs: environments available to build
+
+    Examples:
+        >>> import pyspiel
+        >>> from torchrl.envs import OpenSpielWrapper
+        >>> from tensordict import TensorDict
+        >>> base_env = pyspiel.load_game('chess').new_initial_state()
+        >>> env = OpenSpielWrapper(base_env, return_state=True)
+        >>> td = env.reset()
+        >>> td = env.step(env.full_action_spec.rand())
+        >>> print(td)
+        TensorDict(
+            fields={
+                agents: TensorDict(
+                    fields={
+                        action: Tensor(shape=torch.Size([2, 4672]), device=cpu, dtype=torch.int64, is_shared=False)},
+                    batch_size=torch.Size([]),
+                    device=None,
+                    is_shared=False),
+                next: TensorDict(
+                    fields={
+                        agents: TensorDict(
+                            fields={
+                                observation: Tensor(shape=torch.Size([2, 20, 8, 8]), device=cpu, dtype=torch.float32, is_shared=False),
+                                reward: Tensor(shape=torch.Size([2, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
+                            batch_size=torch.Size([2]),
+                            device=None,
+                            is_shared=False),
+                        current_player: Tensor(shape=torch.Size([]), device=cpu, dtype=torch.int32, is_shared=False),
+                        done: Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.bool, is_shared=False),
+                        state: NonTensorData(data=FEN: rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1
+                        3009
+                        , batch_size=torch.Size([]), device=None),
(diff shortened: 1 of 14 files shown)


kurtamohler:
A few notes:

`categorical_action_encoding=False` is not supported yet.

Also, some of the games in OpenSpiel do not work properly in `OpenSpielWrapper` because the action spec assumes a discrete space of `pyspiel.Game.num_distinct_actions()` (see [OpenSpiel API reference](https://github.com/google-deepmind/open_spiel/blob/82c742c6403aefeebb104736387c19d709fa7255/docs/api_reference.md)). However, for some of the games, `pyspiel.State.legal_actions()` can return more actions than `pyspiel.Game.num_distinct_actions()`. I suppose to support those games we need to allow the action spec's size to change?

torchrl/envs/libs/openspiel.py
```diff
@@ -0,0 +1,358 @@
+import importlib.util
```
vmoens, line 1: missing headers

torchrl/envs/libs/openspiel.py
```diff
@@ -0,0 +1,358 @@
...
+from typing import Dict, Optional
```
vmoens, line 2: I try to upgrade to the new `foo: Smth | SmthElse` type annotation, can we remove the `Optional` and use that instead (make sure you have a `from __future__ import annotations` for older python versions)

torchrl/envs/libs/openspiel.py
```diff
@@ -0,0 +1,358 @@
...
+    Args:
+        env (pyspiel.State): the game to wrap.
+        categorical_action_encoding (bool, optional): if ``True``, categorical
+            specs will be converted to the TorchRL equivalent (:class:`torchrl.data.DiscreteTensorSpec`),
+            otherwise a one-hot encoding will be used (:class:`torchrl.data.OneHotTensorSpec`).
+            Defaults to ``False``.
+
+    Keyword Args:
+        device (torch.device, optional): if provided, the device on which the data
+            is to be cast. Defaults to ``torch.device("cpu")``.
```
vmoens, line 46: device should be `None` by default

torchrl/envs/libs/openspiel.py
```diff
@@ -0,0 +1,358 @@
...
+                    observation_shape
+                )
+                for idx in range(self._env.num_players())
+            ]
+        )
+        reward = self._to_tensor(self._env.returns())
+        done = torch.tensor(
+            self._env.is_terminal(), device=self.device, dtype=torch.bool
+        )
+        state = self._env.serialize()
```
vmoens, line 206: let's put a flag around that to make the state optional. Maybe "stateless=True"?

torchrl/envs/libs/openspiel.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def _update_action_mask(self):
+        action_mask = torch.zeros(
+            self._env.num_distinct_actions(), device=self.device, dtype=torch.bool
+        )
+        action_mask[self._env.legal_actions()] = True
+        self.action_spec.update_mask(action_mask)
+
+    def _make_td_out(self, exclude_reward=False):
+        observation_shape = self._env.get_game().observation_tensor_shape()
```
vmoens, line 193: How expensive is `get_game()`? Can we cache it?

torchrl/envs/libs/openspiel.py
```diff
@@ -0,0 +1,358 @@
...
+            state=NonTensorSpec([]),
+            shape=self.batch_size,
+        )
+
+    def _set_seed(self, seed):
+        if seed is not None:
+            raise NotImplementedError("This environment has no seed.")
+
+    def current_player(self):
+        return self._env.current_player()
```
vmoens, line 183: @matteobettini if we have multiple players, do we have to consider that as a MARL problem?
What's the proper way to represent the env here?

I see various options:
- what we are currently doing, tracking the player in the TD
- Having a nested TD with `"player0"` and `"player1"` (should that be dynamic)
- i also thought: can we have 2 envs that play against each other? I guess for MCTS that could be nice to keep things separated

torchrl/envs/libs/openspiel.py
```diff
@@ -0,0 +1,358 @@
...
+                "terminated": done.clone(),
+                "state": state,
+                "current_player": current_player,
+            },
+            batch_size=self.batch_size,
+            device=self.device,
+        )
+
+        if not exclude_reward:
+            tensordict_out["reward"] = reward
```
vmoens, line 224: setting it after the fact is more expensive than putting it in the dict or not, maybe we could do 
```python
source = {
                "observation": observation,
                "done": done,
                "terminated": done.clone(),
                "state": state,
                "current_player": current_player,
            }
if not exclude_reward:
    source["reward"] = reward
tensordict_out = TensorDict(source, ...)
```

torchrl/envs/libs/openspiel.py
```diff
@@ -0,0 +1,358 @@
...
+        game = self._env.get_game()
+
+        if tensordict is not None and "state" in tensordict:
+            new_env = game.deserialize_state(tensordict["state"])
+        else:
+            new_env = game.new_initial_state()
+
+        self._env = new_env
+        self._update_action_mask()
+        return self._make_td_out(exclude_reward=True)
```
vmoens, line 254: interesting, so openspiel has a reward at reset time?

vmoens (commented):
Great work thanks for that, I left a couple of comments already.
Is the TODO blocking on this?

torchrl/envs/libs/openspiel.py
matteobettini: In pettingzoo we have some boatd games already. I guess the choice here also depends on if the game is turn-based or not (players can  act in parallel)

the way we did it in pettingzoo is to keep the players in 2 nested td groups (your option 2) and have a mask for the non-playing one.

vmoens: the goal is to use it for MCTS so I'd say turn based.
Self-play sort of thing

torchrl/envs/libs/openspiel.py
kurtamohler: It isn't very expensive, but I tried caching it with `functools.lru_cache`, and it does speed it up by about 6x, so I'll update it.

```python
import timeit
iters = 1_000_000
res0 = timeit.timeit(
    setup='import torchrl; env = torchrl.envs.OpenSpielEnv("chess")',
    stmt='env._env.get_game()',
    number=iters)
res1 = timeit.timeit(
    setup='import torchrl; env = torchrl.envs.OpenSpielEnv("chess")',
    stmt='env._get_game_cached()',
    number=iters)
print(f'nocache: {res0 / iters} s')
print(f'cache: {res1 / iters} s')
print(f'speedup: {res0 / res1} x')
```

```
nocache: 4.2987480599913395e-07 s
cache: 6.127979099983349e-08 s
speedup: 7.014952221366127x
```

torchrl/envs/libs/openspiel.py
kurtamohler: Yes, but it's just set to zero

torchrl/envs/libs/openspiel.py
kurtamohler: What if we call the flag `output_state`? I think that name makes it more obvious what exactly it does--it adds the state to the output of `reset` and `step`

torchrl/envs/libs/openspiel.py
kurtamohler: Well, flake8 fails, pointing out that `lru_cache` on an instance method can cause memory leaks: https://docs.astral.sh/ruff/rules/cached-instance-method/

So I decided to just cache it manually using an initially-None data member. Speedup decreased to 5x, but that still makes it worth doing.

torchrl/envs/libs/openspiel.py
kurtamohler: I will push an update to put the players in their own nested td groups with a mask, like in PettingZooEnv

kurtamohler:
At the moment, this only supports games where all actions are taken by the players, like in chess or tic-tac-toe. But I've realized that OpenSpiel also has a concept of chance nodes, where a random non-player action is taken. For instance, in Kuhn poker, the initial dealing of the cards is a chance node. In liar's dice, the outcome of rolling dice is a chance node. OpenSpiel has some methods to obtain all the possible chance actions and the associated probability distribution ([shown in this example](https://openspiel.readthedocs.io/en/latest/concepts.html#playing-a-trajectory)).

For now, I'll raise an error if a loaded game contains chance nodes and leave it as a TODO for me to add support for it in a follow-up PR--unless you would prefer for me to add it in this PR.

torchrl/envs/libs/openspiel.py
kurtamohler: I renamed to `return_state`, because that's what `PettingZooWrapper` calls it

vmoens:
> I suppose to support those games we need to allow the action spec's size to change at each step?

Yes I think having a dynamic space would be the way to go. See #2143

vmoens:
> OpenSpiel has some methods to obtain all the possible chance actions and the associated probability distribution (shown in this example).

That's an amazing feature. Happy to integrate it separately!

kurtamohler:
> > I suppose to support those games we need to allow the action spec's size to change at each step?
> 
> Yes I think having a dynamic space would be the way to go. See #2143

Actually, I'm not so sure that we would need dynamic action specs after all. It is true that `pyspiel.State.legal_actions()` can return a different number of actions than `pyspiel.Game.num_distinct_actions()`, but I'm pretty sure (not 100% sure) that only happens when it's a chance node, in which case the length of `legal_actions()` is `pyspiel.Game.max_chance_outcomes()` instead. Once I add support for chance nodes, it will have its own action spec separate from the players' action specs, and I think all action specs will maintain the same shape throughout the game.

kurtamohler:
I think I've addressed everything that needed to be fixed so far. Let me know if there is anything else

vmoens (approved):
Amazing thanks a mil for this!

PR #2351, opened 2024-08-02
[Feature] Store MARL parameters in module

vmoens:
## Description

We currently store the parameters in MARL modules in `self.params` in a TensorDictParams.
During a call to forward, we call `vmap` and `to_module` to put the batched parameters in place within the module.

This PR proposes to optionally make `self.params` a regular TensorDict (ie, `self.parameters()` will not see them because `self.params` is not within the `self.modules()` anymore), and place them in the `self._empty_net` instead. With that in place, the module has two copies of the parameters, but one is not accessible via `self.parameters()` (so things don't change from the user perspective).

We test that these two scenarios are identical and that sending the module to device does not create multiple distinct copies of the params.

cc @matteobettini

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -898,6 +898,51 @@ class TestMultiAgent:
         mlp.from_stateful_net(snet)
         assert (mlp.params == 1).all()
 
+    @retry(AssertionError, 5)
+    @pytest.mark.parametrize("n_agents", [3])
+    @pytest.mark.parametrize("share_params", [True])
+    @pytest.mark.parametrize("centralized", [True])
+    @pytest.mark.parametrize("n_agent_inputs", [6])
+    @pytest.mark.parametrize("batch", [(4,)])
+    @pytest.mark.parametrize("tdparams", [True, False])
+    def test_multiagent_mlp_tdparams(
+        self,
+        n_agents,
+        centralized,
+        share_params,
+        batch,
+        n_agent_inputs,
+        tdparams,
+        n_agent_outputs=2,
+    ):
+        torch.manual_seed(1)
+        mlp = MultiAgentMLP(
+            n_agent_inputs=n_agent_inputs,
+            n_agent_outputs=n_agent_outputs,
+            n_agents=n_agents,
+            centralized=centralized,
+            share_params=share_params,
+            depth=2,
+            use_td_params=tdparams,
+        )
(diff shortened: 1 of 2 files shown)


matteobettini:
Could you explain why we need this?

also, having 2 copies of the parameters is not error prone?

for example in methods like https://github.com/facebookresearch/BenchMARL/blob/d260eea5d4ef2ff5f0bea8ae36f68638ecb14865/benchmarl/models/common.py#L165 or in any general case where users access `self.parameters()` won’t things break?

vmoens:
We test that nothing breaks. I don't thing it's error prone, you never see two copies (for instance `parameters()` just returns one).

We need this because it makes initialization of the params more natural, mainly.

matteobettini:
So if a user modifies the content of one copy of the parameters the change is reflected in another copy? As in the function is sent.

But apart from being more natural, what use cases is it used for/ envisioned for?

matteobettini:
Maybe I am misreading the PR description: when you say 2 copies you mean:

1. 2 different sets of parameters?
2. 2 objects referring by pointer to the same parameter tensors?

test/test_modules.py
```diff
@@ -898,6 +898,51 @@ def one_outofplace(mod):
         mlp.from_stateful_net(snet)
         assert (mlp.params == 1).all()
 
+    @retry(AssertionError, 5)
```
matteobettini, line 901: What is the effect of this?

torchrl/modules/models/multiagent.py
```diff
@@ -483,6 +502,11 @@ class MultiAgentConvNet(MultiAgentNetBase):
             Defaults to ``2``.
         activation_class (Type[nn.Module]): activation class to be used.
             Default to :class:`torch.nn.ELU`.
+        use_td_params (bool, optional): if ``True``, the parameters can be found in `self.params` which is a
+            :class:`~tensordict.nn.TensorDictParams` object (which inherits both from `TensorDict` and `nn.Module`).
+            If ``False``, parameters are contained in `self._empty_net`. All things considered, these two approaches
+            should be roughly identical but not interchangeable: for instance, a ``state_dict`` created with
+            ``use_td_params=True`` cannot be used when ``use_td_params=False``.
```
matteobettini, line 509: Defaults to True missing from docs

PR #2380, opened 2024-08-08
[Feature] Support wrapping IsaacLab environments with GymEnv

### LINKED ISSUE #2369: [Feature Request] Make GymLikeEnv compatible with IsaacLab Envs
## Motivation

The current implementation of `GymLikeEnv._step(...)` is incompatible with IsaacLab environments due to the conversion to numpy when IsaacLab expects a torch tensor:

```
        action = tensordict.get(self.action_key)
        action_np = self.read_action(action)
        reward = 0
        for _ in range(self.wrapper_frame_skip):
            (
                obs,
                _reward,
                terminated,
                truncated,
                done,
                info_dict,
            ) = self._output_transform(self._env.step(action_np))
```

## Solution

Perhaps we can add a config parameter to `GymLikeEnv` to specify if the underlying environment uses either `np.ndarray` or `torch.tensor` types. 

## Checklist

yu-fz:
## Description

Adds a configuration parameter to `GymLikeEnv` to specify whether or not actions should be converted to numpy arrays before being passed to the step function of the underlying environment to retrieve the next state. Users can now pass `convert_actions_to_numpy=False` when wrapping environments that work with torch tensors. 

Let me know if a new test is required for this. 

## Motivation and Context

Previously, it was assumed that the step function argument expected numpy array types on CPU, which makes torchrl incompatible with environments like IsaacLab that operate on torch tensors and on the GPU. 

Closes #2369 

## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)
- [] Breaking change (fix or feature that would cause existing functionality to change)
- [] Example (update in the folder of examples)

## Checklist

### DIFF
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -3039,6 +3039,7 @@ class _EnvWrapper(EnvBase):
 
         self._constructor_kwargs = kwargs
         self._check_kwargs(kwargs)
+        self._convert_actions_to_numpy = kwargs.pop("convert_actions_to_numpy", True)
         self._env = self._build_env(**kwargs)  # writes the self._env attribute
         self._make_specs(self._env)  # writes the self._env attribute
         self.is_closed = False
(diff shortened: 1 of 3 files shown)


vmoens (approved):
LGTM
I guess that then we should also document better how to use the new Isaac API

Ideally, I'd like not to rely on Gym for Isaac and wrap the simulator directly, but first let's focus on making it work like this!

PR #2391, opened 2024-08-12
[BUG] Allow for composite action distributions in losses

albertbou92:
## Description

At the moment objective classes do not allow to use an actor with a composite distribution.

This PR aims to fix this. I have started by modifying the on-policy objectives and added tests for PPO.

Once these modification are correct, I will move on the the tests of the other on-policy objectives and then to all other objectives.

This PR requires the TensorDict PR https://github.com/pytorch/tensordict/pull/961 to be merged.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/torchrl/objectives/a2c.py
+++ b/torchrl/objectives/a2c.py
@@ -11,7 +11,12 @@ from typing import Tuple
 
 import torch
 from tensordict import TensorDict, TensorDictBase, TensorDictParams
-from tensordict.nn import dispatch, ProbabilisticTensorDictSequential, TensorDictModule
+from tensordict.nn import (
+    CompositeDistribution,
+    dispatch,
+    ProbabilisticTensorDictSequential,
+    TensorDictModule,
+)
 from tensordict.utils import NestedKey
 from torch import distributions as d
 
@@ -383,7 +388,11 @@ class A2CLoss(LossModule):
             entropy = dist.entropy()
         except NotImplementedError:
             x = dist.rsample((self.samples_mc_entropy,))
-            entropy = -dist.log_prob(x).mean(0)
+            if isinstance(dist, CompositeDistribution):
+                log_prob = dist.log_prob(x).get(self.tensor_keys.sample_log_prob)
+            else:
+                log_prob = dist.log_prob(x)
+            entropy = -log_prob.mean(0)
         return entropy.unsqueeze(-1)
 
     def _log_probs(
@@ -402,7 +411,11 @@ class A2CLoss(LossModule):
             self.actor_network
         ) if self.functional else contextlib.nullcontext():
             dist = self.actor_network.get_dist(tensordict_clone)
-        log_prob = dist.log_prob(action)
+        if isinstance(dist, CompositeDistribution):
+            tensordict = dist.log_prob(tensordict)
+            log_prob = tensordict.get(self.tensor_keys.sample_log_prob)
+        else:
+            log_prob = dist.log_prob(action)
         log_prob = log_prob.unsqueeze(-1)
         return log_prob, dist
 
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -477,6 +477,9 @@ class PPOLoss(LossModule):
             raise RuntimeError("tensordict prev_log_prob requires grad.")
 
         if isinstance(dist, CompositeDistribution):
+            if tensordict.get(self.tensor_keys.action).batch_size != tensordict.batch_size:
+                # This condition can be True in notensordict usage
+                tensordict.get(self.tensor_keys.action).batch_size = tensordict.batch_size
             tensordict = dist.log_prob(tensordict)
             log_prob = tensordict.get(self.tensor_keys.sample_log_prob)
         else:
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -451,7 +451,7 @@ class PPOLoss(LossModule):
         except NotImplementedError:
             x = dist.rsample((self.samples_mc_entropy,))
             if isinstance(dist, CompositeDistribution):
-                log_prob = dist.log_prob(x).get(self.tensor_keys.sample_log_prob)
+                log_prob = dist.log_prob(x).get("sample_log_prob")
             else:
                 log_prob = dist.log_prob(x)
             entropy = -log_prob.mean(0)
@@ -478,7 +478,7 @@ class PPOLoss(LossModule):
 
         if isinstance(dist, CompositeDistribution):
             tensordict = dist.log_prob(tensordict)
-            log_prob = tensordict.get(self.tensor_keys.sample_log_prob)
+            log_prob = tensordict.get("sample_log_prob")
         else:
             log_prob = dist.log_prob(action)
 
@@ -1118,12 +1118,8 @@ class KLPENPPOLoss(PPOLoss):
         except NotImplementedError:
             x = previous_dist.sample((self.samples_mc_kl,))
             if isinstance(current_dist, CompositeDistribution):
-                previous_log_prob = previous_dist.log_prob(x).get(
-                    self.tensor_keys.sample_log_prob
-                )
-                current_log_prob = current_dist.log_prob(x).get(
-                    self.tensor_keys.sample_log_prob
-                )
+                previous_log_prob = previous_dist.log_prob(x).get("sample_log_prob")
+                current_log_prob = current_dist.log_prob(x).get("sample_log_prob")
             else:
                 previous_log_prob = previous_dist.log_prob(x)
                 current_log_prob = current_dist.log_prob(x)
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -451,7 +451,7 @@ class PPOLoss(LossModule):
         except NotImplementedError:
             x = dist.rsample((self.samples_mc_entropy,))
             if isinstance(dist, CompositeDistribution):
-                log_prob = dist.log_prob(x).get("sample_log_prob")
+                log_prob = dist.log_prob(x).get(self.tensor_keys.sample_log_prob)
             else:
                 log_prob = dist.log_prob(x)
             entropy = -log_prob.mean(0)
@@ -478,7 +478,7 @@ class PPOLoss(LossModule):
 
         if isinstance(dist, CompositeDistribution):
             tensordict = dist.log_prob(tensordict)
-            log_prob = tensordict.get("sample_log_prob")
+            log_prob = tensordict.get(self.tensor_keys.sample_log_prob)
         else:
             log_prob = dist.log_prob(action)
 
@@ -1118,8 +1118,12 @@ class KLPENPPOLoss(PPOLoss):
         except NotImplementedError:
             x = previous_dist.sample((self.samples_mc_kl,))
             if isinstance(current_dist, CompositeDistribution):
-                previous_log_prob = previous_dist.log_prob(x).get("sample_log_prob")
-                current_log_prob = current_dist.log_prob(x).get("sample_log_prob")
+                previous_log_prob = previous_dist.log_prob(x).get(
+                    self.tensor_keys.sample_log_prob
+                )
+                current_log_prob = current_dist.log_prob(x).get(
+                    self.tensor_keys.sample_log_prob
+                )
             else:
                 previous_log_prob = previous_dist.log_prob(x)
                 current_log_prob = current_dist.log_prob(x)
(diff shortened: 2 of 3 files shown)


torchrl/objectives/a2c.py
```diff
@@ -383,7 +388,11 @@ def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
             entropy = dist.entropy()
         except NotImplementedError:
             x = dist.rsample((self.samples_mc_entropy,))
-            entropy = -dist.log_prob(x).mean(0)
+            if isinstance(dist, CompositeDistribution):
+                log_prob = dist.log_prob(x).get(self.tensor_keys.sample_log_prob)
+            else:
+                log_prob = dist.log_prob(x)
+            entropy = -log_prob.mean(0)
```
vmoens, line 395: what if the dist is wrapped (eg via distributions.TransformedDistribution or smth like that)?
Shouldn't we check if `log_prob` is a tensordict instead? That would be more generic

torchrl/objectives/a2c.py
```diff
@@ -402,7 +411,11 @@ def _log_probs(
             self.actor_network
         ) if self.functional else contextlib.nullcontext():
             dist = self.actor_network.get_dist(tensordict_clone)
-        log_prob = dist.log_prob(action)
+        if isinstance(dist, CompositeDistribution):
+            tensordict = dist.log_prob(tensordict)
+            log_prob = tensordict.get(self.tensor_keys.sample_log_prob)
+        else:
+            log_prob = dist.log_prob(action)
```
vmoens, line 418: ditto

torchrl/objectives/ppo.py
```diff
@@ -449,7 +450,11 @@ def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
             entropy = dist.entropy()
         except NotImplementedError:
             x = dist.rsample((self.samples_mc_entropy,))
-            entropy = -dist.log_prob(x).mean(0)
+            if isinstance(dist, CompositeDistribution):
+                log_prob = dist.log_prob(x).get(self.tensor_keys.sample_log_prob)
+            else:
+                log_prob = dist.log_prob(x)
+            entropy = -log_prob.mean(0)
```
vmoens, line 457: ditto

torchrl/objectives/ppo.py
```diff
@@ -466,12 +471,25 @@ def _log_weight(
...
 
+        if isinstance(dist, CompositeDistribution):
+            if (
+                tensordict.get(self.tensor_keys.action).batch_size
+                != tensordict.batch_size
+            ):
+                # This condition can be True in notensordict usage
+                tensordict.get(
+                    self.tensor_keys.action
+                ).batch_size = tensordict.batch_size
```
vmoens, line 487: get operations do not come for free, if you're accessing the same entity twice (eg action) I would do it just once before.

torchrl/objectives/ppo.py
```diff
@@ -466,12 +471,25 @@ def _log_weight(
             self.actor_network
         ) if self.functional else contextlib.nullcontext():
             dist = self.actor_network.get_dist(tensordict)
-        log_prob = dist.log_prob(action)
 
         prev_log_prob = tensordict.get(self.tensor_keys.sample_log_prob)
         if prev_log_prob.requires_grad:
             raise RuntimeError("tensordict prev_log_prob requires grad.")
 
+        if isinstance(dist, CompositeDistribution):
```
vmoens, line 479: ditto, isn't there a way to test if the dist returns td log-probs other than checking the id of the dist?

torchrl/objectives/ppo.py
```diff
@@ -1107,7 +1125,17 @@ def forward(self, tensordict: TensorDictBase) -> TensorDict:
             kl = torch.distributions.kl.kl_divergence(previous_dist, current_dist)
         except NotImplementedError:
             x = previous_dist.sample((self.samples_mc_kl,))
-            kl = (previous_dist.log_prob(x) - current_dist.log_prob(x)).mean(0)
+            if isinstance(current_dist, CompositeDistribution):
```
vmoens, line 1128: ditto

vmoens (commented):
Good work! 
There's an exception with `test_ppo_notensordict` in `test_cost.py`

Have a look at the couple of comments I left

albertbou92:
Thanks a lot for the feedback!

I have made a few changes:
1. I have removed any specific distribution id check `isinstance(dist, CompositeDistribution)`, which should make the code more general. I am instead simply checking whether the `log_prob` is a `TensorDict` as suggested, or alternatively if action is a `torch.Tensor`. I think using a `TransformedDistribution` to do that is a bit overkill as it requires coding some methods that we don't need, but if you prefer it I can further look into it.
2. Also added tests for A2C. At the moment, a composite action distribution is simply not compatible with `notensordict`. The action becomes a nested structure which makes it difficult. Any suggestion how to solve that?

torchrl/objectives/a2c.py
```diff
@@ -383,26 +383,39 @@ def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
             entropy = dist.entropy()
         except NotImplementedError:
             x = dist.rsample((self.samples_mc_entropy,))
-            entropy = -dist.log_prob(x).mean(0)
+            log_prob = dist.log_prob(x)
+            if isinstance(log_prob, TensorDict):
```
vmoens, line 387: A lazy stack is not a TensorDict but a TensorDict base.
Also ideally we would want this to work with tensorclasses.
The way to go should be to use `is_tensor_collection` from tensordict lib.

torchrl/objectives/ppo.py
```diff
@@ -449,28 +449,38 @@ def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
             entropy = dist.entropy()
         except NotImplementedError:
             x = dist.rsample((self.samples_mc_entropy,))
-            entropy = -dist.log_prob(x).mean(0)
+            log_prob = dist.log_prob(x)
+            if isinstance(log_prob, TensorDict):
```
vmoens, line 453: ditto

torchrl/objectives/ppo.py
```diff
@@ -1107,7 +1117,17 @@ def forward(self, tensordict: TensorDictBase) -> TensorDict:
             kl = torch.distributions.kl.kl_divergence(previous_dist, current_dist)
         except NotImplementedError:
             x = previous_dist.sample((self.samples_mc_kl,))
-            kl = (previous_dist.log_prob(x) - current_dist.log_prob(x)).mean(0)
+            previous_log_prob = previous_dist.log_prob(x)
+            current_log_prob = current_dist.log_prob(x)
+            if isinstance(x, TensorDict):
```
vmoens, line 1122: ditto

vmoens (commented):
I think it's safe to assume that people won't use this feature with non-tensordict inputs, because the action will be a tensordict anyway. 
I would just document it properly in the loss docstrings where we explain how to use the loss without tensordict.

vmoens (commented):
I think it's safe to assume that people won't use this feature with non-tensordict inputs, because the action will be a tensordict anyway. 
I would just document it properly in the loss docstrings where we explain how to use the loss without tensordict.

albertbou92:
Done! I will do the off-policy and offline losses in separate PRs.

torchrl/objectives/a2c.py
```diff
@@ -383,26 +395,39 @@ def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
             entropy = dist.entropy()
         except NotImplementedError:
             x = dist.rsample((self.samples_mc_entropy,))
-            entropy = -dist.log_prob(x).mean(0)
+            log_prob = dist.log_prob(x)
```
vmoens, line 398: Previously, was this a bug or did we sum the log-probs automatically?

torchrl/objectives/a2c.py
```diff
@@ -383,26 +395,39 @@ def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
...
             *self.actor_network.in_keys, strict=False
         ).clone()
         with self.actor_network_params.to_module(
             self.actor_network
         ) if self.functional else contextlib.nullcontext():
             dist = self.actor_network.get_dist(tensordict_clone)
-        log_prob = dist.log_prob(action)
+
+        def check_requires_grad(tensor):
+            if tensor.requires_grad:
```
vmoens, line 418: Tensordict has a requires_grad property

torchrl/objectives/ppo.py
```diff
@@ -449,28 +460,38 @@ def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
...
-                f"tensordict stored {self.tensor_keys.action} requires grad."
-            )
 
         with self.actor_network_params.to_module(
             self.actor_network
         ) if self.functional else contextlib.nullcontext():
             dist = self.actor_network.get_dist(tensordict)
-        log_prob = dist.log_prob(action)
+
+        def check_requires_grad(tensor, key=self.tensor_keys.action):
```
vmoens, line 480: Ditto

vmoens (commented):
Thanks for the PR!
Do we need a simple test for this?
Like a dedicated function in PPOTest that runs it with composite dists?

torchrl/objectives/a2c.py
albertbou92: This is simply because the log_prob() method for a composite dist will return a TD, so we compute the entropy in 2 steps. 

This is the old version:
```
    def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
        try:
            entropy = dist.entropy()
        except NotImplementedError:
            x = dist.rsample((self.samples_mc_entropy,))
            entropy = -dist.log_prob(x).mean(0)
        return entropy.unsqueeze(-1)
```

This is the new version. It simply retrieves the log tensor before computing the entropy.
```
    def get_entropy_bonus(self, dist: d.Distribution) -> torch.Tensor:
        try:
            entropy = dist.entropy()
        except NotImplementedError:
            x = dist.rsample((self.samples_mc_entropy,))
            log_prob = dist.log_prob(x)
            if is_tensor_collection(log_prob):
                log_prob = log_prob.get(self.tensor_keys.sample_log_prob)
            entropy = -log_prob.mean(0)
        return entropy.unsqueeze(-1)
```

albertbou92:
Regarding a dedicated test, how do you usually approach this decision?

When I thought about testing different dists I saw it a bit like testing for different types of ValueEstimators. It should work both with single dists and with composite dists in all the tested situations. So I added it to all tests (except the `notensordict` tests).

We could probably switch to a single dedicated test function though.

vmoens (approved):
LGTM thanks for this!

albertbou92:
Computing the entropy for composite distributions is not fully resolved, particularly when dealing with a composite distribution that includes some distributions with an implemented entropy() method and others without.

We could add an entropy method to `CompositeDistribution` as suggested in this PR: https://github.com/pytorch/tensordict/pull/981

wdyt?

vmoens:
From discord:

> Hey, sorry, was out on Holidays last week. I can have a look at the PR. What I was wondering: Why is the behavior for log_prob different between composite and non-composite distributions? I.e. why not have log_prob always just return the (combined) log-prob and have a separate log_prob_composite function for the specific case of returning individual log-probs in the composite distribution?


I think it's a good idea, I prefer a `log_prob` which always returns a single tensor.

vmoens:
To add on my previous comment, here is how I would address this:

- Add a warning saying that from v0.7 log-prob will return a tensor. Users can achieve this already by constructing the lib with an additional temporary kwarg `return_log_prob_tensor=True` (this will be the default in v0.7). 
- Add a method `log_prob_composite` for those who need it.

albertbou92:
> From [discord](https://discord.com/channels/1171857748607115354/1171857749060112405/1280395570472554528):
> 
> > Hey, sorry, was out on Holidays last week. I can have a look at the PR. What I was wondering: Why is the behavior for log_prob different between composite and non-composite distributions? I.e. why not have log_prob always just return the (combined) log-prob and have a separate log_prob_composite function for the specific case of returning individual log-probs in the composite distribution?
> 
> I think it's a good idea, I prefer a `log_prob` which always returns a single tensor.

makes sense

vmoens:
Merging this to clear space in the PR list but we should take care of https://github.com/pytorch/rl/pull/2391#issuecomment-2326461726 sooner than later! Wanna give a shot at it or should I?

albertbou92:
I will give it a shot, give me a few days

albertbou92:
> Merging this to clear space in the PR list but we should take care of [#2391 (comment)](https://github.com/pytorch/rl/pull/2391#issuecomment-2326461726) sooner than later! Wanna give a shot at it or should I?


I will give it a shot, give me a few days

PR #2433, opened 2024-09-12
[Performance] Faster `PrioritizedSliceSampler._padded_indices`

### LINKED ISSUE #2431: [BUG] `PrioritizedSliceSampler._padded_indices` is slow
## Describe the bug

`PrioritizedSliceSampler`'s `_padded_indices` method seems to take more time than it really needs. In the example below, it takes up ~55% of the runtime when a sample is taken. This can likely be improved.

## To Reproduce

Run the following script with the command `python -m cProfile <script-name>  | grep -w 'ncalls\|_padded_indices\|samplers\.py.*sample\|^time'`

```python
import torch
from torchrl.data.replay_buffers import TensorDictReplayBuffer, LazyTensorStorage, PrioritizedSliceSampler
import timeit
import numpy as np

device = 'cpu'

def make_data(num_traj, traj_len):
    done = torch.zeros(2, num_traj * traj_len, 1, dtype=torch.bool, device=device)
    priority = torch.ones(2, num_traj * traj_len, device=device)
    # This makes half the trajectories traj_len-1, and the other half
    # traj_len+1, so that half of them require padding.
    done[:, (traj_len-1-1)::2*traj_len] = 1
    done[:, (traj_len-1-1) + traj_len+1::2*traj_len] = 1
    priority[:, :traj_len] = 10
    data = TensorDict(
        {"done": done, "priority": priority},
        batch_size=[2, num_traj * traj_len],
        device=device,
    )
    return data

def make_rb(num_traj, traj_len):
    torch.manual_seed(0)
    np.random.seed(0)
    data = make_data(num_traj, traj_len)
    sampler = PrioritizedSliceSampler(
        max_capacity=data.numel(),
        alpha=1.0,
        beta=1.0,
        end_key="done",
        slice_len=10,
    )
    rb = TensorDictReplayBuffer(
        storage=LazyTensorStorage(data.numel(), ndim=2, device=device),
        sampler=sampler,
        batch_size=20,
    )
    rb.extend(data)
    return rb

def measure_time(num_iters, num_traj, traj_len):
    rb = make_rb(num_traj, traj_len)
    timer = timeit.Timer(stmt=lambda: rb.sample())
    t = timer.timeit(num_iters) / num_iters
    return t

num_measurements = 30
num_iters = 30
times = []
for _ in range(num_measurements):
    times.append(measure_time(num_iters, num_traj=100,  traj_len=10_000))

t = np.median(times)
print(f"time: {t * 1000} ms")
```

On my machine, I get the following output:

```
time: 22.23507576660874 ms
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      900    0.020    0.000   11.616    0.013 samplers.py:1845(_padded_indices)
      900    1.685    0.002   20.767    0.023 samplers.py:1904(sample)
      900    0.109    0.000    0.505    0.001 samplers.py:473(sample)
```

This shows that the typical time for the `ReplayBuffer.sample` call is about 22.23 ms.

Also, the profile shows that the script cumulatively spent 20.767 seconds in `PrioritizedSliceSampler.sample`, and 11.616 seconds in `PrioritizedSliceSampler._padded_indices`, so `_padded_indices` apparently takes about 55% of the total runtime of `sample`.

## Checklist

kurtamohler:
## Description

Speeds up `PrioritizedSliceSampler._padded_indices` by about 2x.

Running the performance script given in https://github.com/pytorch/rl/issues/2431#issue-2517641562, my machine gives the following:

```
time: 7.804846733536882 ms
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      900    2.739    0.003    3.114    0.003 samplers.py:1845(_padded_indices)
      900    0.197    0.000    6.895    0.008 samplers.py:1907(sample)
      900    0.067    0.000    0.342    0.000 samplers.py:473(sample)
```

This is a speedup of (22.235 / 7.804) = 2.8x

Although sometimes, the runtime per `sample` call reaches up to 11.8 ms, a speedup of ~1.8x.

## Motivation and Context

close #2431

## Types of changes

## Checklist

### DIFF
--- a/torchrl/data/replay_buffers/samplers.py
+++ b/torchrl/data/replay_buffers/samplers.py
@@ -22,7 +22,7 @@ from tensordict.utils import NestedKey
 
 from torchrl._extension import EXTENSION_WARNING
 
-from torchrl._utils import _replace_last, implement_for, logger
+from torchrl._utils import _replace_last, logger
 from torchrl.data.replay_buffers.storages import Storage, StorageEnsemble, TensorStorage
 from torchrl.data.replay_buffers.utils import _is_int, unravel_index
 
@@ -1842,28 +1842,31 @@ class PrioritizedSliceSampler(SliceSampler, PrioritizedSampler):
     ) -> None:
         return PrioritizedSampler.mark_update(self, index, storage=storage)
 
-    @implement_for("torch", "2.4")
     def _padded_indices(self, shapes, arange) -> torch.Tensor:
         # this complex mumbo jumbo creates a left padded tensor with valid indices on the right, e.g.
         # tensor([[ 0,  1,  2,  3,  4],
         #         [-1, -1,  5,  6,  7],
         #         [-1,  8,  9, 10, 11]])
         # where the -1 items on the left are padded values
-        st, off = torch._nested_compute_contiguous_strides_offsets(shapes.flip(0))
-        nt = torch._nested_view_from_buffer(
-            arange.flip(0).contiguous(), shapes.flip(0), st, off
-        )
-        pad = nt.to_padded_tensor(-1).flip(-1).flip(0)
-        return pad
+        num_groups = shapes.shape[0]
+        max_group_len = shapes.max()
+        pad_lengths = max_group_len - shapes
 
-    @implement_for("torch", None, "2.4")
-    def _padded_indices(self, shapes, arange) -> torch.Tensor:  # noqa: F811
-        arange = arange.flip(0).split(shapes.flip(0).squeeze().unbind())
-        return (
-            torch.nn.utils.rnn.pad_sequence(arange, batch_first=True, padding_value=-1)
-            .flip(-1)
-            .flip(0)
+        # Get all the start and end indices within arange for each group
+        group_ends = shapes.cumsum(0)
+        group_starts = torch.empty_like(group_ends)
+        group_starts[0] = 0
+        group_starts[1:] = group_ends[:-1]
+        pad = torch.empty(
+            (num_groups, max_group_len), dtype=arange.dtype, device=arange.device
         )
+        for pad_row, group_start, group_end, pad_len in zip(
+            pad, group_starts, group_ends, pad_lengths
+        ):
+            pad_row[:pad_len] = -1
+            pad_row[pad_len:] = arange[group_start:group_end]
+
+        return pad
 
     def _preceding_stop_idx(self, storage, lengths, seq_length, start_idx):
         preceding_stop_idx = self._cache.get("preceding_stop_idx")


torchrl/data/replay_buffers/samplers.py
```diff
@@ -1842,28 +1842,31 @@ def mark_update(
...
         )
-        pad = nt.to_padded_tensor(-1).flip(-1).flip(0)
-        return pad
+        for pad_row, group_start, group_end, pad_len in zip(
+            pad, group_starts, group_ends, pad_lengths
+        ):
+            pad_row[:pad_len] = -1
+            pad_row[pad_len:] = arange[group_start:group_end]
 
-    @implement_for("torch", None, "2.4")
```
kurtamohler, line 1859: On my machine, both of the previous implementations had roughly the same performance

torchrl/data/replay_buffers/samplers.py
```diff
@@ -1842,28 +1842,31 @@ def mark_update(
...
+        group_ends = shapes.cumsum(0)
+        group_starts = torch.empty_like(group_ends)
+        group_starts[0] = 0
+        group_starts[1:] = group_ends[:-1]
+        pad = torch.empty(
+            (num_groups, max_group_len), dtype=arange.dtype, device=arange.device
         )
-        pad = nt.to_padded_tensor(-1).flip(-1).flip(0)
-        return pad
+        for pad_row, group_start, group_end, pad_len in zip(
```
kurtamohler, line 1863: It would be nice to be able to get rid of this for loop, but I don't see a good way to do it. I think indexing different sized ranges under each row of a tensor is only possible if you list out all the indices, like for `torch.index_select` or `torch.gather`.

If `torch.nn.utils.rnn.pad_sequence` supported left-padding, we could just use that, but it doesn't.

torchrl/data/replay_buffers/samplers.py
```diff
@@ -1842,28 +1842,31 @@ def mark_update(
...
-    @implement_for("torch", "2.4")
     def _padded_indices(self, shapes, arange) -> torch.Tensor:
         # this complex mumbo jumbo creates a left padded tensor with valid indices on the right, e.g.
         # tensor([[ 0,  1,  2,  3,  4],
         #         [-1, -1,  5,  6,  7],
         #         [-1,  8,  9, 10, 11]])
         # where the -1 items on the left are padded values
-        st, off = torch._nested_compute_contiguous_strides_offsets(shapes.flip(0))
-        nt = torch._nested_view_from_buffer(
-            arange.flip(0).contiguous(), shapes.flip(0), st, off
```
kurtamohler, line 1854: It seems that the main thing that the new implementation improves upon is that it doesn't need to do these `torch.flip` operations, and the time it saves on seems to outweigh the overhead of the for loop

torchrl/data/replay_buffers/samplers.py
kurtamohler: FWIW, I did try getting rid of the for loop by listing out all the indices and using `index_copy_` like so:

```python
shapes = shapes.flatten()
num_groups = shapes.shape[0]
max_group_len = shapes.max()
pad_lengths = max_group_len - shapes
pad_before_groups = pad_lengths.cumsum(0)
pad_before_indices = torch.repeat_interleave(pad_before_groups, shapes)
indices = torch.arange(arange.shape[0], dtype=arange.dtype, device=arange.device) + pad_before_indices
p = torch.full(
    (num_groups * max_group_len,),
    -1,
    dtype=arange.dtype,
    device=arange.device
)
p.index_copy_(0, indices, arange)
pad = p.reshape((num_groups, max_group_len))
```

It worked, but it gave very performance to the old implementations

kurtamohler:
Something is interesting about the cProfile results I got. Before this PR, `_padded_indices` took 11.616 s cumulative time and `sample` was about 20.767. After this PR, `_padded_indices` is 3.114 and `sample` is 6.895.

`_padded_indices` decreased by 8.502 s, while `sample` decreased by 13.872. I would have expected all of the time savings to come from `_padded_indices`, but apparently something outside of that function also gained a performance boost.

I would guess that the reason is that `_padded_indices` is now outputing a contiguous tensor, whereas it used to be noncontiguous, so operations performed on the output can be a little faster

kurtamohler:
One other thing, I tried to measure the cuda performance, but changing the device in my script https://github.com/pytorch/rl/issues/2431#issue-2517641562 to `'cuda'` still ended up calling `_padded_indices` with cpu tensors. I'm not sure if cuda is just not supported in `PrioritizedSliceSampler` or if something is wrong with how I tried to set the device in the script

torchrl/data/replay_buffers/samplers.py
vmoens: IIRC, `torch.nn.utils.rnn.pad_sequence` left padding was mentioned during a review not long ago
cc @mikaylagawarecki

vmoens (commented):
LGTM thanks for taking care of this!

vmoens (approved):
LGTM thanks for taking care of this!

vmoens (approved):
LGTM thanks for taking care of this!

PR #2447, opened 2024-09-22
[Feature] SAC compatibility with composite distributions.

albertbou92:
## Description

This PR adapts the SAC objective to be compatible with composite distributions.

DiscreteSAC seems a little bit more tricky, as it requires computing the log probabilities of all possible actions. In the case of composite distributions this might mean the probabilities of all combinations of actions. 

## Motivation and Context

## Types of changes

## Checklist


torchrl/objectives/sac.py
```diff
@@ -251,7 +264,7 @@ class _AcceptedKeys:
             state_action_value (NestedKey): The input tensordict key where the
                 state action value is expected.  Defaults to ``"state_action_value"``.
             log_prob (NestedKey): The input tensordict key where the log probability is expected.
-                Defaults to ``"_log_prob"``.
+                Defaults to ``"sample_log_prob"``.
```
albertbou92, line 267: I renamed this key to make it work with the new API. It did not seem to break anything, but I imagine there was a reason for the naming.

torchrl/objectives/sac.py
```diff
@@ -450,9 +463,7 @@ def target_entropy(self):
             else:
                 action_container_shape = action_spec.shape
             target_entropy = -float(
-                action_spec[self.tensor_keys.action]
```
albertbou92, line 453: This too I had to change. Shouldn't  action spec already have the correct shape?

PR #2452, opened 2024-09-24
[Feature] Add scheduler for alpha/beta parameters of PrioritizedSampler

### LINKED ISSUE #1575: [Feature Request] Support for alpha and beta parameters' schedule in torchrl.data.PrioritizedReplayBuffer
## Motivation

Since in the original [PER paper](https://arxiv.org/abs/1511.05952) the parameter beta is changing its value during the training, it would be desirable to be able to perform similar experiments with torch-rl's Prioritized Experience Replay.

## Solution

Being able to manually change beta or both alpha and beta by calling a method on a replay buffer or modifying its properties.

## Alternatives

OpenAI baseline's implementation does it by providing an additional parameter beta during sampling (while alpha is fixed). It would be also possible to implement this with a scheduler (similar to lr schedules, like in [torch.optim](https://pytorch.org/docs/stable/optim.html))

## Checklist

LTluttmann:
## Description
Add scheduler for alpha/beta parameters of PrioritizedSampler.

## Motivation and Context

close #1575 

Following the suggestions made by @vmoens in issue #1575, this PR adds different Scheduler classes through which the user can adjust the alpha and beta parameters of the `PrioritizedSampler` during training when using the `PrioritizedReplayBuffer`. This is explicitly suggested in the paper "Schaul, T.; Quan, J.; Antonoglou, I.; and Silver, D. 2015. Prioritized experience replay".

The main reason to use separate scheduler classes for the annealing instead of a simple linear annealing (as also suggested by @vmoens in issue #1575) is the greater flexibility for the users. This way, the annealing can take place for example after taking a sample from the replay buffer or after a full training epoch (depending on where the user places the `scheduler.step()` command). Also, through the `LinearScheduler`, `StepScheduler` and `LambdaScheduler` different annealing schemes can be used (or new ones can be easily created). 

## Types of changes

## Checklist

### DIFF
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -58,6 +58,11 @@ from torchrl.data.replay_buffers.samplers import (
     SliceSampler,
     SliceSamplerWithoutReplacement,
 )
+from torchrl.data.replay_buffers.scheduler import (
+    LinearScheduler,
+    SchedulerList,
+    StepScheduler,
+)
 
 from torchrl.data.replay_buffers.storages import (
     LazyMemmapStorage,
@@ -97,11 +102,6 @@ from torchrl.envs.transforms.transforms import (
     ToTensorImage,
     UnsqueezeTransform,
     VecNorm,
-)
-from torchrl.data.replay_buffers.scheduler import (
-    LinearScheduler, 
-    StepScheduler, 
-    SchedulerList
 )
 
 
@@ -3040,33 +3040,39 @@ def test_prioritized_parameter_scheduler():
     LINEAR_STEPS = 100
     TOTAL_STEPS = 200
     rb = TensorDictPrioritizedReplayBuffer(
-        alpha=INIT_ALPHA, 
-        beta=INIT_BETA, 
-        storage=ListStorage(max_size=2000)
+        alpha=INIT_ALPHA, beta=INIT_BETA, storage=ListStorage(max_size=2000)
     )
-    data = TensorDict(
-        {
-            "data": torch.randn(1000, 5)
-        }, 
-        batch_size=1000
-    )
+    data = TensorDict({"data": torch.randn(1000, 5)}, batch_size=1000)
     rb.extend(data)
     alpha_scheduler = LinearScheduler(
         rb, param_name="alpha", final_value=0.0, num_steps=LINEAR_STEPS
     )
     beta_scheduler = StepScheduler(
-        rb, param_name="beta", gamma=GAMMA, n_steps=EVERY_N_STEPS, max_value=1.0, mode="additive"
+        rb,
+        param_name="beta",
+        gamma=GAMMA,
+        n_steps=EVERY_N_STEPS,
+        max_value=1.0,
+        mode="additive",
     )
-    scheduler = SchedulerList(scheduler=(alpha_scheduler, beta_scheduler))
-    expected_alpha_vals = np.linspace(INIT_ALPHA, 0.0, num=LINEAR_STEPS+1)
-    expected_alpha_vals = np.pad(expected_alpha_vals, (0, TOTAL_STEPS-LINEAR_STEPS), constant_values=0.0)
+    scheduler = SchedulerList(schedulers=(alpha_scheduler, beta_scheduler))
+    expected_alpha_vals = np.linspace(INIT_ALPHA, 0.0, num=LINEAR_STEPS + 1)
+    expected_alpha_vals = np.pad(
+        expected_alpha_vals, (0, TOTAL_STEPS - LINEAR_STEPS), constant_values=0.0
+    )
     expected_beta_vals = [INIT_BETA]
-    for _ in range((TOTAL_STEPS // EVERY_N_STEPS -1)):
+    for _ in range((TOTAL_STEPS // EVERY_N_STEPS - 1)):
         expected_beta_vals.append(expected_beta_vals[-1] + GAMMA)
-    expected_beta_vals = np.atleast_2d(expected_beta_vals).repeat(EVERY_N_STEPS).clip(None, 1.0)
+    expected_beta_vals = (
+        np.atleast_2d(expected_beta_vals).repeat(EVERY_N_STEPS).clip(None, 1.0)
+    )
     for i in range(TOTAL_STEPS):
-        assert np.isclose(rb.sampler.alpha, expected_alpha_vals[i]), f"expected {expected_alpha_vals[i]}, got {rb.sampler.alpha}"
-        assert np.isclose(rb.sampler.beta, expected_beta_vals[i]), f"expected {expected_beta_vals[i]}, got {rb.sampler.beta}"
+        assert np.isclose(
+            rb.sampler.alpha, expected_alpha_vals[i]
+        ), f"expected {expected_alpha_vals[i]}, got {rb.sampler.alpha}"
+        assert np.isclose(
+            rb.sampler.beta, expected_beta_vals[i]
+        ), f"expected {expected_beta_vals[i]}, got {rb.sampler.beta}"
         rb.sample(20)
         scheduler.step()
 
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -98,6 +98,12 @@ from torchrl.envs.transforms.transforms import (
     UnsqueezeTransform,
     VecNorm,
 )
+from torchrl.data.replay_buffers.scheduler import (
+    LinearScheduler, 
+    StepScheduler, 
+    SchedulerList
+)
+
 
 OLD_TORCH = parse(torch.__version__) < parse("2.0.0")
 _has_tv = importlib.util.find_spec("torchvision") is not None
@@ -3024,6 +3030,45 @@ def test_prioritized_slice_sampler_episodes(device):
     assert {1, 3} == set(
         torch.cat(episodes).cpu().tolist()
     ), "after priority update, only episode 1 and 3 are expected to be sampled"
+
+
+def test_prioritized_parameter_scheduler():
+    INIT_ALPHA = 0.7
+    INIT_BETA = 0.6
+    GAMMA = 0.1
+    EVERY_N_STEPS = 10
+    LINEAR_STEPS = 100
+    TOTAL_STEPS = 200
+    rb = TensorDictPrioritizedReplayBuffer(
+        alpha=INIT_ALPHA, 
+        beta=INIT_BETA, 
+        storage=ListStorage(max_size=2000)
+    )
+    data = TensorDict(
+        {
+            "data": torch.randn(1000, 5)
+        }, 
+        batch_size=1000
+    )
+    rb.extend(data)
+    alpha_scheduler = LinearScheduler(
+        rb, param_name="alpha", final_value=0.0, num_steps=LINEAR_STEPS
+    )
+    beta_scheduler = StepScheduler(
+        rb, param_name="beta", gamma=GAMMA, n_steps=EVERY_N_STEPS, max_value=1.0, mode="additive"
+    )
+    scheduler = SchedulerList(scheduler=(alpha_scheduler, beta_scheduler))
+    expected_alpha_vals = np.linspace(INIT_ALPHA, 0.0, num=LINEAR_STEPS+1)
+    expected_alpha_vals = np.pad(expected_alpha_vals, (0, TOTAL_STEPS-LINEAR_STEPS), constant_values=0.0)
+    expected_beta_vals = [INIT_BETA]
+    for _ in range((TOTAL_STEPS // EVERY_N_STEPS -1)):
+        expected_beta_vals.append(expected_beta_vals[-1] + GAMMA)
+    expected_beta_vals = np.atleast_2d(expected_beta_vals).repeat(EVERY_N_STEPS).clip(None, 1.0)
+    for i in range(TOTAL_STEPS):
+        assert np.isclose(rb.sampler.alpha, expected_alpha_vals[i]), f"expected {expected_alpha_vals[i]}, got {rb.sampler.alpha}"
+        assert np.isclose(rb.sampler.beta, expected_beta_vals[i]), f"expected {expected_beta_vals[i]}, got {rb.sampler.beta}"
+        rb.sample(20)
+        scheduler.step()
 
 
 class TestEnsemble:
--- a/torchrl/data/replay_buffers/scheduler.py
+++ b/torchrl/data/replay_buffers/scheduler.py
@@ -1,0 +1,235 @@
+import numpy as np
+from typing import Callable, Dict, Any
+
+from .replay_buffers import ReplayBuffer
+from .samplers import Sampler
+
+
+class ParameterScheduler:
+
+    """Scheduler to adjust the value of a given parameter of a replay buffer's sampler, e.g. the 
+    alpha and beta values in the PrioritizedSampler.
+
+    Args:
+        rb (ReplayBuffer): the replay buffer whose sampler to adjust
+        param_name (str): the name of the attribute to adjust, e.g. `beta` to adjust the beta parameter
+        min_value (Union[int, float], optional): a lower bound for the parameter to be adjusted
+            Defaults to None.
+        max_value (Union[int, float], optional): an upper bound for the parameter to be adjusted
+            Defaults to None
+
+    """
+
+    def __init__(
+        self, 
+        obj: ReplayBuffer | Sampler, 
+        param_name: str,
+        min_value: int | float = None, 
+        max_value: int | float = None
+    ):
+        if not isinstance(obj, ReplayBuffer) and not isinstance(obj, Sampler):
+            raise TypeError(
+                f"ParameterScheduler only supports Sampler class. Pass either ReplayBuffer or Sampler object. Got {type(obj)}"
+            )
+        self.sampler = obj.sampler if isinstance(obj, ReplayBuffer) else obj
+        self.param_name = param_name
+        self._min_val = min_value
+        self._max_val = max_value
+        if not hasattr(self.sampler, self.param_name):
+            raise ValueError(f"Provided class {obj.__name__} does not have an attribute {param_name}")
+        self.initial_val = getattr(self.sampler, self.param_name)
+        self._step_cnt = 0
+
+    def state_dict(self):
+        """Return the state of the scheduler as a :class:`dict`.
+
+        It contains an entry for every variable in self.__dict__ which
+        is not the optimizer.
+        """
+        return {
+            key: value for key, value in self.__dict__.items() if key != "sampler"
+        }
+
+    def load_state_dict(self, state_dict: Dict[str, Any]):
+        """Load the scheduler's state.
+
+        Args:
+            state_dict (dict): scheduler state. Should be an object returned
+                from a call to :meth:`state_dict`.
+        """
+        self.__dict__.update(state_dict)
+
+    def step(self):
+        self._step_cnt += 1
+        # Apply the step function
+        new_value = self._step()
+        # clip value to specified range
+        new_value_clipped = np.clip(new_value, a_min=self._min_val, a_max=self._max_val)
+        # Set the new value of the parameter dynamically
+        setattr(self.sampler, self.param_name, new_value_clipped)
+
+    def _step(self):
+        raise NotImplementedError
+
+
+class LambdaScheduler(ParameterScheduler):
+    """Similar to torch.optim.LambdaLR, this class sets a parameter to its initial value
+    times a given function. 
+
+    Args:
+        obj (ReplayBuffer | Sampler): the replay buffer whose sampler to adjust (or the sampler itself)
+        param_name (str): the name of the attribute to adjust, e.g. `beta` to adjust the 
+            beta parameter
+        lambda_fn (function): A function which computes a multiplicative factor given an integer 
+            parameter step_count
+        min_value (Union[int, float], optional): a lower bound for the parameter to be adjusted
+            Defaults to None.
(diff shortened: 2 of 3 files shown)


torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+from .samplers import Sampler
+
+
+class ParameterScheduler:
+    """Scheduler to adjust the value of a given parameter of a replay buffer's sampler.
+
+    Scheduler can for example be used to alter the alpha and beta values in the PrioritizedSampler.
+
+    Args:
+        rb (ReplayBuffer): the replay buffer whose sampler to adjust
```
vmoens, line 15:
```suggestion
        obj (ReplayBuffer or Sampler): the replay buffer or sampler whose sampler to adjust
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        param_name (str): the name of the attribute to adjust, e.g. `beta` to adjust the beta parameter
+        min_value (Union[int, float], optional): a lower bound for the parameter to be adjusted
+            Defaults to None.
```
vmoens, line 18:
```suggestion
            Defaults to `None`.
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        max_value (Union[int, float], optional): an upper bound for the parameter to be adjusted
+            Defaults to None
```
vmoens, line 20:
```suggestion
            Defaults to `None`.
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+    """
+
+    def __init__(
+        self,
+        obj: ReplayBuffer | Sampler,
+        param_name: str,
+        min_value: int | float = None,
+        max_value: int | float = None,
+    ):
+        if not isinstance(obj, ReplayBuffer) and not isinstance(obj, Sampler):
```
vmoens, line 31:
```suggestion
        if not isinstance(obj, (ReplayBuffer, Sampler)):
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+            raise TypeError(
+                f"ParameterScheduler only supports Sampler class. Pass either ReplayBuffer or Sampler object. Got {type(obj)}"
```
vmoens, line 33:
```suggestion
                f"ParameterScheduler only supports Sampler class. Pass either `ReplayBuffer` or `Sampler` object. Got {type(obj)} instead."
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+            )
+        self.sampler = obj.sampler if isinstance(obj, ReplayBuffer) else obj
+        self.param_name = param_name
+        self._min_val = min_value
+        self._max_val = max_value
+        if not hasattr(self.sampler, self.param_name):
+            raise ValueError(
+                f"Provided class {obj.__name__} does not have an attribute {param_name}"
```
vmoens, line 41:
```suggestion
                f"Provided class {type(obj).__name__} does not have an attribute {param_name}"
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+            )
+        self.initial_val = getattr(self.sampler, self.param_name)
```
vmoens, line 43: do we want to copy that? If it's a tensor its value could change in-place

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        self._step_cnt = 0
+
+    def state_dict(self):
+        """Return the state of the scheduler as a :class:`dict`.
```
vmoens, line 47:
```suggestion
        """Returns the state of the scheduler as a :class:`dict`.
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+
+        It contains an entry for every variable in self.__dict__ which
```
vmoens, line 49:
```suggestion
        It contains an entry for every variable in ``self.__dict__`` which
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        is not the optimizer.
```
vmoens, line 50:
```suggestion
        is not the sampler.
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        self._min_val = min_value
+        self._max_val = max_value
+        if not hasattr(self.sampler, self.param_name):
+            raise ValueError(
+                f"Provided class {obj.__name__} does not have an attribute {param_name}"
+            )
+        self.initial_val = getattr(self.sampler, self.param_name)
+        self._step_cnt = 0
+
+    def state_dict(self):
```
vmoens, line 46: do we want to match the nn.Module.state_dict [signature](https://github.com/pytorch/pytorch/blob/dd4a51b39aa02cba23b3a387b41c5026770d9220/torch/nn/modules/module.py#L2133-L2145) here?

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        """Return the state of the scheduler as a :class:`dict`.
+
+        It contains an entry for every variable in self.__dict__ which
+        is not the optimizer.
+        """
+        return {key: value for key, value in self.__dict__.items() if key != "sampler"}
```
vmoens, line 52:
```suggestion
        sd = dict(self.__dict__)
        del sd["sampler"]
        return sd
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        # Apply the step function
+        new_value = self._step()
+        # clip value to specified range
+        new_value_clipped = np.clip(new_value, a_min=self._min_val, a_max=self._max_val)
+        # Set the new value of the parameter dynamically
+        setattr(self.sampler, self.param_name, new_value_clipped)
+
+    def _step(self):
+        raise NotImplementedError
+
```
vmoens, line 74:
```suggestion
    @abstractmethod
    def _step(self):
        ...
```

vmoens: add `from abc import abstractmethod` at beginning of file

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+
+class LambdaScheduler(ParameterScheduler):
+    """Sets a parameter to its initial value times a given function.
+
+    Similar to torch.optim.LambdaLR.
```
vmoens, line 79:
```suggestion
    Similar to :class:`~torch.optim.LambdaLR`.
```
but is this the proper ref? shouldn't it be in `lr_schedulers` module or similar?

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+
+    Args:
+        obj (ReplayBuffer | Sampler): the replay buffer whose sampler to adjust (or the sampler itself)
```
vmoens, line 82:
```suggestion
        obj (ReplayBuffer or Sampler): the replay buffer whose sampler to adjust (or the sampler itself).
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        param_name (str): the name of the attribute to adjust, e.g. `beta` to adjust the
+            beta parameter
+        lambda_fn (function): A function which computes a multiplicative factor given an integer
+            parameter step_count
+        min_value (Union[int, float], optional): a lower bound for the parameter to be adjusted
+            Defaults to None.
+        max_value (Union[int, float], optional): an upper bound for the parameter to be adjusted
+            Defaults to None
```
vmoens, line 90:
```suggestion
        param_name (str): the name of the attribute to adjust, e.g. `beta` to adjust the
            beta parameter.
        lambda_fn (Callable[[int], float]): A function which computes a multiplicative factor given an integer
            parameter ``step_count``.
        min_value (Union[int, float], optional): a lower bound for the parameter to be adjusted
            Defaults to `None`.
        max_value (Union[int, float], optional): an upper bound for the parameter to be adjusted
            Defaults to `None`.
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+
+    """
+
+    def __init__(
+        self,
+        obj: ReplayBuffer | Sampler,
+        param_name: str,
+        lambda_fn: Callable[[int], float],
+        min_value: int | float = None,
+        max_value: int | float = None,
```
vmoens, line 100:
```suggestion
        min_value: int | float | None = None,
        max_value: int | float | None = None,
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+            Defaults to None
+
+    """
+
+    def __init__(
+        self,
+        obj: ReplayBuffer | Sampler,
+        param_name: str,
+        min_value: int | float = None,
+        max_value: int | float = None,
```
vmoens, line 29:
```suggestion
        min_value: int | float | None = None,
        max_value: int | float | None = None,
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        return self.initial_val * self.lambda_fn(self._step_cnt)
+
+
+class LinearScheduler(ParameterScheduler):
+    """A linear scheduler for gradually altering a parameter in an object over a given number of steps.
+
+    This scheduler linearly interpolates between the initial value of the parameter and a final target value.
+
+    Args:
+        obj (ReplayBuffer | Sampler): the replay buffer whose sampler to adjust (or the sampler itself)
```
vmoens, line 115:
```suggestion
        obj (ReplayBuffer or Sampler): the replay buffer whose sampler to adjust (or the sampler itself).
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        param_name (str): the name of the attribute to adjust, e.g. `beta` to adjust the
+            beta parameter
+        final_value (Union[int, float]): The final value that the parameter will reach after the
+            specified number of steps.
+        num_steps (Union[int, float], optional): The total number of steps over which the parameter
+            will be linearly altered.
```
vmoens, line 121:
```suggestion
        param_name (str): the name of the attribute to adjust, e.g. `beta` to adjust the
            beta parameter.
        final_value (number): The final value that the parameter will reach after the
            specified number of steps.
        num_steps (number, optional): The total number of steps over which the parameter
            will be linearly altered.
```

vmoens: I don't think `num_steps` is supposed to be marked as optional here

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        super().__init__(obj, param_name)
+        self.final_val = final_value
+        self.num_steps = num_steps
+        self._delta = (self.final_val - self.initial_val) / self.num_steps
+
+    def _step(self):
+        if self._step_cnt < self.num_steps:
+            return self.initial_val + (self._delta * self._step_cnt)
+        else:
+            return self.final_val
```
vmoens, line 153: nit: if we ever want this to be compilable without graph breaks we should think of a way to remove the control flow, eg
```python
return torch.where(self._step_cnt < self.num_steps, self.initial_val + (self._delta * self._step_cnt), self.final_val)
```
assuming that all of these are tensors

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        gamma (int | float, optional): The value by which to adjust the parameter,
+            either in a multiplicative or additive way
+        n_steps (int, optional): The number of steps after which the parameter should be altered.
+            Defaults to 1
+        mode (str, optional): The mode of scheduling. Can be either 'multiplicative' or 'additive'.
+            Defaults to 'multiplicative'
+        min_value (int | float, optional): a lower bound for the parameter to be adjusted
+            Defaults to None.
+        max_value (int | float, optional): an upper bound for the parameter to be adjusted
+            Defaults to None
```
vmoens, line 176:
```suggestion
        obj (ReplayBuffer or Sampler): the replay buffer whose sampler to adjust (or the sampler itself).
        param_name (str): the name of the attribute to adjust, e.g. `beta` to adjust the
            beta parameter.
        gamma (int or float, optional): The value by which to adjust the parameter,
            either in a multiplicative or additive way.
        n_steps (int, optional): The number of steps after which the parameter should be altered.
            Defaults to 1.
        mode (str, optional): The mode of scheduling. Can be either `'multiplicative'` or `'additive'`.
            Defaults to `'multiplicative'`.
        min_value (int or float, optional): a lower bound for the parameter to be adjusted.
            Defaults to `None`.
        max_value (int or float, optional): an upper bound for the parameter to be adjusted.
            Defaults to `None`.
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def __init__(
+        self,
+        obj: ReplayBuffer | Sampler,
+        param_name: str,
+        gamma: int | float = 0.9,
+        n_steps: int = 1,
+        mode: str = "multiplicative",
+        min_value: int | float = None,
+        max_value: int | float = None,
```
vmoens, line 201:
```suggestion
        min_value: int | float | None = None,
        max_value: int | float | None = None,
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        super().__init__(obj, param_name, min_value, max_value)
+        self.gamma = gamma
+        self.n_steps = n_steps
+        if mode == "additive":
+            operator = np.add
+        elif mode == "multiplicative":
+            operator = np.multiply
+        else:
+            raise ValueError(
+                f"Invalid mode: {self.mode}. Choose 'multiplicative' or 'additive'."
```
vmoens, line 213:
```suggestion
                f"Invalid mode: {mode}. Choose 'multiplicative' or 'additive'."
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        gamma: int | float = 0.9,
+        n_steps: int = 1,
+        mode: str = "multiplicative",
+        min_value: int | float = None,
+        max_value: int | float = None,
+    ):
+
+        super().__init__(obj, param_name, min_value, max_value)
+        self.gamma = gamma
+        self.n_steps = n_steps
```
vmoens, line 206:
```suggestion
        self.n_steps = n_steps
        self.mode = mode
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+        self.operator = operator
+
+    def _step(self):
+        """Applies the scheduling logic to alter the parameter value every `n_steps`."""
+        # Check if the current step count is a multiple of n_steps
+        current_val = getattr(self.sampler, self.param_name)
+        if self._step_cnt % self.n_steps == 0:
+            return self.operator(current_val, self.gamma)
+        else:
+            return current_val
```
vmoens, line 224: ditto

test/test_rb.py
```diff
@@ -3026,6 +3032,51 @@ def test_prioritized_slice_sampler_episodes(device):
     ), "after priority update, only episode 1 and 3 are expected to be sampled"
 
 
+def test_prioritized_parameter_scheduler():
+    INIT_ALPHA = 0.7
+    INIT_BETA = 0.6
+    GAMMA = 0.1
+    EVERY_N_STEPS = 10
+    LINEAR_STEPS = 100
+    TOTAL_STEPS = 200
```
vmoens, line 3041: let's maybe make these args to the func?

test/test_rb.py
```diff
@@ -3026,6 +3032,51 @@ def test_prioritized_slice_sampler_episodes(device):
...
+        gamma=GAMMA,
+        n_steps=EVERY_N_STEPS,
+        max_value=1.0,
+        mode="additive",
+    )
+    scheduler = SchedulerList(schedulers=(alpha_scheduler, beta_scheduler))
+    expected_alpha_vals = np.linspace(INIT_ALPHA, 0.0, num=LINEAR_STEPS + 1)
+    expected_alpha_vals = np.pad(
+        expected_alpha_vals, (0, TOTAL_STEPS - LINEAR_STEPS), constant_values=0.0
+    )
```
vmoens, line 3062: let's use torch here

test/test_rb.py
```diff
@@ -3026,6 +3032,51 @@ def test_prioritized_slice_sampler_episodes(device):
...
+    expected_beta_vals = (
+        np.atleast_2d(expected_beta_vals).repeat(EVERY_N_STEPS).clip(None, 1.0)
+    )
+    for i in range(TOTAL_STEPS):
+        assert np.isclose(
+            rb.sampler.alpha, expected_alpha_vals[i]
+        ), f"expected {expected_alpha_vals[i]}, got {rb.sampler.alpha}"
+        assert np.isclose(
+            rb.sampler.beta, expected_beta_vals[i]
+        ), f"expected {expected_beta_vals[i]}, got {rb.sampler.beta}"
```
vmoens, line 3075: let's use `torch.testing.assert_close`

vmoens (commented):
Amazing! Great and long awaited feature!
Thanks a mil

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+                from a call to :meth:`state_dict`.
+        """
+        self.__dict__.update(state_dict)
+
+    def step(self):
+        self._step_cnt += 1
+        # Apply the step function
+        new_value = self._step()
+        # clip value to specified range
+        new_value_clipped = np.clip(new_value, a_min=self._min_val, a_max=self._max_val)
```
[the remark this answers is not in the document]
LTluttmann: good catch. Imo there are two ways of handling this 
1) specifying a backend (`torch` or `numpy`) based on the type of the parameter
2) forcing `new_value` to be of type `float`, similar to as its done in torch's `ReduceLROnPlateau` [here](https://github.com/pytorch/pytorch/blob/dd4a51b39aa02cba23b3a387b41c5026770d9220/torch/optim/lr_scheduler.py#L1380). Might be more error prone but in general we would not expect tensors with multiple elements in the scheduler. 

What do you think?

torchrl/data/replay_buffers/scheduler.py
LTluttmann: interesting point. I sticked to the way torch schedulers handle different behavior for different epochs (e.g. [here](https://github.com/pytorch/pytorch/blob/dd4a51b39aa02cba23b3a387b41c5026770d9220/torch/optim/lr_scheduler.py#L479)). You think that is okay for now?

torchrl/data/replay_buffers/scheduler.py
vmoens: Oh wow! Ok then...

torchrl/data/replay_buffers/scheduler.py
vmoens: yeah that's fine, maybe let's add a comment to let someone know in the future that this should be fixed

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,263 @@
...
+        super().__init__(obj, param_name)
+        if isinstance(self.initial_val, torch.Tensor):
+            # cast to same type as initial value
+            final_value = torch.tensor(final_value).to(self.initial_val)
+        self.final_val = final_value
+        self.num_steps = num_steps
+        self._delta = (self.final_val - self.initial_val) / self.num_steps
+
+    def _step(self):
+        if self._step_cnt < self.num_steps:
```
vmoens, line 172:
```suggestion
        # Nit: we should use torch.where instead than if/else here to make the scheduler compatible with compile
        #  without graph breaks
        if self._step_cnt < self.num_steps:
```

torchrl/data/replay_buffers/scheduler.py
```diff
@@ -0,0 +1,263 @@
...
+            raise ValueError(
+                f"Invalid mode: {mode}. Choose 'multiplicative' or 'additive'."
+            )
+        self.operator = operator
+
+    def _step(self):
+        """Applies the scheduling logic to alter the parameter value every `n_steps`."""
+        # Check if the current step count is a multiple of n_steps
+        current_val = getattr(self.sampler, self.param_name)
+        if self._step_cnt % self.n_steps == 0:
```
vmoens, line 244:
```suggestion
        # Nit: we should use torch.where instead than if/else here to make the scheduler compatible with compile
        #  without graph breaks
        if self._step_cnt % self.n_steps == 0:
```

vmoens (commented):
LGTM thanks

vmoens (approved):
LGTM thanks

PR #2463, opened 2024-10-04
[BugFix] Fix Compose input spec transform

louisfaury:
## Description

This PR fixes what I think is a bug in the `transform_input_spec` method of the `Compose` transform. 

## Motivation and Context

The original code goes backward through the list of transforms:

```python
def transform_input_spec(self, input_spec: TensorSpec) -> TensorSpec:
    for t in self.transforms[::-1]:
        input_spec = t.transform_input_spec(input_spec)
    return input_spec
```

while I believe this should go forward to propagate the input spec of the underlying env.

```python
def transform_input_spec(self, input_spec: TensorSpec) -> TensorSpec:
    for t in self.transforms:
        input_spec = t.transform_input_spec(input_spec)
    return input_spec
```

I noticed this issue after trying to compose two inverse transformations: `DiscretiseAction` and `RenameTransform` -- which broke with a "key" error. I also tried to chain some `RenameTransform`s that rename the action key. I added this last examples to the tests.

_Rq_. I have seen that chaining >2 `RenameTransform` still yields some error, but this seems to be on the `RenameTransform` side, not `Compose`. I'll open an issue for this. 

## Types of changes

## Checklist

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -8678,8 +8678,18 @@ class TestTransforms:
     def test_compose_action_spec(self):
         # Create a Compose transform that renames "action" to "action_1" and then to "action_2"
         c = Compose(
-            RenameTransform(in_keys=(), out_keys=(), in_keys_inv=("action",), out_keys_inv=("action_1",)),
-            RenameTransform(in_keys=(), out_keys=(), in_keys_inv=("action_1",), out_keys_inv=("action_2",)),
+            RenameTransform(
+                in_keys=(),
+                out_keys=(),
+                in_keys_inv=("action",),
+                out_keys_inv=("action_1",),
+            ),
+            RenameTransform(
+                in_keys=(),
+                out_keys=(),
+                in_keys_inv=("action_1",),
+                out_keys_inv=("action_2",),
+            ),
         )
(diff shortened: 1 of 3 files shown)


vmoens:
The issue is broken because of a bug in tensordict lib - let me fix this and re-run the tests

louisfaury:
I see I forgot to run the linter on the test 🙈 I'll do that asap.

vmoens:
just did

test/test_transforms.py
```diff
@@ -8675,6 +8675,35 @@ def test_compose_indexing(self):
...
+
+        # Check the `full_action_spec`s
+        assert "action_2" in env.full_action_spec
+        # Ensure intermediate keys are no longer in the action spec
+        assert "action_1" not in env.full_action_spec
+        assert "action" not in env.full_action_spec
+
+        # Final check to ensure clean sampling from the action_spec
+        action = env.rand_action()
+        assert "action_2"
```
antoinebrl, line 8705: This assert is always true because the string is not empty. Instead, do you want to check if this key is present in the tensordict?
```python
        assert "action_2" in action
```

PR #2466, opened 2024-10-04
[Feature] Ensure transformation keys have the same number of elements

antoinebrl:
## Description

From the documentation and tests it is implicit that the number of in and out keys should match in the transformations. I'm proposing to enforce that through a runtime check by adding the `strict=True` when using zip. If the sequences don't have the same number of elements a `ValueError` is raised.

Let me know if you think enforcing this assumption requires some tests. Happy to take your suggestions if you have an idea on how to easily test all the calls to `zip`.

## Motivation and Context

This will ensure the keys are provided by pairs. Currently last elements of the longest list are ignored.

## Types of changes

Not sure if preventing silent errors should be considered a bug fix or a new feature to guide the users.

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -42,6 +42,7 @@ from tensordict import (
 from tensordict.nn import dispatch, TensorDictModuleBase
 from tensordict.utils import (
     _unravel_key_to_tuple,
+    _zip_strict,
     expand_as_right,
     expand_right,
     NestedKey,
@@ -88,7 +89,7 @@ def _apply_to_composite(function):
             _specs = observation_spec._specs
             in_keys = self.in_keys
             out_keys = self.out_keys
-            for in_key, out_key in zip(in_keys, out_keys, strict=True):
+            for in_key, out_key in _zip_strict(in_keys, out_keys):
                 if in_key in observation_spec.keys(True, True):
                     _specs[out_key] = function(self, observation_spec[in_key].clone())
             return Composite(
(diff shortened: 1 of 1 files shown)


torchrl/envs/transforms/transforms.py
```diff
@@ -88,7 +88,7 @@ def new_fun(self, observation_spec):
             _specs = observation_spec._specs
             in_keys = self.in_keys
             out_keys = self.out_keys
-            for in_key, out_key in zip(in_keys, out_keys):
+            for in_key, out_key in zip(in_keys, out_keys, strict=True):
```
vmoens, line 91: We need to be compatible with python 3.8. PyTorch and TensorDict have a zip_strict that falls back onto zip(..., strict=True) for appropriate python versions

vmoens (commented):
It's a great feature! Just need to make it work with earlier versions of python

torchrl/envs/transforms/transforms.py
antoinebrl: Thanks for the pointer, I will make use of it!
Py3.8 is reaching end of life at the end of the month: https://devguide.python.org/versions/. What is the plan for supporting or dropping this version in the next torch and torchrl releases?

vmoens: I think this is a 3.10 feature so we have still a good 6 months until we can fully deprecate zip_strict

antoinebrl: I saw there was already some imports of private functions from `tensordict` so I did it too for `_zip_strict`.

vmoens (approved):
LGTM thanks!

PR #2469, opened 2024-10-07
[Feature] Add env wrapper for Unity MLAgents

### LINKED ISSUE #1110: [Feature Request] Unity MLAgents Support

kurtamohler:
## Description

Adds `UnityMLAgentsWrapper` and `UnityMLAgentsEnv` environment wrapper classes to control MLAgents environments from TorchRL.

## Motivation and Context

close #1110

- [x] Someone has raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

## Checklist

### DIFF
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -5,6 +5,7 @@
 import functools
 import gc
 import importlib.util
+import urllib.error
 
 _has_isaac = importlib.util.find_spec("isaacgym") is not None
 
@@ -18,10 +19,12 @@ import importlib
 import os
 
 import time
+import urllib
 from contextlib import nullcontext
 from pathlib import Path
 from sys import platform
 from typing import Optional, Union
+from unittest import mock
 
 import numpy as np
 import pytest
@@ -36,6 +39,7 @@ from _utils_internal import (
     PENDULUM_VERSIONED,
     PONG_VERSIONED,
     rand_reset,
+    retry,
     rollout_consistency_assertion,
 )
 from packaging import version
@@ -111,6 +115,11 @@ from torchrl.envs.libs.openspiel import _has_pyspiel, OpenSpielEnv, OpenSpielWra
 from torchrl.envs.libs.pettingzoo import _has_pettingzoo, PettingZooEnv
 from torchrl.envs.libs.robohive import _has_robohive, RoboHiveEnv
 from torchrl.envs.libs.smacv2 import _has_smacv2, SMACv2Env
+from torchrl.envs.libs.unity_mlagents import (
+    _has_unity_mlagents,
+    UnityMLAgentsEnv,
+    UnityMLAgentsWrapper,
+)
 from torchrl.envs.libs.vmas import _has_vmas, VmasEnv, VmasWrapper
 
 from torchrl.envs.transforms import ActionMask, TransformedEnv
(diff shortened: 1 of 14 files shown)


torchrl/envs/libs/unity_mlagents.py
```diff
@@ -0,0 +1,859 @@
...
+
+    Attributes:
+        available_envs: list of registered environments available to build
+
+    Examples:
+        >>> from torchrl.envs import UnityMLAgentsEnv
+        >>> env = UnityMLAgentsEnv()
+        >>> td = env.reset()
+        >>> td = env.step(td.update(env.full_action_spec.rand()))
+
```
vmoens, line 527:
```suggestion
```

test/test_libs.py
```diff
@@ -3930,6 +3939,127 @@ def test_chance_not_implemented(self):
...
+        try:
+            env.rollout(
+                max_steps=500, break_when_any_done=False, break_when_all_done=False
+            )
+        finally:
+            env.close()
+
+    @pytest.mark.unity_editor
+    def test_with_editor(self):
+        print("Please press play in the Unity editor")  # noqa: T201
```
vmoens, line 3999: what's this?

kurtamohler: When `UnityMLAgentsEnv` is not given a path to a file or the name of a registered env, it waits for the user to press the play button in the unity editor or run a mlagents-compatible binary. Since we can't easily do that in CI, I figured we can just skip the test by default. To run the test, the cmd line arg `--unity_editor` has to be given. We could remove this test, but I found it useful while working on this PR to have a test that connects to whichever environment I have open in the unity editor

vmoens: gotcha!

kurtamohler:
> Can you share a bit about the current limitations (eg, are the specs expected to be fixed?) and some context about the tests - I see a couple of mocks and pytest marks

Current limitations are:
* The specs are expected to be fixed throughout the run, so no new agents, behaviors, or observations can be added during the run.
* All agents are expected to request a decision in the first step so that they all get included when the specs are built.
* Because of the previous two points, the "Basic" and "DungeonEscape" environments from the example environments included with MLAgents do not work with our wrapper classes. But the ~15 other example envs do work.
* I haven't added the `group_map` argument to the wrappers.


For the tests:
* MLAgents comes with a few built-in environment binaries that can be automatically downloaded and run, so we have a test that uses those. For now, I only enabled testing two of those environments because each download is 180 MB. But thankfully they just get downloaded to `/tmp`, which is cleared on shutdown. We could add more of them to the list.
* The mock tests are modeled after the [tests in the mlagents repo](https://github.com/Unity-Technologies/ml-agents/blob/ac576f9f6d2cb5e7c5d078bd6aea497e6dab4043/ml-agents-envs/tests/test_envs.py#L94). I figured if that's how they test their environment class, we might as well do it that way too. Also, the registered envs are still an "experimental" feature, so I thought maybe we shouldn't only rely on those.

PR #2473, opened 2024-10-09
[Feature] Gymnasium 1.0 compatibility

vmoens:
* __->__ #2473

### DIFF
--- a/.github/unittest/linux_libs/scripts_gym/batch_scripts.sh
+++ b/.github/unittest/linux_libs/scripts_gym/batch_scripts.sh
@@ -118,7 +118,7 @@ do
 done
 
 # For this version "gym[accept-rom-license]" is required.
-for GYM_VERSION in '0.27' '0.28'
+for GYM_VERSION in '0.27' '0.28' '0.29'
 do
   # Create a copy of the conda env and work with this
   conda deactivate
@@ -140,7 +140,7 @@ conda deactivate
 conda create --prefix ./cloned_env --clone ./env -y
 conda activate ./cloned_env
 
-pip3 install 'gymnasium[accept-rom-license,ale-py,atari]<1.0' mo-gymnasium gymnasium-robotics -U
+pip3 install 'gymnasium[accept-rom-license,ale-py,atari]' mo-gymnasium gymnasium-robotics -U
 
 $DIR/run_test.sh
 
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -2133,7 +2133,9 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         if autoreset is not None:
             raise TypeError("autoreset is only compatible with gymnasium<1.0.")
         if apply_api_compatibility is not None:
-            raise TypeError("apply_api_compatibility is only compatible with gymnasium<1.0.")
+            raise TypeError(
+                "apply_api_compatibility is only compatible with gymnasium<1.0."
+            )
         return gymnasium.register(
             id=id,
             entry_point=entry_point,
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -1622,9 +1622,9 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         nondeterministic: bool = False,
         max_episode_steps: int | None = None,
         order_enforce: bool = True,
-        autoreset: bool = False,
+        autoreset: bool | None = None,
         disable_env_checker: bool = False,
-        apply_api_compatibility: bool = False,
+        apply_api_compatibility: bool | None = None,
         **kwargs,
     ):
         """Registers an environment in gym(nasium).
@@ -1811,9 +1811,9 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         nondeterministic: bool = False,
         max_episode_steps: int | None = None,
         order_enforce: bool = True,
-        autoreset: bool = False,
+        autoreset: bool | None = None,
         disable_env_checker: bool = False,
-        apply_api_compatibility: bool = False,
+        apply_api_compatibility: bool | None = None,
         **kwargs,
     ):
         import gym
@@ -1836,9 +1836,9 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
             nondeterministic=nondeterministic,
             max_episode_steps=max_episode_steps,
             order_enforce=order_enforce,
-            autoreset=autoreset,
+            autoreset=bool(autoreset),
             disable_env_checker=disable_env_checker,
-            apply_api_compatibility=apply_api_compatibility,
+            apply_api_compatibility=bool(apply_api_compatibility),
         )
 
     @implement_for("gym", "0.25", "0.26", class_method=True)
@@ -1853,14 +1853,14 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         nondeterministic: bool = False,
         max_episode_steps: int | None = None,
         order_enforce: bool = True,
-        autoreset: bool = False,
+        autoreset: bool | None = None,
         disable_env_checker: bool = False,
-        apply_api_compatibility: bool = False,
+        apply_api_compatibility: bool | None = None,
         **kwargs,
     ):
         import gym
 
-        if apply_api_compatibility is not False:
+        if apply_api_compatibility is not None:
             raise TypeError(
                 cls._GYM_UNRECOGNIZED_KWARG.format(
                     "apply_api_compatibility", gym.__version__
@@ -1901,14 +1901,14 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         nondeterministic: bool = False,
         max_episode_steps: int | None = None,
         order_enforce: bool = True,
-        autoreset: bool = False,
+        autoreset: bool | None = None,
         disable_env_checker: bool = False,
-        apply_api_compatibility: bool = False,
+        apply_api_compatibility: bool | None = None,
         **kwargs,
     ):
         import gym
 
-        if apply_api_compatibility is not False:
+        if apply_api_compatibility is not None:
             raise TypeError(
                 cls._GYM_UNRECOGNIZED_KWARG.format(
                     "apply_api_compatibility", gym.__version__
@@ -1954,14 +1954,14 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         nondeterministic: bool = False,
         max_episode_steps: int | None = None,
         order_enforce: bool = True,
-        autoreset: bool = False,
+        autoreset: bool | None = None,
         disable_env_checker: bool = False,
-        apply_api_compatibility: bool = False,
+        apply_api_compatibility: bool | None = None,
         **kwargs,
     ):
         import gym
 
-        if apply_api_compatibility is not False:
+        if apply_api_compatibility is not None:
             raise TypeError(
                 cls._GYM_UNRECOGNIZED_KWARG.format(
                     "apply_api_compatibility", gym.__version__
@@ -1973,7 +1973,7 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
                     "disable_env_checker", gym.__version__
                 )
             )
-        if autoreset is not False:
+        if autoreset is not None:
             raise TypeError(
                 cls._GYM_UNRECOGNIZED_KWARG.format("autoreset", gym.__version__)
             )
@@ -2010,9 +2010,9 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         nondeterministic: bool = False,
         max_episode_steps: int | None = None,
         order_enforce: bool = True,
-        autoreset: bool = False,
+        autoreset: bool | None = None,
         disable_env_checker: bool = False,
-        apply_api_compatibility: bool = False,
+        apply_api_compatibility: bool | None = None,
         **kwargs,
     ):
         import gym
@@ -2028,11 +2028,11 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
                     "disable_env_checker", gym.__version__
                 )
             )
-        if autoreset is not False:
+        if autoreset is not None:
             raise TypeError(
                 cls._GYM_UNRECOGNIZED_KWARG.format("autoreset", gym.__version__)
             )
-        if apply_api_compatibility is not False:
+        if apply_api_compatibility is not None:
             raise TypeError(
                 cls._GYM_UNRECOGNIZED_KWARG.format(
                     "apply_api_compatibility", gym.__version__
@@ -2056,7 +2056,7 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
             max_episode_steps=max_episode_steps,
         )
 
-    @implement_for("gymnasium", class_method=True)
+    @implement_for("gymnasium", None, "1.0", class_method=True)
     def _register_gym(  # noqa: F811
         cls,
         id,
@@ -2068,9 +2068,9 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         nondeterministic: bool = False,
         max_episode_steps: int | None = None,
         order_enforce: bool = True,
-        autoreset: bool = False,
+        autoreset: bool | None = None,
         disable_env_checker: bool = False,
-        apply_api_compatibility: bool = False,
+        apply_api_compatibility: bool | None = None,
         **kwargs,
     ):
         import gymnasium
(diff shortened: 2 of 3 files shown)


vmoens:
@pseudo-rnd-thoughts @jkterry1 I would need some guidance here for the partial resets.

If I understand correctly, now if an env is done, the `step` returns the next `obs` and when the `step` will be queried next time the action will be ignored and the reset `obs` will be returned.
In torchrl, what we do is (simplified with a single env)
```
def step_and_maybe_reset(self, data):
    next_obs, rew, term, trunc, info = self._env.step(data["action"])
    if term | trunc:
        reset_obs = self._env.reset()
    else:
        reset_obs = next_obs
    data.set("next", TensorDict(obs=next_obs, reward=rew, ...)
    maybe_reset = data.empty().set(obs=reset_obs, ...)
    return data, maybe_reset
```
I'm trying to think about how I could adapt `step_and_maybe_reset` but it isn't easy since I should be calling `self._env.step` with a new action to get `reset_obs`.

I cannot think of any way to make `env_and_maybe_reset` behave correctly since I need the action at t+1 (for envs that should not reset) - am I correct?

Other more metaphysical questions:
- what effect does the action have when passed to step when step means reset?
- Assuming the answer is "it has no effect": If it's expensive to compute the action and many envs are resetting (or envs are resetting frequently), isn't this a loss of resources?
- What values will reward, termination and truncation have when they come from a reset?
- In the example below, we're waiting one step to read what `autoreset` is doing
  ```python
  replay_buffer = []
  obs, _ = envs.reset()
  autoreset = np.zeros(envs.num_envs)
  for _ in range(total_timesteps):
      next_obs, rewards, terminations, truncations, _ = envs.step(envs.action_space.sample())

      for j in range(envs.num_envs):
          if not autoreset[j]:
              replay_buffer.append((
                  obs[j], rewards[j], terminations[j], truncations[j], next_obs[j]
              ))

      obs = next_obs
      autoreset = np.logical_or(terminations, truncations)
  ```
  Am I correct to think that previously, if I was looping N times I would get N steps, now I get anything from N//2 to N steps? Isn't that a problem if I'm designing a training loop where I'm giving myself a budget of `N` steps?
- I now see some control flow appearing in the collection loop. Isn't the whole point of vectorized envs to get rid of that? You could tell me that this is a simplified version and you could mask the obs / rewards etc I guess... but why not adopting a loop like this, where you can probably let users just extend their buffers with the peace of mind of not having to copy some error-prone boilerplate code
  ```python
  for _ in range(N):
      action = policy(obs)
      next_obs, maybe_reset_obs, rewards, term, trunc, info = env.step(action)
      buffer.extend((obs, action, next_obs, rewards, term, trunc, info))
      obs = maybe_reset_obs
  ```
  
Thanks!

pseudo-rnd-thoughts:
@vmoens Thanks for your message, I'll try to answer the best I can

> what effect does the action have when passed to step when step means reset?

The action is discarded or "has no effect"

> Assuming the answer is "it has no effect": If it's expensive to compute the action and many envs are resetting (or envs are resetting frequently), isn't this a loss of resources?

This adds an extra step to vector environments when resetting so this cost is dependent on the reset frequency. Yes, this is a loss of resources, however, in my opinion, very minimally. 

> What values will reward, termination and truncation have when they come from a reset?

The reward will be zero, termination and truncation is False. 

> Isn't that a problem if I'm designing a training loop where I'm giving myself a budget of `N` steps?

Possibly, if you reset every step, this reduces your practical budget to `N//2` however practically, I would imagine this to be very rare. How you would specify that you are precisely learning on `N` steps is less clear, admittedly, now.

> control flow

You are correct, but the boiler plate is assuming that users have a single environment buffer thus each sub-environment data must be extracted. For vectorised environment buffers, this control is unnecessary but it is critical that you are not using the end of one episode and the begining of another in your training data (or it must be masked out).

@RedTachyon and I are planning on releasing a blog post about this more because we understand the complexity of it for projects to adapt

vmoens:
What about an approach that does not need a blog post to adapt? :D 
Like the one I proposed above where you always get a `maybe_reset_obs`?
that would not consume more mem or compute than what you have now I think, and it'd be easier to integrate

vmoens:
@pseudo-rnd-thoughts you don't have partial resets right? Smth like `reset(mask)`?

matteobettini:
@pseudo-rnd-thoughts I'll drop here my 2 cents as the creator of VMAS which is a vectorized env

It seems to me that the new API for autoresets in `step` is quite problematic from multiple points of view:

1. it will cause significant increase in the number of `step` calls, cause it will have to be called once more for every reset
2. it will cause significant increase in the number of calls to the policy, which has to be called even if 99% of the envs are done just to provide the valid 1% actions (keep in mind that the policy could be an expensive LLM and we are through away up to 99% of the outpout)

I think the impact of this is bigger than we can imagine. I work with batches of hunderds of thousands of envs on GPU and it is very probable that in one step at least one env will be done.

I don't know why this is the route that was taken, but I don't think it will be sustainable.

The solution is really simple in my opinion and it is to keep `step` just doing steps on all envs and have something like `env.reset(done)` which resets using a mask (or index or anything)

vmoens:
Since it's an API that is also used by envpool and others we could think of a way to circumvent this when calling step_and_maybe_reset but as of now I think the most immediate solution is going to be to tell people not to use gymnasium 1.0 with torchrl until a fix is found...
 
I agree with Matteo that having a method call `step` that under the hood ignores the input, provides garbage output and actually does something different than `step` looks like a misstep. That being said, I assume it's a decision from the gymnasium community of users in which case we can only acknowledge it and work around that...

pseudo-rnd-thoughts:
@vmoens and @matteobettini Thanks for your input, I'm happy to have a zoom call to work out a solution if that helps. 

To clarify, the previous autoreset (reset within the same step) is still compatible with `gymnasium.vector.VectorEnv`, it is just that the gymnasium vector wrappers will not. 

> you don't have partial resets right? Smth like reset(mask)?

Normally a vector environment from Gymnasium, you only call reset at the beginning with users calling `step` which will autoreset sub-environment that terminate or truncate so it wouldn't make sense to have a `reset(mask)` under this API if I understand correctly. 

@matteobettini I admittedly didn't test this across a wide range of environments, but I didn't expect that adding an additional call to a policy would cause significant problems as most environments take a few hundred steps at minimum before termination, let alone truncation. Meaning that I expected this to cause only a minimal impact on performance.  

> having a method call step that under the hood ignores the input, provides garbage output and actually does something different than step looks like a misstep.

The problem as I see it is that an autoreset must happen under the hood to reset sub-environments, either within the same step as a termination / truncation or on the next step.

matteobettini:
Thanks for the reply, I am available if we set up a call

> The problem as I see it is that an autoreset must happen under the hood to reset sub-environments, either within the same step as a termination / truncation or on the next step.

Yes I think the core problem discussed here is that doing it in the same step is a better solution as distributing it across 2 steps has the problems highlighted above.

In your example of the 200 steps. If you have 200 enviornomnets terminating in 200 steps and they are perfectly out of sync, you will be doing a reset at every step

More in general regarding autoreset, I am a bit constrained by it. I understand that in gym this is how things are done currently, but for example in my use case I have a VMAS env (which has no autoreset) and I would like to wrap it with a gymnasium VectorEnv and I can't because gymansium assumes that everyone will use autoresets, while I would prefer the general API not to prescribe if you do autoresets or not (and thus provide have a reset function that takes an optional mask).
Users can then implement it using autoresets and `env.reset` or using normal steps and `env.step(done)`. **This is more of  my own struggle with the API and not super related to the problem discussed here, although it is a solution for it**

vmoens:
@pseudo-rnd-thoughts is there any way one could implement a `reset(done)` like @matteobettini was suggesting, or anything where we don't have to deal with garbage data?

pseudo-rnd-thoughts:
The vector API doesn't require autoreset to be implemented but it has been designed with this in mind. 
You could pass this data through `reset(options={"mask": ...})` as well.

matteobettini:
I think I spotted anotehr criticality related to this.

The API is assuming that it ispossible to step the sub envs independently (which is not true in batched simulators as it would defeat the purpose)

Here is the example:

- I have 4 envs: [0,1,2,3]
- i called step and 2 and 3 are done 
- now when is call step again, 2 and 3 will be reset 
- but then if my simulator uses a batch of envs (and not a list of processes), I cannot step dimensions 0 and 1 without stepping alo 2 and 3
- so my step will step all envs and I will have to consider the actions and the new data for 2 and 3 as well

So for all enviornments that cannot step subenvs independently the new API cannot work

RedTachyon:
I might address some other points later (discussing with Vincent "offline" in parallel), but one thing I want a sanity check on regarding the last message:
> So for all environemnts that cannot step subenvs independently the new API cannot work

Wouldn't the same be true for the old API? (which also featured an autoreset, just shifted by one step)

matteobettini:
> Wouldn't the same be true for the old API? (which also featured an autoreset, just shifted by one step)

No, the old API should be fine, because it did (correct me if i am wrong):

When step is called:
- step all envs
- reset the done ones
- store the data from the step and reset in 2 different places and output both

So it always has a global step

pseudo-rnd-thoughts:
@matteobettini I don't know how your problem is implemented but can you not pass a no-op like action to the simulator? 
Otherwise, it would seem that you can still use Gymnasium v1.0 but your vector environments use a different autoreset API than default. I can look to see what we can do to support both API's, but there are limitations.

matteobettini:
> @matteobettini I don't know how your problem is implemented but can you not pass a no-op like action to the simulator? Otherwise, it would seem that you can still use Gymnasium v1.0 but your vector environments use a different autoreset API than default. I can look to see what we can do to support both API's, but there are limitations.

I am not talking about my problem here, I am talking about any batched simulation framework (isacc, vmas,brax...)
the simulator step is some function of the imput $o'=f(\pi(o))$, where $o,o'$ have n_envs as the leading dimensions.

the core point of batching and vectorization in RL is that the functions $f$ and $\pi$ can run independently of n_envs and scaling in n_envs makes them more efficient.

What the new API implies is that you will have to compute $\pi$ and $f$ over partial data (partial steps and partial actions when a reset happens) .

This is a major problem because this API is essentially eliminating all the benefits of batching and vectorization
- the benefits of batching in $\pi$ (which could perviously be achieved also without a batched env) are undone because soem actions will be ignored
- the benefits of batching in $f$ (which were enabled by gpu accellerated envs) are undone beacuse you require partial steps

RedTachyon:
I'd really, really like to ask people to turn down the temperature and stop overexaggerating. 

> This is a major problem because this API is essentially eliminating all the benefits of batching and vectorization

This is just not true. I know this for a fact, because I'm working on a project that uses the new API, that benefits from batching and vectorization. Beyond that, Gymnasium ships with some hardware-accelerated Jax-based environments. 

There's probably a more targeted argument that you can make - in which, case, please make that argument instead of saying false things, because there's no point for us to engage with that.

matteobettini:
I did not want to step up the tone, sorry for that, I am just trying to bring this issue to light.

Let me try to prove that there is a case where you would loose all benefits of batching with a worst-case scenario study.

Assume you have $N$ enviornments.

The worst case scenarios happens when all envrionments are done except 1 at every step.

You would then call the policy at every iteration with N obs, but then N-1 are discarded.

Then your step function would call the physical step with 1 action and N-1 no-op actions.

This is why i believe my statement is true.

I know that this is a worst-case scenario, but computer science often studies complexity using worst-case analysis.

I understand that empirically you use jax and are doing fine, but I am just trying to bring out the theoretical implications here.

pseudo-rnd-thoughts:
Maybe I'm too much of an empiricist, but to me, this worst-case analysis should be a minimal problem in practice, so I didn't take it much into consideration. If you find this significantly increases compute time, please let me know. 

> any paper using gymnasium>=1.0 and saying "we trained this algorithm for 1M steps" is lying to you

@vmoens Ironically, the opposite is true (to my knowledge) with the no-op action for terminated / truncated states; then 1 million steps is an upper bound, not the lower bound. In reality, they are probably taking fewer steps than that of the actual environments unless they try to be fancy then yes, they could mess it up but that is their fault not gymnasium's imho.

vmoens:
@RedTachyon I agree let's lower the tone! At the end of the day what we want it to support the community better.

Let me first try to explain why it is so difficult for us to adapt to these changes:

Concretely speaking, this change is impossible to incorporate in torchrl without some heavy, painful and probably bc-breaking refactoring. To put it simply: `env.step_and_maybe_reset` is our main entry point for doing a step and reset combined. It looks at the done states and, if needed runs a reset in the env. If the env is auto-resetting, it fetches the final obs somewhere (previously in info dict) and stores it at it appropriate location. This function returns two objects: one which is the input data updated with the content of the current traj, and another that is either the next step base data or the reset data. You can use the second object to feed your policy.

Our goal is to support the RL community, whether they use gym or something else and here the cost seems too high to pay for the general community. I understand gym wants to be the one default backend for simulation in RL and related fields, but from where we stand we can't ignore other frameworks. 
Unfortunately, any solution that requires users to run the policy on observations that are out of the trajectory and throw away the actions manually isn't compatible with our current API and probably won't be. It would introduce some heavy control flow in training scripts / trainers / buffers and make the library less modular (because one component like the buffer would depend on the backend of the env, which is surprising).
With the previous API, we could fit the gym vectorized envs within a wrapper and all the actions were always valid, so there was no need for the users to worry about the backend they were using. 
We want to allow people to code environments directly with TorchRL: the user base is not as big as Gym, I admit, and I also understand this should not be a worry for you that your API matches anyone else's in any way. But as it is now, the change is too big and too abrupt to be incorporated and, as you guys, we have a very limited bandwidth to refactor things within the library. 
I hope that makes sense!

To understand why we don't encounter the problem of reset vs final obs in torchrl, I invite you to read about torchrl format [here](https://pytorch.org/rl/stable/reference/data.html#torchrl-episode-data-format-ted). I hope this will help you understand where we come from!

To move forward:

- Is there a way to detect whether an env is autoresetting with the new API? If so we could just error when such env is detected by `GymWrapper` and not in other cases. We wouldn't want to have to just prohibit users from using gymnasiun 1.0 across the lib!
- If we can help in the design of new gym wrappers or methods that permit partial reset or a fallback on the previous API (which we did not like much but at least managed to make work) we'd be happy to dedicate time doing that.

matteobettini:
> Maybe I'm too much of an empiricist, but to me, this worst-case analysis should be a minimal problem in practice, so I didn't take it much into consideration. If you find this significantly increases compute time, please let me know.

I believe it could show up in practical examples. Imagine a self-driving simulator where you have 4 actions for driving and one for a hand-break. Probably, pulling the hand-break at the wrong time would lead to a crash. I believe that a scenario like this would come close to what I outlined as a lot of sub envs would be frequently done during exploration and you would see empirically the difference.

I guess the point I am making is that the cost of the previous API did not depend on the task, and now it does, with the worst case being what i outlined and a scenario like the above empirically approaching that

vmoens:
@pseudo-rnd-thoughts 

Let me rephrase and expand a bit on the accuracy point. Here's another simple example of why I think this is an issue:

CartPole (arguably the most popular and simple env ever) has a reset frequency tightly linked to the perf of the algo.
Say I allocate myself a 1M steps budget for training. Now take algo 1 which is not good, giving me many many resets, and algo 2 much more efficient, giving me very few resets.

With gym 1.0, algo 1 will have fewer total steps so probably an even poorer perf. The gap between 1 and 2 will be bigger than it should.
Therefore smth irrelevant (simulator API) will impact the measure of how good an algo is. 

Note that the effect is opposite if the env resets more frequently when it's solved (imagine getting out of a maze the fastest possible => more success = more resets = less training data).

Does that make sense?

> To move forward:
>
> - Is there a way to detect whether an env is autoresetting with the new API? If so we could just error when such env is detected by GymWrapper and not in other cases. We wouldn't want to have to just prohibit users from using gymnasiun 1.0 across the lib!
> - If we can help in the design of new gym wrappers or methods that permit partial reset or a fallback on the previous API (which we did not like much but at least managed to make work) we'd be happy to dedicate time doing that.

Any thoughts on this?

pseudo-rnd-thoughts:
@vmoens I've spent some time investigating adding other autoreset APIs to Gymnasium's built-in environments: https://github.com/Farama-Foundation/Gymnasium/pull/1227. This would allow users to specify the autoreset mode of vector environments and select the autoreset mode for the built-in Sync and Async vector environments. Please test this branch with TorchRL and let me know if there are any problems with it or if there are changes to the API that you would like implemented. 

> Is there a way to detect whether an env is autoresetting with the new API? If so we could just error when such env is detected by GymWrapper and not in other cases. We wouldn't want to have to just prohibit users from using gymnasiun 1.0 across the lib!

This solution should solve this question.  

> If we can help in the design of new gym wrappers or methods that permit partial reset or a fallback on the previous API (which we did not like much but at least managed to make work) we'd be happy to dedicate time doing that.

This has always been true. We are always interested in new proposals and contributions from the community. I want to note that this API change was a key part of an alpha release from February of this year (https://github.com/Farama-Foundation/Gymnasium/releases/tag/v1.0.0a1), 8 months ago. This was not a new change.

vmoens:
@pseudo-rnd-thoughts amazing, looking forward to this!

AlexandreBrown:
@pseudo-rnd-thoughts @RedTachyon @vmoens @matteobettini

As a member of both the Gymnasium and TorchRL communities, I want to thank all of you for working together on this. It’s great to see this kind of collaboration. With many environments and libraries now adjusting to dependency limits like `gymnasium < 1.0.0` (e.g., [ManiSkill2 real2sim](https://github.com/simpler-env/ManiSkill2_real2sim/commit/ef7a4d4fdf4b69f2c2154db5b15b9ac8dfe10682)), this effort is more important than ever.

I really hope this collaboration succeeds, as it would help the whole community. Keep up the good work!

pseudo-rnd-thoughts:
Apologies, it took a very long time to cut a new release, but I just released Gymnasium v1.1, which supports the different autoreset modes with the vector implementations supporting all three. 
See https://github.com/Farama-Foundation/Gymnasium/releases/tag/v1.1.0 and https://farama.org/Vector-Autoreset-Mode for more information. 
I'm looking forward to PyTorch RL supporting Gymnasium, let me know if I can help if anything

vmoens:
Thanks @pseudo-rnd-thoughts this is wonderful news! 
I really appreciate the effort you put into solving that reset issue, looking forward to reintegrate gymnasium within the library!!

JonDum:
Now that gymnasium v1.1.0 has been released, may I ask what the suggested path is for users of gymnasium envs with torchrl? If there are any other blockers maintainers see I'd gladly contribute some time to get this PR merged and cut into a release.

I have been using the same gymnasium 1.0 custom env with RLlib, SKRL, and SB3 without issue. I'd love to run some experiments with TorchRL but so far TorchRL has been the major outlier in that it has been the only framework that has not been easy or straight forward to get running.

vmoens:
Hello @JonDum 
I haven't had the time to patch torchrl yet unfortunately... I will give it a go asap!

vmoens:
@JonDum here is the PR https://github.com/pytorch/rl/pull/2898

Issue #2501, opened 2024-10-18
[Feature Request] Support compiling `ReplayBuffer.extend` without recompile


kurtamohler:
## Motivation

Compiling `ReplayBuffer.extend` and then calling it multiple times can cause the function to be recompiled.

```python
import torch
import torchrl

torch._logging.set_logs(recompiles=True)

rb = torchrl.data.ReplayBuffer(
    storage=torchrl.data.LazyTensorStorage(1000)
)

@torch.compile
def extend(data):
    rb.extend(data)

for idx in range(4):
    print('---------------------')
    print(f'{idx}:')
    extend(torch.randn(idx + 1, 1))
```

Running the above script gives the following, showing that the first two calls cause recompilation.

```
---------------------
0:
/home/endoplasm/miniconda/envs/torchrl-1/lib/python3.11/site-packages/torch/_dynamo/variables/functions.py:733: UserWarning: Graph break due to unsupported builtin None.SemLock.acquire. This function is either a Python builtin (e.g. _warnings.warn) or a third-party C/C++ Python extension (perhaps created with pybind). If it is a Python builtin, please file an issue on GitHub so the PyTorch team can add support for it and see the next case for a workaround. If it is a third-party C/C++ Python extension, please either wrap it into a PyTorch-understood custom operator (see https://pytorch.org/tutorials/advanced/custom_ops_landing_page.html for more details) or, if it is traceable, use torch.compiler.allow_in_graph.
  torch._dynamo.utils.warn_once(msg)
V1017 17:56:51.958000 523255 site-packages/torch/_dynamo/guards.py:2842] [11/1] [__recompiles] Recompiling function _lazy_call_fn in /home/endoplasm/develop/torchrl-1/torchrl/_utils.py:389
V1017 17:56:51.958000 523255 site-packages/torch/_dynamo/guards.py:2842] [11/1] [__recompiles]     triggered by the following guard failure(s):
V1017 17:56:51.958000 523255 site-packages/torch/_dynamo/guards.py:2842] [11/1] [__recompiles]     - 11/0: L['self'].func_name == 'torchrl.data.replay_buffers.storages.TensorStorage.set'
V1017 17:56:51.965000 523255 site-packages/torch/_dynamo/guards.py:2842] [12/1] [__recompiles] Recompiling function torch_dynamo_resume_in__lazy_call_fn_at_394 in /home/endoplasm/develop/torchrl-1/torchrl/_utils.py:394
[...]

PR #2504, opened 2024-10-18
[Feature] Avoid some recompiles of `ReplayBuffer.extend\sample`

kurtamohler:
Part of #2501

* __->__ #2504


torchrl/data/replay_buffers/storages.py
```diff
@@ -149,7 +149,7 @@ def _rand_given_ndim(self, batch_size):
         if self.ndim == 1:
             return torch.randint(
                 0,
-                len(self),
+                self._len,
```
kurtamohler, line 152: For some reason, the `len(self)` call here was causing recompiles. Changing it to `self._len` avoids the recompiles, even though `TensorStorage.__len__` just returns `self._len` anyway

kurtamohler:
FWIW, I looked into disabling the multiprocessing like you suggested. I tried temporarily changing `TensorStorage._len_value` to be a normal int rather than `mp.Value` with the below diff, but that did not fix the recompiles.

<details><summary>Click to expand</summary>

```
diff --git a/torchrl/data/replay_buffers/storages.py b/torchrl/data/replay_buffers/storages.py
index 217229b5..8b588006 100644
--- a/torchrl/data/replay_buffers/storages.py
+++ b/torchrl/data/replay_buffers/storages.py
@@ -458,15 +458,12 @@ class TensorStorage(Storage):
     def _len(self):
         _len_value = self.__dict__.get("_len_value", None)
         if _len_value is None:
-            _len_value = self._len_value = mp.Value("i", 0)
-        return _len_value.value
+            _len_value = self._len_value = 0
+        return _len_value
 
     @_len.setter
     def _len(self, value):
-        _len_value = self.__dict__.get("_len_value", None)
-        if _len_value is None:
-            _len_value = self._len_value = mp.Value("i", 0)
-        _len_value.value = value
+        self._len_value = value
 
     @property
     def _total_shape(self):
@@ -606,7 +603,7 @@ class TensorStorage(Storage):
     def __setstate__(self, state):
         len = state.pop("len__context", None)
         if len is not None:
-            _len_value = mp.Value("i", len)
+            _len_value = len
             state["_len_value"] = _len_value
         self.__dict__.update(state)
```

</details>

With that change, I was still getting recompile records like this:

```
V1023 12:03:57.543000 648354 site-packages/torch/_dynamo/guards.py:2842] [26/2] [__recompiles] Recompiling function _rand_given_ndim in /home/endoplasm/develop/torchrl-0/torchrl/data/replay_buffers/storages.py:147
V1023 12:03:57.543000 648354 site-packages/torch/_dynamo/guards.py:2842] [26/2] [__recompiles]     triggered by the following guard failure(s):
V1023 12:03:57.543000 648354 site-packages/torch/_dynamo/guards.py:2842] [26/2] [__recompiles]     - 26/1: L['self']._len_value == 200  # return torch.randint(  # develop/torchrl-0/torchrl/data/replay_buffers/storages.py:150 in _rand_given_ndim (_ops.py:723 in __call__)
V1023 12:03:57.543000 648354 site-packages/torch/_dynamo/guards.py:2842] [26/2] [__recompiles]     - 26/0: L['self']._len_value == 100
```

torchrl/data/replay_buffers/storages.py
kurtamohler: Though I can see that this change made other tests fail, so this isn't a good solution

torchrl/data/replay_buffers/storages.py
```diff
@@ -146,10 +146,14 @@ def _empty(self):
 
     def _rand_given_ndim(self, batch_size):
         # a method to return random indices given the storage ndim
+        if isinstance(self, TensorStorage):
```
kurtamohler, line 149: This is almost definitely too hacky, but I just wanted to check if this would fix the failures

torchrl/data/replay_buffers/storages.py
```diff
@@ -144,12 +144,19 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
     def _empty(self):
         ...
 
+    # NOTE: This property is used to enable compiled Storages. A `len(self)`
+    # call can cause recompiles, but for some reason, wrapping the call in a
+    # `property` decorated function avoids the recompiles.
+    @property
```
kurtamohler, line 150: It seems that the `property` decorator does something to prevent the recompiles. If I remove the decorator and change `self.len` to `self.len()` below, the recompiles happen again. I'm wondering if this is either some kind of bug in the compiler, or if the compiler treats properties in a special way. Maybe it forces a graph break, forcing `len` to be executed outside of compiled code? I'll look through the verbose logs to try to find an answer

vmoens (approved):
LGTM, I like it (and I'm learning a lot too!)
For the compile tests, we may want to decorate them with some pytorch version check - we have an olddeps CI workflow that runs with PT 2.0 for backward compatibility checks. It's ok to skip compile tests as it is not expected that these will succeed with earlier versions of PT.

torchrl/data/replay_buffers/storages.py
vmoens: That's weird! properties should not be treated differently than any other method I guess...

kurtamohler: Actually, it looks like the root cause really is the `mp.Value` after all! With the current state of this PR, I see this in the logs:

```
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks] torch._dynamo.exc.Unsupported: Graph break due to unsupported builtin None.SemLock.acq
uire. This function is either a Python builtin (e.g. _warnings.warn) or a third-party C/C++ Python extension (perhaps created with pybind). If it is a Python builtin, please file an issue on
 GitHub so the PyTorch team can add support for it and see the next case for a workaround. If it is a third-party C/C++ Python extension, please either wrap it into a PyTorch-understood cust
om operator (see https://pytorch.org/tutorials/advanced/custom_ops_landing_page.html for more details) or, if it is traceable, use torch.compiler.allow_in_graph.
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks] 
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks] from user code:
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks]    File "/home/endoplasm/develop/torchrl-0/torchrl/data/replay_buffers/storages.py", l
ine 159, in _rand_given_ndim
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks]     self.len,
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks]   File "/home/endoplasm/develop/torchrl-0/torchrl/data/replay_buffers/storages.py", li
ne 152, in len
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks]     return len(self)
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks]   File "/home/endoplasm/develop/torchrl-0/torchrl/data/replay_buffers/storages.py", li
ne 812, in __len__
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks]     return self._len
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks]   File "/home/endoplasm/develop/torchrl-0/torchrl/data/replay_buffers/storages.py", li
ne 469, in _len
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks]     return _len_value.value
V1023 15:21:38.467000 663811 site-packages/torch/_dynamo/convert_frame.py:1182] [27/0] [__graph_breaks]   File "<string>", line 3, in getvalue
```

Still, I don't understand what the difference is between decorator versus no decorator. If I change the `self.len` back to `len(self)`, I no longer see a graph break at that point in the logs. I wonder if it could have something to do with the fact that `Storage.__len__` is an abstract method...

torchrl/data/replay_buffers/storages.py
```diff
@@ -144,12 +144,29 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
     def _empty(self):
         ...
 
+    # NOTE: This property is used to enable compiled Storages. Calling
```
kurtamohler, line 147: I added this note to summarize what I know and don't know. I agree that it makes sense to merge this, but I think I can figure out what's going on here if I spend a little more time on it, so I'll submit another PR when I do

vmoens:
I'm happy with these changes!
Let's just make sure the olddeps workflow runs ok.
I think windows failed sporadically

PR #2519, opened 2024-10-27
[Doc] MADDPG bug fix of buffer device and improve explaination

### LINKED ISSUE #2515: [BUG]Competitive MARL (DDPG) with TorchRL Tutorial - Memory map device other than CPU isn't supported
Hi! Thanks for sharing your work on MARL with VMAS! The tutorial was great and included lots of details and explanations! I have been looking for examples with non-sharing parameters between agents within the same group, as well as centralized vs decentralized training. This tutorial is exactly what I need! Thank you!

## Describe the bug

I was trying to test on the performance of the CPU version with PettingZoo vs GPU version with VMAS. Apparently on replay buffer section, I am getting the error "ValueError: Memory map device other than CPU isn't supported." so I follow the instruction to cast the data by adding a line of code "replay_buffer.append_transform(lambda x: x.to(device))". So my question is:

Would this step slows down the training since there is a transform to change the device from CPU to GPU?
Could we use a replay buffer other than storage=LazyMemmapStorage if there is such a limitation? Which one could we use?

Thanks!

## To Reproduce

Run the tutorial "Competitive Multi-Agent Reinforcement Learning (DDPG) with TorchRL Tutorial".  Use GPU and VMAS=True.

ValueError: Memory map device other than CPU isn't supported. To cast your data to the desired device, use `buffer.append_transform(lambda x: x.to(device)` or a similar transform.

## Possible fixes
replay_buffer.append_transform(lambda x: x.to(device))

## Checklist

matteobettini:
Fix #2515

### DIFF
--- a/torchrl/data/replay_buffers/storages.py
+++ b/torchrl/data/replay_buffers/storages.py
@@ -1106,7 +1106,7 @@ class LazyMemmapStorage(LazyTensorStorage):
         if self.device.type != "cpu":
             raise ValueError(
                 "Memory map device other than CPU isn't supported. To cast your data to the desired device, "
-                "use `buffer.append_transform(lambda x: x.to(device)` or a similar transform."
+                "use `buffer.append_transform(lambda x: x.to(device))` or a similar transform."
             )
         self._len = 0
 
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM thanks for this

PR #2521, opened 2024-10-28
[BugFix] Allow expanding TensorDictPrimer transforms shape with parent batch size

albertbou92:
## Description

Often, when defining `TensorDictPrimer` transforms during environment creation, the batch size is unknown. For example when we create the transform with the` torchrl.modules.utils.get_primers_from_module` method.

In `TensorDictPrimer`, we assume that if the parent environment is batch-locked, the specs will already have the appropriate leading shape. However, this can lead to issues if the batch size is unknown at instantiation.

Maybe I am missing something here, but this PR makes sure the primer matches the parent batch size during reset, expanding it if necessary.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -142,6 +142,7 @@ from torchrl.envs.transforms.vc1 import _has_vc
 from torchrl.envs.transforms.vip import _VIPNet, VIPRewardTransform
 from torchrl.envs.utils import check_env_specs, step_mdp
 from torchrl.modules import GRUModule, LSTMModule, MLP, ProbabilisticActor, TanhNormal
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM, just a couple of things:
Since this will affect the transform in-place I think it should be documented in the docstrings of the class. Also, we should quickly test that the behaviour is as such, to make sure we don't modify it inadvertently in the future.

albertbou92:
I have remove the line where I first tried to set the primer batch_size to the parent batch_size. I could see edge cases where the batch_size could be set but still the specs needed to be expanded with and extra dimension. For example in the test I added.

PR #2529, opened 2024-10-31
[Performance] Improve performance of compiled ReplayBuffer

kurtamohler:
## Description

Performance improvements for compiled `ReplayBuffer.extend/sample` with `LazyTensorStorage`. For the cases tested in the benchmark, compiled gives the same or better performance than eager.

This uses the changes from #2426 to avoid using multiprocessing features, but also adds some other changes to avoid graph breaks in other places.

The benchmark and test are updated to include both cases where the storage's max size is reached and not reached.

## Motivation and Context

Part of #2501

## Types of changes

## Checklist

### DIFF
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -252,6 +252,11 @@ class implement_for:
     Keyword Args:
         class_method (bool, optional): if ``True``, the function will be written as a class method.
             Defaults to ``False``.
+        compilable (bool, optional): If ``False``, the module import happens
+            only on the first call to the wrapped function. If ``True``, the
+            module import happens when the wrapped function is initialized. This
+            allows the wrapped function to work well with ``torch.compile``.
+            Defaults to ``False``.
 
     Examples:
         >>> @implement_for("gym", "0.13", "0.14")
@@ -290,11 +295,13 @@ class implement_for:
         to_version: str = None,
         *,
         class_method: bool = False,
+        compilable: bool = False,
     ):
         self.module_name = module_name
         self.from_version = from_version
         self.to_version = to_version
         self.class_method = class_method
+        self._compilable = compilable
         implement_for._setters.append(self)
 
     @staticmethod
@@ -386,18 +393,27 @@ class implement_for:
         self.fn = fn
         implement_for._lazy_impl[self.func_name].append(self._call)
 
-        @wraps(fn)
-        def _lazy_call_fn(*args, **kwargs):
-            # first time we call the function, we also do the replacement.
-            # This will cause the imports to occur only during the first call to fn
+        if self._compilable:
+            _call_fn = self._delazify(self.func_name)
 
-            result = self._delazify(self.func_name)(*args, **kwargs)
-            return result
+            if self.class_method:
+                return classmethod(_call_fn)
 
-        if self.class_method:
-            return classmethod(_lazy_call_fn)
+            return _call_fn
+        else:
 
-        return _lazy_call_fn
+            @wraps(fn)
+            def _lazy_call_fn(*args, **kwargs):
+                # first time we call the function, we also do the replacement.
+                # This will cause the imports to occur only during the first call to fn
+
+                result = self._delazify(self.func_name)(*args, **kwargs)
+                return result
+
+            if self.class_method:
+                return classmethod(_lazy_call_fn)
+
+            return _lazy_call_fn
 
     def _call(self):
 
--- a/torchrl/data/replay_buffers/replay_buffers.py
+++ b/torchrl/data/replay_buffers/replay_buffers.py
@@ -19,6 +19,11 @@ import numpy as np
 
 import torch
 
+try:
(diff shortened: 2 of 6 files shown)


torchrl/data/replay_buffers/storages.py
```diff
@@ -550,6 +550,9 @@ def shape(self):
         if _total_shape is not None:
             return torch.Size([self._len_along_dim0] + list(_total_shape[1:]))
 
+    # TODO: Without this disable, compiler recompiles for back-to-back calls.
+    # Figuring out a way to avoid this disable would give better performance.
+    @torch._dynamo.disable()
     def _rand_given_ndim(self, batch_size):
```
kurtamohler, line 556: I haven't been able to figure this one out yet, unfortunately. Even if remove the `len(storage)` and use `self._len_value` directly, it still puts a guard around `_len_value`. I even tried completely removing any code that seems special about `_len_value` (the property decorator and anything related to `mp.Value`), and it still puts a guard around it.

I can't seem to find anything in the output from TORCH_TRACE that tells exactly why there's a guard on it. I've attached a trace of running the benchmark after commenting out the `@torch._dynamo.disable()`.

[tl_out.zip](https://github.com/user-attachments/files/17582288/tl_out.zip)

torchrl/data/replay_buffers/storages.py
```diff
@@ -674,22 +683,32 @@ def load_state_dict(self, state_dict):
...
+    _torch_20_to_23 = implement_for("torch", "2.0", "2.3")
+
+    def _set_tree_map_impl_20_to_23(self, cursor, data, storage):
         # flatten data and cursor
         data_flat = tree_flatten(data)[0]
         storage_flat = tree_flatten(storage)[0]
         for datum, store in zip(data_flat, storage_flat):
             store[cursor] = datum
 
+    def _set_tree_map(self, cursor, data, storage):
```
kurtamohler, line 704: I know this is not an elegant solution, so I'll try to figure out how to make `implement_for` work well with the compiler

kurtamohler: Oh I think I found a good way to do it, will push it up soon

torchrl/data/replay_buffers/storages.py
vmoens: that's ok, let's deactivate like you do here.
Perhaps we could isolate this and put a test around that graph break to make sure that if the compiler ends up hadling this properly we're aware of it and we can remove the disable() decorator?

torchrl/_utils.py
```diff
@@ -290,11 +295,13 @@ def __init__(
         to_version: str = None,
         *,
         class_method: bool = False,
+        compilable: bool = False,
```
vmoens, line 298: Smart!

torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -600,7 +612,9 @@ def _add(self, data):
         return index
 
     def _extend(self, data: Sequence) -> torch.Tensor:
-        with self._replay_lock, self._write_lock:
+        is_compiling = torch.compiler.is_dynamo_compiling()
```
vmoens, line 615: it's a bit annoying but since we want the code to be compatible with torch 2.0, we need to import `is_dynamo_compiling` like this
```
try:
     from torch.compiler import is_dynamo_compiling
except ImportError:
     from torch._dynamo import is_compiling as is_dynamo_compiling
```

torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -653,7 +667,7 @@ def update_priority(
 
     @pin_memory_output
     def _sample(self, batch_size: int) -> Tuple[Any, dict]:
-        with self._replay_lock:
+        with self._replay_lock if not torch.compiler.is_dynamo_compiling() else contextlib.nullcontext():
```
vmoens, line 670: ditto

vmoens (approved):
LGTM thanks a mil.
Just a couple of quick fixes here and there and we're good to go.
Do you think we should add a benchmark for the E2E replay buffer interactions?
Another next step is to make sure compile works with MemoryMappedTensor from tensordict - I'm pretty sure it won't like it

kurtamohler:
> Do you think we should add a benchmark for the E2E replay buffer interactions?

Yeah I think that's a good idea, but I thought the benchmark I added was exercising the end-to-end interface, so I'm not entirely sure what you mean. Do you have a specific case in mind? Maybe something that actually uses the data for something rather than just expand/sample?

torchrl/data/replay_buffers/replay_buffers.py
kurtamohler: Done

torchrl/data/replay_buffers/storages.py
kurtamohler: Good idea, I've added a test for that

PR #2543, opened 2024-11-06
[BugFix] Support for tensor collection in the `PPOLoss`

priba:
## Description

The function `get_entropy_bonus` was not unpacking the `entropy` from the `TensorDict`.

## Motivation and Context

When using a `CompositeDistribution`, the return type of the entropy can be either a `Tensor` or a `TensorDict` depending on the `aggregate_probabilities`. This break the `get_entropy_bonus` that expects a `Tensor`. We can unpack it accessing by the entropy key.

## Types of changes

## Checklist

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -7650,6 +7650,7 @@ class TestPPO(LossModuleTestBase):
         observation_key="observation",
         sample_log_prob_key="sample_log_prob",
         composite_action_dist=False,
+        aggregate_probabilities=True,
     ):
         # Actor
         action_spec = Bounded(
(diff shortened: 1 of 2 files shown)


moses1998lee:
Hi, can i check if this is in version 0.6.0 of torchrl?

vmoens:
This is an active area of development. This solution should be the one in v0.6, but we're now going to use something slightly different
On nightlies, you can check [this example](https://github.com/pytorch/rl/blob/main/examples/agents/composite_ppo.py) to get a sense of what the v0.7 API will look like

PR #2548, opened 2024-11-11
[Doc] Minor fixes to the docs and type hints

thomasbbrunner:
## Description

Minor improvements to the documentation and type hints of the environment code.

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -516,7 +516,7 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         self,
         transform: "Transform"  # noqa: F821
         | Callable[[TensorDictBase], TensorDictBase],
-    ) -> None:
+    ) -> EnvBase:
         """Returns a transformed environment where the callable/transform passed is applied.
 
         Args:
@@ -1482,7 +1482,8 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
 
     # Single-env specs can be used to remove the batch size from the spec
     @property
-    def batch_dims(self):
+    def batch_dims(self) -> int:
+        """Number of batch dimensions of the env."""
         return len(self.batch_size)
 
     def _make_single_env_spec(self, spec: TensorSpec) -> TensorSpec:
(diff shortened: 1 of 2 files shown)


torchrl/envs/common.py
```diff
@@ -516,7 +516,7 @@ def append_transform(
         self,
         transform: "Transform"  # noqa: F821
         | Callable[[TensorDictBase], TensorDictBase],
-    ) -> None:
+    ) -> EnvBase:
```
thomasbbrunner, line 519: This should actually return `TransformedEnv`.

However, I'm guessing we don't want to import it at the top to prevent loading a lot of modules that won't be accessed?

torchrl/envs/common.py
```diff
@@ -2393,14 +2393,16 @@ def rollout(
                 the call to ``rollout``.
 
         Keyword Args:
-            auto_reset (bool, optional): if ``True``, resets automatically the environment
-                if it is in a done state when the rollout is initiated.
-                Default is ``True``.
```
thomasbbrunner, line 2398: IMO this docstring was wrong and caused some confusion on our side. I hope that I was able to get the correct meaning of this argument.

torchrl/envs/common.py
```diff
@@ -2393,14 +2393,16 @@ def rollout(
...
+            auto_reset (bool, optional): if ``True``, the contained environments will be reset before starting the
+                rollout. If ``False``, then the rollout will continue from a previous state, which requires the
+                ``tensordict`` argument to be passed with the previous rollout. Default is ``True``.
             auto_cast_to_device (bool, optional): if ``True``, the device of the tensordict is automatically cast to the
                 policy device before the policy is used. Default is ``False``.
-            break_when_any_done (bool): breaks if any of the done state is True. If False, a reset() is
-                called on the sub-envs that are done. Default is True.
-            break_when_all_done (bool): TODO
```
thomasbbrunner, line 2403: Took the liberty of adding a docstring here and making some things a bit clearer. Open to suggestions for improvements!

torchrl/envs/common.py
vmoens: Either `EnvBase` or "torchrl.envs.TransformedEnv" but we can't import it 
`EnvBase` could be the best compromise!

torchrl/envs/common.py
vmoens: That looks great!

torchrl/envs/common.py
vmoens: haha yeah much better! 
200w

vmoens (approved):
LGTM thanks!

PR #2567, opened 2024-11-14
[Feature] Add `Stack` transform

### LINKED ISSUE #2566: [Feature Request] Transform that stacks data for agents with identical specs
## Motivation

Some multi-agent environments, like `VmasEnv`, stack all of the tensors for observations, rewards, etc. for different agents that have identical specs. For instance, in one of these stacked environments, if there are 2 agents that each have 8 observations, the observation spec might look like this:

```
Composite(
    group_0: Composite(
        observation: UnboundedContinuous(
            shape=torch.Size([2, 8]),
            ...),
        shape=torch.Size([2])),
    shape=torch.Size([]))
```

In contrast, other environments, like `UnityMLAgentsEnv`, have separate keys for each agent, even if the agents' specs are identical. For instance, with 2 agents that each have 8 observations, the observation spec might look like this:

```
Composite(
    group_0: Composite(
        agent_0: Composite(
            observation: UnboundedContinuous(
                shape=torch.Size([8]),
                ...),
            shape=torch.Size([])),
        agent_1: Composite(
            observation: UnboundedContinuous(
                shape=torch.Size([8]),
                ...),
            shape=torch.Size([])),
        shape=torch.Size([])),
    shape=torch.Size([]))
```

It is not easy to apply the same training script to two environments that use these two different formats. For instance, applying the [multi-agent PPO tutorial](https://pytorch.org/rl/main/tutorials/multiagent_ppo.html) to a Unity env is not straightforward.

## Solution

If we had an environment transform that could stack all the data from different keys, we could convert an environment that uses the unstacked format into an environment that uses the stacked format. Then it should be straightforward to use the same (or almost the same) training script on the two different environments.

## Checklist

kurtamohler:
## Description

Adds a transform that stacks tensors and specs from different keys of a tensordict into a common key.

## Motivation and Context

close #2566

## Types of changes

## Checklist

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -45,6 +45,7 @@ if os.getenv("PYTORCH_TEST_FBCODE"):
         IncrementingEnv,
         MockBatchedLockedEnv,
         MockBatchedUnLockedEnv,
+        MultiAgentCountingEnv,
         MultiKeyCountingEnv,
         MultiKeyCountingEnvPolicy,
         NestedCountingEnv,
@@ -71,6 +72,7 @@ else:
         IncrementingEnv,
         MockBatchedLockedEnv,
         MockBatchedUnLockedEnv,
+        MultiAgentCountingEnv,
         MultiKeyCountingEnv,
         MultiKeyCountingEnvPolicy,
         NestedCountingEnv,
@@ -134,6 +136,7 @@ from torchrl.envs import (
     SerialEnv,
     SignTransform,
     SqueezeTransform,
+    Stack,
     StepCounter,
     TargetReturn,
     TensorDictPrimer,
@@ -141,12 +144,14 @@ from torchrl.envs import (
     ToTensorImage,
     TrajCounter,
     TransformedEnv,
+    UnityMLAgentsEnv,
     UnsqueezeTransform,
     VC1Transform,
     VIPTransform,
 )
 from torchrl.envs.libs.dm_control import _has_dm_control
 from torchrl.envs.libs.gym import _has_gym, GymEnv, set_gym_backend
+from torchrl.envs.libs.unity_mlagents import _has_unity_mlagents
 from torchrl.envs.transforms import VecNorm
 from torchrl.envs.transforms.r3m import _R3MNet
 from torchrl.envs.transforms.rlhf import KLRewardTransform
@@ -159,7 +164,7 @@ from torchrl.envs.transforms.transforms import (
 )
 from torchrl.envs.transforms.vc1 import _has_vc
 from torchrl.envs.transforms.vip import _VIPNet, VIPRewardTransform
-from torchrl.envs.utils import check_env_specs, step_mdp
+from torchrl.envs.utils import check_env_specs, MarlGroupMapType, step_mdp
 from torchrl.modules import GRUModule, LSTMModule, MLP, ProbabilisticActor, TanhNormal
 from torchrl.modules.utils import get_primers_from_module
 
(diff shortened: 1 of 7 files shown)


kurtamohler:
Looks like there is a minor bug if I try to use this on `UnityMLAgentsEnv` and then do a rollout. I'll fix that and add a test

torchrl/envs/transforms/transforms.py
```diff
@@ -4317,6 +4317,150 @@ def __repr__(self) -> str:
...
+    algorithms that expect the tensors for observations, rewards, etc. to
+    contain batched data for all the agents.
+
+    Args:
+        in_keys (sequence of NestedKey): keys to be stacked. If `None` or not
+            provided, the keys will be retrieved from the group map of the
+            environment the first time the transform is used. This behavior will
+            only work if a parent is set.
+        out_key (NestedKey): key of the resulting tensor.
+        dim (int, optional): dimension to insert. Default is ``0``.
```
vmoens, line 4335: there's usually some benefit in using negative dims: if your transform is being used by a single env, multiple envs (ie envs where batch-size is non-empty) and / or replay buffers, the `dim=0` arg will lead to different output shapes, putting a feature dimension before the batch size!
Not sure we want to strictly prevent that, but for instance in `FlattenObservation` we ask the user to explicitly tell us that they want the dim to be non-negative.

test/test_transforms.py
```diff
@@ -2136,6 +2139,121 @@ def test_transform_no_env(self, device, batch):
         pytest.skip("TrajCounter cannot be called without env")
 
 
+class TestStack(TransformBase):
```
vmoens, line 2142: when testing a transform, we need to check that they behave correctly in single env, multi-env, replay buffers and modules. You can go through the abstract methods in the base class and check other tests (eg CatTensor) to see an example

vmoens (commented):
Thanks for this, long awaited feature!
Just left a couple of comments on the default dim and test set

torchrl/envs/transforms/transforms.py
kurtamohler: I see, that makes sense. I can make the default `allow_positive_dim=False, dim=-1`.

It is worth pointing out that if the user wants to stack multi-agent data, the different tensors under the agents may have different numbers of dimensions, so they will have to set `allow_positive_dim=True, dim=0`. I'll add an example of that to the docstring

kurtamohler: Oh I didn't realize that if the elements to be stacked are tensordicts with `batch_size=[]`, `torch.stack(values, dim=-1)` is actually  equivalent to `torch.stack(values, dim=0)`, so the default of `allow_positive_dim=False, dim=-1` works fine

vmoens (commented):
Thanks this is superb
I'd like to discuss the inverse transform:
Would it make sense in the inverse to get an entry (from the `input_spec`) and unbind it?
Like: you have a single action with leading dim of 2, and map it to `("agent0", "action")`, `("agent1", "action")`. The spec seen from the outside is the stack of the 2 specs (as it is for stuff processed in forward).
Would that make sense?

torchrl/envs/transforms/transforms.py
```diff
@@ -4323,6 +4323,183 @@ def __repr__(self) -> str:
...
+            tensordict.exclude(*self.in_keys, inplace=True)
+        return tensordict
+
+    forward = _call
+
+    def _inv_call(self, tensordict: TensorDictBase) -> TensorDictBase:
+        values = torch.split(tensordict[self.in_keys_inv[0]], 1, dim=self.dim)
+        for value, out_key_inv in _zip_strict(values, self.out_keys_inv):
+            tensordict.set(out_key_inv, value.squeeze(self.dim))
+        tensordict.exclude(self.in_keys_inv[0], inplace=True)
```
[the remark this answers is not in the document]
kurtamohler: I'll remove the inplace. I had the mistaken impression that since the forward transformations are inplace, then the inverse would be too

torchrl/envs/transforms/transforms.py
```diff
@@ -4323,6 +4323,183 @@ def __repr__(self) -> str:
...
+        out_tensor = torch.stack(values, dim=self.dim)
+        tensordict.set(self.out_keys[0], out_tensor)
+        if self._del_keys:
+            tensordict.exclude(*self.in_keys, inplace=True)
+        return tensordict
+
+    forward = _call
+
+    def _inv_call(self, tensordict: TensorDictBase) -> TensorDictBase:
+        values = torch.split(tensordict[self.in_keys_inv[0]], 1, dim=self.dim)
```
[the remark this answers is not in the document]
kurtamohler: Good points, I switched to `unbind` and added a check for the key

kurtamohler:
> Would it make sense in the inverse to get an entry (from the `input_spec`) and unbind it?
> Like: you have a single action with leading dim of 2, and map it to `("agent0", "action")`, `("agent1", "action")`. The spec seen from the outside is the stack of the 2 specs (as it is for stuff processed in forward).
> Would that make sense?

Yes, I think that does make sense, and that is exactly what happens. For instance, if I add a line to print the output of `Stack._inv_call` at the end of the function and then run the following script:

<details><summary>Click to expand</summary>

```python
from torchrl.envs import Stack, TransformedEnv
from torchrl.envs import UnityMLAgentsEnv

base_env = UnityMLAgentsEnv(registered_name='3DBall')

try:
    t = Stack(
        in_keys=[("group_0", f"agent_{idx}") for idx in range(12)],
        out_key="agents",
    )   
    env = TransformedEnv(base_env, t)
    action = env.full_action_spec.rand()
    print("-------------------------")
    print(action)
    print("-------------------------")
    env.step(action)

finally:
    base_env.close()
```

</details>

I get the following output:

<details><summary>Click to expand</summary>

```
-------------------------
TensorDict(
    fields={
        agents: TensorDict(
            fields={
                continuous_action: Tensor(shape=torch.Size([12, 2]), device=cpu, dtype=torch.float32, is_shared=False)},
            batch_size=torch.Size([12]),
            device=None,
            is_shared=False),
        group_0: TensorDict(
            fields={
            },
            batch_size=torch.Size([]),
            device=None,
            is_shared=False)},
    batch_size=torch.Size([]),
    device=None,
    is_shared=False)
-------------------------
TensorDict(
    fields={
        group_0: TensorDict(
            fields={
                agent_0: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_10: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_11: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_1: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_2: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_3: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_4: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_5: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_6: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_7: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_8: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                agent_9: TensorDict(
                    fields={
                        continuous_action: Tensor(shape=torch.Size([2]), device=cpu, dtype=torch.float32, is_shared=False)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False)},
            batch_size=torch.Size([]),
            device=None,
            is_shared=False)},
    batch_size=torch.Size([]),
    device=None,
    is_shared=False)
```

<details>

Which shows that the inverse of Stack is able to correctly unbind the stacked actions to the format the the unity env expects. BTW, this functionality is being tested in one of the unit tests I added

vmoens:
Ah got it, sorry I was expecting something like in_keys or such. Maybe let's make clear in the doc strings that in_keys can be part of the input (usually it's reserved to output keys)

Maybe this? If that's what you need you can mention it in the doc of the transform.

https://pytorch.org/rl/main/reference/generated/torchrl.envs.transforms.RemoveEmptySpecs.html#torchrl.envs.transforms.RemoveEmptySpecs

vmoens:
> I'm sorry, I'm not sure what you mean by this. in_keys is always supposed to be part of the input, isn't it? Are you talking about the inverse operation here, or the leftover empty specs/tensordicts, or something else?

Yeah usually anything you iterate over during inverse pass is passed through the inv_keys and anything you process during forward is in the in_keys.
It's useful to keep them separated because it might as well be the case that there is a key in the input named as an output but you just want to process it one way and not the other.

vmoens:
> I don't think the way I implemented it is overly complicated, and it should give significantly better performance than RemoveEmptySpecs, since it knows which keys to check ahead of time.

Ok that sounds good!

kurtamohler:
> Yeah usually anything you iterate over during inverse pass is passed through the inv_keys and anything you process during forward is in the in_keys.

What I think you're saying is that most of the existing transforms that have an inverse allow the user to explicitly set the inverse keys in `__init__`. But with `Stack`, the inverse keys are automatically assigned to `in_key_inv = out_key` and `out_keys_inv = in_keys`. So we should document that difference. Is that what you're saying?

I guess we also might as well allow the user to override the default `in_key_inv` and `out_keys_inv`.

vmoens:
Yep that's what I'm saying.
But also I'm a bit worried about this behaviour: in some cases, the observations are part of the input tensordict but they're not input to the env (not registered in the input spec). With the current state of things, they will be split before being fed to the base env when it's not necessary for them to be (it would only be the case if they were part of the input spec), inducing some overhead (with the unbind function).

So I'd rather ask users to indicate the ˋin_keys` and ˋin_keys_inv` separately.

kurtamohler:
> So I'd rather ask users to indicate the ˋin_keys` and ˋin_keys_inv` separately.

Got it, that makes sense. I'll make that change

kurtamohler:
Actually, I guess there are still some decisions to make about the specific behavior of `in_keys_inv` which we should discuss. What do you think about this?

```
        in_keys_inv (NestedKey or sequence of NestedKey): keys to be unstacked
            during :meth:`~.inv`. To fully invert the output of
            :meth:`~forward`, set equal to ``out_key``. To only invert part of
            the :meth:`~forward` output, specify subkeys of ``out_key``. For
            instance, if ``out_key=("agents",)`` and ``in_keys_inv=[("agents",
            "action"), ("agents", "state")]``, then :meth:`~.inv` will only
            unstack the "action" and "state" keys under "agents", leaving
            everything else under "agents" untouched.
```

We'll probably have to enforce that all of the keys specified in `in_keys_inv` start with the prefix `out_key`. For instance, if `out_key=("agents",)`, then each of the `in_keys_inv` need to be `("agents",) + <some nested key>`.

We can still infer the `out_keys_inv` automatically. But would there be a reason to allow the user to set `out_keys_inv` as well? If we did allow the user to set `out_keys_inv`, then looking at the following example:

```
in_keys=[("agent0",), ("agent1",), ...]
out_key=("agents",)
in_keys_inv=[("agents", "action"), ("agents", "state")]
```

`out_keys_inv` would need to be able to indicate a mapping where both the "action" and "state" gets split into multiple agents, like this:

```
("agents", "action") --> [("agent0", "action"), ("agent1", "action"), ...]
("agents", "state") --> [("agent0", "state"), ("agent1", "state"), ...]
```

To me, it seems that a natural representation of `out_keys_inv` would be a list of lists of nested keys:

```
out_keys_inv = [
    [("agent0", "action"), ("agent1", "action"), ...],
    [("agent0", "state"), ("agent1", "state"), ...]
]
```

where `out_keys_inv[i]` is the list of keys to split the key `in_keys_inv[i]` into. But (I think) none of the other transforms specify keys in the format of a list of lists of nested keys, so it would be out of the ordinary. Plus, maybe it's a bit more verbose than it really needs to be.

Instead, maybe `out_keys_inv` could just be:

```
out_keys_inv = [
    ("agent0",), ("agent1",), ...,
]
```

(Which is just equal to `in_keys` in this case)

But the docs would need to make it clear that `out_keys_inv` specifies the prefixes of the actual keys to which the inverse op assigns all the unstacked elements, and that the mapping between `in_keys_inv` and `out_keys_inv` really goes like:

```
in_keys_inv[i] --> [prefix + in_keys_inv[i][len(out_key) : -1] for prefix in out_keys_inv ]
```

Where `i` iterates over `in_keys_inv`. Note that `out_key` here is the output key of the forward op.

What do you think? Is there a better decision for the behavior of `in_keys_inv`? Can you think of a reason why users would want to specify `out_keys_inv`? If not, then I suppose we can leave it out for now until someone requests that feature

vmoens:
> ```
> in_keys_inv=[("agents", "action"), ("agents", "state")]
> ```

@kurtamohler what about we simply restrict the transform to stack X keys into 1 during forward and split 1 entry into Y during inverse?

If users want to unbind multiple keys they should just have multiple transforms.

kurtamohler:
Ok that sounds good to me!

torchrl/envs/transforms/transforms.py
```diff
@@ -4323,6 +4323,210 @@ def __repr__(self) -> str:
...
 
 
+class Stack(Transform):
+    """Stacks tensors and tensordicts.
+
+    This transform is useful for environments that have multiple agents with
+    identical specs under different keys. The specs and tensordicts for the
+    agents can be stacked together under a shared key, in order to run MARL
+    algorithms that expect the tensors for observations, rewards, etc. to
+    contain batched data for all the agents.
```
vmoens, line 4333: I would rather have a general statement about the transform first - some will use that just to stack tensors together, not in MARL settings (eg multiple images in one)

vmoens: maybe add a note saying that multiple stacks will require multiple transforms?

vmoens (approved):
Very minor comments but otherwise LGTM

torchrl/envs/transforms/transforms.py
kurtamohler: Added, let me know if that's good

PR #2584, opened 2024-11-19
[Refactor] Deprecate recurrent_mode API to use decorators/CMs instead

vmoens:
* __->__ #2584

Also ensures that all public loss functions are using the appropriate exploration mode and recurrent mode.

cc @thomasbbrunner


thomasbbrunner:
Oh, that was super fast. Many thanks!

PR #2592, opened 2024-11-20
[BugFix] action_spec_unbatched whenever necessary

vmoens:
* #2593
* __->__ #2592

cc @matteobettini we now have a more generic `single_action_spec` that is somewhat similar to unbached_action_spec (name is borrowed [from gymnasium](https://github.com/Farama-Foundation/Gymnasium/blob/79a2306dbab6f998b57953bfbc37fc2099c46683/gymnasium/vector/vector_env.py#L439))
There is something a bit fishy about `unbatched_action_spec` though which is that it's the `unbatched_full_action_spec`, and it doesn't (always) match action spec in its sturcture (it will always be composite).

LMK if these changes make sense (implemented these changes to fix `tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py` which was broken)


sota-implementations/multiagent/iql.py
```diff
@@ -91,7 +91,7 @@ def train(cfg: "DictConfig"):  # noqa: F821
             ("agents", "action_value"),
             ("agents", "chosen_action_value"),
         ],
-        spec=env.unbatched_action_spec,
+        spec=env.single_action_spec,
```
matteobettini, line 94:
```suggestion
        spec=env.single_full_action_spec,
```
Wherever `unbatched_action_spec` is used, it should be changed to `single_full_action_spec`

This is valid for all the places here. As the components that were receiving `unbatched_` are expecting a composite

tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py
```diff
@@ -486,8 +486,8 @@
         out_keys=[(group, "action")],
         distribution_class=TanhDelta,
         distribution_kwargs={
-            "low": env.full_action_spec[group, "action"].space.low,
-            "high": env.full_action_spec[group, "action"].space.high,
+            "low": env.single_full_action_spec[group, "action"].space.low,
+            "high": env.single_full_action_spec[group, "action"].space.high,
```
matteobettini, line 490: The line before:
`spec=env.full_action_spec[group, "action"],`
should also be changed then.

tutorials/sphinx-tutorials/multiagent_ppo.py
```diff
@@ -445,13 +445,13 @@
 
 policy = ProbabilisticActor(
     module=policy_module,
-    spec=env.unbatched_action_spec,
+    spec=env.single_action_spec,
```
matteobettini, line 448: Also here the components are expecting the full one.

They might work without a full one but there might be subtle bugs I think.

matteobettini (commented):
Got it.

Love that there is a way to get the unbatched specs easily now. This is super helpful.

I guess the prefix `single_` has already been decided but I will give my 2 cents on this anyway as I think there are some important points.

I think `single_` is a really confusing name. In gymnasium I guess it can work as the library is not multi-agent/multi-task. but in torchrl it sounds super confusing to me:

- it could be meaning single agent
- or single spec (as opposed to full spec). this is why `single_full_spec` sounds oximoric to me
- it does not correlate with the concept of a batch_size

To me the prefix `unbatched_` (or variations of it) drives the concept home better without any of the confusions above

vmoens:
We can name it unbatched but it will clash with VMAS unbatched (which are somewhat different).
It's a nightly feature IIRC so we can do it

matteobettini:
> We can name it unbatched but it will clash with VMAS unbatched (which are somewhat different). It's a nightly feature IIRC so we can do it

Yeah, the vmas ones were there for the absence of this feature. Now that this exists the vmas ones could be removed. 

Your are right that the vmas name implied full so there would be a clash there. 

We could think about how to best deal with this but i think torchrl should not be prevented to use the better name cause of this

matteobettini:
so the difference between full and normal is just present when there is only one agent group.

What we could do is that for a period of time we could warn vmas users that request `unbatched_spec` in envs with just one group that this is now returning the non-compiste version and if they want the old composite one there is `unbatched_full_spec`

vmoens:
That's bc breaking, we need to warn them the behaviour will change and they can make the warning disappear by using the full version

matteobettini:
> Intermediate solution: `single_action_spec` becomes `action_spec_unbatched` that way there is no conflict, and we can just raise a warning in VMAS to let people know about the new API. Wdyt?

Genius! Love it

vmoens:
Ok I also simplified the mock envs that had unbatched specs, LMK if that makes sense!

matteobettini:
I think that makes sense. My comments from the first review still hold tho. I would just make this a refactoring and no logic change

so `unbatched_spec` in vmas should be translated to `full_spec_unbatched`

tutorials/sphinx-tutorials/multiagent_ppo.py
vmoens: Nope I don't think so
1. it's a tutorial so if it passes it passes (it's not a public API that is likely to change)
2. `ProbabilisticActor` will look at your spec and do the required wrapping if it's not a Composite spec. And if the action is composite, so is `unbatched_action_spec`. So in either case it will work.

vmoens:
> I would just make this a refactoring and no logic change

There is not logic change, see my comment above

tutorials/sphinx-tutorials/multiagent_ppo.py
matteobettini: It is that "will do the  required wrapping if it's not a Composite spec" that worries me. I don't think it knows the group name. I am afraid that this could lead to subtle bugs but I trust you

vmoens: if the full actions spec is not trivially "action" it returns the composite, so you're safe

vmoens:
@matteobettini I'll wait until the tests pass but it should be good.
I did not make any deprecation in VMAS, not sure how you want to procede there.

matteobettini:
Maybe we can put a warning to use the new feature?

PR #2593, opened 2024-11-20
[Refactor] Use <spec>_unbatched in VMAS

vmoens:
* (to be filled)


matteobettini (approved):
Cool! So the batched versions are built automatically?

vmoens:
yep I wrote a setter for them

PR #2604, opened 2024-11-24
[BugFix] Wrong spec returned

matteobettini:
Maybe we need some tests that check these automatically with all the variations: full, unbatched, etc


vmoens (approved):
LGTM thanks for this

PR #2606, opened 2024-11-25
[BugFix] Account for terminating data in SAC losses

vmoens:
* __->__ #2606

### DIFF
--- a/torchrl/objectives/sac.py
+++ b/torchrl/objectives/sac.py
@@ -16,7 +16,7 @@ import torch
 from tensordict import TensorDict, TensorDictBase, TensorDictParams
 
 from tensordict.nn import dispatch, TensorDictModule
-from tensordict.utils import NestedKey
+from tensordict.utils import expand_right, NestedKey
 from torch import Tensor
 from torchrl.data.tensor_specs import Composite, TensorSpec
 from torchrl.data.utils import _find_action_space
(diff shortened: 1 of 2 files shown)


torchrl/objectives/sac.py
```diff
@@ -711,13 +711,37 @@ def _compute_target_v2(self, tensordict) -> Tensor:
             with set_exploration_type(
                 ExplorationType.RANDOM
             ), self.actor_network_params.to_module(self.actor_network):
-                next_tensordict = tensordict.get("next").clone(False)
-                next_dist = self.actor_network.get_dist(next_tensordict)
+                next_tensordict = tensordict.get("next").copy()
+                # Check done state and avoid passing these to the actor
+                done = next_tensordict.get(self.tensor_keys.done)
+                if done is not None and done.any():
+                    next_tensordict_select = next_tensordict[~done.squeeze(-1)]
```
matteobettini, line 718: The done shape could be more extended than the batch shape, this line is breaking in multiagent settings

vmoens: Then we need a test that cover this use case!
Can you draft one for me?

matteobettini: The SOTA ci picked up on this. Both SAC scripts are failing

vmoens: Ah i didn't see (SOTA is broken bc of dreamer so I didn't check)
we should have tests that are not in SOTA, SOTA is there to test that scripts run smoothly, not features. The scripts are not part of the core lib - we can arbitrarily decide to ditch them, the rest of the lib should still work.

matteobettini: Yeah I long wanted to make some tests for multiagent data in losses, will get to it when I have time.

Right now just crunching on writing thesis and satisfying BenchMARL users in free time.

Issue #2612, opened 2024-11-27
[BUG] SAC loss masking


matteobettini:
PR https://github.com/pytorch/rl/pull/2606

Introduces indexing of loss tensordict using done signals

https://github.com/pytorch/rl/blob/d537dcb6347e2370fcdaed553bf3474d653cb5a6/torchrl/objectives/sac.py#L718

https://github.com/pytorch/rl/blob/d537dcb6347e2370fcdaed553bf3474d653cb5a6/torchrl/objectives/sac.py#L1223

I have multiple concerns regarding this PR:

1. masking using done signals is not always possible as the done signal could have more dimensions than the loss tensordict, leading to errors in multiagent settings
2. the action network could be expecting a specific imput dimension, masking the input could lead to arbitrary unsupported shapes and crashes in the actor (also not in multiagent settings)
3. in continuous sac, the mask is applied to output actions and logprobabilities (with arbitrary 0 entries). this does not seem necessary, those values will be discarded anyway as the `value_estimate()` will read the dones
4. same in descrete sac, it seems to me that the `value_estimate()` already reads done and discards next_values for done states. plus the target of a done state should be the reward, so here by using 0s we are actually introducing a further bug

In my opinion this change was not needed as the done target_values of done states are already discarded in `value_estimate()`.

Maybe I am wrong in this analysis, please let me know.

I do not think it is possible to avoid submitting inputs of done states to the policy without changing the input shape (which we should avoid as it could lead to errors)

matteobettini:
If the problem is that the observation of done state could be nan (which I argue it shouldn't) we could consider replacing nans with 0 so that the forward can run

vmoens:
I think we also need more coverage of MARL usage of the losses in the tests, because this could have been easily spotted if running `pytest test/test_cost.py -k SAC` had any MARL setting in it

matteobettini:
I understand the orginal issue, but this seems to be a very difficult pickle.

not all networks can be queried with sparse data or data or arbitrary shape (list is long)

matteobettini:
> There should not be any forward pass for values that are nonsensical and where outputs will not be used.

Not sure it makes sense to enforce this here without introducing problems bigger than the original

vmoens:
> Not sure it makes sense to enforce this here without introducing problems bigger than the original

You know what, I'm happy to revert that PR as soon as we can figure out what to do for people who have networks that simply cannot accept "done" observation values!

[...]

PR #2613, opened 2024-11-27
[BugFix] skip_done_states in SAC

vmoens:
* __->__ #2613

### DIFF
--- a/torchrl/objectives/sac.py
+++ b/torchrl/objectives/sac.py
@@ -126,6 +126,10 @@ class SACLoss(LossModule):
             ``"none"`` | ``"mean"`` | ``"sum"``. ``"none"``: no reduction will be applied,
             ``"mean"``: the sum of the output will be divided by the number of
             elements in the output, ``"sum"``: the output will be summed. Default: ``"mean"``.
+        skip_done_states (bool, optional): whether the actor network should only be run on valid, non-terminating
+            next states. If ``True``, it is assumed that the done state can be broadcast to the shape of the
+            data and that masking the data results in a valid data structure. Among other things, this may not
+            be true in MARL settings or when using RNNs. Defaults to ``False``.
 
     Examples:
         >>> import torch
@@ -320,6 +324,7 @@ class SACLoss(LossModule):
         priority_key: str = None,
         separate_losses: bool = False,
         reduction: str = None,
+        skip_done_states: bool = False,
     ) -> None:
         self._in_keys = None
         self._out_keys = None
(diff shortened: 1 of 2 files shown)


torchrl/objectives/sac.py
```diff
@@ -126,6 +126,10 @@ class SACLoss(LossModule):
             ``"none"`` | ``"mean"`` | ``"sum"``. ``"none"``: no reduction will be applied,
             ``"mean"``: the sum of the output will be divided by the number of
             elements in the output, ``"sum"``: the output will be summed. Default: ``"mean"``.
+        skip_done_states (bool, optional): whether the actor network should only be run on valid, non-terminating
```
matteobettini, line 129:
```suggestion
        skip_done_states (bool, optional): whether the actor network used for value computation should only be run on valid, non-terminating
```

torchrl/objectives/sac.py
```diff
@@ -877,6 +891,10 @@ class DiscreteSACLoss(LossModule):
...
```
matteobettini, line 894: same as above

matteobettini (approved):
LGTM thanks!

One other option to consider to avoid changing the input shape in case the user sets this flag is to ask what to use to pad the non-terminated obs

vmoens:
> LGTM thanks!
> 
> One other option to consider to avoid changing the input shape in case the user sets this flag is to ask what to use to pad the non-terminated obs

See [this comment](https://github.com/pytorch/rl/issues/2612#issuecomment-2504137606)

matteobettini:
> > LGTM thanks!
> > One other option to consider to avoid changing the input shape in case the user sets this flag is to ask what to use to pad the non-terminated obs
> 
> See [this comment](https://github.com/pytorch/rl/issues/2612#issuecomment-2504137606)

Yes I saw. I was referring exactly to that. Maybe there are users who have such issue but also need an input with the same shape (cholesky expects a matirx for isntance). in that case they might know what works for them (maybe nana and 0 no but 1 yes)

Just an idea, we don't need to do it

vmoens:
> Yes I saw. I was referring exactly to that. Maybe there are users who have such issue but also need an input with the same shape (cholesky expects a matrix for isntance). in that case they might know what works for them (maybe nana and 0 no but 1 yes)

with cholesky (if we want to take that example), any fixed number filling the matrix will fail. Padding is simply not a solution unfortunately.

matteobettini:
> with cholesky (if we want to take that example), any fixed number filling the matrix will fail. Padding is simply not a solution unfortunately.

got it. but neither is masking if i understand right

matteobettini:
ok got it

PR #2622, opened 2024-12-02
[BugFix] Better account of composite distributions in PPO

vmoens:
* __->__ #2622

Happy to massage this a little bit if the API isn't optimal!
Do you think we're covering all (edge) cases?

cc @louisfaury @albertbou92


louisfaury:
lgtm -- thanks @vmoens for looking into this!

PR #2623, opened 2024-12-02
[Tutorial] Beam search with GPT models

vmoens:
* __->__ #2623


docs/requirements.txt
```diff
@@ -28,3 +28,7 @@ vmas
 onnxscript
 onnxruntime
 onnx
+plotly
+igraph
+transformers
+datasets
```
kurtamohler, line 34: Might be good to mention in the tutorial that these need to be installed

tutorials/sphinx-tutorials/beam_search_with_gpt.py
```diff
@@ -0,0 +1,414 @@
...
+# Because we have hashes that uniquely define the trajectory up to a given step, it's easier to use these hash values
+# to pinpoint the unique nodes rather than using the observations directly.
+#
+# Notice that indexing the relevant nodes is made easy thanks to tensordict's API!
+#
+
+
+def select_unique_obs(td):
+    # Get the obs (the hash)
+    hashes = td["hash"]
```
kurtamohler, line 220: Looks like `LLMHashingEnv` currently uses the key "hashing", not "hash", so this currently fails in the generated tutorial: https://docs-preview.pytorch.org/pytorch/rl/2623/tutorials/beam_search_with_gpt.html#run-policy

vmoens: Good point!
I think we should avoid any "hash" that refers to a hashing value, bc it may override a built-in in some cases (eg `hash = do_smth()`)

PR #2641, opened 2024-12-06
[Feature] ChessEnv

vmoens:
* __->__ #2641

### DIFF
--- a/torchrl/envs/custom/chess.py
+++ b/torchrl/envs/custom/chess.py
@@ -1,0 +1,197 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+from typing import Optional
+
+import torch
+from tensordict import TensorDict, TensorDictBase
+from torchrl.data import Categorical, Composite, NonTensor, Unbounded
+
+from torchrl.envs import EnvBase
+
+from torchrl.envs.utils import _classproperty
+
+
+class ChessEnv(EnvBase):
+    """A chess environment that follows the TorchRL API.
+
+    Requires: the `chess` library. More info `here <https://python-chess.readthedocs.io/en/latest/>`__.
+
+    Args:
+        stateful (bool): Whether to keep track of the internal state of the board.
+            If False, the state will be stored in the observation and passed back
+            to the environment on each call. Default: ``False``.
+
+    .. note:: the action spec is a :class:`~torchrl.data.Categorical` spec with a ``-1`` shape.
+        Unless :meth:`~torchrl.data.Categorical.set_provisional_n` is called with the cardinality of the legal moves,
+        valid random actions cannot be taken. :meth:`~torchrl.envs.EnvBase.rand_action` has been adapted to account for
+        this behavior.
+
+    Examples:
+        >>> env = ChessEnv()
+        >>> r = env.reset()
+        >>> env.rand_step(r)
+        TensorDict(
+            fields={
+                action: Tensor(shape=torch.Size([]), device=cpu, dtype=torch.int64, is_shared=False),
+                done: Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.bool, is_shared=False),
+                fen: NonTensorData(data=rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1, batch_size=torch.Size([]), device=None),
+                hashing: Tensor(shape=torch.Size([]), device=cpu, dtype=torch.int64, is_shared=False),
+                next: TensorDict(
+                    fields={
+                        done: Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.bool, is_shared=False),
+                        fen: NonTensorData(data=rnbqkbnr/pppppppp/8/8/8/2N5/PPPPPPPP/R1BQKBNR b KQkq - 1 1, batch_size=torch.Size([]), device=None),
+                        hashing: Tensor(shape=torch.Size([]), device=cpu, dtype=torch.int64, is_shared=False),
+                        reward: Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.int32, is_shared=False),
+                        terminated: Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.bool, is_shared=False),
+                        turn: Tensor(shape=torch.Size([]), device=cpu, dtype=torch.bool, is_shared=False)},
+                    batch_size=torch.Size([]),
+                    device=None,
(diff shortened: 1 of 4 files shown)


torchrl/envs/custom/chess.py
```diff
@@ -0,0 +1,197 @@
...
+        # Collect data
+        fen = self.board.fen()
+        dest = tensordict.empty()
+        hashing = hash(fen)
+        dest.set("fen", fen)
+        dest.set("hashing", hashing)
+
+        done = board.is_checkmate()
+        turn = torch.tensor(board.turn)
+        reward = torch.tensor([done]).int() * (turn.int() * 2 - 1)
```
kurtamohler, line 184: Just a nit: I think this line is a little hard to read. Maybe something like this is a little easier to grasp?

```
winner = not board.turn
reward = torch.tensor([0 if not done else (-1 if winner == chess.BLACK else 1)])
```

vmoens: Since there are multiple ways of ending the game maybe we should be a bit more comprehensive there.:
Win = 1
Lose = -1
Stalemate ?

kurtamohler: I guess it's worth considering that usually in tournaments, if there's a checkmate or resignation, the winner gets 1 point and the loser gets 0, but if it's a draw of any kind, both players get 0.5 points.

I think whatever the values of win/lose are, the value of draw should be the average of win and lose

kurtamohler: Also chess evaluation engines, like stockfish, use 0 as a draw, negative score for a black advantage, and positive score for white advantage

vmoens: Ok so what about 0 / 0.5 / 1?

Also worth considering: more granular reward
https://www.restack.io/p/reinforcement-learning-answer-chess-bot-cat-ai
Since we have the opportunity of having multiple rewards, we could add another tensor that assign a reward for taking / losing pieces.

kurtamohler: I think that -1/0/1 is probably most consistent with existing chess engines, and it can encapsulate the value of the game for both black and white in a single number.

0/0.5/1 might also be good, but then I guess we'd need two rewards, one for black and one for white.

> a reward for taking / losing pieces

I think that could promote moves that do not optimize actually winning the game. For instance, if the reward decreases when losing a piece, then it could discourage making a sacrifice (or a combination of sacrifices) that leads to a forced checkmate. Likewise, if the reward increases when taking pieces, it could potentially encourage taking a piece rather than properly defending against an imminent checkmate threat.

Although it depends on what you want. Sometimes you don't want a chess bot to play optimally--like if it is meant to play against humans who are no match for the best chess bots

torchrl/envs/custom/chess.py
```diff
@@ -0,0 +1,197 @@
...
+        else:
+            dest = TensorDict()
+
+        if fen is None:
+            self.board.reset()
+            fen = self.board.fen()
+        else:
+            self.board.set_fen(fen.data)
+
+        hashing = hash(fen)
```
kurtamohler, line 142: If I understand correctly, `hash` is not guaranteed to give unique values for string inputs, since the number of possible strings is infinite, greater than the number of hash values. The output of `hash` only has 2^64 possible values.

```
>>> import sys
>>> sys.hash_info.width
64
```

Is uniqueness required for this hash?

vmoens: I think the questions to ask are

1. what are the chances that a collision will occur
2. what happens if a collision occurs

If we build a forest with - say - 10M elements, the number of combinations is still 10^13 times bigger than the capacity of the forest so I think it's safe to assume that the risk of failures to rebuild a tree due to hash collision is going to be small.

Another option on the safe size could also be to tokenize the fen. We could pad the tokens if they're shorter to make sure they all fit contiguously in memory.

Another question to solve is the one of reproducibility which worries me more than collision: if you restart your python process, the hash map will not hold anymore so any save data will be meaningless. IIRC there's a way to set the "seed" of the hash but that'd acting on a global variable which we may want to avoid anyway!

torchrl/envs/custom/chess.py
vmoens: I agree with all of this. With the `chess` lib, it's always white-centric no? ie, `win` means white wins, or is there a way to flip things around and be black-centric?

vmoens:
@kurtamohler I landed all the stack up until here.
If you want to give a shot at this feel free to do so!
You seem like a more skilled player than I am so feel free to edit things as you think it's best, I'll be happy to give it a look later!
Also we don't have tests but it'd be worth having some.


For the record here is the random forest maker I showed the other day
```
from torchrl.data import MCTSForest, Tree
import torch
from torchrl._utils import timeit
from torchrl.envs import ChessEnv
from tensordict import TensorDict, LazyStackedTensorDict, NonTensorData
env = ChessEnv()
f = MCTSForest(observation_keys=["hashing"], max_size=1000000, 
               excluded_keys="fen")

def make_random_forest():
    r = None
    reset = None
    i = 0
    while r is None or r.shape[-1] > 1:
        with timeit("rollout"):
            r = env.rollout(1000, tensordict=reset, auto_reset=reset is None)
        # r_ext = r.exclude("fen", ("next", "fen"))
        with timeit("extend"):
            f.extend(r)
        if i == 0:
            idx = 0
        else:
            idx = torch.randint(r.shape[-1], ())
        i += 1
        reset = r[idx]
    timeit.print()
make_random_forest()
t = f.get_tree(env.reset())
t.plot()
# etc
```

torchrl/envs/custom/chess.py
kurtamohler: True, at 10 million, the risk of collision is insignificant. I was curious where the limit is, so I looked into it. If my reasoning is correct, the risk starts to become significant around the order of 1 billion generated hashes. I think at 1 billion, the probability of collision is 2.7%. (I put up some notes [here](https://github.com/kurtamohler/notes/blob/a410bc6b9e627ab04e94c08fda66e2d3842000c2/math/hash-collision-probability.ipynb)).

So if we expect the tree to have significantly fewer than 1 billion nodes, then Python `hash` should be good enough I suppose.

kurtamohler: It seems like one option for reproducibility is to use `hashlib`, which I think also has features for generating unique hashes if we decide we need that

vmoens: I'm open to it!

kurtamohler:
> If you want to give a shot at this feel free to do so!

Ok I'll give it a shot!

It might be easiest if we merge this as-is first and then I can make a new PR with updates. I looked into whether ghstack has collaboration features, and it looks like it doesn't

vmoens:
Done! Thanks for that

PR #2648, opened 2024-12-13
[Feature] Add Hash transform and UnaryTransform

kurtamohler:
* __->__ #2648

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -2188,7 +2188,7 @@ class TestHash(TransformBase):
             hash_fn = hash
         elif datatype == "str":
             obs = "abcdefg"
-            hash_fn = Hash.reproducible_hash_parts
+            hash_fn = Hash.reproducible_hash
         elif datatype == "NonTensorStack":
             obs = torch.stack(
                 [
@@ -2199,9 +2199,7 @@ class TestHash(TransformBase):
             )
 
             def fn0(x):
-                return torch.stack(
-                    [Hash.reproducible_hash_parts(x_.get("data")) for x_ in x]
-                )
+                return torch.stack([Hash.reproducible_hash(x_.get("data")) for x_ in x])
 
             hash_fn = fn0
         else:
@@ -2311,9 +2309,9 @@ class TestHash(TransformBase):
                 in_keys=["string"],
                 out_keys=["hash"],
                 hash_fn=lambda x: torch.stack(
-                    [Hash.reproducible_hash_parts(x_.get("data")) for x_ in x]
+                    [Hash.reproducible_hash(x_.get("data")) for x_ in x]
                 ),
-                output_spec=Unbounded(shape=(2, 4), dtype=torch.int64),
+                output_spec=Unbounded(shape=(2, 32), dtype=torch.uint8),
             )
             base_env = CountingEnvWithString
 
@@ -2335,9 +2333,9 @@ class TestHash(TransformBase):
                 in_keys=["string"],
                 out_keys=["hash"],
                 hash_fn=lambda x: torch.stack(
-                    [Hash.reproducible_hash_parts(x_.get("data")) for x_ in x]
+                    [Hash.reproducible_hash(x_.get("data")) for x_ in x]
                 ),
-                output_spec=Unbounded(shape=(2, 4), dtype=torch.int64),
+                output_spec=Unbounded(shape=(2, 32), dtype=torch.uint8),
             )
             base_env = CountingEnvWithString
 
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -116,6 +116,7 @@ from torchrl.envs import (
     FrameSkipTransform,
     GrayScale,
     gSDENoise,
+    Hash,
     InitTracker,
     MultiStepTransform,
     NoopResetEnv,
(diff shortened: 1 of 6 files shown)


vmoens (commented):
Do we want to interact with [SipHash](https://github.com/pytorch/rl/blob/7d7cd9538d5e1adafceea2b77971f5f72b046708/torchrl/data/map/hash.py#L75) and other hash modules?

Or perhpas a tokenizer (not strictly a hash but the signature is similar for strings).

Or perhaps a custom hash function?

I'm happy with this being restrictive but if possible I'd prefer to avoid having multple transforms that do the str -> int map

kurtamohler:
Good questions. I agree that the module should allow the user to specify any hashing function or tokenizer they want to use, including `nn.Module`s like SipHash. In fact, I think all the transform really needs to do is this (along with adding the `out_keys` to the spec):

```
for in_key, out_key in zip(in_keys, out_keys):
    td[out_key] = user_specified_fn(td[in_key])
```

The `user_specified_fn` can be any unary function that knows how to handle whatever is in `td[in_key]`. However, the user would be free to specify a function that has nothing to do with hashing or tokenizing--it could even be something like `lambda x: x + 4`, so it doesn't seem right to call this a Hash transform.

It seems like what we really need is just a general key-wise `UnaryOpTransform`. I think its signature would be something like:

```
class UnaryOpTransform(Transform):
    def __init__(self, in_keys, out_keys, fn, output_spec): ...
```

In order to use this to implement a Python-hash transform for ChessEnv, for instance, we would just do:

```
chess_hash_t = UnaryOpTransform(
    in_keys=["fen"],
    out_keys=["hashing"],
    fn=hash,
    output_spec=Unbounded(shape=(), dtype=torch.int64)
)
```

We could even allow the user to optionally give inverse keys and a function for the inverse op. We could probably simplify the implementations of some of the existing transforms by making them inherit from `UnaryOpTransform`.

What do you think?

vmoens:
@kurtamohler I'm up to it up until `PythonHash`, but I'd be cautious about having a big fraction of the transforms all inherit from the same parent.
The design philosophy in the PT ecosystem is (usually) to avoid too many levels of inheritance to keep a clear API for the users, it makes it easier for people to hack their own transforms.

vmoens:
For the record, here is a script that makes it possible to use sha256 hashes (fewer collisions) in a reproducible manner (ie, sort of seeded)
```python
import hashlib


def reproducible_hash_parts(string, seed):
    """
    Creates a reproducible 256-bit hash from a string using a seed and splits it into four 64-bit parts.

    Args:
        string (str): The input string.
        seed (str): The seed value.

    Returns:
        tuple: Four 64-bit integers representing the parts of the 256-bit hash value.
    """
    # Prepend the seed to the string
    seeded_string = seed + string

    # Create a new SHA-256 hash object
    hash_object = hashlib.sha256()

    # Update the hash object with the seeded string
    hash_object.update(seeded_string.encode('utf-8'))

    # Get the hash value as bytes
    hash_bytes = hash_object.digest()

    # Split the hash bytes into four parts
    part1 = hash_bytes[:8]
    part2 = hash_bytes[8:16]
    part3 = hash_bytes[16:24]
    part4 = hash_bytes[24:]

    # Convert each part to a 64-bit integer
    part1_value = int.from_bytes(part1, 'big')
    part2_value = int.from_bytes(part2, 'big')
    part3_value = int.from_bytes(part3, 'big')
    part4_value = int.from_bytes(part4, 'big')

    return part1_value, part2_value, part3_value, part4_value


# Example usage:
string = "Hello, World!"
seed = "my_seed"

part1, part2, part3, part4 = reproducible_hash_parts(string, seed)
print(f"Part 1: {part1}")
print(f"Part 2: {part2}")
print(f"Part 3: {part3}")
print(f"Part 4: {part4}")
```

vmoens:
Another random thought: we could add the option to store a table of hash-to-value within the transform
```python
class HashTransform(...):
    _hash_table: Dict[HashType, str]
    ...
```
and include that in the transform state-dict (or make it easy to save this for future use)

kurtamohler:
Maybe I should add `reproducible_hash_parts` as a class method on the `Hash` transform, so it can be invoked like `t = Hash(..., fn=Hash.reproducible_hash_parts)`?

> Another random thought: we could add the option to store a table of hash-to-value within the transform

What's the purpose you have in mind for that? Is there a reason someone would want to look up the value that corresponds to a hash if that pairing already exists in the output tensordict? Or would this be used to detect hash collision? If there is good reason to have this map, I think we'd have to use weakrefs to the values so that we don't prevent them from being deallocated when all the other references are deleted

vmoens:
The goal I had in mind was to keep track of the existing map for later use, although if `hash("stuff")` doesn't return the same result across runs it's of little use... 

Maybe we should default to hashlib for reproducibility

torchrl/envs/transforms/transforms.py
```diff
@@ -4400,6 +4401,146 @@ def __repr__(self) -> str:
...
+        part4 = hash_bytes[24:]
+
+        # Convert each part to a 64-bit integer
+        part1_value = int.from_bytes(part1, "big", signed=True)
+        part2_value = int.from_bytes(part2, "big", signed=True)
+        part3_value = int.from_bytes(part3, "big", signed=True)
+        part4_value = int.from_bytes(part4, "big", signed=True)
+
+        return torch.tensor(
+            [part1_value, part2_value, part3_value, part4_value], dtype=torch.int64
```
kurtamohler, line 4540: The choice to split it into 4 longs seems kind of arbitrary. I'm wondering if it would be better to use `torch.frombuffer` with `dtype=torch.int8`

vmoens: Oh yeah, I think that'd be better! We can even use `tensor.view` after that and get int64s if we want to

torchrl/envs/transforms/transforms.py
```diff
@@ -4400,6 +4401,132 @@ def __repr__(self) -> str:
...
 
+class UnaryTransform(Transform):
+    """Applies a unary operation on the specified inputs.
+
+    Args:
+        in_keys (sequence of NestedKey): the keys of inputs to the unary operation.
+        out_keys (sequence of NestedKey): the keys of the outputs of the unary operation.
+        fn (Callable, optional): the function to use as the unary operation.
+        output_spec (TensorSpec, optional): the spec of the output of the operation.
+    """
```
vmoens, line 4412:
```suggestion
    """Applies a unary operation on the specified inputs.

    Args:
        in_keys (sequence of NestedKey): the keys of inputs to the unary operation.
        out_keys (sequence of NestedKey): the keys of the outputs of the unary operation.
        fn (Callable): the function to use as the unary operation.
        output_spec (TensorSpec): the spec of the output of the operation.
    """
```

The output spec could perhaps be optional, when transforming the spec, we could use `spec.zero()`, apply `_fn` and then recast the output to a spec (see `torchrl.envs.utils.make_composite_from_td`)

Also, it seems we're transforming all specs but I'm not sure why (perhaps bc we capture the keys using in/out-keys)? What is the spec of one in-key doesn't match the result of another? Maybe `out_spec` could be a composite spec in that case.

kurtamohler: I like the idea of using `make_composite_from_td`. Not having to specify the output spec would make it easier to use. However,  `make_composite_from_td` assumes all values are unbounded, and maybe the user would want it to be bounded. So I guess we probably would still need to allow the user to specify the output spec if they want?

> it seems we're transforming all specs but I'm not sure why

I figured it would be ideal to allow each of the `in_keys` to be in any of the output specs, rather than just observations. I also figured that it could be annoying if the user had to specify each of the reward/obs/done specs as three separate arguments. So I made UnaryTransform detect whether a particular `in_key` is part of the observation, reward, or done spec. Does that make sense?

> What is the spec of one in-key doesn't match the result of another? Maybe out_spec could be a composite spec in that case.

Good point. I didn't consider that, but we should support that case. Yes, we could allow `output_spec` to be composite and require that it has an entry for each of the `in_keys` in that case. I'll make that change

vmoens: Cool! I didn't really get why we can't automatically build the spec though, having access to the spec from the env should be enough to create some fake data and extrapolate a spec from it no? (At least in the case of Hash transform)

kurtamohler: I'm on board with creating the spec automatically, for both Hash and UnaryTransform. I was thinking there might be a reason why the user would want to specify it manually, but I don't have any concrete use case, so I suppose it's probably best to remove the `output_spec` argument for now and only add it back in later if we find a need for it

test/test_transforms.py
```diff
@@ -2180,6 +2183,246 @@ def test_transform_no_env(self, device, batch):
...
+            obs = "abcdefg"
+            hash_fn = Hash.reproducible_hash
+        elif datatype == "NonTensorStack":
+            obs = torch.stack(
+                [
+                    NonTensorData(data="abcde"),
+                    NonTensorData(data="fghij"),
+                    NonTensorData(data="klmno"),
+                ]
+            )
```
vmoens, line 2202: in the future you can also do `NonTensorStack("a", "b", "c")`

PR #2661, opened 2024-12-17
[Test] Add tests and a few fixes for ChessEnv

kurtamohler:
* __->__ #2661

### DIFF
--- a/torchrl/envs/custom/chess.py
+++ b/torchrl/envs/custom/chess.py
@@ -127,10 +127,13 @@ class ChessEnv(EnvBase):
         self._set_action_space(tensordict)
         return super().rand_action(tensordict)
 
+    def _is_done(self, board):
+        return board.is_game_over() | board.is_fifty_moves()
+
     def _reset(self, tensordict=None):
         fen = None
         if tensordict is not None:
-            fen = self._get_fen(tensordict)
+            fen = self._get_fen(tensordict).data
             dest = tensordict.empty()
         else:
             dest = TensorDict()
@@ -139,7 +142,11 @@ class ChessEnv(EnvBase):
             self.board.reset()
             fen = self.board.fen()
         else:
-            self.board.set_fen(fen.data)
+            self.board.set_fen(fen)
+            if self._is_done(self.board):
+                raise ValueError(
+                    "Cannot reset to a fen that is a gameover state." f" fen: {fen}"
+                )
 
         hashing = hash(fen)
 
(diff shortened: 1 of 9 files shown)


torchrl/envs/custom/chess.py
```diff
@@ -127,10 +127,13 @@ def rand_action(self, tensordict: Optional[TensorDictBase] = None):
         self._set_action_space(tensordict)
         return super().rand_action(tensordict)
 
+    def _is_done(self, board):
+        return board.is_game_over() | board.is_fifty_moves()
```
kurtamohler, line 131: There's a pretty important rule that we're neglecting at the moment, which is that if a particular position happens three times in a game, then it's a draw.

The `chess` library has a [feature](https://python-chess.readthedocs.io/en/latest/core.html#chess.Board.is_repetition) to check for this, but it requires the `chess.Board` to have the entire history of the game. Since we're constantly calling `Board.set_fen`, we're erasing the history, so we don't have a way to detect repetition at the moment.

If the main purpose of ChessEnv is to prove MCTS works, then adding this rule may not be worth the effort. But if we specifically want to follow all the standard rules, then we'd have to figure out a way to add it

vmoens: Very good point thanks for flagging it!
I'd be in favor of documenting it in the docstrings for now. 

One option could be to use a state that is a sequence of all previous fen (like we do with LLMs where a state is all the prompt tokens + all generated tokens) and manually check for duplicates. 

I see that chess also has a the option of exporting PGNs but it's a bit convoluted because we'd need to create a 
```python
import chess.pgn

# Create a new Board object
board = chess.Board()

# Make some moves
board.push_uci("e2e4")
board.push_uci("e7e5")
board.push_uci("d2d4")

# Create a new Game object
game = chess.pgn.Game()

# Add the moves to the game
node = game
for move in board.move_stack:
    node = node.add_variation(move)

# Generate the PGN string
pgn_string = str(game)

print(pgn_string)
```
which gives us
```
[Event "?"]
[Site "?"]
[Date "????.??.??"]
[Round "?"]
[White "?"]
[Black "?"]
[Result "*"]

1. e4 e5 2. d4 *
```

So in practice, we could be using `fen` or `pgn` (or else?) as a state representation

```python
env = ChessEnv(representation="fen")
```

I believe there should be a way to make these interchangeable (at the end of the day it's still string-in string-out)

vmoens (approved):
Fantastic I love the shape it's taking
We should add a kwarg for representing the board as a tensor, I will add it later!

torchrl/envs/custom/chess.py
kurtamohler: Oh yes, using a PGN seems like a good solution

PR #2668, opened 2024-12-19
[BugFix] Compatibility of tensordict primers with batched envs (specifically for LSTM and GRU)

vmoens:
* #2674
* #2672
* #2671
* #2670
* __->__ #2668

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -21,11 +21,6 @@ on:
     branches:
       - "nightly"
 
-env:
-  ACTIONS_RUNNER_FORCED_INTERNAL_NODE_VERSION: node16
-  ACTIONS_RUNNER_FORCE_ACTIONS_NODE_VERSION: node16
-  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true  # https://github.com/actions/checkout/issues/1809
-
 concurrency:
   # Documentation suggests ${{ github.head_ref }}, but that's only available on pull_request/pull_request_target triggers, so using ${{ github.ref }}.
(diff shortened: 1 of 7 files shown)


vmoens:
cc @matteobettini

vmoens:
This PR works with https://github.com/pytorch/tensordict/pull/1150

vmoens:
@albertbou92 THere is an annoying behaviour that if the composite spec in the primer can be set to the parent shape it is, oterwise it is expanded. I didn't realize how unstable this was: this basically means that if your spec has shape [2] and by chance you're using a parallel env with 2 workers, you don't expand, but if you change the number of workers you do... 
We should not do this. I think the right API should be
```python
# 1. No indication of shape
primer=TensorDictPrimer(stuff=spec) # no batch_size
parallel_env.append_transform(primer) # expands the specs

# 2. Using a composite spec to indicate the shape
primer=TensorDictPrimer(composite_spec)
parallel_env.append_transform(primer) # expands the spec shape only if the composite shape differs from the env shape

# 3. No indication of shape
primer=TensorDictPrimer(stuff=spec, batch_size=[n_envs]) # batch_size explicitly set
parallel_env.append_transform(primer) # expands the specs
```

It's going to be hard to transition to that behaviour but I think it's worth it.

So what we need is that for the case (1), if the shape can be reset, we warn users that we are changing the spec shape but tht will not be the case anymore in v0.8 (I'm doing just one release cycle here because the behaviour is pretty bad).
If the shape cannot be set, we expand as we should.

vmoens:
> What about having a boolean paramater in TensorDictPrimer that controls whether or not specs batch size is expanded to the parent batch size

that's exactly what I implemented here! glad you agree it's a good idea

PR #2677, opened 2024-12-26
[Feature] multiagent data standardization: PPO advantages

matteobettini:
Hello!

### Context

So, in this library we frequently apply standardizations over data (also called normalization) where we subtract the mean from the data and divide it by a standard deviation.

A few places where we do this that come to mind are in the transforms (where it can be used over observations and rewards) and in the advantages in ppo. Please let me know about other places if you can think of more.

### Issue

The issue is that normally the mean and std statistics are computed over all the dimensions. But in multiagent settings, where for example a reward tensor could have an "agent" dimension, this is not the proper way to do it as different agents might have different reward distributions and applying standardization naively across them will create issues.

### Example

For example consider the case where we have 2 agents each with 2 rewards (agent_1_r = [100,200],  agent_2_r = [0,-1]), structured in a 2x2 tensor `r = torch.tensor([[100,200],[0,-1]])`

If we standardize using `mean(r)` and `std(r)`, we get `r = torch.tensor([[0.3,1.5],[-0.9,-.91]])`, which is bad

If we standardize using `mean(r, dim=1)` and `std(r,dim=1)`, hence excluding the agent dim from the statistics, we get `r = torch.tensor([[-1,1],[1,-1]])`, which is better as it does not cross pollute over agents

### Proposal 

We should adapt all places in the library where standardization is used to take a list of dimensions to exlude from the statistics. This way, multiagent users can provide the agent dimension as input and use the standardization feature

### What I did here

I did this for the advantages in PPO

To do it, I made a nice standardization util function that takes as input a list of dimensions to exlude from the standardization (a more flexible implementation of a functional layer norm)

### DIFF
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -892,14 +892,23 @@ def _standardize(
 
     """
     if eps is None:
-        eps = torch.finfo(torch.float.dtype).resolution
+        if input.dtype.is_floating_point:
+            eps = torch.finfo(torch.float).resolution
+        else:
+            eps = 1e-6
 
     len_exclude_dims = len(exclude_dims)
     if not len_exclude_dims:
         if mean is None:
             mean = input.mean()
+        else:
+            # Assume dtypes are compatible
+            mean = torch.as_tensor(mean, device=input.device)
         if std is None:
             std = input.std()
+        else:
+            # Assume dtypes are compatible
+            std = torch.as_tensor(std, device=input.device)
         return (input - mean) / std.clamp_min(eps)
 
     input_shape = input.shape
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -32,7 +32,7 @@ from packaging.version import parse
 from tensordict import unravel_key
 
 from tensordict.utils import NestedKey
-from torch import multiprocessing as mp
+from torch import multiprocessing as mp, Tensor
 
 try:
     from torch.compiler import is_compiling
@@ -873,7 +873,11 @@ class _ContextManager:
 
 
 def _standardize(
-    input, exclude_dims: Tuple[int] = (), mean=None, std=None, eps: float = None
+    input: Tensor,
+    exclude_dims: Tuple[int] = (),
+    mean: Tensor | None = None,
+    std: Tensor | None = None,
+    eps: float | None = None,
 ):
     """Standardizes the input tensor with the possibility of excluding specific dims from the statistics.
 
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -24,7 +24,7 @@ from copy import copy
 from distutils.util import strtobool
 from functools import wraps
 from importlib import import_module
-from typing import Any, Callable, cast, Dict, Sequence, TypeVar, Union
+from typing import Any, Callable, cast, Dict, Tuple, TypeVar, Union
 
 import numpy as np
 import torch
@@ -872,7 +872,9 @@ class _ContextManager:
             self._mode = type
 
 
-def _standardize(input, exclude_dims: Sequence[int] = (), mean=None, std=None):
+def _standardize(
+    input, exclude_dims: Tuple[int] = (), mean=None, std=None, eps: float = None
+):
     """Standardizes the input tensor with the possibility of excluding specific dims from the statistics.
 
     Useful when processing multi-agent data to keep the agent dimensions independent.
@@ -882,8 +884,12 @@ def _standardize(input, exclude_dims: Sequence[int] = (), mean=None, std=None):
         exclude_dims (Sequence[int]): dimensions to exclude from the statistics, can be negative. Default: ().
         mean (Tensor): a mean to be used for standardization. Must be of shape broadcastable to input. Default: None.
         std (Tensor): a standard deviation to be used for standardization. Must be of shape broadcastable to input. Default: None.
+        eps (float): epsilon to be used for numerical stability. Default: float32 resolution.
 
     """
+    if eps is None:
+        eps = torch.finfo(torch.float.dtype).resolution
+
     input_shape = input.shape
     exclude_dims = [
         d if d >= 0 else d + len(input_shape) for d in exclude_dims
@@ -893,7 +899,7 @@ def _standardize(input, exclude_dims: Sequence[int] = (), mean=None, std=None):
         raise ValueError("Exclude dims has repeating elements")
     if any(dim < 0 or dim >= len(input_shape) for dim in exclude_dims):
         raise ValueError(
-            f"exclude_dims provided outside bounds for input of shape={input_shape}"
+            f"exclude_dims={exclude_dims} provided outside bounds for input of shape={input_shape}"
         )
     if len(exclude_dims) == len(input_shape):
         warnings.warn(
@@ -901,11 +907,14 @@ def _standardize(input, exclude_dims: Sequence[int] = (), mean=None, std=None):
         )
         return input
 
-    # Put all excluded dims in the beginning
-    permutation = list(range(len(input_shape)))
-    for dim in exclude_dims:
-        permutation.insert(0, permutation.pop(permutation.index(dim)))
-    permuted_input = input.permute(*permutation)
+    if len(exclude_dims):
+        # Put all excluded dims in the beginning
+        permutation = list(range(len(input_shape)))
+        for dim in exclude_dims:
+            permutation.insert(0, permutation.pop(permutation.index(dim)))
+        permuted_input = input.permute(*permutation)
+    else:
+        permuted_input = input
     normalized_shape_len = len(input_shape) - len(exclude_dims)
 
     if mean is None:
@@ -916,11 +925,15 @@ def _standardize(input, exclude_dims: Sequence[int] = (), mean=None, std=None):
         std = torch.std(
             permuted_input, keepdim=True, dim=tuple(range(-normalized_shape_len, 0))
         )
-    output = (permuted_input - mean) / std.clamp_min(1e-6)
+    output = (permuted_input - mean) / std.clamp_min(eps)
 
     # Reverse permutation
-    inv_permutation = torch.argsort(torch.LongTensor(permutation)).tolist()
-    output = torch.permute(output, inv_permutation)
+    if len(exclude_dims):
+        inv_permutation = torch.argsort(
+            torch.tensor(permutation, dtype=torch.long, device=input.device)
+        ).tolist()
+        output = torch.permute(output, inv_permutation)
+
     return output
 
 
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -919,31 +919,12 @@ def _standardize(
         )
         return input
 
-    # Put all excluded dims in the beginning
-    permutation = list(range(len(input_shape)))
-    for dim in exclude_dims:
-        permutation.insert(0, permutation.pop(permutation.index(dim)))
-    permuted_input = input.permute(*permutation)
-
-    normalized_shape_len = len(input_shape) - len_exclude_dims
-
+    included_dims = tuple(d for d in range(len(input_shape)) if d not in exclude_dims)
     if mean is None:
-        mean = torch.mean(
-            permuted_input, keepdim=True, dim=tuple(range(-normalized_shape_len, 0))
-        )
+        mean = torch.mean(input, keepdim=True, dim=included_dims)
     if std is None:
-        std = torch.std(
-            permuted_input, keepdim=True, dim=tuple(range(-normalized_shape_len, 0))
-        )
-    output = (permuted_input - mean) / std.clamp_min(eps)
-
-    # Reverse permutation
-    inv_permutation = torch.argsort(
-        torch.tensor(permutation, dtype=torch.long, device=input.device)
-    ).tolist()
-    output = torch.permute(output, inv_permutation)
-
-    return output
+        std = torch.std(input, keepdim=True, dim=included_dims)
+    return (input - mean) / std.clamp_min(eps)
 
 
 @wraps(torch.compile)
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -911,7 +911,7 @@ def _standardize(
         )
     if len_exclude_dims == len(input_shape):
         warnings.warn(
-            "standardize called but all dims were excluded from the statistics, returning unprocessed input"
+            "_standardize called but all dims were excluded from the statistics, returning unprocessed input"
         )
         return input
 
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -24,7 +24,7 @@ from copy import copy
 from distutils.util import strtobool
 from functools import wraps
 from importlib import import_module
-from typing import Any, Callable, cast, Dict, TypeVar, Union
+from typing import Any, Callable, cast, Dict, Sequence, TypeVar, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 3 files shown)


torchrl/objectives/ppo.py
```diff
@@ -311,6 +314,7 @@ def __init__(
         critic_coef: float = 1.0,
         loss_critic_type: str = "smooth_l1",
         normalize_advantage: bool = False,
+        normalize_advantage_exclude_dims: Sequence[int] = (),
```
vmoens, line 317:
```suggestion
        normalize_advantage_exclude_dims: Tuple[int] = (),
```

torchrl/objectives/ppo.py
```diff
@@ -87,6 +88,8 @@ class PPOLoss(LossModule):
             Can be one of "l1", "l2" or "smooth_l1". Defaults to ``"smooth_l1"``.
         normalize_advantage (bool, optional): if ``True``, the advantage will be normalized
             before being used. Defaults to ``False``.
+        normalize_advantage_exclude_dims (Sequence[int], optional): dimensions to exclude from the advantage
```
vmoens, line 91:
```suggestion
        normalize_advantage_exclude_dims (Tuple[int], optional): dimensions to exclude from the advantage
```

torchrl/objectives/ppo.py
```diff
@@ -87,6 +88,8 @@ class PPOLoss(LossModule):
...
+            standardization, can be negative. Useful in multiagent settings to exlude the agent dimension. Default: ().
```
vmoens, line 92:
```suggestion
            standardization. Negative dimensions are valid. This is useful in multiagent settings where the agent dimension may be excluded from the reductions.
            Defaults to ``()``.
```

torchrl/objectives/ppo.py
```diff
@@ -8,7 +8,7 @@
 
 from copy import deepcopy
 from dataclasses import dataclass
-from typing import Tuple
+from typing import Sequence, Tuple
```
vmoens, line 11: Perhaps let's restrict it to tuple? Having a sequence seems just a tiny bit more complex, and tuples seem to be the right class for this.

```suggestion
from typing import Tuple
```

torchrl/objectives/ppo.py
```diff
@@ -711,6 +715,8 @@ class ClipPPOLoss(PPOLoss):
             Can be one of "l1", "l2" or "smooth_l1". Defaults to ``"smooth_l1"``.
         normalize_advantage (bool, optional): if ``True``, the advantage will be normalized
             before being used. Defaults to ``False``.
+        normalize_advantage_exclude_dims (Sequence[int], optional): dimensions to exclude from the advantage
```
vmoens, line 718:
```suggestion
        normalize_advantage_exclude_dims (Tuple[int], optional): dimensions to exclude from the advantage
```

torchrl/objectives/ppo.py
```diff
@@ -802,6 +808,7 @@ def __init__(
         critic_coef: float = 1.0,
         loss_critic_type: str = "smooth_l1",
         normalize_advantage: bool = False,
+        normalize_advantage_exclude_dims: Sequence[int] = (),
```
vmoens, line 811:
```suggestion
        normalize_advantage_exclude_dims: Tuple[int] = (),
```

torchrl/objectives/ppo.py
```diff
@@ -711,6 +715,8 @@ class ClipPPOLoss(PPOLoss):
             Can be one of "l1", "l2" or "smooth_l1". Defaults to ``"smooth_l1"``.
         normalize_advantage (bool, optional): if ``True``, the advantage will be normalized
             before being used. Defaults to ``False``.
+        normalize_advantage_exclude_dims (Sequence[int], optional): dimensions to exclude from the advantage
+            standardization, can be negative. Useful in multiagent settings to exlude the agent dimension. Default: ().
```
vmoens, line 719:
```suggestion
            standardization. Negative dimensions are valid. This is useful in multiagent settings where the agent dimension may be excluded from the reductions.
            Defaults to ``()``.
```

torchrl/_utils.py
```diff
@@ -872,6 +872,58 @@ def set_mode(self, type: Any | None) -> None:
...
+    input_shape = input.shape
+    exclude_dims = [
+        d if d >= 0 else d + len(input_shape) for d in exclude_dims
+    ]  # Make negative dims positive
+
+    if len(set(exclude_dims)) != len(exclude_dims):
+        raise ValueError("Exclude dims has repeating elements")
+    if any(dim < 0 or dim >= len(input_shape) for dim in exclude_dims):
+        raise ValueError(
+            f"exclude_dims provided outside bounds for input of shape={input_shape}"
```
vmoens, line 896: The dim should also be printed here

torchrl/_utils.py
```diff
@@ -872,6 +872,58 @@ def set_mode(self, type: Any | None) -> None:
             self._mode = type
 
 
+def _standardize(input, exclude_dims: Sequence[int] = (), mean=None, std=None):
```
vmoens, line 875: There is a bit of extra compute here compared to before which every run will incur, even if `exclude_dim` is empty. I would treat that as a special case.

torchrl/_utils.py
```diff
@@ -872,6 +872,58 @@ def set_mode(self, type: Any | None) -> None:
...
+
+    if mean is None:
+        mean = torch.mean(
+            permuted_input, keepdim=True, dim=tuple(range(-normalized_shape_len, 0))
+        )
+    if std is None:
+        std = torch.std(
+            permuted_input, keepdim=True, dim=tuple(range(-normalized_shape_len, 0))
+        )
+    output = (permuted_input - mean) / std.clamp_min(1e-6)
```
vmoens, line 919: For `1e-6` we have two options: retrieve the resolution of the dtype, or have it as a kwarg (or a mixture of both: retrieve from dtype if not passed as a kwarg).

torchrl/_utils.py
```diff
@@ -872,6 +872,58 @@ def set_mode(self, type: Any | None) -> None:
...
+
+    # Reverse permutation
+    inv_permutation = torch.argsort(torch.LongTensor(permutation)).tolist()
```
vmoens, line 922: `torch.LongTensor(permutation)` should not be used, use `.to(torch.long)` instead or `torch.tensor(..., dtype=torch.long)`.
The device should also be controlled (do we always want this to be executed on the default device? We could use the device of the input to do this op).

torchrl/objectives/ppo.py
vmoens: This seems like an indirect argument to me and it could be hard for users to understand.
If you're using PPO with this kwarg, you need to think this way:
1. I asked to normalize the advantage
2. My reward has one or more agent dims
3. I want to excldue these from the advantage normalization.
Another way to do this could be to have a kwarg
```
reward_agent_dim: int | Tuple[int] | None = None,
```
If this is passed, the `reward_agent_dim` will be kept during calls to `mean` / `std`. 
Per se, it changes nothing but it's a tiny bit more clear from the name and signature that this indicates a dim that is not a feature dim.

torchrl/objectives/ppo.py
vmoens: I have a question reguarding the "negative" dims.

I guess ultimately we would also like to do the same with this
https://github.com/pytorch/rl/blob/d009835b4fccd1482e9a2bd2b597e00824b95c2d/torchrl/objectives/value/advantages.py#L696-L701

But there we have an extra "time" dimension, so we must be careful that `agent_dim=(-2,)` may actually mean `(-3,)` in some cases (similarly, `feature_dim=(4,)` may be `feature_dim=(5,)`).

Is it an issue in this (ie, PPO) case?

torchrl/objectives/ppo.py
```diff
@@ -1048,6 +1056,7 @@ def __init__(
         critic_coef: float = 1.0,
         loss_critic_type: str = "smooth_l1",
         normalize_advantage: bool = False,
+        normalize_advantage_exclude_dims: Sequence[int] = (),
```
vmoens, line 1059:
```suggestion
        normalize_advantage_exclude_dims: Tuple[int] = (),
```

torchrl/objectives/ppo.py
```diff
@@ -955,6 +961,8 @@ class KLPENPPOLoss(PPOLoss):
             Can be one of "l1", "l2" or "smooth_l1". Defaults to ``"smooth_l1"``.
         normalize_advantage (bool, optional): if ``True``, the advantage will be normalized
             before being used. Defaults to ``False``.
+        normalize_advantage_exclude_dims (Sequence[int], optional): dimensions to exclude from the advantage
+            standardization, can be negative. Useful in multiagent settings to exlude the agent dimension. Default: ().
```
vmoens, line 965: ditto

vmoens (commented):
I can definitely see why this is a problem, thanks for pointing it out and coming up with a solution!

My first thought when looking at this PR is that it can easily be forgotten. 
IMO the ideal API is, by decreasing order of desiderability 
1. The API doesn't break with edge cases (I'm using a overly generic term here, not saying MARL is edge case :) ) but takes care of them automatically. That can be achieved with a trainer class for instance.
2. The API breaks when encoutering an edge case and the error message tells the user exactly why and what to do
3. The run goes through but a warning is raised.
Finally, 4., the run goes through and things are silently wrong. 

I think this is a (4). The ideal solution would be to do it automatically (1) but I'm not sure how we can achieve that with the current API. Any thoughts on how to make it a (2) or (3)?
(As a counter-example, if your data has some specific keys, the loss will simply not work until you call `loss_module.set_keys(**tensor_keys)`. We're not doing it automatically but there is very little chance that this step is forgotten and we can capture the errors and give an informative error message to let people know what to do.)

Another way of saying this is: there are two levels of complexity that make a lib hard to use and I think both are present with the current solution:
1. (weak) There is something that users need to know when they use the library but they may not be aware of that and the lib isn't telling them at runtime;
2. (strong) The lib doesn't take care automaticlly of edge cases and it's up to the user to come up with a solution.
How can we make sure that users are not assuming that this is done automatically when they say "normalize_advantage=True`? Is there a quick check that we can do? 

One thing I have in mind is to look at the reward shape and detecting if the last dim is a singleton (if not, the reward is probably something that isn't from regular single agent)? In principle, anything with shape that is not `(*tensordict.shape, 1)` should have a dedicated reward normalization, no?

Regarding other parts of the lib where this could be needed:

- I think we also need to address this issue with the advantages where there's a lot of normalization of rewards etc.
  https://github.com/pytorch/rl/blob/d009835b4fccd1482e9a2bd2b597e00824b95c2d/torchrl/objectives/value/advantages.py#L1434-L1438
  https://github.com/pytorch/rl/blob/d009835b4fccd1482e9a2bd2b597e00824b95c2d/torchrl/objectives/value/advantages.py#L1804-L1808
- VecNorm needs a big refactoring (it's clunky and inefficient atm), so we can put this in the TODO list.
- For ObservationNorm, you could pass a loc / scale with any broadcastable shape and it should work. 
  The `ObservationNorm.init_stats` should also have the required keyword arguments to account for the example you gave.

torchrl/objectives/ppo.py
matteobettini: I think the user should pass the dimension index for the tensor that goes in the class.

If you give as input [batch,time,agents,n] then agents is -2

also agents is usually after time

matteobettini:
Thanks so much for the in-depth reveiw! I agree! 

Let me trying t give some answers, I'll start from some general clarifications

**This is not a feature purely for multiagent uses.** The need for this feature could be various. Imagine a multi-objective rl setting where you have an array of rewards. Now, even in that case you would not want to mix rewards during normalization. You could also have mutli-objective + multi-agent. So, the goal of this PR is more like saying that using `tensor.mean()` in the lib is a tight fit as there are a lot of usecases where users might want to keep some dimensions independent while computing statistics. So the idea is to provide a way for users to communicate that, regardelss of the usecase. For this, I don't think that naming it something like "agent_dimension" is appropriate as the poor multi-objective usecases or others are not represented

**I agree this should be automatic.** The problem is that it is quite impossible to automatically know what dimensions the user wants to keep independent. What we could do to turn it into a (3) is what you suggest: if  you are asking for normalization and your shape is more than (*tensordict.shape, 1), we can give you a warning saying "hey! you have all those extra dims that we are considering in our statistics, any of those independent?"

torchrl/objectives/ppo.py
matteobettini: See my discussion point on this being useful for more uscases. We can come up with a better name together

matteobettini:
I should have addressed all comments.

Anorther thing I forgot to say is that the `_standardize` ultil function is meant to be used in all those other cases we need to fix you mentioned

vmoens:
Thanks for the answer!

One last thing: If users don't provide a `normalize_advantage_exclude_dims`, I suspect the result will be identical to what it used to be, yet the `_standardize` function will do a lot more stuff. Can we streamline this (avoid any extra compute when the normalization is simple)?

matteobettini:
I thought I had addressed that. When the exclude dims are not provided, all the permutation stuff is skipped. What is the “lot more stuff”?

torchrl/_utils.py
```diff
@@ -872,6 +872,71 @@ def set_mode(self, type: Any | None) -> None:
...
+        raise ValueError("Exclude dims has repeating elements")
+    if any(dim < 0 or dim >= len(input_shape) for dim in exclude_dims):
+        raise ValueError(
+            f"exclude_dims={exclude_dims} provided outside bounds for input of shape={input_shape}"
+        )
+    if len(exclude_dims) == len(input_shape):
+        warnings.warn(
+            "standardize called but all dims were excluded from the statistics, returning unprocessed input"
+        )
+        return input
```
vmoens, line 908: All of this can be skipped if `exclude_dims` is empty

torchrl/_utils.py
```diff
@@ -872,6 +872,71 @@ def set_mode(self, type: Any | None) -> None:
...
+
+    if len(exclude_dims):
+        # Put all excluded dims in the beginning
+        permutation = list(range(len(input_shape)))
+        for dim in exclude_dims:
+            permutation.insert(0, permutation.pop(permutation.index(dim)))
+        permuted_input = input.permute(*permutation)
+    else:
+        permuted_input = input
+    normalized_shape_len = len(input_shape) - len(exclude_dims)
```
vmoens, line 918: This doesn't need to be computed if `exclude_dims` is empty

torchrl/_utils.py
```diff
@@ -872,6 +872,71 @@ def set_mode(self, type: Any | None) -> None:
...
+
+    if mean is None:
+        mean = torch.mean(
+            permuted_input, keepdim=True, dim=tuple(range(-normalized_shape_len, 0))
+        )
+    if std is None:
+        std = torch.std(
+            permuted_input, keepdim=True, dim=tuple(range(-normalized_shape_len, 0))
+        )
```
vmoens, line 927: mean and std have a simpler call if `exclude_dims` is empty

matteobettini:
ok, now first thing i handle the case for no excluded dims

torchrl/_utils.py
```diff
@@ -872,6 +872,80 @@ def set_mode(self, type: Any | None) -> None:
...
+    output = (permuted_input - mean) / std.clamp_min(eps)
+
+    # Reverse permutation
+    inv_permutation = torch.argsort(
+        torch.tensor(permutation, dtype=torch.long, device=input.device)
+    ).tolist()
+    output = torch.permute(output, inv_permutation)
+
+    return output
```
vmoens, line 946: I don't think we need to permute anything. `v2` is slightly faster than `v1` here and more straightforward
```python
import torch
eps = 1e-6
def v1(input, exclude_dims):
    input_shape = input.shape
    len_exclude_dims = len(exclude_dims)
    # Put all excluded dims in the beginning
    permutation = list(range(len(input_shape)))
    for dim in exclude_dims:
        permutation.insert(0, permutation.pop(permutation.index(dim)))
    permuted_input = input.permute(*permutation)

    normalized_shape_len = len(input_shape) - len_exclude_dims

    mean = torch.mean(
        permuted_input, keepdim=True, dim=tuple(range(-normalized_shape_len, 0))
    )
    std = torch.std(
        permuted_input, keepdim=True, dim=tuple(range(-normalized_shape_len, 0))
    )
    output = (permuted_input - mean) / std.clamp_min(eps)

    # Reverse permutation
    inv_permutation = torch.argsort(
        torch.tensor(permutation, dtype=torch.long, device=input.device)
    ).tolist()
    output = torch.permute(output, inv_permutation)

    return output

def v2(input, exclude_dims):
    input_shape = input.shape
    len_exclude_dims = len(exclude_dims)
    dims = tuple(d for d in range(len(input_shape)) if d not in exclude_dims)
    mean = torch.mean(
        input, keepdim=True, dim=dims
    )
    std = torch.std(
        input, keepdim=True, dim=dims
    )
    return (input - mean) / std.clamp_min(eps)

input = torch.randn(10, 20, 30, 40)
exclude_dims = (0, 2)

%%timeit
v1(input, exclude_dims)

%%timeit
v2(input, exclude_dims)

torch.testing.assert_close(v1(input, exclude_dims), v2(input, exclude_dims))
```

matteobettini: Sure, makes sense! I can switch to that

vmoens (approved):
LGTM

Added tests and fixed a couple of bugs in 121bbc16b85f92d37b46f1c9dae6efd383871ba8

PR #2681, opened 2025-01-03
[Feature] Linearise reward transform

louisfaury:
## Description
Adds a transform to linearise multi-objective rewards into a single one, via a weighted sum. Essentially, this is a reproduction of mo-gym's `LinearReward ` wrapper.

## Motivation and Context
This feature provides a naive way to address multi-objective environments.

## Types of changes

## Checklist

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -12630,9 +12630,7 @@ class TestLineariseRewards(TransformBase):
 
         expected = sum(
             w * r
-            for w, r in zip(
-                weights, rollout.get(("next", "reward")).split(1, dim=-1), strict=True
-            )
+            for w, r in zip(weights, rollout.get(("next", "reward")).split(1, dim=-1))
         )
         torch.testing.assert_close(scalar_reward, expected)
 
@@ -12654,7 +12652,7 @@ class TestLineariseRewards(TransformBase):
         td = TensorDict({"reward": torch.randn(num_rewards)}, [])
         model(td)
 
-        expected = sum(w * r for w, r in zip(weights, td["reward"], strict=True))
+        expected = sum(w * r for w, r in zip(weights, td["reward"]))
         torch.testing.assert_close(td["scalar_reward"], expected)
 
     @pytest.mark.parametrize("rbclass", [ReplayBuffer, TensorDictReplayBuffer])
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -12703,7 +12703,7 @@ class TestLineariseRewards(TransformBase):
     )
     def test_reward_spec(
         self,
-        weights: list[float] | None,
+        weights,
         reward_spec: TensorSpec,
         expected_spec: TensorSpec,
     ) -> None:
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -94,6 +94,7 @@ from torchrl.data import (
     Unbounded,
     UnboundedContinuous,
 )
+from torchrl.data.tensor_specs import BoundedContinuous
 from torchrl.envs import (
     ActionMask,
     BinarizeReward,
@@ -414,7 +415,7 @@ class TestBinarizeReward(TransformBase):
         assert ((sample["reward"] == 0) | (sample["reward"] == 1)).all()
 
     def test_transform_inverse(self):
-        raise pytest.skip("No inverse for BinerizedReward")
+        raise pytest.skip("No inverse for BinarizedReward")
 
 
 class TestClipTransform(TransformBase):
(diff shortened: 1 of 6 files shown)


test/test_transforms.py
```diff
@@ -92,7 +92,9 @@
     TensorSpec,
     TensorStorage,
     Unbounded,
+    UnboundedContinuous,
 )
+from torchrl.data.tensor_specs import BoundedContinuous
```
louisfaury, line 97: `BoundedContinuous` is not imported at the root of the `data` package -- happy to add to this PR if that's ok.

vmoens: Yes that'd be cool!

torchrl/envs/transforms/transforms.py
```diff
@@ -9287,3 +9289,118 @@ def transform_observation_spec(self, observation_spec: Composite) -> Composite:
...
         )
         return super().transform_observation_spec(observation_spec)
+
+
+class LineariseRewards(Transform):
+    """Transforms a multi-objective reward signal to a single-objective one via a weighted sum.
+
+    Args:
+        in_keys: The keys under which the multi-objective rewards are found
+        out_keys: The keys under which single-objective rewards should be written. Defaults to `in_keys`.
```
vmoens, line 9299:
```suggestion
        out_keys (List[NestedKey], optional): The keys under which single-objective rewards should be written. Defaults to :attr:`in_keys`.
```

torchrl/envs/transforms/transforms.py
```diff
@@ -9287,3 +9289,118 @@ def transform_observation_spec(self, observation_spec: Composite) -> Composite:
             high=torch.iinfo(torch.int64).max,
         )
         return super().transform_observation_spec(observation_spec)
+
+
+class LineariseRewards(Transform):
+    """Transforms a multi-objective reward signal to a single-objective one via a weighted sum.
+
+    Args:
+        in_keys: The keys under which the multi-objective rewards are found
```
vmoens, line 9298:
```suggestion
        in_keys (List[NestedKey]): The keys under which the multi-objective rewards are found.
```

torchrl/envs/transforms/transforms.py
```diff
@@ -9287,3 +9289,118 @@ def transform_observation_spec(self, observation_spec: Composite) -> Composite:
...
+        out_keys: The keys under which single-objective rewards should be written. Defaults to `in_keys`.
+        weights: Dictates how to weight each reward when summing them. Defaults to `[1.0, 1.0, ...]`.
+
+    Warning:
```
vmoens, line 9302:
```suggestion
    .. warning::
```

torchrl/envs/transforms/transforms.py
```diff
@@ -9287,3 +9289,118 @@ def transform_observation_spec(self, observation_spec: Composite) -> Composite:
...
         return super().transform_observation_spec(observation_spec)
+
+
+class LineariseRewards(Transform):
+    """Transforms a multi-objective reward signal to a single-objective one via a weighted sum.
+
+    Args:
+        in_keys: The keys under which the multi-objective rewards are found
+        out_keys: The keys under which single-objective rewards should be written. Defaults to `in_keys`.
+        weights: Dictates how to weight each reward when summing them. Defaults to `[1.0, 1.0, ...]`.
```
vmoens, line 9300: What's the typing? Is a tensor appropriate? What about a differentiable tensor?

torchrl/envs/transforms/transforms.py
```diff
@@ -9287,3 +9289,118 @@ def transform_observation_spec(self, observation_spec: Composite) -> Composite:
...
+
+    Warning:
+        If a sequence of `in_keys` of length strictly greater than one is passed (e.g. one group for each agent in a
+        multi-agent set-up), the same weights will be applied for each entry. If you need to aggregate rewards
+        differently for each group, use several `AggregateRewardsTransform` in a row.
```
vmoens, line 9305:
```suggestion
        differently for each group, use several :class:`~torchrl.envs.AggregateRewardsTransform` in a row.
```

torchrl/envs/transforms/transforms.py
```diff
@@ -9287,3 +9289,118 @@ def transform_observation_spec(self, observation_spec: Composite) -> Composite:
...
+    @_apply_to_composite
+    def transform_reward_spec(self, reward_spec: TensorSpec) -> TensorSpec:
+        if not reward_spec.domain == "continuous":
+            raise NotImplementedError(
+                "Aggregation of rewards that take discrete values is not supported."
+            )
+
+        *batch_size, num_rewards = reward_spec.shape
+        weights = (
+            torch.ones(num_rewards, device=reward_spec.device)
```
vmoens, line 9367: maybe get the dtype of the reward spec?

torchrl/envs/transforms/transforms.py
```diff
@@ -9287,3 +9289,118 @@ def transform_observation_spec(self, observation_spec: Composite) -> Composite:
...
+
+    def _apply_transform(self, reward: Tensor) -> TensorDictBase:
+        if self.weights is None:
+            return reward.sum(dim=-1)
+
+        *batch_size, num_rewards = reward.shape
+        num_weights = torch.numel(self.weights)
+        if num_weights != num_rewards:
+            raise ValueError(
+                "The number of rewards and weights should match."
```
vmoens, line 9402:
```suggestion
                "The number of rewards and weights should match. "
```

torchrl/envs/transforms/transforms.py
```diff
@@ -9287,3 +9289,118 @@ def transform_observation_spec(self, observation_spec: Composite) -> Composite:
...
+        weights = (
+            torch.ones(num_rewards, device=reward_spec.device)
+            if self.weights is None
+            else self.weights
+        )
+
+        num_weights = torch.numel(weights)
+        if num_weights != num_rewards:
+            raise ValueError(
+                "The number of rewards and weights should match."
```
vmoens, line 9375:
```suggestion
                "The number of rewards and weights should match. "
```

vmoens (commented):
LGTM thanks for this!
Just a few suggestions

torchrl/envs/transforms/transforms.py
louisfaury: I added the typing, `List[float] | Tensor | None`. I didn't give any thoughts for the tensor to be differentiable -- I believe here it would give unexpected results as `weights` are registered as a non-parameter buffer.

torchrl/envs/transforms/transforms.py
```diff
@@ -9287,3 +9289,118 @@ def transform_observation_spec(self, observation_spec: Composite) -> Composite:
...
+        >>> so_env.reward_spec
+        BoundedContinuous(
+            shape=torch.Size([1]),
+            space=ContinuousBox(
+                low=Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.float32, contiguous=True),
+                high=Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.float32, contiguous=True)),
+            ...)
+        >>> td = so_env.rollout(5)
+        >>> td["next", "reward"].shape
+        [5, 1]
```
vmoens, line 9328:
```suggestion
        torch.Size([5, 1])
```

vmoens (approved):
LGTM thanks!

torchrl/envs/transforms/transforms.py
louisfaury: Yes -- my understanding about this is that it would therefore be ignored by any optimizer. 
What I mean by 
> it would give unexpected results

is that if you pass a differentiable tensor `weights` to this module in the hope to optimise it somehow, you won't ever get any gradients for it. 

Maybe I'm misunderstanding your comment?

vmoens: No no, my comment was just this:
If your weight is differentiable, is the reward going to be differentiable too?
I think it will so that's great (I can imagine some researcher, somewhere, trying to hack something like that)

vmoens:
From the failing tests it seems a `from __future__ import annotations` is missing - can you check @louisfaury?

torchrl/envs/transforms/transforms.py
louisfaury: Got it -- the answer is that as it, no. 

I also see the value, but maybe we can keep this to anther PR ?

PR #2698, opened 2025-01-17
[Feature] `example_data` for NonTensor spec

vmoens:
* (to be filled)

### DIFF
--- a/torchrl/data/tensor_specs.py
+++ b/torchrl/data/tensor_specs.py
@@ -2457,6 +2457,7 @@ class NonTensor(TensorSpec):
         shape: Union[torch.Size, int] = _DEFAULT_SHAPE,
         device: Optional[DEVICE_TYPING] = None,
         dtype: torch.dtype | None = None,
+        example_data: Any = None,
         **kwargs,
     ):
         if isinstance(shape, int):
@@ -2467,6 +2468,7 @@ class NonTensor(TensorSpec):
         super().__init__(
             shape=shape, space=None, device=device, dtype=dtype, domain=domain, **kwargs
         )
+        self.example_data = example_data
 
     def cardinality(self) -> Any:
         raise RuntimeError("Cannot enumerate a NonTensorSpec.")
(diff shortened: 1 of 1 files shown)


vmoens:
@kurtamohler I think it would also be useful to have a spec that specified a set of available choices.
I was thinking to do that in `NonTensor` first (like `NonTensor(choices["a", "b", "c"])`, where `rand()` would pick one of the three values) but that seem restrictive (we'd need another one if you can select some tensor values).

One solution I'm contemplating is to have a `Choice` spec that works like this:
```python
choice = Choice(
    torch.stack([
        NonTensor("a"),
        NonTensor("b"),
        NonTensor("c"),
    ])
)
choice.rand() # gives NonTensorData("a"), for instance
```
Using this you could do
```python
choice = Choice(
    torch.stack([
        Bounded(0, 1),
        Bounded(5, 10),
    ])
)
choice.rand() # gives a tensor with value in [0, 5, 6, 7, 8, 9]
```

kurtamohler:
`Choice` sounds like a useful spec.

I like this `example_data` feature--it partially addresses something I saw when writing Hash/UnaryTransform, which caused me to make this stipulation about `None` inputs: https://github.com/pytorch/rl/blob/c5f1565de5b066c688fe1291a9eda69ffe6b7477/torchrl/envs/transforms/transforms.py#L4418-L4419

The only reason that stipulation exists is that when UnaryTransform calls `spec.zero()` to make a fake input to determine the output spec, if the spec has a NonTensor, the result for it is `None`. But after this PR, the `example_data` gets chosen instead of `None`.

Just a thought--maybe `example_data` could have a sensible default other than `None` for some dtypes? Like for `str`, it could be `example_data=""`

kurtamohler (approved):

vmoens:
> Just a thought--maybe `example_data` could have a sensible default other than `None` for some dtypes? Like for `str`, it could be `""` or even `"example_data"`

Oh you mean like `NonTensor(dtype=str).example_data == ""`
That sounds ok, though `str` isn't a dtype so I'm wondering if `NonTensor(dtype=str).dtype == str` won't end up being a footgun...

PR #2699, opened 2025-01-17
[BugFix] patch rand_action in TransformedEnv to read the base_env method

vmoens:
* #2711
* #2709
* #2708
* #2702
* #2700
* __->__ #2699

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -795,6 +795,17 @@ but got an object of type {type(transform)}."""
             input_spec = self.__dict__.get("_input_spec", None)
         return input_spec
 
+    def rand_action(self, tensordict: Optional[TensorDictBase] = None) -> TensorDict:
+        if self.base_env.rand_action is not EnvBase.rand_action:
+            # TODO: this will fail if the transform modifies the input.
+            #  For instance, if PendulumEnv overrides rand_action and we build a
+            #  env = PendulumEnv().append_transform(ActionDiscretizer(num_intervals=4))
+            #  env.rand_action will NOT have a discrete action!
+            #  Getting a discrete action would require coding the inverse transform of an action within
+            #  ActionDiscretizer (ie, float->int, not int->float).
+            return self.base_env.rand_action(tensordict)
+        return super().rand_action(tensordict)
+
     def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
         # No need to clone here because inv does it already
         # tensordict = tensordict.clone(False)


vmoens:
@kurtamohler I'm not super happy with this, as the comment says it's far from accounting for transforms that do some kind of inverse mapping but the point is that without that, a transformed chess env cannot generate random actions (because the `rand_action` is not the same as the one in `EnvBase` that would otherwise be used by `TransformedEnv`).

torchrl/envs/transforms/transforms.py
```diff
@@ -795,6 +795,17 @@ def input_spec(self) -> TensorSpec:
...
         return input_spec
 
+    def rand_action(self, tensordict: Optional[TensorDictBase] = None) -> TensorDict:
+        if self.base_env.rand_action is not EnvBase.rand_action:
+            # TODO: this will fail if the transform modifies the input.
+            #  For instance, if PendulumEnv overrides rand_action and we build a
+            #  env = PendulumEnv().append_transform(ActionDiscretizer(num_intervals=4))
+            #  env.rand_action will NOT have a discrete action!
+            #  Getting a discrete action would require coding the inverse transform of an action within
+            #  ActionDiscretizer (ie, float->int, not int->float).
```
kurtamohler, line 805: Is there a reason we couldn't do `self.action_spec.rand()`?

```python
>>> import torchrl
>>> env = torchrl.envs.PendulumEnv().append_transform(torchrl.envs.ActionDiscretizer(num_intervals=4))
>>> env.action_spec
MultiCategorical(
    shape=torch.Size([1]),
    space=BoxList(boxes=[CategoricalBox(n=4)]),
    device=cpu,
    dtype=torch.int64,
    domain=discrete)
>>> env.action_spec.rand()
tensor([3])
>>> env.action_spec.rand().dtype
torch.int64
```

vmoens: yes, what I meant with that comment is that *if* your base env redefines the `rand_action` then the action you'll get won't be transformed
```python

import torchrl
Pendulum = torchrl.envs.PendulumEnv
rand_action = Pendulum.rand_action
Pendulum.rand_action = lambda *args, **kwargs: rand_action(*args, **kwargs)

env = Pendulum().append_transform(torchrl.envs.ActionDiscretizer(num_intervals=4))

print(env.action_spec.rand())
print(env.action_spec.rand().dtype)
print(env.rand_action())

```

which prints
```
tensor([3])
torch.int64
TensorDict(
    fields={
        action: Tensor(shape=torch.Size([1]), device=cpu, dtype=torch.float32, is_shared=False)},
    batch_size=torch.Size([]),
    device=None,
    is_shared=False)
```

PR #2702, opened 2025-01-17
[Feature,Refactor] Chess improvements: fen, pgn, pixels, san

vmoens:
* #2711
* #2709
* #2708
* __->__ #2702


torchrl/envs/custom/chess.py
```diff
@@ -90,28 +115,76 @@ class ChessEnv(EnvBase):
     """
 
     _hash_table: Dict[int, str] = {}
+    _PNG_RESTART = """[Event "?"]
```
kurtamohler, line 118: nit: 
```suggestion
    _PGN_RESTART = """[Event "?"]
```

torchrl/envs/custom/chess.py
```diff
@@ -132,41 +205,126 @@ def _is_done(self, board):
...
+        node = game
+        for move in board.move_stack:
+            node = node.add_variation(move)
+
+        # Generate the PGN string
+        pgn_string = str(game)
+        return pgn_string
+
     @classmethod
     def _get_fen(cls, tensordict):
```
kurtamohler, line 326: I guess this function can probably be removed, in favor of a direct `TensorDict.get` call

torchrl/envs/custom/chess.py
```diff
@@ -132,41 +205,126 @@ def _is_done(self, board):
...
         fen = None
+        pgn = None
         if tensordict is not None:
-            fen = self._get_fen(tensordict).data
-            dest = tensordict.empty()
+            if self.include_fen:
+                fen = self._get_fen(tensordict).data
+                dest = tensordict.empty()
+            if self.include_pgn:
+                fen = self._get_pgn(tensordict).data
```
kurtamohler, line 214:
```suggestion
                pgn = self._get_pgn(tensordict).data
```

Also `_get_pgn` doesn't exist, but I figure you realize that

torchrl/envs/custom/chess.py
```diff
@@ -132,41 +205,126 @@ def _is_done(self, board):
...
+        if tv is None:
+            import torchvision
+
+            tv = cls._torchvision_lib = torchvision
+        return tv
+
+    @classmethod
+    def _get_tensor_image(cls, board):
+        try:
+            svg = board._repr_svg_()
```
kurtamohler, line 277: Maybe we could have another option that has a more minimal representation as well, if we want to train without vision. For instance, OpenSpiel's chess env has a representation where each square of the board is given a size 20 array of ones and zeros, which is somehow used to represent which piece is in that square and which color it is

vmoens: Sure, I was using this because it's builtin - but a more minimal representation could also be cool!

torchrl/envs/custom/chess.py
vmoens: I was thinking that one day we might let people choose the fen / pgn key for themselves but it's true that as of now it's not incredibly useful

PR #2707, opened 2025-01-20
[Feature] Log each entropy for composite distributions in PPO

louisfaury:
## Description

This PR enables PPO to log the entropy of each individual head of a composite policy separately.  

Concretely, for a composite distribution with, say, a nested discrete and continuous head, the `td_out` is augmented with some detached values.
```python
>>> loss = PPO(td)
>>> loss.keys()
[..., entropy, discrete-head-entropy, continuous-head-entropy, ...]
```
## Motivation and Context

This is an extremely useful debugging tool when training composite policies.

## Types of changes

## Checklist

### DIFF
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -684,7 +684,7 @@ class PPOLoss(LossModule):
             entropy = self._get_entropy(dist)
             if is_tensor_collection(entropy):
                 # Reports the entropy of each action head.
-                td_out.set("composite_entropy", entropy)
+                td_out.set("composite_entropy", entropy.detach())
                 entropy = _sum_td_features(entropy)
             td_out.set("entropy", entropy.detach().mean())  # for logging
             td_out.set("loss_entropy", -self.entropy_coef * entropy)
@@ -984,7 +984,7 @@ class ClipPPOLoss(PPOLoss):
             entropy = self._get_entropy(dist)
             if is_tensor_collection(entropy):
                 # Reports the entropy of each action head.
-                td_out.set("composite_entropy", entropy)
+                td_out.set("composite_entropy", entropy.detach())
                 entropy = _sum_td_features(entropy)
             td_out.set("entropy", entropy.detach().mean())  # for logging
             td_out.set("loss_entropy", -self.entropy_coef * entropy)
@@ -1297,7 +1297,7 @@ class KLPENPPOLoss(PPOLoss):
             entropy = self._get_entropy(dist)
             if is_tensor_collection(entropy):
                 # Reports the entropy of each action head.
-                td_out.set("composite_entropy", entropy)
+                td_out.set("composite_entropy", entropy.detach())
                 entropy = _sum_td_features(entropy)
             td_out.set("entropy", entropy.detach().mean())  # for logging
             td_out.set("loss_entropy", -self.entropy_coef * entropy)
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -687,8 +687,7 @@ class PPOLoss(LossModule):
                 for head_key, head_entropy in entropy.items(
                     include_nested=True, leaves_only=True
                 ):
-                    head_prefix = head_key[0] if len(head_key) == 1 else head_key[-2]
-                    td_out.set(f"{head_prefix}_entropy", head_entropy.detach().mean())
+                    td_out.set("-".join(head_key), head_entropy.detach().mean())
                 entropy = _sum_td_features(entropy)
             td_out.set("entropy", entropy.detach().mean())  # for logging
             td_out.set("loss_entropy", -self.entropy_coef * entropy)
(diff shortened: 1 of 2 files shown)


torchrl/objectives/ppo.py
```diff
@@ -681,10 +679,17 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
             log_weight = log_weight.view(advantage.shape)
         neg_loss = log_weight.exp() * advantage
         td_out = TensorDict({"loss_objective": -neg_loss}, batch_size=[])
+        td_out.set("kl_approx", kl_approx.detach().mean())  # for logging
```
louisfaury, line 682: Unrelated to the main topic of this MR, but afaik the existence and relevance of `kl_approx` is independent of whether we are training with an entropy bonus.

torchrl/objectives/ppo.py
```diff
@@ -681,10 +679,17 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
         if self.entropy_bonus:
-            entropy = self.get_entropy_bonus(dist)
+            entropy = self._get_entropy(dist)
+            if is_tensor_collection(entropy):
+                # Logs the entropy of each action head.
+                for head_key, head_entropy in entropy.items(
+                    include_nested=True, leaves_only=True
+                ):
+                    td_out.set("-".join(head_key), head_entropy.detach().mean())
```
louisfaury, line 690: Choosing under which key to log the individual factor's entropy was a bit of a headache. The way I personally use `CompositeDistribution` yields tensordict that look like:

```
action: {
    head_1: {
        action: ...
        entropy: ...
   }
   head_2 {
        action: ...
        entropy: ...
   }
}
```
which means that using `head_key[-1]` to log each entropy is not really a viable solution (all the factor entropies will be logged under the same name, `entropy`). I'm not sure how to get a one-size-fits-all here, and happy for suggestions. The current solution ensures that there is no collision, at the price of having very verbose keys.

vmoens: Wouldn't the most generic solution just be to log the entropy TD as it comes?
Why do we need to rename it?
BTW it seems to me that what you're doing here amends to
```python
tensordict.flatten_keys("-").detact().mean()
```
Nit: this isn't collision-safe I think (but flatten_keys will tell you if there are any collision):
eg `("key-one", "entropy")` and `("key", "one", "entropy")` will collide

torchrl/objectives/ppo.py
```diff
@@ -976,11 +981,18 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
 
         if self.entropy_bonus:
-            entropy = self.get_entropy_bonus(dist)
+            entropy = self._get_entropy(dist)
+            if is_tensor_collection(entropy):
+                # Logs the entropy of each action head.
+                for head_key, head_entropy in entropy.items(
+                    include_nested=True, leaves_only=True
+                ):
+                    td_out.set("-".join(head_key), head_entropy.detach().mean())
```
vmoens, line 993: ditto, we could not flatten and let users do that, no?

torchrl/objectives/ppo.py
vmoens: also - are we 100% sure all keys are nested? I think so (that's how `CompositeDist` works) but maybe we could just put a safeguard check here to make sure an error is raised if that assumption is violated (eg, users have their own dist class that returns `{"entropy", ("nested", "entropy")}` keys).

vmoens (commented):
LGTM, just a couple of comments to address before we merge it

LMK what you think would be the best way to log the entropies in the output data structure, I think flattening may be a bit surprising

torchrl/objectives/ppo.py
louisfaury: I followed your recommendation: I added a `composite_entropy` key to the loss td. Two remarks:

1.  The composite entropy is not logged under `entropy` to avoid BC (users currently expect a Tensor),
2. I did not `detach()` the composite entropy; this would allow the user to compute a custom entropy bonus when using a composite entropy (e.g. not the same penalty per head).

Wdyt ?

vmoens: Not detaching could be slightly bc-breaking: what if I do `loss_tensordict.sum(reduce=True).backward()`?
Previously, this was giving the right result, now it would also backprop through the entropy. Usually metadata in loss outputs is guaranteed to be non-differentiable, that would be a one off.
But I understand that it could be useful... 
We could add a kwarg in the constructor (which would become a bit overloaded!)

louisfaury: Right. I'll detach for now, let's revisit when there is a need/ask for it?

vmoens: yes, thanks for raising it!
Thinking about it, a way of doing this could be to pass the entropy coefficient as a tensordict and do 
`(td_coef * td_entropy).sum(reduce=True)`
Idea for a follow-up PR ;)

vmoens (approved):
LGTM thanks!

PR #2713, opened 2025-01-22
[Feature] Add `Choice` spec

### LINKED ISSUE #2712: [Feature Request] `Choice` spec
## Motivation

Add a `Choice` spec that chooses between a stack of specs or non-tensor data.

Initially proposed here: https://github.com/pytorch/rl/pull/2698#issuecomment-2598616398

## Solution

When generating a sample, `Choice` spec should choose between the specs in the stack and then generate a sample from that spec.

## Checklist

kurtamohler:
* __->__ #2713

Close #2712


torchrl/data/tensor_specs.py
```diff
@@ -3678,6 +3679,99 @@ def clone(self) -> Categorical:
...
+        domain = None
+        super().__init__(
+            shape=shape, space=None, device=device, dtype=dtype, domain=domain
+        )
+
+    def _rand_idx(self):
+        return torch.randint(0, len(self._stack), ()).item()
+
+    def _sample(self, idx, spec_sample_fn) -> TensorDictBase:
+        res = self._stack[idx]
```
kurtamohler, line 3731: At the moment, we're always just choosing from the first dimension of the stack. But of course, stacks can have more than one dimension. So I'm wondering if `Choice` should have a `dims` argument so we can sample from multiple dims.

That would also make it possible to support `expand` and `unsqueeze`, which I'm not sure would really make sense otherwise. Furthermore, it would allow us to create nested `Choice` specs (implemented as a single `Choice` with multiple sampling dims), which is not currently supported

test/test_specs.py
```diff
@@ -1409,6 +1433,21 @@ def test_non_tensor(self):
...
+            stack = torch.stack([Bounded(0, 2.5, ()), Bounded(10, 12, ())])
+        elif input_type == "nontensor":
+            stack = torch.stack([NonTensorData("a"), NonTensorData("b")])
+        elif input_type == "nontensorstack":
+            stack = torch.stack(
+                [NonTensorStack("a", "b", "c"), NonTensorStack("d", "e", "f")]
+            )
+
+        spec = Choice(stack)
+        with pytest.raises(NotImplementedError):
```
vmoens, line 1448: is this intentional? Why can't we just expand the sub-specs?
I would have thought that this would hold
```python
choice = Choice(torch.stack([Bounded(0, 2.5, (3,)), Bounded(10, 12, (3,))]))
assert choice.shape == (3,)
choice_expand = choice.expand(10, 3)
assert choice_expand.shape == (10, 3,)
assert choice_expand.rand().shape == (10, 3)
```

test/test_specs.py
```diff
@@ -678,6 +685,23 @@ def test_change_batch_size(self, shape, is_complete, device, dtype):
         assert ts["nested"].shape == (3,)
 
 
+class TestChoiceSpec:
+    @pytest.mark.parametrize("input_type", ["spec", "nontensor", "nontensorstack"])
+    def test_choice(self, input_type):
+        if input_type == "spec":
+            stack = torch.stack([Bounded(0, 2.5, ()), Bounded(10, 12, ())])
```
vmoens, line 692: do we need to stack or can't we just use a plain list, like
`Choice(spec1, spec2)`?
`stack` implies that there's a dim along which we stack, but the values generated / tested against don't have that concept of stack dim so perhaps stack is superfluous?

test/test_specs.py
```diff
@@ -2270,6 +2374,29 @@ def test_stack_non_tensor(self, shape, stack_dim):
...
+            )
+        elif input_type == "nontensorstack":
+            stack = torch.stack(
+                [NonTensorStack("a", "b", "c"), NonTensorStack("d", "e", "f")]
+            )
+
+        spec0 = Choice(stack)
+        spec1 = Choice(stack)
+        with pytest.raises(NotImplementedError):
+            torch.stack([spec0, spec1], 0)
```
vmoens, line 2398: We should be able to stack (if we can expand - see above)

torchrl/data/tensor_specs.py
```diff
@@ -3678,6 +3679,99 @@ def clone(self) -> Categorical:
         )
 
 
+class Choice(TensorSpec):
+    """A discrete choice spec for either tensor or non-tensor data.
+
+    Args:
+        stack (:class:`~Stacked`, :class:`~StackedComposite`, or :class:`~tensordict.NonTensorStack`):
+            Stack of specs or non-tensor data from which to choose during
+            sampling.
```
vmoens, line 3688: See test comment: I think we should be able to get away with just a list of specs

torchrl/data/tensor_specs.py
vmoens: My take is this (but maybe I don't have the full picture in mind):
- Let's not use stack, let's just pass a list of specs
- maybe one restriction of Choice would be that all the items have the same type, same shape, same dtype, same device. Like this, `choice.device` and such are defined. The loose version of this is that if dtype or device don't match they default to `None`.
- rand takes one spec at random and return `that_spec.rand()`
- `zero` and `one` are a bit more complex - if device / type / dtype / shape match we're good by doing `specs[0].zero()` but if not, `zero()` is undefined (not good!)

So my take for a first implementation is: let's make sure that device / type / dtype / shape match - like that the behavior is predictable. In the future we can assess if we relax that assumption.

vmoens (commented):
Thanks I left a few comments, feel free to disagree if you think it won't work!

test/test_specs.py
kurtamohler: Fixed

kurtamohler:
I've switched over to using a list instead of a stack, and I enabled expanding and stacking

vmoens (approved):

PR #2714, opened 2025-01-23
[Test] Add tests for CatFrames with PermuteTransform

kurtamohler:
* __->__ #2714

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -693,6 +693,7 @@ class TestCatFrames(TransformBase):
     @pytest.mark.parametrize("cat_N", [3, 10])
     @pytest.mark.parametrize("device", get_default_devices())
     def test_with_permute_no_env(self, cat_dim, cat_N, device):
+        torch.manual_seed(cat_dim * cat_N)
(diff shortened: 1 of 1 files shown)


kurtamohler:
@vmoens, I tried to find a case where CatFrames fails when it's surrounded by two permutes, but the ones I tried all worked. This PR contains what I tried. Let me know how to reproduce the issue you saw, and I can add that case and fix it.

vmoens:
> @vmoens, I tried to find a case where CatFrames fails when it's surrounded by two permutes, but the ones I tried all worked. I figured I'd make this PR to share what I tried. Let me know how to reproduce the issue you saw, and I can add that case and fix it.

Ah that's very cool! I'll sleep better tonight :)
Can you add a seed at the beginning of the test to avoid it being flaky?

kurtamohler:
> Can you add a seed at the beginning of the test to avoid it being flaky?

Done

vmoens (approved):
LGTM thanks so much!

kurtamohler:
Accidentally merged this from github instead of running `ghstack land`, so resubmitting with #2715

PR #2732, opened 2025-01-30
[Feature] TensorDictPrimer with single default_value callable

vmoens:
* #2823
* #2822
* #2821
* #2819
* #2818
* #2817
* #2816
* #2815
* __->__ #2732


vmoens:
Example script:
```python

from torchrl.envs import TensorDictPrimer, PendulumEnv, StepCounter, TrajCounter
import torch
from tensordict import assert_close
# let's build a primer that outputs the same Pendulum initial state 4 times in a row
env = PendulumEnv()

def val_iterator(N = 4, env=env):
    i = 0
    while True:
        i += 1
        r = env.reset()
        for _ in range(N):
            torch.manual_seed(i)
            yield r

iterator = val_iterator()

def val_generator(iterator=iterator):
    return next(iterator)

print('observation_spec', env.observation_spec)
primer = TensorDictPrimer(primers=env.observation_spec, default_value=val_generator, single_default_value=True)
env = env.append_transform(primer)
env = env.append_transform(StepCounter(max_steps=50))
env = env.append_transform(TrajCounter())
r = env.rollout(1000, break_when_any_done=False)
print(r[0]["th"], r[0]["thdot"])
print(r[50]["th"], r[50]["thdot"])

# To compute the GAE with empirical adv, we fist need to collapse the trajs
r["traj_count"] = r["traj_count"]//4
r["next", "traj_count"] = r["next", "traj_count"]//4

assert_close(r[:50], r[50:100])
assert_close(r[:50], r[100:150])

# Reshape r
print(r.shape)
r = r.reshape(-1, 4, 50)
assert r[0]["traj_count"].unique().numel() == 1
```

PR #2738, opened 2025-01-31
[Test] Add tests for Tree

kurtamohler:
* __->__ #2738


kurtamohler:
I mostly wrote this to become more familiar with Tree. Maybe we don't strictly need these tests since a lot of it is also tested in TestMCTSForest, but I figured it might be good to have some tests that are independent of the MCTSForest API. If you think we don't need this, I'm fine with just closing it

vmoens (approved):
Definitely worth having! Thanks for that

PR #2757, opened 2025-02-04
[Feature] Enable `Hash.inv`

kurtamohler:
* #2766
* __->__ #2757


torchrl/envs/transforms/transforms.py
```diff
@@ -4826,25 +4826,23 @@ class Hash(UnaryTransform):
...
-                mapped to ``None``.
-
         out_keys_inv (sequence of NestedKey, optional): the keys of the resulting hashes during inv call.
 
     Keyword Args:
         hash_fn (Callable, optional): the hash function to use. If ``seed`` is given,
-            the hash function must accept it as its second argument. Default is
-            ``Hash.reproducible_hash``.
+            the hash function must accept it as its second argument. The output of
+            the hash function must be a tensor. Default is ``Hash.reproducible_hash``.
```
kurtamohler, line 4834: I thought it would be difficult to support all output types for `hash_fn`, since it's not obvious how to convert any arbitrary type into a dictionary key. But we could try to remove this restriction if needed

vmoens (approved):
LGTM thanks

PR #2763, opened 2025-02-05
ppo chess with llm draft

mikaylagawarecki:
* (to be filled)

### DIFF
--- a/examples/agents/ppo-chess-llm.py
+++ b/examples/agents/ppo-chess-llm.py
@@ -1,0 +1,483 @@
+import queue
+import random
+import re
+import subprocess
+import threading
+import time
+
+import chess
+
+import torch
+from tensordict.nn import (
+    ProbabilisticTensorDictModule as Prob,
+    ProbabilisticTensorDictSequential as ProbSeq,
+    TensorDictModule as Mod,
+    TensorDictSequential as Seq,
+)
+
+from torch.distributions import Categorical
+
+from torchrl.collectors import SyncDataCollector
+from torchrl.data import NonTensor
+from torchrl.data.replay_buffers.samplers import SliceSamplerWithoutReplacement
+from torchrl.data.tensor_specs import Composite
+from torchrl.envs import ChessEnv
+from torchrl.envs.transforms import (
+    ConditionalPolicySwitch,
+    StepCounter,
+    Tokenizer,
+    Transform,
+)
+from torchrl.objectives import ClipPPOLoss
+from torchrl.objectives.value import GAE
+
+from tqdm import tqdm
+
+from transformers import pipeline
+
+
+def random_policy(td):
+    legal_moves = env.get_legal_moves()
+    move = random.choice(legal_moves)
+    san_idx = env.san_moves.index(move)
+    td["action"] = torch.tensor(san_idx)
+    return td
+
+
+class SanHistory(Transform):
+    def __init__(self):
+        super().__init__()
+
+    def _step(self, tensordict, next_tensordict):
+        import numpy as np
+
+        try:
+            history = np.append(tensordict["san"], next_tensordict["san"])
+        except:
+            history = tensordict["san"] + [next_tensordict["san"]]
+        next_tensordict["san"] = history
+        return next_tensordict
+
+    def _reset(self, tensordict, tensordict_reset):
+        return tensordict_reset
+
+
+class LLMInputTransform(Transform):
+    def __init__(self, san_moves):
+        super().__init__()
+        self.san_moves = san_moves
+
+    def _step(self, tensordict, next_tensordict):
+        legal_move_indices = next_tensordict["legal_moves"]
+        legal_moves = [self.san_moves[i] for i in legal_move_indices if i != 29275]
+        existing_moves = (
+            ", ".join(next_tensordict["san"])
+            if not isinstance(next_tensordict["san"], str)
+            else next_tensordict["san"]
+        )
+        if len(legal_moves) > 0:
+            llm_input = (
+                f"You are playing a game of chess. The list of moves so far are [{existing_moves}] and the legal moves are "
+                f"[{', '.join(legal_moves)}]. Please choose one of the legal moves. "
+                f"Respond only with the following sentence, with no additional explanatory text. Example Answer: I choose {legal_moves[0]}!"
+            )
+        else:
+            llm_input = "The game has ended as there are no legal moves left. "
+        next_tensordict["observation"] = llm_input
+        print("step", llm_input)
+        return next_tensordict
+
+    def _reset(self, tensordict, tensordict_reset):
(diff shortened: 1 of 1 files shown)


examples/agents/ppo-chess-llm.py
```diff
@@ -0,0 +1,478 @@
...
+        Mod(
+            LLMWrapper(llm, tokenizer, mode="policy"),
+            in_keys=["obs_tokens"],
+            out_keys=["logits", "hidden"],
+        ),
+        prob_module,
+        # if use lambda: 'function' object has no attribute 'in_keys'
+        Mod(AggregateProb(), in_keys=["sample_log_prob"], out_keys=["sample_log_prob"]),
+        return_composite=True,
+    )
```
mikaylagawarecki, line 350: Does how `data_llm_policy` and `actor_llm_policy` were defined make sense here?

examples/agents/ppo-chess-llm.py
```diff
@@ -0,0 +1,478 @@
...
+            tensor2_padded = F.pad(tensor2, (0, pad2))
+            tensor1_padded = tensor1
+        else:
+            tensor1_padded = tensor1
+            tensor2_padded = tensor2
+
+        return tensor1_padded, tensor2_padded
+
+    for data in tqdm(collector):
+        # FIXME: reward seems to be getting wrongly propagated (e.g. sunfish's win gets reflected as llm's win)
```
mikaylagawarecki, line 432: debugging this one, not sure whether it's an issue with how I applied the transforms

examples/agents/ppo-chess-llm.py
```diff
@@ -0,0 +1,478 @@
...
+        storage=LazyStackStorage(100),
+        batch_size=32,
+        sampler=SliceSamplerWithoutReplacement(slice_len=8, end_key=("next", "done")),
+    )
+    # layout=torch.jagged errors with Qwen
+    # File "/home/mg1998/.conda/envs/rl/lib/python3.10/site-packages/transformers/models/qwen2/modeling_qwen2.py", line 859, in forward
+    # cache_position = torch.arange(
+    #             past_seen_tokens, past_seen_tokens + inputs_embeds.shape[1], device=inputs_embeds.device
+    #         )
+    # AttributeError: 'ConstantIntNode' object has no attribute 'add'
```
mikaylagawarecki, line 380: The takeaway here for me is that we can't generically expect to use NJT for input_tokens to the LLM unless the LLM forward is written in an "NJT friendly" manner without any queries to sizes/implicit assumptions about dense input, in this case Qwen is making some such assumptions..

examples/agents/ppo-chess-llm.py
```diff
@@ -0,0 +1,572 @@
...
+        except:
+            history = tensordict["san"] + [next_tensordict["san"]]
+        next_tensordict["san"] = history
+        return next_tensordict
+
+    def _reset(self, tensordict, tensordict_reset):
+        return tensordict_reset
+
+
+class Score(Transform):
```
mikaylagawarecki, line 78: trying this instead of a randomly initialized critic head

examples/agents/ppo-chess-llm.py
```diff
@@ -0,0 +1,572 @@
...
+    def transform_observation_spec(self, observation_spec: Composite):
+        if not isinstance(observation_spec, Composite):
+            raise ValueError(
+                f"observation_spec was expected to be of type Composite. Got {type(observation_spec)} instead."
+            )
+        observation_spec["observation"] = NonTensor(shape=(), example_data="any")
+        return observation_spec
+
+
+def run_player(input_queue, output_queue):
```
vmoens, line 167: oh wow that's clever! Didn't think it'd take that to use sunfish
Do you think we should upstream that in ChessEnv?

examples/agents/ppo-chess-llm.py
```diff
@@ -0,0 +1,572 @@
...
+    )
+
+    for data in tqdm(collector):
+        # FIXME: reward seems to be getting wrongly propagated (e.g. sunfish's win gets reflected as llm's win)
+        rb.empty()
+        rb.extend(data)
+
+        for data in tqdm(rb):
+
+            data = gae(data)
```
vmoens, line 549: gae should go above the tqdm(rb)

You need the data to be presented sequentially to apply gae - even slices may not be ideal (better to compute it once and for all than at each iter).

Also I would put a no_grad around it for safety.

examples/agents/ppo-chess-llm.py
```diff
@@ -0,0 +1,572 @@
...
+    return actor_llm_policy, data_llm_policy, critic_llm_policy, tokenizer
+
+
+def play(env, data_llm_policy, actor_llm_policy, tokenizer):
+    from torchrl.data import LazyStackStorage, ReplayBuffer
+
+    rb = ReplayBuffer(
+        storage=LazyStackStorage(100),
+        batch_size=48,
+        sampler=SliceSamplerWithoutReplacement(slice_len=8, end_key=("next", "done")),
```
vmoens, line 482: in theory we may not need a slice sampler

examples/agents/ppo-chess-llm.py
```diff
@@ -0,0 +1,572 @@
...
+    optim = torch.optim.Adam(loss_module.parameters(), lr=2.5e-4)
+
+    gae = GAE(
+        value_network=Seq(actor_llm_policy[0], *critic_llm_policy),
+        gamma=0.99,
+        lmbda=0.95,
+        shifted=True,
+    )
+
+    for data in tqdm(collector):
```
vmoens, line 542: usually there are 3 nested loops
```python
for data in collector:
    for n in range(n_epochs):
        rb.extend(gae(data.copy())) 
        for d in rb:
            loss(d) # etc
```

examples/agents/ppo-chess-llm.py
```diff
@@ -0,0 +1,478 @@
...
+                        -1
+                    )  # [B, vocab_size] -> [B] -> [B, 1]
+                    tokenized_observation = torch.cat(
+                        (tokenized_observation, next_token), dim=1
+                    )
+                log_prob = output.logits[:, input_length - 1 :, :]
+                hidden = output.hidden_states[-1][:, input_length - 1, :]
+                return log_prob, hidden
+            else:
+                while True:
```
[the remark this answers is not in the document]
vmoens: Yeah I think it's good!
The while loop for generation is a bit worrisome - because then the samples are not really gathered from the LLM. 
One way to do this could be to terminate the game if the output is not valid and assign a losing reward (like -1 or smth)

examples/agents/ppo-chess-llm.py
mikaylagawarecki: this was something I wanted to talk about! if gae is before putting into replay buffer, we have no way to nicely append_transform to collector (I think)

PR #2773, opened 2025-02-07
[Bug] Fixes bug in GAE advantage estimation when `gammalmbda` is a Tensor

louisfaury:
## Description

## Motivation and Context

The vectorial implementation of GAE was using `done` to compute the exponential averaging weights instead of `terminated`.

## Types of changes

## Checklist

### DIFF
--- a/torchrl/objectives/value/functional.py
+++ b/torchrl/objectives/value/functional.py
@@ -311,16 +311,16 @@ def vec_generalized_advantage_estimate(
         == terminated.shape
     ):
         raise RuntimeError(SHAPE_ERR)
+
     dtype = state_value.dtype
     *batch_size, time_steps, lastdim = terminated.shape
 
-    value = gamma * lmbda
+    gammalmbdas = gamma * lmbda
+    gammalmbdas_is_tensor = (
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM thanks

torchrl/objectives/value/functional.py
```diff
@@ -332,6 +332,8 @@ def vec_generalized_advantage_estimate(
             lmbda=lmbda,
         )
 
+    not_terminated = (~terminated).to(dtype)
```
antoinebrl, line 335: I can't comment outside of the diff but ~15 lines below `not_terminated` is redefined. Maybe worth removing it as part of this PR.

vmoens:
We have a few errors in test_cost.py :)

louisfaury:
Looking more closely at broken test, I'm actually now understanding how the previous approach was sound. Classic Chesterton's fence I guess 🙃

louisfaury:
I guess here a single comment would have gone a long way -- basically simply reminding that a single sequence can hold different trajectories, hence we need to make sure to cut the "eligibility traces" not only on `terminated` but also on `truncated`.

PR #2777, opened 2025-02-10
[Feature] Transform for partial steps

vmoens:
* #2788
* #2784
* #2779
* __->__ #2777


torchrl/envs/transforms/transforms.py
```diff
@@ -923,7 +934,36 @@ def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+                # Batched envs have their own way of dealing with this - batched envs that are not batched-locked may fail here
+                if partial_steps.all():
+                    partial_steps = None
+                else:
+                    tensordict_batch_size = tensordict_in.batch_size
+                    partial_steps = partial_steps.view(tensordict_batch_size)
+                    tensordict_in = tensordict_in[partial_steps]
+            else:
+                if not partial_steps.any():
+                    next_tensordict = self._skip_tensordict(tensordict_in)
```
mikaylagawarecki, line 953: If we skip the step, would it also make sense to skip the transform on line 976?

`next_tensordict = self.transform._step(tensordict, next_tensordict)`

The setup I have is something like this -- `td['no_step']` is the condition for `ConditionalSkip`

(1) `data_policy` generates single token from `td['obs_tokens']`
(2) `append_transform` after (before in `.inv` but after in code) `ConditionalSkip` that overrides `_inv_step` to do the following 
            (a) append new token to `td[obs_tokens]`
            (b) Check whether end token is reached, if so check if answer is a valid move
                        (i) If valid, `td['no_step'] = False`
                        (ii) else reset `td['obs_tokens'] and `td['no_step'] = True`
            (c) If max steps has been reached, also reset
            (d) else `td['no_step'] = True`


Thing is, I have a bunch of transforms that patch `_step` that should only really be applied if the `_step` actually happened (e.g. scoring the move made, appending to the san history etc.)

I was able to write some if conditions that would make sure these transforms would be appropriately skipped, but I think it would be convenient if the transforms were skipped (is there perhaps a way to also wrap a transform with this sort of `Conditional-ness`?

vmoens: > Thing is, I have a bunch of transforms that patch _step that should only really be applied if the environment step actually happened (e.g. scoring the move made, appending to the san history etc.)

Yes that makes sense!
We just need to be extra careful that skipping the transforms won't cause the output tensordict to be of a different format.

torchrl/envs/common.py
```diff
@@ -1933,6 +1933,17 @@ def state_spec_unbatched(self, spec: Composite):
...
         self.state_spec = spec
 
+    def _skip_tensordict(self, tensordict):
+        # Creates a "skip" tensordict, ie a placeholder for when a step is skipped
+        next_tensordict = self.full_done_spec.zero()
+        next_tensordict.update(self.full_observation_spec.zero())
+        next_tensordict.update(self.full_reward_spec.zero())
+        # Copy the data from tensordict in `next`
+        next_tensordict.update(
+            tensordict.select(*next_tensordict.keys(True, True), strict=False)
```
mikaylagawarecki, line 1943: It seems that this line will not select any `NonTensorData` in `next_tensordict`, with the reason being that this line yields nothing I believe

https://github.com/pytorch/tensordict/blame/7dd385b96dc0a75a851ca8a100f4765791cbd7aa/tensordict/_td.py#L4473

Is this by design?

PR #2780, opened 2025-02-12
[Feature] Add `EnvBase.all_actions`

kurtamohler:
* __->__ #2780


vmoens (commented):
Nice feature!
Do you think we could use `Spec.enumerate()` as a default? Presumably, envs that don't have a list of actions are also those where enumerate would fail

kurtamohler:
Ah I didn't know about `enumerate`! Yeah that would be a good default

kurtamohler:
> Have you seen the mask transform by the way?

Nope, I hadn't seen it until now. That's cool!

torchrl/data/tensor_specs.py
```diff
@@ -1810,7 +1814,9 @@ def to_numpy(self, val: torch.Tensor, safe: bool = None) -> np.ndarray:
             return np.array(vals).reshape(tuple(val.shape))
         return val
 
-    def enumerate(self) -> torch.Tensor:
+    def enumerate(self, use_mask: bool = False) -> torch.Tensor:
+        if use_mask:
+            raise NotImplementedError
```
kurtamohler, line 1819: I left this unimplemented for OneHot for now, just because I want to focus on getting MCTS working. I'll submit a followup PR in the near future to add it--or if you'd rather I just add it to this PR, I'm happy to do that

vmoens (approved):
LGTM!

PR #2787, opened 2025-02-13
[BugFix] Apply inverse transform to input of `TransformedEnv._reset`

kurtamohler:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #2787

PR #2789, opened 2025-02-13
[BugFix] Fix MultiAction reset

kurtamohler:
* __->__ #2789

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -1375,7 +1375,7 @@ class Compose(Transform):
         return tensordict_reset
 
     def _reset_env_preprocess(self, tensordict: TensorDictBase) -> TensorDictBase:
-        for t in self.transforms:
+        for t in reversed(self.transforms):
             tensordict = t._reset_env_preprocess(tensordict)
         return tensordict
 
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -297,8 +297,8 @@ class Transform(nn.Module):
     def _reset_env_preprocess(self, tensordict: TensorDictBase) -> TensorDictBase:
         """Inverts the input to :meth:`TransformedEnv._reset`, if needed."""
         if self.enable_inv_on_reset:
-            with _set_missing_tolerance(self.transform, True):
-                tensordict = self.transform.inv(tensordict)
+            with _set_missing_tolerance(self, True):
+                tensordict = self.inv(tensordict)
         return tensordict
 
     def init(self, tensordict) -> None:
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -209,6 +209,7 @@ class Transform(nn.Module):
     """
 
     invertible = False
+    disable_inv_on_reset = False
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM!

torchrl/envs/transforms/transforms.py
```diff
@@ -209,6 +209,7 @@ class Transform(nn.Module):
     """
 
     invertible = False
+    disable_inv_on_reset = False
```
vmoens, line 212: Upon reflection: why not making this `True` by default?
Seems like it's more the exception than the rule (since we haven't been needing it so far)?

torchrl/envs/transforms/transforms.py
```diff
@@ -10437,6 +10442,8 @@ class MultiAction(Transform):
 
     """
 
+    disable_inv_on_reset = True
```
priba, line 10445: PR https://github.com/pytorch/rl/pull/2787 also broke the `ActionDiscretizer`, we should also add the `disable_inv_on_reset = True` into https://github.com/pytorch/rl/blob/c1d894fb6d105c50f99a0d1194a2c339f187e026/torchrl/envs/transforms/transforms.py#L9665

torchrl/envs/transforms/transforms.py
```diff
@@ -1264,6 +1266,9 @@ def map_transform(trsf):
         self.transforms = nn.ModuleList(transforms)
         for t in transforms:
             t.set_container(self)
+        self.disable_inv_on_reset = any(
+            [trsf.disable_inv_on_reset for trsf in transforms]
+        )
```
priba, line 1271: nit: There is no need to generate the whole list :)

```suggestion
        self.disable_inv_on_reset = any(trsf.disable_inv_on_reset for trsf in transforms)
```

torchrl/envs/transforms/transforms.py
kurtamohler: Yeah that makes sense

torchrl/envs/transforms/transforms.py
kurtamohler: I've changed this PR to set `enable_inv_on_reset = False` by default

kurtamohler:
I realized that `Transform` should probably have a method that does the conditional inversion before calling `env.reset`, so I added `Transform._reset_env_preprocess`. Something like this was necessary in order for `Compose` to work correctly.

Before, inside `TransformedEnv._reset`, I was calling `self.transform.inv` conditionally, depending on if the transform had the enable flag set. The issue with that was in the case that the `self.transform` is `Compose`, one of the transforms in the chain doesn't support input inversion (like `MultiAction`), and one of the transforms requires input inversion (like `Hash`). To support cases like that, we need to defer to each transform in the chain to determine whether it needs to invert the input

torchrl/envs/transforms/transforms.py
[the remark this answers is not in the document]
kurtamohler: I believe this is not needed anymore, after my recent change: https://github.com/pytorch/rl/pull/2789#issuecomment-2664497042

priba:
Thanks @kurtamohler ! It looks much cleaner now.

PR #2791, opened 2025-02-17
[BugFix] PPOs with composite distribution

louisfaury:
## Description

I believe there is a bug in PPOs' implementation when both `prev_log_prob` and `log_prob` are TensorDicts.

## Motivation and Context

In the setting were both `prev_log_prob` and `log_prob` are TensorDicts, we were clamping `prev_log_prob - log_prob` directly, instead of their sum over features. 

## Types of changes

## Checklist

### DIFF
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -527,12 +527,9 @@ class PPOLoss(LossModule):
             self.actor_network
         ) if self.functional else contextlib.nullcontext():
             dist = self.actor_network.get_dist(tensordict)
-        if isinstance(dist, CompositeDistribution):
-            is_composite = True
-        else:
-            is_composite = False
 
-        # current log_prob of actions
+        is_composite = isinstance(dist, CompositeDistribution)
+
         if is_composite:
             action = tensordict.select(
                 *(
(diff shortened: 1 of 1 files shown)


torchrl/objectives/ppo.py
```diff
@@ -562,25 +559,26 @@ def _log_weight(
...
+
+                    if is_tensor_collection(log_prob):
+                        log_prob = _sum_td_features(log_prob)
+                        log_prob.view_as(prev_log_prob)
 
         log_weight = (log_prob - prev_log_prob).unsqueeze(-1)
+        if is_tensor_collection(log_weight):
+            log_weight = _sum_td_features(log_weight)
+            log_weight = log_weight.view(adv_shape).unsqueeze(-1)
+
```
louisfaury, line 581: That is the main change for this method, which is also now consistent with type hints.

torchrl/objectives/ppo.py
```diff
@@ -987,8 +982,6 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
             # to different, unrelated trajectories, which is not standard. Still, it can give an idea of the weights'
             # dispersion.
             lw = log_weight.squeeze()
-            if not isinstance(lw, torch.Tensor):
-                lw = _sum_td_features(lw)
             ess = (2 * lw.logsumexp(0) - (2 * lw).logsumexp(0)).exp()
             batch = log_weight.shape[0]
```
louisfaury, line 986: The main error is two lines below; `clamp` was applied to the TensorDict `log_weight` _before_ it is summed over the feature dimension.

louisfaury (commented):
I checked that all PPO tests are still passing; two comments:
1. We should probably have test to catch this,
2. I'm planning to do a docstring pass on the entire PPO stack so things can be much clearer (some operations are a bit obscure at first read).

PR #2794, opened 2025-02-19
[Feature] Add `MCTSForest/Tree.to_string`

kurtamohler:
* #2796
* #2795
* __->__ #2794

### DIFF
--- a/torchrl/data/map/tree.py
+++ b/torchrl/data/map/tree.py
@@ -604,6 +604,42 @@ class Tree(TensorClass["nocast"]):
             f"Unknown plotting backend {backend} with figure {figure}."
         )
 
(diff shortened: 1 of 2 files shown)


torchrl/data/map/tree.py
```diff
@@ -604,6 +604,42 @@ def plot(
             f"Unknown plotting backend {backend} with figure {figure}."
         )
 
+    def to_string(self, node_format_fn):
```
vmoens, line 607: we don't want a default for this?

vmoens (commented):
Looks good! I would just consider either a default value and / or a simple example in the docstrings (you can just copy paste the test, that will be good enough)

vmoens (approved):
LGTM thanks!

PR #2795, opened 2025-02-19
[Doc] Add docstring for `MCTSForest.extend`

kurtamohler:
* #2796
* __->__ #2795
* #2794

### DIFF
--- a/torchrl/data/map/tree.py
+++ b/torchrl/data/map/tree.py
@@ -1003,6 +1003,33 @@ class MCTSForest:
             self.max_size = self.data_map.max_size
 
     def extend(self, rollout, *, return_node: bool = False):
+        """Add a rollout to the forest.
+
+        Nodes are only added to a tree at points where rollouts diverge from
+        each other and at the endpoints of rollouts.
+
+        If there is no existing tree that matches the first steps of the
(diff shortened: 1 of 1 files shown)


torchrl/data/map/tree.py
```diff
@@ -1003,6 +1003,33 @@ def _make_node_map(self, source, dest):
...
+        some step, a new node is created before the step where the rollouts
+        diverge, and a leaf node is created for the final step of the rollout.
+        If all of the rollout's steps match with a previously added rollout,
+        nothing changes.  If the rollout matches up to a leaf node of a tree but
+        continues beyond it, that node is extended to the end of the rollout,
+        and no new nodes are created.
+
+        Args:
+            rollout (TensorDict): The rollout to add to the forest.
+            return_node (bool, optional): If True, the method returns the added
```
vmoens, line 1026:
```suggestion
            return_node (bool, optional): If ``True``, the method returns the added
```

vmoens (commented):
Do we have a quick example that could be copy pasted for here? I think it brings a lot of value for users if users can just do that and run a simple forest in a notebook

kurtamohler:
I've added an example

vmoens (approved):
LGTM thanks a mil!

PR #2796, opened 2025-02-19
[DRAFT, Example] Add MCTS example

kurtamohler:
* (to be filled)

### DIFF
--- a/examples/trees/mcts.py
+++ b/examples/trees/mcts.py
@@ -1,0 +1,220 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+import torchrl
+from tensordict import TensorDict
+
+pgn_or_fen = "fen"
+
+env = torchrl.envs.ChessEnv(
+    include_pgn=False,
+    include_fen=True,
+    include_hash=True,
+    include_hash_inv=True,
+    include_san=True,
+    stateful=True,
+    mask_actions=True,
+)
+
+
+def transform_reward(td):
+    if "reward" not in td:
+        return td
+    reward = td["reward"]
+    if reward == 0.5:
+        td["reward"] = 0
+    elif reward == 1 and td["turn"]:
+        td["reward"] = -td["reward"]
+    return td
+
+
+# ChessEnv sets the reward to 0.5 for a draw and 1 for a win for either player.
+# Need to transform the reward to be:
+#   white win = 1
+#   draw = 0
+#   black win = -1
+env.append_transform(transform_reward)
+
+forest = torchrl.data.MCTSForest()
+forest.reward_keys = env.reward_keys + ["_visits", "_reward_sum"]
+forest.done_keys = env.done_keys
+forest.action_keys = env.action_keys
+forest.observation_keys = [f"{pgn_or_fen}_hash", "turn", "action_mask"]
+
+C = 2.0**0.5
+
+
+def traversal_priority_UCB1(tree):
+    if tree.rollout[-1]["next", "_visits"] == 0:
+        res = float("inf")
+    else:
+        if tree.parent.rollout is None:
+            parent_visits = 0
+            for child in tree.parent.subtree:
+                parent_visits += child.rollout[-1]["next", "_visits"]
+        else:
+            parent_visits = tree.parent.rollout[-1]["next", "_visits"]
+            assert parent_visits > 0
+
+        value_avg = (
+            tree.rollout[-1]["next", "_reward_sum"]
+            / tree.rollout[-1]["next", "_visits"]
+        )
+
+        # If it's black's turn, flip the reward, since black wants to optimize
+        # for the lowest reward.
+        if not tree.rollout[0]["turn"]:
+            value_avg = -value_avg
+
+        res = (
+            value_avg
+            + C
+            * torch.sqrt(torch.log(parent_visits) / tree.rollout[-1]["next", "_visits"])
+        ).item()
+
+    return res
+
+
+def _traverse_MCTS_one_step(forest, tree, env, max_rollout_steps):
+    done = False
+    trees_visited = []
+
+    while not done:
+        if tree.subtree is None:
+            td_tree = tree.rollout[-1]["next"]
+
+            if (td_tree["_visits"] > 0 or tree.parent is None) and not td_tree["done"]:
+                actions = env.all_actions(td_tree)
+                subtrees = []
+
+                for action in actions:
+                    td = env.step(env.reset(td_tree.clone()).update(action)).update(
+                        TensorDict(
+                            {
+                                ("next", "_visits"): 0,
+                                ("next", "_reward_sum"): env.reward_spec.zeros(),
+                            }
+                        )
+                    )
+
+                    new_node = torchrl.data.Tree(
+                        rollout=td.unsqueeze(0),
+                        node_data=td["next"].select(*forest.node_map.in_keys),
+                    )
+                    subtrees.append(new_node)
+
+                tree.subtree = TensorDict.lazy_stack(subtrees)
+                chosen_idx = torch.randint(0, len(subtrees), ()).item()
+                rollout_state = subtrees[chosen_idx].rollout[-1]["next"]
+
+            else:
+                rollout_state = td_tree
+
+            if rollout_state["done"]:
+                rollout_reward = rollout_state["reward"]
+            else:
+                rollout = env.rollout(
+                    max_steps=max_rollout_steps,
+                    tensordict=rollout_state,
+                )
+                rollout_reward = rollout[-1]["next", "reward"]
+            done = True
+
+        else:
+            priorities = torch.tensor(
+                [traversal_priority_UCB1(subtree) for subtree in tree.subtree]
+            )
+            chosen_idx = torch.argmax(priorities).item()
+            tree = tree.subtree[chosen_idx]
+            trees_visited.append(tree)
+
+    for tree in trees_visited:
+        td = tree.rollout[-1]["next"]
+        td["_visits"] += 1
+        td["_reward_sum"] += rollout_reward
+
+
+def traverse_MCTS(forest, root, env, num_steps, max_rollout_steps):
+    """Performs Monte-Carlo tree search in an environment.
+
+    Args:
+        forest (MCTSForest): Forest of the tree to update. If the tree does not
+            exist yet, it is added.
+        root (TensorDict): The root step of the tree to update.
+        env (EnvBase): Environment to performs actions in.
+        num_steps (int): Number of iterations to traverse.
+        max_rollout_steps (int): Maximum number of steps for each rollout.
+    """
+    if root not in forest:
+        for action in env.all_actions(root.clone()):
+            td = env.step(env.reset(root.clone()).update(action)).update(
+                TensorDict(
+                    {
+                        ("next", "_visits"): 0,
+                        ("next", "_reward_sum"): env.reward_spec.zeros(),
+                    }
+                )
+            )
+            forest.extend(td.unsqueeze(0))
+
+    tree = forest.get_tree(root)
+
+    for _ in range(num_steps):
+        _traverse_MCTS_one_step(forest, tree, env, max_rollout_steps)
+
+    return tree
+
+
+def tree_format_fn(tree):
+    td = tree.rollout[-1]["next"]
+    return [
+        td["san"],
+        td[pgn_or_fen].split("\n")[-1],
+        td["_reward_sum"].item(),
+        td["_visits"].item(),
+    ]
+
+
+def get_best_move(fen, mcts_steps, rollout_steps):
+    root = env.reset(TensorDict({"fen": fen}))
+    tree = traverse_MCTS(forest, root, env, mcts_steps, rollout_steps)
+
+    # print('------------------------------')
+    # print(tree.to_string(tree_format_fn))
+    # print('------------------------------')
+
+    moves = []
+
+    for subtree in tree.subtree:
+        san = subtree.rollout[0]["next", "san"]
+        reward_sum = subtree.rollout[-1]["next", "_reward_sum"]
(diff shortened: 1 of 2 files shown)


kurtamohler:
This seems to work, but at the moment, it is about 100x slower than the one I implemented outside of TorchRL [here](https://github.com/kurtamohler/mcts/blob/fe903b47db81214de7a03e534cad9e91b30d16a7/chess_eval.py). I will look into speeding it up.

examples/trees/mcts.py
```diff
@@ -0,0 +1,220 @@
...
+                rollout = env.rollout(
+                    max_steps=max_rollout_steps,
+                    tensordict=rollout_state,
+                )
+                rollout_reward = rollout[-1]["next", "reward"]
+            done = True
+
+        else:
+            priorities = torch.tensor(
+                [traversal_priority_UCB1(subtree) for subtree in tree.subtree]
```
kurtamohler, line 128: Note to self: this calculation can be batched to improve performance

examples/trees/mcts.py
```diff
@@ -0,0 +1,213 @@
...
+pgn_or_fen = "fen"
+
+env = torchrl.envs.ChessEnv(
+    include_pgn=False,
+    include_fen=True,
+    include_hash=True,
+    include_hash_inv=True,
+    include_san=True,
+    stateful=True,
+    mask_actions=True,
```
kurtamohler, line 19: Note to self: from performance profiling, it looks like action masking takes up a significant chunk of the runtime, which makes sense--it's a 29275 element tensor created at each step. So I'll need to try turning this off, which will require adding support for `ChessEnv.all_actions` in the absence of the mask

examples/trees/mcts.py
kurtamohler: Batching this improved performance by around 2x

kurtamohler:
Just wanted to mention that I am actively working on this. It was a little tricky to get ChessEnv working correctly with `mask_actions=False`, but I almost have it.

vmoens:
> Just wanted to mention that I am actively working on this. It was a little tricky to get ChessEnv working correctly with `mask_actions=False`, but I almost have it.

Super cool thanks!

kurtamohler:
I was able to improve the performance a fair bit, but it is still around 17x slower than my standalone example code. But I suppose we can make further performance improvements later down the road. At the moment, this PR is kind of a mess, so I'll fix it up and probably split out the stuff that is not directly related to MCTS into a separate PR

vmoens:
> I was able to improve the performance a fair bit, but it is now around 17x slower (down from 100x) than my standalone example code. But I suppose we can make further performance improvements later down the road. At the moment, this PR is kind of a mess, so I'll fix it up and probably split out the stuff that is not directly related to MCTS into a separate PR

Do you know what's causing the slowdown? TensorDict overhead? 
It'd be great if we could pinpoint what it is that is responsible for that!

kurtamohler:
I've been using [py-spy](https://github.com/benfred/py-spy) to find bottlenecks and improve them.  Here's the flamegraph that it produces right now: 
profile


At the moment, `ChessEnv._legal_moves_to_index` is one of the the biggest slowdowns.

torchrl/envs/transforms/transforms.py
```diff
@@ -1094,6 +1094,30 @@ def rand_action(self, tensordict: Optional[TensorDictBase] = None) -> TensorDict
...
+        return action
+
+    @property
+    def _overrides_action_generator_funcs(self):
+        return self.base_env._overrides_action_generator_funcs
+
+    def all_actions(
+        self, tensordict: Optional[TensorDictBase] = None
+    ) -> TensorDictBase:
+        if not self._overrides_action_generator_funcs:
```
kurtamohler, line 1110: In order to allow ChessEnv to override `all_actions` and `_rand_action` in the case where `mask_actions=False`, I added this hacky `_overrides_action_generator_funcs` setting. I only added that to get it working to confirm that avoiding the mask increases performance. I intended to remove `_overrides_action_generator_funcs` in favor of always calling the base env's `all_actions`/`_rand_action` functions. But that is proving to be pretty difficult and it's slowing down my progress on MCTS itself.

There may be a good solution to this, but taking a step back to consider the bigger picture, I think it would be more pragmatic for me to stop trying to make `mask_actions=False` work for now. The `mcts.py` script currently takes ~9.2 s without the mask versus ~12 s with the mask. It would be nice to achieve that speedup eventually, but in order to develop a good MCTS API, it just don't think it's necessary. Furthermore, once we have the ability to combine MCTS with a neural network evaluation function (which I assume is one of the goals), then I think the time it takes to generate the mask will no longer be a significant bottleneck.

So I'll back out the changes I made with respect to the action mask and start focusing more on the MCTS API

kurtamohler: As as side note, looking at the performance profile, one relatively simple way to improve ChessEnv performance is to avoid the `board.san(m)` call in `_legal_moves_to_index`, which converts all the legal moves in the current position from UCI format to SAN. That function is apparently pretty slow. Instead of indexing SAN moves, we could index the UCI moves to avoid doing this conversion

kurtamohler: As a side note, the `board.san(m)` call in `_legal_moves_to_index` is pretty slow, and avoiding it would improve performance. At the moment, we rely on the SAN format because the action indices correspond to one of all of the possible ~30K SAN strings. But we could choose to index it in a different way that would be more performant.

I would propose that we directly convert each `chess.Move` into an index using its `from_square`, `to_square`, and `promotion` properties, which are all integers. `from_square` and `to_square` are just numbers between 0 and 63. `promotion` is a number that indicates which piece a pawn will promote to, if applicable. In the chess library, `promotion` can specify any of the 6 kinds of pieces or none, so it has 7 possible values (although by the rules of chess, you can only promote to 4 different kinds of pieces). We can convert this to a number like so:

```
index = move.promotion * (64 * 64) + move.from_square * 64 + move.to_square
```

Then we wouldn't have to look up the SAN string in a big list  (which I've modified to be a map in this PR) to obtain the index, and we wouldn't have to look up the index in a big list to covert an index back to a Move object--we would just extract the `Move` properties directly from the index.

kurtamohler: I figured I might as well try out my idea for a more performant action indexing scheme, since it's relatively simple to implement. It seems to have worked fairly well. Runtime of the mcts script is typically ~10-12 s now, so a decent speedup over the previous 12-15 s.

kurtamohler: Actually, that change breaks some of the ChessEnv tests. Changing the index to be something other than SAN-based would mean that in order to convert a SAN string to a move index, as is done in `test_san` and `test_reward`, we would first need to obtain a `chess.Board` object that reflects the state of the board, since the conversion of a SAN string to a `chess.Move` object depends on the state of the board. It's certainly possible to do this, but I'll drop this idea for now since it turns out not to be as straightforward as I thought

kurtamohler:
I think this is ready for review. I'll add MCTS APIs in subsequent PRs and update the example script accordingly

torchrl/modules/mcts/mcts.py
```diff
@@ -0,0 +1,118 @@
...
+
+# TODO: Allow user to specify different priority functions with PR #2358
+def _traversal_priority_UCB1(tree):
+    subtree = tree.subtree
+    visits = subtree.visits
+    reward_sum = subtree.wins
+
+    # If it's black's turn, flip the reward, since black wants to optimize for
+    # the lowest reward, not highest.
+    # TODO: Need a more generic way to do this, since not all use cases of MCTS
```
kurtamohler, line 24: We'll need to come up with a good way to specify how to process the exploitation value for each different player.

During an MCTS traversal, when we visit each node, we have to rank all the child nodes and decide which one to traverse to, using one of the standard exploration/exploitation formulas. I only know about the UCB1 formula at the moment, so I'll focus on that one. To calculate the expoitation value of each child node, one of the values that the UCB1 formula operates on is the "reward average" attached to each child node--the sum of all the rewards of the rollouts that have been performed under the child node divided by the number of times that child node has been visited during traversals.

Let's say that at the end of a rollout, the reward value that the chess environment gives for a white win is 1, black win is -1, and draw is 0.  In order to do the exploitation part correctly, we should assume that each player wants to maximize their chances of winning. So if it's white's turn at a particular node, we want exploitation actions to maximize the reward average.  But if it's black's turn, we want expoitation actions to minimize the reward average, so we have to flip the sign of the reward average.

Or let's say we instead want to use a two-element reward. A white win is `[1, -1]`, black win is `[-1, 1]`, and draw is `[0, 0]`. Now the reward average at each node of the MCTS tree has two elements. When it's white's turn at a particular node, we want to look at the first element of the reward average of each child node. When it's black's turn, we want to look at the second element.

Ideally, our MCTS API should be able to handle both of the above reward schemes, and any other sensible kind of reward scheme that users would want to use.

kurtamohler: I'm thinking of just adding an argument to the `MCTS` function called `select_player_reward_fn`, or something along those lines, which is a callable with the signature `(td, reward_average)`. In the two examples I gave above, the user could specify something like this:

`select_player_reward_fn=lambda td, reward_avg: reward_avg if td["turn"] else -reward_avg` for 1-d reward between 1 and -1

`select_player_reward_fn=lambda td, reward_avg: reward_avg[td["turn"]]` for 2-d reward

kurtamohler: I was curious about how other libraries handle this, so I looked into a couple of them.

The `mcts` Python library has a sign flip for two-player games. The total reward is multiplied by the player value (<https://github.com/pbsinclair42/MCTS/blob/f4b6226ce06840e51b66bdca4bedbe2c4c143012/mcts.py#L112>), which is either 1 or -1 in the example scripts of this repo. So this library seems to be specifically for two player games.

The MCTS implementation within the `open_spiel` library has an n-element reward, one for each player, and chooses the reward for whichever player whose turn it is, like the default behavior I proposed in the last comment: <https://github.com/google-deepmind/open_spiel/blob/8296179b697644cf957c7c9313f594c062cbd17c/open_spiel/python/algorithms/mcts.py#L368-L369>

kurtamohler: I think maybe our implementation of MCTS should require the reward to have one element per player, and also require that the environment output spec has a key indicating which player's turn it is. Or at least that could be the default behavior and we could allow the user to override that with a lambda.

Let me know what you think @vmoens.

vmoens: Usually the way we frame things in torchrl is that if you have 2 agents, we have
```
reward_spec = Composite(
    agent0=Composite(Unbounded(...)),
    agent1=Composite(Unbounded(...)),
)
```
(you can perfectly substitute `agent0` with `white`)

If you have separate actions you can do the same. Now we should even be able to support easily turn-based environments. One way to do this would be to use the new `ConditionalPolicySwitch` transform, or simply to have `action=torch.zeros((0,))` when the player is not acting at this turn.

I guess observations will be shared in the case of chess.

> The mcts Python library has a sign flip for two-player games

That would only scale to two players though, no?

> The MCTS implementation within the open_spiel library has an n-element reward, one for each player, and chooses the reward for whichever player whose turn it is

That's better than flipping the sign but IMO it's also not awesome as it doesn't show explicitly who owns what reward.

> I think maybe our implementation of MCTS should require the reward to have one element per player, and also require that the environment output spec has a key indicating which player's turn it is. Or at least that could be the default behavior and we could allow the user to override it with a callable argument.

So let's imagine that we're at the root (observing a given dispositon of the board) and we ask ourselves what action should be taken.
We can have an observation spec
```python
TensorDict(
  agent0=TensorDict(turn=torch.Tensor(True, ...)),
  agent1=TensorDict(turn=torch.Tensor(False, ...)),
  ...
)
```
that informes us on the current turn. Then we can use that to index the rewards.
Wdyt?

I can give a shot at making Chess look like this if you think that would help.

vmoens: Quick hacky example:
```python
from torchrl.envs import ChessEnv

from torchrl.envs import Transform
from torchrl.data import Unbounded, Composite
import torch
class TurnBasedChess(Transform):
    def transform_observation_spec(self, obsspec):
        obsspec["agent0", "turn"] = Unbounded(dtype=torch.bool, shape=())
        obsspec["agent1", "turn"] = Unbounded(dtype=torch.bool, shape=())
        return obsspec
    def transform_reward_spec(self, reward_spec):
        reward = reward_spec["reward"].clone()
        del reward_spec["reward"]
        return Composite(
            agent0=Composite(reward=reward),
            agent1=Composite(reward=reward),
        )
    def _reset(self, _td, td):
        td["agent0", "turn"] = True
        td["agent1", "turn"] = False
        print(f"{td=}")
        return td
    def _step(self, td, td_next):
        td_next["agent0", "turn"] = ~td["agent0", "turn"]
        td_next["agent1", "turn"] = ~td["agent1", "turn"]
        
        td_next["agent0", "reward"] = td_next["reward"]
        td_next["agent1", "reward"] = -td_next["reward"]
        del td_next["reward"]
        
        print(f"{td_next=}")
        return td_next
base_env = ChessEnv()
env = base_env.append_transform(TurnBasedChess())
env.rollout(3)
```

kurtamohler: To make sure I understand, I think you're suggesting:
  * The MCTS module should expect the following two points and understand how to pull out the reward for whichever agent's turn it is.
  * Reward spec should have one reward per agent, keyed by the agent name. Example key: `("agent0", "reward")`
  * Observation spec should contain a bool for each agent indicating whether it's their turn, also keyed by the agent name. It should only ever be one agent's turn at each step. Example key: `("agent0", "turn")`. (Presumably we would give MCTS an argument like `turn_key` that tells it what key the turn value is under. I suppose in most cases it would just be `turn_key="turn"`.)

That basically seems reasonable to me. I think I should be able to apply a couple transforms on top of ChessEnv to get that working for testing.

I still wonder though, should we provide a way for the user to override the above reward processing behavior and define their own method for pulling out and processing a reward value?

PR #2801, opened 2025-02-21
[BugFix] correct dim for resolving dtype in _split_and_pad_sequence

### LINKED ISSUE #2758: [BUG] dtype in _split_and_pad_sequence is established using incorrect dim
In  [torchrl/objectives/value/utils.py#L287-L290](https://github.com/pytorch/rl/blob/cb37521e1ee81ab186db1de39e4ce8595057772e/torchrl/objectives/value/utils.py#L287-L290), the code currently uses `tensor.shape[-2]` to choose between `torch.int16` and `torch.int32`. This should use the time dimension (`time_dim`) instead, especially since it's used in `_fast_td_lambda_return_estimate` where the inputs are transposed first. As a result, the condition may be incorrect for certain input shapes (when the size of `time_dim` exceeds the `int16` range, but `F` dimension of `_fast_td_lambda_return_estimate` inputs is within this range).

Code Reference:
```
    # int16 supports length up to 32767
    dtype = (
        torch.int16 if tensor.shape[-2] < torch.iinfo(torch.int16).max else torch.int32
    )
```

Proposed Fix:
```
    dtype = (
        torch.int16 if tensor.size(time_dim) < torch.iinfo(torch.int16).max else torch.int32
    )
```

KubaMichalczyk:
Close #2758

### DIFF
--- a/torchrl/objectives/value/utils.py
+++ b/torchrl/objectives/value/utils.py
@@ -286,7 +286,9 @@ def _split_and_pad_sequence(
 
     # int16 supports length up to 32767
     dtype = (
-        torch.int16 if tensor.shape[-2] < torch.iinfo(torch.int16).max else torch.int32
+        torch.int16
+        if tensor.size(time_dim) < torch.iinfo(torch.int16).max
+        else torch.int32
     )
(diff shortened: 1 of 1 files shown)


vmoens (approved):
LGTM thanks!

PR #2818, opened 2025-02-28
[Feature] LLMEnv and DataLoadingPrimer

vmoens:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* (to be filled)

PR #2819, opened 2025-02-28
[Lint] pyupgrade

vmoens:
* #2823
* #2822
* #2821
* __->__ #2819
* #2818
* #2817
* #2816
* #2815
* #2732


build_tools/setup_helpers/__init__.py
```diff
@@ -4,3 +4,5 @@
 # LICENSE file in the root directory of this source tree.
 
 from .extension import CMakeBuild, get_ext_modules  # noqa
```
antoinebrl, line 6: Now that there is an explicit reexport, is the `# noqa` still relevant?

test/mocking_classes.py
```diff
@@ -133,7 +131,7 @@ def __init__(
     def maxstep(self):
         return 100
 
-    def _set_seed(self, seed: Optional[int]):
+    def _set_seed(self, seed: int | None):
```
antoinebrl, line 134: I think this type notation was introduced in py3.10. Do you still targeting 3.9 ?

vmoens: there should be a `from __future__ import annotations` in each file - maybe not this one!

PR #2832, opened 2025-03-05
[Tutorial] LLM integration

vmoens:
* __->__ #2832


sota-implementations/post-training/grpo.py
```diff
@@ -0,0 +1,167 @@
...
+    )
+
+    # Collector
+    train_model = GPT2LMHeadModel.from_pretrained("gpt2")
+    collector = SyncDataCollector(
+        env,
+        policy,
+        frames_per_batch=args.steps_per_batch,
+        total_frames=1_000_000,
+        local_weights_updater=HF2vLLMLocalWeightUpdater(
```
mikaylagawarecki, line 127: I'm curious to see how this would look when train_model is sharded :b

vmoens: Yeah there is still some work to be done!
IIUC your implementation you do a `full_tensor()` then you send that to the vllm weights right?

sota-implementations/post-training/grpo.py
```diff
@@ -0,0 +1,167 @@
...
+    ref_model = GPT2LMHeadModel.from_pretrained("gpt2")
+    TensorDict.from_module(ref_model).data.to_module(ref_model)
+    ref_model = from_hf_transformers(
+        ref_model,
+        tokenizer=tokenizer,
+        from_text=False,
+        generate=False,
+        return_log_probs=True,
+    )
+    env.append_transform(
```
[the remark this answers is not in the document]
mikaylagawarecki: With just reward model and ref_model I think this API is ok

But I don't think this API would be encompassing for the case where there's a more complex graph of dependencies between transforms 


Also, with this API, who is responsible for consolidating the tds ?

PR #2835, opened 2025-03-06
[Feature] RayReplayBuffer

vmoens:
* __->__ #2835


torchrl/data/replay_buffers/replay_buffers.py
```diff
@@ -90,12 +95,17 @@ class ReplayBuffer:
...
+        transform (Transform or Callable[[Any], Any], optional): Transform to be executed when
             :meth:`sample` is called.
             To chain transforms use the :class:`~torchrl.envs.Compose` class.
             Transforms should be used with :class:`tensordict.TensorDict`
             content. A generic callable can also be passed if the replay buffer
             is used with PyTree structures (see example below).
+            Unlike storages, writers and samplers, transform constructors must
+            be passed as separate keyword argument :attr:`transform_factory`,
+            as it is impossible to distinguish a constructor from a transform.
+        transform_factory (Callable[[], Callable], optional): a factory for the
```
mikaylagawarecki, line 107: If we do this here for all of the buffers it kind of feels like the API for collector should match 😄

torchrl/data/replay_buffers/ray_buffer.py
```diff
@@ -0,0 +1,199 @@
...
+
+                ray_init_kwargs = DEFAULT_RAY_INIT_CONFIG
+            ray.init(**ray_init_kwargs)
+
+        remote_cls = ReplayBuffer.as_remote(DEFAULT_REMOTE_CLASS_CONFIG).remote
+        self._rb = remote_cls(*args, **kwargs)
+
+    def sample(self, *args, **kwargs):
+        pending_task = self._rb.sample.remote(*args, **kwargs)
+        return ray.get(pending_task)
```
mikaylagawarecki, line 127: very n00b qn re how sharing this across distributed workers works

is the idea that by initializing RayReplayBuffer in __main__ and then passing it to each DistributedActorWorker, when .sample() in one of the DistributedActorWorker's methods is called it would do the .get on the resources that have been allocated to that specific DistributedActorWorker?

torchrl/collectors/distributed/ray.py
```diff
@@ -312,7 +314,9 @@ def __init__(
         num_collectors: int = None,
         update_after_each_batch=False,
         max_weight_update_interval=-1,
+        replay_buffer: ReplayBuffer = None,
```
mikaylagawarecki, line 317: nit: this is not documented

torchrl/collectors/distributed/ray.py
```diff
@@ -577,17 +591,33 @@ def stop_remote_collectors(self):
             )  # This will interrupt any running tasks on the actor, causing them to fail immediately
 
     def iterator(self):
```
mikaylagawarecki, line 593: Should iterator raise an error for the async case? Or what happens when trying to iterate over this collector when replaybuffer was passed?

torchrl/data/replay_buffers/replay_buffers.py
vmoens: Yeah I agree
The more I think about it the more it's obvious we should just pass a factory to the collectors

torchrl/data/replay_buffers/ray_buffer.py
vmoens: Yes, not on main but within the worker

torchrl/collectors/distributed/ray.py
vmoens: You should get None outputs, and the buffer is filled each time you call next()

torchrl/data/replay_buffers/ray_buffer.py
wbinventor: @vmoens How would one use RayReplayBuffer in the context of PyTorch DDP/FSDP? Is this even possible? For example, can there be a single RayReplayBuffer that each DDP/FSDP process uses?

vmoens: Yes that should work: instantiate somewhere and pass the remote handler around.

PR #2836, opened 2025-03-06
[Setup] Remove `distutils` imports

antoinebrl:
## Description

`setuptools` is no longer available by default starting in Python 3.12. As such, `distutils` is no longer available. `setuptools` still comes as part of `torch`, making it a transitive dependency for `torchrl`. While `setuptools` and `distutils` can still be used as part of the build process, this PR removes all `distutils` imports inside the code. Effectively, it is only about re-implementing the helper function `strtobool`.

### DIFF
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -18,7 +18,6 @@ import traceback
 import warnings
 from contextlib import nullcontext
 from copy import copy
-from distutils.util import strtobool
 from functools import wraps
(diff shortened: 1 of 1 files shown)


vmoens (approved):
LGTM
maybe that will be needed [in tensordict too](https://github.com/search?q=repo%3Apytorch%2Ftensordict%20distutils&type=code)?

PR #2841, opened 2025-03-10
[Feature] policy factory for collectors

vmoens:
* #2848
* __->__ #2841
* #2835


torchrl/collectors/collectors.py
```diff
@@ -315,6 +317,43 @@ def __len__(self) -> int:
             return -(self.total_frames // -self.requested_frames_per_batch)
         raise RuntimeError("Non-terminating collectors do not have a length")
 
+    @classmethod
+    def from_policy_factory(
```
vmoens, line 321: @mikaylagawarecki 

This is a bit convoluted.

We could also have one more kwarg in the constructors, but since there are many kwargs there already, and many collector subclasses, IDK if that's the most economical solution.

Happy to make this a kwargs that is exclusive with the `policy` arg if you think it's more suited

mikaylagawarecki: I think this is ok

Just a question -- do you think that giving the CustomCollectorCls a  __name__ that depended on its cls would be helpful for debugging, or would that make it harder to find where the wrapper class was defined 😅

mikaylagawarecki (approved):
To double check my understanding -- the change is that policy is now instantiated at collector init time rather than before, which is helpful for distributed settings

PR #2851, opened 2025-03-14
[Feature] Add option for auto-resetting envs in GAE

lin-erica:
## Description

Previously, when calculating the GAE, torchrl assumes that there is a valid `s_(t+1)` return from the environment. However, for auto-resetting environments, that is not true at the end of the episode, as it just jumps to the first state of the next episode. This PR adds a feature to the GAE class, allowing users to specify if it's an auto-resetting env, in which case it will use the `s_t`  as an approximation for bootstrapping. 

## Motivation and Context

Without this change, we were getting much slower convergence rates during training for IsaacLab (an autoresetting env).

## Types of changes

## Checklist

### DIFF
--- a/torchrl/objectives/value/advantages.py
+++ b/torchrl/objectives/value/advantages.py
@@ -1239,6 +1239,9 @@ class GAE(ValueEstimatorBase):
             :meth:`~.value_estimate`.
             Negative dimensions are considered with respect to the input
             tensordict.
+        auto_reset_env (bool, optional): if ``True``, the last ``"next"`` state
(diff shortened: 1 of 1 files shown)


lin-erica:
@vmoens Perhaps I missed it, but I didn't see any tests that were testing the GAE class. If there is, I'm happy to add a relevant test as well.

vmoens (approved):
LGTM! Sorry for the delay in reviewing this!

To answer your question: Not sure what unit test we want for this feature, so I'm willing to consider merging it as is.

We do test the GAE in test_costs.py I believe... I you have an idea of what to test we can add it

PR #2856, opened 2025-03-17
[Feature] Set padded token log-prob to 0.0

vmoens:
* #2861
* #2859
* #2857
* __->__ #2856
* #2854
* #2852

### DIFF
--- a/torchrl/modules/llm/vllm_policy.py
+++ b/torchrl/modules/llm/vllm_policy.py
@@ -24,7 +24,7 @@ from tensordict.nn import (
     TensorDictSequential as Seq,
     WrapModule,
 )
-from tensordict.utils import _zip_strict
+from tensordict.utils import _zip_strict, expand_as_right
 
 from torchrl.data import LLMData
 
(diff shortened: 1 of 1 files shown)


torchrl/modules/llm/vllm_policy.py
```diff
@@ -280,10 +281,18 @@ def get_output_tokens_and_log_probs(td, padding_value=padding_value):
...
                 tokens_response_td.rename_key_("logprobs", "log_probs")
-                # td["log_probs"] = outputs.logprobs.unsqueeze(-1)
+                if padded_values.any():
+                    print(
+                        "padded_values:",
+                        padded_values.sum(),
+                        torch.where(padded_values),
+                    )
+                    lps = tokens_response_td["log_probs"]
+                    lps = torch.where(expand_as_right(~padded_values, lps), lps, 0.0)
```
mikaylagawarecki, line 294: should this be 1.0?

vmoens: To me 1 makes more sense, it's the log prob of prob = 1 (you don't do the ratio of log-probs but their difference).
Any value would do, but I don't see why choosing p=2.71 is superior to 1

PR #2859, opened 2025-03-18
[Feature] pass policy-factory in mp data collectors

vmoens:
* (to be filled)


vmoens:
In this PR, I add an example to show how to sync weights when the native weight update class doesn't work (eg serializing mps weights is not possible so you can cast them to cpu then recast to mps)
I also fix the collectors and test the solution.
This should also make other update methods easier to implement.

cc @mikaylagawarecki

vmoens:
We also need to find a way to pass the local update to the sub-collector. With the distributed ones it's easy to do (we can pass any kwarg). For mp ones not so much atm

PR #2865, opened 2025-03-21
v0 param server (using collectives not object store)

mikaylagawarecki:
* __->__ #2865

### DIFF
--- a/param_server_weight_updater.py
+++ b/param_server_weight_updater.py
@@ -1,0 +1,403 @@
+import ray
+
+from argparse import ArgumentParser
+from functools import partial
+
+import torch
+from datasets import load_dataset
+from tensordict import TensorDict
+from torch.utils.data import DataLoader
+from torchrl.collectors.weight_update import RayRemoteWeightUpdater
+from transformers import AutoTokenizer, AutoModel
+from vllm import LLM
+
+from vllm.utils import get_ip, get_open_port
+
+from vllm.worker.worker import Worker
+
+from torchrl.collectors.distributed import RayCollector
+from torchrl.envs import LLMEnv
+from torchrl.modules import from_vllm
+
+parser = ArgumentParser()
+parser.add_argument("--dataset", type=str, default="gsm8k")
+parser.add_argument("--batch_size", type=int, default=4)
+parser.add_argument("--epochs", type=int, default=10)
+parser.add_argument("--repeats", type=int, default=10)
+parser.add_argument("--steps_per_batch", type=int, default=16)
+parser.add_argument("--optim_batch_size", type=int, default=4)
+
+def stateless_init_process_group(
+    master_address: str,
+    master_port: int,
+    rank: int,
+    world_size: int,
+    device: torch.device,
+):
+    """
+    vLLM provides `StatelessProcessGroup` to create a process group
+    without considering the global process group in torch.distributed.
+    It is recommended to create `StatelessProcessGroup`, and then initialize
+    the data-plane communication (NCCL) between external (train processes)
+    and vLLM workers.
+    """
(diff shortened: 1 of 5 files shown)


param_server_weight_updater.py
```diff
@@ -0,0 +1,403 @@
...
+    
+    def _init_model_update_group(self, worker_id):
+        print("in init model update group", worker_id)
+        master_address, master_port = get_ip(), get_open_port()
+        print(master_address, master_port)
+        # FIXME!!!! This needs to be grabbed from each remote collector
+        vllm_tp_size = 1
+        weight_sync_world_size = vllm_tp_size + 1
+        print("calling collective_rpc")
+        self.collector._remote_collectors[worker_id].call_policy_method.remote(
```
mikaylagawarecki, line 269: @vmoens This is one part where I'm trying to call a method on the LLM object to init a process group with the vllm workers

param_server_weight_updater.py
```diff
@@ -0,0 +1,403 @@
...
+    ):
+        if worker_id not in self.param_server_vllm_comm_groups:
+            self._init_model_update_group(worker_id)
+        handles = []
+        for i, (k, v) in enumerate(server_weights.items()):
+            handle = self.collector._remote_collectors[worker_id].call_policy_method.remote(
+                "collective_rpc",
+                ("update_weight",),
+                {'args': (k, v.dtype, v.shape)}
+            )
```
mikaylagawarecki, line 297: @vmoens This is one part where I'm trying to call a method on the LLM object to init a process group with the vllm workers, the second part is below on L293

In this case 
`SyncDataCollector` is remote and has an attribute `.policy`

`.policy` is the ModuleDict object returned by `from_vllm`, and the actual llm instance is in the generate key (the LLM instance is local to the SyncDataCollector)

How can I have a handle to the LLM instance within the SyncDataCollector to call remote methods on it without the hacky `call_policy_method` implementation below?

torchrl/collectors/vllm_weight_update.py
```diff
@@ -0,0 +1,179 @@
...
+        
+    def _get_server_weights(self):
+        return self.state_dict
+    
+    def _maybe_map_weights(self, server_weights):
+        return server_weights
+    
+    def _init_model_update_group(self, worker_id):
+        # here again, I want to grab the tp size from the vLLM worker... :(
+        # llm.llm_engine.parallel_config.tensor_parallel_size
```
mikaylagawarecki, line 159: I keep finding that I want to get info off vllm directly :/, what would you do here?

Should this vLLMRemoteWeightUpdaterBase be aware of all the vllm engines and tp size of each in its __init__?

torchrl/collectors/vllm_weight_update.py
```diff
@@ -0,0 +1,181 @@
+import torch
+import threading
+
+from torchrl.collectors.weight_update import RemoteWeightUpdaterBase
+from torchrl.collectors.weight_update import LocalWeightUpdaterBase
+
+
+VLLM_ERR = None
+try:
```
joecummings, line 9: import `find_spec` might be preferred: https://docs.python.org/3/library/importlib.html#importlib.util.find_spec as it doesn't do the actual import until needed.

torchrl/collectors/vllm_weight_update.py
vmoens: I think you're right, there's no way around it

We need for the main worker to know that stuff about the remote ones

torchrl/collectors/vllm_weight_update.py
vmoens: I agree, that's the proper way to do it

PR #2866, opened 2025-03-22
[Bugfix] Fix VecNorm eps usage

lin-erica:
## Description

In the documentation, `eps` in `VecNorm` is described to be the lower bound of std, but it is done before the `sqrt()` operation, meaning if `eps = 0.0001` then the lower bound of `std` is `0.01`. There is also an additional clamp in the return. I kept the clamp in the `std` so that it can be the same value when using `to_observation_norm()`. 

## Motivation and Context

From reading the docstring, I set the `eps` based on my interpretation, but the actual value was different than I expected.

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -6673,8 +6673,8 @@ class VecNorm(Transform):
             )
 
         mean = _sum / _count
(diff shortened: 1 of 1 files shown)


vmoens:
@lin-erica do we still need this?

gtnv:
hey @lin-erica, is this still on your list? if not, can we close it for now?

PR #2867, opened 2025-03-23
[Feature] VecNormV2

vmoens:
* __->__ #2867


vmoens:
There is a noticeable difference with the first version which is that here we first normalize then update, whereas the other version first updates then normalizes. 
I think it's better to use an unbiased estimate but i'm open to align to the previous version.

List of TODOs:
- [ ] Better coverage of stateless
- [ ] warning about deprec of v1
- [ ] docstrings
- [ ] add to doc
- [ ] test collision between reward and obs specs, esp if 2 vecnorm are stacked 

I guess long-term we should think about collectives for distributed but i'll leave this for enthusiastic young minds

Issue #2872, opened 2025-03-25
[Feature Request] A collector designed for LLMs


vmoens:
## Motivation

We need a [collector](https://pytorch.org/rl/stable/reference/generated/torchrl.collectors.SyncDataCollector.html?highlight=syncdatacollector) that fits well the LLM space. 

We will need to simplify the rollout function greatly - I would rewrite it from scratch.
The LLMEnv and vLLMWrapper can be used to simulate a real-life collector (just create a dummy dataloader that outputs strings and call `LLMEnv.from_dataloader`, see the tests for examples of this).

## Features

We want this collector to have the right arguments: 
- rename "frames" to "steps" (or a more LLM appropriate term)
- We just want a (list of) env constructors and (a list of) policy constructors (Note: currently you can only pass a single policy constructor)

We want to remove the following params:
- "Devices" may not be useful in the LLM space: the policy is unlikely to sit on a single device, and the env should probably handle the device casting on its own (I'm not 100% confident about this one except for the fact that I don't think the current way we handle devices is proper) 
- Remove `create_env_kwargs` as you can always use `functools.partial` to pass kwargs to a specific env constructor
- `init_random_frames` will never be used
- `split_trajs`: not sure that is very useful for LLMs either
- `exploration_type`: no need to explore - the LLM engine has its own exploration logic
- `interruptor`: I would deactivate that in the v0 of the LLM collector
- `use_buffers` should always be set to `False` since we're going to pass non-tensor data
- `trust_policy` and `compile_policy`: I believe the policy constructor should take care of that.
- `no_cuda_sync` will also probably not be useful

Additional options and features (stretch goals):
- We want this buffer to work with ray like we do with the RayCollector. Since the RayCollector also has too many params, we should find a way of dealing with that (one way could be to do this:
  ```python
  class _DCMeta(...):
       def __call__(cls, *args, **kwargs):
           backend = kwargs.pop("backend")
           if backend:
               return RayCollector(**my_cuisine())
           return super().__call__(*args, **kwargs)

  class LLMCollector(DataCollector, metaclass=_DCMeta):
       def __init__(self, ..., backend: Literal["ray"] | None=None):
           ...
   ```
   that way people can just ask for Ray (or not!) and pass the ray kwargs and tada! the collector uses ray (or not!)
[...]

PR #2879, opened 2025-03-26
[Feature] LLM collector

Lucaskabela:
## Description

Add LLM Collector, which is a subset of the SyncDatacollector aimed at trimming to core funcionality in an easy to consume API for LLM devs

## Motivation and Context

#2872 

## Types of changes

## Checklist

## Testing
```
python test/test_collector.py TestLLMCollector
```

### DIFF
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -57,6 +57,7 @@ from torchrl.collectors.utils import split_trajectories
 from torchrl.data import (
     Composite,
     LazyMemmapStorage,
+    LazyStackStorage,
     LazyTensorStorage,
     NonTensor,
     ReplayBuffer,
@@ -3590,33 +3591,22 @@ class TestLLMCollector:
 
         return model, tokenizer
 
-    def test_llm_collector_with_vllm(self, vllm_instance):
+    @pytest.mark.slow
+    @pytest.mark.parametrize("rb", [True, False])
+    @pytest.mark.parametrize("total_steps", [1, 10, 20])
+    def test_llm_collector_with_vllm(self, rb, total_steps, vllm_instance):
         # NOTE: if VLLM fails with CUDA multiprocessing, try setting
         # `export VLLM_WORKER_MULTIPROC_METHOD=spawn`
         policy = vLLMWrapper(vllm_instance)
         tokenizer = vllm_instance.get_tokenizer()
-        bsz = 2
-        dataloader = DummyStrDataLoader(bsz)
+        self._run_collector_test(total_steps, rb, policy, tokenizer)
 
-        env = LLMEnv.from_dataloader(
-            dataloader=dataloader,
-            tokenizer=tokenizer,
-            str2str=True,
-            batch_size=bsz,
-            group_repeats=True,
-        )
-
-        collector = LLMCollector(
-            env=env,
-            policy_factory=lambda: policy,
-            steps_per_batch=env.batch_size[0],
-            total_steps=3,
-        )
-
-        for data in collector:
-            assert data is not None
-
-    def test_llm_collector_with_transformers(self, transformers_instance):
+    @pytest.mark.slow
+    @pytest.mark.parametrize("rb", [True, False])
+    @pytest.mark.parametrize("total_steps", [1, 10, 20])
+    def test_llm_collector_with_transformers(
+        self, rb, total_steps, transformers_instance
+    ):
         model, tokenizer = transformers_instance
         policy = TransformersWrapper(
             model,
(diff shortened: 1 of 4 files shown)


torchrl/collectors/llm_collector.py
```diff
@@ -0,0 +1,134 @@
...
+            data = self._shuttle
+
+        print('data type', type(data))
+        print('data', data)
+        tensordicts = []
+        collected_frames = 0
+        while collected_frames < self.frames_per_batch:
+            policy_input = data
+            print('policy_input type', type(policy_input))
+            env_input = self.policy(policy_input)
```
Lucaskabela, line 83: seems this data causes some exception in serialization format when passed to the LLM - wondering if we need some additional parameters to the constructor for the policy? (cc @vmoens)

vmoens: I'm not seeing any error on cpu (at least for 20 iterations )

Lucaskabela: Interesting; assuming you are running the testplan as is @vmoens? I get the crash listed on the very first iteration

What version of vllm is in your environment out of curiosity?

Lucaskabela:
Originally, this code ran into the error:
```
  File "/home/lucaskabela/.conda/envs/cabernet/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/home/lucaskabela/.conda/envs/cabernet/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/home/lucaskabela/.conda/envs/cabernet/lib/python3.10/site-packages/vllm/v1/engine/core.py", line 429, in process_input_socket
    request = decoder.decode(data_frame.buffer)
  File "/home/lucaskabela/.conda/envs/cabernet/lib/python3.10/site-packages/vllm/v1/serial_utils.py", line 34, in decode
    return self.decoder.decode(obj)
msgspec.ValidationError: Expected `int | null`, got `bool` - at `$[6].logprobs`
```

This seems to have come from a versioning issue; downgrading torch and vllm with
```
pip install --force torch==2.5.1 vllm==0.7.3
```
Then
```
python setup.py clean && python setup.py develop
```
Did the trick and fixed this error :) Previously, I had `torch==2.6.0` and `vllm==0.8.2`, so note these versions may have cause this error

vmoens:
Amazing!
I did a bit of refactoring.
Things that are missing:
- Tests in test/test_collectors.py. These should include the vLLMWrapper and the TransformersWrapper
- docstrings + register in the `docs` directory
- ultimately remove the `if __name__ == "__main__":` from the file

Lucaskabela:
> Amazing! I did a bit of refactoring. Things that are missing:
> 
> * Tests in test/test_collectors.py. These should include the vLLMWrapper and the TransformersWrapper
> * docstrings + register in the `docs` directory
> * ultimately remove the `if __name__ == "__main__":` from the file
> * use `trust_policy=True` as not doing so can trigger checks that we want to avoid


Thanks for the comments here; I think I incorporated the majority of this feedback and updated the test on my end; the only thing I am not sure about is how to register in the `docs` directory - could you provide more details on this?

test/test_collector.py
```diff
@@ -3544,6 +3559,94 @@ def test_weight_update(self):
...
+
+        collector = LLMCollector(
+            env=env,
+            policy_factory=lambda: policy,
+            steps_per_batch=env.batch_size[0],
+            total_steps=3,
+        )
+
+        for data in collector:
+            assert data is not None
```
vmoens, line 3617: maybe let's check that the fields make sense?
ie that we have all the set of tokens / text etc that we want?

test/test_collector.py
```diff
@@ -3544,6 +3559,94 @@ def test_weight_update(self):
...
```
vmoens, line 3647: ditto

torchrl/collectors/llm_collector.py
```diff
@@ -0,0 +1,78 @@
...
+from torchrl.collectors import (
+    LocalWeightUpdaterBase,
+    RemoteWeightUpdaterBase,
+    SyncDataCollector,
+)
+from torchrl.data import ReplayBuffer
+
+
+class LLMCollector(SyncDataCollector):
+    """A simplified version of SyncDataCollector for LLM inference."""
```
vmoens, line 17: missing kwarg description + running example (feel free to pick the one in the tests)

torchrl/collectors/llm_collector.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __init__(
+        self,
+        env,
+        policy_factory,
+        *,
+        steps_per_batch: int = -1,
```
vmoens, line 24:
```suggestion
        steps_per_batch: int,
```
Right now this is a value that we require. Long-term, when we'll go full async, we'll allow people to not specify anything because the output data will be written the buffer as soon as the trajectory is completed.

test/test_collector.py
vmoens: Another thing to check is that the data has the right number of elements and that the total number of steps matches what you set in the constructor. 
We also want to test that we can populate a replay buffer:
```
rb = ReplayBuffer(storage=LazyStackStorage(size))
```
and that the samples "make sense" (eg, they have the collector metadata under the "collector" key)

vmoens (approved):
LGTM thanks!

PR #2891, opened 2025-04-04
[Feature] More options for LLM collectors

vmoens:
* __->__ #2891

### DIFF
--- a/test/test_collector.py
+++ b/test/test_collector.py
@@ -5,10 +5,10 @@
 from __future__ import annotations
 
 import argparse
+import asyncio
 import contextlib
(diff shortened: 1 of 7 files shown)


vmoens:
This PR:
- introduces options to run collectors async using `collector.start()`
- Permits the writing/yielding of finished (done) trajectories only, and subsequently only the last steps of these
- fixes a couple of bugs in str concatenation with LLMEnv when using multiturn
- Adds an eod_token_id option to LLMEnv to write the done state when the token is caught
cc @mikaylagawarecki @Darktex @felipemello1

PR #2898, opened 2025-04-09
[Feature] Gymnasium 1.1 compatibility

vmoens:
* __->__ #2898

### DIFF
--- a/.github/unittest/linux/scripts/run_all.sh
+++ b/.github/unittest/linux/scripts/run_all.sh
@@ -97,9 +97,9 @@ echo "installing gymnasium"
 if [[ "$PYTHON_VERSION" == "3.12" ]]; then
   pip3 install ale-py
   pip3 install sympy
-  pip3 install "gymnasium[accept-rom-license,mujoco]<1.0" mo-gymnasium[mujoco]
+  pip3 install "gymnasium[accept-rom-license,mujoco]>=1.1" mo-gymnasium[mujoco]
 else
-  pip3 install "gymnasium[atari,accept-rom-license,mujoco]<1.0" mo-gymnasium[mujoco]
+  pip3 install "gymnasium[atari,accept-rom-license,mujoco]>=1.1" mo-gymnasium[mujoco]
 fi
 pip3 install "mujoco" -U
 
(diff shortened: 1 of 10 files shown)


JonDum:
Hey @vmoens thanks for getting this out. After a lot of struggle, I was finally able to test this out  on yesterday's nightly build using a reference from the sota/decision_transformer example and I'm running into an error that has me stuck thinking it's part of this transition to gymnasium 1.1. 

I'm largely just adapting this:

https://github.com/pytorch/rl/blob/c3310b87f09dc6bc69167e72528f06c7dab111c9/sota-implementations/decision_transformer/utils.py#L64-L81

Docs are sparse on custom environments but I was able to piece together through other examples that I needed to set the backed to gymnasium then register the env with EnvBase like so:

```
def make_base_env(env_cfg):
    set_gym_backend(env_cfg.backend).set()
    EnvBase.register_gym(
        "MyCustomEnv"
        entry_point=MyCustomEnv
        backend="gymnasium",
    )
    env = GymEnv(env_cfg.name)
    return env
```
where MyCustomEnv is
```
import gymnasium as gym import numpy as np
from gymnasium.spaces import Box

class MyCustomEnv(gym.Env):
    metadata = {
			"render_modes": [],
			"autoreset_mode": gym.vector.AutoresetMode.DISABLED,
    }

    def __init__(self, config: dict):
        self.observation_space = Box(-1.0, 1.0, (5,), dtype=np.float32)
        self.action_space = gym.spaces.Discrete(3)

    def reset(self, seed=None, options=None, **kwargs):
        super().reset(seed=seed, options=options, **kwargs)
        info = {}
				...
        return observation, info

    def step(self, action: np.int32):
        self.num_steps += 1
        reward = 1
        return (observation, reward, terminated, truncated, info)

```

I get the error:
```
TypeError: register() got an unexpected keyword argument 'autoreset'
```

```
 .venv/lib/python3.12/site-packages/torchrl/envs/common.py:2601  
 in _register_gym                                                                                 
                                                                                                  
   2598 │   │   │   transform=transform,                                                          
   2599 │   │   │   **kwargs,                                                                     
   2600 │   │   )                                                                                 
 ❱ 2601 │   │   return gymnasium.register(                                                        
   2602 │   │   │   id=id,                                                                        
   2603 │   │   │   entry_point=entry_point,                                                      
   2604 │   │   │   reward_threshold=reward_threshold,
```

PR #2903, opened 2025-04-15
[BE] `_set_seed` returns `None` + type annotations

antoinebrl:
## Description

`_set_seed` is not expected to return anything as the sequence of seeds is handled by the parent function `set_seed`. The type annotation should make it more explicit that no return value is expected. 

## Motivation and Context

I went into a rabbit hole to figure out how to handle the seeds when dealing with a pool of environments. Certain implementation of `_set_seed` were returning `seed+1` for example which was confusing. I believe the type annotation `-> None` makes it clear.


vmoens (approved):
LGTM thanks

PR #2913, opened 2025-04-22
[BugFix] Base transform applies `_call` on reset

louisfaury:
## Description

## Motivation and Context

In contrast with what the docstring of  `_call` claims, the base `Transform` does not apply in `_call` method on resets. 
This also means that every daughter class must overwrite it to produce the expected results.  

Maybe there is a good reason for this; I'll be happy to improve the doc if its the case! If not, I'll wait for your approval to fix the test this would potentially break :) 

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/transforms/transforms.py
+++ b/torchrl/envs/transforms/transforms.py
@@ -308,7 +308,7 @@ class Transform(nn.Module):
         self, tensordict: TensorDictBase, tensordict_reset: TensorDictBase
     ) -> TensorDictBase:
         """Resets a transform if it is stateful."""
-        return tensordict_reset
+        return self._call(tensordict_reset)
 
     def _reset_env_preprocess(self, tensordict: TensorDictBase) -> TensorDictBase:
         """Inverts the input to :meth:`TransformedEnv._reset`, if needed."""


vmoens (commented):
If you run the tests under this PR not even the first passes haha
I don't know why we say we do it, the reason we don't do it is that you may not have all the data at reset (typically there is no reward).
Here are some additional pieces of info:
- a `_set_missing_tolerance` context manager can be used to tell `_call` that it's ok if a key is missing. This is what you have in `UnaryTransform` for example;
- we know have the capability to transform the data before it's passed to `base_env.reset` through `_reset_env_preprocess`. I think this is the one the docstrings [refers to](https://github.com/pytorch/rl/blob/main/torchrl/envs/transforms/transforms.py#L313-L313).


Sorry for the misleading docstring! We can try to make it clearer, or use 
```python
    def _reset(
        self, tensordict: TensorDictBase, tensordict_reset: TensorDictBase
    ) -> TensorDictBase:
        """Resets a transform if it is stateful."""
        if not self.enable_inv_on_reset:
            with _set_missing_tolerance(self, True):
                tensordict_reset = self._inv_call(tensordict_reset)
        return tensordict_reset

    def _reset_env_preprocess(self, tensordict: TensorDictBase) -> TensorDictBase:
        """Inverts the input to :meth:`TransformedEnv._reset`, if needed."""
        if self.enable_inv_on_reset and tensordict is not None:
            with _set_missing_tolerance(self, True):
                tensordict = self._inv_call(tensordict)
        return tensordict
```
but one should keep in mind that any transform that overrides `_call` may fail under these changes (eg if users don't accommodate for the `_set_missing_tolerance` functionality)

louisfaury:
Thanks for the answer! 
It makes sense; yet I would still argue that the current design is a bit counter-intuitive. A transform that by default applies `_call` in its reset  (potentially under the `_set_missing_tolerance` context)* feels imo more instinctive. I believe this is why many transforms implement this mechanism instead of inheriting it. I know I had a few silent bugs because of this, and I have reimplemented the same `reset()` in several of my transforms.
I don't have a direct solution yet, so I will modified the MR to just correct the docstring for now! 

*: to me not using the `_set_missing_tolerance` context should be fine; the absence of the required fields at reset time does not necessarily need to be supported by the base transform. If some fields are required, inheriting the base method that do not ressort to `_set_missing_tolerance` will yield a clear error.

vmoens:
I agree with all your points!

The transition may be painful for some users though, as we'd be silently adding a call to the transform.

We can think of a patch, I just want to make sure any change of behavior is well documented and announced!

gtnv:
hi @louisfaury, do you plan to continue working on this? if not, should we close it?

PR #2914, opened 2025-04-23
[Refactor] Refactor the weight update logic

vmoens:
* #2926
* __->__ #2914


vmoens (commented):
I'm trying to rethink about sender and receiver one last time.

I think we always need a sender: in some way, you always need to push the weights somewhere (because vllm will never ask for weights, you push the weights to vllm).

In centralized settings, where you have a central collector orchestrating satellites ones, the responsibility of the central collector is to push weights to the workers (note that this is not the schema that we are using, which is decentralized).

The receiver on the other hand is accessory, it's more like the kind of settings where your worker can ask for weights by itself at a given interval or when some conditions are met. 

The `update_policy_weights_` function then looks like
```
def update_policy_weights_(self, *args, **kwargs):
    weights = self.receive(*args, **kwargs) #  this is a no-op if the weights (hanlde) are in the args
    self.send(weights)  # this should never be a no-op, as this is where the weight update actually occurs
```

@mikaylagawarecki @Darktex

docs/source/reference/collectors.rst
```diff
@@ -118,27 +118,44 @@ try to limit the cases where a deepcopy will be executed. The following chart sh
...
 
 The weight synchronization process is facilitated by two main components: :class:`~torchrl.collectors.WeightUpdateReceiverBase`
 and :class:`~torchrl.collectors.WeightUpdateSenderBase`. These base classes provide a structured interface for
 implementing custom weight update logic, allowing users to tailor the synchronization process to their specific needs.
 
-- :class:`~torchrl.collectors.WeightUpdateReceiverBase`: This component is responsible for updating the policy weights on
-  the local inference worker. It is particularly useful when the training and inference occur on the same machine but on
+- :class:`~torchrl.collectors.WeightUpdateSenderBase`: This component handles the distribution of policy weights to
+  the policy or to remote inference workers. Every collector -- server or worker -- should have a `WeightUpdateSenderBase`
+  instance to handle the "push" operation of the weights to the policy.
```
mikaylagawarecki, line 136: I think "push/pull" and "sender/receiver" are confusing 🫤 In particular, for me the Receiver == "Puller" part is tough to wrap my head around.

Pull architecture is request driven: the client sends the request, and the server responds accordingly
Push architecture is event driven: the server pushes data to clients as updates become available

The confusion for me is that I think of sender --> receiver as "sender actively pushes, receiver passively receives".  Hence receiver == puller is not intuitive

docs/source/reference/collectors.rst
```diff
@@ -118,27 +118,44 @@ try to limit the cases where a deepcopy will be executed. The following chart sh
    Policy copy decision tree in Collectors.
 
 Weight Synchronization in Distributed Environments
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+--------------------------------------------------
```
mikaylagawarecki, line 121: I read the diagram above as 

1. `CollectorServer`: main thread of `RayCollector`
2. `Collector Worker {i}`, remote `DataCollector`

If this read is correct, in my mind, it might sometimes make sense to have the receiver on the collector worker rather than the collector server
e.g. If the number of remote workers is sufficiently high, the collector worker might not be colocated with the collector server, in that case it would not make sense to pass the weights "two hops" to get to the worker


Separate qn -- from the diagram it looks like the collector server chooses when to pull from the param server and then "forcefully pushes" to all the workers at once. Is this design intentional? (e.g. Is the purpose of this to batch up workers to different collector servers and update them in batches?)

PR #2917, opened 2025-04-23
[Deprecation] Enact deprecations

vmoens:
* #2921
* #2915
* #2920
* #2919
* #2918
* __->__ #2917
* #2916


vmoens:
cc @matteobettini This may affect you:
As announced, for consistency across all envs that have nested reward / actions we now enforce that `env.action_spec` returns the full action spec if the action key is nested. To get the leaf, you need to do `env.full_action_spec[env.action_key]`.

PR #2928, opened 2025-04-29
[BugFix] SMACv2 action spec

matteobettini:
Fix bug introduced by https://github.com/pytorch/rl/pull/2917


vmoens:
thanks
already accounted for by #2923

PR #2940, opened 2025-05-05
[BugFix] Multiagent PPO tutorial

matteobettini:
Another leftover bug from https://github.com/pytorch/rl/pull/2917

2 questions:
- do we still have a CI for tutos?
- Will this update the docs when merged or we are stuck w the bug until next release?


vmoens (approved):
LGTM thanks for this!
Sorry for missing it, I had to deactive the builds of the tutos a while ago bc the script was hanging forever for no reason :/

PR #2941, opened 2025-05-07
[BugFix] Fix shifted value computation with an LSTM

vmoens:
* __->__ #2941

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -13797,8 +13797,6 @@ class TestValues:
         else:
             raise NotImplementedError
         recurrent_module.eval()
-        for p in recurrent_module.parameters():
-            p.data *= 1 + torch.randn_like(p.data) / 10
         mlp_value = MLP(num_cells=[64], out_features=1)
         value_net = Seq(
             recurrent_module,
@@ -13809,7 +13807,6 @@ class TestValues:
             recurrent_module,
             Mod(mlp_policy, in_keys=["intermediate"], out_keys=["action"]),
         )
-        # value_net.select_out_keys("state_value")
         env = env.append_transform(recurrent_module.make_tensordict_primer())
         vals = env.rollout(1000, policy_net, break_when_any_done=False)
         value_net(vals.copy())
@@ -13825,8 +13822,6 @@ class TestValues:
             r0 = gae_shifted(vals.copy())
         a0 = r0["advantage"]
 
-        # TODO: where to put this?
-        vals["next", "is_init"] = vals["is_init"]
         gae = GAE(
             gamma=0.9,
             lmbda=0.99,
--- a/torchrl/objectives/utils.py
+++ b/torchrl/objectives/utils.py
@@ -557,17 +557,15 @@ def _pseudo_vmap(
     func: Callable,
     in_dims: Any = 0,
     out_dims: Any = 0,
-    randomness: str = "error",
+    randomness: str | None = None,
     *,
     chunk_size=None,
 ):
-    # Test:
-    # import torch
-    # from tensordict import TensorDict
-    # a = torch.randn(3, 4)
-    # b = TensorDict(a=torch.randn(3, 4), batch_size=3)
-    # ds = (0, 0)
-    # v0, v1, v2 = zip(*tuple(tree_map(lambda d, x: x.unbind(d), (0, 0), (a, b))))
+    if randomness is not None and randomness not in ("different", "error"):
+        raise ValueError(
+            f"pseudo_vmap only supports 'different' or 'error' randomness modes, but got {randomness=}. If another mode is required, please "
+            "submit an issue in TorchRL."
+        )
     if isinstance(in_dims, int):
         in_dims = (in_dims,)
     if isinstance(out_dims, int):
--- a/torchrl/objectives/value/advantages.py
+++ b/torchrl/objectives/value/advantages.py
@@ -441,12 +441,14 @@ class ValueEstimatorBase(TensorDictModuleBase):
             value_net = self.value_network
         in_keys = value_net.in_keys
         if single_call:
-            # Reshape to -1 because we cannot guarantee that all dims have the same number of done states
-            for i, name in enumerate(data.names):
-                if name == "time":
-                    ndim = i + 1
-                    break
-            else:
+            # We are going to flatten the data, then interleave the last observation of each trajectory in between its
+            #  previous obs (from the root TD) and the first of the next trajectory. Eventually, each trajectory will
+            #  have T+1 elements (or, for a batch of N trajectories, we will have \Sum_{t=0}^{T-1} length_t + T
+            #  elements). Then, we can feed that to our RNN which will understand which trajectory is which, pad the data
+            #  accordingly and process each of them independently.
+            try:
+                ndim = list(data.names).index("time") + 1
+            except ValueError:
                 if RL_WARNINGS:
                     warnings.warn(
                         "Got a tensordict without a time-marked dimension, assuming time is along the last dimension. "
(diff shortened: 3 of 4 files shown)


torchrl/objectives/utils.py
```diff
@@ -550,6 +553,58 @@ def decorated_module(*module_args_params):
...
+    *,
+    chunk_size=None,
+):
+    # Test:
+    # import torch
+    # from tensordict import TensorDict
+    # a = torch.randn(3, 4)
+    # b = TensorDict(a=torch.randn(3, 4), batch_size=3)
+    # ds = (0, 0)
+    # v0, v1, v2 = zip(*tuple(tree_map(lambda d, x: x.unbind(d), (0, 0), (a, b))))
```
louisfaury, line 570: Should this be removed before merging?

torchrl/objectives/value/advantages.py
```diff
@@ -435,62 +441,70 @@ def _call_value_nets(
             value_net = self.value_network
         in_keys = value_net.in_keys
         if single_call:
+            # Reshape to -1 because we cannot guarantee that all dims have the same number of done states
             for i, name in enumerate(data.names):
                 if name == "time":
                     ndim = i + 1
                     break
```
louisfaury, line 448: Nit: Should we instead rely on `ndim = data.names.index("time") + 1` with a try/catch instead? Would be cleaner imo.

torchrl/objectives/value/advantages.py
```diff
@@ -435,62 +441,70 @@ def _call_value_nets(
             value_net = self.value_network
         in_keys = value_net.in_keys
         if single_call:
+            # Reshape to -1 because we cannot guarantee that all dims have the same number of done states
```
louisfaury, line 444: I am not sure I follow why this comment is here.

torchrl/objectives/value/advantages.py
```diff
@@ -435,62 +441,70 @@ def _call_value_nets(
...
                         "Got a tensordict without a time-marked dimension, assuming time is along the last dimension. "
                         "This warning can be turned off by setting the environment variable RL_WARNINGS to False."
                     )
                 ndim = data.ndim
-                idx = (slice(None),) * (ndim - 1) + (slice(None, data.shape[ndim - 1]),)
-                idx_ = (slice(None),) * (ndim - 1) + (
-                    slice(data.shape[ndim - 1], None),
+            data_copy = data.copy()
+            done = data_copy["next", "done"].clone()
+            done[(slice(None),) * (ndim - 1) + (-1,)].fill_(True)
```
louisfaury, line 458:
```suggestion
            # Mark the last step of every sequence as done.
            done[(slice(None),) * (ndim - 1) + (-1,)].fill_(True)
```

torchrl/objectives/value/advantages.py
```diff
@@ -435,62 +441,70 @@ def _call_value_nets(
...
                     warnings.warn(
                         "Got a tensordict without a time-marked dimension, assuming time is along the last dimension. "
                         "This warning can be turned off by setting the environment variable RL_WARNINGS to False."
                     )
                 ndim = data.ndim
-                idx = (slice(None),) * (ndim - 1) + (slice(None, data.shape[ndim - 1]),)
-                idx_ = (slice(None),) * (ndim - 1) + (
-                    slice(data.shape[ndim - 1], None),
+            data_copy = data.copy()
+            done = data_copy["next", "done"].clone()
```
louisfaury, line 457: I'll admit having quite a hard time following the logic that lays below, but more importantly _why_ it is needed. I think a few comments could really help the next reader :)

test/test_cost.py
```diff
@@ -13755,6 +13761,84 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
...
+                in_keys=["observation", "rs_h"],
+                out_keys=["intermediate", ("next", "rs_h")],
+                python_based=True,
+                dropout=0,
+            )
+        else:
+            raise NotImplementedError
+        recurrent_module.eval()
+        for p in recurrent_module.parameters():
+            p.data *= 1 + torch.randn_like(p.data) / 10
```
thomasbbrunner, line 13801: Curious as to why overwriting the RNN parameters?

test/test_cost.py
```diff
@@ -13755,6 +13761,84 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
...
+        value_net = Seq(
+            recurrent_module,
+            Mod(mlp_value, in_keys=["intermediate"], out_keys=["state_value"]),
+        )
+        mlp_policy = MLP(num_cells=[64], out_features=1)
+        policy_net = Seq(
+            recurrent_module,
+            Mod(mlp_policy, in_keys=["intermediate"], out_keys=["action"]),
+        )
+        # value_net.select_out_keys("state_value")
```
thomasbbrunner, line 13812:
```suggestion
```

test/test_cost.py
```diff
@@ -13755,6 +13761,84 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
...
+        env = env.append_transform(recurrent_module.make_tensordict_primer())
+        vals = env.rollout(1000, policy_net, break_when_any_done=False)
+        value_net(vals.copy())
```
thomasbbrunner, line 13815: I think this call to the value net is not necessary?

```suggestion
```

test/test_cost.py
```diff
@@ -13755,6 +13761,84 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
...
+            lmbda=0.99,
+            value_network=value_net,
+            shifted=True,
+        )
+        with set_recurrent_mode(True):
+            r0 = gae_shifted(vals.copy())
+        a0 = r0["advantage"]
+
+        # TODO: where to put this?
+        vals["next", "is_init"] = vals["is_init"]
```
thomasbbrunner, line 13829: Should we solve this TODO before merging?

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -782,8 +788,8 @@ def _lstm(
             )
 
         # we only need the first hidden state
-        _hidden0_in = hidden0_in[:, 0]
-        _hidden1_in = hidden1_in[:, 0]
+        _hidden0_in = hidden0_in[..., 0, :, :]
+        _hidden1_in = hidden1_in[..., 0, :, :]
```
thomasbbrunner, line 792: This change is interesting. I'm guessing the shape of `hiddenN_in` is `(batch, sequence, num_lstm_layers, hidden_size)`.

Were we before not using the first hidden state of a sequence, but the first hidden layer in the LSTM?

test/test_cost.py
vmoens: it's there for init purposes

test/test_cost.py
vmoens: yes! I think this is something we can safely ask GAE to do for us

torchrl/objectives/value/advantages.py
vmoens: should be moved down

test/test_cost.py
vmoens: I did that to make sure we had enough power to detect slight changes in behavior (ie that every item that could be numerically different than the same computed with the opposite shifted param would be), but in practice it's less useful now

torchrl/objectives/value/advantages.py
vmoens: Fair point. The reason I wasn't not doing it is that the try/except isn't compatible with compile but in practice, once we reach the warning we're in a non-compilable code anyway.

PR #2972, opened 2025-05-22
[Feature] Adds per-head entropy coefficients to PPOLoss

felixsittenauer:
## Description

Adds **per-head entropy coefficients** to `PPOLoss`.

* `entropy_coef` now accepts  
  * **scalar** (legacy behaviour) or  
  * **`Mapping[str, float]`** — one coefficient per action head.
* New helper `_weighted_entropy` applies the correct weighting:
  * scalar path: `-coef * entropy`
  * mapping path: `-Σ coef_head * entropy_head`
* `forward()` switches from direct multiplication to `_weighted_entropy`.
* Keeps `self.entropy_coef` buffer for backward-compat (dummy `0.0` when mapping used).
* Docstrings updated.

## Motivation and Context

Composite / multi-head policies often need different exploration pressure on
each sub-action (e.g. steering vs. throttle).  
This change lets users express that directly via a dict instead of patching the
loss externally.

## Types of changes

## Checklist

### DIFF
--- a/test/test_cost.py
+++ b/test/test_cost.py
@@ -112,6 +112,7 @@ from torchrl.objectives.deprecated import DoubleREDQLoss_deprecated, REDQLoss_de
 from torchrl.objectives.redq import REDQLoss
 from torchrl.objectives.reinforce import ReinforceLoss
 from torchrl.objectives.utils import (
+    _sum_td_features,
     _vmap_func,
     HardUpdate,
     hold_out_net,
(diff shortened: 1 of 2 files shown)


vmoens (commented):
Thanks for this!
I would:
- edit the docstrings
- add a test in `test/test_costs.py`

Otherwise LGTM!

felixsittenauer:
@vmoens Added some tests and more details to the docstrings. Tried to match the current style of the tests, so please let me know if I should rather chop them in smaller chunks.

Also, is there a good E2E sanity check to make sure I didn't mess up the math?

vmoens (approved):
I took the liberty of editing `_weighted_loss_entropy` such that `NestedKey`s could be used (not only strings) in case you have `("group0", "agent0")`

The logic is that the coeffs will be
```python
coefs = TensorDict({
   ("group0", "agent0"): 0.1,
   ("group1", "agent0"): 0.15,
   ("group1", "agent1"): 2,
})
```

I also registered the coefs as buffers using `TensorDictParams` - maybe it's an overkill?

PR #2995, opened 2025-06-09
[CI] Fix gym rendering issues

vmoens:


gtnv:
hi @vmoens, is this implementation still active? if not, can we close it for now?

Issue #1755, opened 2023-12-20
[Feature Request] Add A3C


tushdon2:
## Motivation

Add A3C implementation as presented [here](https://arxiv.org/abs/1602.01783) so that the experiments of [Rainbow](https://arxiv.org/pdf/1710.02298.pdf) can be added to the examples in further contributions.

## Solution

Add A3C code like [A2C](torchrl/objectives/a2c.py) and add its paper reproduction as an example.

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

tushdon2:
I am interested in working on this issue along with @shivank21.

vmoens:
That would be wonderful!

shivank21:
In A3C algorithm we are required to asynchronize the variables across multiple CPU threads. For that we were planning to use the [threading](https://docs.python.org/3.7/library/threading.html)  library of python. Does that work here?
Also it refers in the Documentation that "This module used to be optional, it is now always available."
What does this really mean?

albertbou92:
Having A3C would be very cool!

tushdon2:
Yes.
I am currently working on it.
If you wish we can collaborate.

PaulBeuran:
That would be great thanks. Do you have an open branch I can start looking ?

tushdon2:
Nope.
I was going through the background, understanding theory and how the repo works.
We can connect to discuss if you want.

PaulBeuran:
No problem, doing the same right now.
DM me on Discord @pbeuran_71003

tushdon2:
Sent you a request.

PR #3001, opened 2025-06-15
[Feature, Example] A3C Atari Implementation for TorchRL

simeetnayan81:
## Description

This PR adds an implementation of the Asynchronous Advantage Actor-Critic (A3C) algorithm for Atari environments in the [torchrl/sota-implementations](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) directory. The main files added are:

a3c_atari.py: Contains the A3C worker class, shared optimizer, and main training loop using multiprocessing.
utils_atari.py: Provides utility functions for environment creation, model construction, and evaluation, adapted for Atari tasks.
config_atari.yaml: Configuration file for hyperparameters, environment settings, and logging.

The implementation leverages TorchRL's collectors, objectives, and logging utilities, and is designed to be modular and extensible for research and benchmarking. Some of the utils functions are also borrowed from a2c_atari.

## Motivation and Context

This change is required to provide a strong, reproducible baseline for A3C on Atari environments using TorchRL. It enables researchers and practitioners to benchmark and compare reinforcement learning algorithms within the TorchRL ecosystem. The implementation follows best practices for distributed RL and is compatible with TorchRL's API.

This PR solves the issue: #1755

## Types of changes

## Checklist


vmoens (commented):
This all looks pretty good!
Could you share a (couple of) learning curve?
Another thing to do before landing is to add it to the sota-implementations CI run:
https://github.com/pytorch/rl/blob/main/.github/unittest/linux_sota/scripts/test_sota.py
Make sure the config passed there is as much barebone as we can - we just want to run the script for a couple of collection / optim iters and make sure it runs without error (not that it properly trains).
We also need to add it to the [sota-check](https://github.com/pytorch/rl/tree/main/sota-check) runs

simeetnayan81:
Thanks @vmoens . I'll add the required changes as well as some training curves.

simeetnayan81:
@vmoens, I have added the required scripts as well.

Not getting enough resources and time for hyperparam tuning to generate a proper training curve.

sota-implementations/a3c/a3c_atari.py
```diff
@@ -0,0 +1,230 @@
...
+        logger = get_logger(
+            cfg.logger.backend,
+            logger_name="a3c",
+            experiment_name=exp_name,
+            wandb_kwargs={
+                "config": dict(cfg),
+                "project": cfg.logger.project_name,
+                "group": cfg.logger.group_name,
+            },
+        )
```
vmoens: What I usually see is that the logger is only passed to the first worker. 
Another thing is that you may want to assume that the logger isn't serializable and should be instantiated locally within the worker.

vmoens (commented):
LGTM, just a minor comment on the logger!

sota-implementations/a3c/a3c_atari.py
```diff
@@ -0,0 +1,230 @@
...
+        cfg.env.env_name, device=cfg.loss.device, gym_backend=cfg.env.backend
+    )
+    global_model = nn.ModuleList([global_actor, global_critic_head])
+    global_model.share_memory()
+    optimizer = SharedAdam(global_model.parameters(), lr=cfg.optim.lr)
+
+    num_workers = cfg.multiprocessing.num_workers
+
+    if num_workers is None:
+        num_workers = mp.cpu_count()
```
vmoens: we should have way fewer workers - I think we need users to tell us how many.

sota-implementations/a3c/a3c_atari.py
```diff
@@ -0,0 +1,230 @@
...
+                metrics_to_log[
+                    "train/episode_length"
+                ] = episode_length.sum().item() / len(episode_length)
+
+            with torch.no_grad():
+                data = self.adv_module(data)
+            data_reshape = data.reshape(-1)
+            losses = []
+
+            mini_batches = data_reshape.split(self.mini_batch_size)
```
vmoens, line 198: To shuffle things a bit I usually rely on a replay buffer instance rather than just splitting the data

sota-implementations/a3c/a3c_atari.py
```diff
@@ -0,0 +1,230 @@
...
+            global_param._grad = local_param.grad
+
+        for local_param, global_param in zip(
+            self.local_critic.parameters(), self.global_critic.parameters()
+        ):
+            global_param._grad = local_param.grad
+
+        gn = torch.nn.utils.clip_grad_norm_(
+            self.loss_module.parameters(), max_norm=max_grad_norm
+        )
```
vmoens, line 134: can you explain what we do here? What do we use the _grad for?

sota-implementations/a3c/a3c_atari.py
```diff
@@ -0,0 +1,230 @@
...
+from torchrl.objectives.value.advantages import GAE
+
+from torchrl.record.loggers import generate_exp_name, get_logger
+from utils_atari import make_parallel_env, make_ppo_models
+
+
+torch.set_float32_matmul_precision("high")
+
+
+class SharedAdam(torch.optim.Adam):
```
vmoens: shouldn't we move this to the utils file?

vmoens (commented):
I made a few edits.
Can you explain the way the params are shared and updated? I'm not sure I see the logic

sota-implementations/a3c/a3c_atari.py
simeetnayan81: _grad is used to store the gradients for each parameter.
We copy local gradients to the global model so the global model can be updated with the optimizer.
This is a key step in A3C, where multiple workers asynchronously update a shared global model.

sota-implementations/a3c/a3c_atari.py
simeetnayan81: Sure, will do it

sota-implementations/a3c/a3c_atari.py
simeetnayan81: That can be configured in the config_atari. You want me to explicitly set it to some constant here?

sota-implementations/a3c/a3c_atari.py
simeetnayan81: Oh yea, I did that because I thought logging any single worker should be a good representative of the global model since anyway the weights are being copied. Logging all the worker might not be really useful but that can be done as well.

simeetnayan81:
> I made a few edits. Can you explain the way the params are shared and updated? I'm not sure I see the logic

There is a global model (shared across all workers) and a local model (each worker has its own copy).
The global model’s parameters are placed in shared memory so all workers can access and update them.
Each worker interacts with its own environment using its local model.
After collecting experience, the worker computes gradients (via backpropagation) on its local model.
The gradients from the local model are copied to the corresponding parameters of the global model (global_param._grad = local_param.grad).
The optimizer (e.g., Adam) is called on the global model, using the gradients just copied from the local model.
This updates the global model’s parameters.
The local model can be periodically updated (synced) from the global model to keep it up-to-date.

sota-implementations/a3c/a3c_atari.py
```diff
@@ -0,0 +1,241 @@
...
+        for local_param, global_param in zip(
+            self.local_critic.parameters(), self.global_critic.parameters()
+        ):
+            global_param._grad = local_param.grad
+
+        gn = torch.nn.utils.clip_grad_norm_(
+            self.loss_module.parameters(), max_norm=max_grad_norm
+        )
+
+        self.optimizer.step()
```
theap06, line 136: After this line, we should sync the weights from global.

theap06: from_module(self.global_actor).data.to_module(self.local_actor)
from_module(self.global_critic).data.to_module(self.local_critic)

sota-implementations/a3c/a3c_atari.py
```diff
@@ -0,0 +1,241 @@
...
+            self.local_actor.parameters(), self.global_actor.parameters()
+        ):
+            global_param._grad = local_param.grad
+
+        for local_param, global_param in zip(
+            self.local_critic.parameters(), self.global_critic.parameters()
+        ):
+            global_param._grad = local_param.grad
+
+        gn = torch.nn.utils.clip_grad_norm_(
```
theap06, line 132: I think to be inline with the correct implementation the grad clipping should occur before lines 113-118

theap06 (approved):

theap06:
@simeetnayan81 lmk if you need any help with this pr

gtnv:
hey @simeetnayan81, is this still on your list? if not, can we close it for now?

PR #3022, opened 2025-06-20
[BugFix] ActionMask is compatible with composite action specs

louisfaury:
## Description

`ActionMask` was not compatible with composite action specs because of the lines:
```python
action_spec = self.container.full_action_spec
        keys = self.container.action_keys
        if len(keys) == 1:
            action_spec = action_spec[keys[0]]
```
that afaiu do not serve a real purpose. 

This PR removes this check, and allows tu use `ActionMask` to mask one factor of a composite action spec.

## Types of changes

## Checklist

- [] My change requires a change to the documentation.


torchrl/envs/transforms/transforms.py
```diff
@@ -8773,8 +8775,6 @@ class ActionMask(Transform):
         >>> base_env = MaskedEnv()
         >>> env = TransformedEnv(base_env, ActionMask())
         >>> r = env.rollout(10)
-        >>> env = TransformedEnv(base_env, ActionMask())
-        >>> r = env.rollout(10)
```
louisfaury, line 8777: Already in the lines above.

torchrl/envs/transforms/transforms.py
```diff
@@ -8726,8 +8726,10 @@ def set_container(self, container):
 class ActionMask(Transform):
     """An adaptive action masker.
 
-    This transform reads the mask from the input tensordict after the step is executed,
-    and adapts the mask of the one-hot / categorical action spec.
+    This transform is useful to ensure that randomly generated actions
+    respect legal actions, by masking the action specs.
+    It reads the mask from the input tensordict after the step is executed,
+    and adapts the mask of the finite action spec.
```
louisfaury, line 8732: Took the occasion to refine a bit the docstring.

torchrl/envs/transforms/transforms.py
```diff
@@ -8810,45 +8810,29 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
         raise RuntimeError(FORWARD_NOT_IMPLEMENTED.format(type(self)))
 
     @property
-    def action_spec(self):
-        action_spec = self.container.full_action_spec
-        keys = self.container.action_keys
-        if len(keys) == 1:
-            action_spec = action_spec[keys[0]]
```
louisfaury, line 8817: tbh I could not really see why this check was in place. I hope I did not miss any chesterton fence!

torchrl/envs/transforms/transforms.py
```diff
@@ -8810,45 +8810,29 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
-        action_spec = self.action_spec
-        mask = tensordict.get(self.in_keys[1], None)
-        if mask is not None:
-            mask = mask.to(action_spec.device)
-        action_spec.update_mask(mask)
-
-        # TODO: Check that this makes sense
-        with _set_missing_tolerance(self, True):
-            tensordict_reset = self._call(tensordict_reset)
-        return tensordict_reset
```
louisfaury, line 8851: I think this entire logic is already handled by `_call`

torchrl/envs/transforms/transforms.py
vmoens: It looks good, there is some confusion in the existing code about in_keys and action_keys but this seems to be aligned with what we want. 

We should maybe have a validation step when we register a transform in an env. For example here, when we set the container to the transform, we should check that the `action_key` arg is part of the action_keys of the parent env.

PR #3039, opened 2025-07-07
[Feature] Make dialog_turns_per_batch optional when yield_completed_trajectories=True

albertbou92:
## Description

When using `LLMCollector` with `yield_completed_trajectories=True`, the `dialog_turns_per_batch` parameter is never used but was still required, so it does not make much sense to provide it. This PR makes it optional in this case.

Changes:
- Made dialog_turns_per_batch optional (default None)
- Added validation to require it except when yield_completed_trajectories=True
- Sets dummy value of 0 when not needed to satisfy parent class requirements. This is a good value not to raise this warning:  https://github.com/pytorch/rl/blob/main/torchrl/collectors/collectors.py#L899
- Updated docstring.

## Motivation and Context

## Types of changes

## Checklist


torchrl/collectors/llm/base.py
```diff
@@ -172,6 +172,8 @@ def __init__(
         elif queue is not None:
             # disguise the queue as a replay buffer
             replay_buffer = _QueueAsRB(queue)
+        if dialog_turns_per_batch is None and yield_completed_trajectories:
+            dialog_turns_per_batch = 0
```
vmoens, line 176: Yeah that makes sense

vmoens (approved):
LGTM thanks

PR #3043, opened 2025-07-07
[BugFix] Add reference to policy with state dict

felixy12:
## Description

This PR adds the `_policy_w_state_dict` attribute for the `SyncDataCollector`. The point here is to use this reference to do all saving/loading of the policy state dict. This fixes the bug of not being able to load/save policies that have been compiled/CudaGraph'ed.

## Motivation and Context

Policy models are often saved and loaded via the SyncDataCollector class, which also handles compiling and CudaGraphing the policy. However, this wraps the policy in a few layers that obscures the saving and loading of the policy weights if we attempt the access the policy directly from the `SyncDataCollector.policy` attribute.

The clean solution here is to keep a reference to the unwrapped, original policy object that contains the `state_dict` and perform all model saving/loading via this reference. This way, we don't care about the method in which the object gets wrapped. This all works because every wrapped layer still points to the same underlying set of model weights.

## Types of changes

## Checklist

### DIFF
--- a/torchrl/collectors/collectors.py
+++ b/torchrl/collectors/collectors.py
@@ -689,6 +689,10 @@ class SyncDataCollector(DataCollectorBase):
                 policy = RandomPolicy(env.full_action_spec)
         elif policy_factory is not None:
             raise TypeError("policy_factory cannot be used with policy argument.")
+        # If the underlying policy has a state_dict, we keep a reference to the policy and
+        # do all policy weight saving/loading through it
+        if hasattr(policy, "state_dict"):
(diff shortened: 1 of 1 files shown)


felixy12:
Yes, definitely. I'll throw something up in the near future.

vmoens:
I'm going to go ahead and merge this to include it in the minor to come - if you can submit your test as a separate PR later that's cool!
(doing this as there are few changes that it'll break people's code!)

felixy12:
sounds good, apologies for not getting around to the tests, other aspects of work have been monopolizing my time. :pray:

PR #3046, opened 2025-07-08
Disable unrunnable benchmarks cpu test

ZainRizvi:
## Description

The benchmark_cpu job is currently unrunnable and has been so for  over 3 months.

Changes are:
1. Disables the job from being scheduled, since it's broken.  I'm not deleting the code in case someone is interested in fixing it later on.
2. Fixes the runner label used for benchmark cpu tests to a similar machine that actually exists and has the same amount of cpu/memory, which is the first step towards fixing the job

## Motivation and Context

This buggy runner label was referenced in this PR: https://github.com/pytorch/rl/pull/2930/files#diff-1d24b39978bd364a0913d34dc491f52f4cb244e1efaf28bd0d87ebb80a73e4cdR17

Since this the job has been unable to even start, let alone run. 

## Types of changes

## Checklist

### DIFF
--- a/.github/workflows/benchmarks_pr.yml
+++ b/.github/workflows/benchmarks_pr.yml
@@ -14,7 +14,7 @@ jobs:
 
   benchmark_cpu:
     name: CPU Pytest benchmark
-    runs-on: linux.g5.4xlarge.nvidia.cpu
+    runs-on: linux.4xlarge.nvidia
     defaults:
       run:
         shell: bash -l {0}
--- a/.github/workflows/benchmarks_pr.yml
+++ b/.github/workflows/benchmarks_pr.yml
@@ -14,7 +14,8 @@ jobs:
 
   benchmark_cpu:
     name: CPU Pytest benchmark
-    runs-on: linux.4xlarge.nvidia
+    runs-on: linux.4xlarge
+    if: false 
     defaults:
       run:
         shell: bash -l {0}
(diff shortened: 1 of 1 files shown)


ZainRizvi:
Merging PR

vmoens:
Thanks for taking good care of torchrl's CI @ZainRizvi 

Do you know if instead of disabling faulty runs we could fix them instead?

The way I approach this usually is that I try to fix the CI before the release (hence the 3 monts since it was broken). A red signal in the CI tells me that that there is something to be fixed - even if it's not ideal it's better than no signal IMO. That being said I understand that this has cost implications so I don't want to push too much to keep dead CIs running, but if anything I'd prefer if we could first try to fix them before disabling (if that makes sense).

vmoens:
> Time to follow the fine tradition of this repository and merge without waiting for review or CI. Yolo

About this, if you want to volunteer and review TorchRL's PRs just let me know! Any help is more than welcome

ZainRizvi:
> The way I approach this usually is that I try to fix the CI before the release (hence the 3 monts since it was broken). A red signal in the CI tells me that that there is something to be fixed - even if it's not ideal it's better than no signal IMO. That being said I understand that this has cost implications so I don't want to push too much to keep dead CIs running, but if anything I'd prefer if we could first try to fix them before disabling (if that makes sense).

Gotcha, apologies for disabling the job in that case.  The runner label used is fixed to point to a legitimate runner now, and here's a PR to re-enable the job: https://github.com/pytorch/rl/pull/3048

PR #3058, opened 2025-07-11
[Feature] CompressedStorage

### LINKED ISSUE #2983: [Feature Request] Compressing data stored in the Replay Buffer
## Motivation

Replay buffers are used to store a lot of data and ideally we could put this data on a GPU.
Often raw sensory observations are stored in these buffers, such as images or audio, which consumes many gigabytes of precious VRAM. 
In torchRL, memmapping and prefetching could be considered the fix for this issue but I have instead pursued compression instead. 

I've found that zstd is effective in compressing atari frames (reducing storage use by 150-250x) and my DQN implementation has gone from storing ~4M frames consuming ~20ish GB to less than a GB. 
I would like some pointers on where it might be sensible to integrate compression into the torch rl replay buffer in the hopes that compressing raw sensory observations could be more widely adopted. 

## Solution

I would like feedback on my proposed solutions to implementing compression in the replay buffer. 

**A CompressedReplayBuffer** class that inherents from the ReplayBuffer class could be one option.
This class would compress data "as it comes in", by modifying the `append` and `extend` functions.
The compressed data is then passed to the parent ReplayBuffer class as a uint8 byte stream for storage.
The compressed data is then decompressed in the `sample` function. 
Although, the issue in this approach is that the compressed byte stream can be of variable shape. 
Some images might be compressed into ~150 bytes, but others might be ~188 bytes. 
The TensorStorage object expects to receive objects of a consistent shape, so perhaps this is an inconvenient place to implement compression..

**A compressed storage type** might make more sense, as ultimately we want a storage type that can tolerate storing data of different lengths. 
One approach could be to use the standard replay buffer but with a `ListStorage` type.
By modifying the `get` and `set` functions we could compress and decompress as data comes in and out. 
Alternatively, `TensorStorage` could be modified to use a [nested & jagged layout](https://docs.pytorch.org/docs/stable/nested.html), which might be more efficient.

## Additional context

Although I have used zstd as the compression algorithm in this issue, ideally we don't lock someone into using it. 
Compression is just converting any arbitrary tensor into a uint8 1d tensor (the byte stream) and then decompression is rebuilding the prior tensor and recovering the shape and datatype. 
If the data is sensory observations from a gym-like environment, we can reconstruct the observation's original shape and datatype from the gym environment metadata. 

## Checklist

vmoens:
@AdrianOrenstein feedback on this?


AdrianOrenstein:
I think the PR is great. If I get permissions to modify you could delegate some of the tasks to me if you'd like. 

I've left some comments below:

- The conversion of an object to a bytestream is hardcoded. If we move that into a function, it'll be easier to later extend this class to serialise the other objects like numpy.arrays or a pickle dump of some object. 
- I don't think we had any tests on numpy objects in the tensordict or arbitrary python objects. 

- Is the goal to keep the scope of the PR just for CPU data, CPU storage, CPU compression, and CPU decompression? If so, I might follow up with a PR to include some GPU compression. I like the API and it is in a state where I'd be able to extend it. 

- In `docs/source/reference/data.rst` you mention a 2-10x compression ratio. I was thinking of including [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/latest/) as a dependancy and then we can get a number on compressing atari frames. This might also be out of scope so I could do it later. 

- There is `pip install zstd` that provides a stateless compression and decompression API which might be more convenient.

- If you're open to adding `pytest-benchmark` as a dependancy, we could make this a regression test. It might make sense to add an Atari rollout to evaluate the expected compression ratio.

vmoens:
All of these look good! I think it's going to be an amazing feature!

I don't think we want a separate PR for GPU. Not particularly opinionated though, if you feel like splitting this in separate chunks I'm good with it.
Also you can just branch out of this branch and make a new PR with your changes, I'll close this one.
Thanks!

AdrianOrenstein:
Great, I'll get started shortly on the gpu side of things. Also thanks for getting me started with this pr, it has helped me wrap my head around a lot of the details of the codebase.

vmoens:
Closing in favor of #3062

Issue #3067, opened 2025-07-14
[Feature Request] Add support for loading datasets from local Minari cache


Ibinarriaga8:
## Motivation

The current implementation of `MinariExperienceReplay` requires datasets to be downloaded using the class itself, which creates an `env_metadata.json` file in the target directory. This workflow does not accommodate custom Minari datasets created by users or datasets that have been loaded into the local Minari cache by other means (e.g., through `minari.load_dataset` or custom dataset creation via `DataCollector`). 

As a result, attempting to instantiate `MinariExperienceReplay` with `download=False` for locally available datasets leads to a `FileNotFoundError` due to missing metadata files, even though the dataset exists in the Minari cache. This limitation is frustrating for users who want to leverage their own datasets without redownloading or duplicating data, and it hinders workflows where datasets are managed independently of TorchRL.

This issue is meant to enable loading datasets directly from the local Minari cache (typically `~/.minari/datasets`) without requiring prior setup via `MinariExperienceReplay`'s download workflow, making it more flexible and compatible with custom and preloaded datasets.

## Solution

Add and fully support the argument `load_from_local_minari` to the `MinariExperienceReplay` class. When set to `True`, this argument will instruct the class to:

- Look for the dataset in the user's local Minari cache (e.g., `~/.minari/datasets/{dataset_id}/data/main_data.hdf5`).
- Bypass any download or remote fetching logic.
- If the required files are present, load the dataset and construct any necessary metadata on-the-fly (e.g., from the Minari dataset spec, if possible).
- Raise a clear and informative `FileNotFoundError` if the dataset is not found in the expected local cache location.
- Ensure that custom datasets created by users (such as those with `DataCollector(...).create_dataset(...)`) or datasets first loaded with `minari.load_dataset` can be used seamlessly with `MinariExperienceReplay`.

This solution allows for greater flexibility, avoids unnecessary downloads and data duplication, and makes TorchRL compatible with the wider Minari ecosystem.

## Alternatives

- **Manual copying of files**: Users could manually copy datasets and metadata to the expected TorchRL directory, but this is error-prone and not user-friendly.
- **Automated metadata generation scripts**: Provide standalone tools for generating `env_metadata.json` based on existing Minari datasets. This adds maintenance burden and complexity for users.

## Additional context

- The new `load_from_local_minari` argument should default to `False` to preserve backward compatibility.
[...]

PR #3068, opened 2025-07-14
[Feature] Add support for loading datasets from local Minari cache

Ibinarriaga8:
## Description

This PR adds support for loading datasets directly from the local Minari cache in the `MinariExperienceReplay` class. Specifically, it introduces the `load_from_local_minari` argument, which, when set to `True`, instructs the class to load the dataset from the user's local Minari cache (typically at `~/.minari/datasets`) and skip any fetching from the Minari server (i.e., no remote download or overwrite will occur). After loading from the local cache, all subsequent preprocessing and loading steps continue as usual, ensuring the dataset is processed and made available correctly. This is especially useful for custom Minari datasets or when you want to avoid network access.

**Documentation has been updated in the class docstring** to clearly state the new behavior of the `load_from_local_minari` argument, including details about local cache prioritization and unchanged downstream preprocessing.

This PR also includes **comprehensive test coverage** for the new feature, confirming that datasets created and stored in the local Minari cache can be loaded, sampled, and validated for correctness using the new argument. The provided test (`test_local_minari_dataset_loading`) creates a custom dataset, loads it from cache, verifies sample integrity, and cleans up afterwards.

## Motivation and Context

Previously, `MinariExperienceReplay` required datasets to be downloaded via its own interface, which was incompatible with custom or preloaded Minari datasets, attempting to load these would result in a `FileNotFoundError`. 
This change allows users to work with their own datasets, datasets created with `minari.DataCollector(...).create_dataset(...)`, or any dataset present in the local Minari cache, without requiring redundant downloads or manual metadata copying.

If the dataset is not found in the local cache, a `FileNotFoundError` is raised with a clear message.

Solves #3067

## Types of changes

## Checklist

### DIFF
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -29,12 +29,12 @@ from pathlib import Path
 from sys import platform
 from unittest import mock
 
+import minari
+
 import numpy as np
 import pytest
 import torch
 from minari import DataCollector
-import gymnasium as gym
-import minari
 
 from packaging import version
 from tensordict import (
@@ -3455,9 +3455,10 @@ class TestMinari:
         assert len(dataset) == 100
         assert sample["data"].shape == torch.Size([32, 8])
         assert sample["next", "data"].shape == torch.Size([32, 8])
-    
-        
-    @pytest.mark.skipif(not _has_minari or not _has_gymnasium, reason="Minari or Gym not available")
+
+    @pytest.mark.skipif(
+        not _has_minari or not _has_gymnasium, reason="Minari or Gym not available"
+    )
     def test_local_minari_dataset_loading(self):
 
         if not _minari_init():
@@ -3466,7 +3467,7 @@ class TestMinari:
         dataset_id = "cartpole/test-local-v1"
 
         # Create dataset using Gym + DataCollector
-        env = gym.make("CartPole-v1")
+        env = gymnasium.make("CartPole-v1")
         env = DataCollector(env, record_infos=True)
         for _ in range(50):
             env.reset(seed=123)
@@ -3482,7 +3483,7 @@ class TestMinari:
             code_permalink="https://github.com/Farama-Foundation/Minari",
             author="Farama",
             author_email="contact@farama.org",
-            eval_env="CartPole-v1"
+            eval_env="CartPole-v1",
         )
 
         # Load from local cache
@@ -3499,14 +3500,21 @@ class TestMinari:
         t0 = time.time()
         for i, sample in enumerate(data):
             t1 = time.time()
-            torchrl_logger.info(f"[Local Minari] Sampling time {1000 * (t1 - t0):4.4f} ms")
-            assert data.metadata["action_space"].is_in(sample["action"]), "Invalid action sample"
-            assert data.metadata["observation_space"].is_in(sample["observation"]), "Invalid observation sample"
+            torchrl_logger.info(
+                f"[Local Minari] Sampling time {1000 * (t1 - t0):4.4f} ms"
+            )
+            assert data.metadata["action_space"].is_in(
+                sample["action"]
+            ), "Invalid action sample"
+            assert data.metadata["observation_space"].is_in(
+                sample["observation"]
+            ), "Invalid observation sample"
             t0 = time.time()
             if i == 10:
                 break
-        
+
         minari.delete_dataset(dataset_id="cartpole/test-local-v1")
+
 
 @pytest.mark.slow
 class TestRoboset:
--- a/test/test_libs.py
+++ b/test/test_libs.py
@@ -32,6 +32,9 @@ from unittest import mock
 import numpy as np
 import pytest
 import torch
+from minari import DataCollector
+import gymnasium as gym
+import minari
 
 from packaging import version
 from tensordict import (
@@ -3452,7 +3455,58 @@ class TestMinari:
         assert len(dataset) == 100
         assert sample["data"].shape == torch.Size([32, 8])
         assert sample["next", "data"].shape == torch.Size([32, 8])
+    
+        
+    @pytest.mark.skipif(not _has_minari or not _has_gymnasium, reason="Minari or Gym not available")
+    def test_local_minari_dataset_loading(self):
 
+        if not _minari_init():
+            pytest.skip("Failed to initialize Minari datasets")
+
+        dataset_id = "cartpole/test-local-v1"
+
+        # Create dataset using Gym + DataCollector
+        env = gym.make("CartPole-v1")
+        env = DataCollector(env, record_infos=True)
+        for _ in range(50):
+            env.reset(seed=123)
+            while True:
+                action = env.action_space.sample()
+                obs, rew, terminated, truncated, info = env.step(action)
+                if terminated or truncated:
+                    break
+
+        env.create_dataset(
+            dataset_id=dataset_id,
+            algorithm_name="RandomPolicy",
+            code_permalink="https://github.com/Farama-Foundation/Minari",
+            author="Farama",
+            author_email="contact@farama.org",
+            eval_env="CartPole-v1"
+        )
+
+        # Load from local cache
+        data = MinariExperienceReplay(
+            dataset_id=dataset_id,
+            split_trajs=False,
+            batch_size=32,
+            download=False,
+            sampler=SamplerWithoutReplacement(drop_last=True),
+            prefetch=2,
+            load_from_local_minari=True,
+        )
+
+        t0 = time.time()
+        for i, sample in enumerate(data):
+            t1 = time.time()
+            torchrl_logger.info(f"[Local Minari] Sampling time {1000 * (t1 - t0):4.4f} ms")
+            assert data.metadata["action_space"].is_in(sample["action"]), "Invalid action sample"
+            assert data.metadata["observation_space"].is_in(sample["observation"]), "Invalid observation sample"
+            t0 = time.time()
+            if i == 10:
+                break
+        
+        minari.delete_dataset(dataset_id="cartpole/test-local-v1")
 
 @pytest.mark.slow
 class TestRoboset:
--- a/torchrl/data/datasets/minari_data.py
+++ b/torchrl/data/datasets/minari_data.py
@@ -88,6 +88,14 @@ class MinariExperienceReplay(BaseDatasetExperienceReplay):
             it is assumed that any ``truncated`` or ``terminated`` signal is
             equivalent to the end of a trajectory.
             Defaults to ``False``.
+        load_from_local_minari (bool, optional): if ``True``, the dataset will be loaded directly
+            from the local Minari cache (typically located at ``~/.minari/datasets``),
+            bypassing any remote download. This is useful when working with custom
+            Minari datasets previously generated and stored locally, or when network
+            access should be avoided. If the dataset is not found in the expected
+            cache directory, a ``FileNotFoundError`` will be raised. 
+            Defaults to ``False``.
+
 
     Attributes:
         available_datasets: a list of accepted entries to be downloaded.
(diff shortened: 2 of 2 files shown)


test/test_libs.py
```diff
@@ -3453,6 +3456,65 @@ def fn(data):
...
+        not _has_minari or not _has_gymnasium, reason="Minari or Gym not available"
+    )
+    def test_local_minari_dataset_loading(self):
+
+        if not _minari_init():
+            pytest.skip("Failed to initialize Minari datasets")
+
+        dataset_id = "cartpole/test-local-v1"
+
+        # Create dataset using Gym + DataCollector
```
Ibinarriaga8: Custom minari dataset creation from a gymnasium environment

vmoens (approved):
That's really handy I love it!

vmoens:
@Ibinarriaga8

Do you think we have an opportunity here to reduce the number of datasets we download to test minari, and use customly built datasets instead like you do in your test?

Ibinarriaga8:
> @Ibinarriaga8
> 
> Do you think we have an opportunity here to reduce the number of datasets we download to test minari, and use customly built datasets instead like you do in your test?

Absolutely, instead of downloading 20 datasets from the Minari server, we could generate smaller, custom datasets from any gymnasium environment as part of our tests. This approach is especially valuable for D4RL datasets, which tend to be very large and can significantly slow down testing.

test/test_libs.py
```diff
@@ -29,9 +29,12 @@
 from sys import platform
 from unittest import mock
 
+import minari
```
vmoens: This is a global import - we must avoid these at all cost.
Can you make it local?

vmoens:
> Absolutely, instead of downloading 20 datasets from the Minari server, we could generate smaller, custom datasets from any gymnasium environment as part of our tests. This approach is especially valuable for D4RL datasets, which tend to be very large and can significantly slow down testing.

Do you want to give it a go? Otherwise I can do it no worry

test/test_libs.py
Ibinarriaga8: Ok, I can fix it

Ibinarriaga8:
> > Absolutely, instead of downloading 20 datasets from the Minari server, we could generate smaller, custom datasets from any gymnasium environment as part of our tests. This approach is especially valuable for D4RL datasets, which tend to be very large and can significantly slow down testing.
> 
> Do you want to give it a go? Otherwise I can do it no worry

Yes, no problem

Ibinarriaga8:
@vmoens  Do you think we should replace testing with all Minari datasets by using smaller, custom datasets generated from gymnasium environments? Or should we still download and test with some datasets from the Minari server to ensure that the downloading functionality in MinariExperienceReplay works correctly?

vmoens:
> @vmoens Do you think we should replace testing with all Minari datasets by using smaller, custom datasets generated from gymnasium environments? Or should we still download and test with some datasets from the Minari server to ensure that the downloading functionality in MinariExperienceReplay works correctly?

Maybe just one defined (not random) dataset - to make sure?

test/test_libs.py
```diff
@@ -29,9 +29,11 @@
 from sys import platform
 from unittest import mock
 
+
 import numpy as np
 import pytest
 import torch
+from minari import DataCollector
```
vmoens: Still a global import

Ibinarriaga8:
> > @vmoens Do you think we should replace testing with all Minari datasets by using smaller, custom datasets generated from gymnasium environments? Or should we still download and test with some datasets from the Minari server to ensure that the downloading functionality in MinariExperienceReplay works correctly?
> 
> Maybe just one defined (not random) dataset - to make sure?

Ok, I agree that testing with a single, well-defined dataset makes sense to ensure the download and loading functionality are covered.

However, with most gym environments, creating custom datasets isn’t always straightforward. For example, I ran into this error:

```
ValueError: Dict key mismatch; expected keys: ['reward_ctrl', 'reward_forward', 'reward_survive', 'x_position', 'x_velocity', 'z_distance_from_origin']; dict: {'x_position': np.float64(0.001823518632481435), 'z_distance_from_origin': np.float64(-0.004461789811977868)}
```

I’ll keep investigating to see if there’s a workaround or a more general approach that works across environments.

vmoens:
Ok maybe we could proceed with this in its current state and refactor the tests later!

Ibinarriaga8:
> Ok maybe we could proceed with this in its current state and refactor the tests later!

Ok, I have refactored the tests to use this custom datasets approach. I am still fixing some minor errors and making sure it all works.
I can either commit these changes to this PR or create a separate PR for the test refactor. Whichever you prefer

vmoens:
If you're ready feel free to commit here directly

test/test_libs.py
```diff
@@ -3341,6 +3341,39 @@ def test_d4rl_iteration(self, task, split_trajs):
 
 _MINARI_DATASETS = []
 
+MUJOCO_ENVIRONMENTS = [
```
Ibinarriaga8, line 3344: Explicitly specified current set of Minari supported datasets for integration with Gym environments

test/test_libs.py
```diff
@@ -3373,30 +3406,148 @@ def _minari_init():
         return False
 
 
-# Initialize with placeholder values for parametrization
-# These will be replaced with actual dataset names when the first Minari test runs
-_MINARI_DATASETS = [str(i) for i in range(20)]
+def get_random_minigrid_datasets():
```
Ibinarriaga8, line 3442: Only use Minari datasets from Minigrid

This is because current version of  Minari cannot serialize custom MissionSpace objects, which are used in most Minigrid environments

test/test_libs.py
Ibinarriaga8: Specified datasets are those used in Mujoco and D4RL

test/test_libs.py
```diff
@@ -3373,30 +3406,148 @@ def _minari_init():
...
+            data = MinariExperienceReplay(
+                dataset_id=dataset_id,
+                batch_size=32,
+                split_trajs=split,
+                download="force",
+            )
+            cleanup_needed = False
+
+        else:
+            # Atari environment datasets
```
Ibinarriaga8, line 3583: For Atari environments, select a random subset of environments.

test/test_libs.py
```diff
@@ -29,11 +29,9 @@
 from sys import platform
 from unittest import mock
 
-
```
Ibinarriaga8: Removed global imports

test/test_libs.py
Ibinarriaga8, line 1: Note: new dependencies were added for creating custom Minari datasets from gym environments:

- ale_py (for Atari)
- gymnasium_robotics (for D4RL & Mujoco)

test/test_libs.py
Ibinarriaga8: This means you cannot create your custom dataset from a minigrid environment directly; you will need to modify Mission space

Ibinarriaga8:
> LGTM! This actually makes me think that we should have a similar pipeline where one creates a datacollector with some replay buffer and serializes it as a dataset later on. It's mostly a matter of documenting it rather than creating the feature, everything already exists I think!

Ok, should I provide a tutorial documenting the pipeline from DataCollector with a replay buffer to dataset serialization?

vmoens:
First let's fix this!
From the CI I see there are some global imports of ale_py etc occurring in every workflow. Probably because Minari is imported in some function that is called in the global workspace. 
Every Minari import should be localised within the tests

Ibinarriaga8:
> First let's fix this! From the CI I see there are some global imports of ale_py etc occurring in every workflow. Probably because Minari is imported in some function that is called in the global workspace. Every Minari import should be localised within the tests

I already import minari and ale_py only inside the specific functions or test cases that require them, rather than at the global level.
Is there a particular place where you’re seeing these imports happen globally, or could it be that the function importing them is being executed in the global workspace?

Ibinarriaga8:
> > First let's fix this! From the CI I see there are some global imports of ale_py etc occurring in every workflow. Probably because Minari is imported in some function that is called in the global workspace. Every Minari import should be localised within the tests
> 
> I already import minari and ale_py only inside the specific functions or test cases that require them, rather than at the global level. Is there a particular place where you’re seeing these imports happen globally, or could it be that the function importing them is being executed in the global workspace?

Do you think moving all imports to the _minari_init which only makes necessary imports if dependencies are installed would solve this?

vmoens:
No imports must happen if and only if they're needed. Otherwise spawned multiprocessed jobs all import a ton of useless libs (if they're installed).
We only want local imports without try except blocks.

test/test_libs.py
```diff
@@ -3373,30 +3406,148 @@ def _minari_init():
...
 @pytest.mark.slow
 class TestMinari:
     @pytest.mark.parametrize("split", [False, True])
-    @pytest.mark.parametrize("dataset_idx", range(20))
+    @pytest.mark.parametrize(
+        "dataset_idx",
+        range(
+            len(MUJOCO_ENVIRONMENTS)
+            + len(D4RL_ENVIRONMENTS)
+            + len(get_random_minigrid_datasets())
```
Ibinarriaga8: These two functions import minari and are called at the module level (global workspace), which causes the error during CI.
I will refactor them so that the imports and calls only occur within the relevant test functions.

test/test_libs.py
```diff
@@ -3488,12 +3488,7 @@ class TestMinari:
     @pytest.mark.parametrize("split", [False, True])
     @pytest.mark.parametrize(
         "dataset_idx",
-        range(
-            len(MUJOCO_ENVIRONMENTS)
```
Ibinarriaga8: These decorators were called in global workspace and imported minari and ale_py internally

 Now, I use a static upper bound, so that no function that imports minari or ale_py globally is called

vmoens:
Cool! Looks like the conflicts with other CI runs are resolved. We have some long-lasting failures but I wouldn't pay too much attention to them.
The minari tests are failing though - do you want me to help? I see a "too many requests" error which I think can only be solved by having fewer tests and a missing dependency to be added [here](https://github.com/pytorch/rl/blob/main/.github/unittest/linux_libs/scripts_minari/environment.yml)

test/test_libs.py
Ibinarriaga8, line 1: Reduce number of fetched datasets to avoid 429 Error

.github/unittest/linux_libs/scripts_minari/environment.yml
```diff
@@ -21,3 +21,5 @@ dependencies:
     - hydra-core
     - minari[gcs,hdf5,hf]
     - gymnasium<1.0.0
+    - ale-py
```
Ibinarriaga8, line 24: Added new dependencies

vmoens:
For some reason the CI fails to identify where mujoco has been installed - let me investigate

vmoens:
The CI is almost fixed! Can you pull the changes and have a look at the remaining issues?
I see we pin `gymnasium<1.0.0` - do we want a more recent version?

Ibinarriaga8:
> The CI is almost fixed! Can you pull the changes and have a look at the remaining issues? I see we pin `gymnasium<1.0.0` - do we want a more recent version?

Yes, I believe we are getting gymnasium.error.VersionNotFound errors due to an outdated version of gymnasium
I have tried on local with **gymnasium==1.2.0** and all versions are available

Ibinarriaga8:
@vmoens I pulled your changes and ran tests on local and everything works fine; I think updating gym version would fix CI issues

vmoens:
@Ibinarriaga8 I refactored the minari cache logic in [ba6b299](https://github.com/pytorch/rl/commit/ba6b299a78080b1e6fd515f58beb4f2e233918b5) 
can you check if that makes sense?

Ibinarriaga8:
lgtm

PR #3075, opened 2025-07-16
[Feature] MinariExperienceReplay now can handle text fields like "mission"

### LINKED ISSUE #3071: [Feature Request] MinariExperienceReplay now can handle text fields like "mission"
## Motivation

MinariExperienceReplay at this moment is not able to manage text fields like "missions". This feature is extremely important for BabyAI datasets, that might have diverse missions or even changing missions inside an episode.

## Solution

I have an already functioning fork that adds a new argument to MinariExperienceReplay called string_to_tensor_map. This argument is typed as a dictionary mapping string to Tensor, where each key of the dictionary would be any of the keys of the observation TensorDict.

The original NonTensorData values would be mapped with the function to the desired vector representation of each mission (or text data).

## Alternatives

Can't think of any other solution that does not hardcode a method for obtaining tensors from NonTensorData.

## Additional context

[This is my working fork, with tests and documentation.](https://github.com/marcosgalleterobbva/rl)

## Checklist

- [ x] I have checked that there is no similar issue in the repo (**required**)

marcosgalleterobbva:
## Description

Adds a new argument string_to_tensor_map that allows the user to parse NonTensorData fields like "mission" or any other categorical values inside the observation TensorDict and assign it any Tensor value they might prefer, be it a one hot encoding version of the categorical field or a BERT-like embedding of the text fields.

### Technical Details

I would need @vmoens to take a small look to this specific set of lines. It works, but I don't know if this specific implementation is correct.
https://github.com/marcosgalleterobbva/rl/blob/c5c075d6eb1da858f580392d7567bc78a8679ce0/torchrl/data/datasets/minari_data.py#L386

## Motivation and Context

The motivation is specified in issue [#3071](https://github.com/pytorch/rl/issues/3071)
`close #3071`

- [ x] I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

- [ x] Documentation (update in the documentation)
- [ x] Example (update in the folder of examples)

## Checklist

- [ x] My change requires a change to the documentation.
- [ x] I have updated the documentation accordingly.

### DIFF
--- a/test/test_loggers.py
+++ b/test/test_loggers.py
@@ -280,6 +280,7 @@ class TestWandbLogger:
                 value=scalar_value,
                 name=scalar_name,
                 step=steps[i] if steps else None,
+                commit=True,
             )
 
         assert wandb_logger.experiment.summary["foo"] == values[-1].item()
@@ -315,9 +316,8 @@ class TestWandbLogger:
         assert video_4fps_size > video_16fps_size, (video_4fps_size, video_16fps_size)
 
         # check that we catch the error in case the format of the tensor is wrong
-        video_wrong_format = torch.zeros(64, 2, 32, 32)
-        video_wrong_format = video_wrong_format[None, :]
-        with pytest.raises(Exception):
+        video_wrong_format = torch.zeros(2, 32, 32)
+        with pytest.raises(ValueError, match="Video must be at least"):
             wandb_logger.log_video(
                 name="foo",
                 video=video_wrong_format,
--- a/torchrl/record/loggers/wandb.py
+++ b/torchrl/record/loggers/wandb.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 import importlib.util
 
 import os
-import warnings
 from typing import Sequence
 
 from torch import Tensor
@@ -47,7 +46,6 @@ class WandbLogger(Logger):
 
     @classmethod
     def __new__(cls, *args, **kwargs):
-        cls._prev_video_step = -1
         return super().__new__(cls)
 
     def __init__(
@@ -84,25 +82,19 @@ class WandbLogger(Logger):
             "resume": "allow",
             **kwargs,
         }
-        self._has_imported_wandb = False
+
         super().__init__(exp_name=exp_name, log_dir=save_dir)
         if self.offline:
             os.environ["WANDB_MODE"] = "dryrun"
 
-        self._has_imported_moviepy = False
-
-        self._has_imported_omgaconf = False
-
-        self.video_log_counter = 0
-
-    def _create_experiment(self) -> WandbLogger:
+    def _create_experiment(self):
         """Creates a wandb experiment.
 
         Args:
             exp_name (str): The name of the experiment.
 
         Returns:
-            WandbLogger: The wandb experiment logger.
+            A wandb.Experiment object.
         """
         if not _has_wandb:
             raise ImportError("Wandb is not installed")
@@ -113,7 +105,9 @@ class WandbLogger(Logger):
 
         return wandb.init(**self._wandb_kwargs)
 
-    def log_scalar(self, name: str, value: float, step: int | None = None) -> None:
+    def log_scalar(
+        self, name: str, value: float, step: int | None = None, commit: bool = False
+    ) -> None:
         """Logs a scalar value to wandb.
 
         Args:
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -32,6 +32,9 @@ If the generation of this artifact in MacOs M1 doesn't work correctly or in the
 ARCHFLAGS="-arch arm64" python setup.py develop
 ```
 
+In some MacOs devices, the installation of the required libraries errors if the correct version of
+clang is not used. Using `llvm@16` (installable with brew), may fix your issues.
```
vmoens: Thanks!

torchrl/data/datasets/minari_data.py
```diff
@@ -243,38 +234,72 @@ def _download_and_preproc(self):
             minari.download_dataset(dataset_id=self.dataset_id)
             parent_dir = Path(tmpdir) / self.dataset_id / "data"
 
-            td_data = TensorDict()
+            td_data = TensorDict({}, batch_size=[])
```
vmoens: those two are equivalent, and I personally like `TensorDict()` more :) 
```suggestion
            td_data = TensorDict()
```

tutorials/sphinx-tutorials/minari_data_loading.py
vmoens, line 1: Tutorials have a specific formatting, need to be registered etc.
If it's an example, it should go in `examples/`

torchrl/data/datasets/minari_data.py
```diff
@@ -313,26 +334,105 @@ def _download_and_preproc(self):
...
                                     f"Mismatching number of steps for key {key}: was {steps} but got {val.shape[0] - 1}."
                                 )
-                            data_view["next", match].copy_(val[1:])
-                            data_view[match].copy_(val[:-1])
-                        elif key not in ("terminations", "truncations", "rewards"):
-                            if steps is None:
-                                steps = val.shape[0]
-                            else:
-                                if steps != val.shape[0]:
+                        if key == "observations":
```
vmoens: Can you comment on the change of logic here?
Hard for me to grasp precisely what is going on

torchrl/data/datasets/minari_data.py
```diff
@@ -313,26 +334,105 @@ def _download_and_preproc(self):
...
+                                            #       Tensordict. I would prefer something like the following:
+                                            #           for i in range(len(subval) - 1):
+                                            #               data_view[i].set(("observation", subkey), subval[i])
+                                            #               data_view[i].set(("next", "observation", subkey), subval[i + 1])
+                                            #       But this three previous lines give this error:
+                                            #           RuntimeError: Cannot modify locked TensorDict. For in-place
+                                            #           modification, consider using the `set_()` method and make
+                                            #           sure the key is present.
+                                            #       But this current approach takes incredibly long to complete, maybe
+                                            #       I should do something different?
```
vmoens: I think anything you think should work should work :)

I will give it a stab, but you can also write the code as you'd expect it to work and I'll try to patch tensordict / torchrl to make the necessary amendments!

vmoens (commented):
Great work thanks
Overall LGTM but I left a few comments.
My biggest question is why don't we simply use the NonTensorData feature from the tensordict library to store the text as non-tensor?

marcosgalleterobbva:
> Great work thanks Overall LGTM but I left a few comments. My biggest question is why don't we simply use the NonTensorData feature from the tensordict library to store the text as non-tensor?

Hi, @vmoens . Love to hear that this PR looks good.

The reason to move from NonTensorData to Tensor is mainly due to datasets similar to [minigrid and BabyAI](https://minigrid.farama.org/environments/babyai/index.html).

In these environments, the 'mission' key of a given observation is absolutely esential for the agent to correctly assess which action to take. And we need to make sure we have a method for integrating these observations in the overall process. Prior to this PR, if you downloaded a dataset_id like minigrid/BabyAI-PutNextS7N4/optimal-v0, the mission key would be downloaded as a NonTensorData.

In order to be able to include this field in the agents Q-function approximation, the natural step forward would be to try to leverage the transform argument of the MinariExperienceReplay to apply a transformation to these categorical observations (mainly "mission" and "direction") but, unfortunately, if we tried to apply a Compose transformation to this observation/mission key, we would get that the mission key is not loaded as an observation in the loading process.

I also thought about a different approach. Maybe I could try to modify the loading process and, instead of changing the download process (like this PR does), just modify the data that is loaded from disk with a transform Compose. But this would not work.

If you downloaded a dataset_id with mission observations (like minigrid/BabyAI-PutNextS7N4/optimal-v0) and took a look at the values that are stored for this mission observation, you would see that MinariExperienceReplay is not downloading to disk the correct missions for each episode. Instead, it is just taking the mission for the first episode and copying it to all the other episodes of the dataset.

This PR allows us to transform the incoming NonTensorData at the moment of saving it to disk, so that, when we might want to apply a transformation to this field, we actually get the correct data. The modification to the env_metadata.json is included in the PR.

marcosgalleterobbva:
This issue will be handled in different PR in a different form that does not involve saving a hardcoded version of NonTensorData as tensors in the disk. I will get to this as soon as I can

PR #3076, opened 2025-07-16
[BugFix] Fix in-place modification of specs in TransformedEnv

### LINKED ISSUE #3073: [BUG] `TransformedEnv` modifies its`base_env` input_spec
## Describe the bug

When using a transformed env with a `RenameTransform` over the actions, we have that the `base_env.input_spec` is modified as well, leading to errors due to keys not being found.

## To Reproduce

```python
from collections.abc import Sequence

import pytest
import torch
from torchrl.data import Binary, Categorical, Composite, UnboundedContinuous
from torchrl.envs import EnvBase, GymEnv, TransformedEnv
from torchrl.envs.transforms import RenameTransform

class DummyCompositeEnv(EnvBase):  # type: ignore[misc]
    """A dummy environment with a composite action set."""

    def __init__(self) -> None:
        super().__init__()

        self.observation_spec = Composite(observation=UnboundedContinuous((*self.batch_size, 3)))

        self.action_spec = Composite(
            action=Composite(
                head_0=Composite(action=Categorical(2, (*self.batch_size, 1), dtype=torch.bool)),
                head_1=Composite(action=Categorical(2, (*self.batch_size, 1), dtype=torch.bool)),
            )
        )

        self.done_spec = Categorical(2, (*self.batch_size, 1), dtype=torch.bool)

        self.full_done_spec["truncated"] = self.full_done_spec["terminated"].clone()

        self.reward_spec = UnboundedContinuous(*self.batch_size, 1)

    def _reset(self, tensordict: TensorDict) -> TensorDict:
        return TensorDict({"observation": torch.randn((*self.batch_size, 3)), "done": False})

    def _step(self, tensordict: TensorDict) -> TensorDict:
        return TensorDict(
            {
                "observation": torch.randn((*self.batch_size, 3)),
                "done": False,
                "reward": torch.randn((*self.batch_size, 1)),
            }
        )

    def _set_seed(self, seed: int) -> None:
        pass

class PatchedRenameTransform(RenameTransform):  # type: ignore[misc]
    """
    Fixes a bug in the RenameTransform due to modifying the input_spec of the `base_env` to be transformed.
    This is fixed by adding a clone to break stateful modifications to proapagate to the `base_env`.
    """

    def transform_input_spec(self, input_spec: Composite) -> Composite:
        input_spec = input_spec.clone()
        return super().transform_input_spec(input_spec)

def test__PatchedRenameTransform__is_needed() -> None:
    """
    Tests whether the `RenameTransform` modifies the action spec in a stateful manner.

    Whenever this test fails, substitute all occurences of `PatchedRenameTransform` with `RenameTransform`
    """
    transformed_env = TransformedEnv(
        DummyCompositeEnv(),
        RenameTransform(
            in_keys=[],
            out_keys=[],
            in_keys_inv=[("action", "head_0", "action")],
            out_keys_inv=[("action", "head_99", "action")],
        ),
    )
    td = transformed_env.reset()
    # A second reset with a TD passed fails due to override of the `input_spec`
    with pytest.raises(KeyError):
        td = transformed_env.reset(td)

    transformed_env = TransformedEnv(
        DummyCompositeEnv(),
        PatchedRenameTransform(
            in_keys=[],
            out_keys=[],
            in_keys_inv=[("action", "head_0", "action")],
            out_keys_inv=[("action", "head_99", "action")],
        ),
    )
    transformed_env.reset()
    # Won't fail since the transform always clones the specs
    td = transformed_env.reset(td)

```

## Expected behavior

No fail should happen, since the `transform_input_spec` should not operate on the `base_env.input_spec` in a stateful manner.

## Reason and Possible fixes

I suggest fixing the `transform_input_spec` when reward is called, adding a `.clone()` to `self.base_env.input_spec.unlock_()` to avoid to propagate any modification happening.

https://github.com/pytorch/rl/blob/77c00b910e6fdd85aa94b4d354390b724af4ec94/torchrl/envs/transforms/transforms.py#L1210-L1216

I am available to open a PR.

## Checklist

- [X] I have read the [documentation](https://github.com/pytorch/rl/tree/main/docs/) (**required**)
- [X] I have provided a minimal working example to reproduce the bug (**required**)

vmoens:
Closes #3073

Additionally fixes prints of empty `Composite` specs

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -10383,6 +10383,78 @@ def test_added_transforms_are_in_eval_mode():
 
 
 class TestTransformedEnv:
+    class DummyCompositeEnv(EnvBase):  # type: ignore[misc]
+        """A dummy environment with a composite action set."""
+
+        def __init__(self) -> None:
+            super().__init__()
+
+            self.observation_spec = Composite(
+                observation=UnboundedContinuous((*self.batch_size, 3))
+            )
+
+            self.action_spec = Composite(
+                action=Composite(
+                    head_0=Composite(
+                        action=Categorical(2, (*self.batch_size, 1), dtype=torch.bool)
+                    ),
+                    head_1=Composite(
+                        action=Categorical(2, (*self.batch_size, 1), dtype=torch.bool)
+                    ),
+                )
+            )
+
+            self.done_spec = Categorical(2, (*self.batch_size, 1), dtype=torch.bool)
+
+            self.full_done_spec["truncated"] = self.full_done_spec["terminated"].clone()
+
+            self.reward_spec = UnboundedContinuous(*self.batch_size, 1)
+
+        def _reset(self, tensordict: TensorDict) -> TensorDict:
+            return TensorDict(
+                {"observation": torch.randn((*self.batch_size, 3)), "done": False}
+            )
+
+        def _step(self, tensordict: TensorDict) -> TensorDict:
+            return TensorDict(
+                {
+                    "observation": torch.randn((*self.batch_size, 3)),
+                    "done": False,
+                    "reward": torch.randn((*self.batch_size, 1)),
+                }
+            )
+
+        def _set_seed(self, seed: int) -> None:
+            pass
+
+    class PatchedRenameTransform(RenameTransform):  # type: ignore[misc]
+        """
+        Fixes a bug in the RenameTransform due to modifying the input_spec of the `base_env` to be transformed.
+        This is fixed by adding a clone to break stateful modifications to proapagate to the `base_env`.
+        """
+
+        def transform_input_spec(self, input_spec: Composite) -> Composite:
+            input_spec = input_spec.clone()
+            return super().transform_input_spec(input_spec)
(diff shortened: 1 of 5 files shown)


test/test_transforms.py
```diff
@@ -10383,6 +10383,78 @@ def test_added_transforms_are_in_eval_mode():
...
+
+    class PatchedRenameTransform(RenameTransform):  # type: ignore[misc]
+        """
+        Fixes a bug in the RenameTransform due to modifying the input_spec of the `base_env` to be transformed.
+        This is fixed by adding a clone to break stateful modifications to proapagate to the `base_env`.
+        """
+
+        def transform_input_spec(self, input_spec: Composite) -> Composite:
+            input_spec = input_spec.clone()
+            return super().transform_input_spec(input_spec)
```
LCarmi, line 10438: @vmoens This class is not needed by this test anymore

vmoens: You're right, removed in #3079

PR #3080, opened 2025-07-16
[BugFix] Fix cross-entropy log-prob for batched input

vmoens:


louisfaury:
Would be great if we can have this in a minor too 🙏

vmoens:
It's in the [release/0.9.2 branch](https://github.com/pytorch/rl/commits/release/0.9.2/) and I'll work on shipping the binaries as soon as my plane lands

louisfaury:
hero!

PR #3084, opened 2025-07-18
[BugFix] Fix Binary reshaping

LCarmi:
## Description

Fixes a bug introduced by https://github.com/pytorch/rl/pull/3077 where shapes for newly instantiated `Binary` specs are incompatible with the value of `n` passed.

In particular the following pattern is used in `unsqueeze` and other shape-modifying operations:
```python
def unsqueeze(self, dim: int):
        shape = _unsqueezed_shape(self.shape, dim)
        return self.__class__(
            n=self.shape[-1] if len(self.shape) > 0 else None,
            shape=shape,
            device=self.device,
            dtype=self.dtype,
        )
```
However, this is wrong because we should extract `n` from the new `shape` rather than from the current `self.shape`.

We fix by switching to check on the new `shape`: 
```python
def unsqueeze(self, dim: int):
        shape = _unsqueezed_shape(self.shape, dim)
        return self.__class__(
            n=shape[-1] if len(shape) > 0 else None,
            shape=shape,
            device=self.device,
            dtype=self.dtype,
        )
```

## Motivation and Context

Without this change, we have value errors raised due to shape mismatches:

```python
import pytest
import torch
from torchrl.data import Binary


def test__Binary__if_empty_cannot_expand() -> None:
    spec = Binary(shape=())
    new_shape = (3, 3)

    with pytest.raises(ValueError, match="The last value of the shape must match 'n'"):
        spec.expand(new_shape)
```

NOTE: I have tried to include a test to check for this, but I don't know if that is the correct place and scope.

## Types of changes

## Checklist

- [X] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide

### DIFF
--- a/test/test_specs.py
+++ b/test/test_specs.py
@@ -3913,6 +3913,7 @@ class TestDynamicSpec:
         disc = Categorical(shape=(-1, 1, 2), n=4)
         moneh = MultiOneHot(shape=(-1, 1, 2, 7), nvec=[3, 4])
         mdisc = MultiCategorical(shape=(-1, 1, 2, 2), nvec=[3, 4])
+        binary = Binary(shape=(-1, 1, 2))
 
         spec = Composite(
             unb=unb,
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM! Thanks (and sorry!)

PR #3099, opened 2025-07-28
[BE] Avoid warning users who don't care about PRB

vmoens:


vmoens:
Tested locally and minor cosmetic changes - merging

PR #3101, opened 2025-07-30
Fix composite entropy for nested keys

juandelos:
## Description

Updated the function _weighted_loss_entropy to enable setting the entropy coefficient for every head for different nested keys definitions. Now it uses the nested keys and goes through the leaves. If the entropy coefficient is defined for the leave, it uses that. If not, it goes one step up until it finds a key that is defined in the entropy coefficient map.

## Motivation and Context

Before, the function had problems when we had nested keys of the type:
- agents
  - agent1
    - action
  - agent2
    - action
This would result in a problem as it would loop through the keys and try to find the name agents in the entropy coefficient map.

Also, it was not possible to define different entropy coefficients for the same agent, but different action heads like in this example:
- agent1
  - action1
  - action2
  - 

## Types of changes

## Checklist


torchrl/objectives/ppo.py
```diff
@@ -351,7 +351,7 @@ def __init__(
         *,
         entropy_bonus: bool = True,
         samples_mc_entropy: int = 1,
-        entropy_coeff: float | Mapping[str, float] | None = None,
+        entropy_coeff: float | Mapping[str | tuple | list, float] | None = None,
```
felixsittenauer: Nit: You could consider using `NestedKey` from tensordict here: https://github.com/pytorch/tensordict/blob/875d2fe9f9dc83563a74af7bafc83b75df6fa906/tensordict/_nestedkey.pyi#L9

torchrl/objectives/ppo.py
```diff
@@ -918,15 +919,37 @@ def _weighted_loss_entropy(
...
+        coeff = 0
+        for head_name, entropy_head in entropy.items(
+            include_nested=True, leaves_only=True
+        ):
+            if isinstance(head_name, str):
+                head_name = (head_name,)
+            for i, (head_name_from_map, _coeff) in enumerate(
+                self._entropy_coeff_map.items()
+            ):
+                # Check if distinct head name inisde tuple of nested dict
```
felixsittenauer: Nit: Typo
```suggestion
                # Check if distinct head name inside tuple of nested dict
```

torchrl/objectives/ppo.py
```diff
@@ -918,15 +919,37 @@ def _weighted_loss_entropy(
...
-                coeff = self._entropy_coeff_map[head_name]
-            except KeyError as exc:
-                raise KeyError(f"Missing entropy coeff for head '{head_name}'") from exc
+        coeff = 0
+        for head_name, entropy_head in entropy.items(
+            include_nested=True, leaves_only=True
+        ):
+            if isinstance(head_name, str):
+                head_name = (head_name,)
+            for i, (head_name_from_map, _coeff) in enumerate(
```
felixsittenauer: Nit:
```suggestion
            for i, (head_name_from_map, coeff_from_map) in enumerate(
```

torchrl/objectives/ppo.py
```diff
@@ -918,15 +919,37 @@ def _weighted_loss_entropy(
...
+            ):
+                # Check if distinct head name inisde tuple of nested dict
+                if head_name_from_map in head_name:
+                    coeff = _coeff
+                    break
+                # Check if path of head fully or partially in nested dict
+                if any(
+                    head_name_from_map == head_name[i : i + len(head_name_from_map)]
+                    for i in range(len(head_name) - len(head_name_from_map) + 1)
+                ):
```
felixsittenauer: Unsure about this part, wouldn't we always want to define the full nested path? Could you add a small explanation why this is needed?

juandelos: I added this in case we have something like 
- agents
  - agent1 
    - action1
      - action_entropy
  -  agent2
  .
  .
  .

If we specify the entropy_coeff_map as 
('agent1', 'action1'): value
it wouldn't find it otherwise

amalvaidya: It might be a little more verbose, but I think I agree with Felix. Fully defining the nested path is the safest way to do this.

amalvaidya:
>  If the entropy coefficient is defined for the leave, it uses that. If not, it goes one step up until it finds a key that is defined in the entropy coefficient map.

I think that overall this logic probably isn't necessary. I think it's okay for the user to always have to provide the full, nested key.

vmoens:
@juandelos thanks for this PR! Let me know when I should review it.
Can you sign the CLA when ready?

torchrl/objectives/ppo.py
```diff
@@ -918,15 +919,37 @@ def _weighted_loss_entropy(
             return -self.entropy_coeff * entropy
```
louisfaury, line 918: General comment: I feel like this method could use a bit more inline comments to guide the reader.

torchrl/objectives/ppo.py
louisfaury: I feel the same. This is adding a bunch of unnecessary complexity, and defining the coefficient for each leaf is still the safest. 

With that in mind, I believe we can simplify this method quite a lot (ie check before hand that they keys match and then simply go through them linearly).

juandelos:
@louisfaury you mean defining the whole nested key, right?

torchrl/objectives/ppo.py
```diff
@@ -351,7 +351,7 @@ def __init__(
         *,
         entropy_bonus: bool = True,
         samples_mc_entropy: int = 1,
-        entropy_coeff: float | Mapping[str, float] | None = None,
+        entropy_coeff: float | NestedKey | None = None,
```
vmoens: is it
```suggestion
        entropy_coeff: float | Mapping[NestedKey, float] | None = None,
```
otherwise it's a bc-breaking change and we need to move smoothly

torchrl/objectives/ppo.py
vmoens: Yep agree on that
I can give it a shot

torchrl/objectives/ppo.py
```diff
@@ -1075,7 +1079,7 @@ def __init__(
         clip_epsilon: float = 0.2,
         entropy_bonus: bool = True,
         samples_mc_entropy: int = 1,
-        entropy_coeff: float | Mapping[str, float] | None = None,
+        entropy_coeff: float | Mapping[str | tuple | list, float] | None = None,
```
vmoens: @louisfaury `NestedKey` here too right?

```suggestion
        entropy_coeff: float | Mapping[NestedKey | tuple | list, float] | None = None,
```

torchrl/objectives/ppo.py
```diff
@@ -1369,7 +1373,7 @@ def __init__(
         samples_mc_kl: int = 1,
         entropy_bonus: bool = True,
         samples_mc_entropy: int = 1,
-        entropy_coeff: float | Mapping[str, float] | None = None,
+        entropy_coeff: float | Mapping[str | tuple | list, float] | None = None,
```
vmoens: ditto

vmoens:
Follow-up: #3109

Issue #3105, opened 2025-07-31
[BUG] Download of categorical data in MinariExperienceReplay is completely broken


marcosgalleterobbva:
## Describe the bug

When downloading datasets from Minari that have categorical data (like missions), the download of the dataset is broken, as every step in the dataset will get assigned the array of categorical values for the first episode.

This error originates from treating the Minari categorical data as NonTensorData instead of NonTensorStack.

## To Reproduce

Take this code, for example:

```python
from torchrl.data.datasets.minari_data import MinariExperienceReplay

BATCH_SIZE = 1      # Small batch size to minimize processing
SAVE_ROOT = None    # Set to your desired save location, or None for default


def download_minari_datasets(dataset_id):
    a = MinariExperienceReplay(
        dataset_id=dataset_id,
        batch_size=BATCH_SIZE,
        root=SAVE_ROOT,
    )
    print(f"✓ Successfully downloaded {dataset_id}")


if __name__ == "__main__":
    download_minari_datasets("minigrid/BabyAI-Pickup/optimal-v0")

```
And now watch very closely what happens when we ask for mission values across random steps in random episodes:

```python
In [3]: a[0][('observation', 'mission')][:5], len(a[0][('observation', 'mission')])
Out[3]: 
(array([b'pick up a blue box', b'pick up a blue box',
        b'pick up a blue box', b'pick up a blue box',
        b'pick up a blue box'], dtype=object),
 109)

In [4]: a[200][('observation', 'mission')][:5], len(a[0][('observation', 'mission')])
Out[4]: 
(array([b'pick up a blue box', b'pick up a blue box',
        b'pick up a blue box', b'pick up a blue box',
        b'pick up a blue box'], dtype=object),
 109)

In [5]: a[5000][('observation', 'mission')][:5], len(a[0][('observation', 'mission')])
Out[5]: 
(array([b'pick up a blue box', b'pick up a blue box',
        b'pick up a blue box', b'pick up a blue box',
        b'pick up a blue box'], dtype=object),
 109)
```
It is the same array copied over and over again!

## Expected behavior

Minari datasets that have missions clearly need this dataset to be able to function. And the dataset clearly is intended to be used in a stacked manner.

## Additional context

I opened one PR and one issue in pytorch/tensordict incorrectly handling this issue:
https://github.com/pytorch/tensordict/issues/1399
https://github.com/pytorch/tensordict/pull/1400

[...]

PR #3106, opened 2025-07-31
Bugfix/wrong minari download first element

marcosgalleterobbva:
## Description

PersistentTensorDict.from_h5(h5_path) is the function that loads the h5 data and creates the first TensorDict version of the dataset. I have introduced the neccesary code for the _download_and_preproc function to handle all the substitution of NonTensorData into NonTensorStack. That is the way Minari uses the 'mission' categorical values.

I have introduced a fundamental change in the way MinariExperienceReplay works:

```python
h5_data = PersistentTensorDict.from_h5(h5_path)
h5_data = h5_data.to_tensordict()
```

I think this change brings the tensordict to memory instead of remaining lazy, but it was neccesary for the patching to take place. Just let me know if we can handle this in a different way.

## Motivation and Context

It solves the following issue https://github.com/pytorch/rl/issues/3105

With these changes, and using the example introduced in the issue, we can succesfully retrieve the correct mission for each step and see that it changes throught episodes and steps.

```python
In [3]: a[0][('observation', 'mission')]
Out[3]: b'pick up a blue box'

In [4]: a[200][('observation', 'mission')]
Out[4]: b'pick up the purple key'

In [5]: a[5000][('observation', 'mission')]
Out[5]: b'pick up a green key'
```

## Types of changes

## Checklist


torchrl/data/datasets/minari_data.py
```diff
@@ -281,6 +283,7 @@ def _download_and_preproc(self):
                         f"loading dataset from local Minari cache at {h5_path}"
                     )
                     h5_data = PersistentTensorDict.from_h5(h5_path)
+                    h5_data = h5_data.to_tensordict()
```
vmoens, line 294: do we need this? It's a bit expensive so if we can avoid it it's better (under the hood it copies the entire dataset in memory)

torchrl/data/datasets/minari_data.py
```diff
@@ -539,3 +565,51 @@ def _patch_info(info_td):
     if not source.is_empty():
         val_td_sel.update(source, update_batch_size=True)
     return val_td_sel
+
+
+def patch_nontensor_data_to_stack(tensordict: TensorDictBase):
```
vmoens: Make private? I don't think we want to commit that we'll keep these for eternity

torchrl/data/datasets/minari_data.py
```diff
@@ -539,3 +565,51 @@ def _patch_info(info_td):
...
+        if isinstance(val, TensorDictBase):
+            patch_nontensor_data_to_stack(val)  # in-place recursive
+        elif isinstance(val, NonTensorData):
+            data_list = list(val.data)
+            with set_list_to_stack(True):
+                tensordict[key] = data_list
+    return tensordict
+
+
+def extract_nontensor_fields(td: TensorDictBase, recursive: bool = False) -> TensorDict:
```
vmoens: ditto

torchrl/data/datasets/minari_data.py
```diff
@@ -539,3 +565,51 @@ def _patch_info(info_td):
...
+            extracted[key] = val
+            td.del_(key)
+        elif recursive and isinstance(val, TensorDictBase):
+            nested = extract_nontensor_fields(val, recursive=True)
+            if len(nested) > 0:
+                extracted[key] = nested
+    return TensorDict(extracted, batch_size=td.batch_size)
+
+
+def preallocate_nontensor_fields(td_data: TensorDictBase, example: TensorDictBase, total_steps: int, name_map: dict):
```
vmoens: ditto

torchrl/data/datasets/minari_data.py
```diff
@@ -539,3 +565,51 @@ def _patch_info(info_td):
     if not source.is_empty():
         val_td_sel.update(source, update_batch_size=True)
     return val_td_sel
+
+
+def patch_nontensor_data_to_stack(tensordict: TensorDictBase):
+    """Recursively replaces all NonTensorData fields in the TensorDict with NonTensorStack."""
+    for key in list(tensordict.keys()):
+        val = tensordict.get(key)
```
vmoens: use `items()`?

torchrl/data/datasets/minari_data.py
```diff
@@ -539,3 +565,51 @@ def _patch_info(info_td):
...
+        if isinstance(val, TensorDictBase):
+            patch_nontensor_data_to_stack(val)  # in-place recursive
+        elif isinstance(val, NonTensorData):
+            data_list = list(val.data)
```
vmoens, line 599: are we sure this makes sense? Can val be a `str`? Perhaps we should check what type val.data has?

torchrl/data/datasets/minari_data.py
```diff
@@ -539,3 +565,51 @@ def _patch_info(info_td):
...
+            with set_list_to_stack(True):
+                tensordict[key] = data_list
+    return tensordict
+
+
+def extract_nontensor_fields(td: TensorDictBase, recursive: bool = False) -> TensorDict:
+    extracted = {}
+    for key in list(td.keys()):
+        val = td.get(key)
+        if isinstance(val, (NonTensorData, NonTensorStack)):
```
vmoens: you could do `is_non_tensor(val)`

torchrl/data/datasets/minari_data.py
```diff
@@ -539,3 +565,51 @@ def _patch_info(info_td):
...
+            extracted[key] = val
+            td.del_(key)
```
vmoens:
```suggestion
            del td[key]
```

torchrl/data/datasets/minari_data.py
```diff
@@ -539,3 +565,51 @@ def _patch_info(info_td):
...
+        elif recursive and isinstance(val, TensorDictBase):
```
vmoens:
```suggestion
        elif recursive and is_tensor_collection(val):
```

torchrl/data/datasets/minari_data.py
```diff
@@ -539,3 +565,51 @@ def _patch_info(info_td):
...
+
+def preallocate_nontensor_fields(td_data: TensorDictBase, example: TensorDictBase, total_steps: int, name_map: dict):
+    """Preallocates NonTensorStack fields in td_data based on an example TensorDict, applying key remapping."""
+    with set_list_to_stack(True):
+        def _recurse(src_td: TensorDictBase, dst_td: TensorDictBase, prefix=()):
+            for key, val in src_td.items():
+                mapped_key = name_map.get(key, key)
+                full_dst_key = prefix + (mapped_key,)
+
+                if isinstance(val, NonTensorData):
```
vmoens:
```suggestion
                if is_non_tensor(val):
```

torchrl/data/datasets/minari_data.py
```diff
@@ -539,3 +565,51 @@ def _patch_info(info_td):
...
+                    dummy_val = b"" if isinstance(val.data[0], bytes) else ""
+                    dummy_stack = TensorDict({mapped_key: dummy_val}).expand(total_steps)[mapped_key]
```
vmoens:
```suggestion
                    dummy_val = b"" if isinstance(val.data[0], bytes) else ""
                    dummy_stack = NonTensorStack(*[total_steps for _ in range(total_steps)])
```

vmoens (commented):
Thanks!

Added some quick comments here and there.

A test would be nice! (you can just add a line in an existing test to check that the bug is resolved?)

torchrl/data/datasets/minari_data.py
marcosgalleterobbva: There is nothing I would love more than getting rid of that line. The method to change from NonTensorData to NonTensorStack is basically:

```python
with set_list_to_stack(True):
    tensordict[key] = data_list
```
Unfortunately, if we don't get the h5_data into memory, we face this error upon rewriting each NonTensorData key.

OSError: Can't synchronously write data (no write intent on file)

If anyone knows how to avoid this, I would love to get this thing fixed in a better way.

torchrl/data/datasets/minari_data.py
marcosgalleterobbva: mmmm I get this error when using .items()

```python
  File "/Users/O000142/Projects/rl/torchrl/data/datasets/minari_data.py", line 585, in _extract_nontensor_fields
    for key, val in tensordict.items():
RuntimeError: dictionary changed size during iteration
```

This is due to the fact that we are deleting keys in the dataset as we are iterating them. Looks better with `items()`, but sadly doesn't work

torchrl/data/datasets/minari_data.py
marcosgalleterobbva: As far as I know, `val.data` is a `np.array` of length equal to the number of steps in the episode. In the case of 'mission' keys, it is an array that contains what mission the agent has at each step.

torchrl/data/datasets/minari_data.py
marcosgalleterobbva: These functions look similar but perform different functions. One preallocates keys in a tensordict, the other deletes NonTensorData keys and another one for transforming NonTensorData into NonTensorStack. But they all traverse the tensordict keys. I can try to refactor them a bit more later.

marcosgalleterobbva:
I have applied most of your suggestions (except for the ditto suggestions, those take longer). I have also added a test.

I am also having a problem that I might need help with.

This is the code that I am using to test if all these changes work:

```python
from torchrl.data.datasets.minari_data import MinariExperienceReplay

BATCH_SIZE = 1
SAVE_ROOT = None


def download_minari_datasets(dataset_id):
    a = MinariExperienceReplay(
        dataset_id=dataset_id,
        batch_size=BATCH_SIZE,
        root=SAVE_ROOT,
    )
    print(f"✓ Successfully downloaded {dataset_id}")
    print(a[210][('observation', 'mission')])
    print(a[1210][('observation', 'mission')])
    print(a[2210][('observation', 'mission')])
    print(a[3210][('observation', 'mission')])

if __name__ == "__main__":
    download_minari_datasets("minigrid/BabyAI-Pickup/optimal-v0")
```

If you get single, different missions for each step, then it looks probably alright.
If you have an array of missions and always the same array, repeated over all steps, then it is certainly wrong.
And everything works fine the first time that this code is executed (when using the _download_and_preproc function).

The problem is that the second time you run this code, now that the dataset has already been downloaded, the experience replay keeps loading the same wrong missions from before. Long, repeated missions at each step. It happens as we use the _load function.

I have absolutely no idea what might be the reason for this behaviour.

marcosgalleterobbva:
Hi, @vmoens ! I hope you had (or hopefully still having 😁 ) a great summer.

I managed to solve the last problem with this PR. Now the PR loads the categorical data both on first download and from disk read. I added the missing test and added your suggestions. When your suggestions were not added, I justified it in the comments above.

Let me know what you think and if this PR is merge ready. Thanks!

marcosgalleterobbva:
Hi! Is there any problem with this PR that I should address? Just wanted to check if everything is going fine. Thanks!

PR #3181, opened 2025-10-02
Further fix

matteobettini:
Let's wait for the tests to pass to see if this is the last


vmoens:
Let me revert the PR, it'll be simpler

PR #3199, opened 2025-10-16
[Feature] Enable storing rollouts on a different device

### LINKED ISSUE #3177: [Feature Request] `thrl_envs.EnvBase.rollout` save to different device
## Motivation
For environments with high-dimensional observation (e.g. camera feed of autonomous vehicles) and/or long rollout times (i.e. with large `max_steps`) running on gpu, using `thrl_envs.EnvBase.rollout` to collect trajectories of the policy interacting with the environment could cause vram to explode over time. From my rough understanding, all the intermediate `tensordict` returned by `thrl_envs.EnvBase.step` would all be stored on gpu. Suppose I am currently at $t=99$, the `thrl_envs.EnvBase.step` method only depends on the `tensordict` returned at $t=98$, whereas the rest of the `tensordict`s returned prior to $t=98$ are sitting on vram but not actively participating in any kind of arithmetic operations.

## Solution
When collecting rollout trajectories with `thrl_envs.EnvBase.rollout`, is it possible to move intermediate `tensordict`s returned by `thrl_envs.EnvBase.step` to cpu so that precious gpu vram can be freed for other arithmetic computations that actually requires gpu? That is, perhaps add an argument `trajectory_save_device` to `thrl_envs.EnvBase.rollout` so that its return is on cpu if `trajectory_save_device="cpu"`.

## Alternatives
It is possible, but a pain, to use for loops to replicate the behavior of `thrl_envs.EnvBase.rollout` with move and save on cpu.

## Checklist

- [ *] I have checked that there is no similar issue in the repo (**required**)

Xmaster6y:
## Description

Added a `storing_device` argument to `env.rollout` to save to a different device, like in collectors.

## Motivation and Context

Fix #3177 

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -3375,7 +3375,9 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
             "policy": policy,
             "policy_device": policy_device,
             "env_device": env_device,
-            "storing_device": storing_device,
+            "storing_device": None
+            if storing_device is None
+            else torch.device(storing_device),
             "callback": callback,
         }
         if break_when_any_done or break_when_all_done:


torchrl/envs/common.py
```diff
@@ -3531,7 +3536,7 @@ def _rollout_stop_early(
                 else:
                     tensordict.clear_device_()
             tensordict = self.step(tensordict)
-            td_append = tensordict.copy()
```
vmoens, line 3534: maybe only if `storing_device` is defined to avoid the overhead?

torchrl/envs/common.py
```diff
@@ -3614,7 +3620,7 @@ def _rollout_nonstop(
                 tensordict = self.step(tensordict_)
             else:
                 tensordict, tensordict_ = self.step_and_maybe_reset(tensordict_)
-            tensordicts.append(tensordict)
```
vmoens, line 3617: ditto

Xmaster6y:
What overhead? `.to(None)` does nothing, same for same device `to`no?

vmoens:
> What overhead? `.to(None)` does nothing, same for same device `to`no?

A function call is a function call, I'd prefer if we were avoiding any call to `to` if not necessary (for clarity)

Xmaster6y:
Gotcha!

torchrl/envs/common.py
```diff
@@ -3531,7 +3536,10 @@ def _rollout_stop_early(
                 else:
                     tensordict.clear_device_()
             tensordict = self.step(tensordict)
-            td_append = tensordict.copy()
+            if storing_device is None or tensordict.device == storing_device:
```
vmoens, line 3541: This is a bit loose, 
```python
>>> torch.device("cpu") == "cpu"
False
```

torchrl/envs/common.py
```diff
@@ -3614,7 +3623,10 @@ def _rollout_nonstop(
                 tensordict = self.step(tensordict_)
             else:
                 tensordict, tensordict_ = self.step_and_maybe_reset(tensordict_)
-            tensordicts.append(tensordict)
+            if storing_device is None or tensordict.device == storing_device:
```
vmoens, line 3628: ditto

torchrl/envs/common.py
Xmaster6y: Oh right, my bad

PR #3243, opened 2025-11-28
[Feature] Ensure `MultiSyncDataCollectors` returns data ordered by worker id

LCarmi:
## Description

Changes the `self.buffers` data structure in `MultiSyncDataCollectors` to a list in order to ensure that the returned batch respects the `worker_id` order.

## Motivation and Context

The main motivation behind this change is to enable an id-based retrieval of experience when sampling from multiple, possibly different environments/policies. More in detail, if the user specifies a list of environments/policies, this change guarantees that `batch[i]` corresponds to the experience sampled from the `i`th environment and policy.

This was not previously available because the `self.buffer` was previously a dict, and different processes return experience possibly out-of-order due to having unpredictable latency. Since a dict is iterated in insertion order, the relative order between worker ids was not maintaned.

Note: if preemption is enabled, then the results will just maintain relative order, since we are injecting "bubbles" in our returned batch.

## Types of changes

## Checklist

- [~] My change requires a change to the documentation.
- [~] I have updated the documentation accordingly.

### DIFF
--- a/torchrl/collectors/_multi_sync.py
+++ b/torchrl/collectors/_multi_sync.py
@@ -333,7 +333,7 @@ class MultiSyncDataCollector(_MultiDataCollector):
             if self.replay_buffer is not None:
                 yield
                 self._frames += sum(
-                    self.frames_per_batch_worker(worker_idx)
+                    self.frames_per_batch_worker(worker_idx=worker_idx)
                     for worker_idx in range(self.num_workers)
                 )
                 continue


torchrl/collectors/collectors.py
```diff
@@ -3844,16 +3842,16 @@ def iterator(self) -> Iterator[TensorDictBase]:
                 if preempt:
                     # mask buffers if cat, and create a mask if stack
                     if cat_results != "stack":
-                        buffers = {}
-                        for worker_idx, buffer in self.buffers.items():
+                        buffers = [None] * self.num_workers
+                        for idx, buffer in enumerate(filter(None.__ne__, self.buffers)):
```
LCarmi: We need to filter out buffers who did not return their experience: I use the `enumerate(filter(None.__ne__, self.buffers))` idiom to make this compact and hopefully readable; I'm open to better ideas

torchrl/collectors/collectors.py
```diff
@@ -3863,11 +3861,6 @@ def iterator(self) -> Iterator[TensorDictBase]:
                 else:
                     buffers = self.buffers
 
-                # Skip frame counting if this worker didn't send data this iteration
-                # (happens when reusing buffers or on first iteration with some workers)
-                if idx not in buffers:
-                    continue
-
```
LCarmi: I am puzzled by this code, and I miss where it could happen that `idx` is defined but the related buffer be `None`

An equivalent code here would be `if buffers[idx] is None: continue`

torchrl/collectors/collectors.py
vmoens: I think it's ok

vmoens: but you define idx which was defined earlier (LoC 3829 or 3840). It should be worker_idx I believe
See my comment below

torchrl/collectors/collectors.py
vmoens: This happens during preemption: if we say that we're ok with 80% of the data it could be that we don't have data for one of the workers and we just return whatever we have at this stage.

torchrl/collectors/collectors.py
LCarmi: Thanks, fixed. Also, following some internal comments I am switching to `lambda x: x is None` for readability

torchrl/collectors/collectors.py
LCarmi: I see, thanks for the comment. I originally missed that preemption makes everyone return, possibly returning (worker_id, None) as data (I thought that when preemption hit, no message was added to the queue).

I have now re-added this with the same semantic as before.

torchrl/collectors/_multi_sync.py
```diff
@@ -320,7 +321,7 @@ def iterator(self) -> Iterator[TensorDictBase]:
 
                 # Skip frame counting if this worker didn't send data this iteration
                 # (happens when reusing buffers or on first iteration with some workers)
-                if idx not in buffers:
+                if self.buffers[idx] is None:
                     continue
```
vmoens, line 326: (sorry this should be linked to the for loop declaration 🤣 )

Using `enumerate(filter(lambda x: x is not None, self.buffers))` loses the original worker indices.

The `worker_idx` from `enumerate()` corresponds to the position in the filtered list, not the original worker index in `self.buffers`.

**Example:**
If `self.buffers = [None, buffer1, None, buffer2]`:
- `filter(...)` gives `[buffer1, buffer2]`
- `enumerate(...)` gives `(0, buffer1), (1, buffer2)`
- This sets `buffers[0] = buffer1[..., valid]` and `buffers[1] = buffer2[..., valid]`
- But we actually need `buffers[1] = buffer1[..., valid]` and `buffers[3] = buffer2[..., valid]`

I think you can get away from this by doing the filter above the enumerate and not the enumerate above the filter, if that makes sense, or just write a plain loop with a `continue` if `None`

vmoens (commented):
Hi! I've ported this PR to the new file structure (big messy fat PR I just merged) and I noticed a potential bug in the preempt case when `cat_results != "stack"`.

torchrl/collectors/_multi_sync.py
LCarmi: Thanks, I have fixed it with a continue as you suggested

test/test_collector.py
```diff
@@ -1712,6 +1712,186 @@ def env_fn():
...
+
+        def _set_seed(self, seed: int | None) -> int | None:
+            """Set the seed for reproducibility."""
+            if seed is not None:
+                torch.manual_seed(seed)
+            return seed
+
+    @pytest.mark.parametrize("num_envs,n_steps", [(8, 5)])
+    @pytest.mark.parametrize("with_preempt", [False, True])
+    @pytest.mark.parametrize("cat_results", ["stack", -1])
```
LCarmi, line 1810: I have made the test suite much wider, including the `preemption`/`no-preemption` case and the `stack`/`cat`.  On my machine, the preemption ones last ~10s while the non-preemt ones last ~4s. I am open to try to make them shorted (by tuning the `time.sleep` if you deem so.

PR #3236, opened 2025-11-17
[BugFix] RSSMRollout not advancing state/belief across time steps in Dreamer

cmdout:
## Description

This PR fixes the RSSM rollout state propagation in `RSSMRollout.forward`.  
At each timestep, after computing the prior and posterior, we now:

- shift `("next", *)` keys to current using `step_mdp(..., keep_other=True)`, and
- merge the next step inputs before continuing the rollout.

Concretely:
- add `step_mdp` import, and
- call `step_mdp` inside the time loop before selecting `in_keys` and updating with the next step’s values.

This ensures `("next", "state")` and `("next", "belief")` are correctly advanced to `state` and `belief` for the subsequent step.

## Motivation and Context

Previously, the rollout did not propagate `state` and `belief` from `("next", ...)` back to the current keys, causing the loop to reuse stale `state/belief`. This broke the intended posterior-to-prior chaining in Dreamer’s RSSM and led to incorrect latent rollouts (and potentially training instability and degraded reconstruction/reward predictions).

The fix aligns the implementation with Dreamer’s design and the world model wiring used elsewhere (e.g., in `dreamer_utils._dreamer_make_world_model`), where the prior for step t+1 must consume the posterior outputs of step t.

If there is an open issue tracking this bug, please link it here.  
Example: `close #XXXX`

- [ ] I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds core functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)
- [ ] Example (update in the folder of examples)

## Checklist

- [ ] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] My change requires a change to the documentation.
- [ ] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.

PR #3266, opened 2025-12-30
[BugFix] Fix old pytorch dependencies

vmoens:


vmoens:
@cmdout I reverted #3236 because it was breaking things. Could you explain further why that was needed? Happy to chat about this with you!

PR #3337, opened 2026-01-15
[Feature] Added num_envs parameter in DMControlEnv to call multiple environments in one call

ParamThakkar123:
## Description

Added a `num_envs` option in [DMControlEnv](https://github.com/pytorch/rl/blob/69d44f5cf4bf84eab0f21b0eea98112651f7f9a1/torchrl/envs/libs/dm_control.py#L328) to create a parallel env in just one call, e.g. DMControlEnv(name, task, num_envs=4) would run 4 parallel envs of the name, task dmc env.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -2847,8 +2847,6 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
             reward_threshold=reward_threshold,
             nondeterministic=nondeterministic,
             max_episode_steps=max_episode_steps,
-            order_enforce=order_enforce,
-            disable_env_checker=disable_env_checker,
         )
 
     @implement_for("gymnasium", None, "1.0.0", class_method=True)


vmoens (commented):
I like this! But we should be able to pass some more args from ParallelEnv here. Do you have ideas on how to do that without overloading the docstrings too much?
What about other envs (gym etc) should we do one PR per env?
Can we add some tests?

torchrl/envs/libs/dm_control.py
```diff
@@ -407,16 +408,44 @@ class DMControlEnv(DMControlWrapper):
...
+        else:
+            return super().__new__(cls)
+
+    def __init__(self, env_name, task_name, num_envs=1, **kwargs):
         if not _has_dmc:
             raise ImportError(
                 "dm_control python package was not found. Please install this dependency."
             )
+
+        self.num_envs = num_envs
```
vmoens: why do we store num_envs within the env? It should not necessarily be aware of how many of its type there are

ParamThakkar123: Okay I did that because I wanted to have `num_envs` as a Parameter of DMControlEnv but now if we want to have it in ParallelEnv itself. I will remove it from here

ParamThakkar123:
> I like this! But we should be able to pass some more args from ParallelEnv here. Do you have ideas on how to do that without overloading the docstrings too much? What about other envs (gym etc) should we do one PR per env? Can we add some tests?

Sounds good. We can do one PR per Env. Would love to work on this! I don't have any ideas currently but would definitely look for some ideas on how we can have it that way. It's definitely possible and doable and a great idea indeed!

torchrl/envs/libs/dm_control.py
ParamThakkar123: Yeah. Got an idea. Can we add an optional namespaced dictionary something like `parallel_kwargs` and Pass it to `ParallelEnv`?

vmoens: Options:
1. Delayed init 
  ```python
  env = DMControlEnv(name, task, num_envs=3)
  # lasily init, so as of now parallel env has not started and we can change the params
  env.bached(use_buffers=False) # or any other kwarg -- can also be named parallelize or anything. Easy to document
  # now any op other than batched will start the parallel env
  env.full_observation_spec 
  ```
2. Just pass the `**parallel_kwargs` (not super clean but works). Can use MetaClasses like we do for all envs already to capture that 

3. `@classmethod` factory 
  ```python
  env = DMControlEnv.make_parallel(
      "cheetah", "run",
      num_envs=4,
      from_pixels=True,
      # ParallelEnv kwargs
      mp_start_method="spawn",
      daemon=True,
  )
  ```

**Important** Now that I think about it rather than hacking `__new__` maybe the better way to do this would be with the metaclass, wdyt?

ParamThakkar123: hmm. Yes. Adding a metaclass would be quite good. Now I realize that would be actually slightly cleaner than implementing `__new__` inside this

ParamThakkar123: I think I should go ahead an implement this using metaclasses. That would be great!

ParamThakkar123: or even the `@classmethod` factory seems a lot cleaner

ParamThakkar123:
@vmoens 

```python
class _DMControlMeta(_EnvPostInit):
    def __call__(cls, *args, num_envs: int | None = None, parallel_kwargs: dict | None = None, **kwargs):
        # determine num_envs from explicit kw, kwargs, or positional args
        if num_envs is None:
            if "num_envs" in kwargs:
                num_envs = kwargs.pop("num_envs")
            elif len(args) >= 3:
                num_envs = args[2]
            else:
                num_envs = 1
        if cls.__name__ == "DMControlEnv" and num_envs is not None and int(num_envs) > 1:
            from torchrl.envs import ParallelEnv, EnvCreator

            def make_env():
                env_name = args[0] if len(args) >= 1 else kwargs.get("env_name")
                task_name = args[1] if len(args) >= 2 else kwargs.get("task_name")
                inner_kwargs = kwargs.copy()
                return cls(env_name=env_name, task_name=task_name, num_envs=1, **inner_kwargs)

            return ParallelEnv(int(num_envs), EnvCreator(make_env), **(parallel_kwargs or {}))
        return super().__call__(*args, **kwargs)
```
I implemented this metaclass

vmoens:
MetaClass is good let's keep it! 
But I would remove the parallel_kwargs and used the delayed spec I mentioned earlier, can we do that instead? I feel that 
- it gives people the opportunity to use autocomplete
- it's easier to document
- no loss on the ease of use either because people will still need to look into what parallel_kwargs can contain
I would also implement an abstract class in `EnvBase` that has the same args / kwargs as ParallelEnv constructor has to tell the IDEs what to do with it

ParamThakkar123:
Sure ! I will make those updates right away!

vmoens:
@ParamThakkar123 I did [05a7b63](https://github.com/pytorch/rl/pull/3337/commits/05a7b635f69806f5f18b0abdcc9e5b07ffcbc869) with what I had in mind, can you check if it aligns with what you wanted too?

ParamThakkar123:
@vmoens Yeah. I think it aligns perfectly with what I wanted to do. I think I missed some parts in my implementation 😅

ParamThakkar123:
We don't actually require `make_parallel` method now, right ? as `configure_parallel` handles things ?

ParamThakkar123:
Also does it run without using the `main()` function ? Because in my initial implementation I was getting an error from `spawn` and `multiprocessing` where it mentioned that I can't run a ParallelEnv with multiple envs in one call without writing the main function and then doing : 

```python
if __name__ = "__main__":
        main()    
 ```

ParamThakkar123:
@vmoens Should I start extending to other environments too ? Like we discussed

PR #3372, opened 2026-01-22
[Test] Skip Llama test when tokenizer unavailable instead of xfail

vmoens:
* #3377
* #3376
* #3375
* #3374
* #3373
* __->__ #3372

The test_history_assistant_mask_llama test requires a Llama tokenizer
to properly handle Llama-specific special tokens for accurate assistant
masking. Since Llama models are gated on HuggingFace and cannot be
accessed in CI, change from pytest.xfail() to pytest.skip() with a
clear message explaining why.

This is cleaner than xfail because:
- It clearly explains the limitation
- It doesn't add noise to test results
- The test will run if the Llama tokenizer is available

The History API functionality is still tested via
test_history_assistant_mask_qwen which covers the same API.

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -1,0 +1,216 @@
+# Labels for TorchRL repository
+# Use with: gh label create "Label Name" --description "Description" --color "hex"
+# Or run: .github/scripts/create_labels.sh pytorch/rl
+
+# =============================================================================
+# Issue/PR Type Labels
+# =============================================================================
+- name: bug
+  description: "Something isn't working"
+  color: "d73a4a"
+
+- name: enhancement
+  description: "New feature or request"
+  color: "a2eeef"
+
+- name: documentation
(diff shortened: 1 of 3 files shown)


test/llm/test_data.py
```diff
@@ -419,15 +419,24 @@ def test_history_assistant_mask_qwen(self, test_case):
...
-            tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen2.5-0.5B-Instruct")
+            tokenizer = AutoTokenizer.from_pretrained(
+                "meta-llama/Llama-3.2-1B-Instruct"
+            )
         except Exception:
-            pytest.xfail("Could not load Llama tokenizer - model may be gated")
+            pytest.skip(
+                "Llama tokenizer not available (gated model). "
+                "The History API is tested via test_history_assistant_mask_qwen."
+            )
```
vmoens, line 440: removed in a followup PR

PR #3384, opened 2026-01-26
[Environments] Add FinancialRegimeEnv: A Vectorized Financial Environment

### LINKED ISSUE #3394: [Feature Request] Add a native, vectorized financial trading environment
## Motivation

Currently, TorchRL lacks a native environment specifically designed for financial time-series RL benchmarks. Users often have to wrap external libraries (like OpenAI Gym-based trading envs), which can be cumbersome and inefficient.

Currently, prototyping financial agents is difficult because:

- External environments often lack native TensorDict integration.

- It is difficult to leverage GPU acceleration for high-throughput sampling with existing external wrappers.

- There is no standardized benchmark for testing how agents handle regime shifts (e.g., transitioning from a bull market to a bear market).

## Solution

Adding a new environment, FinancialRegimeEnv, to the library.

This environment should:

- Simulate Dynamics: Use Geometric Brownian Motion (GBM) to model price movements realistically.

- Handle Regimes: Include logic to transition between Bull, Bear, and Sideways markets based on momentum and volatility.

- Be Vectorized: Support fully vectorized execution to allow for high-speed data generation on GPUs.

- Native Integration: Be built directly on top of TensorDict to ensure seamless compatibility with TorchRL’s collectors and trainers.

## Alternatives

- Wrapping External Gym Environments: This is the current workaround, but it introduces overhead and often breaks the vectorized data flow required for efficient training in TorchRL.

- Simple Random Walks: One could use basic random walk environments, but these fail to capture the complexity of market regimes (drifts and volatilities) necessary for robust financial RL benchmarking.

## Checklist

aneesh223:
# Description:

This PR adds FinancialRegimeEnv, a lightweight, vectorized environment for financial reinforcement learning. It simulates price movements using Geometric Brownian Motion with regime-switching dynamics (Bull, Bear, Sideways) derived from momentum and volatility logic.

# Motivation and Context

TorchRL currently lacks a native financial environment. This addition provides a standard benchmark for users testing RL agents on continuous time-series data with regime shifts. It is designed to be fully compatible with TensorDict and supports GPU acceleration.

# Types of changes

[ ] Bug fix (non-breaking change which fixes an issue)

[x] New feature (non-breaking change which adds core functionality)

[ ] Breaking change (fix or feature that would cause existing functionality to change)

[x] Documentation (update in the documentation)

[ ] Example (update in the folder of examples)

# Checklist

[x] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (required)

[ ] My change requires a change to the documentation.

[x] I have updated the tests accordingly (required for a bug fix or a new feature).

[x] I have updated the documentation accordingly.

Fixes #3394

### DIFF
--- a/test/test_custom_envs.py
+++ b/test/test_custom_envs.py
@@ -2,11 +2,17 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+from __future__ import annotations
 
+import random
+
+import pytest
 import torch
+from tensordict import TensorDict
 
+from torchrl.envs import check_env_specs, LLMHashingEnv, PendulumEnv, TicTacToeEnv
 from torchrl.envs.custom.trading import FinancialRegimeEnv
-from torchrl.envs.utils import check_env_specs
+from torchrl.testing import get_default_devices
 
 
 class TestCustomEnvs:
@@ -63,6 +69,8 @@ class TestCustomEnvs:
         r = env.rollout(20, break_when_any_done=False)
         assert r.shape == torch.Size((20,))
 
+    @pytest.mark.gpu
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
     def test_financial_env_device(self):
         """Test environment works on different devices."""
         # Test CPU (always available)
@@ -79,3 +87,70 @@ class TestCustomEnvs:
             env_cuda = FinancialRegimeEnv(device="cuda:0")
             assert env_cuda.device.type == "cuda"
             check_env_specs(env_cuda)
+
+
+class TestTicTacToeEnv:
+    def test_tictactoe_env(self):
+        torch.manual_seed(0)
+        env = TicTacToeEnv()
+        check_env_specs(env)
+        for _ in range(10):
+            r = env.rollout(10)
+            assert r.shape[-1] < 10
+            r = env.rollout(10, tensordict=TensorDict(batch_size=[5]))
+            assert r.shape[-1] < 10
+        r = env.rollout(
+            100, tensordict=TensorDict(batch_size=[5]), break_when_any_done=False
+        )
+        assert r.shape == (5, 100)
+
+    def test_tictactoe_env_single(self):
+        torch.manual_seed(0)
+        env = TicTacToeEnv(single_player=True)
+        check_env_specs(env)
+        for _ in range(10):
+            r = env.rollout(10)
+            assert r.shape[-1] < 6
+            r = env.rollout(10, tensordict=TensorDict(batch_size=[5]))
+            assert r.shape[-1] < 6
+        r = env.rollout(
+            100, tensordict=TensorDict(batch_size=[5]), break_when_any_done=False
+        )
+        assert r.shape == (5, 100)
+
+
+class TestPendulum:
+    @pytest.mark.parametrize("device", [None, *get_default_devices()])
+    def test_pendulum_env(self, device):
+        env = PendulumEnv(device=device)
+        assert env.device == device
+        check_env_specs(env)
+
+        for _ in range(10):
+            r = env.rollout(10)
+            assert r.shape == torch.Size((10,))
+            r = env.rollout(10, tensordict=TensorDict(batch_size=[5], device=device))
+            assert r.shape == torch.Size((5, 10))
+
+    def test_llm_hashing_env(self):
+        vocab_size = 5
+
+        class Tokenizer:
+            def __call__(self, obj):
+                return torch.randint(vocab_size, (len(obj.split(" ")),)).tolist()
+
+            def decode(self, obj):
+                words = ["apple", "banana", "cherry", "date", "elderberry"]
+                return " ".join(random.choice(words) for _ in obj)
+
+            def batch_decode(self, obj):
+                return [self.decode(_obj) for _obj in obj]
+
+            def encode(self, obj):
+                return self(obj)
+
+        tokenizer = Tokenizer()
+        env = LLMHashingEnv(tokenizer=tokenizer, vocab_size=vocab_size)
+        td = env.make_tensordict("some sentence")
+        assert isinstance(td, TensorDict)
+        env.check_env_specs(tensordict=td)
--- a/torchrl/envs/custom/trading.py
+++ b/torchrl/envs/custom/trading.py
@@ -317,6 +317,12 @@ class FinancialRegimeEnv(EnvBase):
 
         self.state_spec = self.observation_spec.clone()
 
+        self.done_spec = Composite(
+            done=Categorical(n=2, shape=(1,), dtype=torch.bool),
+            terminated=Categorical(n=2, shape=(1,), dtype=torch.bool),
+            truncated=Categorical(n=2, shape=(1,), dtype=torch.bool),
+        )
+
         self.action_spec = Categorical(
             n=3,
             shape=(1,),
--- a/test/test_envs.py
+++ b/test/test_envs.py
@@ -11,7 +11,6 @@ import gc
 import importlib
 import os
 import pickle
-import random
 import re
 import signal
 import threading
@@ -53,12 +52,9 @@ from torchrl.envs import (
     DoubleToFloat,
     EnvBase,
     EnvCreator,
-    LLMHashingEnv,
     ParallelEnv,
-    PendulumEnv,
     SerialEnv,
     set_gym_backend,
-    TicTacToeEnv,
 )
 from torchrl.envs.batched_envs import _stackable
 from torchrl.envs.gym_like import default_info_dict_reader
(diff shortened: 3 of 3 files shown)


torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+    This class provides tensor-based implementations of market regime detection,
+    volatility calculations, and adaptive risk parameters for use in RL environments.
+    """
+
+    def __init__(self, device: torch.device | None = None):
+        """Initialize the FinancialMarketDynamics class.
+
+        Args:
+            device: PyTorch device to use for computations (default: CPU)
+        """
```
vmoens: I like grouping these docstrings usually
Giving an example of usage there also goes a long way!

test/test_custom_envs.py
vmoens, line 1: Let's avoid making one file per class tested. If you don't feel like any of the existing files suits the purpose we can make a dedicated `test_custom_envs.py`

torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+        Returns:
+            Returns tensor with NaN values for initial periods
+        """
+        if periods >= prices.shape[-1]:
+            return torch.full_like(prices, float("nan"))
+
+        returns = torch.full_like(prices, float("nan"))
+        returns[..., periods:] = (
+            prices[..., periods:] - prices[..., :-periods]
+        ) / prices[..., :-periods]
```
vmoens: numerical underflow?

torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+        # Use 1D convolution for rolling mean
+        kernel = torch.ones(window_size, device=self.device) / window_size
+
+        # Reshape for conv1d: (batch_size, channels, sequence_length)
+        original_shape = tensor.shape
+        tensor_flat = tensor.view(-1, original_shape[-1]).unsqueeze(1)
+
+        # Apply convolution with padding
+        conv_result = torch.nn.functional.conv1d(
+            tensor_flat, kernel.view(1, 1, -1), padding=window_size - 1
```
vmoens: view can be dangerous and `reshape` will do a no-copy op if possible

torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+        """
+        if window_size > tensor.shape[-1]:
+            return torch.full_like(tensor, float("nan"))
+
+        # Use 1D convolution for rolling mean
+        kernel = torch.ones(window_size, device=self.device) / window_size
+
+        # Reshape for conv1d: (batch_size, channels, sequence_length)
+        original_shape = tensor.shape
+        tensor_flat = tensor.view(-1, original_shape[-1]).unsqueeze(1)
```
vmoens: ditto: let's use reshape?
Do we need unsqueeze + view, not just view/reshape?

torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+
+        # Apply convolution with padding
+        conv_result = torch.nn.functional.conv1d(
+            tensor_flat, kernel.view(1, 1, -1), padding=window_size - 1
+        )
+
+        # Remove extra padding and reshape back
+        conv_result = conv_result[:, 0, : original_shape[-1]]
+        result = conv_result.view(original_shape)
```
vmoens: ditto

torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+        drift = params["drift"]
+        transaction_cost = params["transaction_cost"]
+        max_steps = params["max_steps"]
+
+        # Generate next price using geometric Brownian motion with regime-aware parameters
+        current_price = price_history[..., -1]
+        dt = 1.0 / 252.0  # Daily time step
+
+        # Add some randomness for price generation
+        random_shock = torch.randn_like(current_price) * volatility * math.sqrt(dt)
```
vmoens: The `_step` method uses `torch.randn_like()` without a generator, making the environment non-reproducible. In `PendulumEnv`, `_step` is deterministic - only `_reset` uses the generator. If price generation needs randomness per step, you need to either:
- Pass the rng through the tensordict (complex)
- Store random seeds/noise in the state
- Make price evolution deterministic based on some state variable

torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+            window_size: Size of price history window (default: 50)
+        """
+        if td_params is None:
+            td_params = self.gen_params(device=device, window_size=window_size)
+
+        super().__init__(device=device)
+        self._make_spec(td_params)
+
+        # Initialize market dynamics
+        self.market_dynamics = FinancialMarketDynamics(device=self.device)
```
vmoens: The constructor initializes `self.market_dynamics = FinancialMarketDynamics(device=self.device)`, but this object is never called in `_step` or `_reset`. The entire `FinancialMarketDynamics` class and `MarketRegime` enum are dead code. The environment name is "FinancialRegimeEnv" but regime detection isn't actually integrated.

torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+        self.observation_spec = Composite(
+            price_history=UnboundedContinuous(
+                shape=(window_size,),
+                dtype=torch.float32,
+            ),
+            current_holdings=Binary(
+                shape=(1,),
+                dtype=torch.bool,
+            ),
+            step_count=UnboundedContinuous(
```
vmoens: `UnboundedContinuous` is for floating-point values. For integer step counts, use `UnboundedDiscrete` or remove it from observation_spec entirely (many environments don't expose step count as an observation).

torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+            new_holdings[buy_mask, 0] = True
+            # Pay transaction cost for buying
+            reward[buy_mask] = -transaction_cost[buy_mask]
+
+        # Sell action (2)
+        sell_mask = (action == 2) & (current_holdings.squeeze(-1))
+        if sell_mask.any():
+            new_holdings[sell_mask, 0] = False
+            # Receive profit/loss minus transaction cost
+            reward[sell_mask] = price_change[sell_mask] - transaction_cost[sell_mask]
```
vmoens: When selling, `price_change` is only the single-step return, not the gain/loss from when the position was opened. To properly reward P&L, you'd need to track the entry price in the state.

torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+        valid_returns = returns[..., 1:]  # Skip NaN
+
+        volatility = torch.std(valid_returns, dim=-1) * math.sqrt(252)
+
+        # Clamp volatility to reasonable range
+        return torch.clamp(volatility, min=0.1, max=3.0)
+
+
+class FinancialRegimeEnv(EnvBase):
+    """A financial trading environment with market regime detection.
```
vmoens: The specs shown in the class docstring don't include `step_count`, which is part of the actual observation_spec.

torchrl/envs/custom/trading.py
```diff
@@ -0,0 +1,695 @@
...
+                    reward: UnboundedContinuous(
+                        shape=torch.Size([1]),
+                        device=cpu,
+                        dtype=torch.float32,
+                        domain=continuous),
+                    device=None,
+                    shape=torch.Size([])),
+                full_done_spec: Composite(
+                    done: Categorical(
+                        shape=torch.Size([1]),
```
vmoens, line 73: TorchRL environments often distinguish between `terminated` (episode ended naturally) and `truncated` (time limit reached). Since this environment ends on max_steps (truncation), you might want to set `truncated=True` and `terminated=False` for the max_steps case.

vmoens (commented):
I left some comments. 


Some further question:

1. **What is the intended use of `FinancialMarketDynamics`?** The class has substantial implementation for regime detection, volatility calculation, etc., but none of it is integrated into the environment. Is this planned for a follow-up PR, or should it be removed from this PR?

2. **Should the environment track entry price?** The current reward structure doesn't properly account for position P&L since entry price isn't stored. Is this intentional (reward shaping based on single-step returns) or an oversight?

3. **Is per-step randomness intentional?** Unlike `PendulumEnv` which is deterministic in `_step`, this environment has random price generation in each step. This breaks reproducibility with `set_seed()`. Is this the intended behavior? Should we make this optional?

aneesh223:
I addressed your feedback with commit e308f85:

- Changed test/test_financialenv.py to test/test_custom_envs.py; removed unused FinancialMarketDynamics class, which I may reimplement in a future PR.

- Added tracking of entry_price for correct P&L rewards calculation; switched to seeded randomness for reproducibility.

- Changed step_count attribute type from UnboundedContinuous to UnboundedDiscrete; updated docstrings to match observation specifications.

- Replaced unsafe calls of view() with calls of reshape().

Pre-commit checks passed locally. Ready for another look just in case I missed anything!

test/test_custom_envs.py
aneesh223: That should work! I implemented this.

aneesh223:
@vmoens There was an issue with the scalar shape handling in randn, so I went ahead and fixed that.

PR #3403, opened 2026-01-28
[CI] Add path-based triggers for niche workflows

vmoens:
* __->__ #3403
* #3404
* #3402

Make expensive/niche workflows conditional on labels to reduce CI costs:

- docs.yml: Only builds docs on PRs when Documentation or tutorials/ label
  is present. Always builds on push to main/nightly/release.

- benchmarks_pr.yml: Only runs benchmarks when Performance or
  benchmarks/upload label is present. This is expensive (GPU, runs twice).

- test-linux-sota.yml: Only runs SOTA tests when sota-implementations/
  label is present. Always runs on push to main/nightly/release.

Labels are auto-applied by the file-based labeler (auto-labeler.yml) when
relevant files change, so developers don't need to manually add them.

Core library tests (test-linux.yml) continue to run on all PRs since
components are interconnected (e.g., replay buffers affect collectors).


gtnv:
hi @vmoens, is this still planned? if not, would you like to close it?

PR #3442, opened 2026-02-04
[BugFix] Fix shared mem updater with many policies

vmoens:


vmoens (commented):
## Code Review: [BugFix] Fix shared mem updater with many policies

### Summary

This PR fixes a bug in `_make_policy_factory` where a list of policy factories was incorrectly returned as-is instead of using `worker_idx` to select and invoke the correct factory. It also adds comprehensive support for per-worker weight synchronization when distinct policy factories are used.

### Bug Fix ✅

The core bug fix is correct and important:

```python
# Before (bug): returned the sequence directly
if isinstance(policy_factory, Sequence):
    return policy_factory  # Wrong!

# After (fix): uses worker_idx to get and call the correct factory
if isinstance(policy_factory, Sequence):
    policy = policy_factory[worker_idx]()  # Correct!
```

### Issues and Concerns

#### 1. Missing bounds check in `_make_policy_factory`
**File:** `torchrl/collectors/utils.py`

When `policy_factory` is a `Sequence`, there's no validation that `worker_idx < len(policy_factory)`. If `worker_idx >= len(policy_factory)`, this will raise an unhelpful `IndexError`:

```python
if isinstance(policy_factory, Sequence):
    # Consider adding: if worker_idx >= len(policy_factory): raise ValueError(...)
    policy = policy_factory[worker_idx]()
```

#### 2. Potential memory overhead in per-worker mode
**File:** `torchrl/weight_update/_shared.py`

In per-worker mode, device-based deduplication is disabled:

```python
if per_worker:
    # No deduplication - every worker has unique weights
    for worker_idx, weights in params_map.items():
        self._unique_weights.append(weights)
```

For N workers on the same device with model size M, this means O(N×M) memory instead of O(M). This could be significant for large models. Is this intentional? If so, a note in the docstring would help users understand the trade-off.

#### 3. Unnecessary CUDA synchronization for CPU-only setups
**File:** `torchrl/weight_update/_shared.py`, line 219-220

```python
if torch.cuda.is_available():
    torch.cuda.synchronize()
```

This synchronizes even when all tensors are on CPU. Consider checking if any weights are actually on CUDA before synchronizing, or document why unconditional sync is necessary.

#### 4. Identity check for uniform factory detection
**File:** `torchrl/collectors/_multi_base.py`

```python
has_uniform_policy_factory = any(policy_factory) and all(
    f is policy_factory[0] for f in policy_factory
)
```

Using `is` (identity check) rather than `==` (equality) is intentional for detecting replicated factories, but could be surprising to users who provide equivalent but distinct factory objects. Worth documenting this behavior.

#### 5. Early exit optimization for per-worker dict detection
**File:** `torchrl/weight_update/_shared.py`, lines 173-175

```python
if isinstance(weights, dict) and all(
    isinstance(k, int) for k in weights.keys()
):
```

This iterates all keys. For large dicts, consider using a more efficient check:

```python
if isinstance(weights, dict) and weights and isinstance(next(iter(weights.keys())), int):
```

Though this may be premature optimization if the dict is typically small.

### Tests ✅

The new tests are comprehensive and cover:
- `test_make_policy_factory_with_sequence_uses_worker_idx` - Unit test for the bug fix
- `test_multi_collector_with_distinct_policy_factories` - Integration test
- `test_per_worker_weight_sync_with_distinct_factories` - Per-worker weight sync
- `test_per_worker_weight_sync_multiple_workers_update` - Multi-worker updates

### Documentation ✅

The docstring updates in `_base.py` and `_shared.py` clearly document the new per-worker weight update API.

### Overall

The bug fix is solid and the per-worker weight sync feature is well-implemented. The main concern is the memory overhead in per-worker mode which should be documented. The other points are minor improvements.

PR #3459, opened 2026-02-06
[Example] Dreamer: SerialEnv mode and collector compile config

vmoens:
Add two configuration knobs to the Dreamer example:

1. `env.parallel_env_mode` ("parallel" | "serial"): switches the train
   environment between ParallelEnv (uses IPC) and SerialEnv (no IPC
   overhead, better for cheap envs or when GPU contention between
   collector workers degrades throughput). Parallel mode remains the default.

2. `collector.compile` block (enabled, backend, cudagraphs): passes
   compilation settings to MultiCollector via `compile_policy` and
   `cudagraph_policy` kwargs, enabling `torch.compile` + CUDA graphs for
   the policy in collector workers.

### DIFF
--- a/sota-implementations/dreamer/config.yaml
+++ b/sota-implementations/dreamer/config.yaml
@@ -9,6 +9,10 @@ env:
   image_size : 64
   horizon: 500
   n_parallel_envs: 8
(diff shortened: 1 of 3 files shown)


vmoens:
@torchrlbot rebase

PR #3468, opened 2026-02-09
[Refactor] Refactor optimization API for multi-phase optimization

bsprenger:
## Description

Introduce `OptimizationStepper`, an object that encapsulates one full optimization step in the Trainer. This is intended to eventually replace the list-based hook stages for the optimization step.

The Trainer now accepts an optional `optimization_stepper` parameter. The old hook-based optimization API (`OptimizerHook`, `post_loss`, `process_loss`, and `optimizer` hook stages) continues to work but now emits `FutureWarning` messages indicating the API will be removed.

Existing "standard" optimization behaviour is easily preserved by default using the new `DefaultOptimizationStepper`, to allow for easy transition for users to the new API.

## Motivation and Context

The Trainer's optimization loop computes all losses up-front, then runs a single optimizer stage. This works for single-phase algorithms (DQN, simple PPO) where all losses are computed once and a single optimizer step follows.

However, TD3 (and other multi-phase algorithms like SAC with auto-tuned entropy, or some model-based methods) requires a fundamentally different optimization order:

1. Compute **critic loss** from the batch
2. **Step the critic optimizer** (backward + step)
3. Compute **actor loss** using the **now-updated** critic (the actor loss signal depends on the critic parameters that were just updated)
4. **Step the actor optimizer**, but only every N iterations (delayed policy update)

The hard part is the interleaving of loss computations and optimization steps. The existing list-based, sequential optimization hook did not (cleanly) allow for these types of algorithms.

This PR will eventually replace the single optimizer hook with a single `OptimizationStepper` per Trainer, called directly (not via hook dispatch). This allows for much more flexibility in defining the optimization step and can accommodate algorithms like TD3.

For now, this feature is opt-in. Existing behaviour remains the same, as long as an `OptimizationStepper` is not passed explicitly to the trainer.

- [x] I have raised an issue to propose this change #3469

## Types of changes

## Checklist


bsprenger:
I seemingly cannot edit the title to fix the failing PR label CI job 😖

bsprenger:
Thanks @vmoens , makes total sense. I've updated the PR and the description to make this new feature opt-in for now. I think it's along the lines of what you suggest. If the user does not explicitly pass this new feature, the existing behaviour is unchanged.

Let me know if you want me to get a PR ready and waiting for the next phase ("Release 0.{x+1}: change default behaviour")!

And thanks again for your very quick review! 😃

vmoens (approved):
LGTM 
Let's see what the CI says and merge if appropriate

bsprenger:
As far as I can tell, the failing CI job is a just a permissions issue for publishing the benchmark results

PR #3540, opened 2026-03-04
[Refactor] Upgrade to torchcodec for video export

vmoens:

### DIFF
--- a/torchrl/record/loggers/common.py
+++ b/torchrl/record/loggers/common.py
@@ -16,7 +16,13 @@ from torch import Tensor
 from torchrl._utils import implement_for
 
 _has_tv = importlib.util.find_spec("torchvision") is not None
-_has_torchcodec = importlib.util.find_spec("torchcodec") is not None
+try:
+    from torchcodec.encoders import VideoEncoder  # noqa: F401
+
+    _has_torchcodec = True
+    del VideoEncoder
+except Exception:
+    _has_torchcodec = False
 
 
 __all__ = ["Logger"]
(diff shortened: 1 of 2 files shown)


torchrl/record/loggers/common.py
```diff
@@ -5,17 +5,70 @@
...
+    if crf is not None:
+        to_file_kwargs["crf"] = float(crf)
+    if preset is not None:
+        to_file_kwargs["preset"] = preset
+    if pixel_format is not None:
+        to_file_kwargs["pixel_format"] = pixel_format
+    if options:
+        to_file_kwargs["extra_options"] = options
+
+    VideoEncoder(frames=video_array, frame_rate=fps).to_file(filename, **to_file_kwargs)
```
NicolasHug, line 75: [`write_video` ](https://github.com/NicolasHug/vision/blob/4e0198c36f445ac77e950fb027513bb4bfbc9c15/torchvision/io/video.py#L56-L174)from torchvision was a self-contained helper that relied on `pyav`. If you want to avoid having to maintain multiple version of your own `_write_video` with different sets of dependencies for each, you could consider just vendoring `write_video` inside of torchRL 

You'd have a single optional dependency on pyav instead of having an optional dep on both TV and TC, and you wouldn't have to worry about TV's versions either

vmoens: > [write_video ](https://github.com/NicolasHug/vision/blob/4e0198c36f445ac77e950fb027513bb4bfbc9c15/torchvision/io/video.py#L56-L174)from torchvision was a self-contained helper that relied on pyav. If you want to avoid having to maintain multiple version of your own _write_video with different sets of dependencies for each, you could consider just vendoring write_video inside of torchRL

Thanks!

We still use tv's transforms so tv dep is not a big worry. But the point of removing an intermediary makes total sense!

PR #3548, opened 2026-03-06
[Bugfix] Add MPS float64->float32 downcast

bsprenger:
## Description

Add MPS float64 -> float32 downcast in `_default_dtype_and_device()` to prevent crashes when moving specs to MPS device.

MPS does not support float64 tensors. This is a problem e.g. when gymnasium/MuJoCo specs (which use float64) are created on CPU and later moved to MPS via `.to("mps")`, the spec constructors call `torch.full(..., device="mps", dtype=torch.float64)` which raises an error.

The fix adds a guard in `_default_dtype_and_device()` that automatically downcasts float64 -> float32 with a `UserWarning` when the device is MPS.

## Motivation and Context

Gymnasium and MuJoCo environments produce float64 observation/action spaces by default. When a user creates a collector with `env_device="mps"` (e.g., in the a trainer), it will try to move the specs to MPS with float64, and crashes.

The fix prevents this crash by ensuring float64 specs are always downcast to float32 on MPS.

Fixes: #3549

## Types of changes

## Checklist

### DIFF
--- a/test/test_specs.py
+++ b/test/test_specs.py
@@ -4652,6 +4652,15 @@ def _has_mps():
 class TestMPSDtype:
     """Tests that MPS-incompatible dtypes (float64) are downcast to float32 in tensor specs."""
 
+    def test_mps_does_not_support_float64(self):
+        """Assert that MPS still doesn't support float64.
+
+        If this test fails, MPS has gained float64 support and the downcasts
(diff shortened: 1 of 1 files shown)


bsprenger:
@vmoens Thanks for the review, I added a test for that. Thanks!! 😄

PR #3560, opened 2026-03-23
[Performance] Add _skip_maybe_reset flag to bypass auto-reset in step_and_maybe_reset

vmoens:
* #3567
* #3566
* #3565
* #3564
* #3563
* #3562
* #3561
* __->__ #3560

For environments that handle resets internally in _step() (e.g., GPU-batched
auto-resetting envs), the maybe_reset() call in step_and_maybe_reset() is
redundant overhead: it checks done flags, clones tensors, and potentially
calls reset() again. Adding _skip_maybe_reset = True on such envs skips
this entire codepath.

Made-with: Cursor


torchrl/envs/common.py
vmoens, line 1: We don't want that.
The proper way of doing auto-reset with torch compile should be to ALWAYS compute the reset and then mask reset and non-rest in `tensordict_` using torch.where. We should have a toy auto-reset env that we can compile as an example and we should discuss how to make an extension point out of this but we should NOT skip maybe_reset entirely. I would rather make it a no-op if auto-reset is used with the masking I just talked about, or implement the masking in that `maybe_reset` which I would find more natural.

PR #3562, opened 2026-03-23
[Performance] Add _trust_step_output flag to skip step validation overhead

vmoens:
* #3567
* #3566
* #3565
* #3564
* #3563
* __->__ #3562
* #3561
* #3560

When _trust_step_output is True on an env, step() skips _step_proc_data()
(reward shape checks, _complete_done, type checks) and TransformedEnv._step()
skips the inner _complete_done call on the base env. This eliminates per-step
Python iteration over spec keys for envs that guarantee correct output from
their _step() method.

Made-with: Cursor


torchrl/envs/common.py
```diff
@@ -379,6 +379,10 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         _skip_maybe_reset (bool): if ``True``, :meth:`step_and_maybe_reset` will skip calling
             :meth:`maybe_reset` after each step. This is useful for auto-resetting environments
             that already handle resets inside their :meth:`_step` method. Defaults to ``False``.
+        _trust_step_output (bool): if ``True``, :meth:`step` will skip the :meth:`_step_proc_data`
+            validation (reward shape checks, done-key completion, type checks) after :meth:`_step`.
+            Set this when the environment guarantees that its :meth:`_step` output always has correct
+            shapes, all done keys present, and proper dtypes. Defaults to ``False``.
```
vmoens, line 385: This is a bit of a footgun. I'm kind of ok with it but looking at other PRs down the stack we MUST make sure that all the side effects are super documented, like lack of partial steps support (which require control flow) and such. This should also be marked with a massive "Experimental" flag for anyone who wants to play with it.

PR #3563, opened 2026-03-23
[Performance] Add update_traj_ids flag to Collector to skip trajectory tracking

vmoens:
* #3567
* #3566
* #3565
* #3564
* __->__ #3563
* #3562
* #3561
* #3560

When update_traj_ids=False, the collector skips _update_traj_ids() in the
inner rollout loop. This eliminates per-step done-key iteration and
masked_scatter overhead for use cases that don't need trajectory splitting
(e.g., fixed-length PPO rollouts).

Made-with: Cursor


torchrl/collectors/_single.py
vmoens, line 1: I don't like this, this PR should be dropped entirely

PR #3564, opened 2026-03-23
[Performance] Streamline collector inner loop carrier update

vmoens:
* #3567
* #3566
* #3565
* __->__ #3564
* #3563
* #3562
* #3561
* #3560

Pre-computes _env_returns_carrier flag to avoid per-step identity check.
Skips collector data .copy() when update_traj_ids is False (no mutation
risk). Uses internal _get_str/_set_str for collector data transfer to
bypass public API validation overhead.

Made-with: Cursor


torchrl/collectors/_single.py
```diff
@@ -1731,12 +1730,19 @@ def rollout(self) -> TensorDictBase:
...
-                # carry over collector data without messing up devices
-                collector_data = self._carrier.get("collector").copy()
+                collector_data = self._carrier._get_str("collector", default=None)
+                if self.update_traj_ids:
+                    collector_data = collector_data.copy()
                 self._carrier = env_next_output
                 if self._shuttle_has_no_device:
                     self._carrier.clear_device_()
-                self._carrier.set("collector", collector_data)
+                self._carrier._set_str(
```
vmoens, line 1739: using private tensordict methods is a footgun, let's not do that

PR #3565, opened 2026-03-23
[Performance] Add fast path for step() and TransformedEnv._step() when _trust_step_output is set

vmoens:
* #3567
* #3566
* __->__ #3565
* #3564
* #3563
* #3562
* #3561
* #3560

When _trust_step_output is True, EnvBase.step() skips _assert_tensordict_shape,
partial_steps handling, next_preset logic, and _step_proc_data. Similarly,
TransformedEnv._step() skips partial_steps, next_preset, and _complete_done.
This eliminates all per-step Python validation overhead for well-behaved envs.

Made-with: Cursor


torchrl/envs/transforms/transforms.py
```diff
@@ -1220,12 +1220,15 @@ def rand_action(self, tensordict: TensorDictBase | None = None) -> TensorDict:
         return super().rand_action(tensordict)
 
     def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
-        # No need to clone here because inv does it already
-        # tensordict = tensordict.clone(False)
+        if self.base_env._trust_step_output:
```
vmoens, line 1223: how do we guarantee that there are no partial steps and such? Isn't this a bit of a footgun?

PR #3581, opened 2026-03-27
[BugFix] Fix check_env_specs() failing when state_spec contains keys not in observation_spec

### LINKED ISSUE #3260: [BUG] check_env_specs() fails if the state_spec contains tensors not in the observation
## Describe the bug

I have the below specs. Of particular importance are my observation_spec and state_spec (the first two specs):
```python
self.observation_spec = Composite(
    {
        "timestep": Unbounded(
            shape=torch.Size([n_envs]),
            dtype=torch.int64,
            device=self.device,
        ),
        "agents": Composite(
            {
                "observations": Categorical(
                    n=40,
                    shape=torch.Size([n_envs, n_agents, obs_dim_raw]),
                    dtype=torch.int64,
                    device=self.device,
                ),
                "action_mask": Binary(
                    n=action_mask_dim,
                    shape=torch.Size([n_envs, n_agents, action_mask_dim]),
                    dtype=torch.bool,
                    device=self.device,
                ),
            },
            shape=agent_batch_size,
        ),
    },
    shape=self.batch_size,
)
self.state_spec = Composite(
    {
        "randomStates": Unbounded(
            shape=torch.Size([n_envs, 6, 24, 48]),
            dtype=torch.uint8,
            device=self.device,
        ),
        "agents": Composite(
            {
                "rewards": reward_spec
            },
            shape=agent_batch_size,
        )
    },
    shape=self.batch_size,
)
self.action_spec = Composite(
    {
        "agents": Composite(
            {
                "actions": Categorical(
                    n=action_mask_dim,
                    shape=torch.Size([n_envs, n_agents, 1]),
                    dtype=torch.int64,
                    device=self.device,
                )
            },
            shape=agent_batch_size,
        ),
    },
    shape=self.batch_size,
)
self.reward_spec = Composite(
    {
        "agents": Composite(
            {
                "rewards": reward_spec
            },
            shape=agent_batch_size,
        )
    },
    shape=self.batch_size,
)
self.done_spec = Composite(
    {
        "done": Categorical(
            n=2,
            shape=torch.Size([n_envs, 1]),
            dtype=torch.bool,
            device=self.device,
        ),
        "agents": Composite(
            {
                "stepDones": Categorical(
                    n=2,
                    shape=torch.Size([n_envs, n_agents, 1]),
                    dtype=torch.bool,
                    device=self.device,
                )
            },
            shape=agent_batch_size,
        ),
    },
    shape=self.batch_size,
)
```

`randomStates` contains current random seed used to generate random numbers for the next step's computations. This is necessary to have repeatable states/transitions (like needing to backtrack to previous state for MCTS) but shouldn't be in the observation, so I put it in the state_spec since afaik that is the correct spot. I also include `rewards` in my state_spec because I update my `rewards` tensor in-place in my custom Torch op step.

However, `check_env_specs()` fails due to the below error: 
```
AssertionError: The keys of the specs and data do not match:
- List of keys present in real but not in fake: keys_in_real_not_in_fake={('next', 'agents', 'actions'), ('next', 'randomStates')},
- List of keys present in fake but not in real: keys_in_fake_not_in_real=set().
```
I believe the root-cause of the `('next', 'randomStates')` is [this line in `envs/common.py`](https://github.com/pytorch/rl/blob/v0.10.1/torchrl/envs/common.py#L3875). I think the `next` TensorDict attribute should be the next state after RL step transition. Instead, it clones from the observation which results in a difference in keys. When `check_env_specs()` calls `rollout()` the real TensorDict has `next` including Tensors from the state_spec resulting in the discrepancy.

I am also confused why the `fake_state` gets updated with actions [here](https://github.com/pytorch/rl/blob/main/torchrl/envs/common.py#L3883). Shouldn't actions stay separate from the state and be added to `next_output` later? I think I am misunderstanding something about how specs are intended to work here.

#### Steps to reproduce the behavior.

Call `check_env_specs()` where the `state_spec` includes a separate Tensor besides `rewards`

## Expected behavior

`check_env_specs()` would succeed, where the generated fake TensorDict's `next` contains attributes from the state_spec.

## Screenshots

Can't easily provide screenshots, but maybe the below is useful:
At [this line](https://github.com/pytorch/rl/blob/v0.10.1/torchrl/envs/common.py#L3863), `fake_state` is this:
```
TensorDict(
    fields={
        agents: TensorDict(
            fields={
                rewards: Tensor(shape=torch.Size([4, 4]), device=cuda:0, dtype=torch.float32, is_shared=True)},
            batch_size=torch.Size([4, 4]),
            device=cuda:0,
            is_shared=True),
        randomStates: Tensor(shape=torch.Size([4, 6, 24, 48]), device=cuda:0, dtype=torch.uint8, is_shared=True)},
    batch_size=torch.Size([4]),
    device=cuda:0,
    is_shared=True)
```
At [this line](https://github.com/pytorch/rl/blob/v0.10.1/torchrl/envs/common.py#L3871), `fake_state` has been updated to this (this is `fake_input` as well):
```
TensorDict(
    fields={
        agents: TensorDict(
            fields={
                actions: Tensor(shape=torch.Size([4, 4, 1]), device=cuda:0, dtype=torch.int64, is_shared=True),
                rewards: Tensor(shape=torch.Size([4, 4]), device=cuda:0, dtype=torch.float32, is_shared=True)},
            batch_size=torch.Size([4, 4]),
            device=cuda:0,
            is_shared=True),
        randomStates: Tensor(shape=torch.Size([4, 6, 24, 48]), device=cuda:0, dtype=torch.uint8, is_shared=True)},
    batch_size=torch.Size([4]),
    device=cuda:0,
    is_shared=True)
```
At [this line](https://github.com/pytorch/rl/blob/v0.10.1/torchrl/envs/common.py#L3879), before consolidating with `fake_in_out`, `next_output` is this:
```
TensorDict(
    fields={
        agents: TensorDict(
            fields={
                action_mask: Tensor(shape=torch.Size([4, 4, 79]), device=cuda:0, dtype=torch.bool, is_shared=True),
                observations: Tensor(shape=torch.Size([4, 4, 52]), device=cuda:0, dtype=torch.int64, is_shared=True),
                rewards: Tensor(shape=torch.Size([4, 4]), device=cuda:0, dtype=torch.float32, is_shared=True),
                stepDones: Tensor(shape=torch.Size([4, 4, 1]), device=cuda:0, dtype=torch.bool, is_shared=True)},
            batch_size=torch.Size([4, 4]),
            device=cuda:0,
            is_shared=True),
        done: Tensor(shape=torch.Size([4, 1]), device=cuda:0, dtype=torch.bool, is_shared=True),
        terminated: Tensor(shape=torch.Size([4, 1]), device=cuda:0, dtype=torch.bool, is_shared=True),
        timestep: Tensor(shape=torch.Size([4]), device=cuda:0, dtype=torch.int64, is_shared=True)},
    batch_size=torch.Size([4]),
    device=cuda:0,
    is_shared=True)
```

## System info

 * Describe how the library was installed: Initially installed via `pip` as v0.9, then installed from-source as v10.1. Issue is in both versions.
 * Python version: Python 3.11
 * Versions of any other relevant libraries:
    * torch = 2.7.1
    * tensordict = 0.10.0

## Additional context

My environment uses a custom CUDA kernel to step the vectorized environment, so I need the `randomStates` for cuRAND randomness.

## Reason and Possible fixes

Linked above

## Checklist

- [X] I have read the [documentation](https://github.com/pytorch/rl/tree/main/docs/) (**required**)
- [X] I have provided a minimal working example to reproduce the bug (**required**) - I believe the above is enough, I can try to pare down my environment to an even smaller example if necessary

theap06:
Description
fake_tensordict() builds next_output (the fake next key) by cloning fake_obs, which only contains observation spec keys. This means any keys present in state_spec but not in observation_spec are absent from the fake next, causing check_env_specs() to fail when comparing against a real rollout where next correctly contains those state keys.

The fix initializes next_output from fake_state instead, then layers observations, rewards, and dones on top — matching the real rollout behavior.

Motivation and Context
Environments that store non-observable state (e.g. RNG seeds for a custom CUDA kernel, in-place reward tensors) in state_spec rather than observation_spec would fail check_env_specs() with an assertion error about mismatched keys in next, even though the environment is correctly implemented.

Closes #3260

 I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)
Types of changes
 Bug fix (non-breaking change which fixes an issue)
Checklist
 I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (required)
 My change requires a change to the documentation.
 I have updated the tests accordingly (required for a bug fix or a new feature).
 I have updated the documentation accordingly.
Note: I checked the tests box — you'll want to add a test case to confirm before submitting, or uncheck it if you're skipping that for now.

### DIFF
--- a/test/test_envs.py
+++ b/test/test_envs.py
@@ -302,8 +302,9 @@ class TestEnvBase:
 
         check_env_specs() should succeed when state_spec contains keys that are
         not present in observation_spec (e.g. RNG seeds, auxiliary state).
-        The fake 'next' tensordict must include those state keys so it matches
-        a real rollout.
+        Such keys may be returned by _step and appear in the real rollout's
+        "next", but are intentionally excluded from the key comparison in
+        check_env_specs() since their presence is optional.
         """
 
         class EnvWithExtraState(EnvBase):
@@ -341,14 +342,9 @@ class TestEnvBase:
             def _set_seed(self, seed):
                 pass
 
-        env = EnvWithExtraState()
-        # fake_tensordict()'s "next" must contain the state_spec key
-        ftd = env.fake_tensordict()
-        assert (
-            "hidden_state" in ftd["next"].keys()
-        ), "state_spec key 'hidden_state' missing from fake_tensordict()['next']"
-        # check_env_specs() must not raise
-        check_env_specs(env)
+        # check_env_specs() must not raise even though _step returns
+        # "hidden_state" (a state_spec key) in "next"
+        check_env_specs(EnvWithExtraState())
 
     class MyEnv(EnvBase):
         def __init__(self):
--- a/torchrl/envs/common.py
+++ b/torchrl/envs/common.py
@@ -4029,8 +4029,7 @@ class EnvBase(nn.Module, metaclass=_EnvPostInit):
         # Hence we generate the input, and override using the output
         fake_in_out = fake_input.update(fake_obs)
 
-        next_output = state_spec.zero()
-        next_output.update(fake_obs)
+        next_output = fake_obs.clone()
         next_output.update(fake_reward)
         next_output.update(fake_done)
         fake_in_out.update(fake_done.clone())
--- a/torchrl/envs/utils.py
+++ b/torchrl/envs/utils.py
@@ -771,12 +771,27 @@ def check_env_specs(
         fake_tensordict = LazyStackedTensorDict.lazy_stack(
             [fake_tensordict.clone() for _ in range(3)], -1
         )
-    # eliminate empty containers
+    # state_spec keys may optionally appear in "next" if _step returns them,
+    # so exclude them from all comparisons to avoid false positives in either
+    # direction (real has them but fake doesn't, or vice versa).
+    state_spec_next_keys = {
+        ("next", *k) if isinstance(k, tuple) else ("next", k)
+        for k in env.state_spec.keys(True, True)
+    }
+    # eliminate empty containers, excluding state_spec keys from "next"
     fake_tensordict_select = fake_tensordict.select(
-        *fake_tensordict.keys(True, True, is_leaf=_default_is_leaf)
+        *[
+            k
+            for k in fake_tensordict.keys(True, True, is_leaf=_default_is_leaf)
+            if k not in state_spec_next_keys
+        ]
     )
     real_tensordict_select = real_tensordict.select(
-        *real_tensordict.keys(True, True, is_leaf=_default_is_leaf)
+        *[
+            k
+            for k in real_tensordict.keys(True, True, is_leaf=_default_is_leaf)
+            if k not in state_spec_next_keys
+        ]
     )
     # check keys
     fake_tensordict_keys = set(
@@ -785,6 +800,8 @@ def check_env_specs(
     real_tensordict_keys = set(
         real_tensordict.keys(True, True, is_leaf=_is_leaf_nontensor)
     )
+    fake_tensordict_keys -= state_spec_next_keys
+    real_tensordict_keys -= state_spec_next_keys
     if fake_tensordict_keys != real_tensordict_keys:
         keys_in_real_not_in_fake = real_tensordict_keys - fake_tensordict_keys
         keys_in_fake_not_in_real = fake_tensordict_keys - real_tensordict_keys


vmoens:
@theap06 can you think of a non-regression test?

theap06:
@vmoens just added a test csaes for it. should pass ci now

PR #3537, opened 2026-02-27
[Feature] PILCO

PSXBRosa:
## Description

This PR introduces the implementation of the PILCO (Probabilistic Inference for Learning Control) algorithm to TorchRL.

Key details of the implementation:

- Gaussian Process Regression: I utilized the external libraries botorch and gpytorch for the GPR components. this avoids the overhead and complexity of maintaining a custom GPR implementation.
- Moment Matching: I initially considered and experimented with a Monte Carlo approach for moment matching to simplify the underlying mathematics. While I couldn't get the MC approach to stabilize and work correctly (though it remains an interesting alternative), the current working implementation relies on the analytical forms for moment matching. This aligns directly with what was done in the original PILCO paper by Deisenroth and Rasmussen.
- Credits: I want to give a huge thanks and credit to @nrontsis. I used the code from their repository (https://github.com/nrontsis/PILCO) as a valuable implementation reference and to test/validate different parts of my own PyTorch adaptation.

## Motivation and Context

PILCO is a highly sample-efficient model-based reinforcement learning algorithm, making it a valuable addition to the library's algorithm suite.

close #3513

- [x] I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

What types of changes does your code introduce? Remove all that do not apply:

- [x] New feature (non-breaking change which adds core functionality)

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.
If you are unsure about any of these, don't hesitate to ask. We are here to help!

- [x] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] My change requires a change to the documentation.
- [ ] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.

PR #3582, opened 2026-03-27
PILCO (reopened from #3537)

vmoens:


torchrl/modules/models/gp.py
```diff
@@ -0,0 +1,645 @@
...
+        D = self.obs_dim  # state dimension
+        DF = self.state_action_dim  # D+F, dimension of x̃
+
+        # ---- Build joint state-action distribution p(x̃_{t-1}) (Sec. 2.2) ----
+        # μ̃_{t-1} = [μ_x; μ_u] ∈ R^{B×(D+F)}
+        mu_tilde = torch.cat([mu_x, mu_u], dim=-1)
+
+        # Σ̃_{t-1} = [[Σ_x,        Σ_x C_xu    ],
+        #             [C_xu^T Σ_x^T, Σ_u        ]]  ∈ R^{B×(D+F)×(D+F)}
+        Sigma_x_C_xu = Sigma_x @ C_xu  # upper-right block [B, D, F]
```
vmoens, line 405: `C_xu` is documented here as `cov[x_{t-1}, u_{t-1}]`, and `RBFController.forward()` also returns an actual cross-covariance tensor under `('action', 'cross_covariance')`. Multiplying it by `Sigma_x` again when building `Sigma_tilde` changes the propagated belief unless `Sigma_x == I`. I think the upper-right block should just be `C_xu`, not `Sigma_x @ C_xu`.

sota-implementations/pilco/pilco.py
```diff
@@ -0,0 +1,199 @@
...
+        eval_env,
+        policy_module,
+        eval_horizon=cfg.pilco.eval_horizon,
+        step=0,
+        logger=logger,
+    )
+
+    cost_module = ExponentialQuadraticCost(reduction="none").to(env.device)
+    for epoch in range(cfg.pilco.epochs):
+        base_world_model = GPWorldModel(obs_dim=obs_dim, action_dim=action_dim).to(
```
vmoens, line 105: This path is dtype-fragile rather than unconditionally broken. Here policy_module and initial_observation are explicitly float64, while the world model inherits the dtype of the collected rollout. On InvertedPendulum-v5 that seems to work because the env emits float64 observations, so fit(rollout) also trains the GP in double precision. But on envs where the rollout stays float32, the imagined rollout can later fail with RuntimeError: expected scalar type Double but found Float when the float64 belief is fed into a float32 GP. It would be safer to make the dtype contract explicit and keep the world model, rollout data, and belief state on the same dtype.

torchrl/envs/model_based/imagined.py
```diff
@@ -0,0 +1,195 @@
...
+            return tensordict.copy()
+
+        obs = tensordict.get("observation", None)
+        if obs is None:
+            obs = self._base_observation_spec.rand()
+        if obs.ndim == 1:
+            obs = obs.expand(self.batch_size[0], -1)
+
+        obs = obs.to(self.device)
+        B, D = obs.shape
```
vmoens: This reset path only works for scalar / 1D batch sizes even though the constructor accepts arbitrary `Sequence[int]`. For example, with `batch_size=[2, 3]`, this line raises `ValueError: too many values to unpack (expected 2)`, and the `obs.expand(self.batch_size[0], -1)` branch above also only handles a single batch dimension. If multidimensional batches are meant to be supported here, the reset logic needs to preserve `*batch_shape, D` instead of assuming `(B, D)`.

torchrl/modules/models/gp.py
PSXBRosa: I thought the same at first, but the math is actually fine. This is actually a bug in the docs and naming.

If you check the linear algebra in `RBFController.forward()`, the tensor returned as `('action','cross_covariance')` isnt the real cross-cov. Since it drops the leading Σx when computing `t_scaled` and the final mapping, its actually returning the pre-solved matrix V = Σx⁻¹ cov[x,u].

This is a legacy optimization from the original MATLAB code, it avoids an extra `linalg.solve` when applying the chain rule for the squashed cross-cov.

So `C_xu` is really V. Then `Sigma_x @ C_xu` in `gp.py` becomes
Σx (Σx⁻¹ cov[x,u]) and it recovers the true cov[x,u] for the upper-right block of `Sigma_tilde`.

So the propagation math is correct, it shouldn't be changed. I'll push a commit renaming `cross_covariance` to `input_output_mapping` and updating the docs

PR #3584, opened 2026-03-30
[Feature] Trajectory batcher

### LINKED ISSUE #3234: [Feature Request] **TrajectoryBatcher** high-level util for complete episode batching
## Motivation

When using `MultiSyncDataCollector` with `split_trajs=True`, it is surprisingly difficult to obtain a batch of **N complete trajectories**.
Many RL algorithms (Monte Carlo / REINFORCE, episodic PPO rollouts, imitation learning, some RLHF setups) require full episodes padded into a single batch.

Right now users need to manually:

* track `traj_ids`,
* rebuild trajectories across collector iterations,
* detect completion via `mask` / `done`,
* pad variable-length episodes,
* optionally discard “mixed-policy” episodes when the policy is updated.

This logic is both error-prone and reimplemented repeatedly in user code.

There is currently no high-level utility in TorchRL that provides “give me exactly N full episodes as a padded TensorDict”, even though the underlying collectors already expose all the needed information.

## Solution

I propose a new utility, tentatively named **`TrajectoryBatcher`**, in `torchrl.collectors.utils`.

It would wrap any TorchRL collector and yield batches of exactly `num_trajectories` full episodes, reconstructed across collector steps and padded to `max_length`. Example:

```python
batcher = TrajectoryBatcher(
    collector,
    num_trajectories=32,
    strict_on_policy=True,
)

batch = next(iter(batcher))
batch["observation"].shape  # [32, max_len, obs_dim]
batch["action"].shape       # padded
batch["mask"].shape         # valid-step mask
```

Features:

* reconstruct trajectories split across collector iterations (using `traj_ids`)
* optional **strict on-policy** mode (“burn” ongoing trajectories after a policy update)
* padded output using standard TensorDict conventions
* compatible with multi-worker collectors without resetting environments

I already have a working prototype and can contribute the implementation, documentation, and unit tests.

## Alternatives

* Users manually stitch trajectories in Python (complex and repetitive).
* Resetting environments at every update is possible but costly and problematic with multi-process collectors.
* Using `frames_per_batch` as a proxy for episodes is unreliable and does not guarantee complete trajectories.

These alternatives are not ideal and introduce unnecessary boilerplate for common RL workflows.

## Additional context

This feature would make it much easier to train Monte Carlo / episodic algorithms on top of TorchRL collectors, and aligns with patterns commonly found in other RL libraries (e.g., SB3 rollouts, RLlib sample batches, Acme episodes).

## Checklist

* [x] I have checked that there is no similar issue in the repo (**required**)

theap06:
## Description
Adds TrajectoryBatcher to torchrl/collectors/utils.py (and exports it from torchrl.collectors). It wraps any TorchRL collector and yields batches of exactly num_trajectories complete episodes, reconstructed across collector iterations using ("collector", "traj_ids"). Output is a zero-padded [N, max_len, ...] TensorDict with a ("collector", "mask") boolean field marking valid time steps.

from torchrl.collectors import Collector, TrajectoryBatcher

collector = Collector(env_fn, policy, frames_per_batch=200, total_frames=10_000)
batcher = TrajectoryBatcher(collector, num_trajectories=32, strict_on_policy=True)

for batch in batcher:
    # batch.shape == [32, max_episode_len]
    # batch[("collector", "mask")] marks valid steps
    loss = compute_loss(batch)
    loss.backward()
    optimizer.step()
    batcher.update_policy_weights_()  # burns partial trajectories when strict_on_policy=True
Fixes #3234 

Motivation and Context
Many RL algorithms (Monte Carlo / REINFORCE, episodic PPO, imitation learning, RLHF) require full episodes padded into a single batch. Currently users must manually: track traj_ids, rebuild trajectories across collector iterations, detect episode completion via done/mask, pad variable-length episodes, and optionally discard mixed-policy episodes when the policy is updated. This logic is error-prone and reimplemented repeatedly.

There is no existing high-level utility in TorchRL that provides "give me exactly N full episodes as a padded TensorDict", despite the collectors already exposing all the needed information (traj_ids, done, mask).

### DIFF
--- a/test/test_collectors.py
+++ b/test/test_collectors.py
@@ -2343,7 +2343,8 @@ class TestCollectorDevices:
                     device=None,
                 )
 
-        def _set_seed(self, seed: int | None = None) -> None: ...
+        def _set_seed(self, seed: int | None = None) -> None:
+            ...
 
     class EnvWithDevice(EnvBase):
         def __init__(self, default_device):
@@ -2399,7 +2400,8 @@ class TestCollectorDevices:
                     device=self.default_device,
                 )
 
-        def _set_seed(self, seed: int | None = None) -> None: ...
+        def _set_seed(self, seed: int | None = None) -> None:
+            ...
 
     class DeviceLessPolicy(TensorDictModuleBase):
         in_keys = ["observation"]
@@ -2574,7 +2576,8 @@ class TestCollectorDevices:
         def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
             return self.full_done_specs.zeros().update(self.observation_spec.zeros())
 
-        def _set_seed(self, seed: int | None) -> None: ...
+        def _set_seed(self, seed: int | None) -> None:
+            ...
 
     @pytest.mark.gpu
     @pytest.mark.skipif(not torch.cuda.is_available(), reason="no cuda device")
@@ -3451,7 +3454,8 @@ class TestUpdateParams:
                 {"state": self.state.clone()}, self.batch_size, device=self.device
             )
 
-        def _set_seed(self, seed: int | None) -> None: ...
+        def _set_seed(self, seed: int | None) -> None:
+            ...
 
     class Policy(TensorDictModuleBase):
         def __init__(self):
@@ -4151,9 +4155,9 @@ class TestDynamicEnvs:
 class TestCollectorsNonTensor:
     class AddNontTensorData(Transform):
         def _call(self, next_tensordict: TensorDictBase) -> TensorDictBase:
-            next_tensordict["nt"] = (
-                f"a string! - {next_tensordict.get('step_count').item()}"
-            )
+            next_tensordict[
+                "nt"
+            ] = f"a string! - {next_tensordict.get('step_count').item()}"
             return next_tensordict
 
         def _reset(
--- a/torchrl/collectors/_base.py
+++ b/torchrl/collectors/_base.py
@@ -515,7 +515,8 @@ class BaseCollector(IterableDataset, metaclass=abc.ABCMeta):
         self,
         policy_or_weights: TensorDictBase | TensorDictModuleBase | nn.Module | dict,
         /,
-    ) -> None: ...
+    ) -> None:
+        ...
 
     @overload
     def update_policy_weights_(
@@ -525,7 +526,8 @@ class BaseCollector(IterableDataset, metaclass=abc.ABCMeta):
         *,
         worker_ids: int | list[int] | torch.device | list[torch.device] | None = None,
         model_id: str | None = None,
-    ) -> None: ...
+    ) -> None:
+        ...
 
     @overload
     def update_policy_weights_(
@@ -534,7 +536,8 @@ class BaseCollector(IterableDataset, metaclass=abc.ABCMeta):
         weights: TensorDictBase | dict,
         model_id: str | None = None,
         worker_ids: int | list[int] | torch.device | list[torch.device] | None = None,
-    ) -> None: ...
+    ) -> None:
+        ...
 
     @overload
     def update_policy_weights_(
@@ -543,7 +546,8 @@ class BaseCollector(IterableDataset, metaclass=abc.ABCMeta):
         policy: TensorDictModuleBase | nn.Module,
         model_id: str | None = None,
         worker_ids: int | list[int] | torch.device | list[torch.device] | None = None,
-    ) -> None: ...
+    ) -> None:
+        ...
 
     @overload
     def update_policy_weights_(
@@ -553,7 +557,8 @@ class BaseCollector(IterableDataset, metaclass=abc.ABCMeta):
             str, TensorDictBase | TensorDictModuleBase | nn.Module | dict
         ],
         worker_ids: int | list[int] | torch.device | list[torch.device] | None = None,
-    ) -> None: ...
+    ) -> None:
+        ...
 
     def update_policy_weights_(
         self,
@@ -777,28 +782,32 @@ class BaseCollector(IterableDataset, metaclass=abc.ABCMeta):
 
     # Overloads for receive_weights to support multiple calling conventions
     @overload
-    def receive_weights(self) -> None: ...
+    def receive_weights(self) -> None:
+        ...
 
     @overload
     def receive_weights(
         self,
         policy_or_weights: TensorDictBase | TensorDictModuleBase | nn.Module | dict,
         /,
-    ) -> None: ...
+    ) -> None:
+        ...
 
     @overload
     def receive_weights(
         self,
         *,
         weights: TensorDictBase | dict,
-    ) -> None: ...
+    ) -> None:
+        ...
 
     @overload
     def receive_weights(
         self,
         *,
         policy: TensorDictModuleBase | nn.Module,
-    ) -> None: ...
+    ) -> None:
+        ...
 
     def receive_weights(
         self,
--- a/torchrl/collectors/_multi_base.py
+++ b/torchrl/collectors/_multi_base.py
@@ -334,9 +334,8 @@ class MultiCollector(BaseCollector, metaclass=_MultiCollectorMeta):
     def __init__(
         self,
         create_env_fn: Sequence[Callable[[], EnvBase]],
-        policy: None | (
-            TensorDictModule | Callable[[TensorDictBase], TensorDictBase]
-        ) = None,
+        policy: None
+        | (TensorDictModule | Callable[[TensorDictBase], TensorDictBase]) = None,
         *,
         num_workers: int | None = None,
         policy_factory: (
--- a/torchrl/collectors/_single.py
+++ b/torchrl/collectors/_single.py
@@ -414,9 +414,8 @@ class Collector(BaseCollector):
         create_env_fn: (
             EnvBase | EnvCreator | Sequence[Callable[[], EnvBase]]  # noqa: F821
         ),  # noqa: F821
-        policy: None | (
-            TensorDictModule | Callable[[TensorDictBase], TensorDictBase]
-        ) = None,
+        policy: None
+        | (TensorDictModule | Callable[[TensorDictBase], TensorDictBase]) = None,
         *,
         policy_factory: Callable[[], Callable] | None = None,
         frames_per_batch: int,
@@ -847,13 +846,13 @@ class Collector(BaseCollector):
         """Returns the compiled policy, compiling it lazily if needed."""
         if (policy := self._wrapped_policy_maybe_compiled) is None:
             if self.compiled_policy or self.cudagraphed_policy:
-                policy = self._wrapped_policy_maybe_compiled = (
-                    self._compile_wrapped_policy(self._wrapped_policy_uncompiled)
-                )
+                policy = (
+                    self._wrapped_policy_maybe_compiled
+                ) = self._compile_wrapped_policy(self._wrapped_policy_uncompiled)
             else:
-                policy = self._wrapped_policy_maybe_compiled = (
-                    self._wrapped_policy_uncompiled
-                )
+                policy = (
+                    self._wrapped_policy_maybe_compiled
+                ) = self._wrapped_policy_uncompiled
         return policy
 
     @property


torchrl/collectors/utils.py
```diff
@@ -406,6 +406,187 @@ def _map_weight(
     return weight
 
 
+class TrajectoryBatcher:
```
vmoens: This needs to be added to the doc in `docs/`

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+            batch_size=[n],
+        )
+
+    # ------------------------------------------------------------------
+    # Basic functionality
+    # ------------------------------------------------------------------
+
+    def test_single_batch_complete_trajectories(self):
+        """All trajectories complete within one collector batch."""
+        from torchrl.collectors.utils import TrajectoryBatcher
```
vmoens: no local import unless necessary

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+
+        mask = result[("collector", "mask")]
+        # traj 0: all 3 valid; traj 1: 2 valid, 1 padded
+        assert mask[0].all()
+        assert mask[1, :2].all()
+        assert not mask[1, 2]
+
+    def test_trajectories_span_batches(self):
+        """A trajectory split across two collector batches is reassembled."""
+        from torchrl.collectors.utils import TrajectoryBatcher
```
vmoens: ditto

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+        batcher = TrajectoryBatcher([batch], num_trajectories=1)
+        batches = list(batcher)
+
+        # Only traj 0 is complete
+        assert len(batches) == 1
+        assert batches[0].shape == torch.Size([1, 2])
+
+    def test_mask_field_shape_and_values(self):
+        """Mask shape matches batch shape and values reflect valid steps."""
+        from torchrl.collectors.utils import TrajectoryBatcher
```
vmoens: ditto

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+        mask = result[("collector", "mask")]
+        assert mask.shape == torch.Size([2, 4])
+        assert mask[0].all()  # traj 0: 4 valid steps
+        assert mask[1, :2].all()
+        assert not mask[1, 2]
+        assert not mask[1, 3]
+
+    def test_multiple_emitted_batches(self):
+        """More than num_trajectories complete trajectories → multiple emitted batches."""
+        from torchrl.collectors.utils import TrajectoryBatcher
```
vmoens: ditto

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+        for b in out:
+            assert b.shape[0] == 2
+
+    # ------------------------------------------------------------------
+    # strict_on_policy mode
+    # ------------------------------------------------------------------
+
+    def test_strict_on_policy_discards_partials(self):
+        """Calling update_policy_weights_ with strict_on_policy=True drops partials."""
+        from torchrl.collectors.utils import TrajectoryBatcher
```
vmoens: ditto

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+        )
+        assert len(batcher._partial_trajs) == 1
+        batcher.update_policy_weights_()
+        # Partial trajectories should be cleared
+        assert len(batcher._partial_trajs) == 0
+        assert "updated" in calls
+
+    def test_strict_on_policy_false_keeps_partials(self):
+        """With strict_on_policy=False, partials survive a policy update."""
+        from torchrl.collectors.utils import TrajectoryBatcher
```
vmoens: ditto

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+        # Partial trajectories should still be present
+        assert len(batcher._partial_trajs) == 1
+
+    # ------------------------------------------------------------------
+    # terminated key support
+    # ------------------------------------------------------------------
+
+    def test_terminated_key_treated_as_done(self):
+        """Trajectories ending via 'terminated' (not 'done') are still completed."""
+        from torchrl.collectors.utils import TrajectoryBatcher
```
vmoens: ditto

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+        assert len(batches) == 1
+        assert batches[0].shape == torch.Size([1, 3])
+
+    # ------------------------------------------------------------------
+    # Error handling
+    # ------------------------------------------------------------------
+
+    def test_missing_traj_ids_raises(self):
+        """Missing traj_ids field raises a descriptive KeyError."""
+        from torchrl.collectors.utils import TrajectoryBatcher
```
vmoens: ditto

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+        batcher = TrajectoryBatcher([bad_batch], num_trajectories=1)
+        with pytest.raises(KeyError, match="traj_ids"):
+            list(batcher)
+
+    # ------------------------------------------------------------------
+    # repr
+    # ------------------------------------------------------------------
+
+    def test_repr(self):
+        from torchrl.collectors.utils import TrajectoryBatcher
```
vmoens: ditto

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+    def test_multiple_emitted_batches(self):
+        """More than num_trajectories complete trajectories → multiple emitted batches."""
+        from torchrl.collectors.utils import TrajectoryBatcher
+
+        # 4 short trajectories, num_trajectories=2 → 2 emitted batches
+        batches_in = [
+            self._make_batch([i, i], [False, True], obs_start=i * 2) for i in range(4)
+        ]
+        # Concatenate into one big batch
+        import torch
```
vmoens: ditto

test/test_collectors.py
```diff
@@ -5296,6 +5296,257 @@ def env_fn():
...
+                # after policy update, yield a complete trajectory
+                yield TestTrajectoryBatcher._make_batch(
+                    [1, 1], [False, True]
+                )
+
+            def update_policy_weights_(self):
+                calls.append("updated")
+
+        batcher = TrajectoryBatcher(
+            _MockCollector(), num_trajectories=1, strict_on_policy=True
```
vmoens: I want to see tests with Collector and MultiCollector (sync and async)

torchrl/collectors/utils.py
vmoens: The LLMCollector has a "yield_completed_trajectories" arg. 
Can you elaborate why we're building a new class and not patching existing collectors to only yield trajectories when they're completed? 
https://docs.pytorch.org/rl/main/reference/generated/torchrl.collectors.llm.LLMCollector.html#torchrl.collectors.llm.LLMCollector

theap06: TrajectoryBatcher works with any collector without modifying the existing code. It handles the problem of grouping N trajectories into a padded batch for on-policy training -
yield trajectories streams one trajectory at a time.

theap06:
> Thanks for this!
> Curious to hear your thoughts about the need for a new class instead of a new kwarg in collector build functions?

TrajectoryBatcher is able to be a wrapper that avoids modifying every collector class, and handles the cross-batch trajectory reassembly + provides strict_on_policy bookkeeping; we can't do that easily by adding a kwarg

vmoens:
> TrajectoryBatcher is able to be a wrapper that avoids modifying every collector class, and handles the cross-batch trajectory reassembly + provides strict_on_policy bookkeeping; we can't do that easily by adding a kwarg

Is it as efficient though? Can we make this a post-process rather than a wrapper? 
I'm asking this because wrappers have a lot of unforeseen consequences like breaking isinstance checks, obscuring the nature of the object they wrap etc. For example weight sync, state-dict, etc are all kinds of things that we'll need to wire up. 
Adding the fact that we have collectors that have that `kwarg` already so the "two ways to do the same thing" may be puzzling and drive people away from collectors entirely.

theap06:
> > TrajectoryBatcher is able to be a wrapper that avoids modifying every collector class, and handles the cross-batch trajectory reassembly + provides strict_on_policy bookkeeping; we can't do that easily by adding a kwarg
> 
> Is it as efficient though? Can we make this a post-process rather than a wrapper? I'm asking this because wrappers have a lot of unforeseen consequences like breaking isinstance checks, obscuring the nature of the object they wrap etc. For example weight sync, state-dict, etc are all kinds of things that we'll need to wire up. Adding the fact that we have collectors that have that `kwarg` already so the "two ways to do the same thing" may be puzzling and drive people away from collectors entirely.

TrajectoryBatcher is gone, and the logic now lives as a num_trajectories_per_batch kwarg on all collectors. implemented in BaseCollector iter using private stateful helpers.

docs/source/reference/collectors_single.rst
```diff
@@ -20,6 +20,35 @@ Single node data collectors
     MultiSyncCollector
     MultiAsyncCollector
 
+Trajectory batching
+-------------------
+
+Pass ``num_trajectories_per_batch=N`` to any collector to receive batches of
+exactly *N* complete, zero-padded trajectories instead of fixed-frame batches.
```
vmoens: So frames per batch is ignored?

docs/source/reference/collectors_single.rst
```diff
@@ -20,6 +20,35 @@ Single node data collectors
...
+
+.. code-block:: python
+
+    from torchrl.collectors import Collector
+    from torchrl.envs import GymEnv
+
+    collector = Collector(
+        GymEnv("CartPole-v1"),
+        policy=my_policy,
+        frames_per_batch=200,
```
vmoens: If this is ignored we should comment it
```suggestion
        frames_per_batch=200, # ignored by collector
```
Or `superseded by num_trajectories_per_batch`

docs/source/reference/collectors_single.rst
```diff
@@ -20,6 +20,35 @@ Single node data collectors
...
+        total_frames=10000,
+        num_trajectories_per_batch=4,
```
vmoens: It's a bit clunky no? Could we use traj_per_batch (like frames per batch)?

test/test_collectors.py
```diff
@@ -5296,6 +5302,220 @@ def env_fn():
...
+                assert len(batches) > 0
+                for b in batches:
+                    assert b.shape[0] == 2
+                    mask = b[("collector", "mask")]
+                    assert mask.shape == b.shape
+                    assert mask.any(dim=-1).all()
+            finally:
+                collector.shutdown()
+        finally:
+            env.close()
```
vmoens: There is a flag here to avoid raising if close() fails

test/test_collectors.py
```diff
@@ -5296,6 +5302,220 @@ def env_fn():
...
+        """Missing traj_ids raises a descriptive KeyError."""
+        bad = TensorDict({"obs": torch.arange(3, dtype=torch.float)}, batch_size=[3])
+        with pytest.raises(KeyError, match="traj_ids"):
+            _traj_ingest(bad, {}, [])
+
+    # ------------------------------------------------------------------
+    # Integration tests with real collectors
+    # ------------------------------------------------------------------
+
+    def test_with_collector(self):
```
vmoens: This only loosely tests that trajs are complete. We should check the done states

test/test_collectors.py
```diff
@@ -5296,6 +5302,220 @@ def env_fn():
...
+            env.close()
+
+    @pytest.mark.parametrize(
+        "collector_cls",
+        [
+            functools.partial(MultiSyncCollector, cat_results="stack"),
+            MultiAsyncCollector,
+        ],
+    )
+    def test_with_multi_collector(self, collector_cls):
```
vmoens: This only loosely tests that trajs are complete. We should check the done states

test/test_collectors.py
```diff
@@ -5296,6 +5302,220 @@ def env_fn():
...
+            assert len(batches) > 0
+            for b in batches:
+                assert b.shape[0] == 2
+                mask = b[("collector", "mask")]
+                assert mask.shape == b.shape
+                assert mask.any(dim=-1).all()
+        finally:
+            collector.shutdown()
+
+    def test_with_collector_num_trajectories_one(self):
```
vmoens: Ditto

test/test_collectors.py
```diff
@@ -5296,6 +5302,220 @@ def env_fn():
...
+                for b in collector:
+                    assert b.shape[0] == 1
+                    mask = b[("collector", "mask")]
+                    assert mask[:, 0].all()
+            finally:
+                collector.shutdown()
+        finally:
+            env.close()
+
+    def test_trajectories_span_collector_batches(self):
```
vmoens: I wonder if we could group some of these tests together, it's a bit lengthy.
One or two tests with parametrize decorators should do the trick

torchrl/collectors/_base.py
```diff
@@ -949,11 +950,25 @@ def __iter__(self) -> Iterator[TensorDictBase]:
...
+                yield from self.iterator()
+            else:
+                yield from self._iter_by_trajectories()
         except Exception:
             self.shutdown()
             raise
 
+    def _iter_by_trajectories(self) -> Iterator[TensorDictBase]:
+        """Yield padded batches of exactly ``num_trajectories_per_batch`` complete trajectories."""
+        from .utils import _traj_ingest, _traj_emit
```
vmoens: No local imports

test/test_collectors.py
```diff
@@ -5296,6 +5302,220 @@ def env_fn():
...
+        """MultiSyncCollector and MultiAsyncCollector honour num_trajectories_per_batch."""
+        max_steps = 4
+        env_fn = lambda: TransformedEnv(  # noqa: E731
+            CountingEnv(max_steps=max_steps), StepCounter(max_steps)
+        )
+        probe_env = env_fn()
+        try:
+            policy = RandomPolicy(probe_env.action_spec)
+        finally:
+            probe_env.close()
```
vmoens: Ditto

torchrl/collectors/_base.py
```diff
@@ -153,6 +153,7 @@ class BaseCollector(IterableDataset, metaclass=abc.ABCMeta):
     _weight_sync_schemes: dict[str, WeightSyncScheme] | None = None
     verbose: bool = False
     _profile_config: ProfileConfig | None = None
+    num_trajectories_per_batch: int | None = None
```
vmoens: Not in docstrings?

test/test_collectors.py
theap06: seems reasonable. will do. thanks for the feedback

test/test_collectors.py
theap06: I'll add an extra case covering that. Thanks for the feedback!

docs/source/reference/collectors_single.rst
theap06: yes we control the internal polling frequency, not the batch size.

docs/source/reference/collectors_single.rst
theap06: Yes

vmoens (approved):
Amazing! 
Now I think a follow up PR would be nice and could cover using this with filling up a replay buffer with trajectories (using collector.sart() for async collection).
This would work best with SliceSampler and such traj-oriented tooling in the RB API.

PR #3619, opened 2026-04-11
[Feature] Introduces a new memmap value for the CKPT_BACKEND environment variable.

### LINKED ISSUE #3617: [Feature Request] Implementation of Safetensors for saving model weights
Motivation

During RL training, TorchRL's trainer periodically checkpoints actor and critic weights using torch.save, which serializes via Python's pickle format. Pickle is unsafe — loading a checkpoint from an untrusted source can execute arbitrary code on the machine. This is a real concern when sharing model weights with collaborators or loading community checkpoints.

Safetensors (now a PyTorch / Linux Foundation project) solves this — it only stores raw tensor data and physically cannot execute code on load. It is already the standard format for sharing model weights on Hugging Face and is faster to load than torch.save.

Solution

Add a SafetensorsCheckpointHook as an opt-in alternative to the existing trainer checkpoint hooks in torchrl/trainers/trainers.py. Users who want safer, faster model checkpointing can swap in the new hook with no other changes to their training code.

# Before
trainer.register_op("post_steps", CheckpointHook(checkpoint_interval=1000))

# After (opt-in)
trainer.register_op("post_steps", SafetensorsCheckpointHook(checkpoint_interval=1000))
Checklist

 I have checked that there is no similar issue in the repo (required)
 I have read the documentation (required)
 I have provided a minimal working example (required)
 SafetensorsCheckpointHook implemented in torchrl/trainers/trainers.py
 Graceful fallback or informative error if safetensors is not installed
 Tests alongside existing checkpoint hook tests
 API reference entry in docs/source/reference/
References

Safetensors: https://github.com/huggingface/safetensors

theap06:
Tensor state dicts are persisted via TensorDict(sd, []).dumps() (pickle-free, no extra dependencies). Plain Python metadata is stored as JSON alongside. Also adds weights_only=True default to the torch backend for safer loading.

## Description

Fixes #3617 

## Motivations

Adds a `memmap` option for the `CKPT_BACKEND` environment variable used by `Trainer`.

**Motivation**: The `torch` backend uses `pickle` under the hood, which has security concerns. The `safetensors` approach requires an extra dependency. `TensorDict.dumps()` is pickle-free, requires no new dependencies, and is already part of the TorchRL stack.

### How to use
```bash
CKPT_BACKEND=memmap python train.py
```

## Changes

- **`torchrl/_utils.py`**: registers `"memmap"` as a valid backend in `_Dynamic_CKPT_BACKEND.backends`
- **`torchrl/trainers/trainers.py`**:
  - `_save_trainer()`: saves tensor state dicts via `TensorDict(sd, []).dumps(path/key)`, plain Python metadata (frame counts etc.) via JSON
  - `load_from_file()`: restores tensor state via `TensorDict.load_memmap()` and metadata from JSON
  - torch backend: adds `weights_only=True` default for safer deserialization
- **`test/test_trainer.py`**: extends the existing checkpoint round-trip parametrize to include `"memmap"`
- **`docs/source/reference/trainers.rst`**: documents the memmap backend option

## Checklist

- [x] Tests added for memmap save/load round-trip
- [x] `weights_only=True` added for torch backend (security improvement)  
- [x] No new dependencies introduced
- [x] Docs updated

### DIFF
--- a/torchrl/trainers/trainers.py
+++ b/torchrl/trainers/trainers.py
@@ -77,39 +77,33 @@ TYPE_DESCR = {float: "4.4f", int: ""}
 REWARD_KEY = ("next", "reward")
 
 
-def _serialize_for_json(obj):
-    """Recursively convert a state-dict to a JSON-serializable form.
+def _state_dict_to_td(sd: dict) -> TensorDict:
+    """Convert a state dict to a :class:`~tensordict.TensorDict`.
 
-    Tensors are encoded as ``{"__tensor__": true, "data": [...], "dtype":
-    "<dtype>", "shape": [...]}`` so they can be faithfully restored.
+    Tensor values are stored directly; everything else is wrapped in
+    :class:`~tensordict.NonTensorData` so that :meth:`~tensordict.TensorDict.dumps`
+    can persist the whole state without pickle dependencies.
     """
-    if isinstance(obj, torch.Tensor):
-        return {
-            "__tensor__": True,
-            "data": obj.tolist(),
-            "dtype": str(obj.dtype),
-            "shape": list(obj.shape),
-        }
-    if isinstance(obj, dict):
-        return {k: _serialize_for_json(v) for k, v in obj.items()}
-    if isinstance(obj, (list, tuple)):
-        serialized = [_serialize_for_json(v) for v in obj]
-        return serialized if isinstance(obj, list) else serialized
-    return obj
+    from tensordict import NonTensorData
+
+    return TensorDict(
+        {
+            k: v if isinstance(v, torch.Tensor) else NonTensorData(v)
+            for k, v in sd.items()
+        },
+        [],
+    )
 
 
-def _deserialize_from_json(obj):
-    """Inverse of :func:`_serialize_for_json`."""
-    if isinstance(obj, dict):
-        if obj.get("__tensor__"):
-            t = torch.tensor(
-                obj["data"], dtype=getattr(torch, obj["dtype"].split(".")[-1])
-            )
-            return t.reshape(obj["shape"]) if obj["shape"] else t.squeeze()
-        return {k: _deserialize_from_json(v) for k, v in obj.items()}
-    if isinstance(obj, list):
-        return [_deserialize_from_json(v) for v in obj]
-    return obj
+def _td_to_state_dict(td: TensorDict) -> dict:
+    """Inverse of :func:`_state_dict_to_td`.
+
+    Unwraps :class:`~tensordict.NonTensorData` back to plain Python values and
+    leaves tensors (including :class:`~tensordict.MemoryMappedTensor`) as-is.
+    """
+    from tensordict import NonTensorData
+
+    return {k: v.data if isinstance(v, NonTensorData) else v for k, v in td.items()}
 
 
 class TrainerHookBase:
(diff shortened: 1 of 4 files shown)


torchrl/trainers/trainers.py
```diff
@@ -504,6 +505,25 @@ def _save_trainer(self) -> None:
...
+
+            state = self.state_dict()
+            path = pathlib.Path(self.save_trainer_file)
+            path.mkdir(parents=True, exist_ok=True)
+            # Persist tensor state dicts (loss_module, collector, registered modules)
+            # using TensorDict memmap — no extra dependencies, no pickle.
+            for key in ("loss_module", "collector"):
+                sd = state[key]
+                if sd:
+                    TensorDict(sd, []).dumps(str(path / key))
```
vmoens: batch-size is not necessary here, just `TensorDict(sd)`

torchrl/trainers/trainers.py
```diff
@@ -504,6 +505,25 @@ def _save_trainer(self) -> None:
...
+            for mod_key in self._modules:
+                sd = state[mod_key]
+                if sd:
+                    TensorDict(sd, []).dumps(str(path / mod_key))
```
vmoens: ditto

torchrl/trainers/trainers.py
```diff
@@ -504,6 +505,25 @@ def _save_trainer(self) -> None:
             Snapshot.take(app_state=self.app_state, path=self.save_trainer_file)
         elif _CKPT_BACKEND == "torch":
             torch.save(self.state_dict(), self.save_trainer_file)
+        elif _CKPT_BACKEND == "memmap":
```
vmoens, line 534: ideally, we would like to unflatten the state-dict keys no?
`unflatten_keys(".")`

torchrl/trainers/trainers.py
```diff
@@ -504,6 +505,25 @@ def _save_trainer(self) -> None:
...
+                sd = state[key]
+                if sd:
+                    TensorDict(sd, []).dumps(str(path / key))
+            for mod_key in self._modules:
+                sd = state[mod_key]
+                if sd:
+                    TensorDict(sd, []).dumps(str(path / mod_key))
+            # Persist non-tensor training counters as JSON.
+            with open(path / "state.json", "w") as f:
+                json.dump(dict(state["state"]), f)
```
vmoens, line 548: what about building one single tensordict with all the modules and self.state_dict() data and dump it with tensordict directly? 
json-friendly data will be serialized well in theory. We can reduce the LoCs here I believe.

theap06: makes senes. thx

torchrl/trainers/trainers.py
```diff
@@ -511,16 +546,20 @@ def _save_trainer(self) -> None:
...
                     TensorDict(sd, []).dumps(str(path / key))
+            # Persist registered hook-module state dicts as JSON.
+            # These can contain mixed types (scalars, bools, tensors), so
+            # tensors are stored as {"__tensor__": true, ...} objects.
+            hooks_state = {}
             for mod_key in self._modules:
-                sd = state[mod_key]
-                if sd:
-                    TensorDict(sd, []).dumps(str(path / mod_key))
+                hooks_state[mod_key] = _serialize_for_json(state[mod_key])
```
vmoens: isn't it what tensordict's MetaData / NonTensor do?

vmoens:
LGTM
before approving / merging can we drop the local imports?

theap06:
> LGTM before approving / merging can we drop the local imports?

Done.

vmoens (approved):

PR #3628, opened 2026-04-15
[Feature] VecNormV2 shared-memory support

vmoens:
* #3638
* #3634
* #3633
* #3632
* #3630
* #3637
* __->__ #3628

### DIFF
--- a/torchrl/envs/transforms/vecnorm.py
+++ b/torchrl/envs/transforms/vecnorm.py
@@ -591,11 +591,15 @@ class VecNormV2(Transform):
         loc = self._loc
         var = self._var
         count = self._count
-        # Handle cross-device: shared-memory stats live on CPU, data may be on GPU
         if loc.device != data.device:
-            loc = loc.to(data.device)
-            var = var.to(data.device)
-            count = count.to(data.device)
+            stats = TensorDict(
+                {"loc": loc, "var": var, "count": count},
+                batch_size=[],
+                device=loc.device,
+            ).to(data.device, non_blocking=True)
+            loc = stats["loc"]
+            var = stats["var"]
+            count = stats["count"]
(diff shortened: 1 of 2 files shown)


torchrl/envs/transforms/vecnorm.py
```diff
@@ -571,7 +588,15 @@ def _norm(self, data, loc, var, count):
...
     def _stateful_norm(self, data):
-        return self._norm(data, self._loc, self._var, self._count)
+        loc = self._loc
+        var = self._var
+        count = self._count
+        # Handle cross-device: shared-memory stats live on CPU, data may be on GPU
+        if loc.device != data.device:
+            loc = loc.to(data.device)
+            var = var.to(data.device)
+            count = count.to(data.device)
```
vmoens: these kinds of device movement can be made faster by using tensordict.to(device) or explicitly setting non_blocking=True and perhaps a cuda sync if it's a D2H

torchrl/envs/transforms/vecnorm.py
```diff
@@ -609,8 +634,9 @@ def _stateful_update(self, data):
                 weight = 1 - self.decay
             else:
                 weight = 1 / count
-        loc.lerp_(end=data_mean, weight=weight)
-        var.lerp_(end=data2, weight=weight)
+        # Handle cross-device: shared-memory stats on CPU, data on GPU
+        loc.lerp_(end=data_mean.to(loc.device), weight=weight)
+        var.lerp_(end=data2.to(loc.device), weight=weight)
```
vmoens: ditto, as above
+ device transfer should be done separately, in a dedicated block, and run async. 
Recall that device transfers break compile so we need to isolate that in an if block

PR #3629, opened 2026-04-15
[Feature] Add StalenessAwareSampler and PolicyVersion export

vmoens:
* #3634
* #3633
* #3632
* #3631
* #3630
* __->__ #3629
* #3628


torchrl/data/replay_buffers/samplers.py
```diff
@@ -294,6 +294,162 @@ def __repr__(self):
         return f"{self.__class__.__name__}({perc: 4.4f}% sampled)"
 
 
+def _default_staleness_weight(s: torch.Tensor) -> torch.Tensor:
+    """Default freshness weighting: 1 / (staleness + 1)."""
+    return 1.0 / (s.float() + 1.0)
+
+
+class StalenessAwareSampler(Sampler):
```
vmoens, line 302: should this inherit from PrioritizedSampler?
not sure how that would work since the staleness changes through time (whereas priority there is fixed)... 
Maybe the priority could just be the policy version but then I wonder if we'd get the same difference in priority between an item created when the policy age was 10 vs 20 or 900 vs 910. In theory the weight difference should be the same but I suspect that with a PrioritizedSampler implementation we'd see a bigger diff in the first case than in the second.

torchrl/data/replay_buffers/samplers.py
```diff
@@ -294,6 +294,162 @@ def __repr__(self):
...
+    def max_staleness(self, v: int):
+        self._max_staleness = v
+
+    def sample(self, storage: Storage, batch_size: int) -> tuple[torch.Tensor, dict]:
+        n = len(storage)
+        if n == 0:
+            raise RuntimeError(_EMPTY_STORAGE_ERROR)
+
+        # Read policy versions from all valid storage entries
+        all_data = storage[:n]
```
vmoens, line 381: I suspect that this can be pretty expensive if the storage is big since we'll be slicing every item in the storage. Also it's useless whenever the storage is at capacity and we're rolling over the values.

torchrl/envs/transforms/__init__.py
```diff
@@ -75,6 +75,15 @@
 from .vecnorm import VecNormV2
 from .vip import VIPRewardTransform, VIPTransform
 
+
+def __getattr__(name):
+    if name == "PolicyVersion":
+        from torchrl.envs.llm.transforms.policy_version import PolicyVersion
+
+        return PolicyVersion
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```
vmoens, line 84: this isn't beautiful, why do we need it?

PR #3631, opened 2026-04-15
[Feature] Evaluator collector backend and process shared-memory sync

vmoens:
* #3634
* #3633
* #3632
* __->__ #3631
* #3630
* #3637
* #3628

### DIFF
--- a/torchrl/collectors/_evaluator.py
+++ b/torchrl/collectors/_evaluator.py
@@ -351,19 +351,40 @@ class Evaluator:
         self,
         weights: TensorDictBase | nn.Module | None = None,
         step: int | None = None,
-    ) -> None:
-        """Start an async evaluation (fire-and-forget).
+    ) -> dict[str, Any] | None:
+        """Start an async evaluation, auto-logging the previous result.
 
-        If a previous evaluation is still running its result will be
-        discarded.
+        If a previous evaluation has completed, its result is finalized
+        (logged via the configured logger, video dumped, callback invoked)
+        before the new evaluation is submitted.  This means the caller
+        never needs to call :meth:`poll` explicitly — just keep calling
+        ``trigger_eval`` and previous results are logged automatically.
+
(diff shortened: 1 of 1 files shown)


torchrl/collectors/_evaluator.py
vmoens, line 1: these edits lack a test

torchrl/collectors/_evaluator.py
```diff
@@ -1067,25 +1097,35 @@ def _process_eval_worker(
         **(collector_kwargs or {}),
     )
 
+    import time as _time
```
vmoens: no local imports

torchrl/collectors/_evaluator.py
vmoens: also I don't like it much that we have so much custom logic for weight updates. The evaluator should only wrap the collector and rely on the collector update scheme to send the weights to the model/env/replay buffer.

torchrl/collectors/_evaluator.py
```diff
@@ -1135,8 +1175,10 @@ def __init__(
         ctx = mp.get_context("spawn")
         self._request_queue: mp.Queue = ctx.Queue(maxsize=1)
         self._result_queue: mp.Queue = ctx.Queue(maxsize=1)
+        self._shared_weights_queue: mp.Queue = ctx.Queue(maxsize=1)
```
vmoens: weights can be pretty big, I don't think we should queue anything.

torchrl/collectors/_evaluator.py
```diff
@@ -1135,8 +1175,10 @@ def __init__(
...
         self._pending_flag = False
         self._last_step: int | None = None
+        self._shared_weights: TensorDictBase | None = None
```
vmoens: why "shared weights" everywhere in this PR? We can send weights that are not shared and the weight sync scheme will do its job

PR #3632, opened 2026-04-15
[Feature] Add RandomTruncationTransform

vmoens:
* #3638
* #3634
* #3633
* __->__ #3632
* #3630
* #3637
* #3628


torchrl/envs/transforms/transforms.py
```diff
@@ -8085,6 +8085,177 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+    otherwise the full ``max_horizon`` is used.
+
+    ``first_episode_prob`` controls the truncation probability for each
+    environment's first episode after the initial spread.  Setting it to 1.0
+    (default) ensures every environment goes through a second short-horizon
+    episode before settling to ``prob``, which accelerates decorrelation when
+    batch sizes are large relative to ``max_horizon``.
+
+    .. note:: This transform must be placed **after** :class:`~torchrl.envs.StepCounter`
+        in the transform chain, as it relies on the ``"step_count"`` key.
```
vmoens, line 8109: what happens if there is no step_counter? this is something we can check ahead of time by looking at the parent env's specs if they're available. I would hook the parent registration with that check (hooks like that should be a feature of the Transform class, that could be a nice public method to add as part of this PR).

torchrl/envs/transforms/transforms.py
```diff
@@ -8085,6 +8085,177 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+
+    Args:
+        prob (float, optional): probability of sampling a random horizon on
+            each subsequent reset.  Defaults to ``0.5``.
```
vmoens: make it 0.00 and make sure that we're not doing anything if it's 0.0
The point being that one could just truncate the first episodes.
Advise users to use a low value (order of 1% or so) given that there are many episodes to be collected typically and frequently truncating massively impacts negatively training perf.

torchrl/envs/transforms/transforms.py
```diff
@@ -8085,6 +8085,177 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+        min_horizon (int, optional): minimum horizon for random truncation
+            (inclusive).  Defaults to ``500``.
```
vmoens: no default, very task dependent.

torchrl/envs/transforms/transforms.py
```diff
@@ -8085,6 +8085,177 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+        max_horizon (int, optional): maximum horizon for random truncation
+            (inclusive).  Also used as the full-length horizon when no random
+            truncation is applied.  Defaults to ``1000``.
```
vmoens: no default, very task dependent.
mention that this should match the env horizon, which, in general, cannot be retrieved automatically unfortunately.

torchrl/envs/transforms/transforms.py
```diff
@@ -8085,6 +8085,177 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+                effective_prob = torch.where(
+                    first_ep,
+                    torch.tensor(self.first_episode_prob, device=self._horizons.device),
+                    torch.tensor(self.prob, device=self._horizons.device),
+                )
+                keep_full = torch.rand(n, device=self._horizons.device) > effective_prob
+                new_h[keep_full] = self.max_horizon
+                self._horizons[mask] = new_h.view_as(self._horizons[mask])
+                # First episode is over for these envs
+                self._first_episode[mask] = False
```
vmoens, line 8269: ok i like this because it's compile friendly. Even if guarding when self.prob is 0 could allow us to do fewer ops, it won't be as compile friendly as this so let's keep it that way.

torchrl/envs/transforms/transforms.py
vmoens: We can also make people's life easier by using `__new__` and ABC classes to add a `Compose(StepCounter(), RandomTruncationTransform())` if this is added without that.

PR #3637, opened 2026-04-15
[Feature] Add StalenessAwareSampler

vmoens:
* #3638
* #3634
* #3633
* #3632
* #3630
* __->__ #3637
* #3628


torchrl/data/replay_buffers/samplers.py
```diff
@@ -294,6 +294,178 @@ def __repr__(self):
         return f"{self.__class__.__name__}({perc: 4.4f}% sampled)"
 
 
+def _default_staleness_weight(s: torch.Tensor) -> torch.Tensor:
+    """Default freshness weighting: 1 / (staleness + 1)."""
+    return 1.0 / (s.float() + 1.0)
+
+
+class StalenessAwareSampler(Sampler):
```
vmoens, line 302: We should mention PolicyVersion and give a more complete example in the docstrings of how to use this with policy version, even with a silly pseudo-training loop. It's important that users know exactly what to do.
All new classes must be referenced in the API ref in `docs/`. It would be nice also to comment in the docs about this (how to integrate collectors and replay buffer's sampler to keep track of staleness and policy versions).

PR #3638, opened 2026-04-15
[Feature] Evaluator result callbacks and busy policy

vmoens:
* __->__ #3638
* #3634
* #3633
* #3632
* #3630
* #3637
* #3628

### DIFF
--- a/sota-implementations/ppo-async/train.py
+++ b/sota-implementations/ppo-async/train.py
@@ -13,7 +13,6 @@ from __future__ import annotations
 
 import logging
 import multiprocessing
-import threading
 import time
 from functools import partial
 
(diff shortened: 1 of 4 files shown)


sota-implementations/ppo-async/train.py
```diff
@@ -49,6 +50,38 @@ def _make_eval_policy(env, env_name, device):
...
+        step = int(metrics["eval/step"])
+        started_at = pending_eval_start.pop(step, None)
+        if started_at is not None:
+            dt = time.time() - started_at
+            if dt > 0:
+                metrics["eval/fps"] = (num_eval_envs * 1000) / dt
+        with logger_lock:
+            logger.log_metrics(metrics, step)
+
+    return mark_eval_trigger, on_result
```
vmoens: that's a bit much. Why don't we measure this by default in the evaluator metrics? We have a collector from which we can measure the fps no?

sota-implementations/ppo-async/train.py
```diff
@@ -136,10 +172,12 @@ def train_start(
         num_trajectories=num_eval_envs,
         max_steps=1000,
         backend="process",
+        on_result=on_eval_result,
     )
 
     # Start collection and evaluation
     collector.start()
+    mark_eval_trigger(0)
```
vmoens: this isn't great, it pollutes the code. See my comment on making the fps measure a part of the evaluator

sota-implementations/ppo-async/train.py
```diff
@@ -397,9 +412,12 @@ def train_iterate(
         postproc=adv_module,
     )
 
+    logger_lock = threading.Lock()
```
vmoens: I don't like it that the script needs its own logger lock

PR #3679, opened 2026-04-28
[BugFix] Fix PPOTrainer gamma/lmbda defaults, remove dead code, fix wildcard import

theap06:
##Fix

Wrong gamma default (0.9 → 0.99): The discount factor was set to 0.9, causing the agent to heavily discount future rewards and perform significantly worse than expected on standard benchmarks. Every major PPO reference implementation (Stable Baselines3, CleanRL, the existing sota-implementations/ppo/) uses gamma=0.99.

Wrong lmbda default (0.99 → 0.95): The GAE lambda was inverted from the standard value. lmbda=0.99 produces high-variance advantage estimates close to Monte Carlo returns. The original GAE paper and all standard implementations use lmbda=0.95.

Dead try/except blocks: Two empty try: pass blocks that unconditionally set _has_tqdm=True and _has_ts=True regardless of whether those packages are installed. Removed entirely.

Wildcard import in train.py: from torchrl.trainers.algorithms.configs import * replaced with an explicit import to avoid silent name collisions.

WandB as default logger: Changed default logger from wandb to csv so the training script works out of the box without WandB credentials.

### DIFF
--- a/sota-implementations/ppo_trainer/config/config.yaml
+++ b/sota-implementations/ppo_trainer/config/config.yaml
@@ -27,7 +27,7 @@ defaults:
   - trainer@trainer: ppo
   - optimizer@optimizer: adam
   - loss@loss: ppo
-  - logger@logger: wandb
+  - logger@logger: csv
   - _self_
 
 # Network configurations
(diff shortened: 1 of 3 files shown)


vmoens (approved):
LGTM thanks

PR #3682, opened 2026-04-28
[BugFix] Handle PettingZoo state and add regression test for encoding

### LINKED ISSUE #3681: [Feature Request] Support in pettingzoo envs .state() function for not only ndarray
## Motivation

Enable user, to use .state that are not only ndarray but as flexible as observaiton or action state for pettingzoo environement. Historicly pettingzoo .state official API tell to use ndarray. But the code allow to put any state, state space (which is also partly true for torchrl, at one caviat). Pettingzoo will update this https://github.com/Farama-Foundation/PettingZoo/issues/1326, so the change is futur proof

## Solution

replace in pettingzooWrapper:
`if self.return_state:
            state = torch.as_tensor(self.state(), device=self.device)
            tensordict_out.set("state", state)`
with:
`if self.return_state:
            state = self.observation_spec["state"].encode(self.state())
            tensordict_out.set("state", state)`
to be consistant with the rest

## Checklist

- [ ] I have checked that there is no similar issue in the repo (**required**)

MathieuFonsProjects:
close #3681

## Types of changes

## Checklist

### DIFF
--- a/test/libs/test_multiagent.py
+++ b/test/libs/test_multiagent.py
@@ -167,7 +167,7 @@ class TestPettingZoo:
     @pytest.mark.parametrize("task", ["simple_v3"])
     def test_return_state(self, task):
         """Test return_state=True returns state properly from raw env.state().
-        
+
         Regression test for handling PettingZoo state encoding.
         Verifies that the state returned in tensordict matches the raw state
         from env.state(), ensuring proper conversion of any state type.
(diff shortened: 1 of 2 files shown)


vmoens (commented):
LGTM, let's just fix the linter

MathieuFonsProjects:
Sorry @vmoens my pre-commit wasn't installed correctly... It's good now (I had linter issue in files I havn't touched ?)

PR #3689, opened 2026-04-28
[Feature] Add Safety-Gymnasium environment wrapper

### LINKED ISSUE #509: [DO NOT CLOSE] Call for contributions
This issue is a list of contributions requests from the community. 

## How to use this list

If you're willing to contribute to the library, have a look at the list below and don't hesitate to pick up a task.
If you need guidance, refer to @vmoens for more information!

If you want to add an item to this list, start by raising an issue and mention that you think it would be appropriate to put it in the "call for contributions" stack.

## New algorithms
New algorithms can be coded either in a free form or using torchrl's trainer class.
In any case, we would ask to the user to use hydra for the configuration, and to limit the number of extra dependencies as much as can be.

- [ ] A2C algorithm #17 
- [ ] TD3 algorithm #18 
- [ ] Decision transformers #15 
- [ ] PILCO (no open issue yet)
- [ ] Image augmentation is all you need #32 

## New environment libraries
- [ ] SMAC (Starcraft Multi-agent challenge)
- [ ] Safety gym compatibility

## New modules and features
- [ ] Reward-to-go #16 

Thanks for contributing to TorchRL!

theap06:
Adds SafetyGymnasiumWrapper and SafetyGymnasiumEnv for constrained-RL benchmarks. The underlying step API returns a 6-tuple with a parallel cost signal; the wrapper folds cost into the gymnasium info dict so the existing GymWrapper machinery is reused, then registers an info-dict reader that exposes cost as a top-level tensordict key alongside reward.
##Motivation 
Fixes #509 

Adds SafetyGymnasiumWrapper and SafetyGymnasiumEnv for constrained-RL
benchmarks. The underlying step API returns a 6-tuple with a parallel
cost signal; the wrapper folds cost into the gymnasium info dict so the
existing GymWrapper machinery is reused, then registers an info-dict
reader that exposes "cost" as a top-level tensordict key alongside
"reward".


torchrl/envs/libs/safety_gymnasium.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def _make_cost_reader() -> default_info_dict_reader:
+    cost_spec = Composite(
+        cost=Unbounded(shape=(), dtype=torch.float64), shape=[]
+    )
+    return default_info_dict_reader(["cost"], spec=cost_spec)
+
+
+class SafetyGymnasiumWrapper(GymWrapper):
```
vmoens: Missing from the docs

test/libs/test_safety_gymnasium.py
vmoens, line 1: we need a dedicated libs testing pipeline in the CI

torchrl/envs/libs/safety_gymnasium.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+
+    @_classproperty
+    def available_envs(cls):
+        if not _has_safety_gymnasium:
+            return []
+        # Curated list of canonical safety-gymnasium task ids. The library
+        # registers more (different difficulty levels, vision variants, etc.);
+        # this list mirrors the ones documented as primary benchmarks.
+        return [
```
vmoens: could we discover this automatically?

torchrl/envs/libs/safety_gymnasium.py
```diff
@@ -0,0 +1,188 @@
...
+        super().__init__(env_name=env_name, **kwargs)
+        self.set_info_dict_reader(_make_cost_reader())
+
+    def _output_transform(self, step_outputs_tuple):
+        observations, reward, cost, terminated, truncated, info = step_outputs_tuple
+        info = dict(info) if info is not None else {}
+        # The default info_dict_reader expects values with a `.dtype`
+        # attribute. safety-gymnasium emits cost as a Python float, so we
+        # promote it to a numpy scalar of fixed dtype.
+        info["cost"] = np.asarray(cost, dtype=np.float64)
```
vmoens: why np?

torchrl/envs/libs/safety_gymnasium.py
```diff
@@ -0,0 +1,188 @@
...
+            raise ImportError(
+                "safety-gymnasium not found, install with "
+                "`pip install safety-gymnasium`"
+            ) from err
+
+    _make_specs = set_gym_backend("gymnasium")(GymEnv._make_specs)
+
+    def __init__(self, env_name=None, **kwargs):
+        super().__init__(env_name=env_name, **kwargs)
+        self.set_info_dict_reader(_make_cost_reader())
```
vmoens: do we need an info reader? Shouldn't we package this with the observation instead since it ends up there anyway? 
Both are fine if this is easier / fewer LoC

torchrl/envs/libs/safety_gymnasium.py
theap06: Should i set it to torch tensor?

test/libs/test_safety_gymnasium.py
theap06: Should i add this to this pr or have a separate PR for the CI?

vmoens (approved):
LGTM

ParamThakkar123:
Looks like the original issue for call for contributions got closed because the PR comment mentioned that this PR fixes the issue

PR #3692, opened 2026-04-29
[Feature] Add early stopping trainer hook

### LINKED ISSUE #2370: [Feature Request] `EarlyStopping` for `torchrl.trainers.Trainer`
## Motivation

Often times we only want to train an algorithm until it learned the intended behavior, and a total number of frames is only a proxy for the stopping condition.
I would like to propose adding a new callback that makes early stopping possible, much like the example in [Lightning](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.EarlyStopping.html#lightning.pytorch.callbacks.EarlyStopping).

Currently, no callback can influence the end of the training loop, so my current workaround is setting  `self.total_frames` to 1, which isn't great.

## Solution

- [ ] Adding a new early-stopping-hook
- [ ] Creating a default hook implementation for `TrainerHookBase`
- [ ] Stopping the loop if the hook gives a positive return

The main question for me would be where to call the hook and what it can observe.
I would argue that both the `batch` and the `losses_detached` would be a relevant signal for the stopping hook, as one might want to stop when a certain return was seen n-times or the loss is smaller than a threshold.
Therefore, either the call would happen in `optim_steps(self, batch: TensorDictBase) -> None` and the function returns the stopping signal to `def train(self)` or the hook is called in `train` and `optim_steps` returns the loss such that the early stopping hook can observe it.

An example for the second approach could look something like:
```python
def train(self):
    ...
    average_losses = None
    if self.collected_frames > self.collector.init_random_frames:
        average_losses = self.optim_steps(batch)
    
    if self.early_stopping_hook(batch,average_losses):
      self.save_trainer(force_save=True)
     break
    
    self._post_steps_hook()
   ...
```

I am happy to open a PR if this is something of interest to you.

## Checklist

Xmaster6y:
## Description

Add a new trainer hook which monitors a logged quantity and stops training when conditions are met.

## Motivation and Context

Useful to fail fast or quit when the task is solved.

Closes #2370 

## Types of changes

## Checklist


test/test_trainer.py
```diff
@@ -1436,6 +1437,103 @@ def shutdown_hook():
...
+        self, monitor, values, hook_kwargs, expected_stops, expected_reason
+    ):
+        trainer = mocking_trainer(optimizer=None)
+        early_stopping = EarlyStopping(
+            monitor=monitor,
+            **hook_kwargs,
+        )
+        early_stopping.register(trainer)
+
+        td = TensorDict({}, [])
```
vmoens: Nit: TensorDict() is equally valid

torchrl/trainers/trainers.py
```diff
@@ -2353,3 +2367,143 @@ def state_dict(self) -> dict[str, Any]:
...
+
+    Examples:
+        >>> LogScalar(("next", "reward"), "r_training").register(trainer)
+        >>> EarlyStopping(monitor="r_training", patience=10_000).register(trainer)
+    """
+
+    def __init__(
+        self,
+        *,
+        monitor: str = "r_evaluation",
```
vmoens: NestedKey?

torchrl/trainers/trainers.py
```diff
@@ -2353,3 +2367,143 @@ def state_dict(self) -> dict[str, Any]:
...
+        mode: str = "max",
```
vmoens: Literal?

vmoens (approved):
A few nit to solve otherwise LGTM

PR #3683, opened 2026-04-28
[Feature] Make `DDPG`, `PPO` and `SAC` trainers multi-agent-friendly

Xmaster6y:
## Description

- Fix `VMAS` broken config, add instantiation tests and add `scenario_kwargs ` for passing arbitrary parameters from the config
- Add common utilities to convert `ListConfig` to nested keys and add tests
- Enable passing custom keys to trainers
- Remove average_losses unwanted log
- Add `sota-implementations` examples

## Motivation and Context

This change is required to swiftly use the trainers for multi-agent training.

PR #3693, opened 2026-04-29
[Feature] TrainerConfig / Trainer parity audit + auto_log_optim_steps plumbing

vmoens:
## Summary

- Plumb the new `auto_log_optim_steps` kwarg through all 7 algorithm-specific `Trainer` subclasses (SAC, PPO, DQN, DDPG, IQL, CQL, TD3) so it is reachable from each subclass `__init__` and forwarded into the base `Trainer`.
- Audit every `TrainerConfig` and surface kwargs that the matching `Trainer.__init__` accepts but the Config never exposed via Hydra: `enable_logging`, `log_rewards`, `log_actions`, `log_observations` across all algos; `batch_size` on SAC/PPO; `gamma`/`lmbda` on PPO; `frame_skip`/`progress_bar`/`save_trainer_interval`/`log_interval`/`num_epochs`/`log_timings` on TD3.
- Add bidirectional Sphinx `:class:` cross-references between each `Trainer` subclass and its `TrainerConfig` so the round-trip is discoverable from either side.
- New `CLAUDE.md` / `AGENTS.md` section 16 codifying the Config-class parity invariant (parity, cross-references, where to grep) so future kwarg additions can't silently drift.

The new `auto_log_optim_steps` constructor flag (added in #3683) is a base-`Trainer` kwarg that controls whether `optim_steps` and the averaged-loss tensordict are auto-logged at the end of each optimization loop. It was unreachable from any subclass and from any Hydra config; this PR closes that gap and applies the parity rule retroactively to other long-missing kwargs surfaced during the audit.

## Test plan

- [x] `pytest test/test_trainer.py -q` — 67 passed (was 60 before this PR, +7 from the new parametrized subclass-signature test).
- [x] `pytest test/test_configs.py -k 'not Hydra' -q` — 88 passed (no regression in Hydra-instantiation tests; the slow DDPG subprocess parsing test is a pre-existing macOS slowness unrelated to this change).
- [x] `python -c "import torchrl.trainers.algorithms.configs.trainers"` — clean import.
- [x] `pytest test/test_configs.py::TestHydraParsing::test_dqn_trainer_parsing_with_file` — Hydra DQN round-trip still passes (9.5s).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -160,3 +160,24 @@ all.
 
 Read a recently-merged PR in the same area and match the conventions there
 before inventing your own.
+
+## 16. Config / class parity
+
+Some classes (Trainers under `torchrl/trainers/algorithms/`, losses, replay-buffer
+components, transforms, …) have a Hydra `*Config` dataclass companion under
+`torchrl/trainers/algorithms/configs/`. The Config exists so users can construct
+the class via Hydra YAML — it is a faithful façade.
+
+- **Parity.** Every kwarg accepted by the wrapped class's `__init__` must
+  appear as a field on the Config (with the same default), be popped from
+  `kwargs` inside the matching `_make_*` factory, and be forwarded into the
+  constructor. Adding a kwarg to the class without surfacing it in the Config
+  silently breaks Hydra users.
+- **Cross-references.** Each Config docstring must reference its wrapped class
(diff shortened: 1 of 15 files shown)


vmoens:
@Xmaster6y for review

Xmaster6y:
LGTM, much cleaner

Xmaster6y (changes requested):
In fact, it would be great to extend the parity to all configs to avoid broken ones. For example, previously, `VmasEnvConfig` was broken, but I expect others might be.

PR #3694, opened 2026-04-29
[Feature] Add QMix, VDN and IQL support to DQN trainer

Xmaster6y:
## Description

- Extend `DQNTrainer` to support custom key and potential reward aggregation
- Add mixer networks and loss configs with custom builders
- Expose customisation in `QValueActor `
- Add new sota-implementations for QMix, VDN and IQL
- Sort config tests (module and loss)

## Motivation and Context

This change is required for better trainer support of multi-agent algorithms.

## Types of changes

## Checklist


torchrl/trainers/algorithms/dqn.py
```diff
@@ -189,6 +212,9 @@ def __init__(
         self.log_rewards = log_rewards
         self.log_observations = log_observations
 
+        if self.mixing_strategy in ("qmix", "vdn"):
```
theap06, line 244: if self.mixing_strategy in ("qmix", "vdn"):   # "iql" missing
    self.register_op("batch_process", self._aggregate_agent_rewards)

torchrl/trainers/algorithms/dqn.py
```diff
@@ -199,10 +225,21 @@ def _step_greedy(self):
...
 
+    def _aggregate_agent_rewards(self, batch: TensorDictBase) -> TensorDictBase:
+        for key in (self.reward_key, self.episode_reward_key):
+            next_key = ("next", *key) if isinstance(key, tuple) else ("next", key)
+            if batch.get(next_key, None) is not None:
+                continue
+            source_key = ("next", "agents", key[-1] if isinstance(key, tuple) else key)
+            value = batch.get(source_key, None)
+            if value is not None:
+                batch.set(next_key, value.mean(-2))
```
theap06: batch.set(next_key, value.mean(-2))

torchrl/modules/tensordict_module/actors.py
```diff
@@ -1225,12 +1235,12 @@ def __init__(
             spec = Composite()
         if isinstance(spec, Composite):
             spec = spec.clone()
-            if "action" not in spec.keys():
-                spec["action"] = None
+            if action_key not in spec.keys(True, True):
```
theap06, line 1238: spec logic is inverted and nested key wrapping is wrong. 
The else branch fires when the key already exists and replaces the whole spec with a new Composite wrapping the old one, backwards. Also Composite({("agents", "action"): spec}) won't create proper nesting for tuple keys.

torchrl/trainers/algorithms/dqn.py
Xmaster6y: no mixing for iql

torchrl/modules/tensordict_module/actors.py
Xmaster6y: Didn't get you sorry

torchrl/trainers/algorithms/dqn.py
Xmaster6y: -2 is the agent dim

torchrl/modules/tensordict_module/actors.py
theap06: Like if the value keys aren't passed explicitly, they default to flat "action_value" / "chosen_action_value" instead of ("agents", "action_value") / ("agents", "chosen_action_value"). This breaks QMixerLoss. The value key defaults should mirror action_key's namespace.

Xmaster6y: where do they default to flat?

theap06: tbh not really a big issue - was just thinking more in the situation where checking whether action_key is nested, so they always resolve to flat strings.

PR #3695, opened 2026-04-30
[Feature] sequence rl composability

### LINKED ISSUE #3678: [RFC]Making sequence-based RL first-class in TorchRL
##Motivation 
TorchRL has all the building blocks for sequence-based RL — SliceSampler, split_trajectories, LSTMModule, TensorDictPrimer, traj_ids — but they are not designed to compose automatically. Users who want to train RNN or sequence-based policies (ACT, Decision Transformer, recurrent PPO) must manually wire them together and get the masking, trajectory key alignment, and hidden state resets right themselves. This RFC proposes making that pipeline explicit, safe, and ergonomic by default.

##Solution

1. Standardize trajectory key defaults
Align SliceSampler's default traj_key to auto-detect ("collector", "traj_ids") when present, falling back to "episode" for backward compatibility. This fixes a silent composability bug with no API break.

2. Add episode_start metadata from collectors
Emit ("collector", "episode_start") — a boolean flag marking the first step of each episode. Downstream code should not have to re-derive episode boundaries from done repeatedly.

3. Sequence sampling with built-in mask and lengths
Extend SliceSampler to return [B, T] TensorDicts with:

("collector", "mask")     # bool [B, T] — False on padded timesteps
("collector", "lengths")  # int  [B]
when strict_length=False. Users should not need to call split_trajectories() manually.

4. Safe hidden state reset for recurrent policies
Recurrent policies currently leak hidden state across episode boundaries unless the user manually configures TensorDictPrimer + is_init. The target behavior: if episode_start or is_init is present, hidden state resets automatically at those positions.

5. Mask-aware losses
Common loss modules should respect ("collector", "mask") if present, so padded timesteps do not contribute to the gradient. Start narrow — behavior cloning / imitation losses first.

theap06:
Fixes #3678 

Motivation and Fix

- This PR addresses three composability gaps that made it unnecessarily hard to train RNN/sequence-based policies (LSTM, GRU, ACT, Decision Transformer) with TorchRL's existing building blocks.


examples/replay-buffers/recurrent_slice_sampler_pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+    in_keys=["embed_out"],
+    out_keys=["action_value"],
+)
+qval = QValueModule(action_space="categorical")
+policy = Seq(OrderedDict(embed=embed, gru=gru, mlp=mlp, qval=qval))
+
+# ---------------------------------------------------------------------------
+# Env: one call hooks up InitTracker + TensorDictPrimer for the GRU.
+# ---------------------------------------------------------------------------
+env = TransformedEnv(GymEnv("CartPole-v1"), get_env_transforms_from_module(policy))
```
vmoens: @theap06 what about we add a `policy` arg in `EnvBase`, picked up by the metaclass (such that ALL subclasses can use it) and that calls `get_env_transforms_from_module` within the metaclass call, and returns a TransformedEnv.
Would make this equivalent to
```
env = GymEnv("...", policy=policy)
```

examples/replay-buffers/recurrent_slice_sampler_pipeline.py
vmoens, line 1: after a long fight with Claude i got to this, @theap06 eager to know what you think of it

examples/replay-buffers/recurrent_slice_sampler_pipeline.py
vmoens: We could even get the Collector to do that for us! 
Then we would need to make 100% sure that get_env_transforms_from_module does not give duplicated transforms.

examples/replay-buffers/recurrent_slice_sampler_pipeline.py
theap06: lmao

theap06: the is_init marking at slice starts is something I hadn't thought of. LGTM

examples/replay-buffers/recurrent_slice_sampler_pipeline.py
theap06: Love it! Only concern is duplicate transforms if the Collector also calls get_env_transforms_from_module on the same policy. Not a major concern though

vmoens:
**Heads-up on the breadth of the auto-wrap change** (raised after a CI failure I want to flag explicitly):

The collector-side auto-wrap in this PR is wired up via `_compute_missing_env_transforms`, which walks the *whole policy* (`module.apply(...)`) and calls `make_tensordict_primer()` on every submodule that exposes one. That ended up being more intrusive than I anticipated:

* It's not limited to `LSTMModule` / `GRUModule`. Anything that exposes `make_tensordict_primer()` is invoked.
* Some primer-bearing modules need explicit configuration to build their primer. `ConsistentDropoutModule` is the concrete example we hit: without `input_shape` it raises `RuntimeError`. Pre-PR the collector never ran this code path, so policies containing `ConsistentDropoutModule` worked fine through the collector. With this PR, even the *dry-run* (used to decide whether to emit the `FutureWarning`) called `make_tensordict_primer()` and failed.

Mitigations already in this branch:

1. The default for `auto_register_policy_transforms` is `None` through v0.14: the helper only runs as a dry-run to decide whether to emit a `FutureWarning`. The default flips to `True` in v0.15. Users who want today's behavior immediately pass `True`; users who want to opt out forever pass `False`.
2. The dry-run wraps `get_primers_from_module` in `try/except (RuntimeError, ValueError)` and falls back to "no auto-wired primers" for that policy. `InitTracker` detection still runs. So a policy with a `ConsistentDropoutModule` no longer breaks the collector — it just doesn't get an auto-wired dropout primer (which is the same behavior it had before this PR).

Worth a closer look during review:

* Are there other primer-bearing modules in the wild whose `make_tensordict_primer()` has lazy/conditional init like `ConsistentDropoutModule`? If yes, those will silently not be auto-wired with the current fallback. Acceptable IMO (the user can wire them manually), but worth documenting.
* Is `module.apply(...)` the right granularity? An alternative would be to look only at top-level submodules of `TensorDictSequential`, which would miss nested-but-attached recurrent submodules but also avoid surprises like this one.
* The two-release deprecation window mitigates the BC risk but doesn't eliminate it. Anyone passing a recurrent policy + a bare env to a collector today is currently broken silently — they get a `FutureWarning` and a downstream `KeyError: 'is_init'` rather than a working collector. That's the *intentional* state through v0.14, but the warning is the only thing telling them.

Happy to tighten any of these if it feels too aggressive.

theap06:
> **Heads-up on the breadth of the auto-wrap change** (raised after a CI failure I want to flag explicitly):
> 
> The collector-side auto-wrap in this PR is wired up via `_compute_missing_env_transforms`, which walks the _whole policy_ (`module.apply(...)`) and calls `make_tensordict_primer()` on every submodule that exposes one. That ended up being more intrusive than I anticipated:
> 
> * It's not limited to `LSTMModule` / `GRUModule`. Anything that exposes `make_tensordict_primer()` is invoked.
> * Some primer-bearing modules need explicit configuration to build their primer. `ConsistentDropoutModule` is the concrete example we hit: without `input_shape` it raises `RuntimeError`. Pre-PR the collector never ran this code path, so policies containing `ConsistentDropoutModule` worked fine through the collector. With this PR, even the _dry-run_ (used to decide whether to emit the `FutureWarning`) called `make_tensordict_primer()` and failed.
> 
> Mitigations already in this branch:
> 
> 1. The default for `auto_register_policy_transforms` is `None` through v0.14: the helper only runs as a dry-run to decide whether to emit a `FutureWarning`. The default flips to `True` in v0.15. Users who want today's behavior immediately pass `True`; users who want to opt out forever pass `False`.
> 2. The dry-run wraps `get_primers_from_module` in `try/except (RuntimeError, ValueError)` and falls back to "no auto-wired primers" for that policy. `InitTracker` detection still runs. So a policy with a `ConsistentDropoutModule` no longer breaks the collector — it just doesn't get an auto-wired dropout primer (which is the same behavior it had before this PR).
> 
> Worth a closer look during review:
> 
> * Are there other primer-bearing modules in the wild whose `make_tensordict_primer()` has lazy/conditional init like `ConsistentDropoutModule`? If yes, those will silently not be auto-wired with the current fallback. Acceptable IMO (the user can wire them manually), but worth documenting.
> * Is `module.apply(...)` the right granularity? An alternative would be to look only at top-level submodules of `TensorDictSequential`, which would miss nested-but-attached recurrent submodules but also avoid surprises like this one.
> * The two-release deprecation window mitigates the BC risk but doesn't eliminate it. Anyone passing a recurrent policy + a bare env to a collector today is currently broken silently — they get a `FutureWarning` and a downstream `KeyError: 'is_init'` rather than a working collector. That's the _intentional_ state through v0.14, but the warning is the only thing telling them.
> 
> Happy to tighten any of these if it feels too aggressive.

I'll add more exceptions per submodule and write up test cases to determine the issues when considering the whole walk. Thanks for the feedback! Scope makes sense

theap06:
@vmoens lmk what you think! Thanks for the detailed feedback.

docs/source/reference/data_layout.rst
```diff
@@ -0,0 +1,350 @@
...
+write into the same storage concurrently — e.g.
+:class:`~torchrl.collectors.MultiCollector` ``(sync=False)``,
+:class:`~torchrl.collectors.distributed.RayCollector`, an external pool of
+producers, or any cluster setup where a learner aggregates batches from
+many actors — the order in which workers' batches land in the storage
+**cannot be controlled**. A given row of the ``[N, T]`` storage would end
+up with ``T`` frames stitched together from different workers and
+different episodes, with no marker to tell them apart and the
+``SliceSampler`` silently drawing slices that straddle unrelated
+trajectories. The same issue arises if any postprocessing rearranges the
```
vmoens: there's a set_truncated arg to prevent that.
The only problem is really that we introduce artificial boundaries and reduce the effective samplable size of the trajectories produced by the collectors.

There's a mitigation which is to use an ensemble of replay buffers with one storage / collector. Worth an example IMO.

docs/source/reference/data_layout.rst
```diff
@@ -0,0 +1,350 @@
...
+For multi-process / asynchronous / cross-cluster collection, the only
+robust storage shape is ``ndim=1``.
+
+Concretely, the ``ndim`` arg works for:
+
+* Single-process :class:`~torchrl.collectors.Collector` with a batched
+  env (one process writes; the env dim is stable).
+* Synchronous :class:`~torchrl.collectors.MultiCollector` ``(sync=True)``
+  with ``cat_results="stack"``, which delivers one ``[num_workers, T]``
+  batch at a time *atomically*.
```
vmoens, line 192: will that work if the replay buffer is passed to the collector?
I assume the replay buffer is extended within the individual collectors which invalidates what you're writing here. If not, the difference in how the RB is populated between sync and async collection needs to be made super duper clear.

docs/source/reference/data_layout.rst
```diff
@@ -0,0 +1,350 @@
...
+
+It does **not** work for:
```
vmoens: this needs to be soften given the set_truncated arg i told you about

docs/source/reference/data_layout.rst
```diff
@@ -0,0 +1,350 @@
...
+where transforms live in child envs. Limitations and the v0.15 default
+flip are documented in the
+:ref:`policy= argument page <Environment-policy-arg>`.
+
+Net effect: a recurrent training pipeline rarely needs to touch
+``InitTracker`` or ``TensorDictPrimer`` by hand. The user wires the
+policy, hands it to the env or collector, and the boundary keys + hidden
+state the rest of the stack expects are all in place.
+
+Legacy: ``split_trajectories``
```
vmoens, line 377: another issue with split_traj is that you can have a traj that starts at step>0 (because it spans across the previous and current batch) or ends before the done state (because it spans across the current and next batch). That means we need to reset is_init and/or truncated/done signals. This isn't great and conditions the data formatting on the hyperparams of the collector which we want to avoid

docs/source/reference/data_layout.rst
vmoens: Fixed in 3cefe03b6. Reframed the ndim>=2 paragraph: it now says 'runs into trouble' rather than 'does not work', spells out that the actual failure mode is unmarked artificial boundaries (not garbage data), documents `Collector(set_truncated=True)` as a real-but-lossy mitigation (artificial truncations shorten trajectories and force value estimators to bootstrap where they shouldn't have had to), and adds a per-worker `ReplayBufferEnsemble` pattern with a code-block. Added a follow-up note that routing each worker's writes to its own member buffer needs an extra hook (post_collect_hook or per-worker collector) since the parent ensemble doesn't expose indexed extends out of the box.

_Response generated by Claude_

docs/source/reference/data_layout.rst
vmoens: Fixed in 3cefe03b6. You were right — I checked `_runner.py:463` and confirmed each worker process calls `replay_buffer.extend(...)` directly on the shared storage; the parent never aggregates. I rewrote the section to say so explicitly:

- A bullet list spells out that workers extend the buffer directly in **both** `sync=True` and `sync=False` mode, and that the `sync=` flag controls iteration delivery (whether the for-loop blocks for all workers vs first-come), not buffer population. So the inter-worker write order is uncontrolled in either mode — the same condition that makes `ndim>=2` shared storages unsafe.
- Reframed the role of `trajs_per_batch`: it doesn't change *who* extends or *when*; it changes *what* each extend commits. With `trajs_per_batch=N`, every `rb.extend` call is a complete trajectory, so the inter-worker race becomes harmless on a flat `ndim=1` buffer.

_Response generated by Claude_

docs/source/reference/data_layout.rst
vmoens: Fixed in 3cefe03b6 together with the line-128 thread. The "It does not work for" header is gone — the section now says "It needs the `set_truncated` or ensemble mitigation above for", which scopes the recommendation to those code paths rather than asserting flat impossibility, and the `set_truncated` mitigation itself is documented in the bullet list directly above.

_Response generated by Claude_

docs/source/reference/data_layout.rst
vmoens: Fixed in 3cefe03b6. Added a paragraph to the `Legacy: split_trajectories` section explaining that the function bakes collector hyperparameters into the data shape: a trajectory spanning two collected batches gets cut at the batch boundary, the part before gets an artificial `truncated` / `done`, the part after gets an artificial `is_init`, and changing `frames_per_batch` silently changes the trajectory shape the trainer consumes (effective lengths, return calculations, where value bootstraps land). The contiguous 1-D layout sidesteps this — boundaries come exclusively from the env's own `done`.

_Response generated by Claude_

PR #3712, opened 2026-05-06
[BugFix] Fix LSTMModule padding bug

### LINKED ISSUE #3711: [BUG] LSTMModule Bug
## Describe the bug

In LSTMModule, during the forward pass, the hidden states can be incorrectly removed. We believe this is because `inv_pad_sequence` expects it to be stored at the end of the trajectory step and removes the following steps accordingly, while in reality, the hidden state is stored at the max number of steps of the batch. This means trajectories with fewer steps than the max will see their hidden state incorrectly be zero. 

## To Reproduce

Steps to reproduce the behavior.

Please try to provide a minimal example to reproduce the bug. Error messages and stack traces are also helpful.

Please use the markdown code blocks for both code and stack traces.

```python
import torch
from torchrl.modules import LSTMModule
from torchrl.modules.tensordict_module.rnn import set_recurrent_mode

B = 1  # Batch Size (1 Env)
T = 5  # Time Steps
F = 2  # Observation Features
H = 4  # Hidden Size

is_init1 = torch.zeros(B, T, 1, dtype=torch.bool)
is_init1[0, 3] = True 

obs1 = torch.ones(B, T, F)
obs1[0, 3:] = 2.0

data = TensorDict({
    "obs": obs1,
    "is_init": is_init1
}, [B, T])

lstm = LSTMModule(
    input_size=F, 
    hidden_size=H, 
    in_key="obs", 
    out_keys=["action_features", ("next", "recurrent_state_h"), ("next", "recurrent_state_c")]
)

with set_recurrent_mode(True), torch.no_grad():
    data_recurrent = lstm(data)
print(data_recurrent["next", "recurrent_state_h"])
```

## Expected behavior

I'd expect `data_recurrent["next", "recurrent_state_h"]` to be populated at the steps of the end of each trajectory, but instead we see zeros at the end of trajectory 2.

## System info

 * Describe how the library was installed (pip, source, ...)
 * Python version
 * Versions of any other relevant libraries

```python
import torchrl, numpy, sys
```

## Reason and Possible fixes

We can contribute a PR to fix this, but it's because the hidden states are stored at the max length trajectory location, and then when we remove the padding, we remove the end of the padding for shorter trajectories, removing the stored hidden states. 

## Checklist

lin-erica:
## Description

Fixes [bug](https://github.com/pytorch/rl/issues/3711). 

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/test/test_tensordictmodules.py
+++ b/test/test_tensordictmodules.py
@@ -1054,6 +1054,54 @@ class TestLSTMModule:
 
         assert vmap(call, (None, 0))(data, params).shape == torch.Size((2, 50, 11))
 
+    @pytest.mark.parametrize("python_based", [False, True])
+    @pytest.mark.parametrize("num_layers", [1, 2])
+    def test_recurrent_state_at_traj_end(self, python_based, num_layers):
+        # Regression test for https://github.com/pytorch/rl/issues/3711:
+        # in recurrent mode, when a batch contains trajectories of different
+        # lengths, the recurrent_state stored at the end of each trajectory
+        # must be the LSTM hidden state after consuming that trajectory's
+        # last real observation -- not the hidden state after consuming the
+        # padded tail.
+        torch.manual_seed(0)
+        B, T, F, H = 1, 5, 2, 4
+        is_init = torch.zeros(B, T, 1, dtype=torch.bool)
+        is_init[0, 3] = True  # traj 1: steps 0..2, traj 2: steps 3..4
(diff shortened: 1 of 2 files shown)


vmoens:
Thanks for the PR!
I gave a stab at this based on your solution with Claude but the code it came up with looks pretty suboptimal (throughput-wise) to me. These nested loops freak me out a little.
I'll refine that tomorrow!

PR #3715, opened 2026-05-08
[BugFix] Improve shared replay buffer validation

vmoens:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #3722
* #3721
* #3720
* #3719
* #3718
* #3717
* #3716
* __->__ #3715

PR #3716, opened 2026-05-08
[BugFix] Preserve policy log-prob keys in async collectors

vmoens:
* #3722
* #3721
* #3720
* #3719
* #3718
* #3717
* __->__ #3716
* #3715

### DIFF
--- a/torchrl/collectors/utils.py
+++ b/torchrl/collectors/utils.py
@@ -22,6 +22,39 @@ _NON_NN_POLICY_WEIGHTS = (
 )
 
 
+def _log_prob_key_from_sample_key(key: NestedKey) -> NestedKey:
+    if isinstance(key, tuple):
+        return (*key[:-1], f"{key[-1]}_log_prob")
+    return f"{key}_log_prob"
+
+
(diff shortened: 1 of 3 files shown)


torchrl/collectors/utils.py
vmoens, line 1: If I understand correctly our issue here is that the keys defined by the policy are not registered in the env and are dropped.
Let's not overfit to log-prob but think more broadly about policy producing other non-action entries: how do we capture these?
Should we have policy specs (a bit hard to implement)? Should we validate the policy output the first time it's run and add the supplementary keys to the output then? Should we make a dry-run of the policy first and register the additional keys as tensordict primers or such in the env (that would expose them directly)?

vmoens: Broadened this beyond log-prob. The collector now preserves declared policy-produced extra keys, including non-action metadata, when initializing multi-process replay-buffer storage, and the existing derived-key restoration is framed as a complement to that generic dry-run path. Added a regression for a policy-produced `policy_info` key in MultiAsyncCollector replay-buffer writes; the probabilistic log-prob regression remains too.\n\n_Response generated by Codex_

PR #3717, opened 2026-05-08
[Performance] Reduce trajectory replay buffer write overhead

vmoens:
* #3722
* #3721
* #3720
* #3719
* #3718
* __->__ #3717
* #3716
* #3715

### DIFF
--- a/torchrl/collectors/_base.py
+++ b/torchrl/collectors/_base.py
@@ -288,6 +288,10 @@ class BaseCollector(IterableDataset, metaclass=abc.ABCMeta):
                 collector.start()  # workers fill rb with complete trajectories
 
             Defaults to ``None`` (fixed-frame batches).
+        trajs_per_write (int, optional): When ``trajs_per_batch`` is used with
+            a replay buffer, write this many completed trajectories to the
+            buffer per ``extend`` call. Larger values reduce Python overhead
+            for highly batched environments. Defaults to ``1``.
     """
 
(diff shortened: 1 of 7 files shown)


torchrl/collectors/_base.py
```diff
@@ -288,6 +288,10 @@ class BaseCollector(IterableDataset, metaclass=abc.ABCMeta):
                 collector.start()  # workers fill rb with complete trajectories
 
             Defaults to ``None`` (fixed-frame batches).
+        trajs_per_write (int, optional): When ``trajs_per_batch`` is used with
+            a replay buffer, write this many completed trajectories to the
+            buffer per ``extend`` call. Larger values reduce Python overhead
+            for highly batched environments. Defaults to ``1``.
```
vmoens: shouldn't we default to all trajs? What's the cost, beyond the call to cat()?
Can you make this kwarg more explicit with an example (eg, if we have 10 trajs queing to populate the buffer, a value of trajs_per_write=2 will make 5 writes to the buffer, a value of trajs_per_write=10 or more will do one).

vmoens: Changed the default to write all currently queued completed trajectories. The kwarg is now explicit in the docs: with 10 queued trajectories, `trajs_per_write=2` makes 5 writes, while `10` or larger makes 1 write. Added a regression for the default batching behavior.\n\n_Response generated by Codex_

PR #3719, opened 2026-05-08
[Quality] Make async evaluator skip busy requests by default

vmoens:
* #3722
* #3721
* #3720
* __->__ #3719
* #3718
* #3717
* #3716
* #3715

### DIFF
--- a/torchrl/collectors/_evaluator.py
+++ b/torchrl/collectors/_evaluator.py
@@ -130,13 +130,11 @@ class Evaluator:
       Requires *env* to be a callable and *policy_factory* to be provided.
 
     **Backpressure / overlap policy**: calling :meth:`trigger_eval` while a
-    previous evaluation is still running either raises immediately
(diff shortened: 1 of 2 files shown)


torchrl/collectors/_evaluator.py
vmoens, line 1: is there an option to let the evaluator run at all time?
ie pass a logger and a logging program, start() the evaluator and just let it run evaluations, log, update params, repeat.

vmoens: I left this PR scoped to backpressure semantics for externally triggered async evaluations. An always-running evaluator with its own logger/program loop is a larger API: it would need ownership of the evaluation cadence, logging callback contract, weight refresh policy, and shutdown behavior. The current path supports repeated external `trigger_eval` calls plus `on_result`; I think the self-running loop should be a follow-up design rather than part of this default-change PR.\n\n_Response generated by Codex_

PR #3720, opened 2026-05-08
[Feature] Add ordered replay buffer read write API

vmoens:
* #3722
* #3721
* __->__ #3720
* #3719
* #3718
* #3717
* #3716
* #3715

### DIFF
--- a/test/test_rb.py
+++ b/test/test_rb.py
@@ -146,6 +146,32 @@ TensorDictReplayBufferRNG = functools.partial(
 )
 
 
+def test_replay_buffer_read_write_all_in_order():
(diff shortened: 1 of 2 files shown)


test/test_rb.py
vmoens, line 1: i don't dislike it but we could do `rd[:]` for that no?

vmoens: Kept the explicit API, and documented where appropriate that `read_all_in_order()` mirrors `rb[:]` and `write_all(...)` mirrors slice assignment. Added tests comparing the helper methods with slice reads/writes, including `LazyTensorStorage(..., ndim=2)` coverage.\n\n_Response generated by Codex_

PR #3734, opened 2026-05-10
[Feature] Adds HERReplayBuffer and HindsightStrategy to torchrl.data.

### LINKED ISSUE #3713: [Feature Request] Implementing HER(Hindsight Experience Replay)
Motivation
Goal-conditioned RL is sparse-reward by nature. An agent manipulating objects, navigating to targets, or following language instructions almost always fails early in training — the desired goal is never reached, so the reward signal is nearly always zero and gradient flow stalls.

HER (Andrychowicz et al., 2017, https://arxiv.org/abs/1707.01495) solves this by relabeling failed trajectories at sample time: the achieved state is substituted as the goal, the reward is recomputed under that relabeled goal, and the transition becomes a valid positive training example. This turns failures into learning signal with no architectural changes to the policy or loss.

This is the standard baseline for any robotics or manipulation task. OpenAI Gym's GoalEnv was designed around it. Every major RL library (Stable-Baselines3, CleanRL, RLlib) has an implementation. TorchRL does not.

HER belongs in the replay buffer layer, not the loss or env. Relabeling happens at sample time, which is consistent with how PrioritizedSampler and SliceSampler extend sampling behavior without touching storage.

from torchrl.data import ReplayBuffer, LazyMemmapStorage
from torchrl.data.replay_buffers.samplers import HERSampler, HindsightStrategy

sampler = HERSampler(
    strategy=HindsightStrategy.FUTURE,  # "future" | "episode" | "random" | "final"
    her_ratio=0.8,           # fraction of sampled transitions to relabel
    goal_key="desired_goal",
    achieved_goal_key="achieved_goal",
    reward_fn=recompute_reward,  # Callable[[achieved, desired, info], Tensor]
)

rb = ReplayBuffer(storage=LazyMemmapStorage(100_000), sampler=sampler)
The buffer stores transitions with the original goal. HERSampler intercepts the sample call, selects relabeling candidates according to strategy, substitutes achieved_goal into desired_goal, calls reward_fn to recompute the scalar reward, and returns the modified tensordict. The policy and loss see nothing unusual.

theap06:
## Description

## Motivation and Context
Fixes #3713 

I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

##Summary 

- Adds `HERReplayBuffer` — a `TensorDictReplayBuffer` subclass that applies goal relabeling at sample time, turning failed goal-conditioned trajectories into useful training signal (Andrychowicz et al., NeurIPS 2017)
- Adds `HindsightStrategy` enum with all four canonical strategies: `FUTURE` (recommended), `FINAL`, `EPISODE`, `RANDOM`
- Exports both symbols from `torchrl.data` and `torchrl.data.replay_buffers`


theap06:
@vmoens Results utilizing HER. Wrote up a quick script for DDPG
image

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,359 @@
...
+from typing import Any, Literal
+
+import torch
+from tensordict.base import TensorDictBase
+from tensordict.utils import NestedKey
+
+from torchrl.data.replay_buffers.replay_buffers import TensorDictReplayBuffer
+
+
+class HindsightStrategy(str, Enum):
```
vmoens: StrEnum?

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,359 @@
...
+        RANDOM: sample a random achieved state from the entire buffer.
+    """
+
+    FUTURE = "future"
+    FINAL = "final"
+    EPISODE = "episode"
+    RANDOM = "random"
+
+
+class HERReplayBuffer(TensorDictReplayBuffer):
```
vmoens, line 49: A chance we can make this a RB, not only a TDRB? (Allow list storage etc)
Ok if not, happy with it if it increases the complexity

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,359 @@
...
+        ...             "observation": torch.randn(4),
+        ...             "desired_goal": torch.randn(3),
+        ...             "achieved_goal": torch.randn(3),
+        ...             "reward": torch.zeros(1),
+        ...             "done": torch.tensor([True]),
+        ...         },
+        ...     },
+        ...     batch_size=[],
+        ... )
+        >>> _ = rb.add(td)
```
vmoens: Maybe complete the example by showing mutation of some of the fields during sampling?

torchrl/data/replay_buffers/her.py
vmoens: Can we also make it more modular, through sampler / writer inheritance?

vmoens (commented):
Amazing, I love this.
I left a couple of high level comments can you have a look?

torchrl/data/replay_buffers/her.py
theap06: Was working off an old version of python. Updated it to utilize StrEnum

torchrl/data/replay_buffers/her.py
theap06: Updated it so that HER RB is now a standalone sampler anyone can drop into a plain TensorDictReplayBuffer without using HERReplayBuffer at all. Thanks for the feedback.

torchrl/data/replay_buffers/her.py
theap06: Done

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,508 @@
...
+from collections.abc import Callable
+from typing import Any, Literal
+
+try:
+    from enum import StrEnum
+except ImportError:  # Python < 3.11
+    from enum import Enum
+
+    class StrEnum(str, Enum):
+        """Backport of ``enum.StrEnum`` for Python < 3.11."""
```
vmoens: This breaks the import block.
Let's put it at the end of it.
In these cases I find it more pretty to use the python version from sys than a try/except but it's a matter of taste.

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,508 @@
...
+        RANDOM: sample a random achieved state from the entire buffer.
+    """
+
+    FUTURE = "future"
+    FINAL = "final"
+    EPISODE = "episode"
+    RANDOM = "random"
+
+
+class HERSampler(Sampler):
```
vmoens: Here we have the following alternative: break up the feature between files (samplers.py etc) or bundle everything in the same file. I guess it depends on composability.
For example we did not create a prioritized file to make it explicit that PRB and storage (among other things) are orthogonal features

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,508 @@
...
+        >>> sampler = HERSampler(her_ratio=0.5, strategy="future")
+        >>> rb = TensorDictReplayBuffer(
+        ...     sampler=sampler,
+        ...     storage=LazyMemmapStorage(1000),
+        ... )
+    """
+
+    def __init__(
+        self,
+        base_sampler: Sampler | None = None,
```
vmoens: This is an odd way of doing things from an OOP standpoint.
Alternative options could be a mixin, or a method/function that modifies an existing sampler once constructed. A sampler-in-sampler wrapping is going to have bleeding edges like poorly defined methods or samplers in samplers in samplers...

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,508 @@
...
+        done = all_data.get(self.end_key, None)
+        if isinstance(self.end_key, tuple):
+            terminated_key = self.end_key[:-1] + ("terminated",)
+        else:
+            terminated_key = "terminated"
+        terminated = all_data.get(terminated_key, None)
+
+        if done is None and terminated is None:
+            raise KeyError(
+                f"Neither {self.end_key!r} nor the corresponding 'terminated' key "
```
vmoens: Let's give a link to the doc page that explains done/terminated

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,508 @@
...
+            self._episode_ends_cache = self._build_episode_cache(storage)
+            self._last_cache_key = key
+        return self._episode_ends_cache
+
+    def _build_episode_cache(self, storage: Storage) -> torch.Tensor:
+        n = len(storage)
+        if n == 0:
+            return torch.zeros(0, dtype=torch.long)
+
+        all_data = storage.get(torch.arange(n))
```
vmoens: if n is 1M, we're creating a 1M big tensor on CPU here. Do we want that? Does this work with CUDA storages? What if the storage is n-dimensional (`ndim>1`)? What if the storage is complete and the last trajectory spans across the end of the storage limit?

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,508 @@
...
+        if done is not None and terminated is not None:
+            boundary = (done.bool() | terminated.bool()).view(n)
+        elif done is not None:
+            boundary = done.bool().view(n)
+        else:
+            boundary = terminated.bool().view(n)
+
+        episode_ends = boundary.nonzero(as_tuple=True)[0]
+        if len(episode_ends) == 0 or episode_ends[-1] != n - 1:
+            episode_ends = torch.cat([episode_ends, torch.tensor([n - 1])])
```
vmoens: device of this new tensor?

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,508 @@
...
+            )
+
+        if done is not None and terminated is not None:
+            boundary = (done.bool() | terminated.bool()).view(n)
+        elif done is not None:
+            boundary = done.bool().view(n)
+        else:
+            boundary = terminated.bool().view(n)
+
+        episode_ends = boundary.nonzero(as_tuple=True)[0]
```
vmoens: i'm ok with this but we should keep in mind that these kinds of things break compiled graphs. Sometimes there's no way around it

torchrl/data/replay_buffers/her.py
```diff
@@ -0,0 +1,508 @@
...
+        if strategy == HindsightStrategy.FINAL:
+            return ep_ends.clone()
+
+        if strategy == HindsightStrategy.EPISODE:
+            span = (ep_ends - ep_starts + 1).float()
+            offsets = (torch.rand(n) * span).long()
+            return ep_starts + offsets
+
+        if strategy == HindsightStrategy.RANDOM:
+            return torch.randint(storage_len, (n,))
```
vmoens: device?

vmoens (commented):
Good progress, left a few more comments here

torchrl/data/replay_buffers/her.py
theap06: moved the imports :)

torchrl/data/replay_buffers/her.py
theap06: Removed HERSampler entirely. HERReplayBuffer takes a regular Sampler and owns cache/ relabeling itself.

torchrl/data/replay_buffers/her.py
theap06: I ended up bundling everything - should be fine from a composability angle as well

torchrl/data/replay_buffers/her.py
theap06: Done

torchrl/data/replay_buffers/her.py
theap06: I changed it to be such that it only fetches the done/terminated field via storage[:].get(end_key). There is also a guard + cache key now.

vmoens (approved):
I appreciate you going back and forth with me on this. Good calls, LGTM!

vmoens:
This needs a conflict resolution

theap06:
@vmoens should be good to go! dealt with the merge conflict. currently just failing one test in CI

vmoens (approved):
LGTM

vmoens (approved):
LGTM

PR #3745, opened 2026-05-12
[Doc] Memory-efficient RL training tutorial + cross-refs

vmoens:
## Summary

Ties together the three recently-merged memory-efficiency PRs into a
single story:

- ``compact_obs`` collector flag (#3742)
- ``NextStateReconstructor`` RB transform (#3743)
- NaN-safe value-estimator forward (#3744)

Two parts:

**1. Runnable Sphinx-gallery tutorial** at
``tutorials/sphinx-tutorials/memory_efficient_rl.py``. Sections:

- Where the observation memory goes (concrete ``td.bytes()`` numbers)
- Why ``("next", obs)`` is kept by default — bootstrap target at
  trajectory ends, ``MultiStepTransform`` n-step fallback
- Knob 1 — ``SyncDataCollector(compact_obs=True)``
- Knob 2 — ``NextStateReconstructor`` with the traj_id + done contract
- Knob 2.5 — value-estimator NaN safety
  (``_sanitize_next_obs_nan``), GAE finite everywhere
- When *not* to take this path — ``MultiStepTransform`` incompatibility,
  the ``V(obs[t]) ≈ V(real_next_obs)`` approximation at truncated steps,
  and how ``shifted=True`` interacts
- Knob 3 — ``LazyMemmapStorage`` for buffers ≥ VRAM
- Knob 4 — ``SliceSampler`` + the new ``"scan"`` / ``"triton"``
  recurrent backends for padding-free sequence training
- End-to-end pipeline snippet
- Conclusion + Further reading

Runs end-to-end on CPU (CartPole-v1, 200 frames; <2s wall) and reports
the byte-level savings concretely from ``td.bytes()``.

**2. Docstring cross-references** so a reader landing on any of the
three new APIs finds the other two:

- ``Collector(compact_obs=…)`` (and the multi-process collectors):
  pointers to ``NextStateReconstructor``, the value-estimator
  sanitizer, the ``MultiStepTransform`` incompatibility note, and the
  new tutorial.
- ``NextStateReconstructor``: ``.. seealso::`` block covering
  ``compact_obs``, the sanitizer, ``MultiStepTransform``, and the
  tutorial.
- ``ValueEstimatorBase._sanitize_next_obs_nan``: ``.. seealso::``
  block to ``compact_obs``, ``NextStateReconstructor``, and the
  tutorial.

``docs/source/index.rst`` registers the new tutorial under "Basics".

## Test plan

- [x] Tutorial runs end-to-end with concrete output (memory savings
  reported, NaN at slice boundaries confirmed to coincide with
  trajectory boundaries, GAE advantage finite everywhere, memmap
  roundtrip works).
- [x] All cross-references resolve to existing public symbols
  (verified by reading the rendered class docstrings via
  ``Collector.__doc__`` etc.).


tutorials/sphinx-tutorials/memory_efficient_rl.py
```diff
@@ -0,0 +1,694 @@
+"""
+Memory-Efficient RL Training
+============================
```
vmoens, line 3: A tutorials should always start with a paragraph explaining the problem. Here the issue mainly that with the emergence of highly vectorized envs (mujoco XLA, mujoco warp, mujoco torch, isaaclab, genesis, ...) we are collecting data at impressive rates. Models can also scale to hundreds of millions to billions of parameters. If one is not careful, the memory cost of aquiring data can quickly pile up. There is a tension between fast data collection (ie minimizing data movement, on-device simulation and policy execution) and efficient training at scale. 
The pipelines are difficult to handle as well: in a typical PPO training loop, the model needs to be executed as a policy, the critic during GAE and the whole thing during the loss (incl. backward). Keeping the data on the right device at the right time, using accelerators when needed, fast data transmission between inference and trainer are challenges that need to be addressed carefully.

tutorials/sphinx-tutorials/memory_efficient_rl.py
```diff
@@ -0,0 +1,694 @@
...
+
+**Author**: `Vincent Moens <https://github.com/vmoens>`_
+
+.. _memory_efficient_rl:
+
+.. grid:: 2
+
+    .. grid-item-card:: :octicon:`mortar-board;1em;` What you will learn
```
vmoens, line 11: we should add that keeping data on device during rollouts in the collector isn't the best thing if memory is a problem => explain how to use env device, storing device, policy device... 

For RNNs, give links to tutorials and doc about using scan and triton which allow to remove padding coming from intermediate resets (in contrast with cudnn which does not handle intermediate resets). Explain that GAE does not use padding and is efficient in doing so. `split_trajectories` also has a nested tensor mode that allows to avoid padding.

tutorials/sphinx-tutorials/memory_efficient_rl.py
vmoens: Done in 7edd19338. Added an opening "The problem" section before "Where the memory goes": highly vectorized simulators (MuJoCo XLA/Warp/torch, Isaac Lab, Genesis) collect at staggering rates while policies/critics scale to 100M-1B params, so the memory cost of acquiring data piles up; it spells out the fast-collection (minimize data movement, on-device sim + policy) vs. efficient-training tension and the PPO pipeline's policy/critic-GAE/loss+backward device juggling.

_Response generated by Claude_

tutorials/sphinx-tutorials/memory_efficient_rl.py
vmoens: Done in 7edd19338, split into two parts.

Device placement: added **Knob 3.5 - keep collection off the training memory budget**, explaining that keeping the rollout on the accelerator is fastest but competes with the model for VRAM, and documenting `env_device` / `policy_device` / `storing_device` (and the `device` shorthand). Rule of thumb: simulate + act on the accelerator, but move `storing_device` down the hierarchy (accelerator -> host RAM -> memmap on disk) as the buffer grows. Links the collector-internals page for the cast/sync points, and mentions `no_cuda_sync`.

RNNs: expanded **Knob 4** to explain that cuDNN LSTM/GRU cannot reset the hidden state mid-sequence, so a batch with intermediate resets must be split-and-padded, whereas `scan`/`triton` reset in place from `is_init` (no padding, and lower backward activation memory). Added that GAE/TD(λ) consume the flat contiguous slice layout directly and never materialise a padded `(batch, max_T)` view, and documented `split_trajectories(..., as_nested=True)` as the nested-tensor (no-padding) alternative when a per-trajectory `(B, T)` view is needed. Links the recurrent-state-lifecycle guide and the recurrent DQN tutorial.

_Response generated by Claude_

tutorials/sphinx-tutorials/memory_efficient_rl.py
```diff
@@ -0,0 +1,805 @@
...
+torch.manual_seed(0)
+
+######################################################################
+# The problem
+# -----------
+#
+# With the emergence of highly vectorized simulators (MuJoCo XLA / Warp /
+# the native PyTorch backend, Isaac Lab, Genesis, ...) we now collect data
+# at staggering rates, and policies and critics routinely scale from
+# hundreds of millions to billions of parameters. If one is not careful,
```
vmoens: these are two separate things: envs have scaled up AND policies have scaled up (VLA, ...)

tutorials/sphinx-tutorials/memory_efficient_rl.py
```diff
@@ -0,0 +1,805 @@
...
+#   reconstruct correctly within the round-trip precision of
+#   ``delta_dtype``. No ``NaN``, no need for the value-estimator
+#   sanitizer at Knob 2.5.
+# * **Loss compatibility.** The reconstructed ``("next", obs)`` is
+#   bit-close to the original (subject to ``delta_dtype`` precision).
+#   Truncated-step bootstraps see the real next obs, not the
+#   ``V(obs[t]) ≈ V(real_next_obs)`` approximation.
+# * **MultiStep.** Still incompatible — the n-step transform needs
+#   ``data[t + n]`` materialised at every position; a one-step delta
+#   doesn't carry that information.
```
vmoens: that is not true, you can rebuild the multi-step with delta

tutorials/sphinx-tutorials/memory_efficient_rl.py
```diff
@@ -0,0 +1,805 @@
...
+    mmap_sample = rb_mmap.sample()
+    print("memmap sample shape:", mmap_sample.shape)
+
+######################################################################
+# The data went through disk, but the public API is identical to the
+# in-memory case. See the :ref:`replay buffer tutorial <rb_tuto>` for
+# more on storage choices.
+
+######################################################################
+# Knob 3.5 — keep collection off the training memory budget
```
vmoens: why 3.5? The numbers are confusing

tutorials/sphinx-tutorials/memory_efficient_rl.py
```diff
@@ -0,0 +1,805 @@
...
+# * :class:`~torchrl.data.SliceSampler` returns *contiguous* slices of
+#   pre-specified length. There is no padding; every entry is a real
+#   transition. The trajectory-id key lets the sampler align slices to
+#   trajectory boundaries.
+# * :class:`~torchrl.modules.LSTMModule` and
+#   :class:`~torchrl.modules.GRUModule` accept a
+#   ``recurrent_backend`` argument with three non-default values:
+#
+#     * ``"scan"`` — built on
+#       ``torch._higher_order_ops.scan`` (PyTorch ≥ 2.6). Resets the
```
vmoens: hoptorch only works with >= 2.7

tutorials/sphinx-tutorials/memory_efficient_rl.py
vmoens: Fixed in cb19724ee. Split the intro into two distinct trends: **environments scaled up** (vectorized simulators -> data volume through rollouts/buffers) and **policies scaled up** (VLA policies, large recurrent critics, transformer world models -> parameter/activation footprint), noting either strains memory and together they compound.

_Response generated by Claude_

tutorials/sphinx-tutorials/memory_efficient_rl.py
vmoens: Fixed in cb19724ee. Corrected the MultiStep bullet: the delta path keeps the full per-step transition (full-precision root `obs` retained at every step, `("next", obs)` reconstructs exactly everywhere, rewards/dones never dropped), so `data[t+n]` is recoverable and n-step returns can be rebuilt (reconstruct `("next", obs)` from the delta, then apply `MultiStepTransform`). Also re-scoped the MultiStep-incompatibility statements in 'When not to rehydrate', the Conclusion, and the learning objectives to the *compact* path only, calling out the delta path as the exception.

_Response generated by Claude_

tutorials/sphinx-tutorials/memory_efficient_rl.py
vmoens: Fixed in cb19724ee. Renumbered the knobs to clean sequential integers 1..9 (was 1, 2, 2b, 2.5, 2.6, 2.7, 3, 3.5, 4) and updated every in-text cross-reference accordingly.

_Response generated by Claude_

tutorials/sphinx-tutorials/memory_efficient_rl.py
vmoens: Fixed in cb19724ee. Corrected the scan backend description: it is built on the `hoptorch` scan primitive (`pip install hoptorch>=0.1.4`) and requires **PyTorch >= 2.7**, not `torch._higher_order_ops.scan` / PyTorch >= 2.6.

_Response generated by Claude_

PR #3748, opened 2026-05-13
[Feature] MAPPOLoss + IPPOLoss + MultiAgentGAE + ValueNorm

theap06:
## Context

Multi-agent RL is currently the weakest research surface in torchrl: the only multi-agent loss shipped is `QMixerLoss` (DQN family, discrete actions). For cooperative continuous-control MARL — where most modern benchmarks live (SMAC, VMAS, PettingZoo MPE, Hanabi, Overcooked) — users have to hand-assemble `ClipPPOLoss` + manual `set_keys(done=("agents", "done"), terminated=("agents", "terminated"))` + manual `make_value_estimator(GAE, ...)`. The existing `sota-implementations/multiagent/mappo_ippo.py` recipe shows what this boilerplate looks like.

This PR adds **MAPPO** (Yu et al. 2022) and **IPPO** (de Witt et al. 2020) as first-class objectives, plus the two pieces of supporting infrastructure they need.

## What's new

- **`torchrl.objectives.multiagent.MAPPOLoss`** — centralised-critic, decentralised-actor PPO. Subclasses `ClipPPOLoss`; defaults the value estimator to `MultiAgentGAE`, defaults `normalize_advantage_exclude_dims=(-2,)`, and optionally accepts a `ValueNorm` for the critic-stability trick from the paper.
- **`torchrl.objectives.multiagent.IPPOLoss`** — independent-learner counterpart. Each agent has its own local critic; no centralised state required.
- **`torchrl.objectives.value.MultiAgentGAE`** — `GAE` variant that broadcasts team-shared `reward` / `done` / `terminated` (shape `[*B, T, 1]`) across the agent dim before the vec-GAE recursion, so users don't have to manually replicate signals or override `set_keys`. New `ValueEstimators.MAGAE` enum entry.
- **`torchrl.modules.ValueNorm`** — PopArt-style running value normaliser (van Hasselt et al. 2019), used opt-in by `MAPPOLoss`. Yu et al. 2022 Table 13 credits this trick with the algorithm's strong SMAC results.

## Design notes

**Two classes instead of a `centralized: bool` flag.** The structural code difference between MAPPO and IPPO is small (~20 lines), but I made them separate named classes rather than a single class with a flag because:
- The recent feedback on (HER) was explicit about avoiding wrapper-in-wrapper / "sampler-in-sampler" APIs. A boolean flag on a single class is the same pattern shifted to losses.
- `from torchrl.objectives.multiagent import MAPPOLoss` is self-documenting; the docstring spells out the full recipe (centralised critic construction, etc.) for each algorithm independently.

**MAGAE dispatch in plain PPO / A2C / Reinforce.** Adding `ValueEstimators.MAGAE` to the enum would break every parent test that parametrises over `list(ValueEstimators)` unless every `make_value_estimator` knows the new enum value. Two options: (a) update ~29 test parametrisations to skip MAGAE, or (b) have plain PPO / A2C / Reinforce dispatch MAGAE to `MultiAgentGAE`. I went with (b) — it's ~5 lines per file, leaves the enum exhaustive, and is the right thing semantically (any actor-critic with the right data shapes can use MAGAE).

**ValueNorm placement.** Lives under `torchrl/modules/` rather than `torchrl/objectives/utils/` because it's a stateful learnable component that participates in `.to(device)` / `state_dict()`. Happy to move if reviewers prefer otherwise.

## Out of scope (follow-up)

- HAPPO / sequential update scheme (Kuba et al. 2022)
- Multi-Agent Transformer (MAT)
- Refactoring `sota-implementations/multiagent/mappo_ippo.py` to use the new classes — left untouched in this PR to keep the blast radius small; can be a one-line follow-up.

## Verification

- `pytest test/objectives/test_mappo.py` — 16/16 passing. Synthetic-tensordict tests for forward shapes, backward, centralised-vs-decentralised critic semantics, share-params modes, ValueNorm convergence, and critic-loss bounded-ness under 10× reward inflation.
- `pytest test/test_cost.py -k "ppo or qmixer or a2c or reinforce"` — 2394/2394 passing (no regressions).
- Full `test_cost.py` — 8788 passing, 1 pre-existing unrelated failure (`test_exploration_compile` — `torch.compile` + `torch.utils.mkldnn` deprecation, no MAPPO involvement).
- `examples/multiagent/mappo_vmas.py --algo mappo --frames 200_000` provides a minimal end-to-end smoke recipe on VMAS Navigation.

cc @matteobettini

### DIFF
--- a/docs/source/reference/objectives.rst
+++ b/docs/source/reference/objectives.rst
@@ -50,4 +50,5 @@ Documentation Sections
    objectives_policy
    objectives_actorcritic
    objectives_offline
+   objectives_multiagent
    objectives_other
--- a/docs/source/reference/objectives_common.rst
+++ b/docs/source/reference/objectives_common.rst
@@ -28,6 +28,7 @@ Value Estimators
     TD1Estimator
     TDLambdaEstimator
     GAE
+    MultiAgentGAE
 
 .. currentmodule:: torchrl.objectives
 
--- a/docs/source/reference/objectives_multiagent.rst
+++ b/docs/source/reference/objectives_multiagent.rst
@@ -1,0 +1,58 @@
+.. currentmodule:: torchrl.objectives.multiagent
+
+Multi-Agent Objectives
+======================
+
+Loss modules for multi-agent reinforcement learning algorithms. These losses
+follow the torchrl multi-agent tensordict convention (per-agent tensors
+nested under group keys such as ``("agents", "observation")``; see
+:class:`~torchrl.envs.libs.vmas.VmasEnv` and
+:class:`~torchrl.envs.libs.pettingzoo.PettingZooEnv`).
+
+MAPPO and IPPO
+--------------
+
+:class:`MAPPOLoss` implements Multi-Agent PPO (Yu et al. 2022) — a
+decentralised actor paired with a *centralised critic* that conditions on the
+joint observation / state. :class:`IPPOLoss` is the independent-learner
+counterpart from de Witt et al. 2020: each agent has its own local critic and
+there is no centralised information at training time.
+
+Both are thin specialisations of :class:`~torchrl.objectives.ClipPPOLoss`
+that:
+
+- default the value estimator to
+  :class:`~torchrl.objectives.value.MultiAgentGAE`, which broadcasts
+  team-shared rewards / done flags across the agent dimension before
+  computing returns;
+- default ``normalize_advantage_exclude_dims`` to ``(-2,)`` so the agent dim
+  is excluded from advantage standardisation;
+- optionally accept a :class:`~torchrl.modules.ValueNorm` subclass — either
+  :class:`~torchrl.modules.PopArtValueNorm` (EMA, recommended for drifting
+  reward scales) or :class:`~torchrl.modules.RunningValueNorm` (exact
+  Welford running stats, recommended for stationary scales) — to stabilise
+  the critic loss. The MAPPO paper credits this trick for its strong SMAC
+  results.
+
+See ``sota-implementations/multiagent/mappo_ippo.py`` for a hydra-configured
+recipe and ``examples/multiagent/mappo_vmas.py`` for a minimal one.
+
+.. autosummary::
+    :toctree: generated/
+    :template: rl_template_noinherit.rst
+
+    MAPPOLoss
+    IPPOLoss
+
+QMixer
+------
+
+:class:`QMixerLoss` mixes local per-agent Q values into a global team Q
+value via a learnable mixing network, and trains them jointly with a DQN
+update on the global value (Rashid et al. 2018).
+
+.. autosummary::
+    :toctree: generated/
+    :template: rl_template_noinherit.rst
+
+    QMixerLoss
--- a/examples/multiagent/mappo_vmas.py
+++ b/examples/multiagent/mappo_vmas.py
@@ -1,0 +1,228 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""Minimal MAPPO / IPPO recipe on VMAS using the new
+:class:`~torchrl.objectives.multiagent.MAPPOLoss` /
+:class:`~torchrl.objectives.multiagent.IPPOLoss` classes.
+
+For the full, hydra-configured, wandb-logged version see
+``sota-implementations/multiagent/mappo_ippo.py``. This file is intentionally
+short: it's there to show that the new loss classes collapse the boilerplate
+that previously required ``ClipPPOLoss`` + manual ``make_value_estimator(GAE,
+...)`` into a single construction call. We still need ``set_keys(done=...,
+terminated=...)`` because VMAS writes its done/terminated flags under the
+per-agent group key.
+
+Usage::
+
+    python examples/multiagent/mappo_vmas.py --algo mappo --frames 200_000
+    python examples/multiagent/mappo_vmas.py --algo ippo --frames 200_000
+
+The two should reach similar reward on the easy navigation scenario; MAPPO
+typically pulls ahead on harder coordination tasks (Yu et al. 2022).
+"""
+from __future__ import annotations
+
+import argparse
+import time
+
+import torch
+from tensordict.nn import TensorDictModule
+from tensordict.nn.distributions import NormalParamExtractor
+from torch import nn
+
+from torchrl.collectors import SyncDataCollector
+from torchrl.data import LazyTensorStorage, TensorDictReplayBuffer
+from torchrl.data.replay_buffers.samplers import SamplerWithoutReplacement
+from torchrl.envs import RewardSum, TransformedEnv
+from torchrl.modules import (
+    MultiAgentMLP,
+    PopArtValueNorm,
+    ProbabilisticActor,
+    TanhNormal,
+)
+from torchrl.objectives import IPPOLoss, MAPPOLoss
+
+
+def make_actor(env, *, share_params: bool = True) -> ProbabilisticActor:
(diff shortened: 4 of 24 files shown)


Xmaster6y:
I think you're right about making MA training more straightforward, but I have some concerns:
- `MultiAgentGAE.forward` seems to duplicate most of `GAE.forward`; maybe an additional level of abstraction is needed.
- `ValueNorm` should be more generic and less tied to MAPPO
- We might need a registry for value estimators instead of enums
- We should maybe handle/consider potential compatibility issues with MAGAE for other algs

theap06:
> I think you're right about making MA training more straightforward, but I have some concerns:
> 
> * `MultiAgentGAE.forward` seems to duplicate most of `GAE.forward`; maybe an additional level of abstraction is needed.
> * `ValueNorm` should be more generic and less tied to MAPPO
> * We might need a registry for value estimators instead of enums
> * We should maybe handle/consider potential compatibility issues with MAGAE for other algs

I think the layers of abstractions make sense. I think the value estimators would help because of the utilization of the collectors as well. For the compatibility issues, I can write up some test cases to ensure it doesn't impact existing algos.

itwasabhi:
I'd like to file a request for the MAPPO loss if you are open to it. Centralization should happen over agents across agent groups. The current mappo loss (and all torchrl loss modules applied to MARL)  only centralizes over agents in the same group. 

Theres a few settings where you may want to centralize across groups: 
- heterogenous agents (differenting observation/action specs) make sense to keep agents seperated
- ad-hoc teamwork, where agents in one group are (for example) following a pre-defined policy.

theap06:
> I'd like to file a request for the MAPPO loss if you are open to it. Centralization should happen over agents across agent groups. The current mappo loss (and all torchrl loss modules applied to MARL) only centralizes over agents in the same group.
> 
> Theres a few settings where you may want to centralize across groups:
> 
> * heterogenous agents (differenting observation/action specs) make sense to keep agents seperated
> * ad-hoc teamwork, where agents in one group are (for example) following a pre-defined policy.

Thanks for raising this. I agree cross-group centralization is the right generalization for heterogeneous teams and ad-hoc teamwork. I’d prefer to keep this PR scoped to the current single-group MAPPO path, since the limitation comes from the existing MultiAgentMLP centralization behavior and this PR is already fairly broad.

I’ll open a follow-up issue and take a stab at it separately. Feel free to provide feedback btw! My initial preference is the composable CrossGroupCritic TensorDictModule route, since it keeps the current API untouched and avoids a wider MultiAgentMLP refactor.

theap06:
@vmoens results for the training 
image
image

LGTM!

theap06:
@vmoens I think the CI is failing due to the commit: Add CUDA support for prioritized replay sampling.  This landed on main before this branch was opened. None of the commits in this PR touch any C extension or build system files;the failure reproduces on main independently. Happy to confirm if you want me to rebase and rerun, but I don't think it's blocking for review.

vmoens:
@theap06 I took the liberty of pushing the review fixes directly to your branch (the PR was marked as `maintainer_can_modify: true`). Force-pushed with `--force-with-lease`. The branch now has 3 commits:

1. **`[Compat]`** — generator-kwarg guard on `SafeProbabilisticModule` (cleaned up: now uses a precomputed `extra_kwargs` dict instead of the conditional unpack trick) + walks the MRO in `LossModule.convert_to_functional`'s annotation check, so subclasses don't have to redeclare their parents' annotations. That second fix is what was producing the 4 `UserWarning`s per `MAPPOLoss(...)` instantiation.

2. **`[Refactor]`** — exposes a single `_critic_loss_inputs(target_return, state_value, old_state_value)` hook on `PPOLoss.loss_critic` (identity by default) so that `MAPPOLoss` can inject PopArt-style normalisation **without** duplicating the parent body. This means `value_norm` now composes correctly with `clip_value`, `separate_losses=True`, and `log_explained_variance` — none of which worked in the original loss_critic override. Same idea drives the `_prepare_signals` / `_broadcast_optional` / `_normalize_advantage` hooks on `GAE` that the original PR already had, just cleaned up.

3. **`[Feature]`** — the actual MAPPO/IPPO/MultiAgentGAE/ValueNorm work. Changes from the original:
   - `MAPPOLoss.__init__` no longer calls `add_module("_value_norm_module", ...)` — `nn.Module.__setattr__` already registers it. `state_dict()` now has exactly three `value_norm.*` keys instead of six (`value_norm.*` + `_value_norm_module.*`).
   - `IPPOLoss` is now a 1-line subclass of `MAPPOLoss`; the two `__init__`s were copy-pasted before.
   - `RunningValueNorm.update` uses `math.prod` instead of `torch.tensor([...]).prod()` (no host-device sync). The misleading variance comment is also fixed.
   - `examples/multiagent/mappo_vmas.py` — dropped the dead `if False else` branch and fixed the `set_keys(done=("agents", "done"), terminated=("agents", "terminated"))` call so the example actually works on VMAS (which writes per-agent done flags).

The registry refactor that was bundled into the original PR is now its own PR — **#3780** — extended to convert **every** loss in `torchrl.objectives` (not just PPO/A2C/Reinforce), so `@register_value_estimator` actually delivers on its promise.

New test coverage (5 regression tests) all targeting bugs from the review:
- `test_value_norm_not_double_registered` — exactly 3 keys in `state_dict()`.
- `test_value_norm_state_dict_round_trip` — save/load round-trips the running stats.
- `test_value_norm_composes_with_clip_value` — `value_clip_fraction` shows up in the output dict even with `value_norm` attached.
- `test_value_norm_separate_losses_detaches_actor_grads` — critic loss does not backprop into actor params.
- `test_no_spurious_annotation_warnings_on_instantiation` — verifies the `LossModule` annotation fix above.

Full test suite: `pytest test/objectives/` passes 7280 tests, 0 regressions.

Diff is down from 1794+/171- to 1637+/38- across 24 files.

vmoens (approved):
LGTM thanks

matteobettini:
This looks cool but am i missing something or the 2 losses in https://github.com/pytorch/rl/blob/main/torchrl/objectives/multiagent/mappo.py look the same?

theap06:
@matteobettini Nope, you're not missing anything! They're intentionally the same! The only real difference is what critic you pass in. MAPPOLoss expects a centralised critic, IPPOLoss expects a decentralised one, but the loss math is identical either way.Exposed them as separate classes mostly so the API makes it obvious which algorithm you're running.  Happy to collapse it to IPPOLoss = MAPPOLoss or add some actual structural difference in a follow-up if you think that's cleaner, just lmk!

matteobettini:
> @matteobettini Nope, you're not missing anything! They're intentionally the same! The only real difference is what critic you pass in. MAPPOLoss expects a centralised critic, IPPOLoss expects a decentralised one, but the loss math is identical either way.Exposed them as separate classes mostly so the API makes it obvious which algorithm you're running. Happy to collapse it to IPPOLoss = MAPPOLoss or add some actual structural difference in a follow-up if you think that's cleaner, just lmk!

Mmh i am not sure I am a big fan of 2 names for the same thing. The risk here is that people might think changing the loss name changes the loss while it does not actually change anything and the change has to be done somewhere else.

theap06:
@matteobettini I can draft up a follow up pr addressing this if you would like later tonight.

matteobettini:
> @matteobettini I can draft up a follow up pr addressing this if you would like later tonight.

What do you think? we could (1) drop IPPOLoss keeping only MAPPOLoss or (2) keep only one called `MultiPPOLoss`

theap06:
> > @matteobettini I can draft up a follow up pr addressing this if you would like later tonight.
> 
> What do you think? we could (1) drop IPPOLoss keeping only MAPPOLoss or (2) keep only one called `MultiPPOLoss`

I was thinking we just create a field called MultiPPOLoss and the user can choose.

PR #3754, opened 2026-05-15
[Performance] LSTM/GRU scan: canonical strides + cuDNN flat-storage clones + thread-local recurrent mode

vmoens:
* #3776
* #3775
* #3769
* #3763
* #3771
* #3768
* #3767
* #3764
* #3759
* #3757
* #3762
* #3755
* __->__ #3754
* #3753

Three intertwined fixes to the scan / triton recurrent backends.

- Canonical-stride check. A `[1, 4, 5]` tensor with strides `(5, 5, 1)`
  passes `is_contiguous()` but `torch._higher_order_ops.scan` and the
  triton kernels read strides directly and reject non-canonical layouts.
  Add `_canonical_stride` + `_canonical_contiguous` and re-materialize
  inputs / hidden buffers when strides drift off the C-canonical layout.

- cuDNN flat-storage aliasing. `nn.LSTM` / `nn.GRU` with cuDNN flatten
  all per-layer parameters into a single storage; the scan HOP tracer
  walks the FakeTensor graph and rejects the aliased per-layer views as
  inputs. Clone the weight views before closing the scan body. The
  per-layer carry now also clones `x_t` and the transpose+flatten output
  (the only remaining aliasing edge) so the existing `.clone()` on the
  full `torch.stack(...)` carry can drop.

- Thread-local `recurrent_mode`. `_ContextManager` was a single mutable
  module-level flag, so spawning a collector worker thread saw the
  parent's recurrent_mode setting. Wrap in `_RecurrentModeContextManager`
  using `contextvars.ContextVar` so per-thread state is isolated.

Tests cover: `_canonical_contiguous` no-op vs re-materialize, scan/triton
output parity under non-canonical strides, scan-under-torch.compile
aliasing pin for both LSTM and GRU, LSTM scan/pad PPO-advantage parity,
non-canonical hidden-buffer regression for scan, and a thread-local
`set_recurrent_mode` test.

### DIFF
--- a/torchrl/modules/tensordict_module/rnn.py
+++ b/torchrl/modules/tensordict_module/rnn.py
@@ -2461,10 +2461,17 @@ class _RecurrentModeContextManager(_ContextManager):
         )
 
     def get_mode(self) -> bool | None:
+        # Dynamo can't trace ContextVar.get; fall back to the parent's plain
+        # attribute under torch.compile. set_mode keeps both in sync so this
+        # stays correct (compile traces a single thread).
+        if is_compiling():
+            return self._mode
         return self._context_mode.get()
 
     def set_mode(self, mode: bool | None) -> None:
-        self._context_mode.set(mode)
+        self._mode = mode
+        if not is_compiling():
+            self._context_mode.set(mode)
 
 
 recurrent_mode_state_manager = _RecurrentModeContextManager()
--- a/torchrl/modules/tensordict_module/rnn.py
+++ b/torchrl/modules/tensordict_module/rnn.py
@@ -1119,6 +1119,13 @@ class LSTMModule(ModuleBase):
 
         layer_input = _canonical_contiguous(input)
         is_init = _canonical_contiguous(is_init)
+        # Canonicalize the full hidden buffers once. For num_layers=1
+        # (the common RL case) per-layer slicing then yields a canonical
+        # view for free; for num_layers>=2 adjacent layers are interleaved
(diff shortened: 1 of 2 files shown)


torchrl/modules/tensordict_module/rnn.py
```diff
@@ -1110,8 +1134,8 @@ def _lstm_triton_with_resets(
                 b_ih = zeros if b_ih is None else b_ih
                 b_hh = zeros if b_hh is None else b_hh
 
-            hidden_per_step = hidden0_in[..., layer, :]
-            cell_per_step = hidden1_in[..., layer, :]
+            hidden_per_step = _canonical_contiguous(hidden0_in[..., layer, :])
+            cell_per_step = _canonical_contiguous(hidden1_in[..., layer, :])
```
vmoens, line 1145: what if we were doing that for hidden0_in before the loop? Would the view still be contiguous?

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -2245,7 +2287,7 @@ def _gru_triton_with_resets(
                 b_ih = zeros if b_ih is None else b_ih
                 b_hh = zeros if b_hh is None else b_hh
 
-            hidden_per_step = hidden_in[..., layer, :]
+            hidden_per_step = _canonical_contiguous(hidden_in[..., layer, :])
```
vmoens, line 2301: ditto, should we do it before the loop?

PR #3756, opened 2026-05-15
[Feature] timeit.mark_start/mark_end for non-context-manager timing

vmoens:
* #3761
* #3760
* #3759
* #3758
* #3757
* __->__ #3756
* #3755
* #3754
* #3753

The collector worker loop needs to time spans that begin in one branch
of a control-flow path and end in another — picking up the next
``continue`` command, then ending after the env step. ``with timeit(...)``
can't bridge that.

Refactor ``__exit__`` through a new ``_record(name, elapsed)`` classmethod
so the recording side is callable independently. Add ``timeit.mark_start``
/ ``timeit.mark_end`` (and ``mark_env`` alias) on top: they push and pop
into a class-level ``_MARKS`` dict and feed ``_record`` on end. The
context-manager path is unchanged.

Unit tests cover: round-trip into ``_REG``, mark cleanup on end, alias
identity for ``mark_env``, and that ``with timeit(...)`` still records.

### DIFF
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -243,6 +243,7 @@ class timeit:
     """
 
     _REG = {}
+    _MARKS = {}
 
     def __init__(self, name):
         self.name = name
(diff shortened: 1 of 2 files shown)


torchrl/_utils.py
```diff
@@ -295,16 +296,34 @@ def elapsed(self) -> float:
...
+
+    @classmethod
+    def mark_end(cls, name: str) -> None:
+        """Mark the end of a named timed event and record its elapsed time."""
+        cls._record(name, time.time() - cls._MARKS.pop(name))
+
+    @classmethod
+    def mark_env(cls, name: str) -> None:
+        """Alias for :meth:`mark_end`."""
+        cls.mark_end(name)
```
vmoens, line 325: remove this

PR #3758, opened 2026-05-15
[Feature] Collector final_obs: store true boundary next-obs for shifted-GAE

vmoens:
* #3769
* #3763
* #3771
* #3770
* #3768
* #3767
* #3764
* #3759
* __->__ #3758
* #3757
* #3762
* #3755
* #3754
* #3753

When ``compact_obs=True`` is paired with shifted-GAE, the last step of
each rollout window has no ``("next", obs)`` to read; GAE was falling
back to ``V(s_{T-1})`` via ``_fill_missing_next_inputs`` and biasing
the advantage by 1/T at every window boundary that is not a real done.

Add a ``final_obs`` flag (requires ``compact_obs=True``) on the single
and multi-process collectors. When on, the collector maintains a
per-env side buffer that mirrors the most recent ``("next", k)`` in
place; the rollout returned to the consumer carries each leaf under
``("final", k)`` wrapped in :class:`tensordict.UnbatchedTensor` so the
``[*envs, T]`` batch shape is preserved (the buffer has no time dim).

On the consumer side, ``ValueEstimatorBase`` gets two new static
methods:

- ``_apply_final_obs_to_next_done`` finds the synthetic last-step
  positions inside the interleaved bootstrap batch and substitutes
  ``("final", k)`` values for the bootstrap input. No-op when
  ``("final", ...)`` is absent — non-final-obs callers are unaffected.

- ``_maybe_drop_final_obs`` deletes ``("final", ...)`` from the
  consumed tensordict. ``UnbatchedTensor`` leaves are incompatible
  with contiguous-storage replay buffers (``LazyTensorStorage``,
  ``LazyMemmapStorage``); dropping after consumption keeps the
  collector -> GAE -> ``ReplayBuffer.extend()`` pipeline clean.

All five estimators (TD0 / TD1 / TDLambda / GAE / VTrace) invoke
``_maybe_drop_final_obs`` at the end of ``forward``.

Also documents the memmap-vs-on-device tradeoff on
``LazyMemmapStorage`` — the example below the stack picks memmap to
keep large rollout buffers off-device.

Tests cover: ``final_obs=True`` without ``compact_obs`` raises;
boundary obs match a ``compact_obs=False`` reference (non-done envs);
``UnbatchedTensor`` survives indexing along time; bootstrap parity
between compact+final and a non-compact reference for TD0 / TD1 /
TDLambda / GAE; and ``("final", ...)`` is dropped from the returned
tensordict after consumption.

### DIFF
--- a/torchrl/collectors/_single.py
+++ b/torchrl/collectors/_single.py
@@ -11,7 +11,12 @@ from typing import Any
 
 import torch
 
-from tensordict import LazyStackedTensorDict, TensorDict, TensorDictBase
+from tensordict import (
+    LazyStackedTensorDict,
+    TensorDict,
+    TensorDictBase,
+    UnbatchedTensor,
+)
 from tensordict.nn import CudaGraphModule, TensorDictModule, TensorDictModuleBase
 from torch import nn
 from torchrl import compile_with_warmup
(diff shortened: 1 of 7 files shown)


torchrl/collectors/_single.py
```diff
@@ -275,7 +280,38 @@ class Collector(BaseCollector):
...
+            (a 1/T fraction of corruption). With ``final_obs=True``, GAE reads
+            the true ``s_T`` from ``("final", obs)`` instead.
+
+            The pipeline assumption is:
+            ``collector -> GAE(shifted=True) -> ReplayBuffer.extend()``.
+            :class:`~torchrl.objectives.value.advantages.GAE` consumes and
+            drops ``("final", ...)`` from the returned tensordict, so the
+            downstream replay buffer never sees an
+            :class:`~tensordict.UnbatchedTensor` (which would otherwise be
+            incompatible with a contiguous storage). Defaults to ``False``.
```
vmoens: mark this as experimental (can go at any time). 
You mention GAE but it could be any value estimator.

This feature is for correctness but in practice using the last root observation as trailing obs works well in practice.

gtnv:
hey @vmoens, is this still moving forward? if not, would you like to close it?

PR #3759, opened 2026-05-15
[Feature] MultiCollector: track policy version per fresh continue command

vmoens:
* #3776
* #3775
* #3769
* #3763
* #3771
* #3768
* #3767
* #3764
* __->__ #3759
* #3757
* #3762
* #3755
* #3754
* #3753

When the parent calls ``update_policy_weights_()`` and immediately
sends a ``continue`` to a worker, the resulting batch was getting
tagged with the *pre-update* policy version: the worker's policy
version is only bumped when ``increment_version()`` is called, and
that previously only fired on direct user calls. Off-policy logging,
staleness gating, and prioritized sampling that key on the version
saw a 1-step lag.

Pass ``track_policy_version=(self.policy_version_tracker is not None)``
to the worker, plumb a ``fresh_command`` flag through the recv loop
so the worker knows whether ``msg`` was just received vs replayed
from a timed-out iteration, and increment the version on each fresh
``continue`` / ``continue_random`` (only when not in ``run_free``,
where the worker drives its own pacing).

No new unit test: existing ``MultiCollector`` policy-version coverage
in ``test/test_collectors.py`` continues to assert the per-batch
version. End-to-end coverage comes through the example at the top of
the stack.


vmoens (commented):
Unmergeable, this goes against the idea of policy version tracking. Continue != update policy weight

PR #3760, opened 2026-05-15
[Example] Isaac Lab RNN PPO with compact memory + knowledge-base notes

vmoens:
* #3761
* __->__ #3760
* #3759
* #3758
* #3757
* #3756
* #3755
* #3754
* #3753

End-to-end recurrent PPO example targeting Isaac Lab, exercising the
collector / GAE features added below in the stack:

- ``compact_obs=True`` to drop redundant ``("next", obs)`` from each
  rollout step and keep the buffer affordable for vectorized envs.
- ``final_obs=True`` to carry the true window-boundary observation
  under ``("final", k)``, so ``GAE(shifted=True)`` bootstraps with
  ``V(s_T)`` rather than ``V(s_{T-1})`` at every soft boundary.
- ``recurrent_backend="scan"`` (or ``"triton"``) for the LSTM, with the
  canonical-stride and cuDNN flat-storage fixes in place.
- ``timeit.mark_start`` / ``mark_end`` to instrument the worker loop.
- ``track_policy_version=True`` on the MultiCollector so each batch is
  tagged with the policy version at collection time.

``isaaclab_rnn_ppo_memory_utils.py`` factors the network builder, the
collector factory, the GAE+critic warmup, and the cudagraph fake batch
plumbing so the main script reads as a thin orchestration loop. CLI
flags surface the compile / cudagraph / vectorized-GAE / cudagraph-trees
knobs and a configurable warmup-iterations floor (which auto-bumps to
``cudagraph_warmup + 1``).

``knowledge_base/ISAACLAB.md`` adds the pip-based container flow we use
on the cluster (uv venv when ensurepip is missing, IsaacLab pip wheels,
the EULA / privacy-consent env vars to make Kit non-interactive) and a
recurrent-PPO-on-Isaac recipe pinning the collector-owned policy_factory,
the weight-sync scheme, and the compact_obs + shifted-GAE pairing.

### DIFF
--- a/examples/collectors/isaaclab_rnn_ppo_memory.py
+++ b/examples/collectors/isaaclab_rnn_ppo_memory.py
@@ -1,0 +1,817 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""Throwaway Isaac Lab recurrent PPO memory comparison.
+
+This is intentionally an experiment harness, not a polished example.  It runs
+large-vectorized Isaac Lab PPO with an LSTM policy and records the memory gap
+between an unoptimized path and the current memory-saving path.
+"""
+from __future__ import annotations
+
+import argparse
+import logging
+import math
+import os
+import sys
+from functools import partial
+from pathlib import Path
+
+os.environ.setdefault("TORCHRL_PROFILING", "1")
+
+from isaaclab.app import AppLauncher
+
+
+def _parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        description="Isaac Lab LSTM PPO memory comparison."
+    )
+    parser.add_argument(
+        "--preset",
+        choices=["custom", "full-off", "full-on", "full-on-triton"],
+        default="custom",
+    )
+    parser.add_argument("--task", default="Isaac-Ant-v0")
+    parser.add_argument("--num-envs", type=int, default=16_384)
+    parser.add_argument("--num-collectors", type=int, default=1)
+    parser.add_argument("--rollout-steps", type=int, default=32)
+    parser.add_argument("--max-episode-steps", type=int, default=500)
+    parser.add_argument("--iterations", type=int, default=20)
+    parser.add_argument("--warmup-iterations", type=int, default=2)
+    parser.add_argument("--ppo-epochs", type=int, default=3)
+    parser.add_argument("--mini-batch-steps", type=int, default=8_192)
+    parser.add_argument(
+        "--reshape-sampled-slices",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help=(
+            "Deprecated no-op kept for old launch scripts. Minibatches are "
+            "sampled as [B, T] trajectory batches."
+        ),
+    )
+    parser.add_argument("--hidden-size", type=int, default=256)
+    parser.add_argument("--obs-dim", type=int, default=60)
+    parser.add_argument("--action-dim", type=int, default=8)
+    parser.add_argument("--lr", type=float, default=3e-4)
+    parser.add_argument("--gamma", type=float, default=0.99)
+    parser.add_argument("--gae-lambda", type=float, default=0.95)
+    parser.add_argument("--clip-epsilon", type=float, default=0.2)
+    parser.add_argument("--entropy-coeff", type=float, default=0.0)
+    parser.add_argument("--critic-coeff", type=float, default=1.0)
+    parser.add_argument(
+        "--collector-backend",
+        choices=["single", "sync", "async"],
+        default="async",
+    )
+    parser.add_argument("--storage", choices=["cuda", "cpu", "memmap"], default="cuda")
+    parser.add_argument(
+        "--collector-buffer-storage",
+        choices=["cpu", "memmap"],
+        default="memmap",
+        help="Storage used by the double-buffered collector staging buffers.",
+    )
+    parser.add_argument("--double-buffer-collector", action="store_true")
+    parser.add_argument(
+        "--storing-device",
+        default=None,
+        help=(
+            "Collector output device. Use 'cpu' to move rollout data off the "
+            "collector GPU before yielding it; leave unset for collector default."
+        ),
+    )
+    parser.add_argument(
+        "--compact-obs", action=argparse.BooleanOptionalAction, default=False
+    )
+    parser.add_argument(
+        "--shifted-gae", action=argparse.BooleanOptionalAction, default=False
+    )
+    parser.add_argument(
+        "--vectorized-gae", action=argparse.BooleanOptionalAction, default=True
+    )
+    parser.add_argument(
+        "--deactivate-gae-vmap",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help=(
+            "Use GAE's Python fallback for value-network calls. Recurrent "
+            "value networks contain reset-dependent control flow that is not "
+            "compatible with torch.vmap."
+        ),
+    )
+    parser.add_argument(
+        "--rnn-backend",
+        choices=["cudnn", "scan", "triton"],
+        default="cudnn",
+    )
+    parser.add_argument(
+        "--gae-rnn-backend",
+        choices=["auto", "same", "cudnn", "scan", "triton"],
+        default="auto",
+        help=(
+            "Recurrent backend used for the no-grad GAE value pass. 'auto' uses "
+            "a synced cudnn/pad value backbone for scan/triton training runs to "
+            "avoid repeated scan-HOP compile cost in advantage estimation."
+        ),
+    )
+    parser.add_argument(
+        "--compile-update", action=argparse.BooleanOptionalAction, default=False
+    )
+    parser.add_argument(
+        "--compile-gae",
+        action=argparse.BooleanOptionalAction,
+        default=None,
+        help=(
+            "Compile the value network used by GAE. Defaults to --compile-update "
+            "so scan/triton recurrent backends do not pay a scan body compile on "
+            "every value-network call."
+        ),
+    )
+    parser.add_argument(
+        "--compile-mode",
+        choices=["default", "reduce-overhead", "max-autotune"],
+        default="default",
+    )
+    parser.add_argument(
+        "--cudagraph-update", action=argparse.BooleanOptionalAction, default=False
+    )
+    parser.add_argument("--cudagraph-warmup", type=int, default=8)
+    parser.add_argument(
+        "--cudagraph-trees", action=argparse.BooleanOptionalAction, default=True
+    )
+    parser.add_argument(
+        "--precompile-training",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help=(
+            "Run GAE and update once on fake [B, T] data before the collector is "
+            "started, so compilation/cudagraph capture does not pollute the first "
+            "real batches."
+        ),
+    )
+    parser.add_argument("--empty-cache-after-warmup", action="store_true")
+    parser.add_argument("--train-device", default="cuda:1")
+    parser.add_argument("--seed", type=int, default=0)
+    parser.add_argument(
+        "--memmap-dir", type=Path, default=Path("/tmp/torchrl_isaac_rnn_train_buffer")
+    )
+    parser.add_argument("--wandb-project", default="torchrl-isaac-rnn-memory")
+    parser.add_argument("--wandb-group", default="full-toggle-comparison")
+    parser.add_argument("--wandb-name", default=None)
+    parser.add_argument("--wandb-mode", default=os.environ.get("WANDB_MODE", "online"))
+    parser.add_argument("--log-level", default="INFO")
+    parser.add_argument("--debug-data-flow", action="store_true")
+    AppLauncher.add_app_launcher_args(parser)
+    return parser
+
+
+launch_args = sys.argv[1:]
+if not any(arg == "--headless" or arg.startswith("--headless=") for arg in launch_args):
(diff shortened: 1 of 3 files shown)


examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+
+def _restore_local_editable_paths() -> None:
+    for module_name in ("tensordict", "torchrl"):
+        module = sys.modules.get(module_name)
+        if module is not None and getattr(module, "__file__", None) is None:
+            del sys.modules[module_name]
+
+    for path in ("/root/tensordict", "/root/td", "/root/rl-isaac"):
+        if Path(path).exists() and path not in sys.path:
+            sys.path.insert(0, path)
```
vmoens, line 184: that looks terribly hacky and ad-hoc, remove at all cost

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+import logging
+import math
+import os
+import sys
+from functools import partial
+from pathlib import Path
+
+os.environ.setdefault("TORCHRL_PROFILING", "1")
+
+from isaaclab.app import AppLauncher
```
vmoens, line 23: isaac should only run in the sub-process. Let's ditch the regular collector and just use MultiCollector with sync true or false. No isaac in-process here.

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+
+
+def _parser() -> argparse.ArgumentParser:
```
vmoens, line 26: a proper script is imports -> methods and classes -> execution. This is all mixed up

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+        train_device,
+    )
+    actor_head = make_actor_head(args.action_dim, args.hidden_size, train_device)
+    actor = make_actor_from_modules(
+        backbone,
+        actor_head,
+        args.action_dim,
+        train_device,
+    )
+    critic = make_critic_head(args.hidden_size, train_device)
```
vmoens, line 282: this could be bundled in a single "make_models" util

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+    full_value = make_full_value(backbone, critic)
+    gae_rnn_backend = args.gae_rnn_backend
+    if gae_rnn_backend == "auto":
+        gae_rnn_backend = (
+            "cudnn" if args.rnn_backend in ("scan", "triton") else args.rnn_backend
+        )
+    elif gae_rnn_backend == "same":
+        gae_rnn_backend = args.rnn_backend
```
vmoens, line 290: GAE should be same, always. If that causes an issue we need to solve the issue.

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+    if gae_rnn_backend != args.rnn_backend:
+        gae_backbone = make_backbone(
+            args.obs_dim,
+            args.hidden_size,
+            gae_rnn_backend,
+            train_device,
+        )
+        gae_full_value = make_full_value(gae_backbone, critic)
+    compile_gae_value_network = compile_gae and gae_backbone is None
+    gae_precompile_calls = 2 if compile_gae_value_network else 1
```
vmoens, line 302: my previous comment should make this a bit redundant.

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+        vectorized=args.vectorized_gae,
+        deactivate_vmap=args.deactivate_gae_vmap,
+        device=train_device,
+    )
+
+    def compute_advantages(batch: TensorDictBase):
+        with torch.no_grad(), set_recurrent_mode(True):
+            if gae_backbone is not None:
+                gae_backbone.load_state_dict(backbone.state_dict())
+            return adv_module(batch)
```
vmoens, line 348: no load_state_dict, are we mad? Let's have one single copy of our weights in a given process.

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+
+    def compute_loss(batch: TensorDictBase):
+        with set_recurrent_mode(True):
+            return loss_module(batch)
```
vmoens, line 352: the added value of a function like this is 0 or negative

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+        loss = compute_loss(batch)
+        total_loss = loss["loss_objective"] + loss["loss_entropy"] + loss["loss_critic"]
+        total_loss_value = total_loss.detach()
+        total_loss.backward()
+        actor_grad_norm = torch.nn.utils.clip_grad_norm_(
+            actor_params, max_norm=float("inf")
+        )
+        critic_grad_norm = torch.nn.utils.clip_grad_norm_(
+            critic_params, max_norm=float("inf")
+        )
```
vmoens, line 364: clip grad norm to 1 or 10 or something, use the args for that.

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+            value_data = full_value(mini_batch.clone())
+        state_value = value_data.get(loss_module.tensor_keys.value, None)
+        if state_value is None:
+            raise KeyError(
+                f"Could not find value key {loss_module.tensor_keys.value} in "
+                "the fake warmup value-network output."
+            )
+        mini_batch.set(loss_module.tensor_keys.value_target, state_value.detach())
+        mini_batch.set(loss_module.tensor_keys.advantage, torch.zeros_like(state_value))
+        return _canonicalize_batch(mini_batch)
```
vmoens, line 405: we should be able not to use that now. We've solved the issue that made it necessary separately. Let's just run update() in the training loop.

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+            action_dim=args.action_dim,
+            hidden_size=args.hidden_size,
+            device=train_device,
+        )
+        if args.compact_obs:
+            next_td = fake_rollout.get("next", None)
+            if next_td is not None and "policy" in next_td.keys():
+                next_td.del_("policy")
+        return _canonicalize_batch(fake_rollout)
+
```
vmoens, line 421: ditto looks useless to me now

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+                param.copy_(snapshot)
+        for state in optim.state.values():
+            for key, value in state.items():
+                if torch.is_tensor(value):
+                    value.zero_()
+                elif isinstance(value, int):
+                    state[key] = 0
+                elif isinstance(value, float):
+                    state[key] = 0.0
+        optim.zero_grad(set_to_none=True)
```
vmoens, line 436: ditto, ditch

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+
+    def _precompile_training() -> int:
```
vmoens, line 438: let's not precompile anything. Compile and execute and that's it. We'll swallow the compile time in the training loop, amen

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+        collector_replay_buffer = ReplayBuffer(
+            storage=_make_storage(
+                args,
+                torch.device("cpu"),
+                args.num_envs,
+                storage_kind=args.collector_buffer_storage,
+                name="collector_replay_buffer",
+            ),
+            batch_size=args.num_envs,
+        )
```
vmoens, line 510: let's remove the double buffer option. We can just use `storing_device=cpu`. It already uses a buffer in shared mem to pass the data which is a bit like what our double buffer does but more clearly.

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+
+    make_env_fn = partial(
+        make_env,
+        task=args.task,
+        num_envs=per_collector_envs,
+        max_episode_steps=args.max_episode_steps,
+        device=str(collector_device),
+    )
+    collector_rnn_backend = "cudnn"
```
vmoens, line 519: if set_recurrent_mode is False, the rnn backend should always be cudnn. That should be documented. We should not set it here, it should be automatic

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+            policy_device=collector_device,
+            no_cuda_sync=True,
+            trust_policy=True,
+            storing_device=storing_device,
+            compact_obs=args.compact_obs,
+            auto_register_policy_transforms=True,
+            track_policy_version=True,
+            replay_buffer=collector_replay_buffer,
+            extend_buffer=True,
+        )
```
vmoens, line 543: as pointed above this can be removed

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+            "training_precompile": args.precompile_training,
+            "training_compile_gae_requested": compile_gae,
+            "training_compile_gae": compile_gae_value_network,
+            "training_compile_gae_target": "value_network"
+            if compile_gae_value_network
+            else "none",
+            "training_precompile_gae_calls": gae_precompile_calls,
+            "training_precompile_update_calls": precompile_update_calls,
+        }
+    )
```
vmoens, line 592: do we need all of this? Doesn't make the code more readable

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+    try:
+        for iteration in range(total_iterations):
+            current_iteration = iteration
+            timeit.erase()
+            current_phase = "collector_policy_sync"
+            with timeit("collector_policy_sync_s"):
+                collector.update_policy_weights_(
+                    weights=TensorDict.from_module(actor).data
+                )
+                collector.increment_version()
```
vmoens, line 626: that should be called automatically by update_policy_weights_ IMO - if not we need to patch.

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+    collector_iter = iter(collector)
+    try:
+        for iteration in range(total_iterations):
+            current_iteration = iteration
+            timeit.erase()
+            current_phase = "collector_policy_sync"
+            with timeit("collector_policy_sync_s"):
+                collector.update_policy_weights_(
+                    weights=TensorDict.from_module(actor).data
+                )
```
vmoens, line 625: this looks very standard, we should be able to just pass the actor no? Doesn't update_policy_weights_ do that for us under the hood?

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+                    "collector_policy_version": collector_policy_version,
+                }
+            )
+            with timeit("iter_s"):
+                if collector_replay_buffer is None:
+                    if collected_batch is None:
+                        raise RuntimeError(
+                            "collector returned None but no replay buffer was provided"
+                        )
+                    data = collected_batch
```
vmoens, line 648: this path can be removed

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+                        )
+                        current_phase = "gae"
+                        with timeit(name="adv_s"):
+                            _assert_time_batch_shape(
+                                data,
+                                args.num_envs,
+                                args.rollout_steps,
+                                "gae_batch",
+                            )
+                            epoch_data = _canonicalize_batch(data.to(train_device))
```
vmoens, line 700: _canonicalize_batch should be called in the RNN if anything, not in the loop. Users should not care about this.

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+                            }
+                        )
+                        if args.debug_data_flow and epoch == 0:
+                            stored_data = train_buffer[:]
+                            _add_data_flow_metrics(
+                                batch_metrics, "data_flow/train_buffer", stored_data
+                            )
+                            _add_buffer_compare_metrics(
+                                batch_metrics, epoch_data, stored_data
+                            )
```
vmoens, line 731: remove all debug paths from this file

examples/collectors/isaaclab_rnn_ppo_memory.py
```diff
@@ -0,0 +1,817 @@
...
+                        for mini_batch in train_buffer:
+                            current_phase = "update_minibatch"
+                            mini_batch = _canonicalize_batch(
+                                mini_batch.to(train_device)
+                            )
```
vmoens, line 736: ditto this looks like something users will be distracted by. I believe the RNN does this op for us so we can spare it here

PR #3761, opened 2026-05-15
[Feature] Collector.fake_tensordict() / MultiCollector.fake_tensordict()

vmoens:
* __->__ #3761
* #3760
* #3759
* #3758
* #3757
* #3756
* #3755
* #3754
* #3753

Public method that returns a zero-filled tensordict shaped exactly like
one batch yielded by the collector, useful for storage initialization
and ``torch.compile`` / cudagraph warmup without having to step the env
or spin up the worker processes first.

On ``Collector`` (single):

- Reuses the existing ``_final_rollout`` template; builds it lazily via
  ``_maybe_make_final_rollout(make_rollout=True)`` even when
  ``use_buffers=False`` so the public API is consistent.
- Mirrors the rollout post-pipeline: ``_maybe_attach_final_obs``,
  ``_maybe_set_truncated``, then ``_postproc`` (which runs
  ``split_trajectories``, the user ``postproc``, and private-key
  exclusion).
- Result: env keys + policy out-keys + ``("collector", "traj_ids")``,
  ``compact_obs`` exclusions and ``final_obs`` ``UnbatchedTensor``
  leaves applied, last dim named ``"time"``.

On ``MultiCollector``:

- Builds a per-worker fake from ``create_env_fn[0]`` (mirroring the
  legacy replay-buffer init path), applies ``_add_policy_outputs_to_fake_td``,
  expands to ``(*env.batch_size, frames_per_worker)``, refines ``"time"``.
- For ``MultiSyncCollector``, stacks ``num_workers`` copies along dim 0
  (or concatenates along ``cat_results`` when an integer was provided);
  for ``MultiAsyncCollector``, returns a single worker's shape (async
  yields one batch at a time).
- Applies ``split_trajs`` / ``postproc`` / private-key exclusion to
  match the iterator pipeline.

Tests pin: shape / names / keys / zero-fill parity between
``fake_tensordict()`` and ``next(iter(collector))`` (with and without
buffers); ``compact_obs`` drops ``("next", obs)`` and ``final_obs``
attaches ``("final", obs)`` as ``UnbatchedTensor``; multi-sync stacks
along worker dim 0.

### DIFF
--- a/torchrl/collectors/_multi_base.py
+++ b/torchrl/collectors/_multi_base.py
@@ -1054,6 +1054,156 @@ class MultiCollector(BaseCollector, metaclass=_MultiCollectorMeta):
             fake_td.set(key, policy_output.get(key))
         return fake_td
 
+    @torch.no_grad()
+    def fake_tensordict(self) -> TensorDictBase:
+        """Return a zero-filled tensordict shaped like one batch from this multi-collector.
+
+        Mirrors what one iteration of this collector yields, including
+        policy out-keys, ``("collector", "traj_ids")``, ``compact_obs`` /
+        ``final_obs`` effects, ``split_trajs``, and ``postproc``.
+
+        Shape:
+
+        - :class:`~torchrl.collectors.MultiSyncCollector` with
+          ``cat_results="stack"`` (the default): ``(num_workers,
+          *env.batch_size, frames_per_worker)``, last dim named ``"time"``.
+        - :class:`~torchrl.collectors.MultiSyncCollector` with an integer
+          ``cat_results``: per-worker tensordicts concatenated along that dim.
+        - :class:`~torchrl.collectors.MultiAsyncCollector`:
+          ``(*env.batch_size, frames_per_worker)`` — async yields one
+          worker batch at a time.
+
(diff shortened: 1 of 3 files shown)


torchrl/collectors/_multi_base.py
```diff
@@ -1054,6 +1054,156 @@ def _add_policy_outputs_to_fake_td(self, fake_td):
...
+        - :class:`~torchrl.collectors.MultiAsyncCollector`:
+          ``(*env.batch_size, frames_per_worker)`` — async yields one
+          worker batch at a time.
+
+        Intended for storage initialization and ``torch.compile`` /
+        cudagraph warmup without spinning up the worker processes.
+        """
+        # Build / borrow one env to read fake_tensordict, compact-obs leaf
+        # keys, and final-obs leaf shapes from.
+        env_fn = self.create_env_fn[0]
```
vmoens, line 1081: I'm not a big fan of this implementation.
We are creating an env in the main process which defies the purpose of the MultiCollector.
If we cannot get the fake data from the inner collector easily, we should just raise a NotImplementedError. But we should not pretend we're doing that (which is the right thing) and do it via a ton of custom code on the main process.

PR #3765, opened 2026-05-15
[Feature] Add ActionScaling transform (part 1 of #1209)

Drakit0:
## Summary

Adds `ActionScaling`, a new transform that addresses the first half of #1209.

`ActionScaling` is a spec-driven action normalization layer:

- Reads the bounds of the action spec and derives `loc = (high + low) / 2` and `scale = (high - low) / 2`.
- Exposes a normalized action space to the policy: `[-1, 1]` by default, or `[0, 1]` when `standard_normal=False`.
- On the inv direction (policy -> env), rescales the normalized action back to the env's bounds: `env_action = norm * scale + loc`.
- On the forward direction (env action -> normalized, used inside replay buffers / `nn.Module` chains), applies the inverse mapping.

Unlike `ObservationNorm`, the values are derived from the spec (not from rollouts) and there is no moving-average / `init_stats` path - the issue explicitly asks for a static, spec-based transform for actions.

### Unbounded handling

A `RuntimeError` is raised when the action spec is:

- A non-`ContinuousBox` spec (discrete or categorical).
- An `Unbounded` spec.
- A `Bounded` spec where any entry of `low` is `-inf` / `torch.finfo(dtype).min`, or any entry of `high` is `+inf` / `torch.finfo(dtype).max`.
- Has any entry with `high <= low`.

This covers both fully and partially unbounded specs, including the case where `Unbounded` internally uses the `finfo` extremes (which `torch.isfinite` reports as finite).

### What changed

- `torchrl/envs/transforms/_action.py`: implements `ActionScaling`, with a `_maybe_init_from_parent` hook that walks the transform chain so that `loc` / `scale` are populated lazily on the first inv call when the transform is attached to a `TransformedEnv`. Explicit `loc` / `scale` can also be passed to the constructor (useful for replay-buffer-only usage).
- `torchrl/envs/transforms/transforms.py`, `torchrl/envs/transforms/__init__.py`, `torchrl/envs/__init__.py`: re-export the new class.
- `docs/source/reference/envs_transforms.rst`: add `ActionScaling` to the autosummary list.
- `test/transforms/test_action_transforms.py`: full `TransformBase` coverage plus class-specific tests for the scaling math, per-dimension bounds, `[0, 1]` mode, unbounded / partially-unbounded / `finfo` extremes, explicit `loc`/`scale`, multi-key rejection, and a rollout-bounds invariant.

### How to test

```
pytest test/transforms/test_action_transforms.py::TestActionScaling -x
```

22 tests, all passing locally on Windows / Python 3.12 / Torch 2.8.

Closes the first half of #1209.

### DIFF
--- a/torchrl/envs/transforms/_action.py
+++ b/torchrl/envs/transforms/_action.py
@@ -43,7 +43,6 @@ from torchrl.envs.transforms._base import (
     FORWARD_NOT_IMPLEMENTED,
     Transform,
 )
-from torchrl.envs.transforms.utils import _set_missing_tolerance
 
 __all__ = [
     "ActionDiscretizer",
@@ -869,7 +868,7 @@ class MultiAction(Transform):
 
 
 class ActionScaling(Transform):
-    """Affine-scale a continuous action using the bounds of the action spec.
+    r"""Affine-scale a continuous action using the bounds of the action spec.
 
     Given a bounded action spec with bounds ``[low, high]``, this transform exposes
     a normalized action space to the policy and rescales actions back to the
@@ -879,14 +878,14 @@ class ActionScaling(Transform):
 
     .. math::
 
-        loc = \\frac{high + low}{2}, \\quad scale = \\frac{high - low}{2}.
+        loc = \frac{high + low}{2}, \quad scale = \frac{high - low}{2}.
 
     When ``standard_normal=True`` (default) the normalized action space is
     ``[-1, 1]`` and the inverse mapping (policy action -> env action) is
 
     .. math::
 
-        a_{env} = a_{norm} \\cdot scale + loc.
+        a_{env} = a_{norm} \cdot scale + loc.
 
     The forward mapping (env action -> normalized action, used by replay buffer
     transforms) is the inverse:
@@ -1038,7 +1037,6 @@ class ActionScaling(Transform):
         # that the spec we read here reflects the env-scale bounds at the point
         # in the chain where this transform sits.
         full_action_spec = full_action_spec.clone()
-        container = self.container
         # ``parent`` is an EnvBase; walk through preceding transforms in the
         # Compose (if any) up to ``self``.
         transform = parent.transform
@@ -1061,12 +1059,9 @@ class ActionScaling(Transform):
         if key not in full_action_spec.keys(True, True):
             return
         leaf = full_action_spec[key]
-        try:
-            low, high = self._validate_bounded(leaf)
-        except RuntimeError:
-            # surface the validation error to the caller — this is the same
-            # error that would have been raised when the spec was first queried.
-            raise
+        # surface the validation error to the caller - it is the same error
+        # that would have been raised when the spec was first queried.
+        low, high = self._validate_bounded(leaf)
         dtype = low.dtype if low.dtype.is_floating_point else torch.get_default_dtype()
         loc = ((high + low) / 2).to(dtype)
         scale = ((high - low) / 2).to(dtype)
--- a/torchrl/envs/transforms/_action.py
+++ b/torchrl/envs/transforms/_action.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import functools
+from collections.abc import Sequence
 from copy import copy
 from enum import IntEnum
 from textwrap import indent
@@ -21,10 +22,12 @@ from torchrl.data.tensor_specs import (
     Bounded,
     Categorical,
     Composite,
+    ContinuousBox,
     MultiCategorical,
     MultiOneHot,
     OneHot,
     TensorSpec,
+    Unbounded,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


Drakit0:
CI is blocked by "workflows awaiting approval" (first-time contributor). Could a maintainer please approve workflow runs so the full test suite can execute against this PR? Thanks!

Drakit0:
@vmoens - would you be able to take a look when you have a moment? This PR addresses one half of #1209 (the other is #3766/#3765). I cannot use the "Request review" button as an external contributor, hence the mention. Thanks!

Drakit0:
Pushed a lint-only fix - dropped two unused imports (one of which was leftover plumbing from an earlier draft) and ran black-22.3.0. All lint hooks (ufmt, flake8, pydocstyle, autoflake) now pass locally with the versions from `.pre-commit-config.yaml`. Tests still 22/22 green.

Drakit0:
Heads up - the lint-only fix is on the new commit, but as a first-time contributor each new commit re-enters the "workflows awaiting approval" state. Could a maintainer please re-approve the new workflow run so the full matrix can verify? Lint, ufmt, flake8, pydocstyle and autoflake all pass locally with the versions from `.pre-commit-config.yaml`, and the 22 transform tests still pass. Thanks!

Drakit0:
Quick triage on the Windows failure from run #25917872026 (commit `1bcb6c1`, the one before the lint fix):

```
test/objectives/test_ddpg.py::TestDDPG::test_ddpg_prioritized_weights FAILED
  assert tensor(5.1256e-06) > 1e-05
```

Final tally: `1 failed, 21166 passed, 6238 skipped`. The single failure is a numerical tolerance check on DDPG prioritized-buffer weight std (5.13e-6 vs. the 1e-5 threshold) - statistical flake driven by weight init / sampling, completely unrelated to `ActionScaling`. All 22 `TestActionScaling` tests pass on Windows, and `main`'s recent Windows runs are green, which is consistent with this being a sporadic flake on that DDPG test.

Should be cleared by a re-run; no action needed from me on this side.

torchrl/envs/transforms/_action.py
```diff
@@ -857,3 +865,321 @@ def _transform_observation_spec(
...
+            shape=action_spec.shape,
+            device=action_spec.device,
+            dtype=action_spec.dtype,
+        )
+
+    def _apply_transform(self, action: torch.Tensor) -> torch.Tensor:
+        # env action -> normalized
+        self._check_initialized()
+        loc = self.loc.to(action.device)
+        scale = self.scale.to(action.device)
```
vmoens: let's figure out a way to do this in a compile friendly manner. eg call `to()` only if needed (if device differs).

torchrl/envs/transforms/_action.py
```diff
@@ -857,3 +865,321 @@ def _transform_observation_spec(
...
+        normalized = (action - loc) / scale
+        if not self.standard_normal:
+            normalized = (normalized + 1) / 2
+        return normalized
+
+    def _inv_apply_transform(self, action: torch.Tensor) -> torch.Tensor:
+        # normalized -> env action
+        self._check_initialized()
+        loc = self.loc.to(action.device)
+        scale = self.scale.to(action.device)
```
vmoens: ditto

vmoens (commented):
LGTM, just a minor thing to fix for compile friendliness

torchrl/envs/transforms/_action.py
Drakit0: Done in e9b914a0. Extracted a `_loc_scale(device)` helper that only calls `.to()` when `self.loc.device != device`. The device comparison resolves at trace time (device is static metadata, not data), so the common same-device path is copy-free and inserts no device-copy node in the compile graph. Verified with `torch.compile(fullgraph=True)` that the branch causes no graph break.

torchrl/envs/transforms/_action.py
Drakit0: Same fix applies here - both `_apply_transform` and `_inv_apply_transform` now go through `_loc_scale(device)` (e9b914a0).

torchrl/envs/transforms/_action.py
Drakit0: I just changed it!

torchrl/envs/transforms/_action.py
Drakit0: ditto

Drakit0:
Friendly Monday bump, @vmoens - both PRs for #1209 are currently stuck behind "workflows awaiting approval", so CI hasn't been able to run on the latest commits yet:

- **#3765 (ActionScaling)**: your compile-friendliness feedback is addressed in e9b914a0 - `loc`/`scale` are now moved via a `_loc_scale(device)` helper that only calls `.to()` when the device actually differs. Verified with `torch.compile(fullgraph=True)` that the device check causes no graph break. Both review threads have replies.
- **#3766 (FlattenAction)**: no outstanding review comments; it just needs a first look.

Whenever you have a moment to approve the workflow runs (and re-review), that would be much appreciated. Thanks for the review on Friday!

Drakit0:
CI triage - every unit-test job passed here: `tests-cpu` (3.10-3.14), all `tests-gpu` / `tests-stable-gpu` jobs, `unittests-cpu` (Windows), `tests-olddeps`, `tests-optdeps`, `unittests-gym`, `unittests-sklearn`, SOTA, tutorials and docs are all green.

The only red checks are `build-wheel-py3_10-cpu` / `upload-wheel-py3_10-cpu` on Windows - that is the Windows wheel-builder, unrelated to the diff (the Linux and macOS/aarch64 wheel builds for this PR all succeeded). A re-run should clear them.

So this PR is functionally green - the compile-friendliness fix from your review is in (e9b914a0) and verified.

vmoens (approved):
LGTM thanks

Drakit0:
I tested your refactor on local and it's way cleaner than mine, tku!

PR #3766, opened 2026-05-15
[Feature] Add FlattenAction transform (part 2 of #1209)

Drakit0:
## Summary

Adds `FlattenAction`, a new transform that addresses the second half of #1209.

`FlattenAction` mirrors `FlattenObservation`, but on the action side:

- The policy sees an action space whose adjacent dimensions `[first_dim, last_dim]` are merged into one.
- On the inv direction (policy -> env), the flat action is unflattened back to the original env-side shape before it reaches the base environment.
- On the forward direction (env action -> flat, used inside replay buffers and `nn.Module` chains), the same dims are flattened.
- Both flat (`\"action\"`) and nested (e.g. `(\"nested\", \"action\")`) action keys are supported, which is the issue's mention of dict-structured action spaces.

### Design

The per-key unflatten span is recovered from the action spec when the transform is first attached to a `TransformedEnv` (lazy init via `_maybe_init_from_parent`, mirroring the rest of the spec-driven transforms). For replay-buffer-only usage (no parent env), the user can populate `self._unflatten_shapes` directly.

`first_dim` / `last_dim` follow the same conventions as `FlattenObservation`:

- Default to negative dimensions so the transform composes with envs of different batch sizes.
- Positive dims require `allow_positive_dim=True` and explicit caller intent.

### What changed

- `torchrl/envs/transforms/_action.py`: implements `FlattenAction`. The action spec is rewritten in `transform_action_spec` with manual iteration (instead of `@_apply_to_composite_inv`) so that nested keys work the same way as top-level ones.
- `torchrl/envs/transforms/transforms.py`, `torchrl/envs/transforms/__init__.py`, `torchrl/envs/__init__.py`: re-export the new class.
- `docs/source/reference/envs_transforms.rst`: add `FlattenAction` to the autosummary list.
- `test/transforms/test_action_transforms.py`: full `TransformBase` coverage plus class-specific tests for spec shape, 2-D / 3-D actions (partial and full flatten), 1-D no-op flatten, nested action keys, RB round-trip, positive-dim rejection, and a rollout shape check. Adds a small `_MultiDimActionEnv` helper for the multi-dimensional cases.

### How to test

```
pytest test/transforms/test_action_transforms.py::TestFlattenAction -x
```

22 tests, all passing locally on Windows / Python 3.12 / Torch 2.8.

Closes the second half of #1209.


Drakit0:
CI is blocked by "workflows awaiting approval" (first-time contributor). Could a maintainer please approve workflow runs so the full test suite can execute against this PR? Thanks!

Drakit0:
@vmoens - would you be able to take a look when you have a moment? This PR addresses one half of #1209 (the other is #3765). I cannot use the "Request review" button as an external contributor, hence the mention. Thanks!

Drakit0:
Pushed a lint-only fix - dropped two unused imports (one of which was leftover plumbing from an earlier draft) and ran black-22.3.0. All lint hooks (ufmt, flake8, pydocstyle, autoflake) now pass locally with the versions from `.pre-commit-config.yaml`. Tests still 22/22 green.

Drakit0:
Heads up - the lint-only fix is on the new commit, but as a first-time contributor each new commit re-enters the "workflows awaiting approval" state. Could a maintainer please re-approve the new workflow run so the full matrix can verify? Lint, ufmt, flake8, pydocstyle and autoflake all pass locally with the versions from `.pre-commit-config.yaml`, and the 22 transform tests still pass. Thanks!

Drakit0:
CI triage - none of the red checks are caused by this PR; all three are flakes / infra:

- **`tests-cpu (3.12)` -> `test_custom_envs.py::TestPendulum::test_pendulum_env[None|device1]`**: not touched by this PR (FlattenAction only adds a transform in `torchrl/envs/transforms/_action.py` plus its tests). I reproduced it locally on this exact branch: `test_pendulum_env` passes 2/2, and still passes 24/24 when run in the same process immediately after the full `TestFlattenAction` suite - so it is not test pollution from the new tests either. It is a flaky `check_env_specs` on the tutorial Pendulum env.
- **`tests-gpu-distributed (3.12, 13.0)` -> `test_distributed.py::TestRayCollector::test_ray_replaybuffer`**: `CUDA error: CUDA-capable device(s) is/are busy or unavailable` - a GPU-runner infra flake. The same `tests-gpu-distributed` job failed on `main` itself at the base commit `996387f0`.
- **`build-wheel-py3_10-cpu` (Windows)**: Windows wheel-builder infra flake - the Linux wheel build for the same PR succeeded.

A re-run of the failed jobs should clear all three. The branch also merges cleanly into current `main` (15 commits behind, no conflicts) - happy to rebase onto latest `main` for a fresh run if you'd prefer.

vmoens (approved):
LGTM thanks

Drakit0:
Thanks for the reviews

PR #3777, opened 2026-05-18
[Feature][Performance] NextObservationDelta env transform

vmoens:
## Summary

- Adds `NextObservationDelta`, a stateless env-side transform that stores `("next", obs)` as a low-precision delta from the root `obs` for rollout memory savings on large continuous observations.
- Wires up the previously-stubbed `_post_step_mdp_hooks` extension point in `EnvBase.step_and_maybe_reset` and threads it through `Transform`, `Compose`, and `TransformedEnv`. The hook receives both the post-step and post-step-mdp tensordicts so a transform can rehydrate the flowing td that the policy reads on the next iteration.
- `NextObservationDelta._step` writes `(next_obs - obs).to(delta_dtype)` (default `float16`); `_post_step_mdp_hooks` reconstructs `obs + delta` in `restore_dtype` (default: match root). Stateless — no caching across steps.

## Why this shape

The existing `compact_obs` collector flag + `NextStateReconstructor` RB transform attack the same problem by dropping `("next", obs)` entirely and shifting at sample time. That is zero-storage but lossy at trajectory boundaries (which become `NaN`). The delta variant trades a small precision loss for boundary-preserving reconstruction and an env-side hook that does not need to know about collector internals.

The `_post_step_mdp_hooks` mechanism was already stubbed (commented out) in `common.py`, `transforms/_base.py`, and `llm/chat.py`. This PR enables it. The signature was changed from the original comment (`(tensordict_,) -> tensordict_`) to `(tensordict, tensordict_) -> tensordict_` because rehydration needs read access to the post-step root obs. No caller existed before, so this is not a breaking change.

## v1 limitations (documented on the class)

- **Lossy.** Round-trip error scales with `delta_dtype` precision and observation magnitude.
- **Memory savings require non-pre-allocated stacked output.** `SyncDataCollector(use_buffers=False)` or a lazy RB storage. Pre-allocated `_final_rollout` upcasts the write back to the original dtype and erases the saving.
- **Hook fires from `step_and_maybe_reset` only.** `env.rollout()` is not wired in v1; direct rollout callers must rehydrate manually.
- **`check_env_specs` does not pass on the transformed env.** `observation_spec` is shared between root and `("next", ...)` in TorchRL; the transform does not fork it in v1 (a follow-up could). Tests use a reset+step smoke instead.
- **Batched-env composition.** For `SerialEnv`/`ParallelEnv`, the transform belongs *outside* the batched env (i.e. `TransformedEnv(ParallelEnv(...), NextObservationDelta())`) — that path uses the outer `step_and_maybe_reset` and the hook fires. Putting the transform *inside* each worker is allowed and runs without error, but the outer batched env's `step_and_maybe_reset` does not currently propagate the hook so the stacked output upcasts.

## Out of scope (potential follow-ups)

- Forking `observation_spec` so pre-allocated `_final_rollout` benefits from the compression.
- Wiring the hook in `_rollout_stop_early` and in `batched_envs` / `async_envs` / `envpool` `step_and_maybe_reset`.
- A replay-buffer-side delta transform paired with this one.
- Benchmark entry under `benchmarks/`.

## Test plan

- [x] `pytest test/transforms/test_observation_transforms.py::TestNextObservationDelta` — 14 passed, 2 documented skips.
- [x] `pytest --doctest-modules torchrl/envs/transforms/_observation.py -k NextObservationDelta` — passes.
- [x] `pytest test/envs/test_env_base.py` — 47 passed, 4 skipped (no regressions from the hook wiring).
- [x] Manual smoke against `GymEnv("Pendulum-v1")` confirms `("next", "observation").dtype == torch.float16` post-step and `torch.float32` on the flowing td, with bitwise-exact rehydration (`max diff 0.0`).
- [x] `Compose(NextObservationDelta, RewardSum)` works in both orderings.
- [ ] Wider CI sweep (compose + env-transforms suites) — local disk filled before completing; relying on CI.


vmoens:
@elin-bdai @theap06 maybe you could help review this one?
It reduces the size of `("next", "observation")` fields by 50% when the transform is applied by using low-prec representation of the delta between `o_t` and `o'_t`
That looks like a good compromise between efficiency and correctness to carry the last obs of a trajectory/rollout.

torchrl/envs/transforms/_observation.py
```diff
@@ -1419,3 +1420,236 @@ def __repr__(self) -> str:
...
+
+    Example:
+        >>> import torch
+        >>> from torchrl.envs import GymEnv, TransformedEnv
+        >>> from torchrl.envs.transforms import NextObservationDelta
+        >>> env = TransformedEnv(GymEnv("Pendulum-v1"), NextObservationDelta())
+        >>> td_root = env.reset()
+        >>> _ = td_root.set("action", env.action_spec.rand())
+        >>> td, td_ = env.step_and_maybe_reset(td_root)
+        >>> td["next", "observation"].dtype
```
elin-bdai: If I'm understanding correctly, I think it's confusing here when using `NextObservationDelta()` that what's inside td["next", "observation"] is the delta, but the tensordict is indistinguishable from when you don't use `NextObservationDelta`, so you're not sure if it's the delta or not stored in there. It could lead to confusion when inspecting the outputs at different points.

torchrl/envs/transforms/_observation.py
```diff
@@ -1419,3 +1420,236 @@ def __repr__(self) -> str:
...
+            if obs is None or next_obs is None:
+                continue
+            if self.auto_skip and next_obs.dtype == self.delta_dtype:
+                continue
+            # Subtract in the source (typically full-precision) dtype, then
+            # cast once. This loses fewer significant bits than casting each
+            # operand to ``delta_dtype`` first and subtracting in low precision
+            # (which would risk catastrophic cancellation for nearby values).
+            delta = (next_obs - obs).to(self.delta_dtype)
+            next_tensordict.set(key, delta)
```
elin-bdai: Would it make sense to change the `key` here to `{key}_delta`?

elin-bdai (commented):
Thanks for doing this! I'm going to test this with our longer training jobs this week to make sure loss of precision is not a problem. Just a comment in terms of reducing confusion.

theap06:
@vmoens Thanks for building this out! @elin-bdai on your first point, agree that it's easy to miss. The dtype change is the observable signal, and the docstring example showing td["next", "observation"].dtype helps.(personally, dont think its necessary) On the {key}_delta suggestion, I also agree here as it makes the off-policy footgun explicit.

theap06 (approved):

elin-bdai:
> @vmoens Thanks for building this out! @elin-bdai on your first point, agree that it's easy to miss. The dtype change is the observable signal, and the docstring example showing td["next", "observation"].dtype helps.(personally, dont think its necessary) On the {key}_delta suggestion, I also agree here as it makes the off-policy footgun explicit.

Good point. Yeah we can make the tradeoff of whether it's worth it. Thanks!

elin-bdai (approved):

vmoens:
I agree i'm working on making a `("next", "delta", "observation")` kind of thing such that we never have that hidden bug and things are explicit. Funny that we all converged on a similar idea!
Codex has been pushing back for this but I think it's much cleaner.

PR #3788, opened 2026-05-23
[Refactor] Unify MAPPOLoss + IPPOLoss into MultiPPOLoss with deprecat…

theap06:
Addresses the Proposed field in #3748 

@matteobettini lmk what you think

### DIFF
--- a/torchrl/objectives/multiagent/mappo.py
+++ b/torchrl/objectives/multiagent/mappo.py
@@ -4,8 +4,12 @@
 # LICENSE file in the root directory of this source tree.
 """Multi-agent PPO objectives.
 
-Implements :class:`MAPPOLoss` (centralised critic) and :class:`IPPOLoss`
-(independent / decentralised critic).
+Implements :class:`MultiPPOLoss` — the unified multi-agent PPO loss that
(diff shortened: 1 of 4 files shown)


torchrl/objectives/multiagent/mappo.py
```diff
@@ -204,39 +223,46 @@ def __init__(
...
+    """Deprecated alias for ``MultiPPOLoss(critic_type='independent')``.
+
+    .. deprecated:: 0.9
+        Use :class:`MultiPPOLoss` with ``critic_type='independent'`` instead.
+        ``IPPOLoss`` will be removed in v0.11.
     """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        warnings.warn(
+            "IPPOLoss is deprecated and will be removed in v0.11. "
```
vmoens, line 262: lol the next release is 0.13 ;)

vmoens (commented):
I'm not against having the two loss modules, even if they're (almost) identical. It's clearer from an API perspective. We should not deprecate them

theap06:
> I'm not against having the two loss modules, even if they're (almost) identical. It's clearer from an API perspective. We should not deprecate them

I'll close the PR then!

PR #3825, opened 2026-06-09
[CI] Bound gh-pages clone size: cap benchmark data.js + scheduled squash

vmoens:
## Why

The `gh-pages` branch was the dominant contributor to this repo's clone size — it carried ~83% of the reachable git history. It is fully auto-generated, and two sources on it grow without bound:

1. **Benchmark history.** `benchmark-action/github-action-benchmark` appends to `dev/bench/data.js` on every push to `main` and never trims it. The file had grown to ~36 MB holding 800+ data points (486 CPU + 355 GPU commits), and that single file is re-stored on every benchmark commit.
2. **Doc-build commits.** `docs.yml` commits a fresh `auto-generating sphinx docs` snapshot to `gh-pages` on every push to `main`/tags, so the branch accumulated 1200+ commits whose only lasting value is the latest snapshot.

A one-time history squash + `data.js` trim was applied directly to the `gh-pages` branch (out of band, since `gh-pages` history isn't part of a normal PR), taking a full clone from ~375 MiB to ~58 MiB. **This PR prevents it from creeping back.**

## What

- **`benchmarks.yml`**: add `max-items-in-chart: 100` to both *Store … benchmark results* steps, so `data.js` stays ~8 MB regardless of how many runs accumulate.
- **`gh-pages-squash.yml`** (new): a monthly (and manually dispatchable) job that collapses `gh-pages` to a single, content-identical orphan commit via `--force-with-lease`, dropping accumulated commit history while leaving the published site byte-for-byte unchanged. Guarded with `if: github.repository == 'pytorch/rl'` so it never runs on forks, and a no-op when `gh-pages` is already a single commit.

## Notes

- **No effect on published content** — the benchmark chart's data lives inside `data.js` (capped, not removed), and the squash preserves the current tree exactly.
- The clone-size reduction from the existing history fully materializes after GitHub's own background `gc`.
- Considered but intentionally skipped: rewriting `main` history (would reclaim ~15–20 MiB but rewrites every commit SHA / breaks every fork & open PR — not worth it), and de-tracking the `test/assets/*.zip` fixtures (no clone-download benefit since the blobs remain in history, and it would make `test_rlhf.py` network-dependent).

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -186,6 +186,9 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           gh-pages-branch: gh-pages
           auto-push: true
+          # Cap the on-disk benchmark history (dev/bench/data.js) so it cannot
+          # grow without bound and bloat the repository / clone size.
+          max-items-in-chart: 250
       - name: Store GPU benchmark results
         uses: benchmark-action/github-action-benchmark@v1
         with:
@@ -199,5 +202,8 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           gh-pages-branch: gh-pages
           auto-push: true
+          # Cap the on-disk benchmark history (dev/bench/data.js) so it cannot
+          # grow without bound and bloat the repository / clone size.
+          max-items-in-chart: 250
           # Use regular comments instead of PR reviews to avoid permission issues
           comment-style: 'github'
--- a/.github/workflows/gh-pages-squash.yml
+++ b/.github/workflows/gh-pages-squash.yml
@@ -1,0 +1,62 @@
+name: Squash gh-pages history
+
+# The gh-pages branch is fully auto-generated: docs.yml commits a fresh
+# "auto-generating sphinx docs" snapshot on every push to main/tags, and the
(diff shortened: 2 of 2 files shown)


vmoens:
### Note on CI permissions for the force-push (reviewers / post-merge)

The squash job force-pushes `gh-pages` using the default `GITHUB_TOKEN` (`permissions: contents: write`). Verified so far:

- `gh-pages` has **no** classic branch protection.
- The repo's only ruleset ("Meta CLA Enforcement") **explicitly excludes `refs/heads/gh-pages`**, is in `evaluate` (non-blocking) mode, and only enforces status checks.
- `docs.yml` and `benchmarks.yml` already push to `gh-pages` with the same token, so write access is established.

**Not verifiable without `admin:org`:** whether a pytorch **org-level** ruleset blocks force-pushes (`non_fast_forward`) on all branches. A maintainer's manual force-push can bypass such a rule while `GITHUB_TOKEN` cannot.

**Plan:** after merge, trigger once via `workflow_dispatch` to confirm the force-push succeeds. The job is idempotent and `--force-with-lease`-guarded, so the test run is safe. If an org ruleset blocks it, switch the push to a bypass-capable token (fine-grained PAT with Contents: write, or the `pytorchbot` App) via `actions/checkout`'s `token:` input.

Issue #1062, opened 2023-04-15
[Feature Request] Make sure that all losses work with tensorclasses and regular tensors


vmoens:
## Motivation

We should add tests for all losses with
- [ ] tensors passed via `tensordict.nn.dispatch`. This should be accompanied by an example in the docstrings of how to use the losses without tensordict. 
- [ ] Similarly, it should be easy to pass regular nn.Modules to the losses (eg an actor that reads an obs and outputs an action, a qvalue that reads an obs and an action and outputs a value, etc).
- [ ] We also need to check that losses work with tensorclasses. This essentially requires us to check that no `__getitem__` is used for `get(key)`, ie the latter should always be preferred. Tests should all include a tensorclass version.

andrewldesousa:
hey @vmoens is this still a relevant issue?

vmoens:
Yeah I don't think we really test that!

aehebald:
hi! @vmoens  I am going to grab this if that is okay with you!

PR #3864, opened 2026-06-13
[Feature] Tensorclass support for IQLLoss

aehebald:
## Description
  
  Makes `IQLLoss`/`DiscreteIQLLoss` accept tensorclass inputs, not just `TensorDict`.
  First loss converted as a template for the rest of #1062.

  - Read keys with `.get()` instead of `td[key]` (`[]` is positional indexing on a tensorclass).
  - Add `_make_writable()` in `objectives/utils.py`; route network scratch selections through it
    (a tensorclass rejects undeclared out_keys, so convert to `TensorDict`; dynamic containers pass through).
  - Parametrize `test_iql` over a tensorclass input + add a tensorclass/TensorDict parity test.

  ## Motivation and Context

  #1062 asks for tensorclass support across all losses. This does IQL first so the
  approach can be reviewed before applying it to the rest.

  Part of #1062
  
  ## Types of changes

  ## Checklist

  - [x] I have read the CONTRIBUTION guide (**required**)

### DIFF
--- a/test/objectives/test_iql.py
+++ b/test/objectives/test_iql.py
@@ -18,7 +18,7 @@ from _objectives_common import (
     LossModuleTestBase,
 )
 
-from tensordict import assert_allclose_td, TensorDict
+from tensordict import assert_allclose_td, tensorclass, TensorDict
 from tensordict.nn import (
     NormalParamExtractor,
     ProbabilisticTensorDictModule as ProbMod,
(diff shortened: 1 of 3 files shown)


vmoens (commented):
Hey! Thanks for contributing this.

I'd like to hold off here. To accept a tensorclass we'd need the user to pre-register internal keys like td_error and _log_prob on their class (or change the way the loss writes them). Since these write-backs are how essentially every off-policy loss behaves, this really shapes the whole #1062 approach rather than just IQL, so I'd rather settle the pattern first.

A few options: (1) skip these writes when the input is a tensorclass, (2) error out clearly if the fields aren't pre-declared, or (3) return them as metadata. I lean toward (1) — though td_error is the tricky one, since prioritized replay reads it back from the batch, so that case needs a bit more thought.

aehebald:
@vmoens

Thanks! Dug into the code and here's what I have in mind. The framing I landed on is that the loss could just avoid writing into its input container at all. Reads are fine on a tensorclass, it's only the writes that hit the fixed schema. There are three of them and I think each has a reasonable home:

The network outputs (`loc`/`scale` from the actor, `state_value`/`state_action_value` from the value/qvalue nets) only need somewhere writable to land. So instead of converting the input, I'd have the loss run each network in its own scratch `TensorDict` built from the inputs it reads. The input is only ever read from.

`_log_prob` is purely internal, since `forward` only reads it back to log entropy, so I'd return it through the metadata dict `actor_loss` already returns (the same channel `qvalue_loss` uses for `td_error`) instead of writing it into the input.

`td_error` is the one you flagged, since prioritized replay reads it back. Since `update_priority(index, priority)` takes them as separate args, the priority doesn't need to sit on the batch, so I'd surface it in the loss output, keep writing it in place when the input accepts new keys (TensorDict unchanged), and skip that write for a tensorclass. A tensorclass user then pulls the index off the sample and the priority off the output. The one thing that wouldn't work is the `update_tensordict_priority(batch)` convenience method for a tensorclass, but `update_priority` covers it.

How does this sound to you?

gtnv:
hey @aehebald, is this still something you're working on? if not, should we close it?

PR #3867, opened 2026-06-13
[BugFix] ParallelEnv over MPS envs: default to use_buffers=False, stage pipe data on CPU

### LINKED ISSUE #3066: [BUG] Data collection error on M1 Macs
## Describe the bug

SyncDataCollector RuntimeError on Mac-M1 when using `mps` device. 

## To Reproduce
```python
from pettingzoo.butterfly import pistonball_v6
from torchrl.envs import ParallelEnv
from torchrl.envs.libs.pettingzoo import PettingZooWrapper


DEVICE = "mps"

kwargs = {"n_pistons": 21, "continuous": True}
inner_env_fn = lambda: PettingZooWrapper(
    env=pistonball_v6.parallel_env(**kwargs),
    return_state=True,
    group_map=None,
    device=DEVICE,
)
collector_env = lambda: ParallelEnv(num_workers=2, create_env_fn=inner_env_fn)
collector = SyncDataCollector(collector_env, frames_per_batch=100)

collector.reset()
print(next(iter(collector)))
collector.shutdown()
```
```bash
  File ".../.venv/lib/python3.11/site-packages/torch/storage.py", line 447, in wrapper
    return fn(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.11/site-packages/torch/storage.py", line 526, in _share_filename_cpu_
    return super()._share_filename_cpu_(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: _share_filename_: only available on CPU
```

## Expected behavior
Should be able to create env on mps and collect data. 

## System info

```
>>> print(torchrl.__version__, numpy.__version__, sys.version, sys.platform)
0.8.0+83cb41d 2.0.2 3.11.8 (main, Feb 25 2024, 03:55:37) [Clang 17.0.6 ] darwin
```
```
>>> platform.platform()
'macOS-15.5-arm64-arm-64bit'
```

Running via `uv`

## Additional context

- No issue when collecting directly on cpu or cuda 
- No issue when using SerialEnv over ParallelEnv

## Reason and Possible fixes

Seems related to https://github.com/pytorch/pytorch/issues/87688

## Checklist

discobot:
## Description

`ParallelEnv` over `mps` sub-envs (and hence collectors) crashes with
`RuntimeError: _share_filename_: only available on CPU`. This implements the design
from the issue -- check the spec-leaf devices (via `EnvMetaData.device_map`), warn and
default to `use_buffers=False` when unspecified, raise on explicit `use_buffers=True`,
with `configure_parallel` running the same check and `SerialEnv` untouched (in-process
MPS buffers are fine).

Two findings beyond the thread:

- `use_buffers=False` alone does not avoid the crash: every pipe send of an MPS
  tensordict (worker results in `_run_worker_pipe_direct`, parent actions in
  `_step_no_buffers`/`_reset_no_buffers`) goes through the same mp reduction
  (`reduce_storage` -> `_share_filename_cpu_()`) and fails identically. The no-buffers
  pipe traffic is therefore staged on CPU on both sides and cast back to the env device
  on reception. Staging has to happen *before* `consolidate()`: consolidation pickles
  the pre-consolidation device metadata, so `consolidate(device="cpu")` on its own
  makes the receiver believe the data is already on MPS and skip the cast back.
- The multi-task branch of `_get_metadata` unconditionally overwrote an
  already-resolved `use_buffers=False`; it now honors it, matching the single-task
  branch (without this the MPS default would be reverted; it also affects a
  user-passed `False`).

Tests: `TestParallelEnvMPSBuffers` in `test/envs/test_special.py` covers the
default/raise logic on CPU-only runners by faking MPS device-map entries;
`TestMPSSubEnvs` (gated on MPS availability) runs reset/rollout/collection over MPS
sub-envs end to end, including the collector setup from the issue, with a
parent-worker round-trip check. Verified on an M-series Mac (macOS 26, torch 2.12).

## Motivation and Context

Fixes #3066.

## Types of changes

## Checklist


torchrl/envs/batched_envs.py
vmoens, line 1943: This never casts back 
The worker now stages td/root_next_td to CPU for the step_and_maybe_reset message, but the parent reception in _step_and_maybe_reset_no_buffers never casts back to self.device.
Claude seems to be saying that this is a dead method anyway so maybe we don't really care

vmoens (approved):
LGTM thanks

vmoens:
Looks more involved than I initially thought - iterating on this

PR #3886, opened 2026-06-20
[BugFix] Normalize reward loss over valid pairs

fallintoplace:
## Summary
- Track how many chosen/rejected pairs actually contribute to `GPT2RewardModel.compute_reward_loss`.
- Normalize by that valid pair count instead of the original batch size when identical pairs are skipped.
- Keep all-identical batches returning a scalar zero, while preserving an autograd connection to the reward tensors.

## Root cause
The identical-sequence fix correctly skipped pairs with no preference signal, but the final loss still divided by the original batch size. Mixed batches with skipped samples therefore silently scaled the loss and gradients down.

## Tests
- `/Users/hoangvu/.pyenv/versions/3.10.13/bin/python -m pytest test/test_rlhf.py::test_compute_reward_loss_identical_sequences test/test_rlhf.py::test_compute_reward_loss_normalizes_by_non_identical_sequences`
- `/Users/hoangvu/.pyenv/versions/3.10.13/bin/python -m pytest test/test_rlhf.py` (`2 passed, 75 skipped`; optional LLM dependencies unavailable locally)

### DIFF
--- a/test/test_rlhf.py
+++ b/test/test_rlhf.py
@@ -444,13 +444,15 @@ def test_compute_reward_loss_normalizes_by_non_identical_sequences():
         [
             [0.0, 0.0, 10.0, 10.0, 0.0],
             [0.0, 0.0, 2.0, 2.0, 0.0],
-        ]
+        ],
+        requires_grad=True,
     )
     rejected_rewards = torch.tensor(
         [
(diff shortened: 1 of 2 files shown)


theap06 (approved):

vmoens (approved):
LGTM

PR #3893, opened 2026-06-21
[Refactor] Add structured inference server config objects

vmoens:
* #3942
* #3940
* #3941
* #3897
* #3896
* #3895
* #3894
* __->__ #3893
* #3939


docs/source/reference/modules_inference_server.rst
vmoens, line 1: this is worth a paragraph in the doc somewhere

torchrl/modules/inference_server/_server.py
```diff
@@ -103,8 +113,39 @@ def __init__(
         stats_window_size: int = 1024,
         weight_sync=None,
         weight_sync_model_id: str = "policy",
+        server_config: InferenceServerConfig | None = None,
+        device_config: InferenceDeviceConfig | None = None,
         shutdown_event: threading.Event | MPEvent | None = None,
     ):
+        if server_config is not None:
```
theap06: This doesn't enforce mutual exclusivity the way the error message claims. It compares the kwargs' values to the defaults, so it only fires when a kwarg differs.

torchrl/modules/inference_server/_server.py
```diff
@@ -103,8 +113,39 @@ def __init__(
...
+            if (
```
theap06: As mentioned in the comment above, this doesn't really solve mutual exclusivity. I'd suggest defaulting these kwargs to a private sentinel and checking if they have been set instead of comparing values. We can implemented these changes for ProcessInferenceServer.init and AsyncBatchedCollector.init.

torchrl/modules/inference_server/_server.py
```diff
@@ -103,8 +113,39 @@ def __init__(
...
+                device is not None
+                or policy_device is not None
+                or output_device is not None
+            ):
+                raise ValueError(
+                    "device_config is mutually exclusive with device, "
+                    "policy_device, and output_device."
+                )
+            policy_device = device_config.policy_device
+            output_device = device_config.server_output_device()
```
theap06, line 176: server_output_device() substitutes env_device for output_device when the latter is unset. I'd either reject them when a server consumes the config, or document on the server that those two fields are accepted but ignored.

test/test_inference_server.py
```diff
@@ -257,6 +259,23 @@ def test_stats_accounting(self):
...
+        with InferenceServer(
+            policy,
+            transport,
+            server_config=server_config,
+            device_config=device_config,
+        ) as server:
+            client = transport.client()
+            result = client(TensorDict({"observation": torch.randn(4)}))
+            stats = server.stats()
+        assert result["action"].device.type == "cpu"
```
theap06, line 307: This test would still pass if server_output_device(), the output_device move, or the _env_loop device moves were completely broken. To have them work correctly, either mock the tensordict and assert .to(target) was called with the expected device, or add a @pytest.mark.gpu variant that moves CPU to CUDA and checks the result landed on the right device.

theap06 (changes requested):

torchrl/modules/inference_server/_config.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def _as_device(device: torch.device | str | None) -> torch.device | None:
+    if device is None:
+        return None
+    return torch.device(device)
+
+
+@dataclass
+class InferenceDeviceConfig:
```
vmoens, line 20: this looks like something we should also accept in regular collectors
Should we think about softly deprecating explicit devices in the kwargs and replace them all by this?
How do we handle collectors with many sub-collectors, should they take a list of `InferenceDeviceConfig` objects?

torchrl/modules/inference_server/_config.py
```diff
@@ -0,0 +1,105 @@
...
+        ...     nn.Linear(4, 2), in_keys=["observation"], out_keys=["action"]
+        ... )
+        >>> transport = ThreadingTransport()
+        >>> device_config = InferenceDeviceConfig(
+        ...     policy_device="cpu", output_device="cpu"
+        ... )
+        >>> with InferenceServer(policy, transport, device_config=device_config):
+        ...     client = transport.client()
+        ...     result = client(TensorDict({"observation": torch.randn(4)}))
+        >>> result["action"].device.type
```
vmoens, line 60: the interesting thing here is also how does this play with a collector or a evaluator that have similar device arguments. I'm a bit worried that the API becomes bloated with these classes unless we unify things and decrease (not increase) boilerplate on the user side.

torchrl/modules/inference_server/_config.py
```diff
@@ -0,0 +1,105 @@
...
+        collect_stats (bool, optional): whether to collect lightweight
+            throughput and latency stats. Defaults to ``True``.
+        stats_window_size (int, optional): number of recent timing samples kept
+            for percentile stats. Defaults to ``1024``.
+
+    Examples:
+        >>> from torchrl.modules.inference_server import InferenceServerConfig
+        >>> config = InferenceServerConfig(max_batch_size=8, timeout=0.001)
+        >>> config.max_batch_size
+        8
```
vmoens, line 124: this is 100% useless. I want to see a real example with sever execution

torchrl/modules/inference_server/_config.py
vmoens, line 1: We need to think better about integration of these things. If we want to use `InferenceDeviceConfig` then we need to bring the collector and evaluators to a similar stage and think carefully about integration + deprecation.

docs/source/reference/modules_inference_server.rst
vmoens: Added a Structured Configuration section to the inference-server docs page covering both config objects, their field split between server and collector, and a usage example.

torchrl/modules/inference_server/_server.py
vmoens: Reworked: the batching kwargs now default to None and fall back to the config values, so passing any of them alongside server_config raises even when the value equals the default. No literal tuples left.

torchrl/modules/inference_server/_server.py
vmoens: Done for all three classes (InferenceServer, ProcessInferenceServer, AsyncBatchedCollector): None-defaults act as the sentinel and the defaults live in one place (InferenceServerConfig).

torchrl/modules/inference_server/_server.py
vmoens: Servers now reject storing_device with a clear error pointing at the collector, and the env_device-as-output_device fallback is documented on both server classes and the docs page.

test/test_inference_server.py
vmoens: Strengthened: the config test now asserts the resolved server devices/settings, server_output_device() has a unit test, and a gpu-marked CUDA test verifies the policy actually ran on cuda and the result landed on cpu.

torchrl/modules/inference_server/_config.py
vmoens: Replaced with a runnable example that builds an InferenceServer with the config, serves a request through a client, and checks the resolved settings.

torchrl/modules/inference_server/_config.py
vmoens: Deferred to #3943, which proposes exactly this: promote the config to torchrl.collectors (possibly as DevicePlacementConfig), accept device_config on regular collectors/evaluators, and soft-deprecate the explicit device kwargs on the standard two-release cadence.

torchrl/modules/inference_server/_config.py
vmoens: Deferred to #3943: the interplay with collector/evaluator device kwargs and the API-bloat concern are the core of that RFC. The position taken there is that config objects only land library-wide if they replace boilerplate through a single shared resolution function, not as a second parallel convention.

torchrl/modules/inference_server/_config.py
vmoens: Agreed -- tracked in #3943 with the integration and deprecation story (scope decision, multi-collector list semantics, Hydra rule-14 parity). This PR keeps the config objects server-side only so nothing here prejudges that design.

PR #3894, opened 2026-06-21
[Feature] Add remote policy module for inference server clients

vmoens:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #3942
* #3940
* #3941
* #3897
* #3896
* #3895
* __->__ #3894
* #3893
* #3939

PR #3895, opened 2026-06-21
[Feature] Track behavior policy versions in inference server

vmoens:
* #3942
* #3940
* #3941
* #3897
* #3896
* __->__ #3895
* #3894
* #3893
* #3939


torchrl/modules/inference_server/_server.py
theap06, line 376: This only increments when ws.receive() returns an non-None value, but for SharedMemWeightSyncScheme receive() always returns None; weights are applied by a background thread. instead, the weight should pull from the actual weight application.

torchrl/collectors/_async_batched.py
```diff
@@ -458,7 +470,10 @@ def _ensure_started(self) -> None:
         # inherited by the child process.
         if self._clients is None:
             self._clients = [
-                PolicyClientModule(self._transport.client())
+                PolicyClientModule(
+                    self._transport.client(),
+                    policy_version_key=self._policy_version_key or "policy_version",
```
theap06: The collector forwards policy_version_key but never target_policy_version/max_policy_lag, so _check_policy_lag always early-returns. I'd either give the client a live current-version source and expose max_policy_lag on the collector, or drop the staleness check.

theap06 (changes requested):

torchrl/modules/inference_server/_server.py
vmoens: Fixed: the server now installs itself as the scheme's cascade context so the version is bumped at actual weight application (update_policy_weights_), and it no longer polls receive() for shared-memory schemes (their background receiver applies weights and cascades).

torchrl/collectors/_async_batched.py
vmoens: Resolved by removing the client-side check entirely: raising in the consumer was the wrong place. Version stamping stays; bounded-staleness enforcement moved to the data path as PolicyAgeFilter, a replay-buffer transform that drops too-old elements on extend or dynamically on sample.

PR #3896, opened 2026-06-21
[Feature] Add process inference server control plane

vmoens:
* #3942
* #3940
* #3941
* #3897
* __->__ #3896
* #3895
* #3894
* #3893
* #3939


torchrl/modules/inference_server/_server.py
```diff
@@ -466,11 +469,36 @@ def _process_server_entry(
             shutdown_event=shutdown_event,
             **server_kwargs,
         )
+        server.start()
```
theap06, line 534: The child used to call server._run() on its main thread, so an error happens in the inference loop, propagated to the except BaseException and crashed the child.

torchrl/modules/inference_server/_server.py
```diff
@@ -637,8 +670,44 @@ def start(self) -> ProcessInferenceServer:
...
+    ):
+        if self._process is None:
+            raise RuntimeError("ProcessInferenceServer is not running.")
+        if not self._process.is_alive():
+            raise RuntimeError(
+                "ProcessInferenceServer process is not alive "
+                f"(exitcode={self._process.exitcode})."
+            )
+        request_id = self._next_control_request_id
+        self._next_control_request_id += 1
```
theap06: _request_control shares one response queue and drops any response whose id doesn't matchwith no lock, and _next_control_request_id += 1 is an unguarded read-modify-write. So if a monitoring thread calls stats()/health() while another control call (or shutdown()) runs, each call consumes and discards the other's response. the easy solution is to add a lock around the request/response cycle.

theap06 (changes requested):

torchrl/modules/inference_server/_server.py
vmoens: Addressed: the control loop now monitors the serve thread and exits the child with an error when it dies (after the serve loop's own shutdown path drained and rejected pending requests), so the parent sees a dead process instead of a healthy control plane fronting a dead server.

torchrl/modules/inference_server/_server.py
vmoens: Done: the whole request/response cycle is serialized with a lock, and the response wait converts queue.Empty into TimeoutError. Mismatched ids can then only be stale replies from timed-out requests and are safely discarded.

PR #3897, opened 2026-06-21
[Feature] Add max-inflight guard for remote policy clients

vmoens:
* #3942
* #3940
* #3941
* __->__ #3897
* #3896
* #3895
* #3894
* #3893
* #3939


theap06 (changes requested):
Makes sense overall. my one gripe is the result() call. Releasing the slot only when result() is called works for the synchronous collector path, but it leaks a slot if a future gets dropped, and a result() timeout frees the slot while the request is still running. I'd just release on the future completing add_done_callback instead.

also, for the __getattr__ function, since it forwards to self._future, it'll infinite-loop if _future ever isn't set yet.  Easy fix is to forego the private names.

PR #3898, opened 2026-06-21
[Refactor] LossModule: extract make_value_estimator preamble

theap06:
## Summary
- Refactors `make_value_estimator` across 5 loss modules to use a new
  `LossModule._prepare_value_estimator_kwargs` helper, eliminating ~55
  lines of duplicated preamble and establishing a reusable pattern for
  future losses.


theap06:
drafting a follow up pr that reduces boilerplate when writing new loss modules. will add a register_coeff_buffer helper and a default _forward_value_estimator_keys so you dont have to copy paste setup code from existing losses every time you write a new one

vmoens (approved):
LGTM thanks!

PR #3901, opened 2026-06-22
[Feature] Feat/loss module helpers

theap06:
## Summary

This PR adds reusable `LossModule` helpers for boilerplate that appears across several loss modules, and migrates a small set of callsites to use them.

### New helpers

- `LossModule.register_coeff_buffer(...)`
  - registers scalar coefficient buffers with consistent tensor conversion, device/dtype handling, optional-`None` behavior, and scalar validation.
  - replaces repeated `torch.tensor(...)` / `register_buffer(...)` / `None` handling for coefficient-like fields.
- A default `LossModule._forward_value_estimator_keys(...)`
  - forwards common value-estimator keys when they are present on `tensor_keys`.
  - refreshes `in_keys` via `_set_in_keys()` when available.

### Migrated callsites

- `A2CLoss`
  - `entropy_coeff`, `critic_coeff`, and `clip_value` now use `register_coeff_buffer`.
- `PPOLoss`
  - scalar `entropy_coeff`, optional `critic_coeff`, and `clip_value` now use `register_coeff_buffer`.
  - per-head entropy coefficient mappings keep their existing mapping behavior while using the helper for the compatibility scalar buffer.
- `IQLLoss`
  - relies on the default `_forward_value_estimator_keys` implementation instead of carrying an identical local override.

Losses with non-standard value-estimator key remapping, estimator-specific keys, or other custom behavior keep their explicit overrides.

## Tests

- Added regression coverage for `register_coeff_buffer`, including `None`, dtype preservation, non-scalar rejection, and bool rejection.
- Added regression coverage for the default `_forward_value_estimator_keys` behavior and `set_keys` propagation.


vmoens (approved):
LGTM thanks!

PR #3306, opened 2026-01-08
[Feature] Add _Contiguous module and reshape improvements to encoders/decoders

vmoens:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #3940
* #3909
* #3897
* #3896
* #3895
* #3894
* #3893
* #3939
* __->__ #3306

----

- Add _Contiguous helper module for torch.compile inductor compatibility
- Refactor ObsEncoder.forward and ObsDecoder.forward to use flatten/unflatten with contiguous()
- Add _maybe_record_function_decorator for profiling

PR #3909, opened 2026-06-24
[Feature] Add async VLA GRPO collection modes

vmoens:
* #3940
* __->__ #3909
* #3897
* #3896
* #3895
* #3894
* #3893
* #3939
* #3306

### DIFF
--- a/sota-implementations/vla_grpo/test_openvla.py
+++ b/sota-implementations/vla_grpo/test_openvla.py
@@ -356,6 +356,187 @@ def test_make_collector_can_write_to_replay_buffer(monkeypatch):
     assert captured["kwargs"]["frames_per_batch"] == 4
 
 
+def test_make_collector_async_env_uses_async_batched_collector(monkeypatch):
+    captured = {}
+
+    class _FakeAsyncCollector:
+        def __init__(self, *args, **kwargs):
+            captured["args"] = args
+            captured["kwargs"] = kwargs
+
+        def __iter__(self):
+            return self
+
+        def __next__(self):
+            raise StopIteration
+
+        def server_stats(self, *, reset=False):
+            return {"requests": 0}
+
+        def shutdown(self):
+            captured["shutdown"] = True
+
+    class _FakeEnv:
+        batch_size = torch.Size([1])
+        device = torch.device("cpu")
+
+    cfg = SimpleNamespace(
+        collector=SimpleNamespace(
+            groups_per_iter=4,
+            group_size=2,
+            async_env=True,
+            async_policy=True,
+            server_min_batch_size=2,
+        ),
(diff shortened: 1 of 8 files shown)


theap06 (changes requested):
LGTM left a few nitpicking comments regarding design

sota-implementations/vla_grpo/utils.py
```diff
@@ -61,6 +67,13 @@
 LOG_PROBS_KEY = ("vla_action", "log_probs")
 
 
+def _cfg_get(section, key: str, default=None):
```
theap06, line 70: The new cfg_get function is implemented here; however, re-inlines the old getattr(cfg.collector, "get", lambda ...) pattern three more times. Pick one. I'd say utilize the legacy function for maintainability

sota-implementations/vla_grpo/utils.py
```diff
@@ -247,6 +260,146 @@ def _make_libero_worker(
     )
 
 
+def _num_envs_from_cfg(cfg, *, eval_mode: bool = False) -> int:
```
theap06, line 263: on the num_envs_from_cfg function default = cfg.env.num_envs is read and then handed to _cfg_get(cfg.env, "num_envs", default), so the default is the same attribute being looked up, and a config missing num_envs raises AttributeError before the default can help. Harmless if the other configs always set it; still dead logic.

theap06 (approved):

sota-implementations/vla_grpo/test_openvla.py
vmoens, line 1: we need to use Test classes here

vmoens:
Superseded: the async collection modes were reworked and absorbed into #3940, with the interaction-type propagation split out into #3941, so this intermediate implementation is no longer part of the stack.

sota-implementations/vla_grpo/utils.py
vmoens: Done in the successor commit on #3940: _cfg_get is the single helper and the inlined getattr patterns were replaced with it.

sota-implementations/vla_grpo/utils.py
vmoens: Reworked in the successor commit on #3940: _num_envs_from_cfg now goes through _cfg_get with a literal default.

sota-implementations/vla_grpo/test_openvla.py
vmoens: Done in the successor commit on #3940: the module-level tests are grouped into Test classes (TestAdvantageMetrics, TestRealCheckpoint, TestCollectorFactory, TestEvaluatorFactory, TestReplayBufferFactory, TestTokenizerAndTransforms, TestLiberoWorkers).

PR #3922, opened 2026-06-30
[Feature] Add RewardModelLoss objective for RLHF reward-model training

theap06:
## What

Adds a model-agnostic Bradley-Terry pairwise **reward-model loss** as a first-class objective under `torchrl/objectives/llm/`, following the `SFTLoss` / `GRPOLoss` conventions (`_AcceptedKeys` / `set_keys`, a `TensorClass` output, runnable docstring with paper references).

## Why

The Bradley-Terry reward loss already existed as `GPT2RewardModel.compute_reward_loss` (`torchrl/modules/models/llm.py`), explicitly tagged `# TODO: move to objectives` and tied to GPT2. This promotes it to a proper objective that:

- lives in `objectives/` like every other loss;
- is **model-agnostic** — it owns a swappable `score_network` (any `TensorDictModule` producing a per-sequence scalar, e.g. `AutoModelForSequenceClassification(num_labels=1)` or `GPT2RewardModel`);
- is TensorDict-first with configurable keys.

## Contents

- `torchrl/objectives/llm/reward.py`: `RewardModelLoss`, `RewardModelLossOutput`, and the `reward_model_loss` helper. Supports `mean`/`sum`/`none` reduction, an optional score-centering regularizer, and a detached `accuracy` metric for logging.
- Exports from `torchrl/objectives/llm/__init__.py`.
- Docs: "Reward Model Training" sections in `llms_objectives.rst` and `llms.rst`.
- Tests: `TestRewardModel` in `test/llm/test_llm_objectives.py` (CPU-only, includes a nested-key input case per the contributor guide).

## Tests

`pytest test/llm/test_llm_objectives.py -k RewardModel` → 9 passed. `ruff` + `black` clean; the public-API `__all__` guard passes.

### DIFF
--- a/torchrl/objectives/llm/reward.py
+++ b/torchrl/objectives/llm/reward.py
@@ -111,8 +111,10 @@ class RewardModelLoss(LossModule):
             ``center_coeff * (chosen_score**2 + rejected_score**2)`` that discourages
             the reward model from drifting to large magnitudes. Defaults to ``None``
             (disabled).
-        device (torch.device, optional): the device on which to run ``score_network``.
-            Defaults to ``None``.
+        device (torch.device or str, optional): if provided, chosen/rejected
+            sub-tensordicts are moved to this device before scoring and loss
+            computation. The ``score_network`` is expected to already be on this
+            device. Defaults to ``None``.
 
     .. note::
         The input tensordict is expected to contain the following keys by default:
@@ -202,13 +204,13 @@ class RewardModelLoss(LossModule):
         *,
         reduction: Literal["mean", "sum", "none"] = "mean",
         center_coeff: float | None = None,
-        device: torch.device | None = None,
+        device: torch.device | str | None = None,
     ) -> None:
         super().__init__()
         self.score_network = score_network
         self.reduction = reduction
         self.center_coeff = center_coeff
-        self.device = device
+        self.device = torch.device(device) if device is not None else None
         self._set_in_keys()
 
     def _set_in_keys(self) -> None:
@@ -224,6 +226,8 @@ class RewardModelLoss(LossModule):
                 f"Could not find the sub-tensordict at key {key!r} in the input "
                 f"tensordict with keys {set(tensordict.keys())}."
             )
+        if self.device is not None:
+            sub_td = sub_td.to(self.device)
         if self.score_network is not None:
             with (
                 torch.device(self.device)
--- a/torchrl/objectives/llm/reward.py
+++ b/torchrl/objectives/llm/reward.py
@@ -1,0 +1,272 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import annotations
+
+import contextlib
+
+from dataclasses import dataclass
+from typing import Literal
+
+import torch
+from tensordict import NestedKey, TensorClass, TensorDictBase
+from tensordict.nn import TensorDictModule
+from torchrl.objectives.common import LossModule
+
+
+def reward_model_loss(
+    chosen_scores: torch.Tensor,
+    rejected_scores: torch.Tensor,
+    reduction: Literal["mean", "sum", "none"],
+) -> torch.Tensor:
+    r"""Compute the Bradley-Terry pairwise reward-model loss.
+
+    The loss is computed as ``-log_sigmoid(chosen_scores - rejected_scores)``. It is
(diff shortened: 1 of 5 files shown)


torchrl/objectives/llm/reward.py
```diff
@@ -0,0 +1,276 @@
...
+                if self.device is not None
+                else contextlib.nullcontext()
+            ):
+                sub_td = self.score_network(sub_td)
+        score = sub_td.get(self.tensor_keys.score, default=None)
+        if score is None:
+            raise KeyError(
+                f"Could not find the score at key {self.tensor_keys.score!r} under "
+                f"{key!r}. If score_network is None, the scores must be precomputed."
+            )
```
vmoens, line 263: after set_keys(score=("metrics", "score")), _score invokes an unchanged score_network and then reads only the customized key. A network still writing its declared/default "score" key breaks. The nested-key test covers only score_network=None.

torchrl/objectives/llm/reward.py
```diff
@@ -0,0 +1,276 @@
...
+        *,
+        reduction: Literal["mean", "sum", "none"] = "mean",
+        center_coeff: float | None = None,
+        device: torch.device | str | None = None,
+    ) -> None:
+        super().__init__()
+        self.score_network = score_network
+        self.reduction = reduction
+        self.center_coeff = center_coeff
+        self.device = torch.device(device) if device is not None else None
```
vmoens: self.device is a plain attribute, so loss.to(...) does not keep it synchronized

In general we should avoid to have modules that have a self.device attribute.

torchrl/objectives/llm/reward.py
```diff
@@ -0,0 +1,276 @@
...
+            )
+        if self.device is not None:
+            sub_td = sub_td.to(self.device)
+        if self.score_network is not None:
+            with (
+                torch.device(self.device)
+                if self.device is not None
+                else contextlib.nullcontext()
+            ):
+                sub_td = self.score_network(sub_td)
```
vmoens: ditto, self.device should not really be a thing IMO. Let's think about models that have more than one device for example.

vmoens (commented):
Two additional fixes are needed beyond those that I posted:
- invalid reduction values are rejected only after both score-network passes
- the example scorer averages padding tokens and does not demonstrate the advertised attention_mask contract

Thanks for all the contributions!

theap06:
@vmoens thanks for the feedback! Just addressed it. 
image

torchrl/objectives/llm/reward.py
```diff
@@ -0,0 +1,294 @@
...
+    def _score(self, tensordict: TensorDictBase, key: NestedKey) -> torch.Tensor:
+        """Run the score network on a sub-tensordict and return a per-sequence score."""
+        sub_td = tensordict.get(key, default=None)
+        if sub_td is None:
+            raise KeyError(
+                f"Could not find the sub-tensordict at key {key!r} in the input "
+                f"tensordict with keys {set(tensordict.keys())}."
+            )
+        if self.score_network is not None:
+            sub_td = self.score_network(sub_td)
```
gtnv: this scorer writes score into the caller-owned chosen and rejected TensorDicts. can we run it on a shallow structural clone?

```suggestion
            sub_td = self.score_network(sub_td.clone(False))
```

torchrl/objectives/llm/reward.py
```diff
@@ -0,0 +1,294 @@
...
+        score_network: TensorDictModule | None = None,
+        *,
+        reduction: Literal["mean", "sum", "none"] = "mean",
+        center_coeff: float | None = None,
+    ) -> None:
+        _validate_reduction(reduction)
+        super().__init__()
+        self.score_network = score_network
+        self.reduction = reduction
+        self.center_coeff = center_coeff
```
gtnv, line 226: center_coeff=-1.0 is accepted rn. that makes the squared-score term lower the objective. can we reject negative values?

```suggestion
        if center_coeff is not None and center_coeff < 0:
            raise ValueError("center_coeff must be non-negative.")
        self.center_coeff = center_coeff
```

torchrl/objectives/llm/reward.py
```diff
@@ -0,0 +1,305 @@
...
+
+_REDUCTIONS = ("mean", "sum", "none")
+
+
+def _validate_reduction(reduction: str) -> None:
+    if reduction not in _REDUCTIONS:
+        raise ValueError(f"Invalid reduction: {reduction}.")
+
+
+def reward_model_loss(
```
vmoens: We should bundle all the functional like this one in a functionals.py file

PR #3923, opened 2026-06-30
[Algorithm] Add RLHF reward-model training recipe

theap06:
Stacked on #3922 (base branch `theap06/reward-model-loss`): merge #3922 first, then retarget this PR to `main` (`gh pr edit 3923 --base main`) and merge.

## What

Adds a dedicated `sota-implementations/reward_model_training/` recipe that trains a scalar reward model from pairwise human-preference data using `RewardModelLoss`. This modernizes the legacy `examples/rlhf/train_reward.py` (originally from #1597) into the current `sota-implementations` + LLM-API style, and is model-agnostic: any HF `AutoModelForSequenceClassification` backbone works.

## Stacked on #3922

This PR depends on #3922 (the `RewardModelLoss` objective) and is **stacked on top of it** — the recipe imports `RewardModelLoss` and the CI smoke test exercises it end to end. Please merge #3922 first; once it lands, the diff here reduces to just the recipe + CI files (the first commit shown belongs to #3922).

## Contents

- `sota-implementations/reward_model_training/`: `reward_model.py` (Hydra main), `utils.py`, `config.yaml`, `requirements.txt`, `README.md`.
- **Hermetic CI path**: an empty `model.name` builds a tiny from-config model and an empty `data.dataset_name` generates synthetic preference pairs, so the smoke test needs no download and no HF `datasets` dependency (which the `linux_sota` env does not provide).
- CI: a `reward_model_training` smoke entry in `.github/unittest/linux_sota/scripts/test_sota.py`.
- `sota-check/run_reward_model_training.sh` for full release runs.

## Tests

Ran the exact CI smoke command locally: builds a tiny model, trains on synthetic pairs, evaluates (val loss + accuracy), and checkpoints — exit 0. `ruff` + `black` clean.

Real usage: `python reward_model.py model.name=gpt2 data.dataset_name=CarperAI/openai_summarize_comparisons`.

### DIFF
--- a/docs/source/reference/llms.rst
+++ b/docs/source/reference/llms.rst
@@ -544,6 +544,18 @@ SFT
     SFTLoss
     SFTLossOutput
 
+Reward Model Training
+~~~~~~~~~~~~~~~~~~~~~
+
+.. currentmodule:: torchrl.objectives.llm
+
+.. autosummary::
+    :toctree: generated/
+    :template: rl_template.rst
+
+    RewardModelLoss
+    RewardModelLossOutput
+
 .. currentmodule:: torchrl.data.llm
 
 .. autosummary::
--- a/sota-implementations/reward_model_training/reward_model.py
+++ b/sota-implementations/reward_model_training/reward_model.py
@@ -1,0 +1,159 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""RLHF reward-model training recipe.
+
+Trains a scalar reward model from pairwise human-preference data using the
+Bradley-Terry objective :class:`~torchrl.objectives.llm.RewardModelLoss`. This is the
+reward-modelling stage that precedes policy optimization (e.g. GRPO/PPO) in RLHF
+pipelines.
+
+The backbone is any Hugging Face ``AutoModelForSequenceClassification`` with a
+single-output head. The helper functions live in the ``utils.py`` next to this script.
+
+Examples:
+    Train on the default summarization-comparisons dataset with a GPT2 backbone::
+
+        python reward_model.py model.name=gpt2
+
+    Run a quick hermetic smoke test (tiny from-scratch model, synthetic data)::
+
+        python reward_model.py model.name= data.dataset_name= \\
+            optim.max_iters=3 logger.backend=
+"""
+from __future__ import annotations
+
+import hydra
+import torch
+import tqdm
+from torchrl._utils import get_available_device, logger as torchrl_logger, timeit
+from torchrl.objectives.llm import RewardModelLoss
+from torchrl.record.loggers import generate_exp_name, get_logger
+from utils import (
+    get_vocab_size,
+    log_metrics,
+    make_dataset,
+    make_optimizer,
+    make_replay_buffer,
+    make_reward_model,
+    make_tokenizer,
+)
+
+
+@hydra.main(config_path="", config_name="config", version_base="1.1")
+def main(cfg):  # noqa: F821
+    # Logger
+    exp_name = generate_exp_name("RewardModel", cfg.logger.exp_name)
+    logger = None
+    if cfg.logger.backend:
+        logger = get_logger(
+            logger_type=cfg.logger.backend,
+            logger_name="reward_model_logging",
+            experiment_name=exp_name,
+            wandb_kwargs={
+                "mode": cfg.logger.mode,
+                "config": dict(cfg),
+                "project": cfg.logger.project_name,
+                "group": cfg.logger.group_name,
+            },
+        )
+
+    torch.manual_seed(int(cfg.seed))
+    device = (
+        torch.device(cfg.model.device) if cfg.model.device else get_available_device()
+    )
+
+    # Model + data
+    with timeit("setup/model"):
+        score_network = make_reward_model(cfg, device)
+    tokenizer = make_tokenizer(cfg)
+    vocab_size = get_vocab_size(tokenizer, score_network)
+
+    with timeit("setup/data"):
+        train_data = make_dataset(cfg, tokenizer, cfg.data.split_train, vocab_size)
+        val_data = make_dataset(cfg, tokenizer, cfg.data.split_val, vocab_size)
+        train_rb = make_replay_buffer(train_data, cfg.data.batch_size, device)
+        val_rb = make_replay_buffer(val_data, cfg.data.batch_size, device)
+
+    # Loss + optimizer
+    loss_module = RewardModelLoss(
+        score_network,
+        reduction=cfg.loss.reduction,
+        center_coeff=cfg.loss.center_coeff,
+        device=device,
+    )
+    optimizer = make_optimizer(cfg, score_network)
+
+    n_trainable = sum(p.numel() for p in score_network.parameters() if p.requires_grad)
+    torchrl_logger.info(f"Reward model with {n_trainable} trainable parameters.")
+
+    @torch.no_grad()
+    def evaluate() -> tuple[float, float]:
+        score_network.eval()
+        losses, accuracies = [], []
+        for _ in range(int(cfg.logger.eval_iters)):
+            batch = val_rb.sample()
+            out = loss_module(batch)
+            losses.append(out.loss_reward_model)
+            accuracies.append(out.accuracy)
+        score_network.train()
+        return (
+            torch.stack(losses).mean().item(),
+            torch.stack(accuracies).mean().item(),
+        )
+
+    clip_grad = cfg.optim.clip_grad
+    max_iters = int(cfg.optim.max_iters)
+    pbar = tqdm.tqdm(range(1, max_iters + 1))
+    for it in pbar:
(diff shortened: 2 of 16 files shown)


vmoens (commented):
Thanks for putting this together. I think the `RewardModelLoss` itself is a useful TorchRL primitive, good pieces to have in-tree

My main concern is the positioning of the recipe. Moving forward I'd like to be foreseeing on how we can bring torchrl to a state where we can do serious distributed training for LLM features. As written, this is a single-process training loop, with GPT-2 as the visible example. Fine for a tuto/smoke recipe, but it is not really a serious reward-model training stack for modern LLMs. I guess that, at scale, users will still want TRL/Accelerate/FSDP, TorchTitan/FSDP2, Megatron/NeMo, OpenRLHF, etc. for parallelism, sharded checkpointing, fault tolerance, packed/streamed datasets, mixed precision, activation checkpointing, and multi-node orchestration. So the end goal should be to interface with these.

Don't get me wrong, I do **not** think this PR should grow a Megatron/NeMo/TorchTitan integration. That would be too much scope. But I think we should be very explicit about what TorchRL is providing here:

- a canonical TensorDict preference-data format,
- a canonical reward-model loss,
- a small reference trainer,
- and eventually adapters/scorers that let externally trained reward models plug back into TorchRL GRPO/PPO/SFT pipelines.

Concretely, I would suggest adding a short README section along the lines of “Scaling and integration”. It could say that this recipe is a minimal single-node baseline, and that large-scale RM training should use an external backend while preserving the TorchRL data/scoring contract. Essentially we need to pave the way for the real stuff that is to come. If you're happy to work on this with me we can start a topic in discussion.

I would also avoid presenting GPT-2 as the “real” default. Keeping the tiny synthetic model for CI is great, but the docs/examples should probably use a small Qwen model.

So overall: I like the objective and the idea of a reference recipe. I would just make the scope honest: TorchRL should own the data contract, loss semantics, and scoring interface. Overall, we need to think about how to scale this (and the rest of the LLM stack) up.

theap06:
@vmoens Thanks, this framing makes more sense! On the bigger picture, I fully agree the end goal is the adapters/scorers layer and interfacing with external backends. The way I'm picturing it: TensorDict is the shared format that holds the preference data, rollouts, scores, and rewards, and the backends (TRL/Accelerate/FSDP, etc.) bolt on at the edges. They consume the exported preference data, do the heavy distributed training, and hand back a checkpoint we re-import as a scorer. That way GRPO/SFT/reward all just read and write the same keys, and the parallelism can be done with the backends.

theap06:
Addressed the positioning feedback: the README now defines this as a minimal single-node reference, documents the external scaling/integration boundary, and uses Qwen2.5-0.5B as the visible default while retaining the tiny model for CI.

sota-implementations/reward_model_training/reward_model.py
```diff
@@ -0,0 +1,159 @@
...
+        with timeit("train/optim_step"):
+            optimizer.step()
+
+        metrics_to_log = {
+            "train/loss": loss.item(),
+            "train/accuracy": loss_out.accuracy.item(),
+            "train/grad_norm": grad_norm.item(),
+        }
+        pbar.set_postfix(
+            loss=f"{loss.item():.4f}", acc=f"{loss_out.accuracy.item():.3f}"
```
vmoens: the config permits loss.reduction=none, but the loop calls `loss.backward()`, `loss.item()`, and scalar formatting. A batch-shaped loss fails. Reject or normalize this mode.

sota-implementations/reward_model_training/utils.py
```diff
@@ -0,0 +1,226 @@
...
+def make_replay_buffer(
+    data: TensorDict, batch_size: int, device: torch.device
+) -> TensorDictReplayBuffer:
+    rb = TensorDictReplayBuffer(
+        storage=LazyTensorStorage(data.shape[0], device=device),
+        sampler=SamplerWithoutReplacement(drop_last=True),
+        batch_size=batch_size,
+    )
+    rb.extend(data)
+    return rb
```
vmoens: the entire padded/tokenized dataset is copied into LazyTensorStorage(..., device=device), making accelerator memory scale with dataset size rather than batch size. Keep storage on CPU or stream batches.

(also why is a batch passed in the helper? For init of the lazy storage? Then it's not lazy anymore)

sota-implementations/reward_model_training/utils.py
```diff
@@ -0,0 +1,226 @@
...
+        # Hermetic synthetic dataset (no download, no ``datasets`` dependency).
+        n = int(cfg.data.synthetic_size)
+        gen = torch.Generator().manual_seed(int(cfg.seed) + (split == "train"))
+        chosen_ids = torch.randint(0, vocab_size, (n, max_length), generator=gen)
+        rejected_ids = torch.randint(0, vocab_size, (n, max_length), generator=gen)
+        return _pairwise_td(chosen_ids, rejected_ids)
+
+    # Real preference data. ``datasets`` is imported lazily so the synthetic/CI path
+    # never requires it.
+    from datasets import load_dataset
```
vmoens, line 199: the optional datasets import is function-local and lacks the required module-level availability guard

sota-implementations/reward_model_training/reward_model.py
```diff
@@ -0,0 +1,159 @@
...
+
+        if it % int(cfg.logger.eval_iter) == 0:
+            val_loss, val_acc = evaluate()
+            metrics_to_log["eval/loss"] = val_loss
+            metrics_to_log["eval/accuracy"] = val_acc
+            torchrl_logger.info(
+                f"iter {it}: val_loss={val_loss:.4f}, val_acc={val_acc:.4f}"
+            )
+
+        if it % int(cfg.checkpoint.save_iter) == 0:
```
vmoens: repeated save_pretrained calls overwrite one directory and the tokenizer is not persisted.

sota-implementations/reward_model_training/reward_model.py
theap06: Addressed in e508d6efa0. The recipe now validates the reduction before logger, model, or data setup and accepts only `mean` or `sum`, so `none` fails immediately with a targeted error.

sota-implementations/reward_model_training/utils.py
theap06: Addressed in e508d6efa0. The dataset now backs a pre-populated CPU `TensorStorage` directly (no eager `LazyTensorStorage` allocation/extend and no accelerator-sized dataset copy). Only each sampled train/eval batch is transferred to the model device.

sota-implementations/reward_model_training/utils.py
theap06: Addressed in e508d6efa0. `datasets` availability is now checked at module scope with `importlib.util.find_spec`; the guarded lazy import remains limited to the real-dataset path and raises a targeted installation error when unavailable.

sota-implementations/reward_model_training/reward_model.py
theap06: Addressed in e508d6efa0. Periodic saves now use distinct `checkpoint-<step>` directories, the final save uses `checkpoint-final`, and the matching tokenizer is saved alongside the model whenever one is present. The layout is documented in the README/config.

docs/source/reference/llms.rst
```diff
@@ -553,6 +553,7 @@ Reward Model Training
     :toctree: generated/
     :template: rl_template.rst
 
+    reward_model_loss
```
vmoens, line 556: @theap06 made this public

I recently created a `functional` in `torchrl.modules` and I wonder if we shouldn't generalize this across the lib - free function go in `functional` modules and are used by classes to implement features. Good separation of concerns IMO (I'm thinking `torchrl.objectives.functional` for instance)

vmoens (approved):
LGTM thanks!

sota-implementations/reward_model_training/reward_model.py
```diff
@@ -0,0 +1,191 @@
...
+    save_dir: str | Path,
+    step: int | Literal["final"],
+) -> Path:
+    checkpoint_name = (
+        f"checkpoint-{step:08d}" if isinstance(step, int) else "checkpoint-final"
+    )
+    checkpoint_dir = Path(save_dir) / checkpoint_name
+    score_network.module.model.save_pretrained(checkpoint_dir)
+    if tokenizer is not None:
+        tokenizer.save_pretrained(checkpoint_dir)
```
gtnv: checkpoint-* saves only model and tokenizer weights. it cannot resume optimizer state, step/data position, or rng. can we call these exports or save the full state?

sota-implementations/reward_model_training/utils.py
```diff
@@ -0,0 +1,233 @@
...
+        rejected_texts.append(prompt + sep + sample["rejected"])
+
+    tok_kwargs = {
+        "max_length": max_length,
+        "padding": "max_length",
+        "truncation": True,
+        "return_tensors": "pt",
+    }
+    chosen_tok = tokenizer(chosen_texts, **tok_kwargs)
+    rejected_tok = tokenizer(rejected_texts, **tok_kwargs)
```
gtnv: this builds both full string lists first. it then materializes both padded token tensors. can we tokenize lazily or in bounded batches?

PR #3925, opened 2026-07-01
[BugFix] Fix double-alpha transform in PrioritizedSampler

theap06:
## Summary 
- Prioritized replay stores priorities in the sum-tree, and update_priority is the single place that transform is applied, so every caller is meant to pass a raw priority.
- default_priority returns the raw max, so the single transform in update_priority produces the correct value


vmoens (approved):
LGTM thanks!

PR #3926, opened 2026-07-02
[Feature] Add Learner primitive (LocalLearner, FSDP2Learner)

theap06:
## Summary

Introduces `torchrl.trainers.Learner`: a backend-agnostic entry point for taking
one optimization step on a tensordict batch with a given `LossModule`.
`LocalLearner` is the single-process reference implementation; `FSDP2Learner`
shards the same model with `fully_shard` and reuses the training step unchanged.

It plays the same role for training that `Collector` plays for data collection
and `LLMWrapperBase` plays for generation/scoring: a fixed, TensorDict-native
contract with interchangeable backends, so algorithm code does not need to know
whether the update runs on one device, under sharded training, or on a remote
training process.

## Design

- `Learner.update()` is concrete, in the base class, and touches only
  `self.model` / `self.optimizer` / `self.clip_grad_norm` /
  `self.grad_accum_steps`: zero_grad -> forward -> sum the loss module's
  `"loss"`-prefixed output keys -> backward -> optional grad-norm clip ->
  optimizer step. This is what lets `FSDP2Learner` reuse the exact same step as
  `LocalLearner` -- sharded training only changes model construction and weight
  gathering, not the step itself.
- `get_weights()` is the one place sharding is *not* transparent: it must return
  plain tensors (for `WeightSyncScheme.send`, which already accepts a
  `TensorDictBase`), even when the learner's parameters are sharded.
- `FSDP2Learner` does not decide sharding granularity or device mesh -- it
  accepts a model the caller has already wrapped with `fully_shard`, exactly as
  bare FSDP2 usage works. Keeping that decision in caller code avoids
  `FSDP2Learner` becoming a second place those choices are made.

## Two contracts worth reading before using this

**The optimizer defines what is trained, not `model`.** `model` is the
weight-sync source and the gradient-sync handle; the parameters that get clipped
and stepped are the ones in `optimizer.param_groups`. Many TorchRL losses hold
their trainable parameters on the loss module as `TensorDictParams`, and the
losses that expand their networks (`SACLoss`, `REDQLoss`, ...) hold *copies* of
the modules you passed in:

```python
loss_module = SACLoss(actor, qvalue)
learner = LocalLearner(actor, Adam(loss_module.parameters()))  # correct
learner = LocalLearner(actor, Adam(actor.parameters()))        # critics never train
```

The second line is silent -- the critics are differentiated on every step and
never updated -- so `update()` now checks before its first optimizer step and
raises if any parameter received a gradient that no param group covers. Grad-norm
clipping likewise covers `optimizer.param_groups`, not `model.parameters()`.

A `Learner` owns exactly one optimizer, so algorithms that deliberately use
several (per-network learning rates, a separate entropy-temperature optimizer)
are not expressible as a single `Learner` today; use one per optimizer.

**Checkpointing is `checkpoint()` / `load_checkpoint()`, not `state_dict()`.** A
bare `Optimizer` is not an `nn.Module`, so `nn.Module.state_dict()` structurally
cannot carry its state and a resume would reset Adam's moments. Overriding
`state_dict` to return it anyway would break the `nn.Module` contract -- a parent
module calls `child.state_dict(destination=...)` and discards the return value,
so nesting a `Learner` inside any other module would silently drop all of its
state. Separate names keep both contracts intact.

Gradients are not checkpointed, so a checkpoint taken mid-accumulation-window
resets the accumulation counter to 0 with a warning rather than resuming at a
non-zero step with empty gradients (which would step after fewer micro-batches
than `grad_accum_steps` and under-scale that update).

## Notes

- `update()` raises on a non-scalar summed loss, which is what a loss built with
  `reduction="none"` produces -- previously this surfaced as a torch-internal
  "grad can be implicitly created only for scalar outputs" error.
- The `"loss"` prefix cannot be tightened to `"loss_"`: `"loss"` with no
  underscore is a real out_key (`DQNLoss`, `GAILLoss`, `OnlineDTLoss`).
- With `FSDP2Learner.get_weights()`'s default `cpu_offload=True`, the full
  weights are returned **on rank 0 only** and every other rank gets an empty
  tensordict. That is deliberate (gathering the full model onto every rank does
  not scale), but it means an unconditional
  `scheme.send(learner.get_weights())` sends nothing from non-zero ranks. Pass
  `cpu_offload=False` to gather everywhere.
- With gradient accumulation the output key set differs between accumulation
  calls and step calls (`grad_norm` only appears on the latter).
- No `*Config` companion yet. These classes have no `Trainer` wiring in this PR,
  so a Hydra config would have nothing to instantiate against; the configs land
  with the trainer integration (follow-up 2 below).
- These are new, unreleased classes, so the `checkpoint()` rename carries no
  deprecation shim.

## Planned follow-ups (not in this PR)

1. Multi-rank `FSDP2Learner` verification on real multi-GPU hardware -- the one
   thing I could not test here. The single-rank gloo/CPU tests exercise the
   `fully_shard`/DTensor code paths but not actual cross-rank communication.
2. Refactor an existing recipe's hand-rolled training loop (e.g. the
   reward-model recipe, once #3922 lands) onto `LocalLearner`, as the first real
   consumer, with the `*Config` companions.
3. A `RemoteLearner` design writeup scoping one external backend (TorchTitan is
   the most PyTorch-native of the candidates and the most plausible first
   integration target) before any implementation.

## Tested

`test/test_trainer.py`: **100 passed, 19 skipped** (26 of those are the
`Learner`/`FSDP2Learner` tests). Single-rank gloo/CPU only; multi-GPU is
follow-up 1.

`c75e16b` fixes a checkpoint bug that only surfaced once the suite actually ran:
`get_state_dict(full_state_dict=True, cpu_offload=True)` does **not** guarantee
fresh tensors — `cpu_offload` only copies when the shards are off-CPU, so on a
CPU or single-rank mesh the "gathered" state aliases the live shards and training
on after a checkpoint silently mutates it. `FSDP2Learner.checkpoint` now clones,
like the base implementation.

### DIFF
--- a/torchrl/trainers/learners/common.py
+++ b/torchrl/trainers/learners/common.py
@@ -1,0 +1,93 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from tensordict import TensorDictBase
+from torch import nn
+
+from torchrl.objectives.common import LossModule
+
+
+@dataclass
+class LearnerCapabilities:
+    """Declares what a :class:`Learner` implementation supports.
+
+    Algorithm code is not expected to branch on these (that would defeat the
+    point of the abstraction); they exist for logging, validation, and for
+    orchestration code that needs to know, e.g., whether a learner can be
+    checkpointed independently on every rank.
+
+    Attributes:
+        sharded (bool): whether the learner's parameters are sharded across
+            multiple devices/processes (e.g. FSDP2). Defaults to ``False``.
+        remote (bool): whether :meth:`~Learner.update` dispatches to a
+            separate process rather than running in-line. Defaults to
+            ``False``.
+    """
+
+    sharded: bool = False
+    remote: bool = False
+
+
+class Learner(nn.Module):
+    """Base class for the trainable-policy role.
+
+    A :class:`Learner` owns a trainable model and exposes a single,
+    backend-agnostic entry point, :meth:`update`, for taking one optimization
+    step on a :class:`~tensordict.TensorDictBase` batch with a given
+    :class:`~torchrl.objectives.common.LossModule`. Algorithm code calls
+    ``learner.update(batch, loss_module)`` without knowing whether the update
+    runs locally on one device, under sharded (e.g. FSDP2) training, or on a
+    separate remote training process -- that placement is the ``Learner``
+    subclass's responsibility, not the algorithm's.
+
+    This mirrors the role :class:`~torchrl.collectors.Collector` plays for
+    data collection and :class:`~torchrl.modules.llm.LLMWrapperBase` plays for
+    generation/scoring: a fixed, TensorDict-native contract with multiple
+    interchangeable backends.
+
+    .. note::
+        ``update`` requires ``loss_module.forward`` to follow the
+        :class:`~torchrl.objectives.common.LossModule` convention: every
+        differentiable loss term is returned under a key starting with
+        ``"loss"`` (these are summed and used for the backward pass); any
+        other returned entry (e.g. ``accuracy``, a KL for logging) is treated
+        as a non-differentiable metric and left untouched.
+
+    .. seealso:: :class:`~torchrl.weight_update.WeightSyncScheme` consumes
+        :meth:`get_weights` to synchronize a learner's parameters to remote
+        inference workers, so a ``Learner`` composes with the existing
+        weight-sync machinery without changes on either side.
+    """
+
+    capabilities: LearnerCapabilities = LearnerCapabilities()
+
+    def update(self, batch: TensorDictBase, loss_module: LossModule) -> TensorDictBase:
+        """Take one optimization step on ``batch`` using ``loss_module``.
+
+        Args:
+            batch (TensorDictBase): a batch, in the format expected by
+                ``loss_module``.
+            loss_module (LossModule): computes the loss(es) for ``batch``.
+                Its output's ``"loss"``-prefixed keys are summed and
+                backpropagated; other keys are passed through for logging.
+
+        Returns:
+            TensorDictBase: the tensordict returned by ``loss_module``,
+            augmented with a ``"grad_norm"`` entry when gradient clipping is
+            enabled.
+        """
+        raise NotImplementedError
+
+    def get_weights(self) -> TensorDictBase:
+        """Return the learner's current parameters as a tensordict.
+
+        The returned tensordict is accepted as-is by
+        :meth:`~torchrl.weight_update.WeightSyncScheme.send`, so it is the
+        seam between the training role and the weight-sync/inference roles.
+        """
+        raise NotImplementedError
--- a/torchrl/trainers/learners/common.py
+++ b/torchrl/trainers/learners/common.py
@@ -6,7 +6,8 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 
-from tensordict import TensorDictBase
+import torch
+from tensordict import is_tensorclass, TensorDictBase
 from torch import nn
 
 from torchrl.objectives.common import LossModule
@@ -36,10 +37,10 @@ class LearnerCapabilities:
 class Learner(nn.Module):
     """Base class for the trainable-policy role.
 
-    A :class:`Learner` owns a trainable model and exposes a single,
-    backend-agnostic entry point, :meth:`update`, for taking one optimization
-    step on a :class:`~tensordict.TensorDictBase` batch with a given
-    :class:`~torchrl.objectives.common.LossModule`. Algorithm code calls
+    A :class:`Learner` owns a trainable model and an optimizer and exposes a
+    single, backend-agnostic entry point, :meth:`update`, for taking one
+    optimization step on a :class:`~tensordict.TensorDictBase` batch with a
+    given :class:`~torchrl.objectives.common.LossModule`. Algorithm code calls
     ``learner.update(batch, loss_module)`` without knowing whether the update
     runs locally on one device, under sharded (e.g. FSDP2) training, or on a
     separate remote training process -- that placement is the ``Learner``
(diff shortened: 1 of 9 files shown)


vmoens:
Reviewed the rebased head 3614a44bf1a6, with particular attention to the latest optimizer-coverage/checkpoint contract changes and the final FSDP2 checkpoint-cloning fix. The base and FSDP2 checkpoint paths now clone tensor state before returning it, accumulation restores safely reset partial windows, and the optimizer-coverage check runs before the first step. I ran the Learner/FSDP2-focused test selection locally: 27 passed, including the FSDP2 checkpoint round trip, accumulation/reference comparison, and plain-tensor weight gather. I did not find a blocking issue in the current code. The meaningful residual risk is unchanged from the PR description: these checks use world_size=1 gloo/CPU, so actual multi-rank collective behavior and rank-0-only checkpoint/weight gathering still need real distributed coverage.

torchrl/trainers/learners/local.py
```diff
@@ -0,0 +1,86 @@
...
+        self.optimizer = optimizer
+        self.clip_grad_norm = clip_grad_norm
+        if grad_accum_steps < 1:
+            raise ValueError(f"grad_accum_steps must be >= 1, got {grad_accum_steps}.")
+        self.grad_accum_steps = grad_accum_steps
+        self._accum_step = 0
+        self.capabilities = LearnerCapabilities(sharded=False, remote=False)
+
+    def get_weights(self) -> TensorDictBase:
+        return TensorDict.from_module(self.model)
```
gtnv: my loss_module scalar moved from 2.0 to 1.6. get_weights omitted it. restore left 99.0. how should parameters outside model be published and restored?

theap06: yea just noticed this issue. checkpoint() now addresses this issue

vmoens:
Coordinated this branch with #4140 and force-pushed the rebased stack.

The shared contract is now:
- Learner owns execution placement and weight materialization.
- OptimizationStepper is the single owner of loss reduction, backward, gradient accumulation, clipping, mixed precision, optimizer stepping, and optimizer checkpoint state.
- LocalLearner and FSDP2Learner receive the model, loss module, and stepper at construction, then expose update(batch). This is the same stepper contract used by Trainer and GRPOTrainer.

The reconciliation also moved optimizer-coverage and scalar-loss validation into MixedPrecisionOptimizationStepper, added the FSDP gradient-sync seam there, and made Learner checkpoints compose stepper state instead of maintaining a second accumulation/optimizer implementation.

I replaced the prior Learner test expansion with behavior-focused parity, weight snapshot, checkpoint, optimizer coverage, and real FSDP2/DTensor tests. Relative to #4140, test/test_trainer.py now adds 233 lines instead of 526.

Local validation: the complete test/test_trainer.py suite passes (154 passed, 19 dependency-based skips), including two FSDP2 tests. The full changed-file pre-commit suite passes. I also added ci/olddeps because this feature touches the recent FSDP2 API.

vmoens:
Rebased #3926 onto current main after #4140 merged. Because #4140 was squash-merged, I dropped its duplicated pre-squash commit sequence and replayed only the seven Learner commits. The new head is 132f0bc57; GitHub reports the PR mergeable and its diff is limited to the Learner/OptimizationStepper files. Local validation: 154 trainer tests passed with 19 dependency-based skips, and the changed-file pre-commit suite passed.

torchrl/trainers/learners/fsdp2.py
```diff
@@ -0,0 +1,177 @@
...
+        """Gather a detached plain-tensor snapshot of the sharded model."""
+        dsd = _dist_state_dict()
+        options = dsd.StateDictOptions(full_state_dict=True, cpu_offload=cpu_offload)
+        state_dict = dsd.get_model_state_dict(self.model, options=options)
+        return TensorDict(state_dict).unflatten_keys(".").detach().clone()
+
+    @contextlib.contextmanager
+    def _extras_hidden_from_optimizer(self):
+        """Hide non-model parameters from DCP and carry them explicitly."""
+        extras = self._extra_optimized_parameters()
```
gtnv, line 104: `_extra_optimized_parameters()` selects optimizer parameters outside the FSDP-managed `model`, so under the documented setup their gradients are not synchronized and can diverge across ranks. please reject them or explicitly synchronize them before allowing updates.

torchrl/trainers/learners/fsdp2.py
```diff
@@ -0,0 +1,177 @@
...
+        stepper_state = _clone_tensors(self.optimization_stepper.state_dict())
+        stepper_state.pop("optimizer", None)
+
+        dsd = _dist_state_dict()
+        options = dsd.StateDictOptions(full_state_dict=True, cpu_offload=True)
+        with self._extras_hidden_from_optimizer() as (extras, extra_state):
+            model_state, optimizer_state = dsd.get_state_dict(
+                self.model, self.optimizer, options=options
+            )
+        return {
```
gtnv, line 136: this override omits `loss_module` state outside `model`, so loss-owned target tensors and registered buffers are not restored. please include that state or reject loss modules that own it.

torchrl/trainers/learners/fsdp2.py
```diff
@@ -0,0 +1,177 @@
...
+                optim_state_dict=checkpoint["optimizer"],
+                options=options,
+            )
+
+        if not extras or checkpoint.get("extra_params"):
+            self._restore_extra_params(checkpoint)
+            for parameter, state in zip(
+                extras, checkpoint.get("extra_optim_state", [{}] * len(extras))
+            ):
+                if state:
```
gtnv, line 164: when the saved extra optimizer state is `{}`, this branch leaves the current state restored by `_extras_hidden_from_optimizer()` in place, so an early checkpoint can inherit later moments. please clear the current entry before applying the saved state.

torchrl/trainers/learners/fsdp2.py
```diff
@@ -0,0 +1,177 @@
...
+                group["params"] = parameters
+            self.optimizer.state.update(saved_state)
+
+    def checkpoint(self) -> dict[str, Any]:
+        """Gather model and optimizer state while preserving stepper state."""
+        stepper_state = _clone_tensors(self.optimization_stepper.state_dict())
+        stepper_state.pop("optimizer", None)
+
+        dsd = _dist_state_dict()
+        options = dsd.StateDictOptions(full_state_dict=True, cpu_offload=True)
```
gtnv: this call is collective, but cpu_offload returns the complete state only on rank 0. please document that every rank must enter checkpoint while only rank 0 persists its result.

torchrl/trainers/learners/common.py
```diff
@@ -0,0 +1,173 @@
...
+            setter(requires_sync)
+
+    def update(self, batch: TensorDictBase) -> TensorDictBase:
+        """Update the learner from one TensorDict micro-batch.
+
+        The returned TensorDict contains detached scalar metrics, matching the
+        :class:`~torchrl.trainers.OptimizationStepper` contract used by
+        :class:`~torchrl.trainers.Trainer`.
+        """
+        return self.optimization_stepper._step(self, batch)
```
gtnv, line 153: `TD3OptimizationStepper.step()` directly reads `clip_grad_norm` and `clip_norm`, but `Learner` defines neither, so the first `learner.update()` increments the TD3 counter and then raises `AttributeError`. please define the shared stepper-context contract or reject incompatible steppers at construction.

torchrl/trainers/learners/fsdp2.py
```diff
@@ -0,0 +1,177 @@
...
+        with self._extras_hidden_from_optimizer() as (extras, extra_state):
+            model_state, optimizer_state = dsd.get_state_dict(
```
gtnv, line 133: `get_state_dict()` gathers the full FSDP state collectively, but `full_state_dict=True` with `cpu_offload=True` returns it only on rank 0. please document that every rank must call `checkpoint()`, while only rank 0 should persist the returned checkpoint.

PR #3944, opened 2026-07-06
[Feature] trainer coverage

theap06:
## Summary: 
- Adds configurable trainers for two on-policy algorithms, following the pattern
established by `PPOTrainer`, plus two reusable trainer hooks they are built on.
- Both are runnable from YAML via `trainer@trainer: a2c|reinforce` and
`loss@loss: a2c|reinforce`.

## Motivation: 

- first pr in a planned series extending trainer coverage to the remaining
algorithms
- next pr will be: discrete-SAC/TD3+BC; then offline support for
DT/DiffusionBC; then GAIL


vmoens (approved):
LGTM thanks!

PR #3947, opened 2026-07-07
[Feature] Trajectory query language for replay buffers

theap06:
## Summary:
- Adds a trajectory-first API over replay buffer contents, in two connected pieces: Trajectory and traj. 

Trajectory: Wraps the transitions of one trajectory and exposes stored entries as attributes. 

traj: a query language for filtering.


vmoens (commented):
# Review: PR #3947 — Trajectory query language for replay buffers

Thanks for this PR — the feature is a great addition and the DSL design
(`traj.reward.sum() > 100`, composable predicates, explicit `.any()`/`.all()`
for elementwise comparisons) is clean and discoverable. The test coverage of
the happy path is solid.

I pushed a commit to your branch (`[BugFix] Harden trajectory query
splitting`) with the fixes that had a single unambiguous solution:

- Wraparound correctness: once the buffer is full, `rb[:]` returns storage
  order, not time order, so a trajectory written across the wrap point was
  fragmented and returned as two "complete" trajectories. `query()` now
  reorders chronologically using the storage write cursor before splitting.
- End-signal union: splitting on `("next", "done")` alone misses episodes
  that ended by truncation. The TED convention treats
  `done | terminated | truncated` as the end signal (see `TED2Flat`'s
  `done_keys` default), so all present flags are now OR-ed.
- Silent-merge guard: with no trajectory ids, a trajectory whose last
  transition carries no positive end flag is indistinguishable from the next
  one and gets merged. That can't be detected from the data, so the fallback
  now emits a `UserWarning` pointing users at trajectory ids /
  `trajectory_key`.
- Mechanical items: module-top import in `ReplayBuffer.query` (no circular
  dependency exists), `NestedKey | None` type hints on `trajectory_key`,
  `-> list[Trajectory]`, and an explicit `raise` instead of calling the
  elementwise predicate solely to trigger its error.

The remaining points below are ordered by importance. The first two are the
ones I'd like to see addressed (or at least designed for) before merge; the
rest are smaller decisions and suggestions.

## 1. Reuse the SliceSampler boundary machinery instead of a parallel implementation

`SliceSampler._find_start_stop_traj` / `_end_to_start_stop`
(`torchrl/data/replay_buffers/samplers.py`, ~L1997) already solve trajectory
recovery from flat storage, and handle cases query.py does not:

- `at_capacity` circular semantics: a wrap-spanning trajectory is followed
  through the wrap (`trajectory[-1:] != trajectory[:1]`), the write cursor is
  treated as an implicit truncation (`end[cursor] = 1`), and the last valid
  element is forced to be an end when the buffer is not full.
- ndim > 1 storages: the machinery works on multi-dim end tensors, whereas
  `iter_trajectories` currently raises `ValueError` for `batch_dims != 1` —
  which excludes exactly the `[B, T]` layouts (`LazyTensorStorage(...,
  ndim=2)`) where trajectory queries are most natural. Supporting them via
  per-row splitting falls out of the shared machinery nearly for free.
- Caching (`cache_values`): boundary computation is cached and invalidated on
  writes, instead of recomputed per call.

The wraparound fix I pushed is deliberately a stopgap: it copies the full
buffer (gather) whenever the storage has wrapped. Sharing the sampler's
helper removes that copy and guarantees the sampler and the query language
can never disagree about where trajectories start and end. There is a plan to
extract `_find_start_stop_traj` into a shared documented utility; happy to
coordinate so this PR can consume it (or mirror its semantics until it
lands).

## 2. Scalability: avoid `self[:]` in `query()`

`ReplayBuffer.__getitem__` does three things that make `query()` expensive at
scale (`replay_buffers.py` ~L795): it fetches the whole storage, collates it,
and applies the buffer's transform stack to the entire content.

- For `ListStorage`, collate is `_stack_anything` -> dense stack: a full copy
  of the buffer per `query()` call — and the `is_tensor_collection` guard
  currently runs after that cost is paid.
- Any transform stack runs over the full buffer (e.g. a device-cast transform
  would copy the whole buffer to GPU just to filter it).
- For `TensorStorage`/`LazyMemmapStorage` the read itself is fine (views, no
  copy — I verified this), so the concern is collate + transforms, not raw
  slicing.

Suggested direction, following the access pattern `SliceSampler` already
uses (`_get_stop_and_length` fetches only `storage[:][traj_key]`):

1. Check the storage type before any collate, so unsupported storages fail
   for free.
2. Fetch only the boundary entry (ids or end flags) from the storage, always
   from stored (untransformed) data — consistent with what the sampler sees.
3. Make predicates introspectable: each `_FieldExpr` knows its key, and
   combinators can union them, so `TrajectoryPredicate.required_keys()` is a
   small addition (plain callables return "unknown" and fall back to full
   data). Then `select()` only those keys before doing anything expensive.
4. Transforms: apply them minimally via a backward slice over the `Compose`
   chain using `in_keys`/`out_keys` — keep a transform only if its out_keys
   intersect the required set (adding its in_keys to the set); conservatively
   keep the whole suffix if any transform has unreliable key metadata. This
   preserves "predicates see transformed values" semantics without the
   full-buffer transform pass.
5. Chunked evaluation (opt-in `chunk_size`): compute all boundaries first,
   then group trajectories into chunks and select/transform/evaluate per
   chunk, so peak memory is one chunk of a few columns. Chunk edges aligned
   to trajectory boundaries.
6. `ListStorage`: use `LazyStackedTensorDict.lazy_stack` over the stored
   items (zero tensor copies) instead of a dense stack; `.select()` and
   slicing propagate lazily, and only the fields a predicate names get
   densified, per trajectory.

Longer term, the DSL is declarative enough that reductions can be evaluated
without a per-trajectory Python loop at all: `traj.reward.sum() > 100` over
100k trajectories can compile to a single segment reduction over the flat
column plus a boolean mask. Not a merge requirement, but worth keeping the
expression objects introspectable (rather than opaque closures) so this stays
reachable.

## 3. `Trajectory`: decide the class shape before merge

Consider making `Trajectory` a single-field tensorclass
(`class Trajectory(TensorClass["nocast"]): data: TensorDictBase`) rather than
a free class:

- slicing/indexing would return `Trajectory` (today `trajectory[1:3]`
  silently returns a bare `TensorDict`, losing `length`/`total_reward` and
  the `("next", ...)` fallback);
- query results could be `lazy_stack`-ed into one ragged batched `Trajectory`
  instead of a Python list — stackable, `.to()`-able, and re-`extend`-able
  into a buffer ("filter then train on the result" is the obvious next use);
- `.to()`, memmap/serialization and torch.compile support come for free.

The trade-off is a larger attribute-collision surface (stored keys shadowed
by `TensorDictBase` method names — recoverable via `.get()`/`.data`) and a
different constructor. A fully dynamic tensorclass is not on the table
(namespace collision makes it unsound); the single-field wrapper keeps your
custom `__getattr__` fallback for non-reserved names. Whichever way this
goes, it needs deciding now: converting after release is a breaking change to
the constructor and `isinstance` checks. At minimum, `__getitem__` with
slice-like indexes should return a `Trajectory`.

I'm open to you pushing back on this, it's a design choice and could bloat the class and not bring much useful features.

## 4. Smaller points

- Placement: `query()` on the base `ReplayBuffer` raises a runtime
  `TypeError` for non-tensordict storages. That matches other
  tensordict-conditional behavior on the base class, but moving it to
  `TensorDictReplayBuffer` would surface the constraint at class-selection
  time. Maintainer preference; fine either way. (For the record:
  `query()` on a `ReplayBufferEnsemble` fails gracefully with a clear
  `TypeError`, since `ensemble[:]` returns the ensemble itself.)
- Incomplete trajectories: the splitter closes the last segment at the data
  frontier, so an in-progress episode looks complete, and after wraparound
  the oldest trajectory can be head-truncated. A `Trajectory.is_terminated`
  property (last end flag set) plus an opt-in `drop_incomplete=True` on
  `iter_trajectories`/`query` would let users exclude partials.
- Duplication: the six comparison dunders are defined verbatim in both
  `_ScalarExpr` and `_FieldExpr`, and the two `_compare` methods differ only
  in the returned wrapper class. A shared base with a `_result_cls`
  attribute collapses ~40 lines. Suggestion, not a blocker — explicit dunders
  do keep IDE support.
- Docs: `traj` is referenced with `:data:` in the prose of
  `data_replaybuffers.rst` but absent from the autosummary lists; likely fine
  for a module-level singleton, flagging in case you want it listed.

## Verified non-issues

To save you the worry: full-buffer reads on `TensorStorage`/memmap storages
are views (no RAM spike), and multi-agent done layouts cannot silently
corrupt splitting (keys under `("agents", ...)` are never matched and raise a
clear `KeyError` instead).

theap06:
@vmoens adopted the tensorclass shape. thanks for the feedback

vmoens (approved):
LGTM

PR #3950, opened 2026-07-07
[LLM] WS4 Post-Training Observability: Implement PostTrainingLogger

### LINKED ISSUE #3948: [RFC] Post-training efforts
##Summary
TorchRL's LLM post-training stack has grown quickly: GRPO and SFT objectives
(`torchrl/objectives/llm`), vLLM backends and LLM policy wrappers
(`torchrl/modules/llm`), an inference server, weight-sync schemes, and
recipes for GRPO, expert iteration, VLA-GRPO and reward-model training. What
it does not yet have is an adoption path for teams that already run their
post-training on other stacks (TRL, NeMo-RL, or in-house loops).

This RFC proposes treating **interoperability as the product**: making
TorchRL's post-training components individually consumable from existing
workflows, rather than asking users to migrate whole training loops. Nobody
adopts a full stack at once; they adopt one component that solves a problem,
then a second.

## Motivation

The teams doing LLM post-training today mostly have a working loop already.
Their pain points are component-level: replay/rollout storage that scales,
collectors that decouple generation from training, losses with correct
token-level bookkeeping, and observability that makes RL runs debuggable.
TorchRL has strong answers to each, but they are currently easiest to use
from inside a TorchRL loop. If a TRL or NeMo-RL user could pull in a single
TorchRL component with a thin adapter and no loop rewrite, each component
becomes an on-ramp to the rest of the stack.

## Proposed workstreams

Each is independently useful and separately mergeable; none blocks another.

**1. Stable component boundaries (the foundation).** Define and document the
small set of interfaces external loops consume: "give me a buffer"
(extend/sample over token or trajectory data), "give me rollouts"
(collector/inference-server protocol, including weight sync), "give me a
loss" (tensordict-in, losses-out with documented key contracts). These
mostly exist implicitly; the work is making them explicit, versioned and
tested as public contracts, so adapters stay thin and don't break on either
side's releases.

**2. TRL interop.** Thin adapters so TRL users can use TorchRL replay
buffers and collectors inside TRL trainers, and conversely so TorchRL
recipes can consume HF-ecosystem reward models and datasets with zero
friction. Deliverable: an examples page + adapter module + CI-tested
round-trip ("TRL trainer with TorchRL buffer" and "TorchRL GRPO with HF
reward model").

**3. NeMo-RL interop.** Same shape as (2), scoped after a short discovery
pass on where the natural seams are (rollout storage and weight sync are
the likely candidates). NVIDIA-stack users are heavy post-training adopters;
even a storage-only integration is a meaningful on-ramp.

**4. Post-training observability.** A standardized logging layer for the
metrics post-training practitioners actually watch: per-token/per-sequence
reward stats, KL to reference, entropy, generation throughput, weight-sync
latency, buffer utilization. One hook set, emitting to the existing logger
backends, used uniformly across the GRPO/expert-iteration/RM recipes, and
usable standalone from external loops. This makes TorchRL runs comparable
and debuggable, and it is the component external users can adopt with the
least commitment.

coder-jayp:
## Description

This PR implements the Observability workstream (WS4) from RFC #3948. It introduces a modular `PostTrainingLogger` that unifies metric emission across TorchRL's post-training algorithms, making the observability stack consumable by external training loops.

**Key Changes:**
* **`PostTrainingLogger`**: New backend-agnostic logger for tracking loss components, buffer utilization, and inference staleness. Uses `getattr` duck-typing to safely ingest both `GRPOLossOutput` and `SFTLossOutput`.
* **Recipe Refactoring**: Updated GRPO and Expert Iteration (EI) recipes to delegate to the new logger. 
* **Deprecation Handling**: EI legacy metric keys are temporarily emitted in parallel with explicit `DeprecationWarning`s (targeted for removal in `v0.15.0` per `CLAUDE.md`).
* **Docs & Tests**: Added full Sphinx documentation and comprehensive unit testing in `test_loggers.py`.

## Motivation and Context

Solves the need for component-level interoperability in TorchRL's LLM stack. Users can now drop this logger into their own existing loops without migrating their entire workflow. 

close #3948

## Types of changes

## Checklist

### DIFF
--- a/sota-implementations/expert-iteration/ei_utils.py
+++ b/sota-implementations/expert-iteration/ei_utils.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import time
+import warnings
 from typing import Any, Literal
 
 import torch
@@ -19,6 +20,7 @@ from torchrl.envs.llm.datasets.countdown import CountdownEnv
 from torchrl.envs.llm.datasets.ifeval import IFEvalEnv
 from torchrl.envs.llm.datasets.math import MATHEnv
 from torchrl.modules.llm import TransformersWrapper, vLLMWrapper
+from torchrl.record.loggers.llm import PostTrainingLogger
 from torchrl.weight_update.llm import VLLMWeightSyncScheme
 from transformers.models.auto.modeling_auto import AutoModelForCausalLM
 from transformers.tokenization_utils import PreTrainedTokenizer
@@ -651,30 +653,75 @@ def log_training_metrics(
 ):
     """Log training metrics to wandb.
 
-    Args:
-        wandb_logger: The wandb logger instance
-        replay_buffer: The replay buffer containing collected data
-        batch: The current training batch
-        loss: The computed loss object
-        grad_norm: The gradient norm value
-        global_step: Current global training step
-        data_read_count: Total data read count
-        collector: The collector instance
-        start_time: Training start time
-        gradient_accumulation_steps: Number of gradient accumulation steps
-        history_str: Optional history string for logging
-    """
-    with torch.no_grad():
-        rb_content = replay_buffer[:]
-        batch_policy_version = batch["next", "policy_version"].view(-1).min()
-        batch_policy_age = collector.policy_version - batch_policy_version
+    Delegates standard metrics to :class:`~torchrl.record.loggers.PostTrainingLogger`.
 
-        metrics = {
-            "reward from buffer": float(
+    .. deprecated::
+        The legacy flat metric keys (``"reward from buffer"``,
+        ``"loss_sft, from loss"``, etc.) are deprecated and will be removed
+        in v0.15.0.  The new ``namespace/metric`` keys are already being
+        emitted in parallel — update any dashboards to use those.
+
+    Args:
+        wandb_logger: The wandb logger instance.
+        replay_buffer: The replay buffer containing collected data.
+        batch: The current training batch.
+        loss: The computed loss object.
+        grad_norm: The gradient norm value.
+        global_step: Current global training step.
+        data_read_count: Total data read count.
+        collector: The collector instance.
+        start_time: Training start time (``time.time()`` at loop start).
+        gradient_accumulation_steps: Number of gradient accumulation steps.
+        history_str: Optional history string for logging.
+    """
+    pt_logger = PostTrainingLogger(logger=wandb_logger, start_time=start_time)
+
+    pt_logger.log_training_step(
+        loss=loss,
+        step=global_step,
+        grad_norm=grad_norm,
+        gradient_accumulation_steps=gradient_accumulation_steps,
+    )
+    pt_logger.log_collection_step(
+        batch=batch,
+        replay_buffer=replay_buffer,
+        collector=collector,
+        step=global_step,
+    )
+
+    # Legacy flat keys emitted in parallel until v0.15.0 (see CLAUDE.md §12).
+    # Each key here maps to the new namespace/metric equivalent listed in the
+    # deprecation warning string below.
+    _LEGACY_KEY_MAP = {
+        "reward from buffer": "buffer/reward_mean",
+        "loss_sft, from loss": "training/loss_sft",
+        "loss_kl_to_ref, from loss": "training/loss_kl_to_ref",
+        "kl_to_ref, from loss": "training/kl_to_ref",
+        "grad_norm": "training/grad_norm",
+        "write_count, from buffer": "buffer/write_count",
+        "gradient_step_throughput (gradient step per write)": "throughput/gradient_steps_per_write",
+        "optim_step_throughput (optim step per write)": "throughput/optim_steps_per_write",
+        "data_read_count (total)": "buffer/data_read_count",
+        "current_policy_version (collector)": "inference/policy_version",
+        "throughput (steps per second)": "throughput/gradient_steps_per_second",
+    }
+    for old_key, new_key in _LEGACY_KEY_MAP.items():
+        warnings.warn(
+            f'EI metric key "{old_key}" is deprecated and will be removed in v0.15.0. '
+            f'Use "{new_key}" instead.',
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+    with torch.no_grad():
+        optim_steps = global_step // gradient_accumulation_steps
+        legacy_metrics: dict[str, Any] = {}
+        try:
+            rb_content = replay_buffer[:]
+            legacy_metrics["reward from buffer"] = float(
                 torch.cat(rb_content.get(("next", "reward"), as_list=True)).mean()
-            ),
-            "reward from batch": float(batch["next", "reward"].mean()),
-            "seq_length from buffer": float(
+            )
+            legacy_metrics["seq_length from buffer"] = float(
                 torch.tensor(
                     [
                         t.numel()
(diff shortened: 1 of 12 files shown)


coder-jayp:
@vmoens, I would appreciate your feedback on this when you have a moment. Thanks.

coder-jayp:
Hi @theap06, any feedback on this PR?

torchrl/record/loggers/llm.py
```diff
@@ -0,0 +1,220 @@
...
+            step (int, optional): Global gradient-step counter.
+
+        Returns:
+            dict[str, Any]: The metrics dict that was logged.
+        """
+        metrics: dict[str, Any] = {}
+
+        with torch.no_grad():
+            try:
+                reward_list = batch.get(("next", "reward"), default=None, as_list=True)
```
theap06, line 149: This reads from batch, not replay_buffer[:]. Old code for these three metric families read the whole buffer. Should rename the keys or read from replay_buffer

torchrl/record/loggers/llm.py
```diff
@@ -0,0 +1,220 @@
...
+                    f"PostTrainingLogger: could not read response tokens: {exc}"
+                )
+
+            if replay_buffer is not None:
+                try:
+                    metrics["buffer/write_count"] = int(replay_buffer.write_count)
+                    # _storage is private; max_size is a public plain attribute on Storage.
+                    storage = replay_buffer._storage  # noqa: SLF001
+                    if hasattr(storage, "max_size") and storage.max_size > 0:
+                        metrics["buffer/utilization"] = (
```
theap06: RayReplayBuffer never sets self._storage so it raises an AttributeError. also, the buffer/utilization never logs for grpo since both use the rayreplaybuffer

sota-implementations/expert-iteration/ei_utils.py
```diff
@@ -651,69 +653,120 @@ def log_training_metrics(
...
-        }
-
-        wandb_logger.log_metrics(metrics, step=global_step)
-
-        if history_str is not None:
-            wandb_logger.log_str("history", history_str, step=global_step)
+            )
+        except Exception:  # noqa: BLE001
+            pass
+        legacy_metrics["reward from batch"] = float(batch["next", "reward"].mean())
```
theap06: Only the rb_content block above this is guarded; the loss.loss_sft reads here aren't and nothing wraps the log_training_metrics either.

sota-implementations/grpo/grpo_utils.py
```diff
@@ -919,91 +919,55 @@ def log_training_metrics(
 ):
```
theap06, line 935: A few keys got renamed without any docstrings or deprecation warnings. ei_utils.py keeps everything alive via _LEGACY_KEY_MAP, worth doing the same here

sota-implementations/expert-iteration/ei_utils.py
```diff
@@ -651,69 +653,120 @@ def log_training_metrics(
...
+            f'EI metric key "{old_key}" is deprecated and will be removed in v0.15.0. '
+            f'Use "{new_key}" instead.',
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+    with torch.no_grad():
+        optim_steps = global_step // gradient_accumulation_steps
+        legacy_metrics: dict[str, Any] = {}
+        try:
```
theap06, line 724: All 11 warnings are made every call. Instead, worth a module-level flag so it's not 11 warnings. Warn calls every logged step.

theap06 (changes requested):
Sorry for the late reply! Left some comments.

coder-jayp:
@theap06 Thanks for the detailed review. All points addressed.

vmoens:
Reviewed current head `4a2d797f0700` after rebasing onto the fetched `origin/main`, including the latest corrective commit and all five unresolved review threads. The latest code does address those threads: metrics are batch-namespaced, buffer counters use public `stats()`, EI accesses are guarded, GRPO legacy keys warn with a removal version, and the persistent logger makes warning suppression effective. The 20 PostTrainingLogger tests pass.

There is still an optimizer-step accounting bug across the new logger and its sync call sites. `torchrl/record/loggers/llm.py:114-118` treats `step % gradient_accumulation_steps == 0` as the optimizer boundary and derives `optim_steps = step // accumulation`, while both `sota-implementations/grpo/grpo-sync.py:245` and `sota-implementations/expert-iteration/expert-iteration-sync.py:278` actually step on `(global_step + 1) % accumulation == 0`. With accumulation 4, the optimizer runs at step 3 but the logger omits that gradient norm, then reports a stale norm at step 4 and still reports zero optimizer steps at step 3. The new unit test encodes the opposite schedule from the callers. Please make the callers and logger share one convention, preferably fixing the one-based caller loops to step on `global_step % accumulation == 0`, and add an integration-level assertion around the call site.

Checks: `pytest -q test/test_loggers.py -k PostTrainingLogger` gives 20 passed; complete diff and latest thread state inspected; `git diff --check` is clean.

coder-jayp:
@vmoens Fixed in 9b5c0ecc0 — changed both grpo-sync.py and expert-iteration-sync.py to use global_step % accumulation == 0 to align with the logger's boundary. Added an integration-level test that simulates the actual loop pattern and asserts grad_norm fires at exactly steps 4, 8, 12. 21 tests passing.

torchrl/record/loggers/llm.py
```diff
@@ -0,0 +1,258 @@
...
+            try:
+                reward_list = batch.get(("next", "reward"), default=None, as_list=True)
+                if reward_list is not None:
+                    # as_list=True returns a plain tensor when the batch is
+                    # dense (padded) rather than ragged.
+                    if isinstance(reward_list, torch.Tensor):
+                        reward_tensor = reward_list.reshape(-1).float()
+                    else:
+                        reward_tensor = torch.cat(reward_list).float()
+                    metrics["batch/reward_mean"] = float(reward_tensor.mean())
```
gtnv, line 157: the collector mask is [true, false]. rewards [1, 1000] still log batch/reward_mean=500.5. can we apply the mask before reducing?

coder-jayp:
@vmoens The recent merge of `GRPOTrainer` removed the manual training loops, which conflicts with this branch. 

Does it still make sense to keep `PostTrainingLogger` to preserve the extended LLM-specific metrics (unpadded sequence lengths, policy staleness, buffer stats, throughput)? 

If yes, I'll resolve the conflicts.

vmoens (changes requested):
Thanks for checking. Please don’t resolve the conflicts on the current implementation. With `GRPOTrainer` merged, I don’t think we should retain `PostTrainingLogger` as a second public path for post-training.

I want the canonical API to remain `Trainer(..., logger=...)`: training components produce metrics, the Trainer coordinates them, and `Logger` is only the backend-independent sink. The logger should not receive batches, replay buffers, collectors, accumulation settings, or reconstruct optimizer state. TorchRL has already a broad surface and many APIs, let's be cautious about adding more of them. 

The useful missing metrics should be moved into that pipeline: batch-derived metrics such as masked rewards, unpadded response lengths, and sampled-batch staleness belong in `Trainer` hooks; collector and replay-buffer counters belong in `stats()`/`LoggerMonitor`; weight-sync latency belongs to the existing timed hook path. Training losses, gradient norms, and optimizer-step counts should continue to come from the optimization stepper.

**I suggest** closing or superseding this PR with a small follow-up based on main that adds only the missing metric behavior, settles one canonical namespace, and uses compact end-to-end Trainer coverage. Multiple logging backends are welcome, but there should be only one post-training logging API.

cc @gtnv @theap06

coder-jayp:
@vmoens Makes sense. Closing this PR.

PR #3953, opened 2026-07-08
[CI] Speed up linux unit-test jobs: xdist worksteal parallelism and slimmer container setup

vmoens:
## Motivation

The unit-test workflow's wall clock has grown from ~64 to ~98 min since April, tracking test-suite growth (26,950 to 29,360 collected tests). The structural problem: the five `tests-cpu` jobs each run the full suite **serially** on 48-vCPU `linux.12xlarge` runners (~70 min of pytest each), and `tests-optdeps` does the same on its GPU runner (~87 min - the workflow's critical path). Only the GPU shards use pytest-xdist today.

### Test execution
- **Enable pytest-xdist for the parallelism-friendly bulk of the suite on CPU runners** (the unsharded `all` path used by `tests-cpu`, and shard 1). The multiprocessing-heavy set (`test/envs`, `test_collectors.py` - shard 2's file set) stays serial; on the `all` path it runs as a second serial invocation covering the same union of tests.
- **Switch shard 3 from `--dist loadgroup` to `--dist worksteal`**: there are no `xdist_group` marks in the suite, so loadgroup bought nothing, and worksteal is the right scheduler for a long tail of millisecond tests mixed with a few 60s+ tests (e.g. `TestHydraParsing`).
- **Pin `OMP_NUM_THREADS=1` / `MKL_NUM_THREADS=1` for xdist runs**: with N workers on an N-core runner, letting every torch op fan out to all cores makes workers serialize on thread contention instead of scaling. Serial invocations keep default threading.
- **Raise the per-test timeout to 300s for xdist runs** (serial stays 120s): tests that take ~60s alone can legitimately exceed 120s on a fully loaded machine.
- **optdeps**: split the single 87-min serial run into (1) a `-m 'not gpu'` xdist bulk, (2) a serial multiprocessing-heavy pass, (3) a serial `-m gpu` pass (workers sharing one device could oversubscribe GPU memory). Same union of tests, no overlap. Also: keep going on failure so coverage still combines/uploads, and propagate the exit status at the end (previously `set -e` aborted before coverage upload).
- GPU runners outside shard 3 keep serial execution. `TORCHRL_XDIST=0` restores the previous single-invocation behavior; `TORCHRL_XDIST_WORKERS` overrides `-n auto`.

### Coverage under xdist
xdist's execnet workers are plain subprocesses, invisible to coverage's `concurrency=multiprocessing` (this already silently applied to GPU shard 3). Standard fix: a `coverage_process_startup.pth` in site-packages calling `coverage.process_startup()`, activated by `COVERAGE_PROCESS_START` which `coverage_run_parallel.py` now exports; `coverage combine` then picks up the worker data files.

### Container setup (warm start)
Every job starts from a stock `nvidia/cuda` image and paid for `apt-get upgrade` + `dist-upgrade` of a throwaway container (minutes per job, no test benefit). Dropped in favor of a targeted `--only-upgrade libstdc++6` (the actual reason the dist-upgrade existed: GLIBCXX for torch wheels), and apt installs merged into one pass with `--no-install-recommends`.

## Expected effect

- `tests-cpu`: ~78 min to roughly 25-35 min per job (bulk parallelized across 48 vCPUs; serial envs/collectors remainder ~15 min of it).
- `tests-optdeps` (critical path): ~87 min to roughly 30-40 min.
- A few minutes less setup on every linux job.

The exact numbers need this PR's own CI runs to validate; the per-test timeout bump and thread pinning are there to keep flake risk down under load. If some parallel bulk turns out flaky, `TORCHRL_XDIST=0` on the affected job is an immediate escape hatch.

## Follow-ups (out of scope)

- Prebaked CI docker image with system libs + a warm uv cache (kills the remaining ~5 min setup everywhere, including the libs workflow where robohive/mujoco-playground installs dominate).
- 3-way sharding for CPU jobs if xdist alone does not get them under 30 min.


vmoens:
Superseeded

PR #3959, opened 2026-07-09
[Performance] Memory-map trainer checkpoints on load

theap06:
##Summary: 
- The fix: kwargs.setdefault("mmap", True) in Trainer.load_from_file plus the docstring note
- restoring a checkpoint through TorchRL's default torch backend materializes the entire checkpoint as non-evictable RAM before a single byte reaches the destination
image
image


vmoens (approved):
LGTM thanks!

PR #3966, opened 2026-07-09
[BugFix] Copy tensors in load_state_dict instead of aliasing the source

vmoens:
## Description

Several `load_state_dict` implementations retained caller tensors by reference, aliasing external state into live components. With `Trainer.load_from_file` now defaulting to `torch.load(mmap=True)` (#3959), this pins the checkpoint file mapping for the process lifetime (`PermissionError` on Windows when overwriting the file, unspecified reads if it is rewritten externally) and lets later mutations of the source state dict corrupt the loaded component.

This PR makes `load_state_dict` clone/deepcopy incoming tensors so loaded state is decoupled from the caller's objects, matching `nn.Module.load_state_dict` semantics.

Root cause for the storage paths: `TensorDict.load_state_dict` calls `set(key, item, inplace=True)`, which copies in place for existing keys but assigns by reference for missing ones, so loading onto an empty `TensorDict()` aliased every tensor.

Fixed sites:

- `TensorStorage.load_state_dict`: clone the tensor / TensorDict when loading onto an uninitialized storage. Initialized destinations already copied in place and are unchanged.
- `ListStorage.load_state_dict`: clone tensor elements and the TensorDicts rebuilt from dict elements.
- `CompressedListStorage.load_state_dict`: copy the storage list (cloning tensor entries) and deepcopy the metadata.
- `ConsumingSampler.load_state_dict`: clone `_sample_count`, `_live_mask`, `_free_indices`.
- `SamplerWithoutReplacement.load_state_dict` (also covers `SliceSamplerWithoutReplacement`): clone `_sample_list`.
- `PrioritizedSampler.load_state_dict`: deepcopy the sum/min trees and `_max_priority` (mirroring `state_dict()`), and stop popping keys out of the caller's dict.
- `RewardNormalizer.load_state_dict`: deepcopy loaded values (the reward stats tensors).

Writers were audited as well: they only persist scalar cursors (or raise `NotImplementedError`), so no change is needed. `LazyMemmapStorage` was already safe since its uninitialized paths copy into fresh memmap files.

## Tests

Regression tests added to the existing test files (`test/rb/test_storages.py`, `test/rb/test_samplers.py`, `test/rb/test_rb_core.py`, `test/test_trainer.py`) asserting that mutating the source state dict after loading does not affect the component. All 7 new tests fail without the library fix and pass with it. Also verified end-to-end that a state dict loaded via `torch.load(mmap=True)` no longer shares storage with the checkpoint mapping.

## Checklist

- [x] I have read the CONTRIBUTING document.
- [x] My change requires a change to the documentation: no.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

## Note on the olddeps CI failure

The `tests-olddeps` (torch 2.1) failure previously visible on this PR (`test_resume_and_resave_replay_buffer` failing with `Unsupported class torch.device` under `weights_only=True`) was a pre-existing `main` breakage from #3959. A version- and platform-aware fix landed on `main` in #3974 and is now merged into this branch; the interim fix that briefly lived here was dropped in favor of it during the merge.

### DIFF
--- a/test/rb/test_rb_core.py
+++ b/test/rb/test_rb_core.py
@@ -937,6 +937,33 @@ class TestReplayBufferConsumption:
         assert write_index.tolist() == consumed.tolist()
         assert rb2.storage.get(write_index).tolist() == [100, 101]
 
+    def test_replay_buffer_consume_load_state_dict_decouples_from_source(self):
+        # loading a state dict must clone the incoming tensors so that
+        # mutating the source afterwards does not corrupt the sampler state
+        # (e.g. tensors mmap-backed by torch.load(mmap=True))
+        rb = ReplayBuffer(
+            storage=LazyTensorStorage(5),
+            batch_size=2,
+            consume_after_n_samples=1,
+        )
+        rb.extend(torch.arange(5))
+        rb.sample()
+
+        sampler_sd = rb.sampler.state_dict()
+        sampler2 = ConsumingSampler()
+        sampler2.load_state_dict(sampler_sd)
+
+        before_count = sampler2._sample_count.clone()
+        before_mask = sampler2._live_mask.clone()
+        sampler_sd["_sample_count"].fill_(42)
+        sampler_sd["_live_mask"].fill_(False)
+        assert (sampler2._sample_count == before_count).all()
+        assert (sampler2._live_mask == before_mask).all()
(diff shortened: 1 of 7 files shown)


theap06:
@vmoens thanks

PR #3968, opened 2026-07-09
[Test] Skip PPO vecnorm checkpoint roundtrip test without a gym backend

vmoens:
## Description

`test/test_render.py::TestSotaCheckpointFactories::test_ppo_vecnorm_checkpoint_roundtrip` fails deterministically on main's push CI in the tests-optdeps job and the Windows unittests job with:

```
ModuleNotFoundError: Supported version of 'torchrl.envs.libs.gym.GymEnv._set_gym_args' has not been found.
```

Those jobs install neither gym nor gymnasium. The test builds a `GymEnv("CartPole-v1")` via the PPO sota-implementations `make_env` factory but, unlike its siblings in `TestSotaCheckpointFactories`, carries no gym-backend skip guard. With no backend installed, `implement_for` dispatch on `_set_gym_args` fails in `GymEnv.__init__` before `_build_env` can raise its friendlier "gym not found" error.

The `implement_for` version ranges on `_set_gym_args` are exhaustive, and the test passes in the tests-cpu jobs, so this is purely a missing skip guard rather than a dispatch or version-range bug.

## Fix

Add a `skipif` guard that skips the test when neither gym nor gymnasium is importable. The guard accepts either backend because tests-olddeps runs the test with only `gym==0.13` (no gymnasium) and passes, so guarding on gymnasium alone would regress coverage there.

### DIFF
--- a/test/test_render.py
+++ b/test/test_render.py
@@ -49,6 +49,7 @@ from torchrl.render.mujoco_wasm import extract_qpos_trajectory
 from torchrl.render.video import compose_frame_grid, normalize_frame_output
 
 _has_pil = importlib.util.find_spec("PIL") is not None
+_has_gym = importlib.util.find_spec("gym") is not None
 _has_gymnasium = importlib.util.find_spec("gymnasium") is not None
 _has_pygame = importlib.util.find_spec("pygame") is not None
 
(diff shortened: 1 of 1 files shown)


vmoens:
Superseded by #3962, which landed the same skip guard (plus a guard on the DQN factory test) on main.

PR #3986, opened 2026-07-10
[Performance] Add Collector.clone_data to skip yield clones

theap06:
## Summary: 
- added a clone_data field to the collectors to prevent redundant clones 
- each yield is still a clone so successive batches do not alias the collector’s in-place rollout buffer.
- return_same_td=True still takes precedence; attaching replay_buffer= remains the preferred zero-clone path for off-policy loops.

### DIFF
--- a/test/test_collectors.py
+++ b/test/test_collectors.py
@@ -5297,6 +5297,108 @@ class TestCollectorRemoteHelpers:
             collector.shutdown()
 
 
+class TestCollectorCloneData:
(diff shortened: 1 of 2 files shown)


vmoens (commented):
Thanks! I find this confusing, it's basically the same as the return_same_td arg, and it's not obvious when using one or the other. 
The collector constructor is quite loaded already so if we could spare one additional arg I'd rather take that.
Can you explain why this is needed?

PR #3990, opened 2026-07-13
[BugFix] Restore sampler state in SliceSampler subclasses

theap06:
## Summary: 
- SliceSampler defines its persistence methods as no-ops because its own state is a rebuildable cache. 
- In the multiply-inherited subclasses SliceSampler comes first in the MRO, so those no-ops shadow the co-parent's real implementations. 
- Both subclasses already work around this for some methods, and each misses others.


vmoens (approved):
LGTM thanks!

PR #4008, opened 2026-07-15
[Performance] Write large ParallelEnv rollouts directly

vmoens:
## Summary

- let buffered, same-device `ParallelEnv` return internal shared-memory views to a registered collector
- write root and `next` values directly into their final rollout time slices instead of cloning in `ParallelEnv` and stacking afterward
- hand single collectors a fresh owning allocation, preserving returned-batch immutability without a final clone
- gate the path to tensor-only payloads of at least 256 KiB per batched step; smaller and unsupported cases retain the existing path

## Local benchmarks

Apple Silicon results for `Collector(ParallelEnv(4, ...), compact_obs=True)`, 128 frames per batch, compared with `origin/main` at `808b0ac`:

| Observation per environment | main median | stack median | change |
| --- | ---: | ---: | ---: |
| 256 KiB | 6.10 ms | 5.60 ms | -8.3% |
| 1 MiB | 11.10 ms | 7.39 ms | -33.4% |

Linux benchmark CI is enabled on this PR and is the authoritative performance check.

## Test plan

- `pytest test/test_collectors.py -k 'parallel_env_direct_output or multisync_parallel_env_direct_output'`
- `pytest test/envs/test_parallel.py -k 'compact_collector or direct_collector'`
- `pytest test/test_collectors.py::TestCollectorRB::test_parallel_env_with_multi_collector_and_replay_buffer`
- `pytest test/test_collectors.py::TestHeterogeneousEnvsCollector`
- `pytest benchmarks/test_collectors_benchmark.py::test_parallel_env_compact_obs --benchmark-only`

* __->__ #4008
* #4007
* #4006

### DIFF
--- a/test/envs/test_parallel.py
+++ b/test/envs/test_parallel.py
@@ -122,6 +122,28 @@ class TestParallel:
         finally:
             env.close(raise_if_closed=False)
 
+    def test_direct_collector_borrows_step_output(self, maybe_fork_ParallelEnv):
+        class DirectCollector:
+            _compact_next_keys = ()
+            _write_env_output_directly = True
+
+        env = maybe_fork_ParallelEnv(2, CountingEnv, use_buffers=True)
+        collector = DirectCollector()
+        env.register_collector(collector)
+        try:
+            data = env.rand_action(env.reset())
+            step, post_reset = env.step_and_maybe_reset(data)
+            assert (
+                step.get(("next", "observation")).data_ptr()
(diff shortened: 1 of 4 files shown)


vmoens:
Closing this layer because the Linux benchmark did not show a repeatable gain. The CPU run was -5.4% at 256 KiB and +1.8% at 1 MiB, while the GPU-hosted run was -15.4% at 1 MiB. The borrowed-buffer lifetime complexity is not justified by those results; the lower copy-elision changes remain in the stack.

PR #4015, opened 2026-07-15
[CI] Stabilize PR benchmark confirmations

vmoens:
## Summary

- compare benchmark median duration instead of mean-based throughput
- confirm significant improvements as well as regressions and near-threshold changes
- repeat selected baseline and PR benchmarks in balanced baseline/PR/PR/baseline order
- aggregate repetitions with equal weight and retain the per-run JSON measurements
- require at least 10 rounds for non-pedantic confirmation benchmarks

## Root cause

Follow-up analysis of #4010 showed three sources of false signal:

1. The confirmation planner selected significant regressions and changes near 5%, but omitted improvements above the threshold window. In the completed run, 9 CPU and 16 GPU large improvements therefore remained comparisons between different runners.
2. The comparator preferred pytest-benchmark's mean-based `ops` value. Rare stalls could dominate that value: one CPU replay-buffer result reported +279% by mean but was about +1% by median; two GPU outliers reported +32.5% and +31.1% by mean but were -2.6% and -0.3% by median.
3. A single fixed baseline-then-PR confirmation did not balance cold-start effects or temporal drift on the runner.

The follow-up uses the median for outlier resistance, sends every change that can cross the reporting threshold through confirmation, and balances two repetitions per revision in ABBA order before taking the median of the per-run medians.

## Impact

The opt-in workflow spends more time only when a comparison needs confirmation. Final reports should contain fewer false improvements and regressions, while the uploaded repetition JSON makes remaining variability diagnosable.

## Validation

- `uvx --from pytest pytest -q .github/scripts/test_pr_benchmark_workflow.py` (15 passed)
- `actionlint -ignore 'label .* is unknown' .github/workflows/benchmarks_pr.yml`
- `pre-commit run --files .github/workflows/benchmarks_pr.yml .github/scripts/compare_pr_benchmarks.py .github/scripts/comment_pr_benchmarks.py .github/scripts/test_pr_benchmark_workflow.py`
- replayed confirmation planning and aggregation against the JSON artifacts from run 29431181113

### DIFF
--- a/.github/scripts/comment_pr_benchmarks.py
+++ b/.github/scripts/comment_pr_benchmarks.py
@@ -23,12 +23,15 @@ def _benchmark_name(benchmark: dict) -> str:
 
 def _benchmark_ops(benchmark: dict) -> float | None:
     stats = benchmark.get("stats", {})
-    ops = stats.get("ops")
-    if ops is not None:
-        return float(ops)
+    median = stats.get("median")
+    if median:
+        return 1.0 / float(median)
     mean = stats.get("mean")
     if mean:
         return 1.0 / float(mean)
+    ops = stats.get("ops")
+    if ops is not None:
+        return float(ops)
     return None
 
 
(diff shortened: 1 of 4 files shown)


vmoens:
Superseded by #4019, which keeps one benchmark measurement per revision and device and applies targeted benchmark-definition fixes.

PR #4016, opened 2026-07-15
[Feature] Re-enable prefix caching for GRPO online weight updates

vmoens:
## Motivation

Prefix caching was disabled for GRPO inference backends (#3809) because cached KV prefixes are keyed by prompt content, not by the weights that produced them: after an online weight update, cached prefixes are stale and corrupt behavior-policy logprobs. This PR restores prefix caching safely by invalidating the caches as part of every weight sync.

## Changes

- **[BugFix] Respect enable_prefix_caching in AsyncVLLM launch path** - opting in was silently broken: `AsyncVLLM.__init__` overwrote `engine_args.enable_prefix_caching` with its own `False` default and `launch()` never forwarded the flag, so `enable_prefix_caching=True` never reached the engine.
- **[Feature] Reset vLLM prefix caches on weight updates** - new `reset_prefix_cache()` on `_AsyncLLMEngine`, `AsyncVLLM` (fan-out over replicas, no-op when caching is disabled, warns when a replica cannot fully clear), `RayLLMWorker`, `LocalLLMWrapper`, and a no-op default on `RLvLLMEngine`. Invoked automatically in every weight-update path after weights land and before `wake_up`, so no request can be scheduled against a stale prefix.
- **[Feature] Flush SGLang radix cache on weight updates** - `flush_cache_on_update` on `SGLangWeightSyncScheme` defaults to auto: the flush is requested exactly when the pause mode is `abort`, the only mode under which SGLang (which flushes only when idle) can honor it. Explicit `True` requires `pause_mode="abort"` and raises otherwise; an env-forced flush under other pause modes is downgraded with a warning. `pause_mode` is now a scheme argument alongside the `TORCHRL_SGLANG_PAUSE_GENERATION_MODE` env var.
- **[BugFix] Keep MCAdvantage trajectories lazy when contiguous() drops entries** - found while validating: `TensorDict.contiguous()` silently turns stacked `NonTensorData` entries (the string prompts GRPO groups by) into empty TensorDicts, which broke grpo-async on gsm8k for single-turn trajectories. `_concrete_if_possible` now keeps the lazy input when materialization would lose entries; the per-step advantage path also handles same-length groups whose concatenated rewards come back strided. The underlying data-loss bug will be fixed in tensordict separately.
- **[Feature] Enable prefix caching in GRPO configs** - flips `enable_prefix_caching: false -> true` in the 6 GRPO yaml configs after validation.

## Validation

A/B on an H200 node, grpo-async gsm8k, Qwen2.5-1.5B, 3200 dialog turns per arm, two rounds (second round with weight syncs every 2 optim steps to stress the reset path):

| metric | off r1 | on r1 | off r2 | on r2 |
|---|---|---|---|---|
| reward (avg over run) | 0.050 | 0.029 | 0.026 | 0.034 |
| KL to inference (final) | 0.00029 | 0.00028 | 0.00039 | 0.00049 |
| gradient steps/s | 1.385 | 1.369 | 1.293 | 1.295 |
| automatic cache resets | 0 | 2 | 0 | 7 |

No stale-logprob signature (KL to the inference policy is indistinguishable from baseline even with frequent syncs), no reset refusals or warnings, rewards match within seed noise, throughput neutral on this training-dominated smoke. New unit tests cover the launch-path regression, the reset ordering/gating (mock-based, no GPU), the SGLang flush flag resolution, and the MCAdvantage NonTensor regression; the GPU updater suites now run with prefix caching enabled end-to-end.

## Full-scale run with caching enabled

Stock grpo-async gsm8k on Qwen2.5-3B (8xH200, 100k dialog turns, 3,125 optim batches, ~30 min wall clock) with the flipped config, on post-#1746 tensordict main ([W&B run](https://wandb.ai/vmoens/grpo-async/runs/k6atjz3c)). Reward 0.201 -> 0.532; 40 weight syncs, each followed by an automatic prefix-cache reset, none refused; zero errors.

Mean trajectory reward, raw and EMA-smoothed, climbing from 0.201 to 0.532 over 3125 gradient steps

KL to the inference policy stays at ~1e-4 across all 40 syncs (one transient self-recovering spike at step 830); throughput and sequence-length statistics are steady throughout:

Run health panel: KL to inference policy on log scale flat at 1e-4, mean sequence length, gradient steps per second, and objective loss over training

### DIFF
--- a/sota-implementations/grpo/config/grpo_countdown.yaml
+++ b/sota-implementations/grpo/config/grpo_countdown.yaml
@@ -61,7 +61,7 @@ inference_model:
   max_tokens: 512
   include_stop_str_in_output: true
   enforce_eager: false
-  enable_prefix_caching: false
+  enable_prefix_caching: true  # safe with online updates: caches are reset on weight sync
 
 ref_model:
   gradient_checkpointing: false
--- a/sota-implementations/grpo/config/grpo_gsm8k.yaml
+++ b/sota-implementations/grpo/config/grpo_gsm8k.yaml
@@ -104,7 +104,7 @@ inference_model:
   max_tokens: 1024
   include_stop_str_in_output: true
   enforce_eager: false
-  enable_prefix_caching: false
+  enable_prefix_caching: true  # safe with online updates: caches are reset on weight sync
   # compile: false  # Whether to enable model compilation (disabled by default for GRPO)
   # verbose: true   # Whether to show vLLM throughput statistics (enabled by default)
 
--- a/sota-implementations/grpo/config/grpo_ifeval.yaml
+++ b/sota-implementations/grpo/config/grpo_ifeval.yaml
@@ -103,7 +103,7 @@ inference_model:
   max_tokens: 2048
   include_stop_str_in_output: true
   enforce_eager: false
-  enable_prefix_caching: false
+  enable_prefix_caching: true  # safe with online updates: caches are reset on weight sync
   # compile: false  # Whether to enable model compilation (disabled by default for GRPO)
   # verbose: true   # Whether to show vLLM throughput statistics (enabled by default)
 
--- a/sota-implementations/grpo/config/grpo_math.yaml
+++ b/sota-implementations/grpo/config/grpo_math.yaml
@@ -63,7 +63,7 @@ inference_model:
   max_tokens: 2048
   include_stop_str_in_output: true
   enforce_eager: false
-  enable_prefix_caching: false
+  enable_prefix_caching: true  # safe with online updates: caches are reset on weight sync
   max_model_len: 4096
 
 ref_model:
--- a/sota-implementations/grpo/config/grpo_math_7b.yaml
+++ b/sota-implementations/grpo/config/grpo_math_7b.yaml
@@ -63,7 +63,7 @@ inference_model:
   max_tokens: 2048
   include_stop_str_in_output: true
   enforce_eager: false
-  enable_prefix_caching: false
+  enable_prefix_caching: true  # safe with online updates: caches are reset on weight sync
 
 ref_model:
   gradient_checkpointing: false
--- a/sota-implementations/grpo/config/grpo_math_7b_sync.yaml
+++ b/sota-implementations/grpo/config/grpo_math_7b_sync.yaml
@@ -62,7 +62,7 @@ inference_model:
   max_tokens: 2048
   include_stop_str_in_output: true
   enforce_eager: false
-  enable_prefix_caching: false
+  enable_prefix_caching: true  # safe with online updates: caches are reset on weight sync
 
 ref_model:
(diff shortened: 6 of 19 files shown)


vmoens:
pytorch/tensordict#1746 (merged) fixes the underlying contiguous() NonTensorData data loss. The MCAdvantage guard here is kept for compatibility with the tensordict 0.13.0 release, which still has the bug. Running the MCAdvantage suite against post-fix tensordict main exposed one more latent issue: groups of individually-uniform but cross-trajectory-ragged trajectories now materialize per trajectory and torch.cat raised on the ragged leaves; fixed in 2d3e6994ae with a lazy-concatenation fallback. The suite passes against both tensordict main (post-fix) and the locked pre-fix build.

vmoens:
Additional full-scale evidence beyond the A/B in the description: a stock grpo-async gsm8k run on Qwen2.5-3B (8xH200, 100k dialog turns, 3125 optim batches, ~30 min) with prefix caching enabled via the flipped config, on post-fix tensordict main. Reward climbed 0.201 -> 0.532; 40 weight syncs each followed by an automatic prefix-cache reset (none refused); KL to the inference policy flat at ~1e-4 apart from one transient self-recovering spike; zero errors. W&B: https://wandb.ai/vmoens/grpo-async/runs/k6atjz3c

vmoens:
All three review findings addressed:

1. **SGLang retract vs flush (P1)** - fixed in 309af05e23. `flush_cache_on_update` now defaults to `None` (auto): the flush is sent exactly when the pause mode is `abort`, the only mode that leaves the scheduler idle, so the server-side assert can no longer fire after weights have landed. Explicit `flush_cache_on_update=True` requires `pause_mode="abort"` and raises otherwise (fail-fast at scheme/transport construction); an env-forced flush under retract/in_place is downgraded with a one-time warning that stale prefixes persist. `pause_mode` is now a first-class scheme argument, and the GRPO scripts warn when the radix cache is enabled without the abort pause mode. Net effect: no configuration can pair a trusted flush with a pause mode that cannot honor it.

2. **Double-buffer path (P1)** - fixed in a2c9a4b212. `VLLMDoubleBufferWeightReceiver.apply_weights` now resets the prefix cache after both the RPC and direct load branches. This path has no scheduling barrier, so a hard success requirement would deadlock against in-flight requests; instead the reset is retried once on a partial clear and a warning is emitted if blocks are still held (consistent with the NCCL path's reporting). Covered by a CPU-only unit test.

3. **KV connector caches (P2)** - fixed in a5bde39270. All reset paths (async engine, sync Ray worker, local wrapper) now call `reset_prefix_cache(reset_connector=True)`, falling back to the bare call on vLLM versions without the kwarg.

The updated flush semantics are covered by new CPU-only tests (payload resolution per pause mode, fail-fast validation, env downgrade warning).

PR #4027, opened 2026-07-16
[BugFix] Stage collector weight sync through CPU when device sharing …

### LINKED ISSUE #3985: [BUG] MultiCollector zeroes parameters/buffers of ClipPPOLoss when initialized with GPU modules on Windows
## Description

When initializing a MultiCollector, some internal values of modules that reference the actor/critic are unexpectedly reset to zero. This only happens if the actor, critic, and the modules (ClipPPOLoss, GAE) are already on the GPU before creating the MultiCollector.

I was able to reproduce this consistently on Windows 11, but I could not reproduce it on Linux.

## Expected behavior

Initializing a MultiCollector should not modify the state of existing modules such as ClipPPOLoss or GAE.

## Observed behavior

Image

## Workaround

A workaround is to initialize MultiCollector first, and only afterwards either:

create the ClipPPOLoss/GAE, or
move them to the GPU.

This avoids the issue, although I could not find any documentation mentioning this requirement.

## To Reproduce

```python
import torchrl
from torchrl.modules import MLP, ValueOperator, TanhNormal
from torchrl.modules.tensordict_module import ProbabilisticActor
from torchrl.objectives import ClipPPOLoss
from torchrl.objectives.value import GAE
from torchrl.collectors import MultiCollector

import tensordict
from tensordict.nn import TensorDictModule, TensorDictSequential, NormalParamExtractor, TensorDictModuleBase

import torch

from torchrl.envs import GymEnv

from functools import partial

import torch.multiprocessing as mp



if __name__ == "__main__":


    env_factory = partial(GymEnv,"Pendulum-v1")

    device = 'cuda'
    nb_env = 4

    actor = ProbabilisticActor(
        module=TensorDictSequential(
            TensorDictModule(
                MLP(in_features=3, out_features=128, depth=2, num_cells=32),
                in_keys=["observation"],
                out_keys=["embedding"],
            ),
            TensorDictModule(
                MLP(in_features=128, out_features=2, depth=2, num_cells=32),
                in_keys=['embedding'],
                out_keys=['params'],
            ),
            TensorDictModule(
                NormalParamExtractor(),
                in_keys=['params'],
                out_keys=['loc', 'scale'],
            ),
        ),
        in_keys=['loc', 'scale'],
        distribution_class=TanhNormal,
        distribution_kwargs={'low': -2.0, 'high': 2.0},
        return_log_prob=True,
        cache_dist=True,
    )


    critic = ValueOperator(
        module=TensorDictSequential(
            TensorDictModule(MLP(in_features=3, out_features=128, depth=2, num_cells=32), in_keys=["observation"], out_keys=["embedding"]),
            MLP(in_features=128, out_features=1, depth=2, num_cells=32) 
        ),
        in_keys=["observation"],
    )

    loss = ClipPPOLoss(
            actor_network=actor,
            critic_network=critic,
            clip_epsilon=0.2,
            entropy_bonus=True,
            entropy_coeff=1e-3,
            critic_coeff=1.0,
            loss_critic_type="smooth_l1",
            normalize_advantage=True,
        )

    advantage = GAE(
            gamma=0.99,
            lmbda=0.95,
            value_network=critic,
            average_gae=True,
        )

    advantage = advantage.to(device)
    loss = loss.to(device)
    actor = actor.to(device)
    critic = critic.to(device)


    print('Before collector :')
    print('     loss clip epsilon =', loss.clip_epsilon)
    print('     advantage gamma =', advantage.gamma)   

    collector = MultiCollector(
        create_env_fn=[env_factory] * nb_env,
        sync=True,
        policy = actor,
        frames_per_batch=nb_env*10,
        total_frames=10000,
        policy_device=device,
        storing_device="cpu",
        env_device= 'cpu',
    )

    #advantage = advantage.to(device)
    #loss = loss.to(device)

    print('After collector :')
    print('     loss clip epsilon =', loss.clip_epsilon)
    print('     advantage gamma =', advantage.gamma)   


    collector.shutdown()
```

## System info

OS: Windows 11
PyTorch: 2.12.0+cu126
TorchRL: 0.13.2
TensorDict: 0.13.0

theap06:
## Summary

- Fixes #3985 
The fix. Weights bound for a device that can't cross process boundaries are now staged through CPU shared memory and moved back to the policy device inside each worker, with a UserWarning about the extra host-device copy:

### DIFF
--- a/test/test_collectors.py
+++ b/test/test_collectors.py
@@ -62,8 +62,10 @@ from torchrl.collectors._multi_base import MultiCollector
 from torchrl.collectors.distributed.ray import _has_ray, RayCollector
 
 from torchrl.collectors.utils import (
+    _device_shareable_across_processes,
     _make_policy_factory,
     _maybe_normalize_replay_buffer_tensordict_device,
+    _stage_unshareable_weights_on_cpu,
     _traj_chunk_ends_done,
     _traj_emit,
     _traj_ingest,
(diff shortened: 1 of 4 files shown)


vmoens (commented):
The fix looks right, but I’d simplify the tests.

The important missing test is the actual #3985 regression: on Windows+CUDA, constructing/using MultiCollector must not mutate the parent policy, and workers must receive the correct weights.

I’d remove tests that mostly mock helpers and assert their mocked behavior, especially test_cuda_shareable_iff_cuda_ipc and the current test_stage_unshareable_weights_on_cpu, which doesn’t stage anything.

I’d prefer: one real Windows+CUDA regression test, one unit test for the CPU fallback, and optionally the MPS integration test.

vmoens (approved):

PR #4029, opened 2026-07-16
[Performance] Avoid parent env construction in ParallelEnv startup

vmoens:
## Summary

- use one homogeneous VLA environment factory with per-worker `create_env_kwargs`, preserving task, instruction, group, worker-offset, seed, and init-state assignment
- add opt-in `ParallelEnv(metadata_from_workers=True)` startup so real workers provide metadata before initialization and the parent never creates shadow environments
- validate worker schemas, report startup failures, clean up partially started workers, and reject the unsupported shared-buffer combination explicitly
- always close temporary environments created by ordinary metadata callables while preserving `EnvCreator` and instantiated-environment ownership
- enable worker-originated metadata for the LIBERO recipe and document the generic API and production benchmark workflow

## Why

A sequence of distinct per-worker callables makes `ParallelEnv` treat a homogeneous LIBERO workload as heterogeneous. Metadata discovery then constructs every environment serially in the subcollector parent before constructing the same environments again in the real workers. At 320 environments this creates 320 unnecessary MuJoCo/EGL environments and contaminates the parent with EGL state.

The worker metadata handshake removes that shadow wave while keeping the default `ParallelEnv` behavior unchanged. The initial implementation deliberately uses direct pipe communication; `use_buffers=True` fails clearly rather than selecting an unsafe path.

## Benchmark

Production LIBERO was measured on the 5 subcollector x 64 environment topology. The verified worker-metadata/spawn run produced:

- parent metadata constructions: 0
- real worker constructions: 320
- all subcollectors ready: 367.93 s
- first reset and random step complete: 458.73 s
- peak descendants: 329
- outer exit codes: `[0, 0, 0, 0, 0]`
- remaining environment processes: none
- lingering worker processes: none
- instructions: 64 per subcollector with 8 distinct task instructions and the expected group-id blocks

A controlled one-subcollector/64-environment comparison on the same node was:

| Mode | Parent metadata envs | Real envs | All ready | First batch |
| legacy independent callables, spawn | 64 | 64 | 168.53 s | 183.52 s |
| homogeneous factory, spawn | 1 | 64 | 43.74 s | 59.05 s |
| worker metadata, spawn | 0 | 64 | 39.77 s | 54.19 s |
| worker metadata, forkserver | 0 | 64 | 39.42 s | 54.29 s |

Worker metadata reduced readiness latency by 4.24x and first-batch latency by 3.39x versus the legacy path. Forkserver was not materially faster than spawn in the paired run, so the recipe keeps spawn as the safer default. Outer subcollectors also remain on spawn.

Complete commands, output, construction markers, and JSON summaries are stored under:

```text
/root/artifacts/libero_parallel_env_cold_start/5ab5a17f60/
```

The aggregate is `comparison.json`; each run directory contains `command.txt`, `run.log`, and `summary.json`.

## Validation

- `uv run pytest -q test/envs/test_parallel.py` (83 passed, 66 skipped)
- `uv run pytest -q sota-implementations/vla_grpo/test_openvla.py` (14 passed, 22 skipped for unavailable optional dependencies)
- `uv run pytest -q sota-implementations/vla_grpo/test_openvla.py -k 'factory or instruction or group'` (10 passed)
- focused worker-metadata startup, failure, compatibility, and cleanup tests (7 passed)
- pre-commit hooks on all changed files
- production 320-environment LIBERO startup, first step, and clean shutdown

### DIFF
--- a/torchrl/envs/env_creator.py
+++ b/torchrl/envs/env_creator.py
@@ -260,9 +260,12 @@ def get_env_metadata(
         if kwargs is None:
             kwargs = {}
         env = env_or_creator(**kwargs)
-        if env_validator is not None:
-            env_validator(env)
-        return EnvMetaData.metadata_from_env(env)
+        try:
+            if env_validator is not None:
+                env_validator(env)
+            return EnvMetaData.metadata_from_env(env)
+        finally:
+            env.close()
     elif isinstance(env_or_creator, EnvCreator):
         if not (
             kwargs == env_or_creator.create_env_kwargs
--- a/benchmarks/test_envs_benchmark.py
+++ b/benchmarks/test_envs_benchmark.py
@@ -3,6 +3,8 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 import argparse
+import time
+from functools import partial
 
 import pytest
 import torch
@@ -12,6 +14,30 @@ from torchrl.envs import ParallelEnv, SerialEnv, step_mdp, StepCounter, Transfor
 from torchrl.envs.libs.dm_control import DMControlEnv
 from torchrl.envs.libs.libero import _has_libero, LiberoEnv
 from torchrl.envs.transforms.functional import cat_frames
+from torchrl.testing.mocking_classes import CountingEnv
+
+
+class _SlowStartEnv(CountingEnv):
+    def __init__(self, delay: float = 0.1):
+        time.sleep(delay)
+        super().__init__()
+
+
+def _make_slow_start_env(delay: float) -> _SlowStartEnv:
+    return _SlowStartEnv(delay)
+
+
+def _run_parallel_cold_start(metadata_from_workers: bool) -> None:
+    env = ParallelEnv(
+        4,
+        partial(_make_slow_start_env, 0.1),
+        mp_start_method="spawn",
+        metadata_from_workers=metadata_from_workers,
+    )
+    try:
+        env.reset()
+    finally:
+        env.close(raise_if_closed=False)
 
 
 def make_simple_env():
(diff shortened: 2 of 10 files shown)


theap06 (changes requested):
The ParallelEnv in batched_env throws a TypeError at the start. The metaclass converts single-worker ParallelEnvs to SerialEnv but forwards the new flag, which SerialEnv rejects.

vmoens:
It's weird that our tests did not catch this, the SerialEnv fallback is used in the tests for sure no?

theap06:
@vmoens The fallback is covered, but never together with the new flag: test_serial_for_single predates this PR. The flag only reaches SerialEnv on the num_workers==1 && serial_for_single=True early-return in _PEnvMeta.__call__, which pops serial_for_single but not metadata_from_workers before return SerialEnv(*args, **kwargs).  so the failing line is never executed in CI. minor nitpick tbh

vmoens:
Reviewed the rebased head 2b2c6a0a0e22 against the latest requested-changes review and follow-up discussion. The reported fallback bug is still present at torchrl/envs/batched_envs.py:252-261: when serial_for_single=True and num_workers=1, _PEnvMeta removes mp_start_method but forwards metadata_from_workers to SerialEnv. A direct reproduction with ParallelEnv(1, CountingEnv, serial_for_single=True, metadata_from_workers=True, use_buffers=False) raises TypeError: metadata_from_workers=True is only supported by ParallelEnv. Please remove metadata_from_workers before returning SerialEnv and add a regression test combining these flags. Until that is fixed, the opt-in API breaks an existing supported fallback path; the freshly triggered CI is still running.

torchrl/envs/env_creator.py
```diff
@@ -260,9 +260,12 @@ def get_env_metadata(
...
         env = env_or_creator(**kwargs)
-        if env_validator is not None:
-            env_validator(env)
-        return EnvMetaData.metadata_from_env(env)
+        try:
+            if env_validator is not None:
+                env_validator(env)
+            return EnvMetaData.metadata_from_env(env)
+        finally:
+            env.close()
```
gtnv, line 268: metadata failed first in my probe. close() then failed and replaced that error. can we keep the metadata error as the primary failure?

vmoens:
Close as superseded. #4028 and #4093 already landed the worker-metadata design and fixes; #4029 retains a confirmed serial_for_single fallback bug.

PR #4030, opened 2026-07-17
[BugFix] Persist write_count in RoundRobinWriter checkpoints

theap06:
## Description

`RoundRobinWriter.state_dict()` and `dumps()` only persist the cursor, so `ReplayBuffer.write_count` silently resets to zero after a checkpoint restore, even though the docstring describes it as the total number of items ever written.

This PR persists `_write_count`:

- through `state_dict()` / `load_state_dict()`;
- through `dumps()` / `loads()` metadata (for both `RoundRobinWriter` and `TensorDictMaxValueWriter`).

Loading remains tolerant of checkpoints produced before this change: when the key is absent, the current counter is left untouched instead of raising.

## Motivation

`write_count` is the natural cumulative progress counter of a replay buffer and is increasingly used for scheduling and monitoring decisions. A counter that silently rewinds on resume makes any such logic misbehave and makes a counter decrease indistinguishable from an intentional `empty()`.

### DIFF
--- a/test/rb/test_writers.py
+++ b/test/rb/test_writers.py
@@ -13,6 +13,7 @@ from torch import multiprocessing as mp
 
 from torchrl.data import (
     PrioritizedReplayBuffer,
+    ReplayBuffer,
     TensorDictPrioritizedReplayBuffer,
     TensorDictReplayBuffer,
 )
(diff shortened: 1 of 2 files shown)


vmoens (approved):
Thanks!

PR #4031, opened 2026-07-17
[Feature] Add ReplayBuffer.stats() for lightweight monitoring

theap06:
## Description

First implementation step of the collector/replay-buffer monitoring RFC: a cheap, side-effect-free, serializable `stats()` snapshot on `ReplayBuffer`.

```python
>>> rb.stats()
{'size': 5, 'write_count': 5, 'prefetch_queue_size': 0, 'initialized': True, 'capacity': 10, 'utilization': 0.5}
```

Contract:

- scalar counters and gauges only, never storage content or `state_dict()`-style state;
- side-effect free and safe to call concurrently with writes and samples (snapshot taken under `_replay_lock`);
- calling it on a delayed-init buffer does not trigger initialization (returns an empty snapshot with `initialized=False`);
- JSON-serializable, directly consumable by `Logger.log_metrics()`;
- cumulative counters (`write_count`) are meant to be turned into rates by an external monitor.

Remote implementations:

- `RayReplayBuffer.stats()` performs a single actor round-trip for the whole snapshot instead of one RPC per attribute (the existing per-property pattern);
- the distributed transport client maps its control-channel stats to the same dictionary contract.

### DIFF
--- a/test/rb/test_rb_core.py
+++ b/test/rb/test_rb_core.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import argparse
 import contextlib
 import functools
+import json
 
 import pytest
 import torch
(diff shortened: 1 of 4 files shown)


vmoens (approved):
LGTM thanks!

PR #4033, opened 2026-07-17
[Feature] Add LoggerMonitor and Every for pull-based operational monitoring

theap06:
## Description

Core deliverable of the collector/replay-buffer monitoring RFC: `torchrl.record.loggers.monitoring` with `LoggerMonitor` and `Every`, letting one TorchRL logger observe any number of collectors and replay buffers (local, multiprocessing or Ray) without putting logging work on collection or sampling hot paths.

```python
logger = WandbLogger(exp_name="experiment", project="torchrl")

with LoggerMonitor(logger, poll_interval=1.0) as monitor:
    monitor.watch(collector, name="collector", schedule=Every.counter("frames", 10_000))
    monitor.watch(replay_buffer, name="replay_buffer", schedule=Every.seconds(5), step="write_count")
    collector.start()
    run_training()
```

Design (source / monitor / sink separation from the RFC):

- **Pull-based**: snapshots are requested from the monitor's own daemon thread, or through explicit `monitor.step()` calls (`background=False`) for deterministic tests. Watched objects never call the logger; a slow backend can only delay the next poll, never build an unbounded backlog.
- **Schedules**: `Every.seconds(x)` (wall clock) and `Every.counter(key, n)` (logical counters). Counter thresholds are observed at poll time; a jump across several thresholds logs the latest snapshot once; a counter decrease (reset, `empty()`, state restore) re-baselines instead of producing spurious logs or negative rates.
- **Rates**: derived by the monitor from cumulative counter deltas between logged snapshots (`frames_per_second`, `writes_per_second`, ...), so sources need no timers or rolling windows.
- **Namespacing and steps**: metrics log as `<name>/<metric>`; the per-watch `step` key (defaulting to the counter key for counter schedules) gives each namespace an independent logical step, matching the per-prefix step handling in `WandbLogger`.
- **Lifecycle**: the monitor owns neither the logger nor the targets. `stop()` logs final snapshots (`log_on_close`) and joins the thread; logger shutdown stays explicit. `on_error="warn"|"raise"|"ignore"` controls failure behavior.
- **Custom providers**: `watch(..., stats_fn=..., stats_kwargs=..., rate_keys=...)` for driver-side custom gauges and per-worker views.

## Notes

Stacked on #4030, #4031 and #4032 (their commits are included here; only the last commit is new). I will rebase once those land. Docs added to `trainers_loggers.rst`. Follow-ups per the RFC's phasing: compile-safe sample counters and richer sampler stats, source-side custom providers for remote objects, and composable event observers for batch-derived metrics.

### DIFF
--- a/docs/source/reference/trainers_loggers.rst
+++ b/docs/source/reference/trainers_loggers.rst
@@ -45,6 +45,40 @@ backpressure when several clients log concurrently.
     get_logger
     generate_exp_name
 
+Monitoring collectors and replay buffers
+----------------------------------------
+
+Any collector or replay buffer exposes a cheap
+:meth:`~torchrl.collectors.BaseCollector.stats` /
+:meth:`~torchrl.data.ReplayBuffer.stats` snapshot of operational counters
+(frames collected, buffer size, write count, worker liveness, ...). A
+:class:`~torchrl.record.loggers.monitoring.LoggerMonitor` periodically pulls
+those snapshots on a wall-clock or counter
+(:class:`~torchrl.record.loggers.monitoring.Every`) schedule, derives rates
+such as frames per second, and forwards namespaced metrics to any logger,
+without adding work to collection or sampling hot paths. This works with
+local, multiprocessing and Ray implementations alike.
+
+.. code-block:: python
+
+    from torchrl.record import WandbLogger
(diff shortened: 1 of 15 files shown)


vmoens (approved):
LGTM thanks!

Issue #4039, opened 2026-07-22
[Feature Request] Separate replay selection distributions from sampling units and boundary policies


vmoens:
## Motivation

Transition sampling and trajectory-slice sampling currently expose different sampler classes even
though they combine several orthogonal decisions:

1. how an anchor is selected and weighted;
2. how that anchor expands into a transition, sequence or complete trajectory;
3. how episode and stored-segment boundaries are handled;
4. how short/ragged output is represented.

This makes combinations such as prioritized sequence starts, recurrent burn-in, n-step bootstrap
suffixes and alternative padding policies harder to express consistently. It also makes individual
samplers responsible for both probability distributions and trajectory-range mechanics.

## Proposal

Introduce an optional composition layer with three responsibilities:

- **Anchor sampler**: selects record/trajectory anchors and returns probabilities or weights.
- **Sampling unit / range resolver**: expands anchors into transitions, fixed sequences, n-step
  transitions or complete trajectories.
- **Boundary/output policy**: controls episode/rollout crossing, padding, masks and output layout.

An illustrative API is:

```python
rb = ReplayBuffer(
    storage=storage,
    sampler=PrioritizedSampler(...),
    sample_unit=Sequence(
        learn_length=32,
        burn_in=8,
        bootstrap=1,
        episode_boundary="stop",
        rollout_boundary="cross",
        short_sequence="pad",
        layout="flat",  # preserve current TorchRL-friendly default
    ),
)
```

The exact API need not use these names. The important property is that the distribution chooses
anchors while the sampling unit owns range expansion and boundary enforcement.

## Required semantics

- Transition, fixed-sequence and complete-trajectory units.
- Recurrent burn-in distinct from the learning portion.
- Optional n-step/bootstrap suffix.
- Configurable stride/overlap where meaningful.
- Episode boundary policies such as stop, pad or include-reset.
- Collector-rollout boundary policies independent of episode boundaries.
- Validity and learning masks for padded/ragged data.
- Clearly defined priority/importance-weight semantics: per anchor, per sequence or expanded per
  transition.
- Both current flat concatenated output and an optional structured `[B, T]` index/data layout.

## Backward compatibility

- Existing `RandomSampler`, `PrioritizedSampler`, `SliceSampler` and
  `PrioritizedSliceSampler` APIs remain supported.
- Existing slice samplers can initially be implemented as adapters over the new decomposition, or
  the new API can be introduced alongside them.
[...]

PR #4045, opened 2026-07-24
[Feature] Add SampleUnit composition point for replay sampling

theap06:
## Description

PR 1 of the split proposed in #4039: the interface layer of the sampler decomposition. Replay sampling combines two orthogonal decisions — which anchors the sampler's distribution selects, and what each anchor expands into. This PR adds the composition point for the second decision without changing any behavior.

- `SampleUnit` abstract base: `expand(index, info, storage)` runs inside the buffer's sampling critical section, after the anchor sampler and before the storage read or any index bookkeeping, and returns the expanded index plus (possibly augmented) info. Units own the alignment of per-anchor info entries they change; metadata they add flows into `sample(return_info=True)` and becomes keys of TensorDict samples through the existing info-copy path.
- `Transition`: the identity unit and implicit default. `sample_unit=None` and `sample_unit=Transition()` are behaviorally identical — pinned by tests under seeded generators, so this PR is a pure no-op for existing users.
- `sample_unit=` keyword on `ReplayBuffer`, threaded through `PrioritizedReplayBuffer` and `TensorDictPrioritizedReplayBuffer`; Hydra config companions gain the matching field.

This settles the RFC's first open question (units are buffer-owned, matching the illustrative API in the issue) and gives the follow-ups from the #4039 split — the `Sequence` unit with boundary policies, burn-in/bootstrap, priority semantics — a stable target to build against.

## Testing

`TestSampleUnit` in `test/rb/test_rb_core.py`: seeded-generator equivalence of default vs `Transition()` (data, info keys, indices), no metadata leakage from the identity unit, a toy expanding unit proving the composition end-to-end (expanded batch, `info["index"]` reflects expansion, metadata surfaces as TensorDict sample keys), constructor type validation, and the prioritized-buffer path including `update_tensordict_priority` after sampling through a unit. Full `test/rb/` suite passes (4189 tests).

Part of #4039.


vmoens (approved):
LGTM thanks!

Issue #14, opened 2022-02-11
[Feature Request] Trainer saving and loading utils


vmoens:
The `Trainer` class supports saving model and environment variables in order for them to be re-used. However this workflow is still experimental. 
Eventually, it should be possible to save and load an agent state and restart training where it was left.

Benjamin-eecs:
Now that `agent` class is gone, I think it is safe to close this issue.

vmoens:
You're right. We should edit it with Trainer instead

Benjamin-eecs:
I think it can be considered done in `trainer` class.

PR #4046, opened 2026-07-24
[Feature] Add generation-stamped replay slots for round-robin writer

harryfrzz:
## Summary

A replay buffer index is a *physical slot number*, not a handle on a piece of
data. A round-robin writer reuses slots, so an index sampled at one point may
name completely different data a moment later — and nothing in the buffer tells
a consumer that happened.

This adds opt-in **generation stamps**: one counter per storage slot, advanced on
every write to that slot. A consumer that captured a stamp when it sampled can
compare it against the current stamp to know whether the slot still holds its
data.

```python
rb = ReplayBuffer(
    storage=LazyTensorStorage(8),
    writer=RoundRobinWriter(track_generations=True),
)
rb.extend(torch.arange(8))
_, info = rb.sample(4, return_info=True)
index, generation = info["index"], info["index_generation"]

rb.extend(torch.arange(8, 11))            # overwrites slots 0, 1, 2
stale = rb.writer.generations_of(index) != generation
# index[stale] no longer holds the sampled data
```

This is the primitive that conditional writes ("update this record only if it is
still the one I read") need; it is useful on its own for async training and for
prioritized replay, where priorities are computed after the forward pass and
written back at an index that may since have been reused.

## Design

**Opt-in.** The default is `track_generations=False`. Enabling it allocates one
`int64` per storage slot and adds an `"index_generation"` entry to the sampler
output — which for tensordict buffers is a new *sample key* — so it is not
imposed on buffers that do not need it. `sample()` output is byte-identical to
before for every existing user.

**The counters live on the storage, not on the writer.** Two buffers sharing one
storage overwrite each other's slots, so a per-writer counter would let one
buffer's handles read as live after the other overwrote them — exactly the
staleness the feature exists to detect. The buffer is attached to the storage
object, and a writer registered against a storage that already has one adopts it
rather than replacing it.

**Semantics.** One stamp per write, not per `extend` call: an `extend` that wraps
the storage advances a reused slot once per write it receives. `-1` means "no
usable stamp" (never written, out of range, or tracking disabled), not
"generation zero". `empty()` *advances* stamps rather than resetting them, so
handles taken before it correctly read as stale. Stamps are independent
per-slot counters and say nothing about write order between different slots.

**Allocation.** Storages small enough to allocate up front get one allocation, so
the buffer's shape never changes and the `torch.compile` extend/sample path does
not recompile. Larger and unbounded storages (`ListStorage` with no `max_size`
reports `torch.iinfo(torch.int64).max`) grow geometrically on demand.

**Process-local, deliberately.** The counters are not shared across processes. The
buffer is replaced rather than mutated when it grows, so a shared mapping would
silently stop tracking after the first growth. Cross-process staleness detection
needs a storage-owned fixed-size mapping and is a follow-up, not a half-working
`share_memory_()` call here.

## Docs

`docs/source/reference/data_replaybuffers.rst` gains a "Detecting overwritten
slots: generation stamps" section (`ref_buffers_generations`) with the motivation,
a runnable example, the semantics above and the implementation notes, cross-
referenced from the Writers section. `Writer.generations_of`,
`Writer.tracks_generations` and `RoundRobinWriter` carry full docstrings.

## Not in this PR

- **Cross-process visibility.** Needs a storage-owned, fixed-size mapping; see
  above.
- **`ReplayBuffer(track_generations=True)` passthrough.** Enabling it means
  passing your own writer today, as with samplers and storages. A buffer-level
  kwarg would need matching fields on `ReplayBufferConfig` /
  `TensorDictReplayBufferConfig` (CLAUDE.md #14) and is better added alongside
  the first consumer.
- **Conditional writes** (`update_if_present` with a version check) build on this
  and are proposed separately.

## Tested

`test/rb`: **4201 passed, 1029 skipped**. Pre-PR baseline on the same merge base:
4175 passed, 1025 skipped — so +26 new tests, +4 CUDA-gated skips, no regressions.
The four pre-existing tests this PR previously had to patch pass unmodified again.
Doctest and docs examples verified to produce the output they claim.

### DIFF
--- a/torchrl/data/replay_buffers/writers.py
+++ b/torchrl/data/replay_buffers/writers.py
@@ -163,13 +163,6 @@ class ImmutableDatasetWriter(Writer):
 class RoundRobinWriter(Writer):
     """A RoundRobin Writer class for composable replay buffers.
 
-    Round-robin writers stamp every physical storage slot with an int64
-    generation counter: the first write of a slot has generation ``0`` and every
-    reuse increments it. Sampling returns the generation observed for each index
-    (see :meth:`generations_of`), so a captured ``(index, generation)`` pair
-    identifies one specific record and becomes detectably stale once its slot is
-    reused or the buffer is emptied.
-
     Args:
         compilable (bool, optional): whether the writer is compilable.
             If ``True``, the writer cannot be shared between multiple processes.
@@ -222,12 +215,6 @@ class RoundRobinWriter(Writer):
             )
 
     def generations_of(self, index: int | torch.Tensor) -> torch.Tensor:
-        """Returns the int64 generation stamp for each physical slot in ``index``.
-
-        The first write of a slot has generation ``0`` and every reuse increments
-        it. For multidimensional storages the index is reduced to its first (slot)
-        dimension. Slots that have never been written report ``-1``.
-        """
         if isinstance(index, tuple):
             index = index[0]
         elif (
--- a/torchrl/data/replay_buffers/writers.py
+++ b/torchrl/data/replay_buffers/writers.py
@@ -212,7 +212,9 @@ class RoundRobinWriter(Writer):
             )
         if self._generation is None:
             raise RuntimeError("No data has been written yet; generation is undefined.")
-        if (
+        if isinstance(index, tuple):
+            index = index[0]
+        elif (
             isinstance(index, torch.Tensor)
             and index.ndim
             and self._storage.ndim > 1
--- a/torchrl/data/replay_buffers/writers.py
+++ b/torchrl/data/replay_buffers/writers.py
@@ -39,6 +39,10 @@ except ImportError:
 from torchrl.data.replay_buffers.storages import Storage
 from torchrl.data.replay_buffers.utils import _is_int, _reduce
 
+# Storage capacities at or above this value are treated as unbounded (lazy
+# storages report a sentinel max size), triggering dynamic generation growth.
+_GENERATION_UNBOUNDED = 2**40
+
 
 class Writer(ABC):
     """A ReplayBuffer base Writer class."""
@@ -50,13 +54,19 @@ class Writer(ABC):
         self._storage = None
         self._compilable = compilable
 
+    #: Whether this writer type stamps storage slots with a reuse generation.
+    tracks_generations: bool = False
+
     def register_storage(self, storage: Storage) -> None:
         self._storage = storage
 
-    @property
-    def supports_generation(self) -> bool:
-        """Whether this writer maintains per-slot generation stamps."""
-        return False
+    def generations_of(self, index: int | torch.Tensor) -> torch.Tensor:
+        """Returns the generation stamp for each physical slot in ``index``.
+
+        Writers that do not track slot reuse report ``-1``.
+        """
+        index = torch.as_tensor(index)
+        return torch.full(index.shape, -1, dtype=torch.int64, device=index.device)
 
     @abstractmethod
     def add(self, data: Any) -> int:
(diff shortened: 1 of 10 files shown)


torchrl/data/replay_buffers/writers.py
```diff
@@ -155,16 +170,92 @@ class RoundRobinWriter(Writer):
...
+        else:
+            index = torch.as_tensor(index, dtype=torch.long).reshape(-1)
+            if index.numel() == 0:
+                return
+            capacity = self._storage._max_size_along_dim0(batched_data=data)
+            min_size = (
+                capacity if capacity < _GENERATION_UNBOUNDED else int(index.max()) + 1
+            )
+            self._ensure_generation(capacity, min_size)
+            self._generation.index_put_(
```
theap06, line 345: On my local test, the _generation is causing crashes on non-cpu storage. This is likely due to the fact that _generation is always allocated on CPU without a device= parameter. The extend() and batched bump it with index_put_ using the storage's own device index tensor. Got the same issue with the CUDA storage.

torchrl/data/replay_buffers/writers.py
```diff
@@ -249,16 +353,27 @@ def _update_storage_len_for_write_at(self, index: int | torch.Tensor) -> None:
...
+        generation = state_dict.get("_generation")
+        if generation is not None:
+            generation = generation.clone()
+            if not self._compilable:
+                generation.share_memory_()
+            self._generation = generation
 
     def _empty(self, empty_write_count: bool = True) -> None:
         self._cursor = 0
+        if self._generation is not None:
```
theap06: self._generation += 1 in _empty() is unconditional, so slots that were never written (sentinel -1) become 0

theap06 (changes requested):
Thanks! Left some feedback.

torchrl/data/replay_buffers/writers.py
harryfrzz: Fixed!, _generation now allocates on the storage device and the index is moved there before index_put_

torchrl/data/replay_buffers/writers.py
```diff
@@ -229,6 +364,7 @@ def write_at(self, index: int | torch.Tensor, data: Any) -> int | torch.Tensor:
             batch_size = index.numel()
         self._write_count += batch_size
         self._storage.set(index, data, set_cursor=False)
+        self._bump_generation(index, data)
```
theap06, line 481: Bumping here means positional writes consume outstanding handles. Update the docstring to account for the new record here.

torchrl/data/replay_buffers/writers.py
```diff
@@ -155,16 +170,123 @@ class RoundRobinWriter(Writer):
...
+                return
+            capacity = self._storage._max_size_along_dim0(batched_data=data)
+            min_size = (
+                capacity if capacity < _GENERATION_UNBOUNDED else int(index.max()) + 1
+            )
+            self._ensure_generation(capacity, min_size, device)
+            index = index.to(device)
+            self._generation.index_put_(
+                (index,), torch.ones_like(index), accumulate=True
+            )
```
theap06, line 347: With accumulate=True, a single extend that wraps the storage bumps a reused slot once per write. update docstring here as well and maybe a test case for this change

theap06 (approved):
Left some comments. Otherwise LGTM! cc @vmoens

harryfrzz:
> Left some comments. Otherwise LGTM! cc @vmoens

sure! i’ve pushed the changes you mentioned @theap06.

harryfrzz:
@theap06 @vmoens any changes that i need to do here?

vmoens (approved):
With [84c1146](https://github.com/pytorch/rl/pull/4046/commits/84c1146c65e2763e7617f497c324335214ee3a3d) I made it optional to enable this - we can think about enabling it by default 2 releases after the next but the plan must be announced via deprecation warnings since this would be bc breaking. 
I also added some doc. 
Happy with this version if you are. Deprecation plan can follow in a follow up PR.

harryfrzz:
> With [84c1146](https://github.com/pytorch/rl/pull/4046/commits/84c1146c65e2763e7617f497c324335214ee3a3d) I made it optional to enable this - we can think about enabling it by default 2 releases after the next but the plan must be announced via deprecation warnings since this would be bc breaking. I also added some doc. Happy with this version if you are. Deprecation plan can follow in a follow up PR.

Happy with it, thanks. Opt-in is the right call and the storage owned counters fix the shared storage case. one thing that i want to point out is that the four CUDA tests I added in cfaa0abb still build default writers and assert on _writer._generation, which is None now that tracking is gated. They skip on CPU and are -m gpu-selected on the GPU job. full test/rb is green here (4332 passed, 898 skipped). Should I add track_generations=True to those four?
deprecation plan in a follow-up works for me!

vmoens:
Yes you are correct the tests had to be corrected which i did in 6d3af1aec3bb114fa9144514a717f9e04f909828

PR #4048, opened 2026-07-25
[Test] Add automated Config/Class kwarg-parity checks for Hydra configs

theap06:
## Summary

CLAUDE.md section 14 requires every kwarg of a wrapped class's `__init__` to
appear as a field on its Hydra `*Config` dataclass (same default), but only
one narrow, hand-written parity test existed (`test_checkpoint_config_parity`,
checking a single field's default across 10 of 212 config classes). This adds
a generic, automatically-discovered check:

- `TestConfigClassParity` walks every leaf `*Config` dataclass reachable from
  `torchrl.trainers.algorithms.configs`, resolves the class or factory named
  by its `_target_` (via the factory's return-type annotation for the
  `_make_*`/`make_*` indirection cases), and asserts every required
  constructor parameter has a same-named Config field.
- **128 classes pass strictly today** and will regress immediately if a
  future kwarg addition to a wrapped class isn't mirrored into its Config —
  closing the exact silent-drift risk CLAUDE.md section 14 calls out.
- **62 classes have a pre-existing gap** and are marked
  `xfail(strict=True)` as a ratchet: fixing one of them without also removing
  it from `_CONFIG_PARITY_KNOWN_GAPS` turns CI red (XPASS), so the allowlist
  can only shrink, never silently grow.
- **5 configs are skipped** with an explicit reason instead of silently
  passing: two wrap a composite/polymorphic factory with no single
  constructor to diff against (`BatchedEnvConfig`, `TanhNormalModelConfig`),
  and three have a `_target_` that references a class which no longer exists
  in the codebase (`StorageEnsembleWriterConfig`, `KLRewardTransformConfig`,
  `LionConfig` — pre-existing breakage, out of scope here, left visible via
  the skip reason rather than swept under a blanket exclusion).

### DIFF
--- a/test/test_configs.py
+++ b/test/test_configs.py
@@ -271,8 +271,22 @@ class TestConfigClassParity:
     user setting that kwarg through Hydra has no effect and the failure is
     silent. This walks every concrete Config dataclass reachable from
     ``torchrl.trainers.algorithms.configs``, resolves the class or factory named
-    by its ``_target_``, and asserts every required constructor parameter has a
-    same-named Config field.
+    by its ``_target_``, and asserts every named constructor parameter --
+    required or optional -- has a same-named Config field.
+
+    Two deliberate limitations, left as follow-ups:
+
+    - Only field-name presence is checked; default-value *equality* between a
+      Config field and the corresponding ``__init__`` kwarg is NOT enforced, so
+      a Config default that drifts from the constructor's default still passes.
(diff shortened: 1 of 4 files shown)


vmoens (approved):
Thanks for the contribution — I made missing-optional-dep configs skip instead of fail and clarified the docstring's stated guarantees in 1038a3e.

Issue #4040, opened 2026-07-22
[Feature Request] Add generation-safe conditional updates for mutable replay fields


vmoens:
## Motivation

Some replay algorithms update stored data after sampling, not only sampler priorities. Examples
include:

- refreshing recurrent hidden states;
- caching embeddings or target-model outputs;
- attaching asynchronous labels or post-processing results;
- updating other explicitly mutable replay fields.

Existing physical-index mutation can target the wrong record if a round-robin writer reuses the
slot between sampling and update. Holding sampled records pinned throughout a learner step would
avoid the race but can create severe replay backpressure.

## Dependency

This should build on generation-stamped replay indices. Priority updates remain sampler metadata
and should use their own conditional API; this issue covers stored replay fields.

## Proposal

Add a best-effort conditional update API for storage implementations that support generation
validation:

```python
result = rb.update_if_present(
    index=sample["index"],
    generation=sample["index_generation"],
    patch={"recurrent_state": refreshed_state},
)
```

The operation validates all requested keys, shapes and dtypes before mutating storage, then updates
only records whose `(index, generation)` is still live. It reports updated and stale elements.

An optional compare-version facility could prevent an older asynchronous computation from
overwriting a newer refresh:

```python
result = rb.update_if_present(
    index=index,
    generation=generation,
    patch=patch,
    version_key="state_model_version",
    version=current_model_version,
    require_newer=True,
)
```

The version feature may be deferred, but the base API should not preclude it.

## Schema and storage contract

- Storages may restrict conditional mutation to fields designated mutable at initialization.
- Structural changes such as adding keys or changing shapes are out of scope.
- Generation validation and each record's patch write are atomic with respect to slot reuse.
- The API documents whether atomicity is per record or across the whole batch. Per-record atomicity
  with a returned mask is likely the useful default.
- Validation failure before writing must not leave a partially applied malformed patch.
- Multiple keys in one record's patch must not become observably torn to concurrent readers.

## Backward compatibility

- Existing `set_at_`, `set_` and `update_` APIs remain unchanged.
- Conditional mutation is opt-in and supported first by compatible tensor storages/writers.
- Unsupported storage backends raise a capability error instead of performing an unsafe raw-index
  update.

## Acceptance criteria

- Live stamped indices update successfully.
- Reused slots are skipped and never modified.
- Mixed live/stale batches return an accurate mask and counts.
- Shape, dtype and key validation happen before mutation.
- Multi-key patches are atomic per record.
[...]

PR #4049, opened 2026-07-25
[Feature] Implement Compare-Version facility for update_if_present

coder-jayp:
## Description
Implements the optional compare-version facility inside `ReplayBuffer.update_if_present`. This allows asynchronous updates to be safely ignored if they carry an older model/state version than what is currently stored in a live Replay Buffer slot.

## Motivation and Context
This prevents an older asynchronous computation from overwriting a newer refresh of mutable replay fields. It guarantees generation-safe and version-safe updates.

Part of #4040. (Builds on top of the base API drafted in #4043).

- [x] I have raised an issue to propose this change (required for new features and bug fixes)

## Types of changes

## Checklist
- [x] I have read the CONTRIBUTION guide (**required**)

### DIFF
--- a/benchmarks/test_replaybuffer_benchmark.py
+++ b/benchmarks/test_replaybuffer_benchmark.py
@@ -402,6 +402,35 @@ def test_rb_populate(benchmark, rb, storage, sampler, size):
     )
 
 
+class create_wraparound_rb:
+    """Builds a full buffer so every timed extend reuses slots and bumps generations."""
+
+    def __init__(self, size=10_000, batch=1_000):
+        self.size = size
+        self.batch = batch
+
+    def __call__(self):
+        rb = ReplayBuffer(storage=LazyTensorStorage(self.size))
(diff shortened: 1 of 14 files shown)


vmoens:
Reviewed the rebased head 407ae4ed88d0, including the latest hardening commit. The current implementation normalizes nested key spellings, reads/writes full coordinates for multidimensional storage, resolves duplicate handles by highest incoming version with last-row tie-breaking, and keeps stale/version-rejected masks disjoint even for an all-stale call. I ran the focused core suite locally: 29 update_if_present tests passed, including concurrency, duplicates, multidimensional storage, nested keys, and checkpoint round trips. The Ray delegation test was collected but skipped because Ray is unavailable here; CUDA/MPS device paths were also not exercised locally. I did not find a blocking issue in the current diff, with those distributed/device paths as the remaining validation risk.

coder-jayp:
@vmoens Thanks for the thorough review and for adding that final hardening commit on top to handle the multidimensional and distributed edge cases

vmoens (approved):
LGTM thanks!

PR #4050, opened 2026-07-25
[Feature] Sequence sample unit with exact boundary policies

coder-jayp:
## Description

Implements the `Sequence` sampling unit to expand sampled anchors into fixed-length sequences. 

**Key Features:**
- **Robust Boundary Recovery**: Leverages `utils.find_start_stop_traj` to accurately map anchors to their trajectory endpoints, cleanly handling ring buffer seams and `storage._last_cursor` truncations.
- **Strict Boundary Policies**: 
  - `include_reset`: Blindly sweeps forward crossing boundaries.
  - `pad`: Clamps the sequence at the episode termination and masks trailing steps as invalid.
  - `stop`: Shifts the anchor backward so the sequence ends exactly on the boundary (falls back to `pad` if the episode is shorter than the sequence length).
- **Metadata Flow**: Expands existing per-anchor info and injects accurate `sequence_id`, `step_in_sequence`, and `validity_mask` for every frame in the `B * length` output.

## Motivation and Context

This change separates the trajectory-range mechanics (sequence generation) from the anchor probability distributions (like Prioritized or Random Sampling), solving the problem where individual samplers previously had to handle both. It enables combinations such as prioritized sequence starts and consistent padding/boundary policies.

Addresses piece 2 of #4039

- [x] I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

- [x] New feature (non-breaking change which adds core functionality)

## Checklist

- [x] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.

PR #4051, opened 2026-07-25
[Feature] Extend Sequence with burn-in, bootstrap, dilation and per-anchor priority semantics

theap06:
## Description

Extends the `Sequence` sample unit from #4050 with recurrent-training context and anchor-based priority semantics.

### Burn-in, bootstrap, and dilation

Each sampled anchor expands into `burn_in + length + bootstrap` records. The learning region starts at the sampled anchor; burn-in records precede it and bootstrap records follow it. `dilation` spaces records within the window and does not control spacing or overlap between independently sampled windows.

The returned metadata includes:

- `learning_mask`: true only for the learning region;
- `validity_mask`: false for padded records;
- `anchor_index`: the original sampled anchor repeated across its expanded records;
- `sequence_id` and `step_in_sequence`: flat-batch grouping metadata.

A recurrent learner should replay the full window to reconstruct state and compute loss under `learning_mask & validity_mask`.

`episode_boundary="pad"` clamps records at trajectory boundaries and marks padding invalid. `"stop"` shifts the effective window backward when possible. `"include_reset"` may cross episode boundaries but cannot cross the replay write seam or enter unwritten slots. Defaults preserve the previous `Sequence(length=...)` behavior.

### Priority semantics

Priorities remain attached to sampled anchors. `TensorDictReplayBuffer.update_tensordict_priority` groups expanded per-record priorities by `anchor_index`, excludes records masked by `validity_mask`, max-reduces duplicate anchors, and updates only anchor slots. Per-anchor sampler metadata such as importance weights expands blockwise without changing the anchor sampling distribution.

### Documentation and tests

The reference documentation includes a recurrent Q-learning example and explains how `learning_mask` and `validity_mask` are consumed. Tests cover burn-in, bootstrap, dilation, all boundary policies, partial and full rings, wraparound, duplicate anchors, priority routing, distribution invariance, nested done keys, and config parity.

The focused Sequence suite passes on CPU and NVIDIA GB10 (`36 passed, 447 deselected`). The original full replay-buffer run passed 4,212 tests. CI failures shown on the merged PR were classified as unrelated flaky or broken-trunk failures; GPU, wheel, docs, lint, and targeted jobs passed.

## Post-merge review and follow-ups

Post-merge profiling used recurrent Q-learning replay with one million records, average episode length 256, 256 anchors, `burn_in=40`, `length=64`, `bootstrap=5`, and `dilation=1`.

On NVIDIA GB10:

| Operation | Median |
| --- | ---: |
| Current `Sequence.expand()` | 0.510 ms |
| Current boundary scan | 0.300 ms |
| Hot compressed boundary lookup and expansion | 0.097 ms |
| Compiled hot compressed lookup and expansion | 0.024 ms |
| Full compressed-index rebuild | 0.450 ms |
| Dense-map rebuild | 123.205 ms |

The compressed cache used 62,528 bytes versus 16,000,000 bytes for the dense map. The compressed design is the preferred default; a direct sorted-stop path is preferable for one-dimensional storage, with lane-encoded search reserved for multidimensional storage.

The follow-up work is intentionally split into reviewable PRs:

1. `[Refactor] Centralize replay trajectory-boundary queries`: introduce one internal provider shared by `Sequence`, `SliceSampler`, and other trajectory-aware consumers; establish storage mutation revisions, cache lifecycle, ring-wrap behavior, device placement, and same-process/shared-process invalidation.
2. `[Performance] Cache and compile replay-boundary queries`: add the compressed cache, the specialized one-dimensional stop lookup, the pure compile-friendly tensor kernel, and cold/hot/write-invalidated benchmarks for `Sequence` and `SliceSampler`.
3. `[Feature] Support multidimensional storage in Sequence sample units`: expand only the time coordinate, preserve lane coordinates, return multidimensional `anchor_index`, and max-reduce priorities by full anchor coordinate after validity masking.

Part of #4039.


vmoens:
Thanks for the contribution — I routed priority updates through anchors (max over valid records), kept burn-in inside the ring seam, and documented index/anchor_index semantics in b719832.

vmoens:
Thanks @theap06!

I think that performance and multidimensional-storage support can be handled in follow-up PRs rather than blocking this one, so I'll leave it as TODOs for you (please let me know if you'd rather have someone else do it).

I pushed the missing documentation, including a recurrent Q-learning use case and an explanation of how learning_mask and validity_mask should be consumed. I also renamed stride to dilation: the parameter spaces records inside a window and does not control spacing or overlap between independently sampled windows. The PR title/body should be updated accordingly.

For follow-ups, I suggest:

- Design shared, efficient replay-boundary detection used by Sequence, SliceSampler, and other trajectory-aware components, including caching/invalidation, ring wraparound, device placement, and benchmarks.
- Add ndim > 1 support by expanding the time coordinate while preserving lane coordinates, including multidimensional anchor-priority reduction.

The boundary-detection work should be holistic rather than a local optimization of Sequence (we have that problem in many places in torchrl). Compile can help, it worked nicely with SliceSampler back in the days IIRC.

That's a high difficulty task so I'm happy to take it over.

vmoens (approved):
Thanks @theap06!

I think that performance and multidimensional-storage support can be handled in follow-up PRs rather than blocking this one, so I'll leave it as TODOs for you (please let me know if you'd rather have someone else do it).

I pushed the missing documentation, including a recurrent Q-learning use case and an explanation of how learning_mask and validity_mask should be consumed. I also renamed stride to dilation: the parameter spaces records inside a window and does not control spacing or overlap between independently sampled windows. The PR title/body should be updated accordingly.

For follow-ups, I suggest:

- Design shared, efficient replay-boundary detection used by Sequence, SliceSampler, and other trajectory-aware components, including caching/invalidation, ring wraparound, device placement, and benchmarks.
- Add ndim > 1 support by expanding the time coordinate while preserving lane coordinates, including multidimensional anchor-priority reduction.

The boundary-detection work should be holistic rather than a local optimization of Sequence (we have that problem in many places in torchrl). Compile can help, it worked nicely with SliceSampler back in the days IIRC.

IMO that's a high difficulty task.

Issue #4055, opened 2026-07-31
Prepare TorchRL's Hydra entry points for Hydra 1.4


omry:
Hi! Hydra maintainer here. A heads-up about the upcoming Hydra 1.4 release.

Hydra 1.4 is not ready for production yet. It is a very large and long-delayed release, and it will contain several breaking changes accumulated during that extended development period. The current development release should only be used for compatibility testing.

One planned change is that Hydra 1.4 will stop accepting `version_base` values below `"1.3"`. Omitting `version_base` will select current behavior, while `version_base=None` will be deprecated. The rationale and migration plan are tracked in [Hydra #3237](https://github.com/facebookresearch/hydra/issues/3237).

TorchRL currently:

- Declares `hydra-core>=1.1`, which will allow Hydra 1.4 to be installed once released.
- Has approximately 50 entry points using `version_base="1.1"`, including the [current PPO trainer](https://github.com/pytorch/rl/blob/ae421b98d0dba86e5ab0b24917d1e64f376ee6f9/sota-implementations/ppo_trainer/train.py#L8-L15).
- Has another four entry points using `version_base=None`.

Under Hydra 1.4, the `"1.1"` entry points will fail during application startup.

Please migrate maintained entry points to `version_base="1.3"` and raise the Hydra dependency floor to `hydra-core>=1.3`. The `version_base=None` occurrences should normally remove the argument to select current behavior.

For compatibility testing, the current development release is [`hydra-core 1.4.0.dev6`](https://pypi.org/project/hydra-core/1.4.0.dev6/):

```bash
python -m pip install --upgrade "hydra-core==1.4.0.dev6"
```

Because 1.4 contains other breaking changes, I recommend running TorchRL's relevant test suite against this development release in addition to updating `version_base`.

TorchRL's structured configuration system is an extensive Hydra integration, so I wanted to flag this early enough to avoid surprising users when Hydra 1.4 becomes available.

theap06:
@omry thanks for the heads up! I'll update our test suite and update any endpoints that need to be updated. Is the hydra-core dev6 release current up to date for devs. like will there be any other breaking changes before the prod release. I'll draft up a pr for this.

omry:
Hi @theap06 
There will likely be a few more breaking changes, but this is moving forward, not backward. I think making sure you are ready for dev6 is the right move if you want to upgrade incrementally.
It will also give you the chance to ask questions and provide early feedback before things are locked down.
[...]

PR #4056, opened 2026-08-01
[Versioning] Migrate Hydra entry points to version_base 1.3 for Hydra 1.4 readiness

theap06:
Prepares Hydra entry points for Hydra 1.4 (#4055)

- All `hydra.main` entry points and compose-API test calls now pass `version_base="1.3"` (previously `"1.1"`, `None`, or unset).
- Configs of scripts that ran under 1.1 compat gain `hydra.job.chdir: true`, keeping run-dir behavior unchanged.
- hydra-core floor raised to `>=1.3` in pyproject.toml.

Draft until Hydra 1.4 is released.

### DIFF
--- a/sota-implementations/expert-iteration/config/ei_gsm8k.yaml
+++ b/sota-implementations/expert-iteration/config/ei_gsm8k.yaml
@@ -166,8 +166,6 @@ logging:
   checkpoint_frequency: 10  # Save checkpoint every N batches
 
 hydra:
-  job:
-    chdir: true
   run:
     dir: outputs/${now:%Y-%m-%d}/${now:%H-%M-%S}
   sweep:
--- a/sota-implementations/expert-iteration/config/ei_gsm8k.yaml
+++ b/sota-implementations/expert-iteration/config/ei_gsm8k.yaml
@@ -166,6 +166,8 @@ logging:
   checkpoint_frequency: 10  # Save checkpoint every N batches
 
 hydra:
+  job:
+    chdir: true
   run:
     dir: outputs/${now:%Y-%m-%d}/${now:%H-%M-%S}
   sweep:
--- a/sota-implementations/expert-iteration/config/ei_ifeval.yaml
+++ b/sota-implementations/expert-iteration/config/ei_ifeval.yaml
@@ -164,8 +164,6 @@ logging:
   checkpoint_frequency: 10  # Save checkpoint every N batches
 
 hydra:
-  job:
-    chdir: true
   run:
     dir: outputs/${now:%Y-%m-%d}/${now:%H-%M-%S}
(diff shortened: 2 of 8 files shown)


theap06 (commented):
@omry left a PR addressing all the migrating all entry points.

omry:
Thanks for putting this together. The migration is broadly consistent: all 57 `@hydra.main` entry points now pin `version_base="1.3"`, the former 1.1-style applications receive `hydra.job.chdir: true`, and the updated Python and YAML files parse successfully.

One item to address is propagating the `hydra-core>=1.3` floor to every first-party dependency manifest. The PR updates `pyproject.toml`, but several direct installation paths still use an unconstrained `hydra-core` or `hydra-core>=1.1`, including:

- `examples/rlhf/requirements.txt`
- `sota-implementations/ppo-async/requirements.txt`
- `sota-implementations/grpo/requirements_gsm8k.txt`
- `sota-implementations/grpo/requirements_ifeval.txt`
- `docs/requirements.txt`
- the SOTA CI dependency installation

In an existing environment, an unconstrained requirement can leave Hydra 1.1 installed. The new `version_base="1.3"` decorators would then fail during application startup.

One broader, non-blocking observation: needing to repeat `hydra.job.chdir: true` across 53 configurations suggests these entry points lack a shared base configuration for common Hydra behavior. The explicit additions are fine for this migration, but consolidating shared settings into a base config included through the defaults list could reduce duplication and prevent drift during future migrations. That could be handled separately.

There is also a broader compatibility choice to make. The new requirement is `hydra-core>=1.3`, so it will allow Hydra 1.4, while `version_base="1.3"` only preserves the behaviors controlled by that setting. It does not protect TorchRL from other Hydra 1.4 or OmegaConf changes. Examples include native tuples becoming `TupleConfig` rather than `ListConfig`, changes to `hydra.utils.instantiate()` resolution and container semantics, and different handling of call-site dataclass or attrs instances.

I suggest choosing one of these approaches:

- Add a CI job against the current Hydra 1.4 development release to detect broader incompatibilities; or
- Temporarily constrain the dependency to `hydra-core>=1.3,<1.4`, then remove the ceiling once Hydra 1.4 is sufficiently finalized and TorchRL’s compatibility has been validated.

The current state—allowing Hydra 1.4 without testing it—leaves users exposed to incompatibilities outside the scope of `version_base`.

theap06:
@omry I just temporarily constrained the dep for now. once the 1.4 is finalized and the trl compability has been validated, I will draft a follow up pr for the updated dep.

vmoens (approved):
LGTM let's wait till the CI has finished running
Can you look at the PR once more, I made some edits, just want to check they make sense

PR #3888, opened 2026-06-21
[Refactor] Migrate remaining LossModules to mask-aware reduction (#3866)

coder-jayp:
## Description
Migrates remaining `LossModule` subclasses to use the mask-aware `self._reduce_loss(loss, tensordict=tensordict)` helper instead of direct `_reduce(loss, reduction=self.reduction)` calls, as requested in #3866.

**Modules migrated:**
- `cql.py` (CQLLoss, DiscreteCQLLoss)
- `crossq.py` (CrossQLoss)
- `ddpg.py` (DDPGLoss)
- `decision_transformer.py` (DTLoss, OnlineDTLoss)
- `deprecated.py`
- `dqn.py` (DQNLoss, DistributionalDQNLoss)
- `gail.py` (GAILLoss)
- `iql.py` (IQLLoss, DiscreteIQLLoss)
- `llm/grpo.py` (GRPOLoss)
- `ppo.py` (PPOLoss, ClipPPOLoss, KLPENPPOLoss)
- `redq.py` (REDQLoss)
- `sac.py` (SACLoss, DiscreteSACLoss)
- `td3.py` (TD3Loss)
- `td3_bc.py` (TD3BCLoss)

`bc.py` was intentionally left untouched (already migrated in #3850 as the reference case).

## Motivation and Context
Closes the loss-side gap from the sequence-RL composability work.

- #3695 added `SliceSampler(pad_output=True)` which produces batches with a `("collector", "mask")` key.
- #3850 introduced `LossModule._reduce_loss` and migrated BCLoss.

Behaviour is byte-identical when no mask is present. When the mask exists, padded positions are correctly excluded from reduction.

## Types of changes
- [x] New feature / improvement (non-breaking change which adds core functionality)

## Checklist
- [x] I have read the [CONTRIBUTING](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide
- [x] Tests pass (`pytest test/objectives/`)
- [ ] Documentation updated (if needed)

PR #4057, opened 2026-08-02
[BugFix] honor collector masks in loss reductions

### LINKED ISSUE #3866: [RFC] Mask-aware reduction across LossModule subclasses
## Background

[#3695 added `SliceSampler(pad_output=True)`, which returns sequence batches with a `("collector", "mask")` boolean key flagging real positions (`True`) from duplicated-last-step padding (`False`). [#3850  wired the `LossModule` base class to honor this mask in its reduction helper, and adopted it in `BCLoss` as the reference case.

**Most other loss modules still call `_reduce(loss, reduction=self.reduction)` directly, which silently averages padded positions into the gradient.** Each remaining loss needs a near-mechanical one-line migration to route through `self._reduce_loss(loss, tensordict=tensordict)` instead.

Behavior is byte-identical when no mask key is present, so this is a safe back-compatible change. Together these migrations close the loss-side gap from the sequence-RL composability work in [#3695 

## The pattern

The reference diff is [PR #3850] . Open it side-by-side with your target loss while migrating.

gtnv:
#3888 refactored objectives to route through _reduce_loss but the collector mask change from #3850 never merged, leaving a gap. The relevant code changed as well. This made it so padded timesteps from collector, mask were still included and could change the final loss and grads. 

BC & RND use it directly. A2C, ppo and reinforce pass input via TensorDict and REDQ/OnlineDT deal with shape-specific cases their own way. 

closes #3866.

Ran tests; 9314 passed.

cc @vmoens 

## Tested

`test/objectives`: **7525 passed, 2330 skipped**, matching the pre-PR baseline
(7522 passed with the same skips, plus the 3 new regression tests).

Worth flagging: before the fix in `4cc1e15`, this branch **failed 72 existing
tests** — every `TestTD3::test_td3_batcher` and `TestTD3BC::test_td3bc_batcher`
parametrization. Their fixtures supply `("collector", "mask")` with the standard
trailing-singleton shape `[B, T, 1]` against a `[B, T]` loss, and
`_expand_loss_mask` only handled masks with *fewer* dimensions than the loss, so
`expand_as` raised. `shifted_valid` never hit this because the value estimators
emit it already squeezed. That crash is the reason a mask-discovery change needs
the objectives suite run against a baseline, not just new unit tests.

### DIFF
--- a/torchrl/objectives/common.py
+++ b/torchrl/objectives/common.py
@@ -329,6 +329,20 @@ class LossModule(TensorDictModuleBase, metaclass=_LossMeta):
 
     @staticmethod
     def _expand_loss_mask(mask: torch.Tensor, loss: torch.Tensor) -> torch.Tensor:
+        # Validity masks conventionally carry a trailing singleton dimension, to
+        # broadcast against the [..., 1]-shaped rewards they accompany --
+        # ``("collector", "mask")`` is [B, T, 1] where a per-timestep loss is
+        # [B, T]. Drop those before broadcasting the other way.
+        while mask.ndim > loss.ndim and mask.shape[-1] == 1:
+            mask = mask.squeeze(-1)
+        if mask.ndim > loss.ndim:
+            raise ValueError(
+                f"A mask of shape {tuple(mask.shape)} cannot be applied to a "
+                f"loss of shape {tuple(loss.shape)}: the mask has more "
+                "non-singleton dimensions than the loss. Per-element masking "
+                "requires an elementwise loss, not one whose event dimensions "
+                "have already been reduced (e.g. a log-prob)."
+            )
         if mask.ndim < loss.ndim:
             mask = mask.reshape(mask.shape + (1,) * (loss.ndim - mask.ndim))
         return mask.expand_as(loss)
--- a/torchrl/objectives/common.py
+++ b/torchrl/objectives/common.py
@@ -264,8 +264,6 @@ class LossModule(TensorDictModuleBase, metaclass=_LossMeta):
 
     @staticmethod
     def _expand_loss_mask(mask: torch.Tensor, loss: torch.Tensor) -> torch.Tensor:
-        while mask.ndim > loss.ndim and mask.shape[-1] == 1:
-            mask = mask.squeeze(-1)
         if mask.ndim < loss.ndim:
             mask = mask.reshape(mask.shape + (1,) * (loss.ndim - mask.ndim))
         return mask.expand_as(loss)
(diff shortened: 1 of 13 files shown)


theap06 (commented):
cc @vmoens - can you trigger github workflows?

torchrl/objectives/common.py
```diff
@@ -279,12 +281,24 @@ def _reduce_loss(
     ) -> torch.Tensor:
         if reduction is None:
             reduction = self.reduction
-        if mask is None and tensordict is not None:
-            mask = tensordict.get("shifted_valid", default=None)
+        preserve_shape = mask is None
```
theap06: reduction="none" returns a different shape depending on where the mask came from. Let's implement one contract and drop the shifted_valid case on line 291

torchrl/objectives/redq.py
```diff
@@ -634,20 +633,18 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
                 "next.state_value": next_state_value.detach(),
                 "target_value": target_value.detach(),
             }
```
theap06, line 635: The reduction != none gate means that REDQ still ignores the mask under the reduction set none. This PR should be closing that bug. also, deprecated.py passes the unsqueezed td for actor/qvalue but the plain one for loss_alpha. What was your intuition behind this? Both are right.

torchrl/objectives/common.py
```diff
@@ -264,6 +264,8 @@ def _set_deprecated_ctor_keys(self, **kwargs) -> None:
 
     @staticmethod
     def _expand_loss_mask(mask: torch.Tensor, loss: torch.Tensor) -> torch.Tensor:
+        while mask.ndim > loss.ndim and mask.shape[-1] == 1:
```
theap06, line 326: Is this reachable on any of the current paths? REDQ pre-aligns via unsqueeze(0) and BC raises when mask.ndim > loss.ndim

theap06 (changes requested):
left some feedback cc @vmoens

torchrl/objectives/common.py
gtnv: ack. 👍

torchrl/objectives/common.py
gtnv: ur right. removed bc not reachable. 👍

torchrl/objectives/redq.py
gtnv: > The reduction != none gate means that REDQ still ignores the mask under the reduction set none. This PR should be closing that bug.

acknowledged.

> What was your intuition behind this? Both are right.

actore/qval losses have a leading q-network dim. so the mask would need to unsqueeze 0 to bradcast. loss_alpha only has the original batch dim so the TensorDict is just fine.

vmoens:
I made some edits in `77518041d2a04a66c4807c8bf2b75d4b43e0d8ed`, see the comments in the commit.

vmoens (approved):
LGTM thanks!

PR #4060, opened 2026-08-05
[CI] Install torchcodec nightly wheels instead of building from source

vmoens:
## Description

All CI jobs are currently failing (see e.g. #4059) because the nightly CI paths build torchcodec from source, and that build now hard-fails: torchcodec's CMake requires libheif for HEIC decoding, which the CI images don't ship.

Per Nicolas (torchcodec maintainer), torchcodec publishes nightly wheels on the PyTorch index, so we can install those instead of cloning and building. This switches the three source-build sites (`linux/scripts/run_all.sh`, `tutorials/scripts/run_all.sh`, `docs.yml`) to `--pre torchcodec --index-url https://download.pytorch.org/whl/nightly/<cu>`. Stable/release paths are unchanged.

Nightly wheels exist for the variants CI uses: `cpu` (tests-cpu, docs) and `cu130` (GPU jobs). The olddeps (cu118) jobs don't install torchcodec.

## Tested

- Verified the wheels exist on the index for cpu/cu126/cu130 (`torchcodec-0.16.0.dev20260804`).
- Dry-run resolve of the new install command in a clean venv: pulls only torchcodec, no torch pin, so it won't disturb the torch/torchvision nightlies installed earlier in the scripts.
- `bash -n` on both scripts and YAML-parsed `docs.yml`.
- CI on this PR is the real test.

### DIFF
--- a/.github/unittest/linux/scripts/run_all.sh
+++ b/.github/unittest/linux/scripts/run_all.sh
@@ -252,18 +252,9 @@ else
   uv_pip_install -e . --no-build-isolation --no-deps
 fi
 
-# install torchcodec (from source for nightly, from the PyTorch wheel index for stable)
+# install torchcodec from the PyTorch wheel index (nightly wheels for torch nightly)
 if [[ "$TORCH_VERSION" == "nightly" ]]; then
-  # torchcodec builds with scikit-build-core; --no-build-isolation requires the
-  # build backend to be present in the environment (pybind11/ninja/cmake already are).
-  uv_pip_install "scikit-build-core>=0.10"
-  torchcodec_dir=$(mktemp -d)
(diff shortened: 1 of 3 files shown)


vmoens:
CI update: the torchcodec nightly wheel installs cleanly on all jobs (`torchcodec==0.16.0.dev20260804+cpu` in <1s vs the failing source build), and tests now actually run — tests-cpu (3.14, mp) is fully green.

The one failure so far — tests-cpu (3.10, bulk) — is unrelated: Hydra 1.4 dropped `version_base="1.1"`, which ~50 `sota-implementations` scripts pass to `@hydra.main`, breaking `test_render.py::TestSotaCheckpointFactories`. That needs its own fix (pin `hydra-core<1.4` or migrate the scripts).

vmoens:
cc @NicolasHug

PR #4080, opened 2026-08-09
[BugFix] Fix TanhNormal sample scoring at tanh saturation

### LINKED ISSUE #2199: [BUG] Numerical Instability issues with `torchrl.modules.TanhNormal`
## Describe the bug

When training on `PettingZoo/MultiWalker-v9` with `Multi-Agent Soft Actor-Critic`, **all** losses (`loss_actor`, `loss_qvalue`, `loss_alpha`) explode after ~1M environment steps at most.

This phenomenon occurs regardless of (reasonable) hyperparameter and gradient clipping threshold choice.

## To Reproduce

```python
from copy import deepcopy
import tqdm
import numpy as np
from gymnasium.spaces import Box

import logging
import math

import torch
from torch import nn
import torch.distributions as D

from torchrl.data.replay_buffers import TensorDictReplayBuffer
from torchrl.data.replay_buffers.samplers import RandomSampler
from torchrl.data.replay_buffers.storages import LazyMemmapStorage 

from torchrl.envs import (
    check_env_specs,
    PettingZooEnv, 
    ParallelEnv,
    GymEnv
)

from torchrl.modules import AdditiveGaussianWrapper, ProbabilisticActor
from torchrl.modules.models import MLP
from torchrl.modules.models.multiagent import (
    MultiAgentMLP,
    MultiAgentNetBase
)
from torchrl.collectors import SyncDataCollector, MultiSyncDataCollector, RandomPolicy

from torchrl.objectives import SACLoss, SoftUpdate

from tensordict.nn import TensorDictModule, TensorDictSequential, NormalParamExtractor
from torchrl.envs.utils import ExplorationType, set_exploration_type
from torchrl.envs import EnvCreator, TransformedEnv, Compose, Transform, RewardSum, ObservationNorm, StepCounter
from torchrl.record import CSVLogger, VideoRecorder, PixelRenderTransform

import multiprocessing as mp

EPS = 1e-7
class SMACCNet(MultiAgentNetBase): 
    '''
    This is an MLP policy network for MultiAgent SAC.
    
    This is just a more limited version of MultiAgentMLP.
    (https://pytorch.org/rl/main/_modules/torchrl/modules/models/multiagent.html)
    '''
    def __init__(self, 
                n_agent_inputs: int | None,
                n_agent_outputs: int,
                n_agents: int,
                centralised: bool,
                share_params: bool,
                device = 'cpu',
                activation_class = nn.Tanh,
                **kwargs):

        self.n_agents = n_agents
        self.n_agent_inputs = n_agent_inputs
        self.n_agent_outputs = n_agent_outputs
        self.share_params = share_params
        self.centralised = centralised
        self.activation_class = activation_class
        self.device = device

        super().__init__(
            n_agents=n_agents,
            centralised=centralised,
            share_params=share_params,
            agent_dim=-2,
            device = device,
            **kwargs,
        )

    # Copied over from MultiAgentMLP.
    def _pre_forward_check(self, inputs):
        if inputs.shape[-2] != self.n_agents:
            raise ValueError(
                f"Multi-agent network expected input with shape[-2]={self.n_agents},"
                f" but got {inputs.shape}"
            )
        # If the model is centralized, agents have full observability
        if self.centralised:
            inputs = inputs.flatten(-2, -1)
        return inputs

    def init_net_params(self, net):
        def init_layer_params(layer):
            if isinstance(layer, nn.Linear):
                weight_gain = 1. / (100 if layer.out_features == self.n_agent_outputs else 1)
                torch.nn.init.xavier_uniform_(layer.weight, gain = weight_gain)
                if 'bias' in layer.state_dict():
                    torch.nn.init.zeros_(layer.bias)
        net.apply(init_layer_params)
        return net

    def _build_single_net(self, *, device, **kwargs):
        n_agent_inputs = self.n_agent_inputs
        if self.centralised and n_agent_inputs is not None:
            n_agent_inputs = self.n_agent_inputs * self.n_agents
        
        model = nn.Sequential(
            nn.Linear(n_agent_inputs, 400),
            self.activation_class(),
            nn.Linear(400, 300),
            self.activation_class(),
            nn.Linear(300, self.n_agent_outputs)
        ).to(self.device) # Bandaid fix to use MultiSyncDataCollector

        model = self.init_net_params(model)

        return model

class TqdmLoggingHandler(logging.Handler):
    def __init__(self, level=logging.NOTSET):
        super().__init__(level)

    def emit(self, record):
        try:
            msg = self.format(record)
            tqdm.tqdm.write(msg)
            self.flush()
        except Exception:
            self.handleError(record)  

# Main Function
if __name__ == "__main__":
    logging.basicConfig(level = logging.INFO)
    logger = logging.getLogger(__name__)
    logger.propagate = False
    logger.addHandler(TqdmLoggingHandler())

    mp.set_start_method("spawn", force = True)
    
    NUM_AGENTS = 3
    NUM_CRITICS = 2
    NUM_EXPLORE_WORKERS = 8
    EXPLORATION_STEPS = 30000
    MAX_EPISODE_STEPS = 1000
    DEVICE = "cuda"
    REPLAY_BUFFER_SIZE = int(1e6)
    VALUE_GAMMA = 0.99
    MAX_GRAD_NORM = 1.0
    BATCH_SIZE = 256
    LR = 1e-4
    UPDATE_STEPS_PER_EXPLORATION = 1500
    WARMUP_STEPS = 0 #int(2e5)
    TRAIN_TIMESTEPS = int(1e7)
    EVAL_INTERVAL = 1 #int(9e4 // EXPLORATION_STEPS) # Every 500k steps or so, evaluate once.

    SEED = 42
    torch.manual_seed(SEED)
    np.random.seed(SEED)

    # https://pytorch.org/rl/stable/tutorials/multiagent_competitive_ddpg.html
    # More tutorials: https://pytorch.org/tutorials/advanced/pendulum.html
    # Toy test: https://pettingzoo.farama.org/environments/sisl/multiwalker/def env_fn(mode, parallel = True):
        def base_env_fn():
            return PettingZooEnv(task = "multiwalker_v9", 
                                    parallel = True,
                                    seed = 42,
                                    n_walkers = NUM_AGENTS, 
                                    terminate_reward = -5.0,
                                    forward_reward = 1.0,
                                    fall_reward = -1.0,
                                    shared_reward = False, 
                                    max_cycles = MAX_EPISODE_STEPS, 
                                    render_mode = mode, 
                                    device = "cpu"
                                )

        if parallel:
            # Don't use.
            # https://discuss.pytorch.org/t/pettingzoo-trouble-running-multiple-marl-environments-in-parallel/203706/

            env = lambda: ParallelEnv(num_workers = 4,  # noqa: E731
                                        create_env_fn = base_env_fn, 
                                        device = "cpu",
                                        mp_start_method = "spawn",
                                        serial_for_single = True
                                    )
        else:
            env = base_env_fn # noqa: E731

        def env_with_transforms():
            # dummy_env = base_env_fn()
            # dummy_obs_transform = ObservationNorm(in_keys = [("walker", "observation")], standard_normal = True)
            # dummy_env = TransformedEnv(dummy_env, dummy_obs_transform)
            # dummy_obs_transform.init_stats(10000)


            init_env = env()
            # obs_transform = ObservationNorm(loc = dummy_obs_transform.loc + EPS,
            #                                 scale = dummy_obs_transform.scale + EPS,
            #                                 in_keys = [("walker", "observation")], 
            #                                 standard_normal = True
            #                             )
            init_env = TransformedEnv(init_env, Compose(
                                            StepCounter(max_steps = MAX_EPISODE_STEPS),
                                            RewardSum(
                                                in_keys = [init_env.reward_key for _ in range(NUM_AGENTS)], 
                                                out_keys = [("walker", "episode_reward")] * NUM_AGENTS, 
                                                reset_keys = ["_reset"] * NUM_AGENTS
                                            ),
                                            # obs_transform
                                        )
                                    )
            # del dummy_env, dummy_obs_transform
            return init_env

        return env_with_transforms

    train_env = env_fn(None, parallel = False)()

    if train_env.is_closed:
        train_env.start()
    eval_env = env_fn("rgb_array", parallel = False)()
    video_recorder = VideoRecorder(
                                    CSVLogger("multiwalker-toy-test", video_format = "mp4"), 
                                    tag = "rendered", 
                                    in_keys = ["pixels_record"]
                                )
    
    # Call the parent's render function
    eval_env.append_transform(PixelRenderTransform(out_keys = ["pixels_record"]))
    eval_env.append_transform(video_recorder)

    if eval_env.is_closed:
        eval_env.start()

    check_env_specs(train_env)
    check_env_specs(eval_env)

    print(f"Action: {train_env.full_action_spec}, Reward: {train_env.full_reward_spec}, Done: {train_env.full_done_spec}, Observation: {train_env.full_observation_spec}")
    
    print(f"group_map: {train_env.group_map}")
    print(f"Action: {train_env.action_keys}, Reward: {train_env.reward_keys}, Done: {train_env.done_keys}")
    
    # NOTE: The input and output spaces to be fed in are on a PER-AGENT basis.
    # Basically, if you have 16 agents observing 3D velocity and outputting speed (the magnitude), 
    # n_agent_inputs = 3, n_agent_outputs = 1.  
    obs_dim = train_env.full_observation_spec["walker", "observation"].shape[-1]
    action_dim = train_env.full_action_spec["walker", "action"].shape[-1]

    policy_net = nn.Sequential(
                        SMACCNet(n_agent_inputs = obs_dim,
                          n_agent_outputs = 2 * action_dim, 
                          n_agents = NUM_AGENTS,
                          centralised = False, 
                          share_params = True, 
                          device = "cpu",
                          activation_class = nn.LeakyReLU, 
                        ),
                        NormalParamExtractor(),
                    )
    
    critic_net = SMACCNet(n_agent_inputs = obs_dim + action_dim,
                          n_agent_outputs = 1,
                          n_agents = NUM_AGENTS,
                          centralised = True,
                          share_params = True, 
                          device = "cpu",
                          activation_class = nn.LeakyReLU, 
                        )

    # Hook our networks to TensorDictModules so they can be a part of the TensorDict pipeline...
    policy_net_td_module = TensorDictModule(module = policy_net,
                                            in_keys = [("walker", "observation")],
                                            # NOTE: These outputs must match with the parameter names of the 
                                            # distribution you are using!
                                            out_keys = [("walker", "loc"), ("walker", "scale")]
                                        )
 
    obs_act_module = TensorDictModule(lambda obs, act: torch.cat([obs, act], dim = -1),
                                        in_keys = [("walker", "observation"), ("walker", "action")],
                                        out_keys = [("walker", "obs_act")]
                                    )
    critic_net_td_module = TensorDictModule(module = critic_net,
                                            in_keys = [("walker", "obs_act")],
                                            out_keys = [("walker", "state_action_value")]
                                        )

    # Attach our raw policy network to a probabilistic actor
    policy_actor = ProbabilisticActor(
        module = policy_net_td_module,
        spec = train_env.full_action_spec["walker", "action"],
        in_keys = [("walker", "loc"), ("walker", "scale")],
        out_keys = [("walker", "action")],
        # TanhNormal is based off of pytorch, which as far as we know, 
        # implements a numerically stable log det jacobian.
        distribution_class = TanhNormal,
        distribution_kwargs = {
            "min": train_env.full_action_spec["walker", "action"].space.low,
            "max": train_env.full_action_spec["walker", "action"].space.high,
        },
        return_log_prob = True,
    )
    
    with torch.no_grad():
        fake_td = train_env.fake_tensordict()
        policy_actor(fake_td)

    dora = AdditiveGaussianWrapper(
        policy = policy_actor,
        action_key = ("walker", "action"),
        sigma_init = 0.3,
        sigma_end = 0.1,
        annealing_num_steps = TRAIN_TIMESTEPS // 2
    )

    critic_actor = TensorDictSequential(
                            obs_act_module, critic_net_td_module
                        ) 

    collector = MultiSyncDataCollector(
                    [env_fn(None, parallel = False) for _ in range(NUM_EXPLORE_WORKERS)], 
                    policy = dora,
                    frames_per_batch = BATCH_SIZE,
                    max_frames_per_traj = 0,
                    total_frames = TRAIN_TIMESTEPS,
                    device = "cpu",
                    reset_at_each_iter = False
                )

    replay_buffer = TensorDictReplayBuffer(
        storage = LazyMemmapStorage(
            REPLAY_BUFFER_SIZE, device = "cpu",
        ),  # We will store up to memory_size multi-agent transitions
        sampler = RandomSampler(),
        batch_size = BATCH_SIZE,  # We will sample batches of this size
    )

    sac_loss = SACLoss(policy_actor.to(DEVICE), 
                        qvalue_network = critic_actor.to(DEVICE), 
                        num_qvalue_nets = 2,
                        loss_function = "l2",
                        delay_qvalue = True,
                        alpha_init = 0.1
                        )
    sac_loss.set_keys(
        action = ("walker", "action"),
        state_action_value = ("walker", "state_action_value"),
        reward = ("walker", "reward"),
        done = ("walker", "done"),
        terminated = ("walker", "terminated"),
    )

    sac_loss.make_value_estimator(gamma = VALUE_GAMMA)

    polyak_updater = SoftUpdate(sac_loss, tau = 0.005) 

    critic_params = list(sac_loss.qvalue_network_params.flatten_keys().values())
    actor_params = list(sac_loss.actor_network_params.flatten_keys().values())

    optimizer_actor = torch.optim.Adam(
        actor_params,
        lr = LR,
        weight_decay = 5e-4,
        eps = EPS,
        betas = (0.9, 0.98)
    )
    optimizer_critic = torch.optim.Adam(
        critic_params,
        lr = LR,
        weight_decay = 5e-4,
        eps = EPS,
        betas = (0.9, 0.98)
    )
    optimizer_alpha = torch.optim.Adam(
        [sac_loss.log_alpha],
        lr = LR,
        eps = EPS,
        betas = (0.9, 0.98)
    )

    # breakpoint()
    num_frames = 0
    pbar = tqdm.tqdm(total = TRAIN_TIMESTEPS)
    total_frames = 0
    backprop_ctr = 0
    train_rews, ep_lengths = [], []
    EXPLORATION_BATCHES = EXPLORATION_STEPS // BATCH_SIZE

        collector.update_policy_weights_()

        pbar.update(tensordict.numel())

        tensordict = tensordict.reshape(-1)
        current_frames = tensordict.numel()
        # Add to replay buffer
        replay_buffer.extend(tensordict.cpu())
        total_frames += current_frames

        backprop_ctr += 1
        # Optimization steps
        if total_frames >= WARMUP_STEPS and backprop_ctr > EXPLORATION_BATCHES:
            backprop_ctr = 0
            losses = TensorDict({}, batch_size = [UPDATE_STEPS_PER_EXPLORATION])
            alphas = TensorDict({}, batch_size = [UPDATE_STEPS_PER_EXPLORATION])
            for j in range(UPDATE_STEPS_PER_EXPLORATION):
                # Sample from replay buffer
                sampled_tensordict = replay_buffer.sample()

                if str(sampled_tensordict.device) != DEVICE:
                    sampled_tensordict = sampled_tensordict.to(DEVICE, non_blocking = False)
                else:
                    sampled_tensordict = sampled_tensordict.clone()

                try:
                    # Compute loss
                    loss_td = sac_loss(sampled_tensordict)
                except KeyError:
                    raise Exception(f"wtf {sampled_tensordict}\n{obs_act_module(sampled_tensordict)['walker', 'obs_act']}")


                alpha_loss = loss_td["loss_alpha"]

                # Update actor
                optimizer_actor.zero_grad()
                actor_loss.backward()
                actor_grad_norm = torch.nn.utils.clip_grad_norm_(actor_params, max_norm = MAX_GRAD_NORM)
                optimizer_actor.step()

                # Update critic
                optimizer_critic.zero_grad()
                q_loss.backward()
                q_grad_norm = torch.nn.utils.clip_grad_norm_(critic_params, max_norm = MAX_GRAD_NORM)
                optimizer_critic.step()

                # Update alpha
                optimizer_alpha.zero_grad()
                alpha_loss.backward()
                alpha_grad_norm = torch.nn.utils.clip_grad_norm_([sac_loss.log_alpha], max_norm = MAX_GRAD_NORM)
                optimizer_alpha.step()

                losses[j] = loss_td.select(
                    "loss_actor", "loss_qvalue", "loss_alpha"
                ).detach()

                alphas[j] = loss_td.select("alpha")


                # Update qnet_target params
                polyak_updater.step()

            # Some other stuff I ripped out from https://github.com/pytorch/rl/blob/main/sota-implementations/sac/sac.py
            episode_end = (
                tensordict["next", "done"]
                if tensordict["next", "done"].any()
                else tensordict["next", "truncated"]
            )
            opening_banner = "-" * 10 + f" Batch {i + 1} " + "-" * 10

            def get_mean(src, key): 
                return src.get(key).mean().item()

            logger.info(opening_banner)
            logger.info(f"Average Actor Loss: {get_mean(losses, 'loss_actor')}")
            logger.info(f"Average Q Loss: {get_mean(losses, 'loss_qvalue')}")
            logger.info(f"Average Alpha: {get_mean(alphas, 'alpha')} (Loss: {get_mean(losses, 'loss_alpha')})")
            logger.info("-" * len(opening_banner))
            
            
            ep_length = tensordict['next', 'step_count'][episode_end].to(dtype = torch.float64)

            if ep_length.numel():
                ep_lengths.append(ep_length.mean().item())

        agent_terminated = torch.stack(
            [
                tensordict["next", "walker", "done"][:, agent_id, 0]
                if tensordict["next", "walker", "done"][:, agent_id, 0].any()
                else tensordict["next", "walker", "truncated"][:, agent_id, 0]
                for agent_id in range(NUM_AGENTS)
            ], 
            dim = 1
        )
        train_reward = tensordict['next', 'walker', 'episode_reward'][agent_terminated]

        if train_reward.numel():
            train_rews.append(train_reward.mean().item())

        if not ((i + 1) % (EVAL_INTERVAL * EXPLORATION_BATCHES)):
            logger.info(
                        f"Mean Train Reward Across Past {EVAL_INTERVAL} Collections: " +
                        (
                            f"{sum(train_rews) / len(train_rews)}" 
                            if len(train_rews) 
                            else f"NA (Training starts @ {WARMUP_STEPS} steps)"
                        )
                    )
            with set_exploration_type(ExplorationType.MODE), torch.no_grad():
                eval_rollout = eval_env.rollout(
                    MAX_EPISODE_STEPS,
                    policy_actor,
                    auto_cast_to_device=True,
                    break_when_any_done=True,
                )

                mean_eval_length = eval_rollout["next", "step_count"][-1].to(dtype = torch.float64).mean().item()
                logger.info(f"Mean Eval Reward: {eval_rollout['next', 'walker', 'episode_reward'][-1].mean().item()}")
                logger.info(f"Eval Length: {mean_eval_length}")
            ep_reward_list = []
            train_rews = []
            eval_env.transform.dump()

    collector.shutdown()
    train_env.close()
```

## Expected behavior

Loss values stay within `~ +/- 10^2` throughout training and do not increase to `~ +/- 10^x` where x >> 1.

## System info
```python
>>> import torchrl, numpy, sys
>>> print(f"TorchRL: {torchrl.__version__}\nNumPy: {numpy.__version__}\nPython3 Ver: {sys.version}\nPlatform: {sys.platform}")
TorchRL: 0.4.0 
NumPy: 1.25.0 
Python3 Ver: 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] 
Platform: linux
```

```bash
> lsb_release -a
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.3 LTS
Release:	22.04
Codename:	jammy
```

## Reason and Possible fixes

Though the environment's observation space is not normalized and carries unbounded entries, the issue does not appear to entirely arise from the poor observation scaling, since adding a `torchrl.envs.ObservationNorm` does not mitigate the issue.

Debugging reveals that unusually large and negative values for `log_prob` are somehow being fed into the `SACLoss` calculations from the reimplementation of `torch.distributions.transforms.TanhTransform`.
https://github.com/pytorch/rl/blob/3e6cb8419df56d9263d1daa48f9c3be5f01eaea6/torchrl/modules/distributions/continuous.py#L289-L382

Since this reimplementation does not change much from the original `TanhTransform`, it is plausible that the reimplementation is *NOT* the root cause of the error. Nevertheless, replacing the reimplementation with an alternative variant gets rid of the issue altogether:
```python

class CustomTanhTransform(D.transforms.TanhTransform):

    def _inverse(self, y):
        # from stable_baselines3's `common.distributions.TanhBijector`
        """
        Inverse of Tanh

        Taken from Pyro: https://github.com/pyro-ppl/pyro
        0.5 * torch.log((1 + x ) / (1 - x))
        """

        y = y.clamp(-1. + EPS, 1. - EPS)
        return 0.5 * (y.log1p() - (-y).log1p())

    def log_abs_det_jacobian(self, x, y):
        # From PyTorch `TanhTransform`
        '''
        tl;dr log(1-tanh^2(x)) = log(sech^2(x)) 
                               = 2log(2/(e^x + e^(-x))) 
                               = 2(log2 - log(e^x/(1 + e^(-2x)))
                               = 2(log2 - x - log(1 + e^(-2x)))
                               = 2(log2 - x - softplus(-2x)) 
        '''
        
        return 2.0 * (math.log(2.0) - x - nn.functional.softplus(-2.0 * x))

class TanhNormalStable(D.TransformedDistribution):
    def __init__(self, loc, scale, event_dims = 1):
        self._event_dims = event_dims
        self._t = [
            CustomTanhTransform()
        ]
        self.update(loc, scale)

    def log_prob(self, value):
        """
        Scores the sample by inverting the transform(s) and computing the score
        using the score of the base distribution and the log abs det jacobian.
        """
        if self._validate_args:
            self._validate_sample(value)
        event_dim = len(self.event_shape)
        log_prob = 0.0
        y = value
        for transform in reversed(self.transforms):
            x = transform.inv(y)
            event_dim += transform.domain.event_dim - transform.codomain.event_dim
            log_prob = log_prob - D.utils._sum_rightmost(
                transform.log_abs_det_jacobian(x, y),
                event_dim - transform.domain.event_dim,
            )
            y = x

        log_prob = log_prob + D.utils._sum_rightmost(
            self.base_dist.log_prob(y), event_dim - len(self.base_dist.event_shape)
        )

        log_prob = torch.clamp(log_prob, min = math.log10(EPS)) # <- **CLAMPING THIS SEEMS TO RESOLVE THE ISSUE**
        return log_prob

    def update(self, loc: torch.Tensor, scale: torch.Tensor) -> None:
        self.loc = loc
        self.scale = scale
        if (
            hasattr(self, "base_dist")
            and (self.base_dist.base_dist.loc.shape == self.loc.shape)
            and (self.base_dist.base_dist.scale.shape == self.scale.shape)
        ):
            self.base_dist.base_dist.loc = self.loc
            self.base_dist.base_dist.scale = self.scale
        else:
            base = D.Independent(D.Normal(self.loc, self.scale), self._event_dims)
            super().__init__(base, self._t)

    @property
    def mode(self):
        m = self.base_dist.base_dist.mean
        for t in self.transforms:
            m = t(m)
        return m
``` 
although such a fix flies in the face of [this comment from the PyTorch devs](https://github.com/pytorch/pytorch/blob/main/torch/distributions/transforms.py#L654-L687).

## Checklist

gtnv:
## Description

Fix `TanhNormal.rsample() -> log_prob()` when finite-precision tanh saturation prevents `atanh(action)` from recovering the generating preimage, corrupting SAC scores and gradients.

`TanhNormal` now retains one forward-produced preimage and reuses it only for its unchanged sample. External or mutated values use the ordinary inverse path, and `update()` clears the cache. Compilation, inference, and CUDA capture validate the sample through a tensor snapshot. There are no clamps, flags, or training-specific defaults.

### Testing

- Direct final-head checks: `main` misses the score by `16,312` and the scale gradient by `165,992`; this PR is within `7.63e-6` and `9.54e-7` in eager, Dynamo eager, and Inductor.

tanhnormal_exact_preimage_h100

- Matched MultiWalker SAC, 3 seeds × 2M frames: every `main` run develops score corruption and exploding losses; corrected runs keep score error at `0.0` and raw critic loss at or below `23.6`. E2E ran at `efb3a592b`; final head only adds the CUDA-capture dispatch validated below.

multiwalker_main_vs_fix

- Final head `be7a48dba`: focused H100 `18 passed`; distributions `974 passed`; pre-commit passes. Objectives at the behavior-identical E2E head: `7528 passed, 2324 skipped, 6 deselected`.
- Safe/unsafe tanh with ordinary/custom bounds passes inference and CUDA graph within `7.63e-6`; scale-gradient error is at most `4.55e-7`, with zero allocated-memory growth across 20,000 iterations.

The cached sample keeps its generating loc/scale gradient path. Cloned and external actions retain ordinary action gradients.

Closes #2199

cc @vmoens @theap06

### DIFF
--- a/docs/source/reference/modules_distributions.rst
+++ b/docs/source/reference/modules_distributions.rst
@@ -18,3 +18,16 @@ Custom distribution classes for RL, extending PyTorch distributions.
     TanhDelta
     TanhNormal
     TruncatedNormal
+
+Sampling utilities
+==================
+
+.. currentmodule:: torchrl.modules.distributions.utils
+
+.. autosummary::
+    :toctree: generated/
+    :template: rl_template_fun.rst
+
+    sample_and_log_prob
+    rsample_and_log_prob
+    composite_entropy
--- a/test/modules/test_actor.py
+++ b/test/modules/test_actor.py
@@ -30,7 +30,6 @@ from torchrl.modules import (
     MultiStepActorWrapper,
     ProbabilisticActor,
     SafeModule,
-    SafeProbabilisticModule,
     TanhDelta,
     TanhModule,
     TanhNormal,
@@ -46,33 +45,6 @@ from torchrl.modules.vla import LeRobotPolicyWrapper, TinyVLA, VLAWrapperBase
 
 from torchrl.testing import get_default_devices
 from torchrl.testing.mocking_classes import CountingEnv, NestedCountingEnv
-
-
-@pytest.mark.parametrize("device", get_default_devices())
-@pytest.mark.parametrize("num_samples", [None, 3])
-def test_probabilistic_module_tanhnormal_joint_log_prob(device, num_samples):
-    loc = torch.tensor([[20.0, -20.0]], device=device)
-    scale = torch.full_like(loc, 0.1)
-    module = SafeProbabilisticModule(
-        in_keys=["loc", "scale"],
-        out_keys=["action"],
-        distribution_class=TanhNormal,
-        default_interaction_type=InteractionType.RANDOM,
-        return_log_prob=True,
-        log_prob_key="sample_log_prob",
-        num_samples=num_samples,
-        generator=torch.Generator(device=device).manual_seed(0),
-    )
-
-    torch.manual_seed(0)
-    sample_shape = torch.Size() if num_samples is None else (num_samples,)
-    expected_action, expected_log_prob = TanhNormal(loc, scale).rsample_and_log_prob(
-        sample_shape
-    )
-    result = module(TensorDict({"loc": loc, "scale": scale}, batch_size=[1]))
-
-    torch.testing.assert_close(result["action"], expected_action)
-    torch.testing.assert_close(result["sample_log_prob"], expected_log_prob)
 
 
 @pytest.mark.parametrize(
--- a/test/modules/test_actor.py
+++ b/test/modules/test_actor.py
@@ -30,6 +30,7 @@ from torchrl.modules import (
     MultiStepActorWrapper,
     ProbabilisticActor,
     SafeModule,
+    SafeProbabilisticModule,
     TanhDelta,
     TanhModule,
     TanhNormal,
@@ -45,6 +46,40 @@ from torchrl.modules.vla import LeRobotPolicyWrapper, TinyVLA, VLAWrapperBase
 
 from torchrl.testing import get_default_devices
 from torchrl.testing.mocking_classes import CountingEnv, NestedCountingEnv
+
+
+@pytest.mark.parametrize("device", get_default_devices())
+@pytest.mark.parametrize("num_samples", [None, 3])
+def test_probabilistic_module_tanhnormal_joint_log_prob(
+    device, num_samples, monkeypatch
+):
+    loc = torch.tensor([[20.0, -20.0]], device=device)
+    scale = torch.full_like(loc, 0.1)
+    module = SafeProbabilisticModule(
+        in_keys=["loc", "scale"],
+        out_keys=["action"],
+        distribution_class=TanhNormal,
+        default_interaction_type=InteractionType.RANDOM,
+        return_log_prob=True,
+        log_prob_key="sample_log_prob",
+        num_samples=num_samples,
+        generator=torch.Generator(device=device).manual_seed(0),
+    )
+
+    sample_shape = torch.Size() if num_samples is None else (num_samples,)
+    torch.manual_seed(0)
+    expected_action, expected_log_prob = TanhNormal(loc, scale).rsample_and_log_prob(
+        sample_shape
+    )
+
+    def fail_log_prob(*args, **kwargs):
+        raise AssertionError("a freshly sampled TanhNormal was inverse-scored")
+
+    monkeypatch.setattr(TanhNormal, "log_prob", fail_log_prob)
+    result = module(TensorDict({"loc": loc, "scale": scale}, batch_size=[1]))
+
+    torch.testing.assert_close(result["action"], expected_action)
+    torch.testing.assert_close(result["sample_log_prob"], expected_log_prob)
 
 
 @pytest.mark.parametrize(
--- a/test/modules/test_actor.py
+++ b/test/modules/test_actor.py
@@ -30,6 +30,7 @@ from torchrl.modules import (
     MultiStepActorWrapper,
     ProbabilisticActor,
     SafeModule,
+    SafeProbabilisticModule,
     TanhDelta,
     TanhModule,
     TanhNormal,
@@ -45,6 +46,33 @@ from torchrl.modules.vla import LeRobotPolicyWrapper, TinyVLA, VLAWrapperBase
 
 from torchrl.testing import get_default_devices
 from torchrl.testing.mocking_classes import CountingEnv, NestedCountingEnv
+
+
+@pytest.mark.parametrize("device", get_default_devices())
+@pytest.mark.parametrize("num_samples", [None, 3])
+def test_probabilistic_module_tanhnormal_joint_log_prob(device, num_samples):
+    loc = torch.tensor([[20.0, -20.0]], device=device)
+    scale = torch.full_like(loc, 0.1)
+    module = SafeProbabilisticModule(
+        in_keys=["loc", "scale"],
+        out_keys=["action"],
+        distribution_class=TanhNormal,
+        default_interaction_type=InteractionType.RANDOM,
+        return_log_prob=True,
+        log_prob_key="sample_log_prob",
+        num_samples=num_samples,
+        generator=torch.Generator(device=device).manual_seed(0),
+    )
+
+    torch.manual_seed(0)
+    sample_shape = torch.Size() if num_samples is None else (num_samples,)
+    expected_action, expected_log_prob = TanhNormal(loc, scale).rsample_and_log_prob(
+        sample_shape
+    )
+    result = module(TensorDict({"loc": loc, "scale": scale}, batch_size=[1]))
+
+    torch.testing.assert_close(result["action"], expected_action)
+    torch.testing.assert_close(result["sample_log_prob"], expected_log_prob)
 
 
 @pytest.mark.parametrize(
--- a/test/objectives/test_cql.py
+++ b/test/objectives/test_cql.py
@@ -148,6 +148,22 @@ class TestCQL(LossModuleTestBase):
         )
         self.reset_parameters_recursive_test(loss_fn)
 
+    def test_cql_actor_uses_joint_sample_log_prob(self, monkeypatch):
+        torch.manual_seed(self.seed)
+        td = self._create_mock_data_cql()
+        loss_fn = CQLLoss(
+            actor_network=self._create_mock_actor(),
+            qvalue_network=self._create_mock_qvalue(),
+        )
+
+        def fail_log_prob(*args, **kwargs):
+            raise AssertionError("CQL inverse-scored a freshly sampled TanhNormal")
+
+        monkeypatch.setattr(TanhNormal, "log_prob", fail_log_prob)
+        actor_loss, metadata = loss_fn.actor_loss(td)
+        assert actor_loss.isfinite().all()
+        assert metadata[loss_fn.tensor_keys.log_prob].isfinite().all()
+
     @pytest.mark.parametrize("delay_actor", (True, False))
     @pytest.mark.parametrize("delay_qvalue", (True, True))
     @pytest.mark.parametrize("max_q_backup", [True, False])
--- a/test/objectives/test_sac.py
+++ b/test/objectives/test_sac.py
@@ -1281,6 +1281,28 @@ class TestSAC(LossModuleTestBase):
         )
         loss.load_state_dict(state)
 
+    def test_sac_uses_joint_sample_log_prob(self, version, monkeypatch):
+        torch.manual_seed(self.seed)
+        td = self._create_mock_data_sac()
+        actor = self._create_mock_actor()
+        qvalue = self._create_mock_qvalue()
+        value = self._create_mock_value() if version == 1 else None
+        loss = SACLoss(
+            actor_network=actor,
+            qvalue_network=qvalue,
+            value_network=value,
+        )
+
+        def fail_log_prob(*args, **kwargs):
+            raise AssertionError("SAC rescored a newly sampled action")
+
+        monkeypatch.setattr(TanhNormal, "log_prob", fail_log_prob)
+        with pytest.warns(
+            UserWarning, match="No target network updater"
+        ) if rl_warnings() else contextlib.nullcontext():
+            result = loss(td)
+        assert result.isfinite().all()
+
     @pytest.mark.parametrize("action_dim", [1, 2, 4, 8])
     def test_sac_target_entropy_auto(self, version, action_dim):
         """Regression test for issue #3291: target_entropy='auto' should be -dim(A)."""
--- a/test/objectives/test_sac.py
+++ b/test/objectives/test_sac.py
@@ -1281,28 +1281,6 @@ class TestSAC(LossModuleTestBase):
         )
         loss.load_state_dict(state)
 
-    def test_sac_uses_joint_sample_log_prob(self, version, monkeypatch):
-        torch.manual_seed(self.seed)
-        td = self._create_mock_data_sac()
-        actor = self._create_mock_actor()
-        qvalue = self._create_mock_qvalue()
-        value = self._create_mock_value() if version == 1 else None
-        loss = SACLoss(
-            actor_network=actor,
-            qvalue_network=qvalue,
-            value_network=value,
-        )
-
-        def fail_log_prob(*args, **kwargs):
-            raise AssertionError("SAC rescored a newly sampled action")
-
-        monkeypatch.setattr(TanhNormal, "log_prob", fail_log_prob)
-        with pytest.warns(
-            UserWarning, match="No target network updater"
-        ) if rl_warnings() else contextlib.nullcontext():
-            result = loss(td)
-        assert result.isfinite().all()
-
     @pytest.mark.parametrize("action_dim", [1, 2, 4, 8])
     def test_sac_target_entropy_auto(self, version, action_dim):
         """Regression test for issue #3291: target_entropy='auto' should be -dim(A)."""
--- a/test/objectives/test_sac.py
+++ b/test/objectives/test_sac.py
@@ -1281,6 +1281,28 @@ class TestSAC(LossModuleTestBase):
         )
         loss.load_state_dict(state)
 
+    def test_sac_uses_joint_sample_log_prob(self, version, monkeypatch):
+        torch.manual_seed(self.seed)
+        td = self._create_mock_data_sac()
+        actor = self._create_mock_actor()
+        qvalue = self._create_mock_qvalue()
+        value = self._create_mock_value() if version == 1 else None
+        loss = SACLoss(
+            actor_network=actor,
+            qvalue_network=qvalue,
+            value_network=value,
+        )
+
+        def fail_log_prob(*args, **kwargs):
+            raise AssertionError("SAC inverse-scored a freshly sampled TanhNormal")
+
+        monkeypatch.setattr(TanhNormal, "log_prob", fail_log_prob)
+        with pytest.warns(
+            UserWarning, match="No target network updater"
+        ) if rl_warnings() else contextlib.nullcontext():
+            result = loss(td)
+        assert result.isfinite().all()
+
     @pytest.mark.parametrize("action_dim", [1, 2, 4, 8])
     def test_sac_target_entropy_auto(self, version, action_dim):
         """Regression test for issue #3291: target_entropy='auto' should be -dim(A)."""
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -149,7 +149,7 @@ class TestTanhNormal:
     @pytest.mark.parametrize("safe_tanh", [False, True])
     @pytest.mark.parametrize("compiled", [False, True])
     @pytest.mark.parametrize("device", get_default_devices())
-    def test_tanhnormal_rsample_log_prob_consistency(
+    def test_tanhnormal_rsample_and_log_prob(
         self, dtype, low, high, safe_tanh, compiled, device
     ):
         if compiled and sys.version_info >= (3, 14):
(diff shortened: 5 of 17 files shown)


theap06 (approved):
LGTM! Thanks

vmoens:
Reviewed latest head `be7a48dba7`, including the CUDA-capture follow-up commit and the current approval. The single-sample cache is invalidated on `update()`, detects eager in-place mutation via tensor versioning, uses a snapshot for compile/inference/capture paths, and falls back to the ordinary inverse path for external or stale samples; custom affine bounds are included because the composed transform is cached as a whole. The focused eager/compiled consistency and invalidation matrix passes locally (`18 passed`), and `git diff --check` is clean. No blocking finding from this review; the remaining platform-specific coverage is the CUDA-capture path already described in the PR.

vmoens-ii (dismissed):
[duplicated review]

vmoens (changes requested):
The fix appears correct only for the narrow `rsample() -> log_prob()` path on the exact tensor returned by the most recent sampling-like call. I do not think we should make these stateful semantics the default for `TanhNormal`.

The cache is single-entry and identity-based. A second call to `rsample()`, `sample()`, or `deterministic_sample` replaces it, and `update()` clears it. Cloning, slicing, or otherwise replacing the tensor object also misses it. Consequently, benign control-flow changes such as

```python
a1 = dist.rsample()
a2 = dist.rsample()
log_prob1 = dist.log_prob(a1)
```

silently restore the saturated inverse path and its incorrect score/gradients. More generally, `log_prob(value)` becomes dependent on call history and Python object identity.

The gradient behavior is also inconsistent. For the unchanged cached sample, gradients with respect to `loc` and `scale` follow the cached preimage and appear mathematically correct. However, the cached `log_prob` is constructed from `cached_preimage`, not through the supplied `value`, so `autograd.grad(log_prob.sum(), sampled_action)` can report the action as unused. A cloned or external action instead uses the inverse path and has the ordinary action gradient. I do not think the same public operation should have different differentiation semantics based on identity and cache state.

There is also unconditional overhead on every sample: a full `detach().clone()`, retained references to the sample and preimage (and potentially their autograd graph), plus the snapshot. Compile/inference/capture additionally execute the equality reduction and selection/inverse machinery. Showing no allocator growth over repeated iterations rules out an unbounded leak, but does not establish acceptable peak memory, graph lifetime, or throughput.

I would prefer an explicit atomic operation such as `rsample_and_log_prob()`, computed directly from the pre-tanh value without persistent mutable state. The ordinary `log_prob(external_value)` can then remain a history-independent operation. TorchRL's probabilistic-module path can use the joint operation where required.

vmoens (commented):
Thanks for replacing the cache-based approach with an atomic sample-and-score operation. This is much safer, and the gradient path now appears correct because the sample and log-probability are computed from the same reparameterized preimage.

I still have two concerns before approving:

1. **Documentation**

`rsample_and_log_prob()` is a new public API, but it currently has only a one-line docstring. Please document:

- The numerical and gradient contract.
- Why this should be used instead of `rsample()` followed by `log_prob()`.
- The meaning and shape of both returned tensors.
- Interaction with event dimensions, bounds, and `safe_tanh`.
- The conditions under which `SafeProbabilisticModule` uses this operation, and when it silently falls back to the old behavior.

An example demonstrating the saturated-action case would also be useful.

Adding it to the repo's API doc would be nice as well!

2. **Library-wide adoption**

The PR integrates the operation into `SafeProbabilisticModule` and SAC, but several objectives obtain the distribution directly and still generate a sample followed immediately by `log_prob(sample)`. These include paths in CQL, CrossQ, REDQ, and the Monte Carlo entropy implementations in A2C, PPO, and Decision Transformer.

Those paths bypass `SafeProbabilisticModule`, so they can still encounter the same numerical problem with a saturated `TanhNormal`. Composite distributions and multi-output probabilistic modules also currently fall back to the two-step behavior.

I'm not sure what the proper API could be since that method isn't present everywhere (maybe we could expose it in all distributions within torchrl somehow!?).
We should audit the internally generated sample-and-score paths and use a shared helper (either free function or method) where applicable. This should only affect actions generated from the same distribution; scoring externally supplied or replay-buffer actions should remain a normal `log_prob()` call. Representative integration tests outside SAC would help ensure the fix is consistently adopted.

gtnv:
@vmoens alright: I've implemented the changes and expanded to the other use-cases.

I also tested on h100s and double-checked that the issue reproduced on main, and that the added code works like it should.

Ran tests for tanhnormal cases, all passed.
Actor/SAC integration tests: 6 passed  
CPU tests: 5,835 passed

vmoens (approved):
LGTM thanks for working on this!

PR #4084, opened 2026-08-10
[Performance] Cache and vectorize replay-boundary queries

theap06:
Depends on #4083. GitHub shows the cumulative diff until that PR lands.

## Summary

- cache boundaries by storage revision, including cross-process writes
- replace dense scans with compile-friendly `searchsorted` queries
- add focused replay-boundary benchmarks

GB10, 1M records: 0.536 ms to 0.161 ms (3.3x); invalidated query 0.478 ms.

## Tests

Replay core, samplers, storages, multiprocessing invalidation, and compiled kernels.

PR #4085, opened 2026-08-10
[Feature] Support multidimensional storage in Sequence sample units

theap06:
Depends on #4084. GitHub shows the cumulative diff until that PR lands.

## Summary

- expand time while preserving lane coordinates for `storage.ndim > 1`
- return full-coordinate indices and reduce priorities per anchor coordinate
- cover masks, nested keys, ring wraparound, and CUDA placement

GB10, `[100000, 64]` storage: 0.195 ms cached; 1.004 ms invalidated.

## Tests

Replay core, samplers, storages, CUDA, and full-graph boundary kernels.

### DIFF
--- a/benchmarks/test_replaybuffer_benchmark.py
+++ b/benchmarks/test_replaybuffer_benchmark.py
@@ -125,6 +125,28 @@ def test_sequence_boundary_query_benchmark(
     benchmark(query)
 
 
+@pytest.mark.parametrize("lanes", [8, 64])
+def test_sequence_multidimensional_boundary_query_benchmark(benchmark, lanes):
+    device = _replay_boundary_device()
+    time = 100_000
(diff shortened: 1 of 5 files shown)


vmoens:
Reviewed incremental head `a7edffdb72` on top of #4084 and the current PR state (no human review threads; CI awaits approval). The expansion changes consistently keep coordinate 0 as time, preserve lane coordinates in both `index` and `anchor_index`, and flatten/recover full anchors for lane-specific priority reduction. Locally, the five CPU multidimensional Sequence tests pass and the CUDA case skips as expected; the prioritized end-to-end case could not run in this checkout because the optional `SumSegmentTreeFp32` extension is unavailable, rather than from a Python assertion failure. `git diff --check` is clean. No code blocker found; the remaining confidence gap is that extension-backed priority test plus CUDA CI.

vmoens (approved):
LGTM amazing work!

PR #4093, opened 2026-08-11
[Performance] Enable worker metadata for automatic parallel envs

vmoens:
## Summary

- automatically enable worker-originated metadata when environment convenience constructors create a `ParallelEnv` from `num_workers`
- cover Gym, DMControl, Brax, Habitat, LIBERO, Genesis, MJLab, MuJoCo Playground, and process-based custom MuJoCo environments
- retain parent-side metadata for known heterogeneous LIBERO schemas and Habitat device assignments
- document the automatic behavior and extend the existing integration tests

## Motivation

`metadata_from_workers=True` was introduced in #4028 to avoid constructing temporary environments in the parent process. Convenience constructors such as `GymEnv(..., num_workers=N)` synthesized a `ParallelEnv` without enabling that mode, so they still paid the parent-side construction cost unless users built the `ParallelEnv` manually.

This follow-up makes the optimized path automatic when the generated workers have compatible tensor schemas. Gym's native `num_envs` vectorization remains unchanged.

## User impact

Automatically generated process-based environments now start their workers during construction and use pipe-based communication (`use_buffers=False`), as required by worker-originated metadata. Expensive simulator environments are no longer instantiated solely for metadata in the parent process.

## Validation

- `pytest -q test/envs/test_parallel.py -k metadata_from_workers`
- focused `num_workers` tests across Gym, LIBERO, Habitat, Genesis, MJLab, custom MuJoCo, Brax, DMControl, and MuJoCo Playground
- pre-commit hooks for all changed files (`ufmt` run directly from the cached Python 3.11 hook environment)

Local result: 16 passed. Seven optional-dependency cases were collected and skipped because those dependencies or CUDA were unavailable locally; the `ci/optdeps` label requests their full CI coverage.

### DIFF
--- a/docs/source/reference/envs_vectorized.rst
+++ b/docs/source/reference/envs_vectorized.rst
@@ -75,6 +75,11 @@ needed.
   should keep the default metadata path. At shutdown, workers started in this
   mode are closed one at a time to bound teardown resource spikes; the
   per-worker grace period is controlled by the ``shutdown_timeout`` argument.
+  Convenience constructors that create a :class:`ParallelEnv` from
+  ``num_workers`` (for example ``GymEnv("Pendulum-v1", num_workers=4)``)
+  enable worker-originated metadata automatically when the worker schemas are
+  compatible. Native vectorization arguments such as ``GymEnv(...,
+  num_envs=4)`` do not create a :class:`ParallelEnv` and are unaffected.
   Use one common factory with ``create_env_kwargs`` for
   worker-specific arguments:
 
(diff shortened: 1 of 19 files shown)


vmoens:
@ParamThakkar123 @itwasabhi, would you be able to review this follow-up? It automatically enables `metadata_from_workers=True` for the `num_workers` constructor paths. Param, you implemented constructor-level parallelization across Gym, DMControl, Brax, Habitat, and Genesis; Abhishek, you introduced the MuJoCo Playground integration and its parallel constructor path. Feedback on eager worker startup, `use_buffers=False`, and the heterogeneous-schema fallback would be especially valuable.

ParamThakkar123:
Sure @vmoens, I will review this one

PR #4073, opened 2026-08-08
[BugFix] Match DreamerV3 categorical and KL semantics

vmoens:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.17.0) (oldest at bottom):
* #4095
* #4075
* #4074
* __->__ #4073
* #4072
* #4068
* #4067
* #4066
* #4065

Summary:
- add configurable categorical unimix to the DreamerV3 prior and posterior
- add separate dynamics and representation KL terms with aggregated free nats
- configure the maintained example for 1% unimix and 1.0/0.1 KL weighting

Rationale:
The reference objective mixes categorical probabilities with a uniform
component, sums KL across stochastic variables before applying free nats, and
optimizes stopped-gradient dynamics and representation terms separately. The
previous balanced aggregate used different gradient weights and free-bit
semantics, which prevented numerical and training-curve parity.

Test plan:
- pytest test/objectives/test_dreamer_v3.py -q
- pytest test/modules/test_dreamer_components.py -k DreamerV3Components -q
- pytest --doctest-modules torchrl/objectives/dreamer_v3.py -q
- run the maintained DreamerV3 SOTA smoke with the worktree virtual environment
- verify frozen 1% unimix KL values, stopped-gradient routing, aggregated free
  nats, separate loss keys, and end-to-end optimization

PR #4094, opened 2026-08-12
[BugFix] Keep replay storage bookkeeping compilable

vmoens:
## Summary

- represent compilable storage revision and cursor bookkeeping as scalar tensors
- update the revision in place and carry tensor cursors through the compiled graph
- preserve Python integer access for eager cache logic and multiprocessing values for non-compilable storage
- cover repeated writes with `fullgraph=True` and verify revision/cursor semantics in the recompilation test

## First bad commit

The regression starts at `416941983` (`[Performance] Cache and vectorize replay-boundary queries`, #4084), merged on August 11.

Using the same Torch 2.11 environment:

- parent `cbc435e09`: `test__rand_given_ndim_recompile` passes
- `416941983`: the test fails with 12 recompilation records

That commit introduced two mutable Python scalars on the compiled storage write path:

1. `_mutation_revision_value` was incremented as a Python integer on every write, so Dynamo guarded each observed value and recompiled the resumed `TensorStorage.set()` frame.
2. `_set_last_cursor()` converted a tensor cursor with `.item()`, forcing a graph break.

The fix keeps both values as zero-dimensional tensors when `compilable=True`. Revision updates use `add_`, cursor updates stay tensor operations, and eager readers still receive Python integers. No compile-disable decorator is added.

This failure was observed in the CPU bulk job for #4073 and is unrelated to that PR's changes: https://github.com/pytorch/rl/actions/runs/31532919040/job/93917137645?pr=4073

## Validation

- first-bad-commit check: parent passes; `416941983` fails with 12 records
- recompilation test after fix: passes with one recorded recompilation locally; revision and cursor values remain correct
- repeated initialized writes compile with `fullgraph=True`
- focused compile tests pass on Torch 2.11 and Torch 2.13
- `pytest test/rb/test_storages.py -q`: 90 passed, 42 skipped
- storage suite plus cache invalidation tests: 92 passed, 42 skipped
- replay-buffer compile matrix: 12 passed with zero post-warmup recompilations
- compilable storage pickle/fullgraph round trip passes

### DIFF
--- a/test/rb/test_storages.py
+++ b/test/rb/test_storages.py
@@ -317,8 +317,8 @@ class TestStorages:
         sys.version_info >= (3, 14),
         reason="torch.compile is not supported on Python 3.14+",
     )
-    # This test checks if the `torch._dynamo.disable` wrapper around
-    # `TensorStorage._rand_given_ndim` is still necessary.
+    # This test checks that mutable storage bookkeeping does not cause
+    # excessive recompilation when writes and sampling are compiled together.
     def test__rand_given_ndim_recompile(self):
         torch._dynamo.reset_code_caches()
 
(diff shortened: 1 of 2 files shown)


vmoens:
Hi @theap06 — I traced this regression to 416941983793f7cbfbee3be5366617b9d07b4a47 from #4084. This PR keeps the replay-storage bookkeeping inside the compiled graph while restoring limited recompilation behavior. Would you mind reviewing the fix?

PR #4097, opened 2026-08-12
[Feature] Add joint sample-and-score distribution adapter

vmoens:
## Summary

- add `ensure_rsample_and_log_prob()` as an identity-preserving capability adapter
- expose `sample_and_log_prob()` and `rsample_and_log_prob()` on arbitrary PyTorch and third-party distributions through a lightweight proxy
- preserve concrete `isinstance` behavior and transparent `torch.distributions.kl_divergence` dispatch
- adapt distributions at TorchRL probabilistic-module and objective boundaries, allowing consumers to use the method contract without concrete-distribution branches or unwrapping
- document the method-oriented API and preserve the existing free-function fallback
- cover native identity, wrapped gradients, `torch.compile`, serialization, KL dispatch, composites, non-reparameterizable distributions, actor/objective integration, and sampling overhead

## Motivation

#4080 introduced atomic sample-and-score operations for `TanhNormal` and generic free-function fallbacks. Distribution-consuming modules are modular, however, and cannot assume which concrete distribution they will receive. This follow-up establishes a uniform internal method contract:

```python
action, log_prob = dist.rsample_and_log_prob()
```

Distributions that already provide the method retain object identity. Other distributions are adapted without modifying PyTorch classes or maintaining TorchRL copies of the `torch.distributions` namespace.

## Compatibility

The proxy delegates ordinary distribution behavior, including sampling, scoring, entropy, support, statistics, expansion, enumeration, mapping behavior for composite distributions, and distribution-specific attributes. Concrete `isinstance` checks remain transparent, including `CompositeDistribution` branches. PyTorch KL dispatch is registered for adapted/adapted and mixed adapted/raw pairs, so consumers do not need an unwrap API.

Structured `CompositeDistribution` samples continue to use the centralized component aggregation in the existing free functions. Distributions with `has_rsample=False` expose `sample_and_log_prob()` and continue to reject reparameterized sampling.

## Validation

- 2,023 passed across distribution, actor, CQL, A2C, PPO, and Reinforce suites
- direct non-skipped `KLPENPPOLoss` standard and composite executions passed
- adapter KL tests cover adapted/adapted, adapted/raw, and raw/adapted pairs
- GRPO dependency-free unit passed; transformer-backed cases require the optional `transformers` dependency unavailable locally
- sampling microbenchmark: adapted method 19.5 us mean; free function 20.0 us mean in the local run
- pinned `ufmt`, `flake8`, `pydocstyle`, and `git diff --check` passed

### DIFF
--- a/benchmarks/test_objectives_benchmarks.py
+++ b/benchmarks/test_objectives_benchmarks.py
@@ -28,6 +28,10 @@ from torchrl.modules import (
     SymLogValueTransform,
     TanhNormal,
 )
+from torchrl.modules.distributions.utils import (
+    ensure_rsample_and_log_prob,
+    rsample_and_log_prob,
+)
 from torchrl.objectives import (
     A2CLoss,
     ClipPPOLoss,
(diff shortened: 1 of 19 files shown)


vmoens:
not planned

PR #4103, opened 2026-08-13
[Feature] WS1: Add stable component boundary Protocols for external-l…

### LINKED ISSUE #4092: [Feature] WS1: Stable Component Boundaries for External Loops
## Motivation

Spin-off from the Post-training efforts RFC (#3948) specifically for **Workstream 1 (Stable component boundaries)**.

While working on the TRL adapters (WS2), it became clear that TorchRL components currently rely on implicit contracts. For example, knowing exactly what shape and keys a `ReplayBuffer` will yield, or exactly what keys a `Loss` module expects in a `TensorDict` (like `"tokens"` vs `"inputs"`), requires reading the implementation source code. 

For external users to confidently pull a single TorchRL component into their own training loop (like TRL or NeMo-RL), these interfaces need to be explicit, versioned, and documented public contracts so that adapters stay thin and don't break unexpectedly on version bumps.

## Solution

I recommend formalizing these boundaries through documentation and Python types. Specifically, I propose:

**1. Architectural Documentation**
Create a dedicated Sphinx guide (e.g., "Integrating TorchRL Components into External Loops") that explicitly documents the expected I/O contracts for:
* **Buffers:** Extend/sample expectations over token or trajectory data.
* **Rollouts:** Collector/inference-server protocol (including weight sync logic).
* **Losses:** Explicit definitions of the `TensorDict` keys expected as input and the loss keys yielded as output.

**2. Formal Protocols (Optional but recommended)**
Introduce Python `typing.Protocol` or lightweight base classes for these interfaces. This allows external adapters (and internal checks) to use static type checking (e.g., `mypy`) to verify they are satisfying the TorchRL contract before runtime.

## Alternatives

An alternative is to leave these contracts implicit and rely on users reading the source code of recipes like `grpo_utils.py` to figure out how to wire up a collector or buffer. However, this creates a high barrier to entry and leads to brittle external code that might break when internal TorchRL keys or shapes change.

## Additional context

Having just built the TRL interop adapters, I have a fresh perspective on exactly which contracts an external loop needs to consume. I would like to build these documentations and protocols if the approach looks good to you.

## Checklist

coder-jayp:
## Description

Implements Workstream 1 (Stable Component Boundaries) from RFC #3948.

Based on feedback from @vmoens, this PR avoids structural typing (`typing.Protocol`) to prevent false positives. Instead, it adds `torchrl/data/llm/contracts.py` with three explicit validation helper functions that formalise the implicit I/O contracts external training loops (TRL, NeMo-RL, custom) need to consume TorchRL components:

- `assert_buffer_contract(buffer)` — asserts `extend`, `sample`, and `write_count` exist, and explicitly documents expected TensorDict keys in the docstring.
- `assert_collector_contract(collector)` — asserts `update_policy_weights_`, `__iter__`, and `__next__` exist, and explicitly documents expected TensorDict keys in the docstring.
- `assert_loss_contract(loss_output, loss_type)` — asserts the required loss fields exist based on the loss type (e.g. `loss_objective` for GRPO).

These development utilities raise a helpful `TypeError` if an object does not satisfy the contract, providing a safe way to validate external adapters.

## Motivation and Context

Without explicit contracts, external adapter authors must read recipe source code to understand what keys a buffer sample carries or what a loss output exposes. This leads to brittle code that breaks silently on TorchRL version bumps. By providing explicit validation helpers and documentation, we give users a stable public contract to build against.

close #4092

## Types of changes

## Checklist

### DIFF
--- a/torchrl/data/llm/contracts.py
+++ b/torchrl/data/llm/contracts.py
@@ -2,13 +2,15 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-"""
-Defines validation helper functions for replay buffers, collectors, and
-loss outputs so external training loops (TRL, NeMo-RL, custom) can consume
-individual TorchRL components with confidence.
+"""Defines validation helper functions for TorchRL LLM components.
+
+Provides helpers for replay buffers, collectors, and loss outputs so external
+training loops (TRL, NeMo-RL, custom) can consume individual TorchRL components
+with confidence.
 """
 from __future__ import annotations
 
+import collections.abc
 from typing import Any
 
 __all__ = [
@@ -31,13 +33,15 @@ def assert_buffer_contract(buffer: Any) -> None:
         buffer: The buffer object to validate.
 
     Raises:
-        TypeError: If the buffer is missing ``extend``, ``sample``, or ``write_count``.
+        TypeError: If the buffer is missing a callable ``extend``, a callable ``sample``, or ``write_count``.
     """
-    missing = [
-        attr
-        for attr in ("extend", "sample", "write_count")
-        if not hasattr(buffer, attr)
-    ]
+    missing = []
+    if not callable(getattr(buffer, "extend", None)):
+        missing.append("extend (callable)")
+    if not callable(getattr(buffer, "sample", None)):
+        missing.append("sample (callable)")
+    if not hasattr(buffer, "write_count"):
+        missing.append("write_count")
     if missing:
         raise TypeError(
             f"Object of type '{type(buffer).__name__}' does not satisfy the "
@@ -60,13 +64,13 @@ def assert_collector_contract(collector: Any) -> None:
         collector: The collector object to validate.
 
     Raises:
-        TypeError: If the collector is missing ``update_policy_weights_``, ``__iter__``, or ``__next__``.
+        TypeError: If the collector is missing a callable ``update_policy_weights_``, or ``__iter__``.
     """
-    missing = [
-        attr
-        for attr in ("update_policy_weights_", "__iter__", "__next__")
-        if not hasattr(collector, attr)
-    ]
+    missing = []
+    if not callable(getattr(collector, "update_policy_weights_", None)):
+        missing.append("update_policy_weights_ (callable)")
+    if not isinstance(collector, collections.abc.Iterable):
+        missing.append("__iter__")
     if missing:
         raise TypeError(
             f"Object of type '{type(collector).__name__}' does not satisfy the "
--- a/torchrl/data/llm/contracts.py
+++ b/torchrl/data/llm/contracts.py
@@ -4,30 +4,23 @@
 # LICENSE file in the root directory of this source tree.
 """Stable component boundaries for LLM post-training external-loop interop (RFC #3948, WS1).
 
-Defines ``typing.Protocol`` interfaces for replay buffers, collectors, and
+Defines validation helper functions for replay buffers, collectors, and
 loss outputs so external training loops (TRL, NeMo-RL, custom) can consume
-individual TorchRL components without depending on concrete classes.
+individual TorchRL components with confidence.
 """
 from __future__ import annotations
 
-from typing import Any, Iterator, runtime_checkable
-
-import torch
-from tensordict import TensorDictBase
-from typing_extensions import Protocol
+from typing import Any
 
 __all__ = [
-    "PostTrainingBufferProtocol",
-    "PostTrainingCollectorProtocol",
-    "GRPOLossOutputProtocol",
-    "SFTLossOutputProtocol",
-    "assert_satisfies_protocol",
+    "assert_buffer_contract",
+    "assert_collector_contract",
+    "assert_loss_contract",
 ]
 
 
-@runtime_checkable
-class PostTrainingBufferProtocol(Protocol):
-    """Minimal interface a replay buffer must expose for post-training loops.
+def assert_buffer_contract(buffer: Any) -> None:
+    """Assert that a replay buffer exposes the necessary methods for post-training.
 
     **TensorDict key contract** — samples must carry:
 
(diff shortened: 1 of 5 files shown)


vmoens (commented):
Thanks, here are the changes that are required:

  - **High, confirmed:** `PostTrainingLossOutputProtocol` claims that `GRPOLossOutput` and `SFTLossOutput` satisfy it structurally, but the protocol declares both `loss_objective` and `loss_sft` as required members at [`contracts.py:81`](https://github.com/pytorch/rl/blob/96039cad69e7e675ce07c5eec8498db13168675e/torchrl/data/llm/contracts.py#L81). `GRPOLossOutput` exposes only `loss_objective`; `SFTLossOutput` exposes only `loss_sft`. The bespoke runtime helper accepts either field, masking the contradiction rather than providing the advertised static protocol. Split the protocols, define a real common interface, or make both implementations satisfy one shared contract.
  - **Medium, confirmed:** the brand-new tutorial uses multiple `print()` calls, including [`torchrl_component_contracts.py:62`](https://github.com/pytorch/rl/blob/96039cad69e7e675ce07c5eec8498db13168675e/tutorials/sphinx-tutorials/torchrl_component_contracts.py#L62), despite the repository rule that tutorials use prose comments rather than prints. It also introduces a module import mid-file at [line 155](https://github.com/pytorch/rl/blob/96039cad69e7e675ce07c5eec8498db13168675e/tutorials/sphinx-tutorials/torchrl_component_contracts.py#L155).
  - **Medium, confirmed:** the new test file uses function-level imports at [`test_llm_contracts.py:175`](https://github.com/pytorch/rl/blob/96039cad69e7e675ce07c5eec8498db13168675e/test/test_llm_contracts.py#L175) and has no required `if __name__ == "__main__": pytest.main(...)` block.
  - **Low:** the three new public protocol classes lack the runnable `>>>` examples required for new public classes.

tutorials/sphinx-tutorials/torchrl_component_contracts.py
```diff
@@ -0,0 +1,120 @@
+"""
+TorchRL Component Contracts for External Loops (WS1)
```
vmoens: WS1 is a reference to PR structures and has nothing to do with the lib core.

tutorials/sphinx-tutorials/torchrl_component_contracts.py
```diff
@@ -0,0 +1,120 @@
...
+=====================================================
+
+Documents the stable component boundaries from RFC #3948 WS1.
```
vmoens: Ditto

tutorials/sphinx-tutorials/torchrl_component_contracts.py
```diff
@@ -0,0 +1,120 @@
...
+from torchrl.objectives.llm.sft import SFTLossOutput
+
+# %%
+# Buffer contract
+# ---------------
+# Buffers must expose ``extend``, ``sample``, and ``write_count``.
+# TorchRL's built-in ``ReplayBuffer`` already satisfies this:
+
+rb = ReplayBuffer(storage=LazyTensorStorage(1_000), batch_size=8)
+assert isinstance(rb, PostTrainingBufferProtocol)
```
vmoens: I think that's the wrong contract.
This is what I don't like about protocols and why I did not use them so far. It makes no sense that this object returns true for that isinstance in this case.

vmoens (commented):
Not a big fan of Protocols in general. I think the tutorial shows the situation we *don't* want to end up in

coder-jayp:
@vmoens That makes total sense. I understand the concern that structural typing here creates a false sense of security. What approach would you recommend for making these component boundaries explicit? Documentation-only or something else. Happy to rework this entirely.

vmoens:
> @vmoens That makes total sense. I understand the concern that structural typing here creates a false sense of security. What approach would you recommend for making these component boundaries explicit? Documentation-only or something else. Happy to rework this entirely.


I don't know. What is the problem we are trying to solve here? I'm a bit confused by the scope of the PR.

coder-jayp:
@vmoens people building adapters for TRL or NeMo-RL don't have a clear way to know what TensorDict keys our components expect or yield. They have to read through the recipe source code just to figure out if they need ("next", "reward") or ("tokens", "full").

vmoens:
then maybe we just want a bunch of helpers or shallow "real" classes with no content (rather than protocols)?

coder-jayp:
I lean towards helpers.

Shallow base classes would solve the false-positive problem via explicit inheritance, but injecting LLM-specific base classes into core components like ReplayBuffer feels intrusive.

With helpers (e.g. validate_buffer_contract(buffer)), we can actually sample a dummy batch and assert that ("tokens", "full") is present. That validates the true semantic contract without touching TorchRL's core class hierarchies.

What do you think?

coder-jayp:
@vmoens Pivoted the PR to use explicit helper functions instead of Protocols—ready for another look!

torchrl/data/llm/contracts.py
```diff
@@ -0,0 +1,107 @@
...
+        buffer: The buffer object to validate.
+
+    Raises:
+        TypeError: If the buffer is missing ``extend``, ``sample``, or ``write_count``.
+    """
+    missing = [
+        attr
+        for attr in ("extend", "sample", "write_count")
+        if not hasattr(buffer, attr)
+    ]
```
gtnv: extend=3, sample=None, and a string write_count all pass here. that is an attribute check, not the advertised contract. can we rename it or validate callability and a sampled schema?

torchrl/data/llm/contracts.py
```diff
@@ -0,0 +1,107 @@
...
+        collector: The collector object to validate.
+
+    Raises:
+        TypeError: If the collector is missing ``update_policy_weights_``, ``__iter__``, or ``__next__``.
+    """
+    missing = [
+        attr
+        for attr in ("update_policy_weights_", "__iter__", "__next__")
+        if not hasattr(collector, attr)
+    ]
```
gtnv: a normal iterable may return a separate iterator. it then fails this __next__ check. can we validate iter(collector) instead?

torchrl/data/llm/contracts.py
coder-jayp: I've updated the check to validate that both extend and sample are callable().

PR #4108, opened 2026-08-14
[Versioning] Bump 0.14.0

vmoens:
## Description

Prepares TorchRL 0.14.0 against TensorDict 0.14, following the release runbook (compat patches + deprecation enactment + version bump + pin, one PR on main before the branch cut).

### TensorDict 0.14 compatibility patches

TensorDict 0.14 flipped `TensorDict.to_module`'s default to `preserve_module_state=True` (pytorch/tensordict#1764). TorchRL's functional param-swap machinery relies on the historical replacement semantics, so every such call site now passes `preserve_module_state=False` explicitly (the kwarg exists since tensordict 0.13):

- `MultiAgentNetBase` (`torchrl/modules/models/multiagent.py`): the constructor strips its template net to meta and swaps the shared params back in; under the new default the template kept **registered meta parameters**. Downstream, collectors' `has_meta_params` probe then mistook any policy containing a multi-agent net for a deferred weight-sync policy and skipped output-key inference — silently dropping policy outputs such as `("agents", "action_log_prob")` from rollouts. This is what broke the multiagent SOTA scripts (mappo/ippo, qmix_vdn, iql: `KeyError: "Couldn't find the log-prob ('agents', 'action_log_prob')"`), the multiagent tutorial, and with it the docs build.
- Collectors' meta-strip/device-cast paths (`_base.py`, `utils.py`, `_multi_base.py`): the device-cast wrote parameters without moving buffers under the new default (qmix on CUDA: `Expected action and e-greedy module to be on the same device`).
- RNN python path, inference server, `ModuleTransform`, testing helpers: same pattern, same explicit kwarg.
- `torchrl/weight_update/weight_sync_schemes.py` is intentionally left on the new preserve semantics (it was written for them).

`test/modules/test_multiagent_models.py` (red on main) passes locally against tensordict main after this change: 203/203, and the three failing multiagent SOTA scripts complete with the CI's smoke configuration.

### Deprecation enactment for 0.14

- Removed `MultiCollector.postprocs` property/setter alias (deprecated in favor of `.postproc`, scheduled for removal in v0.14). This is the only deprecation naming v0.14 in the codebase (`git log --grep="[Deprecation]" v0.13.0..main` is empty; grep for `removed in v0.14` shows no other hit).

### Version bump

- `version.txt`, `pyproject.toml` `fallback_version`, `.github/scripts/td_script.sh`, `.github/scripts/version_script.bat`: 0.13.0 -> 0.14.0
- tensordict pin: `>=0.14.0,<0.15.0` (tensordict 0.14.0 must be on PyPI before the torchrl release workflow runs; main CI installs tensordict from git and is unaffected)
- README CUDA-wheel install example bumped to 0.14.0

### Not in scope

- `TensorDict.load_memmap`'s new `allow_pickle` deprecation window (pytorch/tensordict#1763) only warns during 0.14; making torchrl's trusted internal loads explicit is deferred to the 0.15 preparation.
- The `test_rnn.py` GRU-scan failures observed locally on macOS/torch-2.13 are not reproducible in CI and are unrelated.


GaetanLepage:
Hi @vmoens !
It looks like the GitHub tag and release were not properly created for 0.14.0

PR #4109, opened 2026-08-14
[Feature] Add shared-memory exchange to AsyncEnvPool

theap06:
Adds opt-in shared-memory slot exchange, deadline-based batching, health statistics, configuration support, tests, documentation, and benchmark coverage.

### DIFF
--- a/torchrl/envs/async_envs.py
+++ b/torchrl/envs/async_envs.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 import abc
 import multiprocessing
 import threading
-import time
 from collections.abc import Callable, Mapping, Sequence
 from concurrent.futures import as_completed, FIRST_COMPLETED, ThreadPoolExecutor, wait
 from multiprocessing import Queue
@@ -24,6 +23,7 @@ from tensordict import (
 from tensordict.tensorclass import NonTensorData, NonTensorStack
 from tensordict.utils import _zip_strict, expand_as_right
 
+from torchrl._utils import timeit
 from torchrl.data.tensor_specs import NonTensor
 from torchrl.envs._async_exchange import _receive_batch, _SharedSlotExchange
 from torchrl.envs.common import _EnvPostInit, EnvBase
@@ -1118,18 +1118,18 @@ class ThreadingAsyncEnvPool(AsyncEnvPool):
         limit = len(futures) if max_get is None else max_get
         pending = set(futures)
         completed = []
-        deadline = None
+        deadline_timer = None
         while pending and len(completed) < min_get:
             done, pending = wait(pending, return_when=FIRST_COMPLETED)
             completed.extend(list(done)[: limit - len(completed)])
-            if deadline is None and timeout is not None:
-                deadline = time.monotonic() + timeout
+            if deadline_timer is None and timeout is not None:
+                deadline_timer = timeit("async_env_future_batch_deadline").start()
         while pending and len(completed) < limit:
             done = {future for future in pending if future.done()}
             if not done:
-                if deadline is None:
+                if deadline_timer is None:
                     break
-                remaining = deadline - time.monotonic()
+                remaining = timeout - deadline_timer.elapsed()
                 if remaining <= 0:
                     break
                 done, _ = wait(pending, timeout=remaining, return_when=FIRST_COMPLETED)
--- a/torchrl/envs/async_envs.py
+++ b/torchrl/envs/async_envs.py
@@ -7,13 +7,10 @@ from __future__ import annotations
 import abc
 import multiprocessing
 import threading
-
+import time
 from collections.abc import Callable, Mapping, Sequence
-from concurrent.futures import as_completed, ThreadPoolExecutor
-
-# import queue
+from concurrent.futures import as_completed, FIRST_COMPLETED, ThreadPoolExecutor, wait
 from multiprocessing import Queue
-from queue import Empty
 from typing import Literal
 
 import torch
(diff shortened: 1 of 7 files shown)


theap06:
@torchrlbot reviewer @vmoens

vmoens:
@torchrlbot rebase

vmoens:
@theap06 can you run some benchmarks with this to check how fast it is compared to the queue transmission?

theap06:
@vmoens 
image

vmoens (commented):
Here are some comments 

## High — Per-environment receive returns worker-owned shared storage after the environment is released

[`read_one`](https://github.com/pytorch/rl/blob/41a188edc8e06ec0560c5557ac5b01856f3c11e4/torchrl/envs/_async_exchange.py#L230-L235) and [`read_pair_one`](https://github.com/pytorch/rl/blob/41a188edc8e06ec0560c5557ac5b01856f3c11e4/torchrl/envs/_async_exchange.py#L261-L274) return `select(...)` results directly from the worker's shared result slots. Immediately afterward, the receive path removes that worker from `_busy`, allowing the caller to send another action/reset. The worker can then overwrite the same slot while the previously returned TensorDict is still in user code.

The documentation only opts users into view lifetime semantics for `stack="lazy"`; an `env_index=...` receive aliases shared storage even with the default dense stack because it bypasses `_stack_func`. This can silently mutate an already-returned transition. Clone per-environment results before releasing the worker, or introduce an explicit lease/acknowledgement contract and document it for this path. Add a regression test that holds a per-environment result, sends the next command, and verifies the held result remains unchanged under the default stack mode.

## Low — Exchange metrics use ad hoc timing instead of the project timing facility

The new module imports `time` and records several `time.monotonic_ns()` intervals, for example in [`publish`/`publish_pair`](https://github.com/pytorch/rl/blob/41a188edc8e06ec0560c5557ac5b01856f3c11e4/torchrl/envs/_async_exchange.py#L154-L171). The contribution instructions direct timing through `torchrl.timeit`. If these user-facing metrics genuinely need private counters rather than the common timing facility, document that rationale; otherwise use the project helper to keep timing behavior and reporting consistent.

## Questions and residual risks

- Lazy-stack aliasing is explicitly documented as valid only until actions are sent back. That is a sharp lifetime contract; coverage should include misuse boundaries and per-environment calls, not just dense aggregate receives.
- `timeout` begins only after the first result and does not bound the wait to reach `min_get`. This matches the new documentation, but callers may read it as an overall deadline; naming or examples could make the hard lower-bound behavior more prominent.

theap06:
Addressed the feedback: per-environment reads now clone before releasing the worker, exchange timing uses torchrl.timeit, and the timeout plus lazy-view lifetime contracts are explicit and covered. Focused result: 4 passed.

torchrl/envs/async_envs.py
```diff
@@ -618,23 +699,45 @@ def async_step_send(
...
+            max_get,
+            timeout,
+            track_action=True,
+        )
         self._current_step = self._current_step - len(r)
+        if self._slot_exchange is not None:
+            idx = [item[0] for item in r]
+            result = self._slot_exchange.read(r, self._stack_func)
+            self._busy.difference_update(idx)
+            return result
```
vmoens, line 740: full-batch shared-memory step/reset reads bypass `_sort_results`. `_SharedSlotExchange.read` preserves queue arrival order, so workers with unequal latency silently permute row-to-environment alignment in synchronous results

torchrl/trainers/algorithms/configs/envs.py
```diff
@@ -32,9 +32,11 @@ class BatchedEnvConfig(EnvConfig):
     create_env_fn: Any = MISSING
     num_workers: int = 1
     create_env_kwargs: dict = field(default_factory=dict)
-    batched_env_type: str = "parallel"
+    batched_env_type: Literal["parallel", "serial", "async"] = "parallel"
     device: str | None = None
-    # batched_env_type: Literal["parallel", "serial", "async"] = "parallel"
+    backend: Literal["threading", "multiprocessing", "asyncio"] = "threading"
+    stack: Literal["dense", "maybe_dense", "lazy"] = "dense"
+    exchange: Literal["queue", "shm"] = "queue"
```
vmoens: the structured config uses Literal fields unsupported by the supported OmegaConf stack, reproducing ValidationError during config collection. Keep precise Literal types on public functions, but use supported config-field types plus explicit validation

vmoens (commented):
Still one apparent bug to fix
almost there

torchrl/envs/async_envs.py
theap06: Fixed in 8d17230d16. Shared descriptors are sorted by environment index before materialization, so full-batch reads match the queue path. Added a delayed-worker regression covering reset and step ordering.

torchrl/trainers/algorithms/configs/envs.py
theap06: Fixed in 8d17230d16. Config fields are plain strings for OmegaConf compatibility; make_batched_env keeps the Literal annotations and validates all four choices explicitly. Added schema and validation coverage.

theap06:
> Still one apparent bug to fix almost there

Thanks for the feedback! Shared-memory descriptors are now sorted by environment index before results are materialized. Also, I added a delayed-worker regression covering full-batch reset and step ordering.

vmoens (approved):
LGTM
Since this is a perf related feature we should make sure it's easily discoverable, well documented, well benchmarked, and that we have a plan to turn automatically to the fasted inference mode by default in a couple of releases

Issue #4061, opened 2026-08-05
[Feature Request] AsyncEnvPool v2


vmoens:
# [Feature] AsyncEnvPool v2: shared-memory slot exchange, deadline batch former, and a SEED-style inference path

## Motivation

`AsyncEnvPool` already has the right *semantics* for high-throughput acting: envs advance independently, `min_get` allows partial-batch harvesting, and `step_and_maybe_reset` inside the worker hides reset latency from the trainer. What limits it today is the *data plane*:

- Every step ships a tensordict through an `mp.Queue` (pickling + Python object churn),
- results carry `env_index` as `NonTensorData` and are re-assembled with `_sort_results` + `torch.stack` on every `recv` (a fresh allocation + copy per batch),
- there is no batch-forming policy (`min_get` is "at least N", with no max/deadline) and no instrumentation to tell whether a run is env-bound or inference-bound,
- the pool ends at one host, and the consumer half (gather → policy → scatter, overlapped with batch forming) is left entirely to user Python.

None of this matters for cheap Python envs on a single node — queue overhead vanishes under env step time. It matters a lot in the regime of many slow envs (sim gangs, game engines, browsers), pixel observations, and a policy pinned to a large accelerator. That is the SEED RL regime ([[Espeholt et al., 2019](https://arxiv.org/abs/1910.06591)](https://arxiv.org/abs/1910.06591)), and nothing in the PyTorch ecosystem currently serves it out of the box. I have seen an internal SEED-style stack built along the lines below sustain fleets of heavyweight simulator envs feeding a central learner; every item in this proposal is a generalization of something that measurably mattered there.

## Proposal

Ordered by expected payoff. Items 1–3 are the core; 4–6 build on them.

### 1. Slot-based shared-memory exchange (replace queues on the hot path)

Pre-allocate one contiguous shared (and optionally pinned) tensordict with a `num_envs` leading dimension — the request buffer — plus a matching response buffer for actions. Each worker owns slot `i`:

- worker writes its `(obs, step_type, reward)` record in place (`update_`) and raises a ready flag (semaphore / eventfd / lock-free ready ring);
- `recv` collects ready indices and returns a **view** into the request buffer plus the index list — no pickle, no sort, no stack;
- `send` scatters actions into the response buffer and signals the corresponding slots — no message back either.

[...]

PR #4110, opened 2026-08-15
[Feature] Add FixedBatchedInference helper (AsyncEnvPool v2 item 3)

coder-jayp:
## Description

Adds `FixedBatchedInference`, an opt-in consumer utility that bridges variable-size observation batches from `AsyncEnvPool.async_step_and_maybe_reset_recv` with fixed-shape policy inference.

Key features:
- Pads incoming batches to the nearest pre-defined bucket size (e.g. `[16, 32, 64]`)
- Injects a boolean `valid_mask` key to distinguish real vs padding rows
- Double-buffered pinned CPU staging to overlap H2D copy with GPU compute
- Async H2D transfer on a dedicated CUDA stream
- Returned CUDA outputs remain valid when a policy returns aliases or views of the reusable input buffer
- Lazy initialisation — no spec info required at construction time
- CPU fallback when CUDA is unavailable
- Context-manager support for prompt pinned-memory release

## Motivation and Context

Partial batches produced by `async_step_and_maybe_reset_recv(min_get=1, max_get=64, timeout=...)` break `torch.compile` and CUDA graphs, which require fixed input shapes. Users currently rebuild this logic manually in their training loops. This helper provides a standard, correct-by-default implementation.

Stacks on #4109 (shared-memory exchange + deadline batching). Closes part of #4061 (AsyncEnvPool v2, item 3).

## Types of changes

## Checklist


coder-jayp:
@torchrlbot reviewer @vmoens @theap06

vmoens:
@torchrlbot rebase

torchrl/envs/batched_inference.py
```diff
@@ -0,0 +1,227 @@
...
+
+        bucket = self._pick_bucket(B)
+        buf_idx = self._buf_idx[bucket]
+        staging = self._staging[bucket][buf_idx]
+        event = self._events[bucket][buf_idx]
+
+        if event is not None:
+            event.synchronize()
+
+        staging[:B].update_(batch)
```
theap06: `env_index` is non-tensor metadata, and `update_` does not preserve it here. I get `[0, 0, 0]` back from an AsyncEnvPool batch that started as `[0, 1, 2]`. Keep routing metadata out of staging and add it back to the result.

torchrl/envs/batched_inference.py
```diff
@@ -0,0 +1,227 @@
...
+            device_batch = staging.to(self.device)
+            output = self.policy(device_batch)
+
+        self._buf_idx[bucket] = (buf_idx + 1) % self._num_buffers
+
+        result = output[:B]
+        if self.add_valid_mask and self._MASK_KEY in result.keys():
+            result = result.exclude(self._MASK_KEY)
+
+        return result
```
theap06, line 350: The forward runs on `_stream`, but the caller stream never waits for it. Reading the result on another CUDA stream can return zeros. Add an event or `wait_stream` before returning.

torchrl/envs/batched_inference.py
```diff
@@ -0,0 +1,227 @@
...
+            mask[:B] = True
+            staging.set(self._MASK_KEY, mask, inplace=False)
+
+        if self._stream is not None:
+            with torch.cuda.stream(self._stream):
+                device_batch = staging.to(self.device, non_blocking=True)
+                new_event = torch.cuda.Event()
+                new_event.record(self._stream)
+                self._events[bucket][buf_idx] = new_event
+                output = self.policy(device_batch)
```
theap06, line 306: The example policy is still on CPU when `device_batch` is CUDA, so the documented path fails with a device mismatch. Move module policies to `device`, or require and show `policy.to(device)`.

torchrl/envs/batched_inference.py
```diff
@@ -0,0 +1,227 @@
...
+            staging[B:].zero_()
+
+        if self.add_valid_mask:
+            mask = torch.zeros(bucket, dtype=torch.bool)
+            mask[:B] = True
+            staging.set(self._MASK_KEY, mask, inplace=False)
+
+        if self._stream is not None:
+            with torch.cuda.stream(self._stream):
+                device_batch = staging.to(self.device, non_blocking=True)
```
theap06: `valid_mask` is added after the buffers are pinned, so that leaf stays pageable. Copy and forward also use the same stream, so H2D cannot overlap compute as described.

test/envs/test_batched_inference.py
```diff
@@ -0,0 +1,172 @@
...
+        assert self.helper._pick_bucket(1) == 8
+        assert self.helper._pick_bucket(9) == 16
+        assert self.helper._pick_bucket(17) == 32
+
+    def test_exceeds_max_raises(self):
+        with pytest.raises(ValueError, match="exceeds the largest bucket"):
+            self.helper._pick_bucket(33)
+
+
+class TestCPUHotPath:
```
theap06: These tests only hit CPU. Please add a `gpu`-marked CUDA test for stream handoff and device placement, plus an AsyncEnvPool test that checks `env_index` survives.

theap06 (changes requested):
Blocking issues in the AsyncEnvPool and CUDA paths. Reproductions are inline.

coder-jayp:
@theap06 All review comments have been fixed. Let me know what you think.

torchrl/envs/batched_inference.py
```diff
@@ -0,0 +1,283 @@
...
+                f"got batch_dims={batch.batch_dims}."
+            )
+        B = batch.batch_size[0]
+        if B == 0:
+            raise ValueError("Received an empty batch (B=0).")
+
+        if not self._initialized:
+            self._init_from_batch(batch)
+
+        # Separate tensor leaves from non-tensor metadata (e.g. env_index).
```
theap06: The recursive tensor keys are tuples and the batch.keys() returns the top-level key "nested". The code therefore mistakes the entire nested TensorDict for non-tensor metadata and reattaches the original value at /envs.

torchrl/envs/batched_inference.py
theap06: Waiting establishes execution order, but it does not tell the caching allocator that the tensor remains in use on another stream. If a policy returns a new action-only TensorDict, device_batch can go out of scope while compute is still reading it.

theap06 (changes requested):
Left some follow up comments. btw can you @pytest.mark.gpu as we don't have any CI workflows for cuda right now.

coder-jayp:
@theap06 Follow-up comments addressed (nested TensorDict fix, record_stream, @pytest.mark.gpu). All 35 tests pass including CUDA. Let me know what you think.

torchrl/envs/batched_inference.py
```diff
@@ -0,0 +1,308 @@
...
+        staging = self._staging[bucket][buf_idx]
+        copy_event = self._copy_events[bucket][buf_idx]
+
+        # Block until the H2D copy that read from this staging buffer is done,
+        # so it is safe to overwrite the pinned CPU memory.
+        if copy_event is not None:
+            copy_event.synchronize()
+
+        # Fill valid rows in-place with tensor leaves only (zero allocation).
+        staging[:B].update_(batch.select(*tensor_keys))
```
gtnv: the second call omitted aux. staging still passed the first call's sevens to every valid row. can we reject tensor-key changes before update_?

```suggestion
        expected_tensor_keys = set(staging.keys()) - (
            {self._MASK_KEY} if self.add_valid_mask else set()
        )
        if set(tensor_keys) != expected_tensor_keys:
            raise ValueError(
                f"Tensor keys changed after staging initialization: expected "
                f"{expected_tensor_keys}, got {set(tensor_keys)}."
            )
        staging[:B].update_(batch.select(*tensor_keys))
```

coder-jayp: Added the validation guard right before update_ to raise a ValueError if the keys don't match, and also included a regression test to ensure it catches changed keys properly.

vmoens:
Thanks for this @coder-jayp -- the stream/event choreography is correct (I traced the event and `record_stream` lifecycle) and the test coverage is solid. Before this merges I would like the data plane reworked in a few places so the helper actually delivers what the docstring promises, plus some smaller items. Detailed checklist:

**Blocking**

1. **CUDA graphs are not actually supported by the current data plane.** `staging.to(self.device, non_blocking=True)` allocates a fresh `device_batch` on every call, so a policy captured with CUDA graphs would replay against stale pointers. Please switch to persistent per-bucket device buffers written with `copy_(non_blocking=True)`. This also removes both `record_stream` calls (a known caching-allocator footgun: deferred frees and allocator-event churn) and lets the pre-allocated events be reused instead of constructing a new `torch.cuda.Event()` per call. Alternatively, soften the docstring to `torch.compile` only -- but the persistent-buffer version is strictly better and not much more code.
2. **The helper stages every tensor key in the incoming batch.** A batch from `async_step_and_maybe_reset_recv` carries `done`/`terminated`/`truncated`/`reward` and the previous `action` alongside the observation; all of it is memcpy'd into pinned staging and shipped H2D on every call. Read `policy.in_keys` when available (fall back to a `select_keys` kwarg) and stage only those.
3. **Test the target combination.** `test_async_env_pool_routes_inference_by_env_index` uses `backend="threading"`; the pipeline this helper exists for is `backend="multiprocessing", exchange="shm"`. Please add that variant.

**Non-blocking but requested**

4. The H2D/compute overlap claim only materializes with a pipelined consumer: in the canonical `recv -> helper -> send` loop the caller syncs to bring actions back to CPU before `send`, so by the next call the previous compute is already done. Double buffering today only overlaps the CPU staging write. Fine to keep the mechanism, but the docstring should say this plainly; the pipelined API (submit batch N, collect actions N-1) is planned as a follow-up on the exchange work.
5. `staging[B:].zero_()` rewrites the full padding region on every call; with pixel observations and a half-empty bucket that is a large host memset. Track a high-water mark and zero only rows dirtied beyond the current batch (or make zeroing opt-out).
6. Repo conventions: please fold the tests into `test/envs/test_special.py` (existing AsyncEnvPool coverage lives there) and add a benchmark entry -- `benchmarks/test_envs_benchmark.py` is growing AsyncEnvPool series you can extend.
7. Naming/docs: `reset()` is confusing next to env semantics -- `close()` reads better; document that the largest bucket must cover the pool's `max_get`.

Measured context for 2 and 5: on the CPU path the helper adds ~33 us/call for small vector obs and ~154 us/call for 64x64x3 float pixels (B=48, bucket 64), most of it staging traffic that items 2 and 5 remove.

Happy to push these changes to the branch if you prefer.

coder-jayp:
@vmoens Pulled your commits and verified locally—all tests pass (including CUDA). Thanks for pushing the fixes directly. LGTM.

vmoens:
Note for the record: no commits landed on this branch after my review above (the last commit predated it), so the earlier "pulled your commits / LGTM" exchange did not correspond to any pushed fixes. Per my offer, I have now pushed the rework directly to this branch (main merged in first):

- Persistent per-bucket device buffers written with `copy_(non_blocking=True)` replace the per-call `.to()` allocation. With `double_buffer=False` every call reuses the same device storage, which is what manual CUDA-graph capture requires. Buffer reuse is guarded by a compute-done event enqueued on the copy stream (no host block); input `record_stream` calls and per-call event allocations are gone.
- Only the keys the policy reads are staged and transferred (`select_keys` kwarg > `policy.in_keys` > all tensor keys). Staged inputs are no longer echoed in the output, since they would alias reused device buffers.
- Padding is zeroed to a per-buffer high-water mark; `reset()` is renamed `close()` and synchronizes the helper streams before releasing buffers; docstrings now state precisely what is and is not delivered (compile vs graphs vs overlap).
- The pool integration test now also covers `backend="multiprocessing", exchange="shm"` -- the combination this helper exists for -- plus new key-selection tests, and a benchmark entry tracks the helper's per-call overhead.

One checklist item I dropped deliberately: folding the tests into `test_special.py` would conflict with two in-flight PRs editing that file; the dedicated test file stays.

All 15 CPU tests pass locally; the 4 CUDA tests (including a new device-buffer-persistence test) need the GPU CI legs.

coder-jayp:
@vmoens my apologies for the git mix-up . Thanks for taking over and pushing this massive rework. I pulled your squashed commit and ran the test suite locally. I pushed a quick fix for a CUDA stream race condition during device buffer initialization to ensure the new CUDA test passes reliably on GPU. 

With this fix, all 15 CPU tests and all 4 CUDA tests pass perfectly. 

LGTM

PR #4113, opened 2026-08-15
[BugFix] Map sparse masked modes to original actions

### LINKED ISSUE #4112: [BUG] Sparse MaskedCategorical mode uses compressed indices
## Describe the bug

`MaskedCategorical` with sparse `indices` reports its mode in the compressed index space rather than the original action space. The result can be an action excluded by the mask, and `log_prob(dist.mode)` can consequently be `-inf`. `MaskedOneHotCategorical` has the same mapping problem and also returns a one-hot vector whose width is the number of valid indices rather than the original number of actions.

## To Reproduce

```python
import torch
from torchrl.modules.distributions import MaskedCategorical, MaskedOneHotCategorical

logits = torch.tensor([0.0, 1.0, 10.0, 2.0])
indices = torch.tensor([0, 2])

for cls in (MaskedCategorical, MaskedOneHotCategorical):
    dist = cls(logits=logits, indices=indices)
    print(cls.__name__, dist.mode, dist.log_prob(dist.mode), dist.sample())
```

On current `main` (`6e18e35b2`), this prints a categorical mode of `1` with log-probability `-inf`, while samples correctly map back to action `2`. The one-hot mode is `[0, 1]` instead of `[0, 0, 1, 0]`.

## Expected behavior

`mode` and `deterministic_sample` should use the same original action space as `sample`: `2` for `MaskedCategorical` and `[0, 0, 1, 0]` for `MaskedOneHotCategorical`. Their log-probabilities should be finite.

## System info

- Installed from source at `6e18e35b2`
- Python 3.13.3
- PyTorch 2.11.0+cu130 (CPU reproduction)
- Windows 11

## Reason and Possible fixes

Sparse construction gathers logits to the valid-index subset. The inherited categorical mode is therefore an offset into that gathered tensor. `sample()` maps this offset back through `self._mask`, but the mode properties do not. Mapping the categorical mode through the sparse indices, then creating the one-hot result with `num_samples` classes, restores consistency.

## Checklist

- [x] I have checked that there is no similar issue in the repo
- [x] I have read the documentation
- [x] I have provided a minimal working example to reproduce the bug

## Generative-tool disclosure

I used OpenAI Codex to assist with repository auditing, reproduction, duplicate searches, and drafting this report. I reviewed the reproduced behavior and proposed scope.

aswanth-07:
## Description

Sparse `MaskedCategorical` distributions gather logits into a compact list of valid actions, but their deterministic mode was returned as an offset into that compact list. This maps the mode back through the sparse indices, matching the action space used by `sample()`. `MaskedOneHotCategorical` now builds its sparse mode with the original action count as well.

The regression tests cover batched sparse indices for both categorical forms and verify `mode`, `deterministic_sample`, output shape, and finite `log_prob(mode)`.

## Motivation and Context

Fixes #4112.

Before this change, a four-action distribution restricted to actions `[0, 2]` could report action `1` as its mode even though that action was masked and had log-probability `-inf`. The one-hot variant returned a two-element vector instead of a four-element action.

## Types of changes

## Checklist

- [x] I have read the CONTRIBUTING guide
- [ ] My change requires a change to the documentation
- [x] I have updated the tests accordingly
- [ ] I have updated the documentation accordingly

## Validation

- `py -3.13 -m pytest test/test_distributions.py -q -k "TestMaskedCategorical or TestMaskedOneHotCategorical"`: 68 passed
- `py -3.13 -m pytest test/test_distributions.py -q -k "not test_tanhnormal_rsample_and_log_prob"`: 962 passed, 16 deselected
- Relevant pre-commit hooks on both changed files: passed (mixed-line-ending, ufmt, flake8, pydocstyle, pyupgrade, codespell, autoflake, docstring checks)
- `git diff --check`: passed

A full `test/test_distributions.py` run completed with 970 passed and 8 unrelated failures in `TestTanhNormal.test_tanhnormal_rsample_and_log_prob`; this Windows PyTorch build fails during TorchDynamo CUDA/Triton device discovery (`IndexError` in `get_device_properties`) before exercising this patch.

## Generative-tool disclosure

I used OpenAI Codex to assist with repository auditing, reproduction, duplicate searches, implementation, and test execution. I reviewed the complete diff and the affected sparse-mode mapping.

### DIFF
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -881,6 +881,16 @@ class TestMaskedCategorical:
         sample_probs = torch.bincount(samples) / num_samples
         torch.testing.assert_close(sample_probs, ref_probs, rtol=1e-5, atol=1e-2)
 
+    def test_sparse_mode_uses_original_indices(self) -> None:
+        logits = torch.tensor([[0.0, 1.0, 10.0, 2.0], [0.0, 9.0, 1.0, 8.0]])
+        indices = torch.tensor([[0, 2], [1, 3]])
+        dist = MaskedCategorical(logits=logits, indices=indices)
+        expected = torch.tensor([2, 1])
+
+        torch.testing.assert_close(dist.mode, expected)
+        torch.testing.assert_close(dist.deterministic_sample, expected)
+        assert torch.isfinite(dist.log_prob(dist.mode)).all()
+
     @pytest.mark.parametrize("neg_inf", [-1e20, float("-inf")])
     @pytest.mark.parametrize("sparse", [False, True])
     @pytest.mark.parametrize("ndim", [2, 1, 3])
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM thanks

PR #4114, opened 2026-08-15
[BugFix] Fix DiffusionActor sampling and multidimensional training

theap06:
## Summary: 
Fixes multidimensional batch handling in `DiffusionBCLoss` and corrects `DiffusionActor` behavior for reduced-precision and deterministic inference.

Also keeps diffusion schedule calculations in float32 for numerical stability.

### DIFF
--- a/test/modules/test_actor.py
+++ b/test/modules/test_actor.py
@@ -502,8 +502,9 @@ class TestDiffusionActor:
             assert p.grad is not None
 
     @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
-    def test_dtype(self, dtype):
+    def test_reduced_precision_schedule(self, dtype):
         actor = DiffusionActor(action_dim=2, obs_dim=3, num_steps=3).to(dtype)
+        assert actor.module.alphas_cumprod.dtype is torch.float32
         td = TensorDict({"observation": torch.randn(4, 3, dtype=dtype)}, batch_size=[4])
         action = actor(td)["action"]
         assert action.dtype is dtype
@@ -520,6 +521,25 @@ class TestDiffusionActor:
                 TensorDict({"observation": observation.clone()}, batch_size=[4])
             )["action"]
         torch.testing.assert_close(action0, action1, rtol=0, atol=0)
+
+    @pytest.mark.parametrize(
+        ("dtype", "num_steps"),
+        [(torch.bfloat16, 258), (torch.float16, 2050)],
+    )
+    def test_timestep_dtype_resolution(self, dtype, num_steps):
+        actor = DiffusionActor(action_dim=2, obs_dim=3, num_steps=num_steps).to(dtype)
+        td = TensorDict({"observation": torch.randn(4, 3, dtype=dtype)}, [4])
+        with pytest.raises(ValueError, match="cannot be represented"):
+            actor(td)
+
+    def test_schedule_buffers_are_authoritative(self):
+        actor = DiffusionActor(action_dim=2, obs_dim=3, num_steps=3)
+        actor.module.alphas_cumprod.fill_(1)
+        clean_action = torch.randn(4, 2)
+        noisy_action, _ = actor.module.add_noise(
+            clean_action, torch.zeros(4, dtype=torch.long)
+        )
+        torch.testing.assert_close(noisy_action, clean_action)
 
 
 @pytest.mark.parametrize("device", get_default_devices())
--- a/test/modules/test_actor.py
+++ b/test/modules/test_actor.py
@@ -501,6 +501,26 @@ class TestDiffusionActor:
         for p in actor.parameters():
             assert p.grad is not None
 
+    @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
+    def test_dtype(self, dtype):
+        actor = DiffusionActor(action_dim=2, obs_dim=3, num_steps=3).to(dtype)
+        td = TensorDict({"observation": torch.randn(4, 3, dtype=dtype)}, batch_size=[4])
+        action = actor(td)["action"]
+        assert action.dtype is dtype
+        assert action.isfinite().all()
+
+    def test_deterministic(self):
+        actor = DiffusionActor(action_dim=2, obs_dim=3, num_steps=3)
+        observation = torch.randn(4, 3)
+        with set_exploration_type(ExplorationType.DETERMINISTIC):
+            action0 = actor(
+                TensorDict({"observation": observation.clone()}, batch_size=[4])
+            )["action"]
+            action1 = actor(
+                TensorDict({"observation": observation.clone()}, batch_size=[4])
+            )["action"]
+        torch.testing.assert_close(action0, action1, rtol=0, atol=0)
+
 
 @pytest.mark.parametrize("device", get_default_devices())
 def test_actorcritic(device):
--- a/test/objectives/test_diffusion_bc.py
+++ b/test/objectives/test_diffusion_bc.py
@@ -4,6 +4,8 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import annotations
 
+import argparse
+
 import pytest
 import torch
 from tensordict import TensorDict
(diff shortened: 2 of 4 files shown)


vmoens (commented):
Thanks!

Here are some comments

## High — the loss regresses on tensordicts with an empty batch size that hold batched data

[`diffusion_bc.py#L143-L160`](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/torchrl/objectives/diffusion_bc.py#L143-L160)
replaces `batch_size = clean_action.shape[0]` with `batch_shape = tensordict.batch_size`. A
`TensorDict` constructed without an explicit `batch_size` gets `torch.Size([])` even when every leaf
is batched, which is a very common way to hand-build a batch.

Concrete failure (both variants run locally against the head diff, applied by monkey-patching the
two methods onto the installed package):

```python
td = TensorDict({"observation": torch.randn(8, 4), "action": torch.randn(8, 2)})  # batch_size=[]
loss_fn(td)
```

- current `main`: returns a finite scalar loss (verified, `0.6607`).
- this PR: `RuntimeError: Tensors must have same number of dimensions: got 2 and 1` — `t` is 0-d, so
  `t_float` is 1-d while `noisy_action` is 2-d and the `torch.cat` at
  [L160](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/torchrl/objectives/diffusion_bc.py#L160)
  fails.

The same expression also still fails for a time dimension (`batch_size=[8]` with action `(8, 5, 2)`),
before and after this PR.

Suggested fix: derive the shape from the action instead of the tensordict.

```python
batch_shape = clean_action.shape[:-1]
```

I ran all five shape cases against that variant — `bs=[]` unbatched, `bs=[8]`, `bs=[2,3]`,
`bs=[]` with batched leaves, and `bs=[8]` with `(B, T, A)` — and all five pass, i.e. it is a strict
superset of both the old and the new behaviour. The new `test_batch_shapes` case would keep passing;
please also add the `batch_size=[]`-with-batched-data case as a regression test.

## Medium — bfloat16 collapses the timestep conditioning for large `num_steps`

`t_tensor` at
[`actors.py#L2814`](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/torchrl/modules/tensordict_module/actors.py#L2814)
and `t_float` at
[`diffusion_bc.py#L159`](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/torchrl/objectives/diffusion_bc.py#L159)
now inherit the data dtype where they were previously pinned to float32. Measured round-trip of
`torch.arange(n)` through bfloat16:

| `num_steps` | distinct bf16 values | distinct fp16 values |
|---|---|---|
| 100 (default) | 100 / 100 | 100 / 100 |
| 512 | 385 / 512 | 512 / 512 |
| 1000 (canonical DDPM) | 507 / 1000 | 1000 / 1000 |

Failure scenario: `DiffusionActor(action_dim=..., num_steps=1000).to(torch.bfloat16)` feeds the same
conditioning value for several adjacent timesteps (e.g. 996-1003 all collapse), at both training and
inference time, so the score network cannot tell the noise levels apart and denoising degrades
silently — no exception, just worse actions. float16 is safe up to 2048.

The cast itself is necessary (mixed-dtype `torch.cat` promotes to float32 and then breaks the half
score network), so the fix belongs elsewhere: either document the constraint and validate
`num_steps` against the module dtype, or stop passing a raw integer timestep (a normalized or
sinusoidal embedding computed in float32 and cast once would keep resolution).

## Medium — the deterministic mode now starts from an off-manifold latent, undocumented

[`actors.py#L2807-L2811`](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/torchrl/modules/tensordict_module/actors.py#L2807-L2811)
substitutes `torch.zeros` for `torch.randn` when the interaction type is `DETERMINISTIC`. This does
fix reproducibility (verified: identical outputs after the patch, differing outputs on `main`), but
`x_T = 0` is not a draw from the `N(0, I)` prior the reverse chain was trained against — the prior
mass concentrates on `||x|| ≈ sqrt(action_dim)`, so the score network is evaluated off-distribution
at the first steps. For small action dims this is benign; for larger ones the "deterministic" action
is not the mode of the learned policy in any defensible sense.

Both docstrings still describe the mode only as skipping noise injection
([`_DDPMModule.forward`](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/torchrl/modules/tensordict_module/actors.py#L2786-L2800)
and the `DiffusionActor` class docstring), so the zero-latent start is invisible to users. Please
document it, and consider offering a seeded `torch.Generator` as the reproducibility mechanism
instead of (or alongside) the zero start.

## Low — `alphas` and `alphas_cumprod` buffers are now dead

[`actors.py#L2741-L2743`](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/torchrl/modules/tensordict_module/actors.py#L2741-L2743)
still registers both buffers, but after this PR every read goes through
[`_schedule`](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/torchrl/modules/tensordict_module/actors.py#L2745-L2753),
which recomputes them from `betas`. Consequences: a subclass or user that overwrites
`alphas_cumprod` (e.g. a cosine schedule) is now silently ignored, and `torch.cumprod` runs on every
`forward`/`add_noise` call. Note also that the float32 promotion only recovers what `betas` still
holds — after `.to(torch.float16)` the stored `beta_start` is already rounded to `9.998e-05`.
Either keep the buffers authoritative (upcast them in `_schedule` instead of recomputing) or store
`beta_start`/`beta_end` and rebuild the schedule in float32; dropping the buffers outright would
break `load_state_dict(strict=True)` on existing checkpoints.

## Low — test-file conventions and coverage gaps

- [`test/objectives/test_diffusion_bc.py`](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/test/objectives/test_diffusion_bc.py#L160-L164)
  has no `if __name__ == "__main__": pytest.main(...)` block (CLAUDE.md 7). Pre-existing, but this
  PR extends the file and could fix it in passing.
- `DiffusionBCLoss._AcceptedKeys` types `action`/`observation` as `NestedKey`, and this PR reworks
  exactly the shape logic driven by those entries, yet `test_custom_keys` only exercises flat
  strings. CLAUDE.md 7 asks for a nested-key case.
- The new
  [`test_batch_shapes`](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/test/objectives/test_diffusion_bc.py#L121-L136)
  sits next to the pre-existing `test_batch_sizes`; the two names are hard to tell apart. Consider
  `test_multidim_and_scalar_batch`.

## Low — `ci/olddeps` label is probably required

[`test_dtype`](https://github.com/pytorch/rl/blob/f370370775db07fe87d4de5f5cf8fa67c4fca3cb/test/modules/test_actor.py#L504-L510)
runs a CPU float16 `nn.Linear`. The `tests-olddeps` suite pins `torch==2.1.1`
(`.github/unittest/linux_olddeps/scripts_gym_0_13/install.sh:42`) and its "remainder" shard collects
`test/modules/test_actor.py`. CPU half `addmm` was not implemented in older torch releases
(`"addmm_impl_cpu_" not implemented for 'Half'`); I could not verify whether 2.1.1 has it, since the
only local interpreter carries torch 2.11. Per CLAUDE.md 7b the safe move is to add the
`ci/olddeps` label before the next push so the signal appears on the PR instead of post-merge.

theap06:
Addressed the review feedback and added reduced-precision coverage.

theap06:
Addressed the review points: action-derived batch shapes now cover implicit and time-dimension batches, reduced-precision timestep limits are validated, deterministic zero-latent behavior is documented, schedule buffers stay authoritative, and nested-key/file-entry coverage was added. Focused result: 35 passed.

vmoens (approved):
LGTM! Thanks

PR #4118, opened 2026-08-16
[BugFix] Fix dreamerv3 GPU usage, loss tensors in memory

### LINKED ISSUE #4117: [BUG] Dreamerv3 walker doesn't use available GPU, and retains loss tensors
## Describe the bug

The DreamerV3 SOTA walker implementation uses the CPU for all training by default, even when a local GPU is available. The DMC env should run on CPU, but the models, losses, and policy can run on the accelerator.

The trainnig loop also keeps detached loss tensors in memory, 5 scalar tensors per optimisation update. Walker uses ~1M updates, which, if using GPU, makes all these CUDA tensors stay in memory till training ends

## To Reproduce

Install and run on a machine with GPU
```bash
python sota-implementations/dreamer_v3/dreamer_v3.py \
  --config-name=config_dmc_walker
```

Open a second terminal. Monitor the active CUDA processes:

```bash
watch -n 0.5 \
  'nvidia-smi --query-compute-apps=pid,process_name,used_gpu_memory --format=csv'
```

GPU usage stays low, training stays on the CPU

The training loop adds detached tensors to lists after each optimization
update:

```python
loss_hist["kl"].append(model_kl.detach())
loss_hist["reco"].append(m_td["loss_model_reco"].detach())
loss_hist["reward"].append(m_td["loss_model_reward"].detach())
loss_hist["actor"].append(a_td["loss_actor"].detach())
loss_hist["value"].append(v_td["loss_value"].detach())
```

The `detach()` operation removes a tensor from the autograd graph. It does not
move the tensor to the CPU. It does not release the tensor storage.

## Expected behavior

The real DMC env runs on CPU, the models, losses, policy, replay samples run on an available accelerator by default. User can select CPU for training if needed

Only a limited amount of loss data is kept on the GPU. Keep only full CPU loss history is plot is needed

## Screenshots

Not applicable

## System info

```text
Python: 3.11.16
TorchRL: 0.14.0+ga46e872e
PyTorch: 2.15.0.dev20260816+cu130
TensorDict: 0.14.0+g8e4daaa
NumPy: 2.4.6
dm-control: 1.0.44
MuJoCo: 3.11.0
Platform: Linux
GPU: NVIDIA A100 80GB PCIe
```

```python
import sys

import numpy
import torch
import torchrl

print(torchrl.__version__, torch.__version__, numpy.__version__)
print(sys.version, sys.platform)
print(torch.cuda.is_available(), torch.cuda.device_count())
```

## Additional context

Raising a PR to fix this

## Reason and Possible fixes

The config uses env device as the collector and training device. Training script doesnt move the training modules to an auto selected accelerator. Use separate settings for the real env and training

- an `optimization.device` setting. Use `null` to select an available
  accelerator.
- Keep the real collection and evaluation environments on the CPU.
- Set separate `policy_device`, `env_device`, and `storing_device` values for
  the collector.
- Move the models, losses, and sampled replay batches to the optimization
  device.
- Set `auto_cast_to_device=True` for evaluation rollouts.
- Create `DreamerEnv` on the optimization device. Its base constructor calls
  `world_model.to(self.device)` on shared model parameters.
- Store the loss metrics in one tensor for each batch. Move the loss history to
  the CPU only when the code must make a plot.

## Checklist

unography:
## Description

Fixes https://github.com/pytorch/rl/issues/4117

## Motivation and Context

Allows Dreamer v3 walker to use an available accelerator by default
Avoids loss tensors from staying in CUDA memory for the whole training duration

## Types of changes

## Checklist

### DIFF
--- a/sota-implementations/dreamer_v3/README.md
+++ b/sota-implementations/dreamer_v3/README.md
@@ -21,6 +21,11 @@ of length 64, a replay ratio of 1024, and 1.1 million environment steps. It logs
 evaluation return against environment steps to JSON so curves can be compared
 without relying on wall-clock-dependent training iterations.
 
+Real collection and evaluation environments run on CPU; `optimization.device`
+selects where the models, losses and policy run and defaults to `null`, which
+auto-selects an available accelerator. Pass `optimization.device=cpu` to force
+CPU execution.
+
 For a three-seed median and interquartile reproduction run:
 
 ```bash
--- a/sota-implementations/dreamer_v3/config.yaml
+++ b/sota-implementations/dreamer_v3/config.yaml
@@ -3,7 +3,6 @@ env:
   name: Pendulum-v1
   task: null
   seed: 0
-  device: cpu
   max_episode_steps: 200
 
 collector:
@@ -41,6 +40,7 @@ networks:
   policy_max_std: 1.0
 
 optimization:
+  device:            # training device (models, losses, policy). null=auto
   lr: 4.0e-5
   adam_eps: 1.0e-20
   adaptive_grad_clip: 0.3
--- a/sota-implementations/dreamer_v3/dreamer_v3.py
+++ b/sota-implementations/dreamer_v3/dreamer_v3.py
@@ -41,7 +41,7 @@ from tensordict.nn import (
     TensorDictSequential,
 )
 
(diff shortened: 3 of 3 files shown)


vmoens (approved):
LGTM thanks!

PR #4121, opened 2026-08-17
[Performance] Lock TransformedEnv specs by default

mathieuorhan:
## Summary

- Prevent transform spec hooks from mutating specs owned by their parent environment.
- Lock `TransformedEnv` specs by default while keeping spec materialization lazy.
- Cache derived spec values such as reward keys and `_step_mdp` when specs are stable.
- Preserve explicit opt-outs through `spec_locked=False` and dynamic-spec behavior through `cache_specs=False`.

## Why this is needed

Environment specs are executable contracts, not just metadata. TorchRL uses them to discover observation, reward, and done keys; allocate rollout tensordicts; validate environment outputs; and configure operations such as `step_mdp`. If a transform changes the wrong spec, the visible failure can occur much later than the mutation and look unrelated to the transform.

A transform spec hook receives a cloned input spec and must describe only the effect of that transform. Several hooks instead read from and wrote to `self.parent.output_spec`. Because `TransformedEnv` specs were unlocked by default, these invalid writes did not fail at the mutation site: they silently changed the contract of the upstream environment.

This could produce user-visible correctness bugs in composed environments:

- `StepCounter` and truncation transforms could add or rewrite done and truncated entries on the parent spec. The parent could then advertise keys that it does not actually emit, leading to rollout allocation or environment-spec validation mismatches.
- Clipping and reward-sign transforms could replace reward bounds on the parent spec. Code inspecting the upstream environment could then see post-transform bounds even though the parent still emits untransformed rewards.
- Reusing a corrupted parent spec in another spec computation could make results depend on which transformed spec was accessed first.
- Leaving transformed specs unlocked disabled the existing derived-value cache, so helpers such as reward-key discovery and `_step_mdp` were rebuilt repeatedly on hot paths.

Locking specs makes accidental mutation fail immediately and enables safe caching. Updating the affected hooks to mutate only their provided clone removes the underlying parent-corruption bug.

## Implementation

- Record the requested lock state at construction and apply it when input and output specs materialize.
- Update the affected clipping, reward-sign, step-counter, and truncation hooks to modify and return their input specs.
- Keep transient specs locked when `cache_specs=False`, while bypassing derived-value caching so transforms with changing specs continue to recompute keys correctly.
- Invalidate spec and derived-value caches when transforms are appended.
- Honor both constructor-level and post-construction spec-lock opt-outs.

## User-facing behavior

The default behavior now matches other environments: materialized specs are locked. Users with intentionally dynamic transform specs can continue to use `cache_specs=False`, and users who need mutable specs can opt out with `spec_locked=False` or `set_spec_lock_(False)`.

## Testing

Added regression coverage for:

- deferred locking with lazy spec materialization;
- caching of derived spec values;
- cache invalidation after appending a transform;
- constructor-level and post-construction lock opt-outs;
- lazy `ObservationNorm` initialization;
- rebuilt transform parents;
- dynamic transform specs with `cache_specs=False`;
- parent spec immutability.

Local validation of the affected suites:

- `test/transforms/test_compose_and_env.py`
- `test/envs/test_env_base.py`
- 235 passed, 20 skipped.

### DIFF
--- a/test/transforms/test_compose_and_env.py
+++ b/test/transforms/test_compose_and_env.py
@@ -204,12 +204,42 @@ class TestTransformedEnv:
         assert "other_count" in env.observation_spec.keys()
         assert env.is_spec_locked
 
-    def test_spec_lock_opt_out(self) -> None:
-        env = TransformedEnv(ContinuousActionVecMockEnv(), StepCounter())
-        env.set_spec_lock_(False)
+    @pytest.mark.parametrize("set_after_init", [False, True])
+    def test_spec_lock_opt_out(self, set_after_init: bool) -> None:
+        env = TransformedEnv(
+            ContinuousActionVecMockEnv(), StepCounter(), spec_locked=set_after_init
+        )
+        if set_after_init:
+            env.set_spec_lock_(False)
         assert not env.is_spec_locked
         assert not env.observation_spec.is_locked
         assert env._step_mdp is not env._step_mdp
+
+    def test_spec_lock_without_spec_cache(self) -> None:
+        class DynamicKeyTransform(Transform):
+            enabled = False
+
+            def transform_observation_spec(
+                self, observation_spec: TensorSpec
+            ) -> TensorSpec:
+                if self.enabled:
+                    observation_spec["dynamic"] = Unbounded(shape=(1,))
+                return observation_spec
+
+        transform = DynamicKeyTransform()
+        env = TransformedEnv(
+            ContinuousActionVecMockEnv(), transform, cache_specs=False
+        )
+
+        assert env.is_spec_locked
+        assert env.observation_spec.is_locked
+        assert env.input_spec.is_locked
+        assert "dynamic" not in env.observation_keys
+
+        transform.enabled = True
+
+        assert "dynamic" in env.observation_keys
+        assert env.observation_spec.is_locked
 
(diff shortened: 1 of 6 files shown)


vmoens:
I pushed commit 785f0312d to address the spec-lock review findings.

Changes made:
- honor the constructor-level spec_locked argument instead of always locking TransformedEnv;
- lock transient input and output Composite specs when cache_specs=False;
- bypass derived-value caching when cache_specs=False, preserving dynamic transform spec updates;
- cover constructor and setter opt-outs, transient spec locking, and dynamic observation-key refreshes with regression tests.

Validation:
- 8 focused spec-lock tests passed;
- test/transforms/test_compose_and_env.py and test/envs/test_env_base.py: 235 passed, 20 skipped;
- git diff --check passed.

The pre-commit environment could not bootstrap its pinned libcst because a Rust compiler is unavailable on this machine.

vmoens (approved):
LGTM Thanks!

vmoens:
Ran some benchmarks 
Great improvement image

mathieuorhan:
Thanks @vmoens! I'm sorry I couldn't add the description yesterday, GH was down most of the day.

vmoens:
> Thanks @vmoens! I'm sorry I couldn't add the description yesterday, GH was down most of the day.

I know, I had to hang out on X instead 🫩

PR #4100, opened 2026-08-13
[Feature] Wire value transforms into actor-critic losses

vmoens:
## Summary

- add a shared `value_transform` contract to scalar value estimators and DDPG, TD3, continuous SAC v1/v2, PPO, ClipPPO, and KLPENPPO
- keep critic outputs in transformed prediction space while Bellman arithmetic, actor objectives, entropy terms, returns, advantages, PPO clipping, and diagnostics remain in raw units
- propagate the exact transform object to lazy/prebuilt estimators, add Hydra config support, and reject concurrent MAPPO/IPPO ValueNorm scaling
- document the end-to-end dataflow and add transformed eager/compiled objective benchmarks

## Key semantics

- `state_value` / `state_action_value`: transformed prediction space
- `advantage` / `value_target`, explicit estimator `next_value=`, PPO `clip_value`, and diagnostic values: raw reward-return space
- replay `td_error`: existing absolute/squared convention, computed from the transformed residual
- `value_transform=None`: existing identity path, with no added tensor operations
- unsupported: DiscreteSAC, distributional/two-hot critics, and concurrent ValueNorm/PopArt scaling

## Validation

- `pytest test/objectives/test_values.py test/objectives/test_ddpg.py test/objectives/test_sac.py test/objectives/test_ppo.py test/objectives/test_mappo.py test/compile/test_compile_value.py -k 'not prioritized_weights'`: 5,211 passed, 2,316 skipped
- targeted Hydra nested-instantiation and config parity checks: 9 passed
- transformed benchmark smoke tests: eager and compiled forward/backward passed for DDPG, TD3, SAC, and trajectory-shaped PPO/GAE
- guide examples executed end to end
- flake8, pydocstyle, docstring-argument, codespell, and Sphinx-section checks passed

The prioritized-replay tests excluded above require the optional segment-tree extension, which is unavailable in this local environment; transformed priority equations are covered directly in the DDPG/TD3/SAC regression tests.

## Local CPU benchmark snapshot

Representative eager forwards (mean):

| Objective | None | symlog | Delta |
| --- | ---: | ---: | ---: |
| DDPG | 1.247 ms | 1.309 ms | +5.0% |
| TD3 | 2.885 ms | 3.109 ms | +7.8% |
| SAC | 2.847 ms | 2.925 ms | +2.7% |
| PPO/GAE | 2.693 ms | 2.701 ms | +0.3% |

[...]

PR #4126, opened 2026-08-18
[Feature] Add PercentileValueNorm and refactor DreamerV3 retnorm onto it

vmoens:
## Description

Second PR of the return-normalization stack (based on #4125; merge that first, then rebase this onto `main`).

DreamerV3's 5th-95th percentile return scaling was private to `DreamerV3ActorLoss`. This PR promotes it to a reusable primitive in the existing `ValueNorm` family, next to `PopArtValueNorm` and `RunningValueNorm`:

- **`PercentileValueNorm`**: EMA of a low/high quantile pair (`quantiles=(0.05, 0.95)`, `rate=0.01` by default), with the normalisation scale clamped from below (`scale = max(min_scale, high - low)`), following DreamerV3 (https://arxiv.org/abs/2301.04104). By default it is scale-only (`center=False`), which is the DreamerV3 recipe for advantages (already centred by the value baseline); `center=True` additionally subtracts the low-percentile EMA, mapping the tracked range onto `[0, 1]`. Quantiles are tracked per trailing element (`shape`), so per-agent statistics work the same way as in the other normalisers.
- **`ValueNorm.scale()`**: new interface method (implemented by all three normalisers) returning the multiplicative scale applied by `normalize`. Consumers that rescale advantages need the scale without the centering; DreamerV3 uses it here, and the follow-up PPO/A2C PR uses it as well.
- **`DreamerV3ActorLoss` refactor**: the loss now holds a `PercentileValueNorm` submodule (`self.retnorm`) instead of private `return_low`/`return_high` buffers and a hand-rolled quantile EMA. `return_low`, `return_high`, `return_normalization_rate`, `return_normalization_quantiles`, and `return_normalization_min_scale` remain accessible as read-through properties, so the constructor signature, produced losses, diagnostics, and logging are unchanged. Since MAPPO/IPPO already accept any `ValueNorm` via `value_norm=`, they can now use percentile normalisation for the critic with no further changes.

Note on checkpoints: state-dict keys move from `return_low`/`return_high` to `retnorm.low`/`retnorm.high` (shape `(1,)` instead of 0-dim). DreamerV3 has not shipped in a release, so no deprecation cycle applies; flagging it for anyone with in-flight training runs on main.

## Tests

- `TestPercentileValueNorm` in `test/objectives/test_mappo.py` (where the ValueNorm family is tested): EMA tracking of exact batch quantiles, min-scale clamping on degenerate spans, scale-only vs center semantics with denormalize round-trips, per-element (per-agent) shapes, and argument validation.
- `TestValueNormBase.test_scale_is_normalize_slope`: interface contract `normalize(a) - normalize(b) == (a - b) / scale()` for all three normalisers.
- Existing DreamerV3 retnorm tests (including the torch.compile check on `_return_scale` and the state-dict round-trip) pass unchanged: `test/objectives/test_mappo.py` + `test/objectives/test_dreamer_v3.py`: 95 passed.

## Stack

1. #4125 (base: `main`)
2. this PR (base: `dreamerv3-retnorm-reparam`)
3. `ppo-advantage-norm` - opt-in percentile advantage scaling for PPO/A2C

Related: #4100 wires the stateless invertible `ValueTransform` family (symlog et al.) into losses in prediction space; this stack is the complementary stateful normaliser in raw return space. No file overlap.

### DIFF
--- a/torchrl/modules/value_norm.py
+++ b/torchrl/modules/value_norm.py
@@ -14,6 +14,11 @@ implementations:
 - :class:`RunningValueNorm` — exact Welford running mean / variance with no
   decay. Cheaper and more stable when value targets are stationary; tends to
   be the better default for shorter / non-curriculum runs.
+- :class:`PercentileValueNorm` — exponential-moving-average of a low / high
+  quantile pair, rescaling by the clamped span between them (the DreamerV3
+  "return normalisation", Hafner et al. 2023,
+  https://arxiv.org/abs/2301.04104). Scale-only by default, which suits
+  advantage scaling.
 
 Plug any subclass into :class:`~torchrl.objectives.multiagent.MAPPOLoss` (or
 your own actor-critic loss) via ``value_norm=...``.
@@ -73,6 +78,15 @@ class ValueNorm(nn.Module, metaclass=ABCMeta):
     @abstractmethod
     def denormalize(self, normalised_value: torch.Tensor) -> torch.Tensor:
         """Inverse of :meth:`normalize` — recover real-scale values."""
+
+    @abstractmethod
+    def scale(self) -> torch.Tensor:
+        """Multiplicative scale currently applied by :meth:`normalize`.
+
+        Exposed separately so consumers can rescale quantities that must not
+        be re-centred, e.g. advantages (already centred by the value
+        baseline), for which only the division by the scale applies.
+        """
 
     # ------------------------------------------------------- shared helpers
 
@@ -161,6 +175,10 @@ class PopArtValueNorm(ValueNorm):
         mean, var = self._running_stats()
         return normalised_value * var.sqrt() + mean
 
+    def scale(self) -> torch.Tensor:
+        _, var = self._running_stats()
+        return var.sqrt()
+
 
 class RunningValueNorm(ValueNorm):
     """Exact running mean / variance (Welford's online algorithm).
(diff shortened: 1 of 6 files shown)


torchrl/modules/value_norm.py
```diff
@@ -74,6 +79,15 @@ def normalize(self, value_target: torch.Tensor) -> torch.Tensor:
...
         """Inverse of :meth:`normalize` — recover real-scale values."""
 
+    @abstractmethod
+    def scale(self) -> torch.Tensor:
+        """Multiplicative scale currently applied by :meth:`normalize`.
+
+        Exposed separately so consumers can rescale quantities that must not
+        be re-centred, e.g. advantages (already centred by the value
+        baseline), for which only the division by the scale applies.
+        """
```
gtnv: legacy subclasses implement update, normalize, and denormalize. they no longer instantiate because scale() is now abstract. can scale() stay non-abstract for a compatibility window?

torchrl/objectives/dreamer_v3.py
```diff
@@ -698,21 +702,11 @@ def __init__(
...
-        self.return_normalization_rate = return_normalization_rate
-        self.return_normalization_quantiles = return_normalization_quantiles
-        self.return_normalization_min_scale = return_normalization_min_scale
-        self.register_buffer("return_low", torch.tensor(0.0))
-        self.register_buffer("return_high", torch.tensor(0.0))
+        self.retnorm = PercentileValueNorm(
+            quantiles=return_normalization_quantiles,
+            rate=return_normalization_rate,
+            min_scale=return_normalization_min_scale,
+        )
```
gtnv, line 720: old checkpoints still contain return_low and return_high. they now report missing retnorm.low/high and unexpected old keys. can we migrate them before registering retnorm?

PR #4127, opened 2026-08-18
[Feature] Opt-in percentile advantage scaling for PPO and A2C

vmoens:
## Description

Third PR of the return-normalization stack (based on #4126; merge #4125 then #4126 first, then rebase this onto `main`).

Adds an `advantage_norm: ValueNorm | None` keyword argument to `PPOLoss` (inherited by `ClipPPOLoss`, `KLPENPPOLoss`, and through them MAPPO/IPPO) and to `A2CLoss`:

- In training mode, the normaliser's statistics are updated with the batch **value targets** once per forward call; the advantage is then divided by `advantage_norm.scale()` with no re-centering (the advantage is already centred by the value baseline).
- With `PercentileValueNorm` (from #4126) this is exactly the DreamerV3 return-normalization recipe (https://arxiv.org/abs/2301.04104): an EMA of the 5th-95th return-percentile span with `max(1, span)` clamping, which scales large returns down without amplifying small or noisy ones. Unlike mean/std advantage standardization (`normalize_advantage=True`), the percentile span is robust to heavy-tailed advantage distributions - the failure mode where a single outlier batch collapses the scale of every other advantage.
- Any `ValueNorm` works: the new `scale()` interface method is all that is used, so `PopArtValueNorm`/`RunningValueNorm` give an EMA/Welford std-based scale-only variant.
- `normalize_advantage=True` and `advantage_norm` are mutually exclusive (constructor `ValueError`).
- If the advantage is precomputed and the value target is absent from the input tensordict, the loss raises a clear `KeyError` instead of silently skipping the update.

Refactor: the three identical mean/std normalization blocks in `PPOLoss.forward`, `ClipPPOLoss.forward`, and `KLPENPPOLoss.forward` are factored into a shared `_maybe_normalize_advantage` helper. Behavior with `advantage_norm=None` is byte-for-byte unchanged, and the feature adds no per-call overhead when unset.

Config parity (CLAUDE.md section 14): `advantage_norm` fields on `PPOLossConfig` and `A2CLossConfig`, instantiated by `_make_ppo_loss` / `_make_onpolicy_loss` when given a config object; the automated `TestConfigClassParity` suite passes.

Scope note: A2C had no advantage-normalization pathway before this PR (no `normalize_advantage` there), so for A2C this is a new opt-in rather than an extension. GRPO subclasses `LossModule` directly with its own forward and is deliberately untouched.

## Tests

In `test/objectives/test_ppo.py`:
- `test_ppo_advantage_norm` (parametrized over the three PPO losses) and `test_a2c_advantage_norm`: the loss with `advantage_norm` equals the reference loss with a manually rescaled advantage, and the EMA buffers land on the exact batch quantiles of the value target.
- Eval mode leaves the statistics untouched; mutual exclusion raises; missing value target raises; a nested `value_target` key set via `set_keys` is honored.
- `test/objectives/test_ppo.py` + `test/objectives/test_mappo.py` + `test/objectives/test_loss_module.py`: 766 passed. Config parity subset of `test/test_configs.py`: 10 passed. (The `TestPPO` class is skipped locally for lack of `transformers`; the new PPO tests were additionally executed directly and pass.)

## Stack

1. #4125 (base: `main`)
2. #4126 (base: `dreamerv3-retnorm-reparam`)
3. this PR (base: `percentile-value-norm`)

### DIFF
--- a/torchrl/objectives/ppo.py
+++ b/torchrl/objectives/ppo.py
@@ -39,6 +39,7 @@ from torchrl.objectives.utils import (
     _maybe_add_or_extend_key,
     _maybe_get_or_select,
     _sum_td_features,
+    _valid_value_target_rows,
     _validate_clip_epsilon,
     dispatch_value_estimator,
     distance_loss,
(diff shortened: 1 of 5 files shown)


torchrl/objectives/ppo.py
```diff
@@ -961,6 +975,36 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+            raise KeyError(
+                f"advantage_norm requires the value target "
+                f"({self.tensor_keys.value_target!r}) in the input tensordict "
+                "to update the return statistics. Let the loss compute the "
+                "advantage with its value estimator, or write the value "
+                "target under that key."
+            )
+        if self.training:
+            self.advantage_norm.update(value_target)
+        return advantage / self.advantage_norm.scale()
```
gtnv: a masked 1e9 value enters the stats. two forwards also update the same rollout twice. can we use only valid rows and update once per rollout?

PR #4128, opened 2026-08-18
[BugFix] Match DreamerV3 reference semantics

unography:
## Summary

This PR is now the focused DreamerV3 correctness patch. The end-to-end reproduction, performance work, default changes, and broad test rewrite have been extracted so the semantic fixes can be reviewed on their own.

Retained fixes:

- reset the previous action together with RSSM state and belief at episode boundaries, including nested action keys;
- use the DreamerV3 RMS normalization without an additive offset;
- initialize block-linear recurrent kernels from the full input fan-in rather than independently per block;
- expose the separate dynamics/representation KL formulation and categorical unimix without changing compatibility defaults in this PR;
- include the root continuation in imagination discount weights;
- train the critic on real replay sequences with a closed-form replay target and configurable detached outputs;
- document the paper as the normative algorithmic reference while identifying later author-maintained JAX choices separately.

The original correctness commits remain authored by @unography. The final small refactor removes unrelated changes, corrects the replay-data wording, and replaces the oversized test rewrite with focused regression tests.

## Extracted follow-ups

- #4132 isolates RSSM fast paths, compilation, mixed precision, and benchmarks.
- #4133 isolates the pinned JAX Walker Walk reproduction and explicitly compares its protocol with the paper.
- #4134 is a draft community discussion that prototypes canonical public defaults before implementing TorchRL's required deprecation schedule.

## Scope reduction

The original mixed change was 18 files and roughly 7,864 changed lines. This PR is now 5 files and 476 changed lines.

## Test plan

```text
PYTHONPATH=. python -m pytest -q \
  test/modules/test_dreamer_components.py \
  test/objectives/test_dreamer_v3.py
275 passed
```

Formatting checks and `git diff --check` also pass.

### DIFF
--- a/sota-implementations/dreamer_v3/dreamer_v3.py
+++ b/sota-implementations/dreamer_v3/dreamer_v3.py
@@ -57,7 +57,7 @@ from torchrl.envs.transforms import (
 from torchrl.envs.utils import ExplorationType, set_exploration_type
 from torchrl.modules import DreamerV3MLP, SymExpTwoHot, WorldModelWrapper
 from torchrl.modules.distributions.continuous import TanhNormal
-from torchrl.modules.models.model_based_v3 import (
+from torchrl.modules.models.dreamer_v3 import (
     RSSMPosteriorV3,
     RSSMPriorV3,
     RSSMRolloutV3,
--- a/test/objectives/test_dreamer_v3.py
+++ b/test/objectives/test_dreamer_v3.py
@@ -715,13 +715,17 @@ class TestDreamerV3(LossModuleTestBase):  # type: ignore[misc]
     # Value loss tests
     # ------------------------------------------------------------------ #
 
-    def test_dreamer_v3_replay_value_target(self, device):
+    @pytest.mark.parametrize("compiled", [False, True])
+    def test_dreamer_v3_replay_value_target(self, device, compiled):
         reward = torch.tensor([[0.0, 1.0, 2.0, 3.0]], device=device)
         bootstrap = torch.tensor([[10.0, 20.0, 30.0, 40.0]], device=device)
         done = torch.zeros_like(reward, dtype=torch.bool)
         terminated = torch.zeros_like(done)
 
-        target = _replay_value_target(
+        target_fn = _replay_value_target
+        if compiled:
+            target_fn = torch.compile(target_fn, backend="eager", fullgraph=True)
+        target = target_fn(
             reward,
             done,
             terminated,
--- a/test/objectives/test_dreamer_v3.py
+++ b/test/objectives/test_dreamer_v3.py
@@ -46,6 +46,7 @@ from torchrl.objectives import (
 from torchrl.objectives.dreamer_v3 import (
     _default_bins,
     _match_trailing_dim,
+    _replay_value_target,
     categorical_kl_balanced,
     categorical_kl_terms,
     symexp,
@@ -591,6 +592,18 @@ class TestDreamerV3(LossModuleTestBase):  # type: ignore[misc]
         }
         self.tensordict_keys_test(loss_fn, default_keys=default_keys)
 
+    @pytest.mark.parametrize("detach_output", [True, False])
+    def test_dreamer_v3_model_loss_detach_output(self, device, detach_output):
+        world_model = self._create_world_model().to(device)
+        loss_module = DreamerV3ModelLoss(
+            world_model,
+            num_reward_bins=self.num_reward_bins,
+            detach_output=detach_output,
+        )
+        _, features = loss_module(self._create_world_model_data().to(device))
+        posterior = features["next", "posterior_logits"]
+        assert posterior.requires_grad is not detach_output
+
     # ------------------------------------------------------------------ #
     # Actor loss tests
     # ------------------------------------------------------------------ #
@@ -664,7 +677,8 @@ class TestDreamerV3(LossModuleTestBase):  # type: ignore[misc]
         loss_td, fake_data = loss_module(
             self._create_actor_data().to(device).reshape(-1)
         )
-        expected_weight = torch.tensor([1.0, 0.5, 0.25], device=device)
+        # The initial state is weighted by its own continuation probability.
+        expected_weight = torch.tensor([0.5, 0.25, 0.125], device=device)
         torch.testing.assert_close(
             fake_data["discount_weight"][0, :, 0], expected_weight
         )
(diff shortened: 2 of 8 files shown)


unography:
@vmoens made some changes to match the JAX implementation. Currently ran for smaller duration, 16k steps, will run for longer 200k steps and update figures
wanted to get eyes on it early if the direction seems correct

test/objectives/test_dreamer_v3.py
vmoens, line 1: Over 1700 LoC here, can we shrink?

sota-implementations/dreamer_v3/dreamer_v3.py
vmoens, line 1: Over 1700 LoC here too

vmoens (commented):
Thanks, the code is concentrated in two big edits. Can we simplify/assess if code can be distributed/chunked/slimmed?

vmoens:
@unography I rechecked the five correctness areas against the branch after reducing it, and the focused patch still covers them:

1. RSSM resets now mask the previous action as well as state and belief; focused tests cover both flat and nested action keys.
2. DreamerV3 RMS normalization no longer has an additive offset.
3. Block-linear recurrent initialization now uses the full input fan-in and is invariant to the block partition.
4. Imagination discount weights now include the root continuation; the regression test checks the exact cumulative sequence.
5. The critic now has the real-replay training term and corrected replay lambda target, with gradient and nested-key coverage.

I also corrected the documentation that called real replay states "on-policy" and documented the paper/current-JAX distinction in this PR.

To keep this review small, I moved useful independent work to:

- #4132: RSSM fast paths, compile/mixed-precision handling, benchmark, and focused parity tests.
- #4133: pinned JAX Walker reproduction. Its body records the extraction fixes: correct LaProp-style optimizer wording, `optimizer_eps` naming, explicit paper-versus-JAX protocol, and repository-style `__init__` signatures.
- #4134: draft prototype and community discussion for canonical constructor defaults, to be converted to the required deprecation schedule if accepted.

The original PR is now 5 files / 476 changed lines instead of 18 files / roughly 7,864 changed lines. The two focused test files pass: 275 tests.

vmoens:
@unography I investigated the replay value recurrence against TorchRL value functionals. It is exactly a one-step-shifted TD(lambda) return for valid episode data: done stops lambda propagation, while terminated suppresses the bootstrap. Commit 5e4d09188 now delegates to td_lambda_return_estimate and keeps only the Dreamer-specific shift/shape adapter.

I did not add an associative-scan core change to this PR. TorchRL already has the core API; its scalar implementation captures as a full graph, while the existing vectorized implementation is intentionally not full-graph compilable because trajectory splitting uses data-dependent shapes. PyTorch associative_scan is still a prototype, its optimized mode is CUDA/compile-only, and TorchRL supports versions back to 2.1. That performance work needs its own benchmark-backed design once the operator is sufficiently portable, rather than being coupled to DreamerV3.

Validation:
- all 63 DreamerV3 objective tests pass
- eager and full-graph results agree for replay fields shaped [B,T] and [B,T,1]
- ufmt, flake8, autoflake, and git diff checks pass for the changed files

vmoens:
@unography Follow-up commit 5333a888d aligns DreamerV3ValueLoss with the standard loss reduction API:

- adds reduction with none, mean, and sum modes; mean remains the default
- applies it consistently to imagined value loss, slow-critic diagnostics, and replay value loss
- uses the keyword TensorDict output constructor
- preserves the existing replay boundary weighting semantics

All 69 DreamerV3 objective tests pass, including per-element output shapes and backward coverage for every reduction. Formatting, flake8, pydocstyle, pyupgrade, codespell, autoflake, docstring, and diff checks pass.

vmoens:
@unography Commit 3858a2130 refactors DreamerV3 RMS normalization to torch.nn.functional.rms_norm on PyTorch 2.4 and newer, while retaining the explicit formula through implement_for for TorchRL supported versions 2.1 through 2.3. The FP32 compute and output-dtype restoration are preserved, including learned scale behavior.

Validation:
- native and explicit formulations are bit-for-bit identical locally for FP32, FP64, FP16, and BF16, with matching gradients
- full-graph forward and backward pass for FP32, FP16, and BF16
- 211 module tests pass; the remaining local Inductor test is blocked by a stale macOS precompiled-header cache unrelated to this change
- formatting and all available lint checks pass

The ci/olddeps label was added before the push so PyTorch 2.1 exercises the fallback.

vmoens:
@unography I renamed torchrl/modules/models/model_based_v3.py to torchrl/modules/models/dreamer_v3.py in dbf4eafbb. The previous name was too generic and version-suffixed; dreamer_v3.py describes the module directly without merging nearly 1,000 lines of V3-specific categorical RSSM code into the older model_based.py module. All first-party imports and examples now use the new path. The path shipped in TorchRL 0.13 remains available through a package-level compatibility alias, so existing imports and pickle lookups keep working without a duplicate shim file. GitHub recognizes this as a 99% rename (31 insertions, 23 deletions). The complete configured linter passes; 280 focused tests passed locally. The remaining compile test was blocked by a local stale TorchInductor PCH timestamp error, unrelated to the rename, and will be covered by CI.

vmoens:
@unography Follow-up performance fix in 4daf91160: the replay critic target now uses vec_td_lambda_return_estimate in eager mode and falls back to td_lambda_return_estimate only during Dynamo tracing. The fallback is necessary because the vectorized implementation discovers episode lengths and splits/pads trajectories with data-dependent shapes, which currently fails fullgraph compilation. On the DMC replay shape B=16, T=63, the vectorized path measured 64-89 microseconds versus 684-695 microseconds for the scalar recurrence on CPU (about 8-11x faster, with and without internal episode boundaries). At T=15 it is slightly slower, so this is specifically valuable for the length-64 DMC preset. The existing target test now covers both eager and torch.compile(fullgraph=True) execution. The full DreamerV3 objective suite passes (70 tests), as does the complete configured linter.

vmoens:
@unography Follow-up cleanup in c02fa4cde: _replay_value_target is now decorated with torch.no_grad() instead of building the complete lambda-return autograd graph and detaching only the final tensor. This preserves the target's non-differentiable contract and gradients through the critic prediction path while reducing saved autograd intermediates. The full DreamerV3 objective suite passes (70 tests, including the eager and fullgraph-compiled target cases), and the complete configured linter passes.

vmoens (approved):
LGTM

vmoens:
Updated the olddeps fix after discussion in 65c7a5531. Direct `implement_for` dispatch remains on `_DreamerV3RMSNorm.forward`; compile compatibility with old PyTorch is not a target. The block-GRU compile regression test now runs only on Torch >=2.4, where it exercises the native `F.rms_norm` path. Verified with a full-graph trace on PyTorch 2.11, 281 remaining Dreamer tests, and the complete configured linter.

PR #4132, opened 2026-08-20
[Performance] Speed up DreamerV3 RSSM rollout

vmoens:
## Stack

1. **#4132 — RSSM performance and recurrence backends (this PR)**
2. #4133 — pinned JAX Walker reproduction
3. #4134 — canonical-defaults prototype

The branches are chained linearly; merge this PR first.

## Summary

This extracts the RSSM performance work from #4128 so it can be reviewed independently from the correctness fixes that have now landed.

The changes:

- add tensor-level RSSM prior, posterior, and rollout paths that avoid TensorDict bookkeeping inside the recurrent loop;
- implement a genuine `torch._higher_order_ops.scan` backend through the same `hoptorch` compatibility path as TorchRL's RNN modules;
- lift RSSM parameters once as scan closure inputs, instead of materializing time-expanded parameter and buffer views as scanned inputs;
- preserve loop/scan forward and backward parity, including reset handling and identical categorical draws for a fixed seed;
- retain the explicit eager loop as the production default and expose compiled loop-step and compiled higher-order-scan modes explicitly because CPU speed and CUDA speed/memory favor different choices;
- make compiled scan unrolling configurable, including non-divisible sequence lengths, while keeping the library default at `1`;
- consolidate all model-based components in `torchrl/modules/models/model_based.py`, while preserving the former import paths as module aliases;
- rename the SOTA entrypoint to `sota-implementations/dreamer_v3/train.py`;
- fold the RSSM benchmark into the existing recurrent-backend benchmark suite instead of keeping a standalone Dreamer benchmark file;
- keep categorical sampling and sensitive normalization/loss calculations in FP32 under mixed precision.

## Loop versus higher-order scan

Benchmark shape: batch 16, sequence length 64, belief width 512. Each measurement is a warm steady-state full TensorDict rollout call. "Compiled loop" uses `compile_rollout("step")`; "compiled scan" uses `compile_rollout("scan")` and compiles the higher-order operation. Times below are means.

### A10G CUDA

Environment: PyTorch `2.15.0.dev20260821+cu126`, A10G. The targeted upstream run and all eight cases passed: [workflow run 32489864701](https://github.com/pytorch/rl/actions/runs/32489864701).

| Execution | Inference | Forward + backward | Inference peak allocation delta | Training peak allocation delta |
|---|---:|---:|---:|---:|
| eager loop | 90.1 ms | 216.2 ms | 12.34 MiB | 43.24 MiB |
| compiled loop step | 49.9 ms | 134.6 ms | 12.34 MiB | 36.35 MiB |
| eager higher-order scan | 92.7 ms | 1875.1 ms | 6.96 MiB | 272.62 MiB |
| compiled higher-order scan | **22.3 ms** | **97.2 ms** | **7.63 MiB** | 210.18 MiB |

On CUDA, compiled scan is 2.24x faster than compiled loop-step for inference and 1.38x faster for forward + backward. It also uses less incremental inference memory, but its incremental training peak is 5.78x larger. Eager scan is not suitable for training.

### Compiled scan unrolling on A10G

A follow-up A10G sweep compared manual unroll factors at the same shape. All 14 targeted RSSM inference/training cases passed in [workflow run 32559624244](https://github.com/pytorch/rl/actions/runs/32559624244); the overall benchmark workflow was red only because unrelated storage benchmarks failed against nightly TensorDict.

| Scan unroll | Inference | Forward + backward | Inference peak allocation delta | Training peak allocation delta |
|---:|---:|---:|---:|---:|
| 1 | 22.78 ms | 101.00 ms | 7.63 MiB | 210.18 MiB |
| 2 | 22.86 ms | 105.56 ms | 7.84 MiB | 120.66 MiB |
| 4 | 19.30 ms | 98.08 ms | 8.20 MiB | 76.41 MiB |
| 8 | **18.01 ms** | **93.04 ms** | 8.95 MiB | **58.89 MiB** |

Relative to `unroll=1`, `unroll=8` is about 21% faster for inference and 8% faster for forward + backward, while reducing incremental training allocation by 72%. It does increase the compiled body and made cold compilation/warmup roughly four times longer in this run. `compile_rollout("scan", unroll=...)` therefore remains configurable and defaults to `1`; the stacked Dreamer reproduction PR selects `8` for long CUDA training runs. Padding and trimming inside the final chunk support horizons that are not divisible by the unroll factor.

### macOS arm64 CPU

Environment: PyTorch 2.11.0, Apple arm64 CPU.

| Execution | Inference | Forward + backward |
|---|---:|---:|
| eager loop | 22.1 ms | 53.5 ms |
| compiled loop step | **24.0 ms** | **72.5 ms** |
| eager higher-order scan | 21.8 ms | 606.7 ms |
| compiled higher-order scan | 25.3 ms | 112.3 ms |

On this CPU, compiled scan is 5.6% slower for inference and 54.9% slower for forward + backward than compiled loop-step. This is why the PR keeps the backend selectable instead of declaring one universal default: compiled scan is the clear CUDA throughput choice, while compiled loop-step is the better CPU and lower-training-memory choice.

The initial scan implementation produced misleadingly poor compiled results because it expanded every RSSM parameter across the time dimension and passed those views as scanned inputs. TorchRL's RNN scan implementation closes over weights and lets the higher-order operator lift them once. Matching that implementation reduced the local compiled-scan training time from roughly 255 ms to 112 ms and exposed the CUDA speedup above. The scan path also uses `hoptorch` to probe and, when needed, patch native scan backward before compilation.

`associative_scan` is not a valid backend for this recurrence. It requires an associative binary combine operation. The Dreamer RSSM step is a nonlinear, parameterized block-GRU transition followed by observation-conditioned posterior logits and a stochastic straight-through categorical sample; arbitrary trajectory segments cannot be summarized as the same fixed-size associative operator without changing the model.

## Test plan

```text
python -m pytest test/modules/test_dreamer_components.py -q -k 'not compile'
213 passed, 3 deselected

python -m pytest test/modules/test_dreamer_components.py -q \
  -k 'rssm_rollout_compile or higher_order_scan_matches_loop'
5 passed, 213 deselected

python -m pytest test/objectives/test_dreamer_v3.py -q \
  -k 'real_rssm or sota_shares_imagination_parameters'
2 passed, 68 deselected

python -m pytest \
  benchmarks/test_rnn_reset_backends_benchmark.py::test_rssm_rollout_backend \
  -q --benchmark-min-rounds=5 --benchmark-warmup=off
the final unroll-8 inference/training smoke passed locally; all 14 backend/unroll cases passed on A10G CUDA
```

The compile test uses native Inductor for both loop-step and higher-order-scan scopes and covers forward and backward. On this Mac I disabled Inductor's precompiled-header optimization for the run to bypass a stale global `.pch` mtime; this changes compilation latency only, not measured steady-state execution. Exact ufmt, flake8, pydocstyle, pyupgrade, codespell, autoflake, docstring-argument checks, and `git diff --check` pass.


vmoens:
Correction to my earlier CPU-only conclusion: I had not tested the RSSM scan path on GPU, and the initial implementation did not follow the existing RNN scan pattern closely enough.

The first version passed time-expanded views of every RSSM parameter and buffer as scanned inputs. I have replaced that with the RNN approach: close over module weights, let the higher-order operator lift them once, and use the same `hoptorch` scan backward probe/patch path.

Targeted A10G results with PyTorch `2.15.0.dev20260821+cu126` (batch 16, T=64, belief=512, warm full-rollout means):

| Execution | Inference | Forward + backward | Training peak allocation delta |
|---|---:|---:|---:|
| compiled loop step | 49.9 ms | 134.6 ms | 36.35 MiB |
| compiled higher-order scan | **22.3 ms** | **97.2 ms** | 210.18 MiB |

So compiled scan is 2.24x faster for inference and 1.38x faster for training on this GPU. The important counterweight is 5.78x higher incremental training allocation. All eight eager/compiled loop/scan inference/training cases passed in [the targeted CUDA run](https://github.com/pytorch/rl/actions/runs/32489864701). The PR body now contains the complete CPU/CUDA timing and memory tables.

PR #4134, opened 2026-08-20
[Deprecation] Make DreamerV3 defaults canonical

vmoens:
## Stack

1. #4132 — RSSM performance
2. #4133 — pinned JAX Walker reproduction
3. **#4134 — canonical-defaults prototype (this PR)**

This branch is based directly on #4133 and adds one defaults-focused commit. Merge #4132 and #4133 first.

## Summary

This draft makes the public DreamerV3 loss constructors select the canonical paper/JAX recipe without extra keyword arguments:

- separate dynamics and representation KL losses;
- 1% categorical uniform mixture;
- REINFORCE actor gradients;
- two-hot critic loss;
- slow-critic regularization with weight 1.

Historical behavior remains available explicitly through `kl_mode="balanced"`, `unimix=0`, `use_reinforce=False`, `value_loss="symlog_mse"`, and `slow_critic_regularization=0`.

This discussion came out of the review and extraction of #4128.

## Paper versus JAX

My proposed rule is:

1. Document the paper as the normative algorithmic contract.
2. Document later JAX choices in named, pinned presets such as #4133.
3. Use paper-canonical public defaults when the paper is unambiguous; do not silently inherit every subsequent JAX experiment as an API requirement.

For these five settings, the paper and the current JAX recipe point in the same direction. The JAX-specific environment size, step count, replay ratio, optimizer schedule, RNG, and logging choices stay in the named SOTA preset rather than becoming loss API defaults.

## Community feedback requested

This is deliberately a draft prototype, not merge-ready deprecation code. Changing these defaults directly would violate TorchRL's compatibility policy. Before replacing this implementation with staged warnings and versioned default changes, I would like feedback on:

- Is the five-setting default bundle the right canonical contract?
- Should any setting remain compatibility-first even after a deprecation cycle?
- Is explicit legacy configuration sufficient, or is a named compatibility preset/factory warranted?
- Should all five defaults migrate together, or should model, actor, and critic defaults move separately?

Once the direction is agreed, the direct switches in this draft should be converted to the repository's two-minor-release warning/default-change schedule with explicit target versions.

## Test plan

```text
PYTHONPATH=. python -m pytest -q test/objectives/test_dreamer_v3.py
71 passed
```

### DIFF
--- a/torchrl/objectives/dreamer_v3.py
+++ b/torchrl/objectives/dreamer_v3.py
@@ -227,15 +227,15 @@ class DreamerV3ModelLoss(LossModule):
             non-terminal continuation targets, for encoding the finite-horizon
             discount in the continuation model. Defaults to 1.0.
         kl_mode ("balanced" or "separate", optional): KL formulation.
-            ``"balanced"`` preserves the historical weighted aggregate;
-            ``"separate"`` emits the reference dynamics and representation
-            losses. Defaults to ``"balanced"``.
+            ``"balanced"`` selects the historical weighted aggregate;
+            ``"separate"`` emits the paper's dynamics and representation
+            losses. Defaults to ``"separate"``.
         lambda_dynamic (float, optional): Dynamics KL weight in separate mode.
             Defaults to 1.0.
         lambda_representation (float, optional): Representation KL weight in
             separate mode. Defaults to 0.1.
         unimix (float, optional): Uniform mixture used by the categorical KL
-            distributions. Defaults to 0.0 for compatibility.
+            distributions. Defaults to 0.01.
         kl_alpha (float, optional): KL balancing factor (alpha in the paper).
             Default: 0.8.
         free_bits (float, optional): Minimum KL per categorical in nats.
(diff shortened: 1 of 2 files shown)


torchrl/objectives/dreamer_v3.py
```diff
@@ -339,10 +339,10 @@ def __init__(
         lambda_reco: float = 1.0,
         lambda_reward: float = 1.0,
         lambda_continue: float = 0.0,
-        kl_mode: Literal["balanced", "separate"] = "balanced",
+        kl_mode: Literal["balanced", "separate"] = "separate",
```
theap06, line 351: changing the value_loss to two_hot requires 255-bin value head and the previous accepted a scalar head. I think for now the best move would be to preserve the existing defualts for this release and raise some warnings in the next release for people leveraging these.

theap06 (changes requested):

torchrl/objectives/dreamer_v3.py
vmoens: Yeah this PR was to keep @unography param changes somewhere but I wouldn't merge this PR just now.

In general before 0.14 we need a bunch of training curves with dreamer v3 to make sure we're good with it.
Some check on perf and throughput would also be valuable.

unography: these were the training curves I got when i raised my earlier PR
walker walk on 200k steps, 3 seeds
combined plot, then separately for all 3, and then comparing combined with the JAX reference  
the `main` branch at that point was too slow to run for the full 200k
  
can run more if needed  
  

walker_run
fixed_seed0_200k
fixed_seed1_200k
fixed_seed2_200k

jax200k

vmoens:
@unography @theap06 what's our take on this PR? Merge?

unography:
> @unography @theap06 what's our take on this PR? Merge?

looks good to me, closer to the paper. someone using this would expect to match the paper's defaults. I'll have some GPUs over the weekend where I can do some training runs, get perf metrics, etc - if we want to do that first before merging

theap06:
> @unography @theap06 what's our take on this PR? Merge?

Sorry for late reply! I think for this pr to be merged the approach should be to just stage the default flips behind the futurewarnings.

PR #4137, opened 2026-08-21
[BugFix] Preserve TruncatedNormal sample dtype

### LINKED ISSUE #4136: [BUG] TruncatedNormal rsample upcasts float16 inputs to float32
## Describe the bug

`TruncatedNormal.rsample()` upcasts `float16` parameters to `float32`, even when `loc`, `scale`, and truncation bounds all use `float16`. The same upcast occurs with the default scalar bounds. This can silently change the action dtype in mixed-precision policies.

## To Reproduce

```python
import torch
from torch.distributions import Normal
from torchrl.modules import TruncatedNormal

loc = torch.zeros(3, dtype=torch.float16)
scale = torch.ones_like(loc)
low = torch.full_like(loc, -1)
high = torch.full_like(loc, 1)

print(Normal(loc, scale).rsample().dtype)
print(TruncatedNormal(loc, scale).rsample().dtype)
print(TruncatedNormal(loc, scale, low=low, high=high).rsample().dtype)
```

Current output on pristine `main`:

```text
torch.float16
torch.float32
torch.float32
```

## Expected behavior

Both `TruncatedNormal` samples should preserve the promoted parameter/bound dtype and return `torch.float16` in this example, matching `torch.distributions.Normal`.

## System info

- Installed from a source checkout at `b0eab87eb17ac6a910758444e716992739e1440b`
- Python 3.13.3
- PyTorch 2.11.0+cu130
- TensorDict 0.13.0
- Windows 11

## Additional context

The problem affects both scalar and tensor bounds. `float32` and `float64` inputs already return samples in their expected dtype.

## Reason and Possible fixes

Two constructors introduce the upcast:

1. The public distribution converts scalar `low` and `high` without inheriting `loc.dtype`.
2. The standard truncated-normal sampler creates its inverse-CDF probabilities with `torch.empty(..., device=...)`, which uses the global default dtype rather than the normalized bound dtype.

Using `loc.dtype` for scalar bounds and a dtype-preserving tensor constructor for the probability samples fixes the issue while retaining normal PyTorch promotion for explicitly typed tensor bounds.

## Checklist

- [x] I have checked that there is no similar issue in the repo
- [x] I have read the documentation
- [x] I have provided a minimal working example to reproduce the bug

aswanth-07:
## Description

Preserve the promoted dtype of samples returned by `TruncatedNormal.rsample()`.

The upcast had two independent sources:

- scalar `low` and `high` bounds were materialized in the global default dtype instead of `loc.dtype`;
- the inverse-CDF sampler created its uniform probabilities with `torch.empty`, also using the global default dtype.

This change gives scalar bounds the location dtype and creates probability samples from the normalized bound tensor. Explicit tensor bounds retain their own dtype, so normal PyTorch promotion semantics are preserved. The regression test covers scalar and tensor bounds across `float16`, `float32`, and `float64`, and checks finite samples plus finite same-dtype gradients.

## Motivation and Context

Closes #4136.

Mixed-precision policies can use `TruncatedNormal` to generate bounded actions. Returning `float32` from `float16` parameters silently changes the action dtype and diverges from `torch.distributions.Normal`. The fix is applied at the two dtype-introducing constructors rather than casting the final sample, so explicitly wider parameters or tensor bounds still promote naturally.

## Types of changes

## Checklist

- [x] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide

## Validation

- The new six-case regression failed on pristine `main` for both `float16` paths before the fix.
- `py -3.13 -m pytest test/test_distributions.py -q` — 984 passed; one local warning that the optional TorchRL C++ extension is unavailable.
- All changed-file pre-commit hooks passed. The Windows-incompatible `python` invocation for `check-docstring-args` was also run directly with `py -3.13` and passed.
- `py -3.13 -m py_compile` passed for all three changed files.
- Manual mixed-dtype checks verified that wider `scale` or explicit tensor bounds still promote the output, while all-`float16` inputs remain `float16`.
- `git diff --check` passed.

## Notes

A `torch.compile` smoke test was attempted but is not counted as passing: the local Windows host has no MSVC `cl`, and full-graph eager compilation reaches an existing `Tensor.tolist()` graph break in `TruncatedStandardNormal.__init__` before sampling.

### DIFF
--- a/torchrl/modules/distributions/continuous.py
+++ b/torchrl/modules/distributions/continuous.py
@@ -244,11 +244,11 @@ class TruncatedNormal(D.Independent):
 
         if not isinstance(high, torch.Tensor):
             high = torch.as_tensor(high, device=self.device, dtype=loc.dtype)
-        else:
+        elif high.device != self.device:
             high = high.to(self.device)
         if not isinstance(low, torch.Tensor):
             low = torch.as_tensor(low, device=self.device, dtype=loc.dtype)
-        else:
+        elif low.device != self.device:
             low = low.to(self.device)
         self.low = low
         self.high = high
--- a/torchrl/modules/distributions/continuous.py
+++ b/torchrl/modules/distributions/continuous.py
@@ -242,8 +242,14 @@ class TruncatedNormal(D.Independent):
         self.device = loc.device
         self.upscale = torch.as_tensor(upscale, device=self.device)
 
-        high = torch.as_tensor(high, device=self.device)
-        low = torch.as_tensor(low, device=self.device)
+        if not isinstance(high, torch.Tensor):
+            high = torch.as_tensor(high, device=self.device, dtype=loc.dtype)
+        else:
+            high = high.to(self.device)
+        if not isinstance(low, torch.Tensor):
+            low = torch.as_tensor(low, device=self.device, dtype=loc.dtype)
+        else:
+            low = low.to(self.device)
         self.low = low
         self.high = high
         self.update(loc, scale)
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -542,6 +542,30 @@ class TestIndependentNormal:
 
 
 class TestTruncatedNormal:
+    @pytest.mark.parametrize("device", get_default_devices())
+    @pytest.mark.parametrize("dtype", [torch.float16, torch.float32, torch.float64])
+    @pytest.mark.parametrize("tensor_bounds", [False, True])
+    def test_truncnormal_rsample_dtype(self, device, dtype, tensor_bounds):
+        torch.manual_seed(0)
+        loc = torch.zeros(3, device=device, dtype=dtype, requires_grad=True)
+        scale = torch.ones(3, device=device, dtype=dtype, requires_grad=True)
+        if tensor_bounds:
+            low = torch.full_like(loc, -1)
+            high = torch.full_like(loc, 1)
+        else:
+            low = -1.0
+            high = 1.0
+
+        dist = TruncatedNormal(loc, scale, low=low, high=high)
+        sample = dist.rsample((4,))
+
+        assert sample.dtype == dtype
+        assert sample.isfinite().all()
+        sample.sum().backward()
+        for grad in (loc.grad, scale.grad):
+            assert grad is not None and grad.dtype == dtype
+            assert grad.isfinite().all()
+
     @pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


aswanth-07:
CI triage: all six new `TruncatedNormal` dtype cases passed in the Python 3.14 bulk shard.

The red checks are outside this PR's scope:
- `test_replay_buffer_prefetch_pickle_roundtrip` failed its first random sample comparison and then passed on pytest's immediate rerun in the same job.
- The Gym compatibility matrix failed only for Gymnasium 0.28 at `TestGym::test_gym[True-True-1-HalfCheetah-v4]` with `Did not find pixels with norm > 1`; the Gym 0.13 and latest Gymnasium suites passed. This PR does not touch replay buffers, environments, rendering, or Gym integration.

I attempted to rerun only the failed workflow, but GitHub restricts that action to repository admins. No code change is warranted for either failure.

torchrl/modules/distributions/continuous.py
```diff
@@ -242,8 +242,14 @@ def __init__(
         self.device = loc.device
         self.upscale = torch.as_tensor(upscale, device=self.device)
 
-        high = torch.as_tensor(high, device=self.device)
-        low = torch.as_tensor(low, device=self.device)
+        if not isinstance(high, torch.Tensor):
+            high = torch.as_tensor(high, device=self.device, dtype=loc.dtype)
+        else:
+            high = high.to(self.device)
```
vmoens, line 248: Maybe guard this under a device check to make sure it doesn't break under compile?

torchrl/modules/distributions/continuous.py
```diff
@@ -242,8 +242,14 @@ def __init__(
...
+        if not isinstance(low, torch.Tensor):
+            low = torch.as_tensor(low, device=self.device, dtype=loc.dtype)
+        else:
+            low = low.to(self.device)
```
vmoens, line 252: Ditto

vmoens (commented):
Thanks 
One minor comment

torchrl/modules/distributions/continuous.py
aswanth-07: Good point — I've guarded the transfer with `high.device != self.device`, so same-device tensor bounds now bypass `.to()` under compile.

torchrl/modules/distributions/continuous.py
aswanth-07: Done for `low` as well; both tensor bounds now use the same guarded device-transfer path.

vmoens (approved):
LGTM thanks!

PR #4143, opened 2026-08-23
[BugFix] Fix cudagraph_policy instability and add action std test

### LINKED ISSUE #3002: [BUG] CUDAgraph policy changes the learning process
## Describe the bug

I have a single collector and I am training with ClipPPO a simple path finding policy. The agent must find a path in 3D around a simple maze while maximizing the coverage of the maze. It receives as input small patches centered around its current position and outputs the parameters to a Beta distribution (see example). From the distribution, we sample a value which is mapped to some action delta ($a_t \in R^3 $) to determine the next position. 

 To accelerate the experience sampling, I opted to apply CUDA graphs, but it seems to lead to unstable learning.

**A specific pathology I notice is that the standard deviation of the actions determined by the CUDAgraph'd policy tends to zero.**

## To Reproduce

I can't provide my full code, but the only difference is turning the cudagraph_policy=True in the collector. Maybe good to know that the policy module is `torch.compile`'d BEFORE it is passed onto the collector.

Policy (CNN extraction head + MLP, Beta distribution):
```python
class IndependentBeta(Independent):
    def __init__(
        self,
        alpha: torch.Tensor,
        beta: torch.Tensor,
        min: Union[float, torch.Tensor] = 0.0,
        max: Union[float, torch.Tensor] = 1.0,
        event_dims: int = 1,
    ):
        self.min = torch.as_tensor(min, device=alpha.device).broadcast_to(alpha.shape)
        self.max = torch.as_tensor(max, device=alpha.device).broadcast_to(alpha.shape)
        self.scale = self.max - self.min
        self.eps = torch.finfo(alpha.dtype).eps
        base_dist = Beta(alpha, beta)
        super().__init__(base_dist, event_dims)

    def sample(self, sample_shape: torch.Size = torch.Size()):
        return super().sample(sample_shape) * self.scale + self.min

    def rsample(self, sample_shape: torch.Size = torch.Size()):
        return super().rsample(sample_shape) * self.scale + self.min

    def log_prob(self, value: torch.Tensor):
        return super().log_prob(((value - self.min) / self.scale).clamp(self.eps, 1.0 - self.eps))

policy_module = ProbabilisticActor(
        module=TensorDictSequential(
            actor_cnn_module,  # Outputs TD with "alpha, beta"
        ),
        spec=action_spec,
        in_keys=["alpha", "beta"],
        out_keys=["action"],
        distribution_class=IndependentBeta,
        return_log_prob=True,
        default_interaction_type=InteractionType.RANDOM,
).to(device)

policy_module.compile(fullgraph=True, dynamic=False)
```

Collector:
```python
        create_env_fn=env_maker,  # Function to create environments
        policy=policy_module,
        # Total frames (steps) to collect in training
        total_frames=total_timesteps - collected_frames,
        frames_per_batch=config.frames_per_batch,
        # No initial random exploration phase needed if policy handles exploration
        init_random_frames=-1,
        split_trajs=False,  # Process rollouts as single batch
        device=device,  # Device for collector ops (usually same as models/env)
        # Device where data is stored (can be CPU if memory is tight)
        storing_device=device,
        max_frames_per_traj=config.max_episode_steps, 
        # num_threads=8
        # cudagraph_policy=True
    )
```

Training code (AMP+BF16):

```python
    for i, batch_data in enumerate(collector, start=collected_frames):
        current_frames = batch_data.numel()  # Number of steps collected in this batch
        collected_frames += current_frames

        for _ in range(config.update_epochs):
            batch_data = batch_data.reshape(-1)

            with (
                torch.no_grad(),
                torch.autocast(device.type, amp_dtype, enabled=config.amp),
            ):
                    adv_module(batch_data)

            for j in range(0, config.frames_per_batch, batch_size):
                minibatch = batch_data[j : j + batch_size]
                with torch.autocast(device.type, amp_dtype, enabled=config.amp):
                    loss_dict = loss_module(minibatch)

                    actor_loss = loss_dict["loss_objective"] + loss_dict["loss_entropy"]
                    critic_loss = loss_dict["loss_critic"]

                optimizer.zero_grad()
                scaler.scale(actor_loss).backward()
                scaler.scale(critic_loss).backward()
                scaler.unscale_(optimizer)
                grad_norm = torch.nn.utils.clip_grad_norm_(
                    loss_module.parameters(), config.max_grad_norm
                )
                scaler.step(optimizer)
                scaler.update()
```

## Expected behavior

The loss/training curves to be the same.

## Screenshots

Environment is a custom environment. Red is with `cudagraph_policy=False`, gray is `cudagraph_policy=True`

Image

Page 2:

Image

## System info

* torchrl==0.8.0 (pip, uv)
* Python 3.12.3
* Linux

## Reason and Possible fixes

No real idea why that might be.

## Checklist

coder-jayp:
## Description

This PR fixes a bug where the standard deviation of actions from a `cudagraph_policy` collapses to zero during training when using a stochastic policy compiled with `torch.compile`. 

When a compiled module is captured into a `CudaGraphModule` by the `SyncDataCollector`, Inductor's internal Philox RNG state (seed and offset) gets frozen as static buffers. Consequently, the random samples drawn (e.g. from `Beta` or `Normal` distributions) are identical on every step replay.

Fix injects `torch.compiler.cudagraph_mark_step_begin()` immediately before the compiled policy is called during rollout. This signals PyTorch to emit an offset-increment kernel, ensuring fresh random numbers are generated on every execution of the CUDA graph. Additionally, a defensive `.clone()` is added to the policy outputs to prevent aliasing with the static CUDA graph buffers inside the TensorDict.

## Motivation and Context

Fixes the learning instability and action standard deviation collapse observed when `cudagraph_policy=True` is used with a compiled stochastic policy.

close #3002

- [x] I have raised a bug report to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

- [x] Bug fix (non-breaking change which fixes a bug)

## Checklist

### DIFF
--- a/test/test_collectors.py
+++ b/test/test_collectors.py
@@ -287,7 +287,7 @@ class ParametricPolicy(Actor):
         )
 
 
-class RandomPolicy(torch.nn.Module):
+class MockRandomPolicy(torch.nn.Module):
     def __init__(self, action_dim):
         super().__init__()
         self.action_dim = action_dim
@@ -296,7 +296,6 @@ class RandomPolicy(torch.nn.Module):
         return torch.randn(
             *observation.shape[:-1], self.action_dim, device=observation.device
         )
-
 
 
 class DeterministicZeroPolicyNet(nn.Module):
@@ -2345,8 +2344,18 @@ if __name__ == "__main__":
 
     @pytest.mark.gpu
     @pytest.mark.skipif(not torch.cuda.is_available(), reason="requires cuda")
-    @pytest.mark.skipif(IS_WINDOWS, reason="torch.compile not fully supported on windows")
+    @pytest.mark.skipif(
+        IS_WINDOWS, reason="torch.compile not fully supported on windows"
+    )
     def test_cudagraph_policy_stochastic_diversity(self):
+        try:
+            from torch.compiler import cudagraph_mark_step_begin
+
+            _ = cudagraph_mark_step_begin
+        except ImportError:
+            pytest.skip(
+                "cudagraph_mark_step_begin not available in this PyTorch version."
+            )
 
         def create_env():
             return ContinuousActionVecMockEnv(device="cuda:0")
@@ -2355,7 +2364,7 @@ if __name__ == "__main__":
         n_actions = env.action_spec.shape[-1]
 
         policy = TensorDictModule(
-            RandomPolicy(n_actions), in_keys=["observation"], out_keys=["action"]
+            MockRandomPolicy(n_actions), in_keys=["observation"], out_keys=["action"]
         )
 
         collector = Collector(
@@ -2380,7 +2389,9 @@ if __name__ == "__main__":
         assert actions is not None, "Collector yielded no data"
         # If RNG is frozen by cudagraph, standard deviation will be exactly 0 across steps
         std = actions.std(dim=0).mean().item()
-        assert std > 0.1, f"Actions have abnormally low variance (std={std}). cudagraph RNG state might be frozen."
+        assert (
+            std > 0.1
+        ), f"Actions have abnormally low variance (std={std}). cudagraph RNG state might be frozen."
 
     @pytest.mark.parametrize("use_async", [False, True])
     @pytest.mark.parametrize(
--- a/test/test_collectors.py
+++ b/test/test_collectors.py
@@ -287,6 +287,18 @@ class ParametricPolicy(Actor):
         )
 
 
+class RandomPolicy(torch.nn.Module):
+    def __init__(self, action_dim):
+        super().__init__()
+        self.action_dim = action_dim
+
+    def forward(self, observation):
+        return torch.randn(
+            *observation.shape[:-1], self.action_dim, device=observation.device
+        )
+
+
+
 class DeterministicZeroPolicyNet(nn.Module):
     """A simple policy that always outputs action 0 (for discrete action spaces)."""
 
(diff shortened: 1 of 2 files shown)


coder-jayp:
@torchrlbot reviewer @vmoens

test/test_collectors.py
```diff
@@ -2331,6 +2331,46 @@ def env_fn(seed):
         assert_allclose_td(data1, data20)
         assert_allclose_td(data10, data20)
 
+    @pytest.mark.gpu
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="requires cuda")
+    def test_cudagraph_policy_stochastic_diversity(self):
+        from unittest.mock import patch
+
+        from tensordict.nn import TensorDictModule
+        from torchrl.envs.utils import ExplorationType, set_exploration_type
```
vmoens: No local imports

test/test_collectors.py
```diff
@@ -2331,6 +2331,46 @@ def env_fn(seed):
...
+
+        def create_env():
+            return ContinuousActionVecMockEnv()
+
+        env = create_env()
+        n_actions = env.action_spec.shape[-1]
+
+        policy = TensorDictModule(
+            torch.nn.LazyLinear(n_actions), in_keys=["observation"], out_keys=["action"]
+        ).to("cuda:0")
```
vmoens: Define device during construction

test/test_collectors.py
```diff
@@ -2331,6 +2331,46 @@ def env_fn(seed):
...
+            "torchrl.collectors._single.cudagraph_mark_step_begin"
+        ) as mock_mark_step:
+            with set_exploration_type(ExplorationType.RANDOM):
+                for _ in collector:
+                    break
+
+            # Ensure cudagraph_mark_step_begin was called to properly advance Inductor state
+            assert (
+                mock_mark_step.called
+            ), "cudagraph_mark_step_begin was not called for cudagraphed_policy"
```
vmoens: This doesn't test that the old behavior (the bug) existed and is solved, it just tests we called what we implemented which is circular thinking

vmoens (commented):
We're not really testing that we solved the bug

coder-jayp:
@vmoens Updated the test suite to address the feedback:

- Added a behavioral variance check to verify the CUDA graph Philox RNG state advances across rollout steps (`std > 0.1`)
- Moved the policy class to module scope, defined devices explicitly at construction, and added clean collector shutdown handling

Tested locally and ready for another look.

test/test_collectors.py
```diff
@@ -287,6 +287,18 @@ def __init__(self):
         )
 
 
+class RandomPolicy(torch.nn.Module):
```
vmoens: Maybe the one in the repo could be used?

vmoens (commented):
Almost here
CI is failing and one minor comment on RandomPolicy

test/test_collectors.py
coder-jayp: I tried it, but the built-in RandomPolicy isn't an nn.Module, causing CudaGraphModule to crash when capturing its RNG allocations. The custom one subclasses nn.Module to fix the tracing.

coder-jayp:
@vmoens Pushed a CI fix.

Unconditionally calling cudagraph_mark_step_begin() caused NotImplementedError on PT < 2.4 for native CUDA graphs. I updated _single.py to gracefully catch this for cudagraphed_policy, preserving backward compatibility.

vmoens (approved):

Issue #4142, opened 2026-08-22
Intermittent segfault in libcuda.so.1 during PPO training with custom EnvBase + ParallelEnv workers


aakhtarshenas:
Training crashes with a segmentation fault at unpredictable step counts (observed between 25 and ~1019 steps) during PPO training using a custom EnvBase subclass with num_workers parallel environments. The crash occurs inside the NVIDIA CUDA driver itself (libcuda.so.1), triggered by an ordinary tensor multiplication during the policy/value forward pass — not during my environment's own code.

Environment:
- torch: 2.5.1+cu121 (also reproduced identically on 2.7.1+cu128)
- torchrl: 0.13.3
- NVIDIA driver: 535.309.01
- GPU: NVIDIA RTX 6000 Ada Generation
- CUDA (driver-supported): 12.2
- OS: Ubuntu 24.04
- Python: 3.11
- num_workers: reproduces with both 1 and 4

What I've ruled out:
- A debug print statement calling repr() on a CUDA tensor every step (removed, crash persisted)
- OpenMP/MKL/OpenBLAS thread oversubscription (OMP_NUM_THREADS=1 etc., no effect)
- Heap corruption via glibc malloc (MALLOC_CHECK_=3, no abort triggered, same crash)
- CUDA/driver version mismatch (downgraded torch to match driver's max supported CUDA 12.2, crash persisted, even at a lower step count)
- Number of parallel workers (crashes identically with num_workers=1 and num_workers=4)
- Disabling the evaluation rollout entirely (crash still reproduces without eval ever running)
- A custom free_memory() utility calling torch.cuda.ipc_collect() — confirmed not even in the executed code path
- Classic GPU memory violations — ran under compute-sanitizer --tool memcheck, which reported 0 errors, yet the process still segfaulted

Backtrace (gdb) — crash occurs inside libcuda.so.1 with no exported symbol:

Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007ffea842c680 in ?? () from /lib/x86_64-linux-gnu/libcuda.so.1

Frames above it show the call chain through PyTorch's autograd dispatcher:

#13 c10::impl::wrap_kernel_functor_unboxed_<...mul_Tensor...>::call(...)
#14 at::_ops::mul_Tensor::call(at::Tensor const&, at::Tensor const&) ()
#15 torch::autograd::THPVariable_mul(...)

Python-level faulthandler traceback (representative example):

Current thread (most recent call first):
[...]

PR #4146, opened 2026-08-24
[BugFix] preserve caller CUDA stream in Collector

gtnv:
### TLDR

Addresses #4142.

Collector.iterator() yielded while inside its private CUDA stream context, so default cloned batches ran caller PPO code on TorchRL's stream. This gates the private stream path to `return_same_td=True`. cloned batches stay on the caller stream.

<details><summary>Testing + visuals</summary>
<p>

### rtx 6000

```
parent ae9719f17b1517af5c2523a36c4616f003ee55cf
fixed  68f945c20fe26189d07d1717737c3b8ce5e66035
PyTorch 2.7.1+cu128, Warp 1.16.0
seed 0, 1 and 4 ParallelEnv workers
1,100 PPO updates / 70,400 frames per run
```

stream-context

- Parent: 0/2,200 batches stayed on the caller stream.
- Fixed: 2,200/2,200 batches stayed on the caller stream.

ppo-control

- Loss and reward histories are exactly equal before and after the patch.
- All four runs completed with finite values.
- The substitute Warp workload did not reproduce the intermittent `libcuda.so.1` segfault on either revision. It does deterministically reproduce the TorchRL stream leak.

Ran regression which passed.

```bash
python -m pytest -q -p no:cacheprovider \
  test/test_collectors.py::TestCollectorDevices::test_no_synchronize
```

</p>
</details>

ai disclosure: i used an agent to write the description, though I modified it myself.

### DIFF
--- a/test/test_collectors.py
+++ b/test/test_collectors.py
@@ -3298,8 +3298,10 @@ class TestCollectorDevices:
                 no_cuda_sync=no_cuda_sync,
             )
             assert collector.env.device == torch.device(env_device)
+            caller_stream = torch.cuda.current_stream()
             i = 0
             for d in collector:
+                assert torch.cuda.current_stream() == caller_stream
                 for _d in d.unbind(0):
                     u = _d["observation"].unique()
                     assert u.numel() == 1, i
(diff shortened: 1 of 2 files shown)


vmoens (approved):
Looks correct to me! Thanks

PR #4151, opened 2026-08-24
[Feature] Add COMA multi-agent objective

Iliamsou:
## What

Adds a dedicated `COMALoss` objective for multi-agent reinforcement learning. The loss implements the COMA counterfactual baseline with a decentralised actor and a centralised Q-value network, including n-step Q-value targets and optional advantage normalisation.

## Contents

* `torchrl/objectives/multiagent/coma.py`: `COMALoss` implementation and helpers to construct the joint observation, joint action without the current agent, and masked joint action inputs used by the centralised critic.
* `torchrl/objectives/multiagent/__init__.py` and `torchrl/objectives/__init__.py`: expose `COMALoss`.
* `test/objectives/test_coma.py`: tests for the counterfactual baseline, Q-value targets, n-step returns, input construction helpers, advantage normalisation, and diagnostics.

## Tests

Ran:

`python -m pytest test/objectives/test_coma.py`

8 passed.

### DIFF
--- a/torchrl/objectives/multiagent/coma.py
+++ b/torchrl/objectives/multiagent/coma.py
@@ -24,8 +24,6 @@ class COMALoss(LossModule):
         action_value: NestedKey = ("agents", "action_value")
         chosen_action_value: NestedKey = ("agents", "chosen_action_value")
         logits: NestedKey = ("agents", "logits")
-        sample_log_prob: NestedKey = ("agents", "sample_log_prob")
-        advantage_old: NestedKey = ("agents", "advantage_old")
         reward: NestedKey = ("agents", "reward")
         done: NestedKey = ("agents", "done")
         terminated: NestedKey = ("agents", "terminated")
@@ -60,13 +58,10 @@ class COMALoss(LossModule):
         entropy_coef: float = 0.0,
         n_step: int = 1,
         normalize_advantage: bool = False,
-        clip_epsilon: float | None = None,
     ) -> None:
         super().__init__()
         if n_step < 1:
             raise ValueError(f"n_step must be >= 1, got {n_step}.")
-        if clip_epsilon is not None and not 0.0 < clip_epsilon < 1.0:
-            raise ValueError(f"clip_epsilon must be in (0, 1), got {clip_epsilon}.")
         self.convert_to_functional(actor_network, "actor_network")
         self.convert_to_functional(qvalue_network, "qvalue_network", create_target_params=True)
         self.gamma = gamma
@@ -74,7 +69,6 @@ class COMALoss(LossModule):
         self.entropy_coef = entropy_coef
         self.n_step = n_step
         self.normalize_advantage = normalize_advantage
-        self.clip_epsilon = clip_epsilon
 
     def forward(self, tensordict: TensorDictBase) -> TensorDict:
         td_copy = tensordict.clone(False)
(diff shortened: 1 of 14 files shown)


vmoens (changes requested):
Thanks for adding a dedicated COMA objective. I found several issues that need to be addressed before this can merge:

1. **[P1] Shift the actual time dimension.** In `compute_value_target`, `next_value[:, :-1] = value_target[:, 1:]` only shifts time for exactly `[B, T, ...]` inputs. A standard collector over an unbatched environment returns `[T, n_agents, 1]`, in which case this shifts the agent dimension. Multiple environment batch dimensions also shift the wrong axis. Please resolve the named/final TensorDict time dimension and add a `[T]` regression case.

2. **[P1] Bootstrap rollout boundaries and truncations correctly.** Zero-filling shifted values makes the last `n_step` transitions of every non-terminal fixed-length rollout use incomplete returns. Masking with `done` also suppresses a valid bootstrap at truncation; TorchRL return semantics distinguish this with `terminated`. The target critic needs to evaluate the transition next observation at boundaries rather than treating the end of the sampled batch as terminal.

3. **[P1] Honor collector validity masks.** Direct `.mean()` and `F.mse_loss(...)` reductions bypass `LossModule._reduce_loss`, so `("collector", "mask")` and `shifted_valid` do not exclude padded transitions. COMA commonly consumes episodic sequences, so this can train both actor and critic on padding. Please keep losses elementwise and reduce through `_reduce_loss(..., tensordict=tensordict)`.

4. **[P2] Preserve flat `NestedKey` values.** `("next",) + tuple(self.tensor_keys.reward)` works for tuple keys but turns `reward="reward"` into `("next", "r", "e", "w", "a", "r", "d")`. Please use the repository nested-key normalization pattern for both reward and termination keys.

The test volume is proportionate, but the current target tests only use synthetic `[1, T]` batches, which hides the time-axis failure. Please cover `[T]`, a truncated boundary, automatic loss masking, and both flat and nested `set_keys()` values.

The public API integration also needs the repository-required documentation: a complete Sphinx-style class docstring with runnable example and paper reference, an entry in `docs/source/reference/objectives_multiagent.rst`, and the applicable tutorial/SOTA recipe. Since this is a new test file, please also add the executable `if __name__ == "__main__": pytest.main(...)` block.

torchrl/objectives/multiagent/coma.py
```diff
@@ -0,0 +1,468 @@
...
+        ).to(chosen_action_value.dtype)
+        not_terminated = 1.0 - terminated
+
+        value_target = chosen_action_value
+        valid = torch.ones_like(terminated, dtype=torch.bool)
+        for _ in range(self.n_step):
+            next_value = _shift_time(value_target, time_dim, fill_value=0.0)
+            next_valid = _shift_time(valid, time_dim, fill_value=False)
+            value_target = reward + self.gamma * not_terminated * next_value
+            valid = terminated.to(torch.bool) | next_valid
```
gtnv: done=True, terminated=False still shifts a value across the reset. changing the next episode's first action moved the previous target from 0.5 to 2.0. can we invalidate this bootstrap at every done boundary?

torchrl/objectives/multiagent/coma.py
```diff
@@ -0,0 +1,468 @@
...
+            self.qvalue_network(td_copy)
+
+        chosen_action_value = self._chosen_action_value(td_copy)
+
+        advantage = self._counterfactual_advantage(td_copy, chosen_action_value).detach()
+
+        if self.normalize_advantage:
+            # MAPPO-style per-batch standardisation: same counterfactual
+            # advantage, rescaled so the actor step size is batch-invariant.
+            advantage = (advantage - advantage.mean()) / advantage.std().clamp_min(1e-6)
```
gtnv: one valid actor value returns nan here. masked padding also changes the scale. can we use the effective loss mask with correction=0?

torchrl/objectives/multiagent/coma.py
Iliamsou: Fixed, thanks !

torchrl/objectives/multiagent/coma.py
Iliamsou: Fixed, thanks!

PR #4154, opened 2026-08-25
[Performance] Specialize GRU scan backward

vmoens:
## Summary
- Hoist the GRU input projection out of the recurrent scan.
- Use a GRU-specific reverse scan that carries only the hidden-state gradient.
- Reduce input, parameter, and bias gradients over flattened batch-time dimensions outside the scan.
- Preserve the existing recompute path and public API.
- Document the one-kwarg backend selection, first-order limitations, and the hardware benchmark workflow.

## Testing
- Full RNN test module: 222 passed, 48 CUDA/Triton-only tests skipped.
- Added pad/scan output and gradient parity coverage for one and two layers, with and without bias, including initial and reset recurrent-state gradients.
- Extended the compiled scan regression to exercise backward and parameter gradients.
- Added BF16 autocast backward coverage and explicit double-backward rejection coverage.
- RST section-underline and whitespace validation passed for the recurrent-module documentation.

## Benchmark
Local CPU eager forward and backward, B=64, T=32, input=32, hidden=128, one layer:
- Before: 100.7 ms median
- After: 10.2 ms median

CUDA performance remains covered by `benchmarks/bench_rnn_backward.py`; the recurrent-module documentation now includes a runnable command for comparing cuDNN, scan, and Triton on the target hardware.

### DIFF
--- a/torchrl/modules/tensordict_module/rnn.py
+++ b/torchrl/modules/tensordict_module/rnn.py
@@ -153,6 +153,158 @@ def _gru_cell_from_gate_params(
     return newgate + inputgate * (hx - newgate)
 
 
+class _GRUScanFunction(torch.autograd.Function):
+    """GRU sequence with a hidden-gradient-only reverse scan.
+
+    The input projection and all shared-parameter gradient reductions run over
+    the flattened batch-time dimensions. Only the recurrent hidden-state
+    derivative remains inside the reverse scan.
+    """
+
+    @staticmethod
+    def forward(ctx, x, initial, reset_hidden, is_init, w_ih, w_hh, b_ih, b_hh):
+        time, batch, _ = x.shape
+        hidden_size = initial.shape[-1]
+        gate_chunks = (hidden_size, hidden_size, hidden_size)
+        gates_x = F.linear(x.flatten(0, 1), w_ih, b_ih).view(
+            time, batch, 3 * hidden_size
+        )
+
+        def step(carry, inputs):
+            gates_x_t, reset_hidden_t, init_t = inputs
+            h_previous = torch.where(init_t.unsqueeze(-1), reset_hidden_t, carry)
+            gates_h = F.linear(h_previous, w_hh, b_hh)
+            i_r, i_z, i_n = gates_x_t.split(gate_chunks, -1)
+            h_r, h_z, h_n = gates_h.split(gate_chunks, -1)
+            resetgate = (i_r + h_r).sigmoid()
+            updategate = (i_z + h_z).sigmoid()
+            newgate = (i_n + resetgate * h_n).tanh()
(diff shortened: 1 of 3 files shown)


torchrl/modules/tensordict_module/rnn.py
```diff
@@ -153,6 +153,158 @@ def _gru_cell_from_gate_params(
...
+
+        d_gates_x_flat = d_gates_x.reshape(time * batch, 3 * hidden_size)
+        d_gates_h_flat = d_gates_h.reshape(time * batch, 3 * hidden_size)
+        x_flat = x.reshape(time * batch, input_size)
+        previous_hidden_flat = previous_hidden.reshape(time * batch, hidden_size)
+        d_x = F.linear(d_gates_x_flat, w_ih.t()).view_as(x)
+        d_w_ih = d_gates_x_flat.t() @ x_flat
+        d_w_hh = d_gates_h_flat.t() @ previous_hidden_flat
+        d_b_ih = d_gates_x_flat.sum(0)
+        d_b_hh = d_gates_h_flat.sum(0)
```
gtnv: bf16 input and hidden state with fp32 parameters finish the autocast forward. backward then reaches this bf16/fp32 linear and fails. can we preserve the autocast dtype here?

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -153,6 +153,158 @@ def _gru_cell_from_gate_params(
...
+            initial,
+            reset_hidden,
+            is_init,
+            w_ih,
+            w_hh,
+            hidden,
+            resetgate,
+            updategate,
+            newgate,
+            h_n,
```
gtnv, line 213: first-order gradcheck passes. gradgradcheck gets the wrong jacobian because these saved gates are constants. can we recompute the differentiable gate state or use once_differentiable?

torchrl/modules/tensordict_module/rnn.py
```diff
@@ -153,6 +153,158 @@ def _gru_cell_from_gate_params(
...
+class _GRUScanFunction(torch.autograd.Function):
+    """GRU sequence with a hidden-gradient-only reverse scan.
+
+    The input projection and all shared-parameter gradient reductions run over
+    the flattened batch-time dimensions. Only the recurrent hidden-state
+    derivative remains inside the reverse scan.
+    """
+
+    @staticmethod
+    def forward(ctx, x, initial, reset_hidden, is_init, w_ih, w_hh, b_ih, b_hh):
```
gtnv, line 172: jacrev fails before this function runs. torch.func requires setup_context for this form. can we support that or reject these transforms explicitly?

vmoens:
thanks @gtnv 
Addressed your comments

PR #4158, opened 2026-08-26
[Feature] Add public DreamerV3 block-GRU sequence module

vmoens:
Stacked on #4154. Review only the commits above codex/optimize-gru-scan-backward.

## Summary

- expose public DreamerV3BlockGRUCell and batch-major DreamerV3BlockGRU modules
- keep the ordinary-autograd reference loop as the default and provide an opt-in optimized scan backend
- share the stateless block-GRU equations with RSSMPriorV3 while preserving its module hierarchy and checkpoint keys
- support resets, multi-layer block dynamics, FP32 parameters, and BF16 recurrent compute

## Validation

- public step/reference/scan forward parity across block, batch, horizon, dtype, reset, and device grids
- nonuniform-cotangent input, hidden-state, and parameter gradient parity
- fullgraph torch.compile next-step prediction loss and backward coverage
- existing RSSM reference and compile coverage
- focused stacked-GRU scan-backward tests
- formatting, lint, syntax, exports, and benchmark CLI checks

## Benchmarks

Extends the recurrent benchmark with block-GRU reference/scan grids, forward and forward-plus-backward timings, synchronized warmups, and 95% confidence intervals.

## Dependency

Depends on #4154.


torchrl/modules/models/model_based.py
```diff
@@ -174,25 +253,697 @@ def forward(
...
+            hidden_weight,
+            hidden_norm_weight,
+            *tensors,
+            outputs,
+            *saved,
+        )
+        return outputs, final_hidden
+
+    @staticmethod
+    def backward(ctx, grad_outputs, grad_final_hidden):
```
gtnv, line 447: the second-gradient probe differed from the reference by up to 65%. this is silent rn. can we preserve the graph or use once_differentiable?

PR #4159, opened 2026-08-26
[Performance] Optimize DreamerV3 block-GRU scan recurrence

vmoens:
Stacked on #4158. Review only the commits above codex/public-dreamerv3-block-gru.

## Summary

- hoist the projected-input slice of the first block-linear contraction across flattened batch and time
- cast recurrent operands and prepare split weight layouts once per sequence
- keep only carry and projected-hidden work inside the recurrence
- retain FP32 parameter contractions outside the reverse scan
- add benchmark selection for default, reduce-overhead, and max-autotune compile modes

The public API, default reference backend, parameter hierarchy, checkpoint keys, and RSSM behavior are unchanged.

## Testing

- full Dreamer component suite: 227 passed
- focused stacked RNN scan-backward coverage: 4 passed
- FP32 and BF16 forward/gradient parity, resets, one/eight blocks, and multilayer dynamics
- fullgraph compile and backward coverage
- full repository pre-commit
- benchmark CLI and syntax checks

Depends on #4158, which depends on #4154.

### DIFF
--- a/torchrl/modules/models/model_based.py
+++ b/torchrl/modules/models/model_based.py
@@ -153,6 +153,21 @@ def _dreamer_v3_block_linear(
     return value.reshape(*batch_shape, num_blocks * block_out) + bias.to(value.dtype)
 
 
+def _dreamer_v3_block_linear_compute(
+    value: torch.Tensor,
+    weight: torch.Tensor,
+    bias: torch.Tensor | None,
+) -> torch.Tensor:
+    """Apply block-linear operands that already share a compute dtype."""
+    num_blocks, _, block_out = weight.shape
+    batch_shape = value.shape[:-1]
+    value = value.reshape(-1, num_blocks, value.shape[-1] // num_blocks)
+    value = value.transpose(0, 1)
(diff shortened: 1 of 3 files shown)


torchrl/modules/models/model_based.py
```diff
@@ -350,12 +365,43 @@ def forward(
...
+        compute_dtype = projected_input.dtype
+
+        hidden_weight_compute = hidden_weight.to(compute_dtype)
+        hidden_bias_compute = hidden_bias.to(compute_dtype)
+        dynamic_compute = [
+            (weight.to(compute_dtype), bias.to(compute_dtype), norm_weight)
+            for weight, bias, norm_weight in dynamic
+        ]
+        gate_weight_compute = gate_weight.to(compute_dtype)
+        gate_bias_compute = gate_bias.to(compute_dtype)
```
gtnv, line 379: bf16 input with fp32 hidden works in the parent backend. this path fails after choosing projected_input's dtype alone. can we use both dtypes or reject the mix?

PR #4160, opened 2026-08-26
[Performance] Add DreamerV3 block-GRU Triton backend

vmoens:
## Summary

- add an explicit CUDA-only Triton backend for DreamerV3BlockGRU
- fuse the time recurrence into one forward and one reverse-time kernel
- preserve FP32 parameters and accumulation while supporting FP32 and BF16 inputs (mixed input/hidden dtypes are promoted like the reference backend)
- cover supported activations, reset behavior, multiple blocks/layers, non-square projections, mixed dtypes, gradients, errors, and compilation
- document the reference/scan/triton backend trade-offs (portability, hardware, precision, double-backward behavior) in the DreamerV3 reference docs, and keep `benchmarks/bench_rnn_backward.py` able to compare the backends on user shapes and hardware

Depends on #4159.

## GB200 performance

NVIDIA GB200, BF16, B=16, H=P=512, 8 blocks, 1 layer, 10 warmups and 30 synchronized samples. Values are mean +/- 95% confidence half-width.

| T | backend | forward (ms) | forward + backward (ms) | peak memory (GB) |
|---:|:---|---:|---:|---:|
| 64 | scan | 65.426 +/- 0.106 | 115.099 +/- 0.915 | 0.160 |
| 64 | triton | 4.223 +/- 0.014 | 7.443 +/- 0.010 | 0.149 |
| 512 | scan | 540.584 +/- 0.578 | 855.082 +/- 5.762 | 0.782 |
| 512 | triton | 28.273 +/- 0.040 | 62.608 +/- 0.023 | 0.690 |

The Triton upper confidence bound is below the scan lower confidence bound at both representative horizons. On the tested PyTorch 2.13 nightly image, the outer torch.compile mode sweep remained in graph compilation for more than two minutes; the direct Triton and scan paths above both completed normally.

## Validation

- full pre-commit suite for the backend implementation
- focused Dreamer component and recurrent tests
- CUDA FP32/BF16 forward and gradient parity, including resets, non-square projections, mixed dtypes, and multilayer cases
- compiled next-step prediction loss parity on the supported CI toolchain
- GB200 performance gate via benchmarks/bench_rnn_backward.py


torchrl/modules/models/_dreamer_v3_block_gru_triton.py
```diff
@@ -0,0 +1,1310 @@
...
+        dynamic_inv_rms = torch.empty(
+            num_layers, batch, time, device=projected_input.device, dtype=torch.float32
+        )
+        gate_state = torch.empty(
+            batch,
+            time,
+            4 * hidden_size,
+            device=projected_input.device,
+            dtype=compute_dtype,
+        )
```
gtnv: save_state=False compiles out these buffer reads. inference still allocates gate_state with 4 * B * T * H elements. can it use zero-sized dummies for the backward-only buffers?

torchrl/modules/models/_dreamer_v3_block_gru_triton.py
```diff
@@ -0,0 +1,1310 @@
...
+            batch, projected_size, device=projected_input.device, dtype=compute_dtype
+        )
+        dynamic_normalized = torch.empty(
+            num_layers,
+            batch,
+            time,
+            hidden_size,
+            device=projected_input.device,
+            dtype=compute_dtype,
+        )
```
gtnv, line 999: this scratch is only read for the current timestep when state is not saved. it still keeps every T slice. can its inference shape be [num_layers, batch, hidden_size]?

PR #4161, opened 2026-08-26
[Doc] Fold corrected TRL interoperability end-to-end section into trl_interop tutorial

coder-jayp:
This PR extends the existing `trl_interop.py` tutorial with a corrected 
end-to-end section (Part 4) that demonstrates the complete TorchRL ↔ TRL 
user journey.

Specific changes:
- Adds Part 4 to `trl_interop.py` covering the full preference-pair pipeline:
  collect (prompt + chosen + rejected) into a ReplayBuffer, bridge to a trl
  trainer via TorchRLBufferDataset, and score rollouts with HFRewardModelWrapper
  using the canonical ("tokens", "full") / ("masks", "all_attention_mask") keys.
- Shows LLMCollector as the production data-collection path.
- Removes the standalone `coding_grpo_trl.py` file.
- Removes the `coding_grpo_trl` entry from `index.rst` and `llms.rst`.

Motivation and Context:

Addresses review feedback on the previous version. The new section:
- Preserves full conversational context (prompt is no longer discarded).
- Exercises the actual TorchRL-to-TRL boundary as described in the RFC.
- Runs in CI using toy models only (no network access or GPU required).

## Types of changes

## Checklist


coder-jayp:
@torchrlbot reviewer @vmoens

tutorials/sphinx-tutorials/coding_grpo_trl.py
```diff
@@ -0,0 +1,325 @@
...
+# would happen inside the collector's data-processing transforms.
+
+N_SAMPLES = 128  # small enough for CI, large enough to show a real trend
+BATCH_SIZE = 8
+MODEL_NAME = "distilbert-base-uncased"
+
+if _has_trl:
+    # Enable TensorDict to store Python lists (strings) transparently.
+    # Scoped here so it only applies when trl is available.
+    set_list_to_stack(True).set()
```
gtnv: set() changes TensorDict's process-global mode and LIST_TO_STACK. the change lasts for the rest of the process. can we scope it to buffer construction?

tutorials/sphinx-tutorials/coding_grpo_trl.py
```diff
@@ -0,0 +1,325 @@
...
+import matplotlib
+
+matplotlib.use("Agg")  # non-interactive backend; Sphinx-Gallery captures the figure
+import matplotlib.pyplot as plt
+import torch
+from tensordict import set_list_to_stack, TensorDict
+from torchrl.data import ListStorage, ReplayBuffer
+from torchrl.modules.llm import HFRewardModelWrapper, TorchRLBufferDataset
+
+warnings.filterwarnings("ignore")
```
gtnv: filterwarnings("ignore") affects the full gallery process. every later warning is hidden. can we use warnings.catch_warnings() around the known warning?

tutorials/sphinx-tutorials/coding_grpo_trl.py
coder-jayp: Updated. Scoped using with set_list_to_stack(True): around the ReplayBuffer population loop so it doesn't leak globally

tutorials/sphinx-tutorials/coding_grpo_trl.py
coder-jayp: Updated. Removed the global filter and wrapped trainer.train() in with warnings.catch_warnings().

vmoens:
Can you rebase?

coder-jayp:
Done @vmoens

tutorials/sphinx-tutorials/coding_grpo_trl.py
```diff
@@ -0,0 +1,325 @@
...
+#
+# In production the buffer would be filled online by a
+# :class:`~torchrl.collectors.llm.LLMCollector`; the chat-template step
+# would happen inside the collector's data-processing transforms.
+
+N_SAMPLES = 128  # small enough for CI, large enough to show a real trend
+BATCH_SIZE = 8
+MODEL_NAME = "distilbert-base-uncased"
+
+if _has_trl:
```
vmoens: that's not amazing for a tutorial, how do we avoid that kind of guard and block?

tutorials/sphinx-tutorials/coding_grpo_trl.py
```diff
@@ -0,0 +1,325 @@
...
+                )
+            )
+
+    print(f"ReplayBuffer populated: {len(rb)} preference pairs.")
+    print(f"  Sample chosen  : {rb[0]['chosen'][:120]}...")
+    print(f"  Sample rejected: {rb[0]['rejected'][:120]}...")
+
+# %%
+# 2. Bridging TorchRL → TRL with TorchRLBufferDataset
+# ----------------------------------------------------
```
vmoens:
```suggestion
# ---------------------------------------------------
```

coder-jayp:
cc @vmoens

vmoens (commented):
The real-TRL example is useful and fits the interoperability RFC, but the current tutorial does not yet demonstrate the user journey that motivates TorchRL.

It round-trips a static HF dataset through TorchRL, discards the prompt when constructing reward-model examples, skips the preference-pair creation step attributed to LLMCollector, and finally scores random tokens outside the canonical History/Tokens representation.

The real path is also skipped by both docs and tutorial CI. Could we either fold a corrected end-to-end section into the existing interop tutorial, or position this within a coherent LLM tutorial track, while preserving conversational context and exercising the actual TorchRL-to-TRL boundary?

coder-jayp:
@vmoens Removed the standalone file and folded a corrected end-to-end section (Part 4) into the existing trl_interop.py. Full prompt context included, LLMCollector shown as the production path, and HFRewardModelWrapper uses the canonical ("tokens", "full") / ("masks", "all_attention_mask") keys. Runs in CI with toy models only.

coder-jayp:
cc @vmoens @theap06

vmoens:
Thanks for iterating on this. I pushed a small cleanup to make the scope and contracts accurate: `LLMCollector` produces rollouts rather than preference labels, `RewardTrainer` and `GRPOTrainer` have different dataset requirements, and the toy reward model was not actually trained by the preceding example. The final section now presents the preference buffer as an interoperability boundary rather than claiming a complete end-to-end pipeline.

At a higher level, I still think this raises an organizational question for TorchRL. This is useful as a focused recipe, but our LLM story is still forming and I would rather see it become part of a broader sequence covering collection/conversation state, preference construction, training, and scoring. In particular, a follow-up should explore how this composes with History and the existing conversational abstractions instead of establishing standalone string fields as the apparent workflow. I don’t think that necessarily needs to block this PR once CI is green, but I would like us to agree on that broader direction so this does not remain an isolated pattern.

cc @theap06: we need and RFC on our LLM APIs to clean it up, make it more suitable for large scale training (incl. distributed APIs, titan/megatron integration, etc). One thing I'm particularly worried/thinking about is how do we guarantee that we have one way and only one way of doing things such that it's not confusing for users.

I think the tutorials should be the last step of our LLM post training integration.

coder-jayp:
@vmoens I completely agree. This tutorial serves as a focused recipe for bridging to TRL right now, but it makes total sense that it should eventually integrate with the History abstractions. 

Once the larger LLM API RFC is finalized (especially around distributed APIs and unified conversational state), we can revisit and update these tutorials so they demonstrate the single, canonical workflow.

PR #4164, opened 2026-08-26
[Quality] Strengthen LLM policy loss tests

quinnarnold:
## Summary

- distinguish asymmetric GRPO clipping from symmetric clipping with exact ratios
- exercise KL masking with tokens on both sides of the threshold
- verify the CISPO clipped-importance objective numerically

## Motivation

Related to #4144. The prior shape and finiteness checks did not distinguish these public behaviors. This PR intentionally does not close the broader audit issue.

## Testing

- 69 focused tests passed, with 3 optional-dependency skips
- targeted mutation checks rejected incorrect behavior
- git diff --check passed

cc @vmoens

### DIFF
--- a/test/llm/test_llm_objectives.py
+++ b/test/llm/test_llm_objectives.py
@@ -854,9 +854,9 @@ class TestLosses:
             sample_log_prob=[0.0],
             advantage=[1.0],
         )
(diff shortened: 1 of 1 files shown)


vmoens (approved):
LGTM

PR #4165, opened 2026-08-26
[Quality] Test RND update fraction behavior

quinnarnold:
## Summary

- use fixed predictor and target outputs to produce distinguishable per-sample losses
- control the update mask and verify update_fraction selects the expected samples

## Motivation

Related to #4144. The previous scalar-output assertion could pass if every update fraction behaved as 1.0. This PR intentionally does not close the broader audit issue.

## Testing

- 10 focused RND tests passed
- git diff --check passed

cc @vmoens


vmoens (approved):
LGTM

PR #4166, opened 2026-08-26
[Quality] Test replay buffer data behavior

quinnarnold:
## Summary

- verify prefill_replay_buffer uses the requested chunk sizes
- verify min, max, mean, and median trajectory-priority reductions with distinguishable priorities
- remove a redundant reduction parameterization that did not assert reduction behavior

## Motivation

Related to #4144. These tests now validate public chunking and prioritized replay behavior rather than only final shapes and lengths. This PR intentionally does not close the broader audit issue.

## Testing

- 13 focused replay and data tests passed
- git diff --check passed

cc @vmoens


vmoens (approved):
LGTM

PR #4167, opened 2026-08-26
[Quality] Test UCB1-Tuned score calculation

quinnarnold:
## Summary

- add hand-computed low-variance and high-variance score cases
- distinguish both sides of the documented min(0.25, V_i) clamp

## Motivation

Related to #4144. The prior shape and finiteness checks could not detect an incorrect variance formula or clamp. This PR intentionally does not close the broader audit issue.

## Testing

- 10 focused UCB1-Tuned tests passed
- git diff --check passed

cc @vmoens


vmoens (approved):
LGTM

PR #4169, opened 2026-08-26
[BugFix] Implement shared MLP output bias

quinnarnold:
## Summary

- implement the documented single_bias_last_layer option as one scalar bias shared across output features\n- honor bias_last_layer, lazy layers, device and dtype, and shaped outputs\n- reject single shared bias with NoisyLinear because its per-output stochastic bias has incompatible semantics\n- reset the shared scalar through the standard recursive reset path\n- activate the existing parameterized coverage and add focused shared-bias assertions\n\n## Motivation\n\nRelated to #4144. The maintainer confirmed that the public option raising NotImplementedError was a bug and should follow its documented behavior. This PR intentionally does not close the broader audit issue.\n\n## Testing\n\n- all 1,299 focused TestMLP cases passed\n- git diff --check passed\n\ncc @vmoens


vmoens (approved):
LGTM

PR #4180, opened 2026-08-28
[Feature] Support fragmented trajectory slices

vmoens:
## Stack

This is the foundational PR for #4130. The geometric trajectory-window sampler is stacked on this branch and consumes the shared fragmented-trajectory index introduced here.

## Summary

- add an internal fragmented-trajectory index that maps logical `(trajectory, step)` adjacency independently of physical replay storage order
- add `SliceSampler(fragmented=True, step_key=...)` for fixed-length slices from interleaved or otherwise non-contiguous trajectories
- split missing logical steps into separate sampleable runs and refresh trajectory links incrementally after tracked writes
- preserve the existing flattened SliceSampler output, truncation, padding, and `is_init` metadata contracts
- add focused nested-key, wraparound, missing-step, and cached-sampling coverage
- continuously compare ordinary contiguous sampling with fragmented sampling across hot-cache, write-invalidated, and cold-start states

## Scope

Fragmented mode currently supports single-dimensional TensorDict-backed storages and SliceSampler with replacement. `span` and `compile` are left for follow-up work. End flags alone cannot reconstruct fragmented ordering; trajectory and step identifiers are required.

## Benchmark comparison

The benchmark constructs the same eight logical trajectories for both layouts. The baseline stores them contiguously and uses ordinary `SliceSampler`; the fragmented case applies a deterministic permutation to physical storage and uses `fragmented=True`. Both use the same tensor storage, metadata, batch size, slice length, and RNG seed. CUDA is synchronized at setup boundaries and after each timed sample.

Three cache states are measured:

- `hot-cache`: repeated sampling from an unchanged, already-indexed buffer
- `write-invalidated`: one tracked overwrite runs before every timed sample, outside the timed region
- `cold-start`: a newly populated replay buffer is prepared before every round and its first sample is timed

Representative local CPU medians:

| State | Contiguous, 1k | Fragmented, 1k | Contiguous, 100k | Fragmented, 100k |
| hot cache | 87.6 us | 71.3 us | 94.8 us | 73.5 us |
| write invalidated | 141.3 us | 222.1 us | 304.0 us | 16.5 ms |
| cold start | 156.5 us | 452.4 us | 309.7 us | 41.6 ms |

The fragmented selector is faster once its packed trajectory index is hot, but rebuilding or repacking that index dominates when sampling immediately after writes. The benchmark uses fixed round counts for the slower states; all 12 local CPU cases complete in approximately 1.9 seconds.

## Validation

- `pytest test/rb/test_samplers.py -k fragmented -q` (3 passed)
- focused SliceSampler compatibility selection (21 passed)
- `pytest benchmarks/test_replaybuffer_benchmark.py::test_slice_sampler_sample --benchmark-only -q` (12 passed in approximately 1.9 seconds)
- exact repository ufmt check and `git diff --check`

The broader SliceSampler selection reached 87 passes locally; its prioritized cases cannot run because this environment lacks the native `SumSegmentTreeFp32` extension.

### DIFF
--- a/benchmarks/test_replaybuffer_benchmark.py
+++ b/benchmarks/test_replaybuffer_benchmark.py
@@ -164,35 +164,110 @@ def test_slice_sampler_boundary_query_benchmark(
     benchmark(sampler._get_stop_and_length, rb.storage)
 
 
+@pytest.mark.parametrize(
+    "cache_state",
+    [
+        pytest.param("hot", id="hot-cache"),
+        pytest.param("write_invalidated", id="write-invalidated"),
+        pytest.param("cold", id="cold-start"),
+    ],
+)
+@pytest.mark.parametrize(
+    "layout",
+    [
+        pytest.param("contiguous", id="contiguous"),
+        pytest.param("shuffled", id="fragmented-shuffled"),
+    ],
+)
 @pytest.mark.parametrize("size", [1_000, 100_000])
-def test_fragmented_slice_sampler_cached_sample(benchmark, size):
+def test_slice_sampler_sample(benchmark, size, layout, cache_state):
     device = _replay_boundary_device()
     num_trajectories = 8
-    storage_order = torch.arange(size, device=device)
-    rb = TensorDictReplayBuffer(
-        storage=LazyTensorStorage(size, device=device),
-        sampler=SliceSampler(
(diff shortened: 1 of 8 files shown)


vmoens:
cc @theap06

PR #4189, opened 2026-08-30
[BugFix] Fix Trainer optimizer checkpointing

### LINKED ISSUE #4188: [BUG] Trainer checkpoints do not preserve optimizer state
## Describe the bug

Trainer checkpoints do not consistently preserve optimizer state.

When an optimizer is passed directly to `Trainer` without a default stepper, the trainer may install an optimizer hook. In this case, the trainer does NOT separately serialize its optimizer, but the optimizer hook's state dict is empty.  Checkpoints are then restored with their model parameters etc. but without the optimizer state. 

It also occurs when using a default optimization stepper, because the default stepper does not own the optimizer, and no hook is added to `_modules`.

The same issue could occur when an optimizer is provided through an explicitly registered hook while `trainer.optimizer` is None (although maybe this is an antipattern).

## To Reproduce

```python
from __future__ import annotations

import torch

from torchrl.trainers import DefaultOptimizationStepper, Trainer


class DummyCollector:
    def state_dict(self):
        return {}

    def load_state_dict(self, state_dict):
        pass


def make_trainer(model, optimizer):
    return Trainer(
        collector=DummyCollector(),
        total_frames=None,
        frame_skip=None,
        optim_steps_per_batch=None,
        loss_module=model,
        optimizer=optimizer,
        optimization_stepper=DefaultOptimizationStepper(),
        progress_bar=False,
    )


# Create a trainer and initialize Adam's internal state.
model = torch.nn.Linear(1, 1)
optimizer = torch.optim.Adam(model.parameters())
trainer = make_trainer(model, optimizer)

optimizer.zero_grad()
model(torch.ones(1, 1)).sum().backward()
optimizer.step()

checkpoint = trainer.state_dict()

# Restore into a fresh trainer.
model_restored = torch.nn.Linear(1, 1)
optimizer_restored = torch.optim.Adam(model_restored.parameters())
trainer_restored = make_trainer(model_restored, optimizer_restored)
trainer_restored.load_state_dict(checkpoint)

print("Optimizer in checkpoint:", "optimizer" in checkpoint)
print("Original optimizer state entries:", len(optimizer.state_dict()["state"]))
print(
    "Restored optimizer state entries:",
    len(optimizer_restored.state_dict()["state"]),
)
```

```bash
  File ... 
```

## Expected behavior

Trainer checkpoints should restore the optimizer state along with model params etc regardless of whether the optimizer is:
- owned by the trainer
- owned by an optimizer hook
- part of a checkpointable trainer component

## System info

```python
import torchrl, numpy, sys
0.14.0+gc80e397b0 2.2.6 3.10.18 | packaged by conda-forge | (main, Jun  4 2025, 14:46:00) [Clang 18.1.8 ] darwin
```

## Reason and Possible fixes

See above

## Checklist

bsprenger:
## Description

Fix Trainer optimizer checkpointing across the supported Trainer checkpoint paths.

As described in #4188 , there are several cases in which the optimizer is silently not checkpointed. Namely:
- When an optimizer is passed directly to Trainer without a default stepper, the trainer may install an optimizer hook. In this case, the trainer does not checkpoint its optimizer (it checkpoints the hook instead), but the optimizer hook's state dict is empty.
- When using a default optimization stepper, the default stepper does not own the optimizer, and no hook is added to _modules. Therefore, the trainer's optimizer is not checkpointed.
- When an optimizer is provided through an explicitly registered hook, trainer.optimizer is None. But, the optimizer hook's state dict is empty.

This can pose a large problem when reloading the trainer checkpoint to resume training etc.

This PR ensures that in all paths (legacy optimizer hooks and optimization steppers) the optimizer state will be checkpointed.

## Motivation and Context

Closes #4188.

## Types of changes

## Checklist

### DIFF
--- a/test/test_trainer.py
+++ b/test/test_trainer.py
@@ -163,6 +163,7 @@ _mocking_optim = MockingOptim()
 def mocking_trainer(
     file=None,
     optimizer=_mocking_optim,
+    optimization_stepper=None,
     checkpoint=None,
     checkpoint_rotation=None,
     checkpoint_metadata=None,
@@ -180,6 +181,7 @@ def mocking_trainer(
         optim_steps_per_batch=None,
         loss_module=loss_module,
         optimizer=optimizer,
+        optimization_stepper=optimization_stepper,
         logger=logger,
         save_trainer_file=file,
         checkpoint=checkpoint,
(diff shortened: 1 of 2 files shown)


vmoens:
Patched directly in `5aec792fa`:

- made optimizer restoration tolerant of legacy checkpoints that do not contain an `optimizer` payload, retaining the destination optimizer state in that case;
- deduplicated optimizer handling in the memmap component iteration;
- strengthened coverage with a real legacy torch-file roundtrip through `DefaultOptimizationStepper`, a missing-optimizer compatibility case, and the existing hook-owned path.

Validation: the checkpoint-component and nearby file-loading selections pass (`12 passed`), and the repository-pinned formatter plus flake8 pass on the changed files.

vmoens (approved):
LGTM

PR #4191, opened 2026-08-31
[Algorithm] Add backend-neutral MicroDuck task and notebook

vmoens:
## Stack

- Base: #4183
- This PR contains the backend-neutral follow-up only.

## Summary

- define one commanded longitudinal-velocity task against `MujocoEnv`, selectable across the `mujoco`, `mjx`, and `mujoco-torch` physics backends
- represent standing, forward motion, and backward motion with exact-zero, positive, and negative body-frame velocity commands sampled at reset
- expose the command directly in the TensorDict and include measured plus commanded longitudinal velocity in the 50-value policy observation
- rotate MuJoCo's world-frame free-joint linear velocity into the duck's body frame while using its already-local angular velocity convention
- prioritize smooth signed velocity tracking, retain uprightness/height as stabilizers, penalize lateral drift, roll/yaw rate, joint velocity, and action rate, and keep termination limited to physical falls or non-finite state
- smoothly gate a small nominal-pose reward to the zero command so it supports standing without fighting forward or backward gait motion
- retain normalized 14-joint position actions and the existing `0.35`-radian action scale; no torque or energy penalty is introduced
- replace only the upstream collision-class meshes with tight runtime box proxies while preserving visual meshes and the upstream source assets; this removes the accelerated-backend convex edge-pair memory blow-up
- train every command combination through the same script using an actor/value model with a shared GRU backbone
- attach a roughly 16K-transition replay buffer directly to the collector, write only complete trajectories, sample whole recurrent sequences with `SliceSampler`, run 10 PPO epochs, and erase the on-policy buffer before the next collection
- add deterministic fixed-command, multi-seed evaluation with return, tracking error, survival, episode length, and signed displacement metrics plus best-checkpoint retention
- log collection/inference and training throughput, process/device telemetry, reward, episodic return, gradient norm, ESS, clip fraction, KL, entropy, and PPO losses to W&B
- require an explicit W&B entity so logging cannot silently fall back to an unintended default workspace
- keep the notebook on the same recurrent complete-trajectory training path and preserve native/WASM rendering for signed commands

The shared robot assets remain in `pollen-robotics/microduck_rl`; nothing is translated or vendored. This intentionally does not claim parity with the richer MJLab task. Observation history, sensor delays, curricula, domain randomization, and BAM actuator behavior remain separate task layers.

## Validation

- focused behavior/PPO tests pass across native MuJoCo, MJX, and the fixed `mujoco-torch` main branch, covering signed reward direction, body-frame velocity rotation, zero-command-only pose stabilization, physical fall termination, partial command resets, collision proxying, recurrent PPO updates, complete-trajectory buffer fill, deterministic evaluation, best-checkpoint restoration, and explicit W&B entity selection
- the real upstream MicroDuck MJCF resets, steps, trains, evaluates, and renders on native MuJoCo CPU
- a full run completed 10,005,235 complete-trajectory transitions in 617 approximately 16K-transition collection/update cycles with 10 PPO epochs per cycle
- Ruff, notebook JSON validation, `git diff --check`, and the targeted MuJoCo tests pass locally

### Backend benchmark

The original two detailed sole meshes create roughly 107 million convex edge pairs per environment. Runtime collision boxes eliminate the OOM. A physics-only Apple-silicon CPU check measured:

| backend | environments | transitions/s |
| --- | ---: | ---: |
| native MuJoCo | 1 | 4,236 |
| native MuJoCo | 8 | 3,731 |
| MJX | 4 | 1,069 |
| MJX | 8 | 1,221 |
| `mujoco-torch` | 8 | 39 |

The `mujoco-torch` check used its current main branch because release 0.2.0 does not import with current MuJoCo enum layouts. It used less memory than MJX in the isolated run but was substantially slower in eager CPU execution. With the 128-unit GRU included, native collection measured about 974, 1,777, 1,969, and 1,866 transitions/s for 1, 4, 8, and 16 environments, so eight native environments are selected for laptop training.

### Training\n\n[View the complete 10M-transition training run on the author’s personal W&B project](https://wandb.ai/vmoens/torchrl-microduck-ppo/runs/mduck10m).\n\nThe stabilized native run completed 10,005,235 transitions. Its last collection/update cycle measured 2,161 collected transitions/s, 14,913 recurrent PPO training transitions/s, and 2.30 GB process RSS. The untrained deterministic baseline fell at step 61 for all commands and displaced -0.083 m.

Periodic deterministic evaluation selected the checkpoint at 1,634,584 transitions. Later policies improved stochastic episode lengths but collapsed toward forward motion for every command, demonstrating why best-checkpoint retention is needed. Across controlled reset-noise seeds 0–7, the selected checkpoint moved backward for 8/8 negative-command rollouts and forward for 5/8 positive-command rollouts. Mean episode lengths were 90.6, 153.2, and 117.0 steps for backward, stand, and forward commands; none survived the 500-step horizon. Mean signed displacements were -0.088, -0.055, and -0.008 m, and mean tracking errors were 0.194, 0.075, and 0.304 m/s.

[Watch the selected commanded-policy rollout](https://github.com/pytorch/rl/raw/refs/heads/codex/microduck-backend-neutral/examples/mujoco/assets/microduck_commanded_ppo.mp4). It uses the identical zero-noise `STAND` reset in all three panels. Backward and forward displace -0.090 and +0.031 m in the requested directions before falling at steps 90 and 130; standing drifts -0.085 m and falls at step 148.

An earlier `3e-4` run was intentionally stopped after clip fractions reached 0.33–0.47 and approximate KL reached 0.05–0.11. A five-buffer rehearsal at the stabilized `1e-4` default measured ESS 0.976–0.978, clip fraction 0.168–0.180, approximate KL 0.010–0.012, and pre-clip gradient norm 10–13. The full run linearly decayed that rate.

No standing, forward-walking, or backward-walking policy is claimed solved. CUDA validation was not run.


vmoens:
Superseeded

Issue #4192, opened 2026-09-01
[Feature Request] Transformers for first-class support in torchRL


theap06:
I'm proposing TransformerModule: the transformer parallel of the LSTMModule, built on the same four aspects.  Collection runs one step at a time against a fixed-shape KV cache carried per env; training processes full windows under a block-diagonal causal mask so attention never crosses episode boundaries. The load-bearing invariant, enforced by a parity test, is that the two paths produce identical outputs. One deliberate rule throughout: the KV cache is inference state, not experience. It lives in the rollout, never enters replay, and training recomputes state under current weights.

I am self-assigning this feature to myself. I have some experiments running for this

PR #4193, opened 2026-09-01
[Feature] Transformer Module 

theap06:
## Description

`TransformerModule`, the temporal-transformer analogue of `LSTMModule`: one causal transformer runs either over a `[B, T]` window (training) or one step at a time against a key/value cache (collection), with matching outputs, and the path is selected with `set_recurrent_mode` like the recurrent modules. Ships a reference `CausalTransformer` backbone, the `positions_from_is_init` and `segment_causal_mask_from_is_init` helpers, tests, a hot-path benchmark and docs. RFC in #4192.

**Where the cache lives.** Nothing travels in the tensordict. The key/value cache is inference state owned by the module instance (option B of the review): the backbone allocates it on the first cached step, one stream per batch position, in the dtype of its own projections; the module clears the streams flagged by `is_init`, restarts every stream when the parameters change, in place or swapped for other tensors (cached keys and values from previous weights are never mixed with new projections), and releases the cache on `reset_cache()`. Copies and pickled instances start with an empty cache, so a collector that ships the policy to workers never serializes a cache. Rollouts and replay buffers hold observations and features only, whatever the context length. The backbone contract is `forward(features, positions, mask=None, kv_cache=None)`, `new_kv_cache(batch_size, device=None)` and `reset_kv_cache(kv_cache, mask)`; the cache object is opaque to the module, so an adapter over an inference engine can keep it in the engine's own representation (option C), and a stream-keyed variant for asynchronous collectors (option A) can reuse the same contract once stable request context is exposed. Batches whose composition changes between calls are documented as unsupported for now.

**Training windows are episode-aligned.** Every row of a window must start with `is_init=True`, which complete-trajectory sampling provides; a window that starts mid-episode raises a `ValueError` rather than silently recomputing the prefix from position 0. Burn-in from a raw prefix under current weights is the planned mid-episode path (PR 3 of the RFC plan).

**Cache dtype.** The reference cache follows the projection weights, so `bfloat16` and `float64` modules get matching caches; under autocast the projected keys and values are cast to the cache dtype on write and the cache to the query dtype on read. Cached entries are detached and the cached-step path runs under `torch.no_grad()`: it is inference only, which is also what lets it compile (an in-place cache write inside an autograd-tracked forward is rejected by Inductor's functionalization). Both paths compile with `fullgraph=True` and match eager.

**Weight updates.** Parameter edits in place or swapped parameter tensors are detected on the next eager step (`(data_ptr, version)` per parameter). TorchRL's own weight-synchronization paths write through `.data` and do not bump the version counter, so they now notify the policy explicitly: `torchrl._utils.mark_weight_update(policy)` calls `mark_weight_update()` on every submodule that defines it, and is invoked from `Collector.update_policy_weights_` (the path multiprocess workers also take) and from `InferenceServer._mark_weight_update`. This also covers compiled modules, where the version check is skipped. User code that updates parameters by other means calls `mark_weight_update()` or `reset_cache()`.

**Window validation under compile.** The alignment check is data-dependent and no longer skipped when compiling: under default `torch.compile` it costs one graph break and still raises; `fullgraph=True` rejects it at compile time. `validate_windows=False` compiles the window path as a single graph and makes the caller responsible for alignment.

**Autocast.** The cache is allocated in the active autocast dtype (falling back to the projection dtype), so no cache-sized conversion happens per layer per step; an autocast dtype change restarts the streams. The cached-step benchmark has an `autocast_bf16` variant, and all timed callables synchronize on CUDA.

Compared with the first revision of this PR: the cache no longer travels under `("transformer_state", ...)`, `make_tensordict_primer` is gone, and the `TensorDictPrimer` dtype fix that revision carried is now its own PR (#4318).

## Motivation and Context

Gap 1 of the VLA roadmap: a generic temporal-transformer contract that reuses the recurrent modules' pillars (`is_init` resets, `set_recurrent_mode`, TensorDict I/O) without turning key/value caches into experience. See #4192.

## Types of changes

## Checklist

## Tests

`test/modules/test_transformer.py`:

- step-versus-window parity with per-environment mid-window resets, on CPU and CUDA, for `[3]` and `[2, 3]` batches;
- no state key written in either mode; reset forgets history; `max_seq_len` exceeded raises in both modes;
- window rows that do not start with `is_init=True` raise;
- an in-place parameter update, or a functional swap through `to_module`, restarts every stream (output equals a fresh episode start);
- the real `Collector.update_policy_weights_` path restarts every stream, with an eager and a compiled policy;
- the window alignment check still raises under default `torch.compile`;
- pickled and deep-copied instances carry no cache (pickle size is unchanged by stepping) and start from an episode start;
- `reset_cache()` and a batch-size change reallocate correctly;
- `bfloat16` and `float64` modules, and bf16 autocast, keep parity between the two paths;
- both paths under `torch.compile(fullgraph=True)` match eager;
- `env.rollout` and `SerialEnv` rollouts match the window recompute, with no state in the rollout;
- a `Collector` episode spanning two batches: the second batch alone is rejected by the window path, and the concatenated episode recomputed as a window matches the step outputs collected across both batches.

`benchmarks/test_transformer_module_benchmark.py`: cached-step throughput in fp32 and under bf16 autocast, window throughput, and a rollout whose returned storage is asserted to scale with the rollout length and not with the context size; timed callables synchronize on CUDA.


theap06:
@torchrlbot reviewer @vmoens

vmoens (changes requested):
I found three blocking correctness and scalability issues.

1. **The complete KV cache is materialized on every transition.**

   Each single-step call writes full K and V tensors of shape `[B, L, H, S, D]` under the next-state keys. Collector rollouts then stack both the current and next copies across time. This makes cache storage `O(T * L * S * hidden_size)`, or quadratic in trajectory length when `S ~= T`. For representative dimensions `L=24`, `hidden_size=896`, and `S=T=256` in fp16, current plus next caches alone require about 10.5 GiB per environment rollout. Excluding the keys before replay insertion does not avoid allocation in the collector rollout buffer.

   I do not think the K/V tensors should be environment carry data or replay data. The backbone or inference library should own their representation and update mechanics, while TorchRL only needs to own stream identity, reset, and lifetime when it is orchestrating inference. Several viable designs are available:

   - **Option A: server-side stateful inference adapter.** Keep one backend-native cache handle per inference client or environment stream on the policy device. The inference server gathers the active streams for a batch, invokes the cached decode path, and scatters the updated handles back to those streams. Requests carry observations plus a compact stream identity and `is_init`; responses contain policy outputs, not caches. `AsyncBatchedCollector` already creates one client per environment, and slot transports already have stable slot IDs, so this is the most natural fit for TorchRL auto-batching. It would require exposing stable request context to the stateful policy adapter.

   - **Option B: collector-local cached policy wrapper.** For a direct, non-server collector, give each collector worker its own inference-only wrapper with cache state indexed by the fixed environment batch slot. The wrapper consumes `is_init` and clears selected slots. This is simpler and keeps the transformer training module stateless, but the wrapper must not be shared across unrelated collectors or actor streams.

   - **Option C: delegate cache and continuous batching to an inference engine.** TorchRL passes an opaque request or session ID plus reset/cancel events; the engine owns paged cache allocation, batching, eviction, dtype, and device placement. This is likely the best scalable path for large pretrained transformers. TorchRL would provide a narrow adapter protocol rather than standardizing a K/V tensor layout.

   - **Option D: stateless bounded-context inference.** Keep a bounded raw observation/token history per stream and recompute the window on each policy call. This avoids a cache API and is useful as a simple reference path, at the cost of `O(S^2)` repeated attention work. The history should still be policy/inference state rather than transition state.

   Options A and C are compatible: TorchRL can own stream lifecycle while the selected backend owns the actual cache. Whichever design is chosen should reset on `is_init`, release state when a client or episode ends, and invalidate affected caches when inference weights change. Please add a cached-collection benchmark that measures peak memory and verifies that returned rollout storage does not scale as `T * S`.

2. **Recurrent/window mode silently loses prefix state when a batch starts mid-episode.**

   The window path ignores incoming cache and position state, while `positions_from_is_init` unconditionally assigns position zero to the first token. A normal collector batch can begin with `is_init=False` when an episode spans batches. In a two-batch reproduction, the first batch matched collected step outputs within `2.4e-7`, but the second batch started mid-episode and differed by `2.83`. Training therefore sees different positions and attention context from the behavior policy after the first chunk.

   There are also several defensible training semantics:

   - Require episode-aligned windows, validate that every row begins with `is_init=True`, and document/use complete-trajectory collection or sampling.
   - Store enough raw prefix context for a burn-in and recompute it under the current training weights before evaluating a truncated window.
   - Define fixed context boundaries and reset the inference cache and positions at those exact same boundaries, so collection and training intentionally share truncated-context semantics.

   I would avoid feeding collected inference K/V tensors into training: they were produced by behavior-policy weights, become stale after updates, and do not provide a gradient path through the prefix. Please extend the collector integration test to process the second batch of a continuing episode and compare recurrent outputs with the outputs collected step by step, or assert the selected episode-alignment contract.

3. **Cache dtype is incompatible with mixed precision and non-float32 modules.**

   Primer-created cache specs default to float32, and the fallback cache is created from the raw input dtype. Under autocast, projected K/V can be bfloat16 while the destination cache remains float32, so the indexed write raises `Index put requires the source and destination dtypes match`. The same failure occurs for a module converted to float64 or bfloat16 when its primer supplies float32 state.

   Suggested direction: let the selected transformer backend or inference engine establish the cache dtype from the actual K/V projection dtype. If the reference implementation retains a tensor cache, its allocation must follow that dtype. Add one compact parameterized non-float32 test plus an autocast collection test.

The submitted transformer tests and targeted primer test pass, and both execution modes pass a full-graph compile smoke test; the cases above are missing from the current coverage.

Please also add the required hot-path benchmark, remove the explicit `__init__ -> None` annotations, and inline the trivial `make_tuple` helper per the repository guidelines.

vmoens (changes requested):
Update after re-checking the current revision: it already resolves the original cache-placement and cache-representation concerns. No K/V state travels in the TensorDict, rollout, or replay data, and the cache object is deliberately opaque to `TransformerModule`: the backbone controls its representation, layout, dtype, device, allocation, and reset. The documented backbone contract also already allows an adapter over an inference engine. My previous ownership paragraph was therefore too broad.

The current tests pass locally (28 transformer tests; 10 benchmark smoke cases), and `git diff --check` is clean. I still see the following correctness and performance issues within the advertised fixed-batch, episode-aligned scope.

## 1. [P1] Cache invalidation misses TorchRL's actual weight-sync paths

`_restart_mask` attempts to detect parameter changes using `(data_ptr, _version)`, and skips that check while compiling. TorchRL's collector and inference-server update paths copy weights with `TensorDict.from_module(...).data.update_(weights.data)`. That path does not bump the destination `Parameter._version`.

I reproduced both variants:

- After the actual collector-style update, the detector sees no change and retains the cache; continuing versus a fresh cache differs by `1.2193`.
- Under `torch.compile`, even an ordinary parameter mutation is not checked; continuing versus a fresh cache differs by `2.9585`.

Please make invalidation explicit at TorchRL's policy-update boundary rather than relying exclusively on parameter internals. The existing `InferenceServer._mark_weight_update` / `update_policy_weights_` lifecycle can reset the cache or bump a cache generation. Direct collectors need the equivalent hook, or the public contract must require `reset_cache()` after a weight update and TorchRL's own sync paths must call it automatically.

One focused regression should exercise the real collector/server weight-update path and compare the next result with a fresh-cache baseline. If compiled weight updates are supported, include that mode in the same test.

## 2. [P1] The training-window alignment check disappears under compilation

The eager path rejects a training window whose first timestep is not an episode start, but the check is guarded by `not is_compiling()`. The same invalid input therefore raises eagerly and silently succeeds under `torch.compile(fullgraph=True)`, with the prefix position reset to zero. This changes the stated episode-alignment contract in a path for which compile support is advertised.

Possible fixes are a compile-compatible runtime assertion, validation immediately outside the compiled region, or an explicit documented compile precondition if runtime validation cannot be retained.

## 3. [P2] Autocast copies the full cache on every layer and token

The reference cache is allocated in the projection-weight dtype, then `cache_k.to(q.dtype)` / `cache_v.to(q.dtype)` converts the full cache during every attention call. With FP32 weights under AMP, this makes two full cache-sized copies per layer per generated token. At the benchmark shape (`B=32`, `L=4`, `S=256`, hidden size 256), the persistent cache is 64 MiB and the profiler reports about 34 MiB of `_to_copy` allocation per token, about 32 MiB of which is the cache cast.

Options are to allocate the reference cache lazily in the active K/V dtype, pass the compute dtype into cache creation, or let a backend-native implementation avoid the conversion. A compute-dtype change should recreate or invalidate the reference cache. An autocast throughput/peak-memory benchmark would protect this behavior.

## 4. [P2] The CUDA benchmark measures enqueue time

The benchmark selects CUDA when available but does not synchronize around the timed callable, so pytest-benchmark measures kernel enqueue latency rather than execution. Please synchronize, use CUDA events, or keep this benchmark CPU-only.

## Integration notes, not blockers for the documented scope

The current module explicitly scopes cached inference to a stable batch composition and explicitly says that partial async batches need a stream-keyed cache. Given that documentation, I no longer consider implementing async routing a requirement for this PR.

The limitation is nevertheless relevant to the merged async stack: `InferenceServer` retains stable slot IDs, but compacts whichever requests are ready into changing model-batch rows. In a real `InferenceServer` / `SlotTransport` reproduction, serving A and B together and then only B causes B's continuing output to differ from an independent per-stream baseline by `3.0291743`; a same-sized reorder can assign the wrong history silently. The remaining integration problem is only how stable server stream identity reaches the cache, not who owns the K/V representation.

Reasonable follow-up options are:

1. Add an inference-server stateful-policy adapter that passes stable stream IDs plus reset/cancel/release and weight-generation events to the opaque backend cache.
2. Extend `TransformerModule` with a cache bank keyed by stable stream IDs.
3. Keep the present fixed-composition scope and have `InferenceServer` reject this module/configuration rather than silently accepting unsupported dynamic batching.

Likewise, `StreamingSliceSampler` is not currently compatible with the episode-aligned training contract: it marks each sampled slice start as `is_init=True`, but later slices do not contain the preceding transformer context. For an 8-step episode split into two 4-step windows, the first window matches full-episode execution (`2.38e-7`) and the second differs by `3.2958`. This does not need to expand this PR if whole-episode sampling is the intended supported path. Future streaming support can use raw-prefix burn-in, identical collection/replay context boundaries, or whole episodes with padding/attention masks. Collected K/V tensors should remain out of replay.

`CudaGraphModule` integration is also follow-up unless this PR claims it. Any future coverage should use a GPU-marked test with partial/reordered serving; the test needs both `@pytest.mark.gpu` and the CUDA `skipif` under the TorchRL CI convention.

Finally, now that #4318 is merged, this PR should be retargeted from `theap06/primer-spec-dtype` to `main`.

vmoens:
Thanks for this. We have a bunch of tutorials for RNNs, would be nice to have one for Transformers as well.
Along the same line, I've found that GTrXL is a good algo to implement and I'd be curious to see it integrated with this module.
EDIT
Working on a version of this, will ping you when it's ready

torchrl/collectors/_single.py
```diff
@@ -1731,6 +1732,7 @@ def _maybe_fallback_update(
             and self._policy_w_state_dict is not None
         ):
             TensorDict.from_module(self._policy_w_state_dict).data.update_(weights.data)
+            mark_weight_update(self._policy_w_state_dict)
```
vmoens, line 1735: [P1] Invalidate the cache after legacy weight updates too

This notification is confined to _maybe_fallback_update, which Collector.update_policy_weights_ bypasses when a legacy weight_updater is configured. With VanillaWeightUpdater.from_policy(policy) and a torch.compile-wrapped transformer, the weights are updated but the cache is retained: the parameter-version fallback is disabled while compiling. In a two-batch collector reproduction, synchronizing weights between batches leaves the next embedding with a maximum absolute difference of 1.10456 against an otherwise identical fresh-cache module. The eager version restarts correctly. Please notify the policy after every supported local weight-application path, including the still-supported legacy updater, and extend the collector regression to cover this route.

torchrl/modules/tensordict_module/transformer.py
```diff
@@ -0,0 +1,752 @@
...
+            self.num_layers,
+            self.num_heads,
+            self.max_seq_len,
+            self.head_dim,
+        )
+        device = weight.device if device is None else device
+        dtype = weight.dtype if dtype is None else dtype
+        return (
+            torch.zeros(shape, dtype=dtype, device=device),
+            torch.zeros(shape, dtype=dtype, device=device),
```
vmoens, line 271: [P2] Keep the mutable cache usable after leaving inference mode

If the first cached step runs inside torch.inference_mode(), these allocations create inference tensors. On the next ordinary module call, the internal torch.no_grad() context does not permit their in-place mutation, and reset_kv_cache fails at masked_fill_ with 'Inplace update to inference tensor outside InferenceMode is not allowed'. This also fails when the next input has is_init=True, so an episode reset does not recover. Allocate persistent mutable cache state outside inference mode, or explicitly handle this transition, and add a regression that performs an inference-mode step followed by an ordinary cached step.

torchrl/modules/tensordict_module/transformer.py
```diff
@@ -0,0 +1,752 @@
...
+            or positions.device != value.device
+            or cache_dtype != self._cache_dtype
+        )
+        if not stale and not is_compiling():
+            version = self._current_weights_version()
+            stale = version != self._weights_version
+        if stale:
+            self._kv_cache = self.transformer.new_kv_cache(
+                batch_size, device=value.device, dtype=cache_dtype
+            )
```
vmoens, line 684: [P2] Align cache creation with the documented backbone protocol

CausalTransformer documents the interchangeable-backbone contract as new_kv_cache(batch_size, device=None), but this call now unconditionally supplies dtype, including dtype=None outside autocast. A custom backbone implementing exactly that advertised signature passes the constructor checks and then raises TypeError on its first cached step. The current custom-backbone test passes another CausalTransformer, so it cannot catch this mismatch. Please either include the optional dtype argument in the documented adapter contract or avoid requiring it from conforming backbones, and exercise an independent adapter in the test.

vmoens (changes requested):
Re-reviewed f60073b17dff1b50b167ae5205c4d376b9062c7e. The fixes from my previous review are present, but I found three remaining issues, detailed inline: compiled policies retain stale caches with the legacy weight updater; a cache allocated under inference mode breaks subsequent ordinary calls; and the custom-backbone call does not match the documented protocol.

Validation: all 26 transformer tests and 14 benchmark smoke cases passed locally on CPU, including the existing compile tests. I separately reproduced each reported issue; the legacy-updater reproduction also fails with the default Inductor backend. The benchmark run used --benchmark-disable, so it validates execution and assertions, not performance. CUDA execution was not verified.

PR #4196, opened 2026-09-01
[Algorithm] Add DMC Walker reproduction script for DreamerV3

theap06:
## Summary

- add `reproduce_dmc_walker.sh`, a single entry point that runs the walker preset's benchmark seeds and acceptance gate through `benchmark.py`
- `--smoke` validates the same pipeline (training, per-seed jsonl metrics, windowed aggregation, gate check) with tiny settings in minutes instead of the full 1.1M-step runs

## Results

Three seeds on a single MI300X reach final evaluation returns of 978.7 / 971.8 / 973.7 at 1.1M environment steps, final median 973.7, at or above the JAX reference plateau. The reference reports stochastic training-episode returns and ends at 490k steps. TorchRL crosses 700 roughly 100-150k steps after the reference median; final performance matches.

## Validation

- `reproduce_dmc_walker.sh --smoke` runs end to end on CPU (exit 0, summary.json written, gate evaluated)
- pre-commit passes on the added file
- the full reproduction was intentionally not rerun for this PR; the smoke validates the mechanics

image

### DIFF
--- a/sota-implementations/dreamer_v3/README.md
+++ b/sota-implementations/dreamer_v3/README.md
@@ -53,7 +53,20 @@ CPU execution.
 For a three-seed median and interquartile reproduction run:
 
 ```bash
-python sota-implementations/dreamer_v3/benchmark.py --output-dir dmc_walker_runs
+./sota-implementations/dreamer_v3/reproduce_dmc_walker.sh
+```
+
+For the fastest supported accelerator path, enable the compiled RSSM scan
+(unrolled eight steps at a time):
+
+```bash
+./sota-implementations/dreamer_v3/reproduce_dmc_walker.sh --fast
+```
+
+Compilation has an up-front cost, so the short validation remains eager:
+
+```bash
+./sota-implementations/dreamer_v3/reproduce_dmc_walker.sh --smoke
 ```
 
 The benchmark writes one metrics file per seed plus `summary.json`, aggregates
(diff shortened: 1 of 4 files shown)


unography:
@theap06 do you have timings for how long the torch run took as compared to the JAX? 
with my benchmarks, torch was still quite slow so I couldn't run for the whole 1M steps, working on a PR to improve the speed

vmoens:
> @theap06 do you have timings for how long the torch run took as compared to the JAX? with my benchmarks, torch was still quite slow so I couldn't run for the whole 1M steps, working on a PR to improve the speed

Good question. We don’t yet have a same-hardware, end-to-end JAX vs. TorchRL timing, so the current 1.1M-step results don’t support a defensible “optimal speed” claim.

I’ve updated this PR with an easy fast path:
`./sota-implementations/dreamer_v3/reproduce_dmc_walker.sh --fast`

This selects compiled RSSM scan with `rssm_scan_unroll=8`, forwards additional Hydra overrides, and logs the resolved device/backend. The smoke path remains eager.

The Walker-shaped RSSM microbenchmark in #4132 improved forward+backward from roughly 216 ms to 93 ms on an A10G, but that is not an end-to-end comparison with JAX. The Triton BlockGRU work in #4160 is not wired into the DreamerV3 RSSM yet and is NVIDIA-only, so it does not apply to the MI300X runs.

The remaining work is a same-hardware end-to-end comparison, including compilation warm-up, and a full three-seed acceptance run, since compiled scan changes the RNG stream.

vmoens (approved):
LGTM thanks for running this!

PR #4202, opened 2026-09-02
[BugFix] Compile the mujoco-torch physics step with fullgraph=True by default

vmoens:
## Summary

`_TorchBackend._build_step_fn` wrapped `torch.vmap(mujoco_torch.step)` in `torch.compile` without `fullgraph`. On torch nightly a graph break inside the batched step was papered over by splitting it into eager fragments, and once Dynamo's recompile limit was hit the whole step silently fell back to eager, so `compile_step=True` gave no speed-up and hid the underlying tracing bug (fixed on the mujoco-torch side in vmoens/mujoco-torch: scan cache keys and `cholesky_ex`).

## Changes

- `_TorchBackend._build_step_fn`: compile with `fullgraph=True` by default, still overridable through `compile_kwargs`.
- `MujocoEnv` docstring: document the default.
- `test_torch_backend_compile_smoke`: reset Dynamo, and assert the batched Hopper step compiles into exactly one graph across the rollout (a recompile would otherwise degrade into an eager fallback).

## Tested

- `pytest test/libs/test_mujoco.py -k test_torch_backend_compile_smoke` passes with torch `2.15.0.dev20260901`, tensordict main, mujoco-torch with the companion fix (47 s on CPU, single graph).
- `HopperEnv(num_envs=1|2, backend="mujoco-torch", compile_step=True).rollout(3)` completes with one graph.
- 8-env MicroDuck (`MicroDuckVelocityEnv`, `frame_skip=4`) with `compile_step=True`: two full 500-step episode batches after compile; warm throughput 180–193 transitions/s vs 44 eager on CPU.

### DIFF
--- a/test/libs/test_mujoco.py
+++ b/test/libs/test_mujoco.py
@@ -1026,10 +1026,26 @@ class TestMujoco:
 
     @pytest.mark.skipif(not _has_mujoco_torch, reason="mujoco-torch not installed")
     def test_torch_backend_compile_smoke(self):
-        """``compile_step=True`` must not raise on the default backend."""
+        """``compile_step=True`` compiles the batched step as a single graph.
+
+        The backend compiles with ``fullgraph=True`` by default, so a graph
+        break raises here instead of silently splitting the step into eager
+        fragments.  The graph counter guards against recompiles across steps,
+        which would otherwise degrade into an eager fallback once Dynamo's
+        recompile limit is hit.
+        """
+        import torch._dynamo
+        import torch._dynamo.utils
+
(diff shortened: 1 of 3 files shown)


vmoens:
CI note: the single failure in `tests-cpu (3.13, bulk)` is `test/rb/test_rb_core.py::test_replay_buffer_load_state_dict_waits_for_in_flight_samples` (replay-buffer concurrency test, `assert False`), unrelated to the MuJoCo backend change here; `main` is red on its last three runs as well. The `unittests-gym` job failed at the docker exec step before reporting tests.

PR #4204, opened 2026-09-02
[Feature] Add MicroDuckEnv and MuJoCo contact and site accessors

vmoens:
## Description

- `MujocoEnv.geom_contacts(names)` and `MujocoEnv.site_positions(names)`, backed by new `mj_model`, `geom_contacts` and `site_positions` methods on the native, mjx and mujoco-torch physics backends, so tasks can use contact-based metrics without reaching into backend internals.
- `torchrl.envs.MicroDuckEnv`: a commanded longitudinal-velocity locomotion task for Pollen Robotics' MicroDuck biped built on `MujocoEnv`, so the same task runs on all three backends.
  - The walking MJCF is resolved from `microduck_root`, then `MICRODUCK_RL_ROOT`, then an installed `mjlab_microduck` package, then a pinned upstream commit under `root` (default `~/.cache/torchrl/microduck`) that `download=True` fetches on demand, following the dataset `root`/`download` convention (`"force"` re-downloads). Nothing is vendored, and the default raises an error listing every option. The scene is resolved once before the native backend spawns its workers. Collision-class meshes are replaced with box proxies at load time so the accelerated backends fit in memory.
  - `DEFAULT_BACKEND = "mujoco"`; the velocity command is part of `state_spec` so a command passed at reset survives `TransformedEnv`.
  - Reward: the mjlab velocity-task recipe, per-second weights times the 50 Hz control period. Gaussian tracking of the commanded velocity (std 0.1, lateral to zero) and of zero yaw rate, Gaussian uprightness, a nominal-pose term that is tight when standing and loose when walking, and contact-based gait terms under a nonzero command: foot air time inside a swing window, the height of the gait clock's swing foot toward a 2 cm target (dense, contact or not), correct single support with respect to the clock (planted feet earn nothing), and a double-support penalty. Small motion costs and a fixed fall penalty. Each of these choices removed a measured failure mode of from-scratch PPO: directional reward ignored the command magnitude, half phase credit and contact-gated swing reward made standing the optimum.
  - Training aids: `command_range`, `warm_start_velocity`/`warm_start_fraction` (resets already in motion, as upstream), `joint_reset_noise_scale` (diverse reset poses), `gait_frequency_per_mps` (clock cadence follows the command), `observe_lateral_velocity`, and `reward_scales` to override any reward attribute per instance.
  - Reward-component, pose and contact diagnostics are opt-in (`diagnostics=True`); `foot_contacts()` and `foot_heights()` expose gait signals.

Task parameters live in `MicroDuckTask`, a frozen dataclass (commands, reset distribution, action scale, gait clock, observation and reward options); the constructor keeps only the asset and simulator arguments (`microduck_root`, `root`, `download`, `backend`, `low_cost_collisions`, `max_episode_steps` and the `MujocoEnv` kwargs). `MicroDuckEnv.tracking_task`, `standing_task` and `speed_range_task` return presets that accept field overrides; `speed_range_task` encodes the command-following gait clock validated in the ablations. `compute_reward=False` zeroes the built-in reward so a transform can supply another one, and `diagnostics=True` still exposes every term. The pin only governs what `download` fetches: any `microduck_rl` revision works through `microduck_root`, `MICRODUCK_RL_ROOT` or the installed package, and the load-time layout check guards compatibility. The class docstring separates `Args` from `Keyword Args` with a type on every entry, and its examples show download, scaling across backends, task selection, video recording with `VideoRecorder`, the reward breakdown and a reward override. Both classes are listed under `docs/source/reference/envs_libraries.rst`.

The backend name alone decides how `num_envs` simulators are batched: `mujoco-torch` (the `MujocoEnv` default, which `MicroDuckEnv` now keeps) and `mjx` vectorize them inside the simulator; `mujoco` runs the C bindings with one process per env and is the explicit CPU-only fallback. The earlier `DEFAULT_BACKEND = "mujoco"` override is gone.

The warm start points along the commanded direction: the command is drawn while the initial state is sampled and consumed by the reset hooks, so a negative command starts the robot moving backward. The mujoco-torch backend turns an import failure caused by a MuJoCo version mismatch into an ImportError that names both versions.

## Motivation and Context

Part of the MicroDuck stack: #4183 -> #4203 -> this PR -> #4205. Moves the task out of `examples/` into the library next to `SatelliteEnv` and `CubeBowlEnv`. With this env and the example in #4205, a GRU policy with no gait prior learns to walk from scratch on a laptop CPU in about 3.5M transitions (see the #4205 README).

## Types of changes

## Checklist

## Validation

- `pytest test/libs/test_mujoco.py -k "microduck or geom_contacts"` passes locally on the native, mjx and mujoco-torch (main) backends with a synthetic 14-actuator fixture that stands on two colliding feet.
- The full `test/libs/test_mujoco.py` passes locally except `test_torch_backend_compile_smoke`, the pre-existing mujoco-torch Dynamo recursion addressed by #4202.
- On the real `microduck_rl` scene the closed-form gait from #4205 clears the contact-based walking gate on every seed, and from-scratch PPO reaches forward walking (see #4205).
- Pinned linters pass on the changed files.


test/libs/test_mujoco.py
vmoens: 579 LoC for one env is a lot.

torchrl/envs/custom/mujoco/microduck.py
```diff
@@ -0,0 +1,1089 @@
...
+from tensordict import TensorDictBase
+from torchrl._utils import logger as torchrl_logger
+from torchrl.data.tensor_specs import Binary, Bounded, Composite, Unbounded
+from torchrl.envs.custom.mujoco._backends import BackendName
+from torchrl.envs.custom.mujoco.base import _MujocoMeta, MujocoEnv
+
+MICRODUCK_RL_COMMIT = "d424a0c899f6b33cbd3daeb279913134349c0b63"
+MICRODUCK_RL_ARCHIVE_URL = (
+    "https://github.com/pollen-robotics/microduck_rl/archive/{commit}.zip"
+)
```
vmoens: why is this needed? We pin a version of microduck? How flexible is this?

torchrl/envs/custom/mujoco/microduck.py
```diff
@@ -0,0 +1,1089 @@
...
+    commit under ``root``, which ``download=True`` fetches when absent. A
+    ``microduck_rl`` checkout, its package directory, or the ``scene_walk.xml``
+    file itself are all accepted.
+
+    Args:
+        microduck_root: ``microduck_rl`` checkout, ``mjlab_microduck`` package
+            directory, or path to ``scene_walk.xml``. Defaults to the
+            :attr:`ROOT_ENV_VAR` environment variable, the installed package,
+            or a download under ``root``.
+        root: directory holding downloaded ``microduck_rl`` checkouts. Defaults
```
vmoens: from here below these are keyword args, not args

vmoens: always indicate the type in the docstrings. For example it's not clear at all what `commanded_x_velocity` is: from the docstrings i assume it's a float but it looks like a tuple is explected. What happens if the same value is duplicated? Do we sample randomly from that? Can I override it by passing a tensordict?

torchrl/envs/custom/mujoco/microduck.py
```diff
@@ -0,0 +1,1089 @@
...
+            the class defaults on this instance.
+        gait_phase_offset: phase of the gait clock at the first step, in
+            radians. Defaults to ``-1.5237``.
+        gait_ramp_duration_s: duration over which the gait ramp feature grows
+            from zero to one after a reset. Defaults to ``0.4``.
+        max_episode_steps: truncation horizon. Defaults to ``500``.
+        \*\*kwargs: forwarded to :class:`~torchrl.envs.MujocoEnv`. ``xml_path``
+            and ``patch_xml`` are not accepted.
+
+    Examples:
```
vmoens: honestly this is not even an example, merely a comment, and not a useful one. Give some flesh: it should get people excited. Example: show how you can render (maybe with a torchrl transform). Show how to scale, how to compile, how to switch backends, how to select a task...

torchrl/envs/custom/mujoco/microduck.py
```diff
@@ -0,0 +1,1159 @@
...
+        "left_hip_pitch",
+        "left_knee",
+        "left_ankle",
+        "neck_pitch",
+        "head_pitch",
+        "head_yaw",
+        "head_roll",
+        "right_hip_yaw",
+        "right_hip_roll",
+        "right_hip_pitch",
```
vmoens: we should be able to pass the reward function we want and override the default one.
This can be done via a transform, so maybe we don't want to do it in the env itself - but then we should be able to toggle off the reward computation.
There are MANY kwargs in this constructor. Should we have a config object, returned as a classmethod?
This class method would take the kwargs that we have here and return the config object. 
Then we could implement more of these class methods to encode various tasks.

`commanded_x_velocity`

torchrl/envs/custom/mujoco/microduck.py
```diff
@@ -0,0 +1,1257 @@
...
+            ``diagnostic_*`` keys. Off by default because it roughly doubles
+            the per-step task cost.
+
+    Examples:
+        >>> from torchrl.envs import MicroDuckEnv, MicroDuckTask
+        >>> task = MicroDuckEnv.speed_range_task(0.1, 0.3, action_scale=1.0)
+        >>> task.command_range, task.gait_frequency_per_mps, task.action_scale
+        ((0.1, 0.3), 5.0, 1.0)
+        >>> MicroDuckTask(commanded_x_velocity=0.2).commanded_x_velocity
+        (0.2,)
```
vmoens: you should at least show how to use it with the env

PR #4205, opened 2026-09-02
[Algorithm] MicroDuck recurrent PPO and closed-form gait examples

vmoens:
## Description

`examples/microduck/`, replacing the `examples/mujoco/` content of #4191:

- `heuristic_gait.py`: `MicroDuckGaitActor`, a `TensorDictModuleBase` that computes the closed-form gait from the `MicroDuckEnv` observation; it is the baseline (`env.rollout(steps, actor)`), the `rlrender` policy and the PPO prior. `gait_metrics` reads a `diagnostics=True` rollout and counts swing phases from foot contacts with tensor ops, so forward motion alone is not mistaken for walking.
- `ppo_mujoco.py`: recurrent PPO with a GRU backbone shared by actor and critic, configured by Hydra from `config.yaml` like the sota-implementations. `policy.head=gaussian` trains from scratch; `policy.head=gait-residual` learns a bounded residual around the gait. A `Collector` with `trajs_per_batch=1` writes every finished episode whole into a `TensorDictReplayBuffer`, GAE runs once over the buffer in recurrent mode, `SliceSampler` draws whole episodes for the minibatches, and the buffer is emptied before collecting again. `KLAdaptiveLR` keeps the KL near target; one `torchrl.collectors.Evaluator` per velocity command runs the deterministic evaluations (`microduck_metrics` reads the padded trajectory batch) and the best checkpoint is kept by survival, direction, forward speed and return, plus the current one at every evaluation with `evaluation.latest_checkpoint_path`. Checkpoints are unified TorchRL checkpoints written with `save_render_checkpoint` (actor as the policy, critic alongside, `MicroDuckTask` and policy kwargs as environment metadata, the Hydra config), so `rlrender` rebuilds actor and env from the checkpoint alone and `policy.init_from` resumes from it. `make_env` takes the `env` section of the config (or the config a checkpoint recorded) and builds the `MicroDuckTask` from `env.task`; logging goes through `get_logger` (W&B, csv or tensorboard). The config defaults to 1,024 vectorized `mujoco-torch` simulators on the available accelerator; `env.backend=mujoco env.parallel=true env.num_envs=16` is the CPU fallback the README commands use, and `smoke=true` pins one native simulator on CPU.
- `README.md`, including the rlrender commands that generate the MuJoCo WASM notebooks. No notebooks or binaries are shipped.

## Motivation and Context

Top of the MicroDuck stack: #4183 -> #4203 -> #4204 -> this PR. Supersedes #4191.

## Types of changes

## Checklist

## Validation

- `pytest test/libs/test_mujoco.py -k microduck_example` passes: the PPO test asserts that only complete episodes reach the buffer, that the gait-residual head starts at the closed-form gait, that best and latest checkpoints are written and restored, and that `make_env`/`make_render_policy` rebuild the trained configuration from a checkpoint.
- `ppo_mujoco.py --smoke` (both heads, `--init-from`, the env flags) and `heuristic_gait.py` run on the real `microduck_rl` scene; `rlrender` exports mp4 and notebooks from the checkpoints.
- From-scratch result on an Apple-silicon CPU with 16 native workers (about 3,300 transitions/s): with action scale 1 rad, unit initial exploration scale, joint reset noise, warm starts and the contact-based reward of #4204, the GRU policy walks forward by about 3.5M transitions and reaches about 1.5 m in 10 s by 7.5M with every evaluation rollout surviving ([W&B run](https://wandb.ai/vmoens/torchrl-microduck-ppo/runs/9jgfqj8p)). Four earlier from-scratch attempts converged to standing still; each was diagnosed with the env's reward diagnostics and fixed in #4204.
- 5M-per-run ablations (README sections "Training from scratch" and "Ablations"). From scratch, the command-scaled gait clock (`--gait-frequency-hz 1.0 --gait-frequency-per-mps 5.0`, [kkz2ctnm](https://wandb.ai/vmoens/torchrl-microduck-ppo/runs/kkz2ctnm)) is the only change that makes the speed follow the command (0.11 / 0.15 / 0.17 m/s at 0.1 / 0.2 / 0.3 m/s; speed error 0.076 against 0.109 for the baseline at 5M), so the README recipe includes it. Smaller minibatches ([vwk0hgu2](https://wandb.ai/vmoens/torchrl-microduck-ppo/runs/vwk0hgu2)) learn faster early but end level with the baseline; a lower entropy coefficient ([fdit1i9n](https://wandb.ai/vmoens/torchrl-microduck-ppo/runs/fdit1i9n)) and extra velocity observations ([iirlbbkv](https://wandb.ai/vmoens/torchrl-microduck-ppo/runs/iirlbbkv)) do not help. Continuing the baseline checkpoint with halved contact shaping and a doubled tracking weight ([ahqvrivb](https://wandb.ai/vmoens/torchrl-microduck-ppo/runs/ahqvrivb)) gives the best policy: speed error 0.053 m/s, speeds 0.15 / 0.19 / 0.21 m/s that follow the command, every rollout surviving; continuing unchanged ([4ae0vl4t](https://wandb.ai/vmoens/torchrl-microduck-ppo/runs/4ae0vl4t)) plateaus at 0.085. All policies drift in heading (50 to 160 degrees in 10 s) because the yaw-rate term is loose, so the README judges speed by the tracking error and names a tighter yaw or heading term as the next reward change. Defaults are unchanged: the clock flags would desynchronize the reward clock from the closed-form gait that the default `gait-residual` head follows.


examples/microduck/microduck_ppo.ipynb
vmoens: Let's try not to commit ipynbs

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+def heading_xy(quaternion: torch.Tensor) -> torch.Tensor:
+    """Return the planar body x-axis of a wxyz quaternion in the world frame."""
+    quaternion = quaternion / quaternion.norm(dim=-1, keepdim=True).clamp_min(1e-8)
+    w, x, y, z = quaternion.unbind(-1)
+    return torch.stack(
+        (1.0 - 2.0 * (y.square() + z.square()), 2.0 * (x * y + w * z)), dim=-1
+    )
+
+
+class MicroDuckGaitPolicy:
```
vmoens: Should be a TDModule

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+    def env_kwargs(self) -> dict[str, float]:
+        """Return the :class:`~torchrl.envs.MicroDuckEnv` gait-clock arguments."""
+        return {
+            "gait_frequency_hz": self.frequency_hz,
+            "gait_phase_offset": self.phase_offset,
+            "gait_ramp_duration_s": self.ramp_duration_s,
+        }
+
+
+def gait_action(config: MicroDuckGaitConfig, observation: torch.Tensor) -> torch.Tensor:
```
vmoens: Looks like something I'd have bundled in an actor class

vmoens: The actor class would take the config as input

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+    action[..., 3:4] = config.knee_amplitude * left_swing
+    action[..., 12:13] = -config.knee_amplitude * right_swing
+    action[..., 4:5] = ankle_pitch_correction - config.ankle_amplitude * left_swing
+    action[..., 13:14] = -ankle_pitch_correction + config.ankle_amplitude * right_swing
+    action[..., 1:2] = config.lateral_amplitude * lateral_wave
+    action[..., 10:11] = config.lateral_amplitude * lateral_wave
+    return action.clamp(-1.0, 1.0)
+
+
+def heading_xy(quaternion: torch.Tensor) -> torch.Tensor:
```
vmoens: Ditto

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+    def __init__(self, config: MicroDuckGaitConfig):
+        self.config = config
+
+    def __call__(self, tensordict: TensorDictBase) -> TensorDictBase:
+        tensordict.set("action", gait_action(self.config, tensordict["observation"]))
+        return tensordict
+
+
+def make_render_policy(spec: RenderPolicySpec) -> MicroDuckGaitPolicy:
+    """Build the closed-form policy for an ``rlrender`` rollout.
```
vmoens: Unclear, can you give an example?

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+        camera_id=camera_id,
+        render_width=render_width,
+        render_height=render_height,
+        **gait.env_kwargs(),
+    )
+
+
+@dataclass(frozen=True)
+class GaitTrial:
+    """Contact-derived metrics of one fixed-seed rollout."""
```
vmoens: This doesn't say anything about what this is.
Hard to justify its existence with so few words

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+        elif swing_is_left != current_swing_is_left:
+            if support_run_max >= MIN_SINGLE_SUPPORT_STEPS:
+                if current_swing_is_left:
+                    left_swing_phases += 1
+                else:
+                    right_swing_phases += 1
+            current_swing_is_left = swing_is_left
+            support_run = support_run_max = 0
+
+        tensordict.set("action", gait_action(config, observation))
```
vmoens: We need a policy that's a TDModule here

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+
+    max_abs_pitch = abs(float(tensordict["observation"][0, 0].clamp(-1, 1).asin()))
+    heights = env.foot_heights()[0]
+    left_height_max, right_height_max = float(heights[0]), float(heights[1])
+    left_swing_phases = right_swing_phases = 0
+    left_support = right_support = 0
+    current_swing_is_left: bool | None = None
+    support_run = support_run_max = 0
+    survived = True
+    episode_length = 0
```
vmoens: What's all of this? Can't we just have an env reset?

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+            self.survived
+            and self.signed_displacement > 0.0
+            and self.max_abs_pitch < MAX_WALKING_PITCH
+            and self.left_swing_phases >= MIN_SWING_PHASES
+            and self.right_swing_phases >= MIN_SWING_PHASES
+        )
+
+
+@torch.no_grad()
+def run_trial(
```
vmoens: I don't like this much. A rollout should be as simple as `env.rollout(policy=policy)`. If you need intermediate transforms and logging, use env transforms. Stay within the contracts and APIs of torchrl. All this boilerplate drives me mad. It's unreadable. 
Why do we output a `GaitTrial`? What does this do? Why would anyone want to see that?

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+        left_swing_phases=left_swing_phases,
+        right_swing_phases=right_swing_phases,
+        left_single_support_steps=left_support,
+        right_single_support_steps=right_support,
+        left_foot_height_max=left_height_max,
+        right_foot_height_max=right_height_max,
+    )
+
+
+def evaluate_gait(
```
vmoens: this is the kind of thing that has 0 value

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+            config,
+            seed=seed,
+            steps=steps,
+            commanded_x_velocity=commanded_x_velocity,
+        )
+        for seed in seeds
+    ]
+
+
+def summarize(trials: Sequence[GaitTrial]) -> dict[str, float]:
```
vmoens: THIS IS EXACTLY WHY I INVENTED TENSORDICT 
I mean - ok you have min and mean - but aside from that the reductions could be made much more compact. You could do `td.apply(lambda key, val: val.min() if key.endswith("min") else val.mean(), named=True)` or similar.

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+        ),
+        "left_foot_height_max_min": float(min(t.left_foot_height_max for t in trials)),
+        "right_foot_height_max_min": float(
+            min(t.right_foot_height_max for t in trials)
+        ),
+        "wrong_way_rollouts": float(np.count_nonzero(displacements <= 0.0)),
+    }
+
+
+def search_key(trials: Sequence[GaitTrial]) -> tuple[float, ...]:
```
vmoens: again this seems to have very little value. Why do we have so many functions in this file?

examples/microduck/heuristic_gait.py
```diff
@@ -0,0 +1,549 @@
...
+        summary["survival_rate"],
+        summary["episode_length_mean"],
+        min(summary["left_swing_phases_min"], summary["right_swing_phases_min"]),
+        summary["walking_success_rate"],
+        -summary["max_abs_pitch_max"],
+        summary["forward_speed_mean"],
+    )
+
+
+def search_configs(
```
vmoens: what does this do?

examples/microduck/ppo_mujoco.py
```diff
@@ -0,0 +1,1111 @@
...
+    return actor
+
+
+# ----------------------------------------------------------------------
+# Evaluation
+# ----------------------------------------------------------------------
+
+
+@torch.no_grad()
+def evaluate_policy(
```
vmoens: let's use an evaluator since we have it in torchrl (perhaps)? Wdyt?

examples/microduck/ppo_mujoco.py
```diff
@@ -0,0 +1,1111 @@
...
+
+    history: list[dict[str, float]] = []
+    collected = 0
+    iteration = 0
+    best_score: tuple[float, ...] | None = None
+    best_state: tuple[dict, dict] | None = None
+    checkpoint_path = Path(best_checkpoint_path) if best_checkpoint_path else None
+    latest_path = Path(latest_checkpoint_path) if latest_checkpoint_path else None
+
+    def save_checkpoint(path: Path, step: int, metrics, score) -> None:
```
vmoens: do we want to use torchrl checkpointing API?

examples/microduck/ppo_mujoco.py
```diff
@@ -0,0 +1,1111 @@
...
+                "evaluation_score": list(score),
+                "actor": actor.state_dict(),
+                "critic": critic.state_dict(),
+                "policy_kwargs": dict(policy_kwargs or {}),
+                "env_kwargs": dict(env_kwargs or {}),
+            },
+            path,
+        )
+
+    def evaluate(step: int) -> dict[str, float]:
```
vmoens: again, maybe we could use the evaluator class? Is there any loss of functionality?

examples/microduck/ppo_mujoco.py
```diff
@@ -0,0 +1,1111 @@
...
+        critic.load_state_dict(best_state[1])
+    return history
+
+
+# ----------------------------------------------------------------------
+# Command line
+# ----------------------------------------------------------------------
+
+
+def parse_args(args: Sequence[str] | None = None) -> argparse.Namespace:
```
vmoens: what about using hydra?
We do it for all the sota-implementation scripts

examples/microduck/ppo_mujoco.py
```diff
@@ -0,0 +1,1111 @@
...
+    parser.add_argument("--wandb-name")
+    parser.add_argument(
+        "--wandb-mode", choices=("online", "offline", "disabled"), default="online"
+    )
+    parser.add_argument("--seed", type=int, default=0)
+    parser.add_argument("--smoke", action="store_true", help="Tiny pipeline check.")
+    return parser.parse_args(args)
+
+
+def main(args: argparse.Namespace) -> None:
```
vmoens: the arg parsing could be simplified if we were using hydra

examples/microduck/ppo_mujoco.py
```diff
@@ -0,0 +1,951 @@
...
+        "compile_step": cfg.env.compile_step,
+        "hidden_size": cfg.policy.hidden_size,
+        "max_episode_steps": cfg.env.max_episode_steps,
+    }
+    env = make_env(
+        cfg.env.microduck_root,
+        num_envs=cfg.env.num_envs,
+        parallel=cfg.env.parallel,
+        **env_kwargs,
+    )
```
vmoens: looks like make_env could just take `cfg` no? or `cfg.env`
I don't like this parsing back and forth to dicts, it's ugly and brittle.

examples/microduck/ppo_mujoco.py
```diff
@@ -0,0 +1,975 @@
...
+                            "kl_approx",
+                            "clip_fraction",
+                            "ESS",
+                        ):
+                            sums[f"ppo/{key}"] = sums.get(f"ppo/{key}", 0.0) + float(
+                                losses[key].detach()
+                            )
+                        sums["ppo/grad_norm"] = sums.get("ppo/grad_norm", 0.0) + float(
+                            grad_norm
+                        )
```
vmoens: use tensordict primitives for this

PR #4213, opened 2026-09-04
[Feature] PPO-EWMA: decouple the proximal policy from the behavior policy

vmoens:
## Description

Implements the decoupled PPO objectives of [Batch size-invariance for policy optimization](https://arxiv.org/abs/2110.00641) (Hilton, Cobbe, Schulman, 2021; reference code: [openai/ppo-ewma](https://github.com/openai/ppo-ewma)).

PPO uses the policy that collected the data for two distinct purposes: as the denominator of the importance-sampling ratio, and as the reference the trust region (clipping or KL penalty) pulls the updated policy towards. Only the first role requires the *behavior* policy; the second can use any recent *proximal* policy. Decoupling them orthogonalizes inference (data collection) from training: the strength of the trust region no longer depends on how, or how recently, the data was collected.

### What changes

- `PPOLoss` / `ClipPPOLoss` / `KLPENPPOLoss` gain `delay_actor: bool = False`. When set, a detached copy of the actor parameters is created under `target_actor_network_params` (the standard TorchRL target-network machinery, `create_target_params=True`) and used as the proximal policy:
  - `ClipPPOLoss`: `loss = -(pi_prox / pi_behav) * min(r * A, clip(r, 1 - eps, 1 + eps) * A)` with `r = pi_theta / pi_prox`, i.e. the decoupled clipped objective (Section 2 of the paper).
  - `KLPENPPOLoss`: the surrogate keeps the behavior ratio `pi_theta / pi_behav` while the KL penalty is measured against the proximal policy.
  - Any `TargetNetUpdater` controls the age of the proximal policy. A `SoftUpdate(loss, eps=beta_prox)` stepped after every optimizer step gives the EWMA of the paper (PPO-EWMA); a `HardUpdate` freezes it for a fixed number of steps.
- New `max_importance_ratio: float | None = None` caps the behavior ratio `pi_theta / pi_behav`. As in the reference implementation the cap lifts the detached behavior log-prob rather than clamping the ratio, so capped samples keep a rescaled policy gradient instead of a zero one. Available for all three losses (useful on its own with stale/async data).
- Diagnostics: with a decoupled proximal policy, `clip_fraction` and `kl_approx` describe `r` (the ratio the trust region acts on) while `ESS`, `max_ratio` and `mean_ratio` describe the behavior ratio the data is actually weighted by. With the defaults everything is bit-identical to before.
- `_log_weight` now returns a fourth element, the (gradient-free) log importance correction, `None` in the default configuration.
- Docs: kwarg docstrings for the three losses, a PPO-EWMA note with a usage snippet on `ClipPPOLoss`, and a "Decoupled proximal policy (PPO-EWMA)" section in `objectives_policy.rst`.

### Usage

```python
loss_module = ClipPPOLoss(actor, critic, delay_actor=True, max_importance_ratio=100.0)
updater = SoftUpdate(loss_module, eps=0.889)  # eps is the EWMA decay rate beta_prox
for batch in replay_buffer:
    losses = loss_module(batch)
    (losses["loss_objective"] + losses["loss_critic"] + losses["loss_entropy"]).backward()
    optimizer.step()
    optimizer.zero_grad()
    updater.step()
```

`MAPPOLoss` / `IPPOLoss` inherit the feature through `ClipPPOLoss`.

### Tests

New tests in `test/objectives/test_ppo.py`:

- `test_ppo_decoupled_proximal_policy` (flat and composite action distributions): the decoupled loss reduces to the standard `ClipPPOLoss` when the proximal policy equals the behavior policy; matches the hand-computed decoupled objective, clip fraction, `kl_approx` and ratio statistics when it does not; and follows the proximal policy after a `SoftUpdate.step()`.
- `test_ppo_max_importance_ratio`: exact values of both `min()` branches with and without the cap, the capped sample keeps a non-zero policy gradient, and the cap acts on the total ratio with a decoupled proximal policy.
- `test_ppo_kl_pen_decoupled_proximal_policy`: the KL is measured against the target-parameter policy (the behavior distribution parameters stored in the data are ignored) while the surrogate keeps the behavior ratio.
- `test_ppo_delay_actor_requires_functional`.

The full `test/objectives/test_ppo.py` and `test/objectives/test_mappo.py` pass locally.

### Follow-up

A stacked PR adds the Hydra config fields (`PPOLossConfig.delay_actor` / `max_importance_ratio`, `PPOTrainerConfig.target_net_updater`) and the `OnPolicyTrainer` wiring so PPO-EWMA can be run from the trainer.

### Update: CI hygiene commits

Two small commits ride along so the stack gets a green CI: the `ufmt` reformat of `sota-implementations/dreamer_v3/train.py` (the Lint workflow is red on `main` for this file) and a `sqlalchemy<2.1` pin in `.github/unittest/linux/scripts/run_all.sh` (uv resolved the pre-release `2.1.0b3`, which breaks the MLflow logger fixtures in the `mp` partition). The remaining red jobs, `unittests-gym` (Atari ROM download 403) and a triton tolerance flake in `test_dreamer_components.py`, are environmental.

### DIFF
--- a/.github/unittest/linux/scripts/run_all.sh
+++ b/.github/unittest/linux/scripts/run_all.sh
@@ -132,6 +132,7 @@ uv_pip_install \
   "huggingface-hub>=1.5.0,<2.0" \
   wandb \
   mlflow \
+  "sqlalchemy<2.1" \
   av \
   coverage \
   transformers \
--- a/docs/source/reference/objectives_policy.rst
+++ b/docs/source/reference/objectives_policy.rst
@@ -14,3 +14,36 @@ Loss modules for policy gradient algorithms.
     KLPENPPOLoss
     A2CLoss
     ReinforceLoss
+
+Decoupled proximal policy (PPO-EWMA)
+------------------------------------
+
+PPO uses the policy that collected the data (the ``sample_log_prob`` entry)
+for two distinct purposes: as the denominator of the importance-sampling
+ratio, and as the reference the trust region (clipping or KL penalty) pulls
+the updated policy towards. `Hilton et al. (2021)
+<https://arxiv.org/abs/2110.00641>`_ observe that only the first role requires
+the *behavior* policy; the second can use any recent *proximal* policy.
+
+Passing ``delay_actor=True`` to :class:`ClipPPOLoss` or :class:`KLPENPPOLoss`
+keeps a detached copy of the actor parameters under
+``target_actor_network_params`` and uses it as the proximal policy: the ratio
+being clipped (or the KL reference) becomes ``pi_theta / pi_prox`` and the
+surrogate is re-weighted by ``pi_prox / pi_behav`` so that the gradient
+estimate stays unbiased for the data at hand. Stepping a
+:class:`~torchrl.objectives.SoftUpdate` after every optimizer step turns the
+proximal policy into an exponentially-weighted moving average of the policy,
+which is PPO-EWMA; a :class:`~torchrl.objectives.HardUpdate` freezes it for a
+fixed number of steps instead. ``max_importance_ratio`` caps the behavior
+ratio for numerical stability with stale data. This decouples the strength of
+the trust region from how, and how recently, the data was collected.
+
+    >>> loss_module = ClipPPOLoss(actor, critic, delay_actor=True, max_importance_ratio=100.0)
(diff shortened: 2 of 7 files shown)


coder-jayp (approved):
I tested the code locally and all tests passed successfully. LGTM!

PR #4214, opened 2026-09-04
[Feature] PPO-EWMA: trainer wiring and Hydra configs

vmoens:
## Description

Stacked on #4213 (base branch `ppo-ewma`): merge that one first, then retarget this PR to `main`.

Wires the decoupled proximal policy / PPO-EWMA feature of #4213 into the trainer and the Hydra configs so it can be run from `PPOTrainer`.

### What changes

- `OnPolicyTrainer` (and thus `PPOTrainer`, `A2CTrainer`, `ReinforceTrainer`) accepts `target_net_updater: TargetNetUpdater | None`. It is handed to the base `Trainer` (so it is checkpointed like the other trainers' updaters) and registered as a `TargetNetUpdaterHook` at the `post_optim` stage, i.e. stepped after every optimizer step, which is what the PPO-EWMA proximal policy requires (`post_steps` would only step it once per collected batch).
- `PPOLossConfig` gains `delay_actor: bool = False` and `max_importance_ratio: float | None = None`, forwarded to all three loss flavors (matching the loss signatures).
- `OnPolicyTrainerConfig` gains `target_net_updater: Any = None`; `_make_onpolicy_trainer` instantiates a partial config (e.g. `SoftUpdateConfig`) with the loss module, as the off-policy trainer factories do.
- `sota-implementations/ppo_trainer/config/config_ewma.yaml`: composed example (`defaults: - config`) enabling PPO-EWMA on the Pendulum PPO trainer recipe: `loss.delay_actor=true`, `loss.max_importance_ratio=100.0`, `target_net_updater@target_net_updater: soft` with `eps: 0.889` (the paper's default `beta_prox`). Run with `python train.py --config-name config_ewma`.
- The `ppo_trainer` base `config.yaml` carried two wandb-only keys (`offline`, `project`) in its csv logger section, which made it fail to compose under Hydra's struct mode (`Key 'offline' not in 'CSVLoggerConfig'`). They are dropped so the base config, and therefore the example above, composes. `a2c_trainer` and `reinforce_trainer` have the same pre-existing issue and are left for a separate fix.

### Tests

- `test/test_trainer.py::TestOnPolicyTargetNetUpdater`: a real `PPOTrainer` with a `ClipPPOLoss(delay_actor=True)` steps a `HardUpdate` exactly once per optimizer step (counter equals the number of optimizer steps, proximal policy untouched below the interval), and a `SoftUpdate` moves the proximal policy without copying the policy.
- `test/test_configs.py::test_ppo_loss_config_delay_actor` (clip and kl): the new fields reach the instantiated loss.
- `test/test_configs.py::test_ppo_ewma_trainer_parsing_with_file`: a yaml with `target_net_updater@target_net_updater: soft` and `loss.delay_actor: true` produces a `PPOTrainer` holding the `SoftUpdate` and a `post_optim` `TargetNetUpdaterHook`.
- `config_ewma.yaml` composes and its loss and updater instantiate locally.

### Update: the example config now runs end to end

`OnPolicyTrainer` requires `weight_update_map` when the collector ships weight-sync schemes, which the `multi_async` collector of `config.yaml` now does by default, so `train.py` could not start (`ValueError: Collector has weight_sync_schemes configured, but weight_update_map was not provided`). `config.yaml` now maps the collector's `policy` destination to `loss_module.actor_network`, and `OnPolicyTrainerConfig.weight_update_map` is typed `Any` because OmegaConf cannot merge a mapping into a `dict[str, str] | None` field defaulting to `None`, neither from yaml nor from the CLI. `python train.py --config-name config_ewma` runs end to end.

Local sanity check on a MacBook (Pendulum-v1, 1M frames, seed 42, mean per-step reward over the last 100k frames; a random policy scores about -6): `config.yaml` (PPO) reaches -4.62 with the mean log-ratio blowing up to 130 and a clip fraction of 0.40 late in training; `config_ewma.yaml` with the paper's default `eps=0.889` reaches -5.03 with stable diagnostics; with `eps=0.941`, i.e. the paper's rule `1/(1-eps) - 1 = optimizer steps per collected batch` (16 here), PPO-EWMA reaches -3.62 with a clip fraction of 0.25 and a mean log-ratio of 1.0. Single seed, so indicative only. Note that the example's `tanh_normal` model config exposes no action bounds, so actions are limited to [-1, 1] while Pendulum accepts [-2, 2], which makes the task harder than the standard benchmark for every variant.

### Update: tuned, runnable example

Three further commits make the Pendulum example learn properly: `TanhNormalModelConfig` gained `low`/`high` (the example policy was capped at torque +-1 while Pendulum accepts +-2) and `tanh_loc`; the example now uses a std floor of 0.1 and `tanh_loc` (without them the std collapses, the tanh saturates and the log-probabilities of actions at the bound overflow to NaN after 300k to 550k frames), lr 3e-4, 10 epochs, minibatches of 64, gradient clip 1 and entropy coefficient 1e-4; `config_ewma.yaml` sets `eps=0.99`.

Local results (1M frames, single seed, mean per-step reward over the last 100k frames; random is about -6, solved is about -1): PPO -1.64 (first reaches -1 at 447k frames), PPO-EWMA `eps=0.99` -1.32 (reaches -3 per step at 319k frames vs 342k for PPO), `eps=0.995` -2.41, `eps=0.97` -3.22, `eps=0.941` -4.15. Decays whose center of mass is of the order of the optimizer steps per collected batch work best; short decays lag. W&B project `vmoens/torchrl-ppo-ewma`, runs `e_ppo`, `f_ewma_e99`, `f_ewma_e995`, `e_ewma_e97`.

### DIFF
--- a/sota-implementations/ppo_trainer/config/config.yaml
+++ b/sota-implementations/ppo_trainer/config/config.yaml
@@ -53,6 +53,12 @@ models:
     # Pendulum-v1 accepts torques in [-2, 2]; TanhNormal defaults to [-1, 1]
     low: -2.0
     high: 2.0
+    # floor on the policy std: without it the std collapses after ~300k
+    # frames and the log-prob ratios overflow to NaN
+    scale_lb: 0.1
+    # bound the location too: an unbounded loc saturates the tanh and the
+    # log-probability of actions at the bound overflows
+    tanh_loc: true
 
   value_model:
     in_keys: ["observation"]
@@ -88,11 +94,11 @@ training_env:
 loss:
   actor_network: ${models.policy_model}
   critic_network: ${models.value_model}
-  entropy_coeff: 0.01
+  entropy_coeff: 1.0e-4
 
 # Optimizer configuration
 optimizer:
-  lr: 0.001
+  lr: 3.0e-4
 
 # Collector configuration
 collector:
@@ -114,7 +120,7 @@ replay_buffer:
     shuffle: true
   writer:
     compilable: false
-  batch_size: 128
+  batch_size: 64
 
 logger:
   exp_name: ppo_pendulum_v1
@@ -129,14 +135,14 @@ trainer:
   total_frames: 1_000_000
   frame_skip: 1
   clip_grad_norm: true
-  clip_norm: 100.0
+  clip_norm: 1.0
   progress_bar: true
   seed: 42
   save_trainer_interval: 100
   log_interval: 100
   save_trainer_file: null
   optim_steps_per_batch: null
-  num_epochs: 2
+  num_epochs: 10
   async_collection: false
   # the multi-process collector syncs weights through weight-sync schemes;
   # map its "policy" destination to the trainer's actor
--- a/sota-implementations/ppo_trainer/config/config.yaml
+++ b/sota-implementations/ppo_trainer/config/config.yaml
@@ -50,6 +50,9 @@ models:
     param_keys: ["loc", "scale"]
     out_keys: ["action"]
     network: ${networks.policy_network}
+    # Pendulum-v1 accepts torques in [-2, 2]; TanhNormal defaults to [-1, 1]
+    low: -2.0
+    high: 2.0
 
   value_model:
     in_keys: ["observation"]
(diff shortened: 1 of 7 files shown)


coder-jayp (approved):
The trainer integration looks great. I tested the config and trainer logic locally. All tests passed successfully. LGTM!

PR #4216, opened 2026-09-04
[Performance] Accelerate DreamerV3 training with CUDA graphs

vmoens:
## Summary

- add an opt-in CUDA-graph learner path to the DreamerV3 DMC reproduction
- make fixed-horizon imagination and replay lambda returns CUDA-graph compatible
- cover compiled RSSM scan unroll 1 and CUDA graph capture with regression tests
- add reproducible steady-state RSSM and complete-learner benchmarks that exclude compilation and capture warmup
- compare two complete uncaptured and CUDA-graph learner updates across losses, parameters, optimizer state, and slow targets

The `--fast` reproduction mode now selects the compiled RSSM scan with unroll 8 and captures the fixed-shape learner forward/backward after five warmup calls. Optimizer and target-network updates remain outside capture so their Python-side schedules continue to advance.

## Why

Profiling the compiled scan path on one NVIDIA GB200 showed about 7,500 CUDA launches per learner update, despite only about 14.7 ms of self CUDA kernel time. The dominant gap was CPU dispatch and scheduling overhead rather than arithmetic throughput.

A same-hardware, steady-state full-learner comparison at batch 16 x sequence 64, BF16 compute and FP32 parameters gave:

| implementation | median ms/update | transitions/s | relative to JAX |
| --- | ---: | ---: | ---: |
| upstream DreamerV3 JAX | 22.04 | 46,468 | 1.00x |
| TorchRL compiled scan | 358.51 | 2,856 | 16.3x slower |
| TorchRL compiled scan + CUDA graph | 17.83 | 57,415 | 1.24x faster |

The checked-in complete-learner benchmark reproduces the TorchRL rows with 10 warmup updates and 50 measured updates. It includes model, actor, value, and replay-value losses, imagination, backward, the optimizer step, and the slow-target update. It excludes compile/capture latency, replay sampling, and environment collection.

The JAX baseline is pinned to `danijar/dreamerv3@e3f02248693a79dc8b0ebd62c93683888ddaccfe`. A JAX compatibility shim only adapts the older positional `jax.jit` call convention to JAX 0.11 for GB200 support.

Triton is therefore not required for parity in this configuration. It remains a possible follow-up for residual kernel-time improvements.

## Test plan

- `test/modules/test_dreamer_components.py -k rssm_rollout_compile`: 3 passed
- `test/objectives/test_dreamer_v3.py -m "not gpu"`: 74 passed
- existing GB200 GPU capture regressions: 2 passed
- GB200 complete-learner equivalence regression: 1 passed
- DreamerV3 SOTA smoke command: passed
- `benchmarks/ad_hoc/bench_dreamer_v3_learner.py` on GB200:
  - compiled scan: 358.51 ms/update
  - compiled scan + CUDA graph: 17.83 ms/update
- `benchmarks/ad_hoc/bench_dreamer_v3_rssm.py` on GB200:
  - compiled scan: 218.04 ms/update
  - CUDA graph: 10.40 ms/update

### DIFF
--- a/.github/unittest/helpers/coverage_run_parallel.py
+++ b/.github/unittest/helpers/coverage_run_parallel.py
@@ -28,6 +28,11 @@ def write_config(config_path: Path, argv: list[str]) -> None:
     """
     assert not config_path.exists(), "Temporary coverage config exists already"
     cmdline = shlex.join(argv[1:])
+    thread_concurrency = (
+        "    thread\n"
+        if os.environ.get("TORCHRL_COV_TRACE_THREADS", "1") != "0"
+        else ""
+    )
     with open(str(config_path), "w", encoding="utf-8") as fh:
         fh.write(
             f"""# .coveragerc to control coverage.py
@@ -35,8 +40,7 @@ def write_config(config_path: Path, argv: list[str]) -> None:
 parallel=True
 concurrency=
     multiprocessing
-    thread
-command_line={cmdline}
+{thread_concurrency}command_line={cmdline}
 """
         )
 
--- a/.github/unittest/linux/scripts/run_all.sh
+++ b/.github/unittest/linux/scripts/run_all.sh
@@ -500,7 +500,10 @@ run_non_distributed_tests() {
       ;;
     collectors)
       echo "Running shard collectors: ${collector_tests} (serial)"
-      python .github/unittest/helpers/coverage_run_parallel.py -m pytest ${collector_tests} \
+      # Coverage's thread hook can crash multiprocessing.Queue feeder threads
+      # on Python 3.10. Process coverage remains enabled for this shard.
+      TORCHRL_COV_TRACE_THREADS=0 \
+        python .github/unittest/helpers/coverage_run_parallel.py -m pytest ${collector_tests} \
         "${GPU_MARKER_FILTER[@]}" \
         ${json_report_args} \
         ${common_args} ${serial_timeout}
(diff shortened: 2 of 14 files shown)


vmoens:
@unography, would you mind reviewing this follow-up to your timing question on #4196? I would especially value your take on the same-hardware benchmark setup and the CUDA-graph fast path.

unography:
thanks, will take a look at it
some initial suggestions though - 
1. would be good to benchmark with the complete learner updates as well, either as part of this PR or separately: so we measure  optimizer, slow-target updates, imagination, all loses, etc - the whole update
2. same for the test case, worth comparing the full learner update: normal v/s CUDA-graph, maybe warmup + 2 learner updates, to compare losses, params, optimizer state, etc

vmoens:
Addressed in 91671591.

The benchmark now exercises the exact production learner update, including model, actor, value, and replay-value losses, imagination, backward, the optimizer step, and the slow-target update. On one NVIDIA GB200 with PyTorch 2.12.0 / CUDA 13.0, BF16, batch 16 x sequence 64, scan unroll 8, 10 warmup updates, and 50 measured updates:

| learner path | median ms/update | transitions/s | speedup |
| --- | ---: | ---: | ---: |
| compiled scan | 358.51 | 2,856 | 1.00x |
| compiled scan + CUDA graph | 17.83 | 57,415 | 20.10x |

Compile and capture time, replay sampling, and environment collection are excluded. The benchmark is now reproducible with `python benchmarks/ad_hoc/bench_dreamer_v3_learner.py`.

I also added the complete-learner GPU regression. It finishes capture warmup, restores identical initial module state, runs two full updates through the uncaptured and CUDA-graph paths with aligned seeds, and compares all six reported losses, refreshed latent outputs, the full module state, and optimizer state. It additionally checks that both optimizer step counters are exactly two and that the slow target actually changed. This passes on the GB200 run.

PR #4220, opened 2026-09-05
[Doc] Clarify TensorDictPrimer is required for RNN collection

### LINKED ISSUE #2362: [BUG] Clarify non-opitionality of `TensordictPrimer`
## Without the primer, the collector does not feed any hidden state to the policy

in the [RNN tutorial ](https://github.com/pytorch/rl/blob/main/tutorials/sphinx-tutorials/dqn_with_rnn.py)it is stated that the primer is optional and it is used just to store the hidden states in the buffer.

This is not true in practice. Not adding the primer will result in the collector not feeding the hidden states to the policy during execution. Which will silently cause the rnn to loose any recurrency.

In the tutorial it seems like the only reason the Primer is there is to store hidden states in the buffer (which I would also highly advise against, as this makes any algorithm on-policy and can lead to severe issues).

I think it needs to be changed and remove any claims on optionality of the primer. Instead strong claims about its non-optionality should be made as, if a user removes it, the tutorial will silently run without recurrency.

### Reproduce

To reproduce, comment out this line

https://github.com/pytorch/rl/blob/0063741839a3e5e1a527947945494d54f91bc629/tutorials/sphinx-tutorials/dqn_with_rnn.py#L269

and print the policy input at this line

https://github.com/pytorch/rl/blob/0063741839a3e5e1a527947945494d54f91bc629/torchrl/collectors/collectors.py#L733

you will see that no hidden state is fed to the rnn during execution and no errors or warnings are thrown

### The future I want to live in

In the future I want to live in there are no primers. The torchrl components are able to look at the policy outputs and carry forward whatever is in "next". The primer for me is a unique pain point particular to torchrl, as users doing recurrency in other libs won't have any equivalent of this and will probably forget it  causing major silent bugs

YeonwooSung:
## Description

The recurrent DQN tutorial described :class:`~torchrl.envs.transforms.TensorDictPrimer` as optional / not required. That is incorrect for the collector path used in the tutorial: without a primer (or equivalent auto-registration), the collector does not feed hidden state to the policy. The LSTM zeros or drops recurrent state, recurrency silently dies, and training still runs but is worse.

This PR rewrites those tutorial comments so they match the actual contract:

- On this tutorial's collector path, a primer is required if hidden states should travel through ``env.reset`` → collector → policy → replay.
- Omitting it does not crash; it zeros/drops recurrent state.
- ``auto_register_policy_transforms=True`` is the collector-side alternative. This tutorial still builds the primer explicitly via ``lstm.make_tensordict_primer()``.

One-line cross-links were added from ``recurrent_state_lifecycle.rst`` and ``modules_rnn.rst``. Training logic is unchanged.

### Code example

Attach the LSTM primer before creating the collector so hidden state survives between policy calls. This Pendulum example requires Gymnasium.

```python
from tensordict.nn import TensorDictSequential
from torchrl.collectors import Collector
from torchrl.envs import Compose, GymEnv, InitTracker, TransformedEnv
from torchrl.modules import LSTMModule

lstm = LSTMModule(input_size=3, hidden_size=1, in_key="observation", out_key="action")
env = TransformedEnv(
    GymEnv("Pendulum-v1"), Compose(InitTracker(), lstm.make_tensordict_primer()),
)
collector = Collector(
    env, TensorDictSequential(lstm), frames_per_batch=8, total_frames=8,
)
batch = next(iter(collector))
assert "recurrent_state_h" in batch.keys()
assert ("next", "recurrent_state_h") in batch.keys(True, True)
collector.shutdown()
# Alternatively, let Collector register policy transforms with
# auto_register_policy_transforms=True; keep InitTracker on the environment.
```

## Motivation and Context

Without this clarification, readers following the tutorial can drop the primer, get no error, and train a non-recurrent policy.

close #2362

## Types of changes

## Checklist

### DIFF
--- a/docs/source/reference/modules_rnn.rst
+++ b/docs/source/reference/modules_rnn.rst
@@ -204,6 +204,8 @@ See also
 
 * :ref:`data-layout` for the contiguous trajectory layout and replay-buffer
   handoff.
+* :ref:`Recurrent state lifecycle <ref_recurrent_state_lifecycle>` for
+  the primer / ``auto_register_policy_transforms`` collection path.
 * :class:`LSTMModule` and :class:`GRUModule` for constructor arguments and
   examples.
 * :class:`set_recurrent_mode` for switching between single-step and recurrent
--- a/docs/source/reference/recurrent_state_lifecycle.rst
+++ b/docs/source/reference/recurrent_state_lifecycle.rst
@@ -365,6 +365,9 @@ See also
   a runnable tutorial that walks through this lifecycle end to end:
   collection, :class:`~torchrl.data.replay_buffers.SliceSampler` replay, and
   a full-sequence ``set_recurrent_mode(True)`` forward.
+- :ref:`Recurrent DQN <RNN_tuto>` — builds the primer explicitly with
+  :meth:`~torchrl.modules.LSTMModule.make_tensordict_primer` (the
+  collector-side alternative is ``auto_register_policy_transforms=True``).
 - ``examples/replay-buffers/recurrent_slice_sampler_pipeline.py`` — a
   minimal, runnable script version (GRU policy, the collector writing
   directly into the buffer, ``SliceSampler`` auto-detecting the trajectory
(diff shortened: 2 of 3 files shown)


vmoens (approved):
LGTM

PR #4229, opened 2026-09-05
[Doc] Document CatFrames with collectors and image replay buffers

### LINKED ISSUE #2618: [Feature Request] Provide documentation on how to use CatFrames with a data collector and replay buffer for images
## Motivation
Using CatFrames for inference is fairly straightforward and is already well documented.  
That being said, using CatFrames to reconstruct a stack of frames when sampling from the replay buffer is not so straightforward I find (subjective) and is not explicitly documentd for images (objective).
Using frame stacking for Visual RL is very common practice so I feel like the community would benefit from getting a better documentation on how to use CatFrames **for images**.  

## Solution
Provide a clear documentation that explains everything in details (no magic flags/magic values) on how to use CatFrames with a data collector and replay buffer (both extend and sample() method should be shown) for images.

I have created a gist of my attempt to use CatFrames for images and while the inference part works, the stack frames retrieved from the replay buffer do not make sense.  

https://gist.github.com/AlexandreBrown/fe378f26a87bdc40c5995dcc7d42f482  

Any help on how to make the last part where we sample from the replay buffer return the correct CatFrames data is appreciated.  

## Contributions  
I am willing to work on the PR for the documentation update if someone can help me get the [MVP script](https://gist.github.com/AlexandreBrown/fe378f26a87bdc40c5995dcc7d42f482) working.

## Checklist

YeonwooSung:
## Description

Adds a focused Sphinx recipe for using `CatFrames` with a data collector and a replay buffer on **images**, which was missing from the reference docs.

Two legitimate placements are documented and cross-linked:

- **Env-side** (`docs/source/reference/envs_transforms.rst`): stateful stacking for the policy, reset / `InitTracker` behavior, and `dim=-3` for CHW images.
- **Buffer-side** (`docs/source/reference/data_replaybuffers.rst`): store unstacked raw `pixels`, rebuild the stack in `rb.sample()`, and wire a `Collector` through `extend` / `sample`. Shows both `CatFrames.make_rb_transform_and_sampler` and the explicit `SliceSampler` form.

The recipe also covers why raw uint8 frames are what you store, how the collector's `("collector", "traj_ids")` interact with the slice sampler, and the common pitfalls of stacking twice (env **and** buffer on the same key) or using the vector default `dim=-1` on pixels.

Copy-paste snippets use `# doctest: +SKIP` where gym / pixel rendering is required. The `CatFrames` class docstring now points at both sections.

### Code example

Rebuild four-frame CHW stacks when sampling, while storing individual frames. Synthetic numbered grayscale pixels make sequence continuity visible without rendering dependencies.

```python
import torch
from torchrl.envs import CatFrames

pixels = torch.arange(100, dtype=torch.float32).reshape(100, 1, 1, 1)
data = TensorDict({
    "pixels": pixels,
    ("next", "pixels"): pixels + 1,
    ("collector", "traj_ids"): torch.zeros(100, dtype=torch.long),
    ("next", "done"): torch.zeros(100, 1, dtype=torch.bool),
    ("next", "terminated"): torch.zeros(100, 1, dtype=torch.bool),
}, batch_size=[100])
catframes = CatFrames(
    N=4, dim=-3, in_keys=["pixels"], out_keys=["pixels_stack"], reset_key="_reset",
)
transform, sampler = catframes.make_rb_transform_and_sampler(
    batch_size=8, traj_key=("collector", "traj_ids"),
)
rb = ReplayBuffer(
    storage=LazyTensorStorage(100), batch_size=8, sampler=sampler, transform=transform,
)
rb.extend(data)  # In training, extend with each collector batch.
batch = rb.sample()
assert batch["pixels_stack"].shape == (8, 4, 1, 1)
frames = batch["pixels_stack"][..., 0, 0]
assert (frames[:, 1:] - frames[:, :-1] == 1).all()
```

## Motivation and Context

Using `CatFrames` for inference was already documented; reconstructing a frame stack when sampling from a replay buffer was not, especially for images (stack dim `-3`, not the vector default). Visual RL relies on this pattern, and the collector + `extend` + `sample` path is easy to get wrong (double stacking, wrong dim, no time axis).

close #2618

## Types of changes

## Checklist

### DIFF
--- a/docs/source/reference/data_replaybuffers.rst
+++ b/docs/source/reference/data_replaybuffers.rst
@@ -506,6 +506,9 @@ trajectory-edge padding, while ``rebin``'s stack keeps a separate frame axis::
     ...     ),
     ... )  # doctest: +SKIP
 
+The same ``dim=-3`` convention, and the collector / ``extend`` / ``sample``
+wiring, is spelled out in :ref:`catframes-collector-replay`.
+
 **Multiple files.** A clip is often split across many small files (one per episode)
 rather than one large mp4. :meth:`VideoClipRef.from_files` addresses a list of files
 as a single logical sequence, so slicing, :meth:`rebin` and decoding work across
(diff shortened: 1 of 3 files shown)


docs/source/reference/data_replaybuffers.rst
```diff
@@ -529,3 +532,228 @@ When camera and control loops run at different rates, prefer
...
+    from torchrl.data import SliceSampler
+    from torchrl.envs import ExcludeTransform
+
+    rb = ReplayBuffer(
+        storage=LazyTensorStorage(100_000),
+        sampler=SliceSampler(
+            slice_len=frame_stack,
+            traj_key=("collector", "traj_ids"),
+        ),
+        batch_size=batch_size * frame_stack,  # B windows of length N
```
vmoens, line 703: [P2] Reshape sampled windows before CatFrames and retain their last step

SliceSampler returns the B windows flattened as [B*N]; it does not make a [B, N] TensorDict for this transform. Consequently CatFrames treats the entire sample as one time sequence, so the first N-1 rows of later windows incorporate frames from an unrelated preceding window. A synthetic increasing-frame replay reproduces stacks such as [52, 53, 54, 66]; 12 of 16 rows were wrong with N=4. Match the helper by reshaping to [-1, N] before CatFrames and selecting [:, -1] afterward, or remove this purported equivalent recipe.

vmoens (changes requested):
The helper-based collector/replay example runs and returns [32, 4, 84, 84] stacks. The explicit alternative is missing the reshape/window-selection steps and mixes unrelated samples.

docs/source/reference/data_replaybuffers.rst
YeonwooSung: Fixed in f5ef733132e02f854a55d2aa7997d82a8030e068. The explicit recipe now matches `make_rb_transform_and_sampler`: `reshape(-1, N)` before `CatFrames` and `[:, -1]` after it. A synthetic increasing-frame buffer reproduced the bleed (`[54, 55, 56, 61]` and only 16/64 consecutive rows with `N=4`); after the reshape/last-step pair every window is consecutive and the sample shape is `[B, N]` like the helper. The surrounding prose and the pitfalls list call out that `SliceSampler` returns `[B * N]`, not `[B, N]`.

YeonwooSung:
Addressed the CHANGES_REQUESTED review in f5ef733132e02f854a55d2aa7997d82a8030e068.

The helper-based collector/replay example was already correct. The explicit `SliceSampler` alternative was missing the reshape / last-step pair, so `CatFrames` treated the flattened `[B * N]` sample as one time sequence and mixed frames across windows. That recipe now does `reshape(-1, N)` then `[:, -1]`, matching `make_rb_transform_and_sampler`.

vmoens (approved):
Re-reviewed f5ef733. The manual replay recipe now restores the [batch, time] slice layout before CatFrames and selects the last item of each slice afterward. Sampling batch_size * frame_stack raw transitions matches the helper recipe.

I ran both rendered CartPole replay examples with a smaller collection/storage size; both return 32 samples with pixels_trsf shape [32, 4, 84, 84]. No remaining actionables.

Issue #3032, opened 2025-07-03
[Feature Request] Documentation about weight update propagation


valterschutz:
First of all, thank you so much to everyone working on this project. I'm currently using it in several projects and it's (mostly) been a very pleasant experience.

But there is one thing in torchrl that I find particularly confusing.

## Motivation

I'm confused about whether weight updates in `LossModule`s get propagated back to the original module that was passed in to the `LossModule`. Let's take `DQNLoss` as an example and suppose we have a `TensorDictModule` called `module` that we pass in as the `value_network` argument. As I understand it, the loss module has its own weight copies of the network that was passed in as `value_network`. So when we call `loss_module.backward()`, the weights of the original network are **not** updated.

When we give the original module (or rather an epsilon-greedy version of it) to a `SyncDataCollector`, we have to call `collector.update_policy_weights_()` in order for the loss module weights to get copied to the original module. But what if we want to do inference without a collector?

I've seen snippets like
```python
with self.loss_module.qvalue_network_params["module"]["0"].to_module(self.module):
  # self.module can now be used with updated weights
```
but I'm not sure if that's the proper way to do it...

I've tried looking into the source code for `SyncDataCollector` and `VanillaWeightUpdater` but I think that this would be a very useful piece of documentation to have.

## Solution

A documentation page that briefly mentions
- whether weight updates from the loss module are propagated back to the original module
- an example of how to use a policy for inference (without using a collector)

## Alternatives

Maybe I'm not understanding something that is obvious to everyone else. In that case, please point me to a source that explains these things.

## Checklist

- [X] I have checked that there is no similar issue in the repo (**required**)

valterschutz:
After some experimenting, it does seem like the weight updates get propagated back to the original module. But I'm still uncertain about *how* and *when* this happens.

vmoens:
Hey @valterschutz 
Yes once you pass a module to a LossModule the weights are updated in-place. The module itself is used as-is - there is no copy happening.

If that isn't clear, where should we document this?

If you pass things like `device` args to the data collectors they will change the storage location though!

Does this help?

valterschutz:
Alright, that is nice to hear. Although this was also my first guess of what should be happening when I started using torchrl, I've had cases when the original module was *not* updated, although the loss module parameters were. But I guess there must have been some other issue in that case. Maybe it had something to do with collectors, although I tend to avoid using the `device` argument there and instead ensuring that everything is on the same device to begin with.

[...]

PR #4230, opened 2026-09-05
[Doc] Document checkpointing to resume RL training

### LINKED ISSUE #3044: [Feature Request] Documentation explaining the recommended way to save and load agents
## Motivation

I would like a concrete documented example of how to save RL agents after they have been trained, and also how to load them afterwards to resume training. In my current projects, I always switch between different methods and I have not yet found an approach that feels "right". For example,
- Should only the `nn.Module`s be saved or the `TensorDictModule`s?
- Should the loss module be saved?
- Saving the state_dict is usually recommended for `nn.Module`s. Is it the same way here?

I know that https://docs.pytorch.org/rl/stable/tutorials/export.html already explains how to export only the policy for inference, but I'm looking for something more general which would allow to resume training of the agent.

 #2620 and #1089 are related.

## Solution

I'm proposing that one of the following pages gets added to the documentation:
1. A separate page that describes how to save a DQN agent (or some other off-policy method). Network weights, optimizer state  *and* the replay buffer should be saved so that training can resume.
2. In the PPO tutorial (https://docs.pytorch.org/rl/stable/tutorials/coding_ppo.html#), add a section at the end that explains how to save the trained agent. The drawback with this approach is that it doesn't show how to save a replay buffer.

## Alternatives

I don't think there is. Saving models after they have been trained is an essential part of the RL workflow and the recommended way to do this should definitely be documented somewhere.

## Checklist

YeonwooSung:
## Description

Adds a **Resume training** guide to `docs/source/reference/checkpoint.rst` so users can save and restore an RL agent with the existing `:class:`~torchrl.checkpoint.Checkpoint`` API.

The new section:

- Lists the components a resume checkpoint should contain (training model, optimizer, target networks; collector policy / env-transform state / frame count; replay-buffer writer, sampler, transforms, and storage; RNG) and why each one matters in RL.
- Shows constructing `Checkpoint(policy=..., optimizer=..., replay_buffer=..., collector=..., loss_module=..., exploration=..., rng=GlobalRNGState())`, then `save` / `load` with `components={...}` and `map_location`.
- Clarifies that the payload is the `TensorDictModule` / `state_dict` (or the buffer's `dump` / `load`), not a pickled mystery container.
- Cross-links the shared-module case from #3032: if the loss holds the same module, save the policy or the loss once; do not persist two independent copies of the online weights.
- Contrasts **checkpoint** (interrupt and resume) with **export** (deploy for inference) and links the export tutorial.

Includes a compact CartPole DQN snippet a reader can copy.

### Code example

Minimal model/optimizer checkpoint round trip. The full RL recipe also registers the loss (including targets), replay buffer, collector, exploration state, and target updater.

```python
from tempfile import TemporaryDirectory
import torch
from torch import nn
from torchrl.checkpoint import Checkpoint, GlobalRNGState

policy = nn.Linear(4, 2)
optimizer = torch.optim.Adam(policy.parameters())
policy(torch.ones(1, 4)).square().sum().backward()
optimizer.step()  # Populate optimizer state before saving.
with TemporaryDirectory() as directory:
    path = f"{directory}/checkpoint"
    Checkpoint(policy=policy, optimizer=optimizer, rng=GlobalRNGState()).save(path)
    restored_policy = nn.Linear(4, 2)
    restored_optimizer = torch.optim.Adam(restored_policy.parameters())
    restored = Checkpoint(
        policy=restored_policy, optimizer=restored_optimizer, rng=GlobalRNGState(),
    )
    restored.load(path, components={"policy", "optimizer", "rng"}, map_location="cpu")
    torch.testing.assert_close(restored_policy.weight, policy.weight)
    assert restored_optimizer.state_dict()["state"]
    restored_optimizer.zero_grad()
    restored_policy(torch.ones(1, 4)).square().sum().backward()
    restored_optimizer.step()  # Continue training with restored optimizer state.
```

## Motivation and Context

close #3044

The export tutorial is inference-only. Users asked for a copy-paste guide to interrupt training and continue it later. `Checkpoint` already exists; this PR documents how to use it rather than inventing a parallel API.

## Types of changes

## Checklist

### DIFF
--- a/docs/source/reference/checkpoint.rst
+++ b/docs/source/reference/checkpoint.rst
@@ -7,6 +7,11 @@ TorchRL checkpoints use one manifest-driven format for standalone scripts,
 trainers, and policy-only consumers. Components are registered independently,
 so a checkpoint may contain only a policy or a complete training state.
 
+To interrupt a run and continue it later, see :ref:`checkpoint_resume`. To
+ship a trained policy for inference, see the :ref:`export tutorial <export_tuto>`.
+Those are different jobs: a resume checkpoint keeps optimizer, collector,
+replay-buffer, and target-network state; an export drops them on purpose.
+
 The directory and archive containers share the same logical layout. Directory
 checkpoints are the default and are best suited to large replay buffers;
 archives are convenient single-file artifacts. Loading either container is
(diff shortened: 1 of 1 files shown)


vmoens (approved):
Executed the new DQN example through save, object reconstruction, load, backward, optimizer.step and target_updater.step. Reviewed the component ownership and checkpoint/export guidance. No actionable findings.

YeonwooSung:
Rebased onto current `main` so this resume guide sits next to the stop/resume work from #4380–#4385.

- Scripts that assemble components themselves still use `Checkpoint` as documented here. Recipe users are pointed at `resume=` / `RunCheckpointer`.
- The CartPole DQN snippet now attaches `InitTracker` so it does not warn under the current collector policy-transform check.
- Re-ran save, object reconstruction, load, backward, `optimizer.step`, and `target_updater.step` after the rebase.

The previous collectors / gym / bulk CI failures were on a stale merge base and are outside this docs-only diff.

docs/source/reference/checkpoint.rst
```diff
@@ -48,6 +53,339 @@ Set ``save_components={"policy", "optimizer", "trainer_state"}`` on a
...
+
+**Collector**
+
+* **Inference / exploration policy.** The module the collector actually
+  executes, including exploration state such as the ``eps`` buffer of
+  :class:`~torchrl.modules.EGreedyModule`. This is often a wrapper around
+  the training model rather than a second network.
+* **Environment and transform state.** Observation-norm statistics,
+  frame-stack buffers, step counters. A collector ``state_dict`` stores
+  the env/transform state of a :class:`~torchrl.envs.TransformedEnv`.
```
vmoens: [P2] Qualify which environment/transform state is restored

This explicitly lists step counters as checkpointed state, but for a `TransformedEnv` the collector saves only `env.transform.state_dict()`. `StepCounter` keeps its episode count in the carrier TensorDict, which is not saved, and simulator state is not included either. I reproduced this with `CountingEnv` plus `StepCounter`: after saving five transitions, transform state is empty; loading preserves `frames=5`, but the first resumed transition starts at `step_count=0` and observation `0`. Please describe restoration of registered transform parameters/buffers and collector progress, and state that live episode/simulator continuation requires separately supported state handling. The nearby RNG wording should likewise distinguish process-global RNG restoration from environment/worker RNG state and exact trajectory reproducibility. This only needs a documentation qualification, not a simulator-checkpointing implementation.

vmoens (changes requested):
Rechecked the latest head `bdb2c6b06aaf4cce3a115f3e1994eccb87e84b23`. The replay-buffer initialization issue is fixed, and the current DQN example runs through save, restore, backward, optimizer update, and target update. One documentation correction remains: distinguish restored training progress/registered transform state from continuation of a live environment episode.

docs/source/reference/checkpoint.rst
YeonwooSung: Qualified the collector bullets: a `TransformedEnv` collector saves `env.transform.state_dict()` (registered transform parameters/buffers) plus `frames`/`iter`. `StepCounter` lives in the carrier TensorDict and is not restored; neither is the base simulator. Reproduced `CountingEnv` + `StepCounter`: after five transitions, transform state is empty, `frames=5` is restored, and the first resumed transition starts at `step_count=0` and observation `0`. `GlobalRNGState` is now described as process-global only, not env/worker RNG or exact trajectory reproducibility. Docs only; no simulator-checkpointing implementation.

YeonwooSung:
Addressed the remaining review note on `bdb2c6b06` and rebased onto current `main`.

The collector resume bullets now distinguish registered transform parameters/buffers and collector progress (`frames`/`iter`) from live episode/simulator continuation. `GlobalRNGState` is process-global only. Reproduced the `CountingEnv` + `StepCounter` case: transform state empty, `frames=5` restored, first resumed transition at `step_count=0` and observation `0`.

vmoens (commented):
Cool improvement. 

What do you think about the current API of checkpointing Trainer classes? Is it well integrated enough? Should we document it better or start by making the API sharper?

If you want to work on something (that requires a little bit of thought!) using rlrender with a checkpoint would be a cool feature (ie, call `rlrender --checkpoint checkpoint [other args to build the env]`)
Not sure how to bind the checkpoint and the network constructors but that'd be a fun thing to nail down.

vmoens (approved):

PR #4231, opened 2026-09-05
[Doc] Document LossModule and collector weight sharing

### LINKED ISSUE #3032: [Feature Request] Documentation about weight update propagation

YeonwooSung:
## Description

Document how `LossModule` and collectors share policy weights, and when `collector.update_policy_weights_()` is actually required.

The objectives landing page (and the DDPG tutorial) said losses "contain a copy of the trainable parameters". That wording made it look like an optimizer step on the loss would not update the original module, and that a collector always needed an explicit sync. Maintainer answers on #3032 already settled the semantics; this PR writes them down.

- Passing a module into a `LossModule` does **not** copy it. The same parameters are used in-place; optimizer steps on the loss update the original module.
- A CPU policy passed to a collector is moved to shared memory in-place. A CUDA policy is already shared.
- `update_policy_weights_()` is **required** when `policy_device` or `device` remaps the policy (a `.to(...)` copy is made). Calling it anyway is good practice but not load-bearing unless devices were remapped.

Cross-links: `LossModule` / objectives landing page ↔ collectors weight-sync page.

### Code example

An optimizer registered on the loss updates the original actor because their trainable parameters share storage.

```python
import torch
from torch import nn
from torchrl.modules import Actor, ValueOperator
from torchrl.objectives import DDPGLoss

actor = Actor(nn.Linear(3, 1))
critic = ValueOperator(nn.Linear(4, 1), in_keys=["observation", "action"])
loss = DDPGLoss(actor, critic, delay_actor=False, delay_value=False)
assert actor is loss.actor_network
optimizer = torch.optim.SGD(loss.parameters(), lr=0.1)
weight = next(actor.parameters())
before = weight.detach().clone()
weight.square().sum().backward()  # Small objective to demonstrate the storage link.
optimizer.step()
assert not torch.equal(weight.detach(), before)
# A collector using copied policy storage additionally needs update_policy_weights_().
```

## Motivation and Context

close #3032

## Types of changes

## Checklist

### DIFF
--- a/docs/source/reference/collectors_weightsync.rst
+++ b/docs/source/reference/collectors_weightsync.rst
@@ -3,6 +3,8 @@
 Weight Synchronization
 ======================
 
+.. _ref_collectors_weightsync:
+
 RL pipelines are typically split in two big computational buckets: training, and inference.
 While the inference pipeline sends data to the training one, the training pipeline needs to occasionally
 synchronize its weights with the inference one.
@@ -63,6 +65,28 @@ Each of these classes is detailed below.
     - Implement custom weight sync schemes for specialized use cases (e.g., new distributed backends, custom serialization)
     - Debug synchronization issues in complex distributed setups
     - Use weight sync schemes outside of collectors for custom multiprocessing scenarios
+
+When is ``update_policy_weights_`` required?
+--------------------------------------------
+
+:class:`~torchrl.objectives.LossModule` does not copy the policy (see
+:ref:`ref_lossmodule_weight_sharing`). Whether the collector needs an explicit
+sync after an optimizer step depends on how the policy was placed:
+
+- **CPU policy, no device remap.** Weights are moved to shared memory
+  **in-place** when the policy is passed to the collector. Workers see the
+  same storage; no extra sync is needed in the common case.
+- **CUDA policy, no device remap.** Parameters are already shared. No extra
+  sync is needed.
+- **``policy_device`` or ``device`` was passed.** The collector makes a
+  ``.to(...)`` copy. :meth:`~torchrl.collectors.Collector.update_policy_weights_`
+  is **required** after training so the collector's copy matches the trained
+  weights.
+
+Calling :meth:`~torchrl.collectors.Collector.update_policy_weights_` anyway
+is good practice, but it is not load-bearing unless devices were remapped.
+Use it also to push weights through a :class:`~torchrl.weight_update.WeightSyncScheme`
+(the rest of this page).
 
 Lifecycle of Weight Synchronization
 -----------------------------------
--- a/docs/source/reference/collectors_weightsync.rst
+++ b/docs/source/reference/collectors_weightsync.rst
@@ -71,22 +71,25 @@ When is ``update_policy_weights_`` required?
 
 :class:`~torchrl.objectives.LossModule` does not copy the policy (see
 :ref:`ref_lossmodule_weight_sharing`). Whether the collector needs an explicit
-sync after an optimizer step depends on how the policy was placed:
+sync after an optimizer step depends on whether its inference policy shares
+parameter storage with the training policy:
 
-- **CPU policy, no device remap.** Weights are moved to shared memory
-  **in-place** when the policy is passed to the collector. Workers see the
-  same storage; no extra sync is needed in the common case.
-- **CUDA policy, no device remap.** Parameters are already shared. No extra
-  sync is needed.
-- **``policy_device`` or ``device`` was passed.** The collector makes a
-  ``.to(...)`` copy. :meth:`~torchrl.collectors.Collector.update_policy_weights_`
-  is **required** after training so the collector's copy matches the trained
-  weights.
+- **Shared storage.** A policy passed directly can retain its parameter storage
+  when no device transfer or other copy is needed. No extra sync is required.
+- **Distinct storage.** Worker policies created by ``policy_factory``, copies
+  placed on another device, and remote policies have independent storage and
+  require a configured synchronization path.
+
+Passing ``policy_device`` or ``device`` is not by itself evidence of a copy. If
+the policy is already on the requested device, the collector can keep the same
+object and storage. If a move is needed, it creates a separate policy and
+:meth:`~torchrl.collectors.Collector.update_policy_weights_` must propagate the
+trained weights.
 
 Calling :meth:`~torchrl.collectors.Collector.update_policy_weights_` anyway
-is good practice, but it is not load-bearing unless devices were remapped.
-Use it also to push weights through a :class:`~torchrl.weight_update.WeightSyncScheme`
-(the rest of this page).
+is conservative good practice. Use it to push weights through a
+:class:`~torchrl.weight_update.WeightSyncScheme` or another configured transport
+whenever the policies do not share storage (the rest of this page).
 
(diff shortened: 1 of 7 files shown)


vmoens (approved):
Verified the parameter-sharing implementation and collector copy conditions, and executed the new weight-sharing example. No actionable findings.

YeonwooSung:
Rebased onto current `main`. The weight-sharing notes still match the LossModule and collector copy conditions after the collector progress / replay-write and trainer-resume work that landed since this PR.

Re-ran the DDPG identity / shared-storage / optimizer-step example after the rebase.

The previous collectors / gym / GPU CI failures were on a stale merge base and are outside this documentation diff.

docs/source/reference/collectors_weightsync.rst
```diff
@@ -3,6 +3,8 @@
 Weight Synchronization
 ======================
 
+.. _ref_collectors_weightsync:
```
vmoens: [P2] Place the reference target before the heading

This target follows the `Weight Synchronization` heading, so Sphinx has no title/caption for the new bare `:ref:` links elsewhere in the PR. A minimal Sphinx 8.1.3 build using this unchanged page opening emits `Failed to create a cross reference. A title or caption not found: 'ref_collectors_weightsync' [ref.ref]` and fails with `-W`. Moving the target immediately before the heading makes the same build pass.

docs/source/reference/objectives.rst
```diff
@@ -39,6 +40,55 @@ Quick Example
...
     total_loss = sum(v for k, v in loss_vals.items() if k.startswith("loss_"))
 
+.. _ref_lossmodule_weight_sharing:
+
+Weight sharing
+--------------
+
+Passing a module into a :class:`~torchrl.objectives.LossModule` does **not**
+copy it. The loss stores the same module and the same parameter tensors; an
+optimizer step on ``loss.parameters()`` updates the original network in-place.
```
vmoens: [P2] Exempt expanded ensembles from the parameter-sharing guarantee

This guarantee is broader than `LossModule` supports. `convert_to_functional(..., expand_dim=N)` clones/resamples independent ensemble parameters, as used by default `SACLoss` for its critics. In a live check, `loss.qvalue_network is critic` was true, but the original critic weight had shape `(1, 4)` while the optimized parameter had shape `(2, 1, 4)`; an optimizer step changed the ensemble and left the original critic unchanged. Calling the original critic can therefore use untrained parameters despite object identity. Please scope the guarantee to the non-expanded shared parameters demonstrated by the DDPG/actor example, and distinguish expanded/stacked ensembles and delayed targets. Apply the same qualification to the new wording in `LossModule` and `convert_to_functional` docstrings.

vmoens (changes requested):
Rechecked the latest head `95b21b4511faab4bed6b2fd2949b6334f305edba`. The earlier device-argument/storage-ownership issue is fixed, and the DDPG sharing example passes. Two remaining issues are reproduced: the new weight-sync reference has no section caption, and the general no-copy claim does not hold for expanded critic ensembles.

YeonwooSung:
Addressed the two review notes on `95b21b451` and rebased onto current `main`.

- `.. _ref_collectors_weightsync:` now sits immediately before the Weight Synchronization heading so the `:ref:` links have a title.
- The no-copy guarantee is limited to non-expanded parameters (DDPG actor). Expanded ensembles (`expand_dim`, default SAC critics) and delayed targets are called out in the Weight sharing section, the `LossModule` docstring, and `convert_to_functional`.

docs/source/reference/collectors_weightsync.rst
YeonwooSung: Moved `.. _ref_collectors_weightsync:` to immediately before the Weight Synchronization heading so the new `:ref:` links pick up that title.

docs/source/reference/objectives.rst
YeonwooSung: Scoped the no-copy claim to non-expanded parameters, matching the DDPG actor example. The Weight sharing section, `LossModule` docstring, and `convert_to_functional` now distinguish expanded/stacked ensembles (`expand_dim`, default SAC critics) and delayed targets (`create_target_params=True`). Reproduced the SAC critic case: `loss.qvalue_network is critic` holds, the original weights stay unexpanded and untrained, and the optimizer updates the ensemble.

docs/source/reference/collectors_weightsync.rst
```diff
@@ -64,6 +66,32 @@ Each of these classes is detailed below.
...
+Passing ``policy_device`` or ``device`` is not by itself evidence of a copy. If
+the policy is already on the requested device, the collector can keep the same
+object and storage. If a move is needed, it creates a separate policy and
+:meth:`~torchrl.collectors.Collector.update_policy_weights_` must propagate the
+trained weights.
+
+Calling :meth:`~torchrl.collectors.Collector.update_policy_weights_` anyway
+is conservative good practice. Use it to push weights through a
+:class:`~torchrl.weight_update.WeightSyncScheme` or another configured transport
+whenever the policies do not share storage (the rest of this page).
```
vmoens, line 93: I think we should have a bunch of example blocks here, that'd be much clearer.

vmoens: these should include various things, like using the device kwargs in collector, using remote collectors, using policy factories, etc.

YeonwooSung: Added example blocks for the cases that decide whether the call is load-bearing: shared storage on a direct collector, same-device `policy_device`/`device` (no copy), a device kwargs copy, local process workers, `policy_factory` workers, and a remote `backend="ray"` collector (`rpc`/`distributed` follow the same pattern).

YeonwooSung:
Addressed the request for example blocks in the `update_policy_weights_` section and rebased onto current `main`.

The new snippets cover shared storage, same-device vs copying `policy_device`/`device`, local process workers, `policy_factory`, and a remote Ray collector.

PR #4232, opened 2026-09-05
[Doc] Document multi-agent _step TensorDict layout

### LINKED ISSUE #2425: Document the TensorDict structure of the return of the _step() function for a multi agent environment
## Motivation

It is not very clear what should be the structure of TensorDict of the return of the _step() function for a multi agent environment.

If there are two agents A and B and they are in separate groups, what would be the structure of the TensorDict that is returned by the _step() function?

## Solution

Update the documentation and provide an example of a multi agent environment that is written natively in torchrl and it is not translated from other frameworks like petting zoo  

## Checklist

@matteobettini

YeonwooSung:
## Description

Extends `docs/source/reference/envs_multiagent.rst` with a **Multiple agent groups** section so a native `EnvBase._step()` is documented beyond the single `"agents"` group used by VMAS.

The new section:

- Shows the key tree a custom `_step()` must write for several groups (`"red"` / `"blue"`, or `"agents"` / `"adversaries"`): per-group nested tensordicts with a stacked agent dimension.
- States where **action** (input only), **observation**, and **reward** live, and where **done** / **terminated** / **truncated** live when they are shared (root reset signal), per-group / per-agent (stacked), or one-group-per-agent.
- Requires specs to mirror that nesting, with a two-group spec example and an optional per-group done spec.
- Includes a copy-paste native `TwoTeamEnv(EnvBase)` sketch (not a PettingZoo wrapper) that reads `(group, "action")` and returns observations, rewards, and shared done flags without a `"next"` wrapper.
- Cross-links the [competitive MADDPG tutorial](https://pytorch.org/rl/stable/tutorials/multiagent_competitive_ddpg.html) as the canonical multi-group training loop.

The existing VMAS single-group example is kept as the simple case. The intro now treats `"agents"` as one group name, not a required key.

### Code example

Example return value from a custom `_step`: two groups with different agent counts and shared environment termination. EnvBase.step adds the outer `"next"` key.

```python
import torch

batch_size = 4
out = TensorDict({
    group: TensorDict({
        "observation": torch.zeros(batch_size, n_agents, 8),
        "reward": torch.zeros(batch_size, n_agents, 1),
    }, batch_size=[batch_size, n_agents])
    for group, n_agents in {"red": 2, "blue": 3}.items()
}, batch_size=[batch_size])
out["done"] = torch.zeros(batch_size, 1, dtype=torch.bool)
out["terminated"] = torch.zeros_like(out["done"])
assert out["red", "reward"].shape == (4, 2, 1)
assert out["blue", "observation"].shape == (4, 3, 8)
# return out from _step; group actions are read from the input TensorDict.
```

## Motivation and Context

close #2425

The current multi-agent env page only documented the single `"agents"` group. Matteo Bettini agreed it needed a more general multi-group description; a later comment pointed at the competitive DDPG tutorial rollout as the pattern.

## Types of changes

## Checklist

### DIFF
--- a/docs/source/reference/envs_multiagent.rst
+++ b/docs/source/reference/envs_multiagent.rst
@@ -407,8 +407,12 @@ transition. In ``_step``, take devices from the action tensors.
 
         def _initial_obs(self):
             # Replace with a real reset distribution.
-            red_obs = torch.zeros(*self.batch_size, self.n_red, self.obs_dim)
-            blue_obs = torch.zeros(*self.batch_size, self.n_blue, self.obs_dim)
+            red_obs = torch.zeros(
+                *self.batch_size, self.n_red, self.obs_dim, device=self.device
+            )
+            blue_obs = torch.zeros(
+                *self.batch_size, self.n_blue, self.obs_dim, device=self.device
+            )
             return red_obs, blue_obs
 
         def _apply_dynamics(self, red_action, blue_action):
--- a/docs/source/reference/envs_multiagent.rst
+++ b/docs/source/reference/envs_multiagent.rst
@@ -20,11 +20,16 @@ Some of the main differences between these paradigms include:
 - **done** (and ``"truncated"`` or ``"terminated"``) can be per-agent or shared.
 
 TorchRL accommodates all these possible paradigms thanks to its :class:`tensordict.TensorDict` data carrier.
-In particular, in multi-agent environments, per-agent keys will be carried in a nested "agents" TensorDict.
-This TensorDict will have the additional agent dimension and thus group data that is different for each agent.
-The shared keys, on the other hand, will be kept in the first level, as in single-agent cases.
+Per-agent keys live in nested **group** tensordicts. Each group has an extra agent dimension
+so that data that differs across agents can be stacked. Shared keys stay at the root,
+as in single-agent cases.
 
-Let's look at an example to understand this better. For this example we are going to use
+The simplest layout uses a single group named ``"agents"`` (the VMAS default shown
+below). When agents belong to several groups -- for example two competing teams --
+each group is its own nested tensordict. See :ref:`MARL-multiple-groups` for the
+general contract a native :class:`~torchrl.envs.EnvBase` ``_step()`` must implement.
+
+Let's look at the single-group case first. For this example we are going to use
 `VMAS <https://github.com/proroklab/VectorizedMultiAgentSimulator>`_, a multi-robot task simulator also
 based on PyTorch, which runs parallel batched simulation on device.
 
@@ -67,7 +72,8 @@ and **info** are present in the nested "agents" tensordict with batch size `(num
 which represents the additional agent dimension.
 
 Multi-agent tensor specs will follow the same style as in tensordicts.
-Specs relating to values that vary between agents will need to be nested in the "agents" entry.
+Specs relating to values that vary between agents will need to be nested in the
+group entry (here, ``"agents"``).
 
 Here is an example of how specs can be created in a multi-agent environment where
 only the done flag is shared across agents (as in VMAS):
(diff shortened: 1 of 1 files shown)


docs/source/reference/envs_multiagent.rst
```diff
@@ -130,6 +139,358 @@ single agent standards.
...
+- **Reward.** Output of ``_step()`` only (not ``_reset()``). Per-agent or
+  per-group rewards live at ``(group, "reward")`` with shape
+  ``(*batch, n_agents_in_group, 1)``. A fully shared reward is a root
+  ``"reward"`` with shape ``(*batch, 1)``.
+- **Done / terminated / truncated** can sit at three levels, which can be
+  combined:
+
+  * **Shared (root).** ``"done"``, ``"terminated"`` and (if used)
+    ``"truncated"`` at the root, shape ``(*batch, 1)``. This is the signal
+    TorchRL uses to reset the environment. A native env must write at least
```
vmoens: [P2] Do not require a root done flag for every native env

EnvBase also supports environments whose done/terminated/reset keys are entirely nested; NestedCountingEnv(nest_done=True) and the existing nested StepCounter tests exercise that contract. Root shared flags are the convention for this example and some wrappers, not a universal native-env requirement. Qualify this paragraph so implementers are not told to add a second termination hierarchy unnecessarily.

docs/source/reference/envs_multiagent.rst
```diff
@@ -130,6 +139,358 @@ single agent standards.
...
+    this shared flag (VMAS writes only this).
+  * **Per-group / per-agent, stacked.** ``(group, "done")`` (and the
+    ``terminated`` / ``truncated`` siblings) with shape
+    ``(*batch, n_agents_in_group, 1)``. PettingZoo writes these *and*
+    aggregates them into the root flag (``any`` or ``all``, controlled by
+    ``done_on_any``).
+  * **One group per agent.** With
+    :attr:`~torchrl.envs.MarlGroupMapType.ONE_GROUP_PER_AGENT` each group
+    has a single agent, so ``(agent_name, "done")`` has shape
+    ``(*batch, 1)`` and there is no extra agent dimension.
```
vmoens: [P2] Keep the singleton agent dimension for one-agent groups

ONE_GROUP_PER_AGENT changes group membership, not the group TensorDict layout. PettingZoo constructs each group Composite with shape [n_agents] and each done leaf with [n_agents, 1], so a one-agent group still has trailing shape [1, 1] (or [*batch, 1, 1] with an env batch). Saying there is no agent dimension will make readers build incompatible specs/policies. Retain that singleton dimension in the documented shape.

vmoens (changes requested):
The TwoTeamEnv example runs and passes check_env_specs. Two statements in the new layout contract do not match the supported environment/spec behavior.

docs/source/reference/envs_multiagent.rst
YeonwooSung: Addressed in 7ac9875cc. ONE_GROUP_PER_AGENT is documented as a grouping change only. PettingZoo still builds each group Composite with shape [n_agents] and each done leaf with [n_agents, 1], so a one-agent group keeps trailing shape [*batch, 1, 1]. The previous "no extra agent dimension" wording is gone.

docs/source/reference/envs_multiagent.rst
YeonwooSung: Addressed in 7ac9875cc. The paragraph now states that root done/terminated/truncated is the convention used by the TwoTeamEnv sketch and wrappers such as VMAS, not a native-env requirement. EnvBase already resets from nested done/_reset keys (NestedCountingEnv(nest_done=True, has_root_done=False) and the nested StepCounter tests), so the text no longer tells implementers to add a second termination hierarchy.

YeonwooSung:
Updated the layout contract in 7ac9875cc for the two CHANGES_REQUESTED items:

- Root done/terminated/truncated is documented as a convention (TwoTeamEnv, VMAS), not a native-env requirement. EnvBase already supports fully nested done/_reset keys.
- ONE_GROUP_PER_AGENT keeps the singleton agent dimension: Composite shape `(*batch, 1)`, done leaf `(*batch, 1, 1)`.

The TwoTeamEnv sketch is unchanged and still uses shared root done flags.

vmoens (approved):
Re-reviewed 7ac9875. The text now allows nested-only done specs instead of requiring a root done universally, and the one-group-per-agent example retains the required singleton done dimension.

I reran the TwoTeamEnv example and check_env_specs successfully. The revised examples and shape descriptions are consistent. No remaining actionables.

PR #4235, opened 2026-09-05
[BugFix] Assert History chat-template round-trip role and content

### LINKED ISSUE #4221: [BUG] History.from_text round-trip test never asserts role/content
## Describe the bug

`History.from_text` round-trip coverage in `test/llm/test_data.py` is effectively dead. `TestHistory.test_history_round_trip` is supposed to check that `History -> apply_chat_template -> History.from_text` recovers the original roles and contents (whitespace-normalized). The role/content assertions never run, so a broken inverse parser still passes CI.

This is a **test bug**, not a request to change `History`.

## To Reproduce

The live test is `TestHistory.test_history_round_trip` in `test/llm/test_data.py` (starts at line 840). After the round-trip parse:

```884:899:test/llm/test_data.py
        def norm(x):
            if isinstance(x, list):
                return [re.sub(r"\\s+", " ", str(xx).strip()) for xx in x]
            return re.sub(r"\\s+", " ", str(x).strip())
            # Compare roles and content
            assert norm(parsed.role) == norm(
                history.role
            ), f"Roles do not match!\nOriginal: {history.role}\nParsed: {parsed.role}"
            assert norm(parsed.content) == norm(
                history.content
            ), f"Content does not match!\nOriginal: {history.content}\nParsed: {parsed.content}"

        # All messages should be complete
        assert all(
            parsed.is_complete
        ), f"All messages should be complete after round-trip. is_complete: {parsed.is_complete}"
```

Two independent problems:

1. **Dead asserts.** `norm()` returns immediately after `re.sub` (line 887). The role/content `assert`s on lines 889-894 sit after that `return`, so they are unreachable. `norm()` is never even called from the test body.
2. **Wrong regex.** `r"\\s+"` is a raw string that matches a literal backslash-s (`\s+`), not whitespace. Even if the asserts were moved out of the function, whitespace normalization would not happen. The same regex is copied into `test_history_round_trip_incomplete` at lines 965-968.

The only assertion that actually executes in `test_history_round_trip` is `assert all(parsed.is_complete)` (lines 897-899). Completeness of messages does not check that the parser recovered the right roles or contents.

A chat-template inverse parser that returned any complete `History` of the right length (or even a single complete dummy message, depending on `is_complete` broadcasting) would still pass this test.

## Expected behavior

After `History.from_text` of a string produced by `apply_chat_template`, the recovered roles and contents should match the original history, compared after real whitespace normalization (`re.sub(r"\s+", " ", ...)`). That is what the comments next to the dead asserts already describe.

`test_history_round_trip_incomplete` should use the same (fixed) normalizer if it keeps comparing roles.

Please un-indent / move the asserts out of `norm()`, fix the regex, and keep this as a test-only change.

## Screenshots

N/A

## System info

Observed on a source checkout of pytorch/rl at `1d3de3db`.

```python
# test/llm/test_data.py, TestHistory.test_history_round_trip
```

## Additional context

`test_history_round_trip_incomplete` (line 916) has the same broken `r"\\s+"` pattern, but its role asserts sit *outside* `norm()` so they do run. They still do not normalize whitespace as intended.

## Reason and Possible fixes

Move the two asserts out of `norm()` so they execute, and change `r"\\s+"` to `r"\s+"`. No `History` API change is needed.

## Checklist

YeonwooSung:
## Description

`TestHistory.test_history_round_trip` was supposed to check that `History -> apply_chat_template -> History.from_text` recovers the original roles and contents. The role/content asserts sat after `return` inside `norm()`, so they never ran, and `r"\\s+"` matched a literal backslash-s rather than whitespace. Only `assert all(parsed.is_complete)` executed.

This moves those asserts into the test body, uses `r"\s+"` for whitespace normalization, and applies the same regex fix in `test_history_round_trip_incomplete`. Qwen chats that do not start with a system message are xfailed: both History's qwen template and the official Qwen tokenizer template inject a default system turn, so that path cannot round-trip faithfully. The Qwen instruct / standard chat path still asserts role and content.

No `History` implementation change.

### Code example

Round-trip regression with a Qwen tokenizer (requires Transformers and the tokenizer files). The role/content assertions execute outside the normalization expression.

```python
import re
from transformers import AutoTokenizer
from torchrl.data.llm import History

tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen2.5-0.5B-Instruct")
history = History.from_chats([
    {"role": "system", "content": "Be brief."},
    {"role": "user", "content": "Hello  there"},
    {"role": "assistant", "content": "Hi!"},
])
text = history.apply_chat_template(
    tokenizer=tokenizer, add_generation_prompt=False, return_dict=False,
)
parsed = History.from_text(text, tokenizer=tokenizer)
assert list(parsed.role) == list(history.role)
assert [re.sub(r"\s+", " ", str(x).strip()) for x in parsed.content] == [
    re.sub(r"\s+", " ", str(x).strip()) for x in history.content
]
assert all(parsed.is_complete)
```

## Motivation and Context

A broken inverse parser that returned any complete `History` would still pass CI. The comments next to the dead asserts already described the intended check.

close #4221

## Types of changes

## Checklist

### DIFF
--- a/test/llm/test_data.py
+++ b/test/llm/test_data.py
@@ -843,7 +843,15 @@ The result is""",
         """Test round-trip conversion: History -> string -> History for various templates and tokenizers."""
         from transformers import AutoTokenizer
 
-        # Example chats
+        # Qwen chat templates (History's built-in template and the official
+        # tokenizer template) insert a default system message when the
+        # conversation does not start with one. The inverse parser then
+        # recovers that extra turn, so role/content cannot match the original.
+        if chat[0]["role"] != "system" and "qwen" in tokenizer_name.lower():
+            pytest.xfail(
+                "Qwen chat templates inject a default system message when the "
+                "conversation does not start with one"
+            )
 
         tokenizer = AutoTokenizer.from_pretrained(
             tokenizer_name, trust_remote_code=True
@@ -880,18 +888,19 @@ The result is""",
             chat_template_name=chat_template_name,
         )
 
-        # Normalize whitespace for comparison
+        # Normalize whitespace for comparison. Keep these asserts in the test
+        # body so a return inside the helper cannot make them dead again.
         def norm(x):
             if isinstance(x, list):
-                return [re.sub(r"\\s+", " ", str(xx).strip()) for xx in x]
-            return re.sub(r"\\s+", " ", str(x).strip())
-            # Compare roles and content
-            assert norm(parsed.role) == norm(
-                history.role
-            ), f"Roles do not match!\nOriginal: {history.role}\nParsed: {parsed.role}"
-            assert norm(parsed.content) == norm(
-                history.content
-            ), f"Content does not match!\nOriginal: {history.content}\nParsed: {parsed.content}"
+                return [re.sub(r"\s+", " ", str(xx).strip()) for xx in x]
+            return re.sub(r"\s+", " ", str(x).strip())
+
+        assert norm(parsed.role) == norm(
+            history.role
+        ), f"Roles do not match!\nOriginal: {history.role}\nParsed: {parsed.role}"
+        assert norm(parsed.content) == norm(
+            history.content
+        ), f"Content does not match!\nOriginal: {history.content}\nParsed: {parsed.content}"
 
         # All messages should be complete
         assert all(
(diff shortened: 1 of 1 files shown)


vmoens (approved):
LGTM!

PR #4255, opened 2026-09-07
[BugFix] Do not invent vLLM prompt log-probabilities

### LINKED ISSUE #4226: [BUG] vLLMWrapper prompt log-probs are incorrect
## Describe the bug

`vLLMWrapper` documents that the prompt log-probs it returns are not correct: the vLLM path used here does not provide usable `prompt_logprobs`, but the wrapper still writes a (padded / empty / placeholder) tensor into `LogProbs.prompt`. Users of GRPO / KL with `return_log_probs=True` can therefore consume silently wrong prompt log-probabilities.

This looks like an engine constraint, not a simple typo. The question for maintainers is what the public contract should be, not a request for a large rewrite.

## To Reproduce

When `return_log_probs=True` and `num_samples is None`, `_get_log_probs` (async / generate path) does:

```1830:1832:torchrl/modules/llm/policies/vllm_wrapper.py
            if self.num_samples is None:
                # TODO: this is not correct, we should use the prompt_logprobs
                #  but they're not returned by vLLM
```

It then reads `request_output_tc["prompt_logprobs"]` (`vllm_wrapper.py:1834-1861`). Nearby comments already describe the failure mode:

```1866:1883:torchrl/modules/llm/policies/vllm_wrapper.py
                    # Only set prompt log-probs if they actually contain
                    # data (vLLM V1 may produce all-zero padded tensors
                    # from empty per-request prompt_logprobs).
                    ...
                    # vLLM V1 may return prompt_logprobs=None per request,
                    # which from_request_output converts to empty tensors.
```

`prompt_logprobs` is requested when `return_log_probs=True`:

```610:620:torchrl/modules/llm/policies/vllm_wrapper.py
        prompt_logprobs = 1 if return_log_probs else None
        ...
        vllm_kwargs.setdefault("prompt_logprobs", prompt_logprobs)
```

If the engine still does not populate that field on the path used here, the wrapper either leaves `LogProbs.prompt` unset and later zero-fills, or stores a padded / empty tensor. Either way, a caller that reads `td["log_probs"].prompt` (GRPO KL, `RetrieveKL`, anything that scores the prompt) can get a silent wrong value rather than an error.

## Expected behavior

Please confirm the intended contract:

1. Can / should we request vLLM `prompt_logprobs` in a way that actually returns them on this generate path (and then use those values, dropping the "this is not correct" branch)?
2. If the engine cannot provide them, should `LogProbs.prompt` be omitted or set to `None`, with a warning, instead of a silently wrong tensor?

I am not proposing a rewrite of the wrapper. A documented answer (and a small change to match it) is enough.

## Screenshots

N/A

## System info

Observed on a source checkout of pytorch/rl at `1d3de3db`. Relevant when `vllm` is installed and `vLLMWrapper(..., return_log_probs=True)` is used for GRPO / KL.

## Additional context

`_build_prompt_logprobs` (`vllm_wrapper.py:2312`) already special-cases `request.prompt_logprobs is None` and zero-pads a shorter prefix. That helper makes the tensor look well-formed even when the engine returned nothing useful.

## Reason and Possible fixes

If vLLM can return `prompt_logprobs` on this path, plumb them through and delete the TODO. If it cannot, do not invent a dense tensor: leave `prompt` unset / `None` and warn once when `return_log_probs=True`. That is a contract decision for maintainers.

## Checklist

YeonwooSung:
## Description

`vLLMWrapper` requested `SamplingParams.prompt_logprobs` but the vLLM V1 generate path often returns nothing useful. The wrapper then wrote zero-filled prompt log-probs so `LogProbs.full` could be `prompt || response`. Zero log-prob is probability 1, so GRPO / KL callers that read `log_probs.prompt` (or `full` as if it contained prompt scores) got a silently wrong tensor.

This PR addresses the missing prompt-score case discussed in #4226:

- If the engine returned usable `prompt_logprobs`, keep using them.
- If it did not (None, empty tensors, all-zero padded placeholders), leave `LogProbs.prompt` unset. Keep `LogProbs.full` at prompt+response length by filling the unavailable prompt positions with NaN for alignment.
- Warn once via `torchrl_logger`.
- The prompt positions in this fallback `full` tensor are alignment placeholders, not engine-provided scores. GRPO excludes unavailable positions through a local score-validity mask, without mutating the policy distribution. KLComputation zeros NaN alignment pads while preserving real infinite log-ratios. Other consumers must likewise exclude unavailable positions or obtain actual prompt scores.

The `generate=False` prompt-scoring path is unchanged. Prefix-cache zero-padding inside `_build_prompt_logprobs` (partial real returns) is also unchanged.

### Code example

Given a configured generating `vLLMWrapper` named `policy` and its prompt TensorDict `td`, inspect `prompt` to distinguish real prompt scores from alignment padding. This example assumes padded output with prompt tokens and a nonempty response available.

```python
import torch

# policy = vLLMWrapper(..., generate=True, return_log_probs=True, pad_output=True)
# td contains the prompt expected by that wrapper.
out = policy(td)
log_probs = out["log_probs"]
response_scores = log_probs.response
assert response_scores is not None
assert log_probs.full.shape == out["tokens"].full.shape

if log_probs.prompt is None:
    # Only the response slice contains engine-provided scores.
    n_response = response_scores.shape[-1]
    torch.testing.assert_close(log_probs.full[..., -n_response:], response_scores)
    # The preceding prompt positions are NaN placeholders, not log-probabilities.
    # Exclude them from loss/KL scoring or obtain real prompt scores first.
else:
    torch.testing.assert_close(
        log_probs.full, torch.cat([log_probs.prompt, response_scores], dim=-1)
    )
```

### Async collector and CI fixes

Completed-trajectory collection keeps the pool dimension when it selects a single environment near the dialog-turn budget. This prevents lazy policy output from reaching a child environment as a scalar TensorDict:

```python

pool_batch = TensorDict(batch_size=[4, 1])
# Keep the pool axis until AsyncEnvPool unbinds it for the worker.
child_batch = pool_batch[:1].unbind(0)[0]
assert child_batch.batch_size == (1,)
# Indexing with pool_batch[0] instead would leave a scalar after unbind(0).
```

The CPU collector regression now uses the lazy batches returned by LLM wrappers. SGLang checks its resolved backend dependencies before installing TorchRL against TensorDict main; source-build version metadata can be lower than TorchRL's released TensorDict requirement. Release jobs retain the final full dependency check. The NaN-pad docstring also satisfies pydocstyle.

## Motivation and Context

close #4226

The issue asked maintainers to choose between plumbing real vLLM `prompt_logprobs` and omitting a silently wrong tensor. The `prompt` field now signals unavailable scores explicitly. Full-sequence alignment is preserved, and GRPO/KL exclude the unavailable prompt positions from scoring. If vLLM populates `prompt_logprobs` on generate, they are stored as before.

## Types of changes

Callers that assumed `log_probs.prompt` is always a dense tensor of prompt length will now see `prompt is None` when the engine omitted scores. `full` keeps prompt+response length, with NaN placeholders in the unavailable prompt slice. GRPO masks unavailable scores out of loss aggregation; KLComputation gives NaN pads zero contribution. Engine-provided prompt log-probs are preserved.

## Checklist


torchrl/modules/llm/policies/vllm_wrapper.py
```diff
@@ -1905,40 +1909,13 @@ def _generate_from_tokens(
...
                     if prompt_lp is not None and response_lp is not None:
                         log_probs_obj_flat.full = self._cat_tensors(
                             prompt_lp, response_lp
                         )
                     else:
-                        # Last resort: use response as full to avoid
-                        # missing key downstream
+                        # Prompt scores were not returned. Do not invent
+                        # zeros (log-prob 0 == probability 1).
                         log_probs_obj_flat.full = response_lp
```
vmoens: [P1] Preserve full-sequence alignment for downstream losses

On the missing-prompt-logprobs path this changes `log_probs.full` from prompt+response length to response length, while `tokens.full`, the assistant masks, and training-policy log-probs still cover the full sequence. GRPOLoss._log_weight explicitly requires matching shapes: a three-token prompt plus two-token response now gives current [1, 5] versus previous [1, 2] and raises ValueError. KL transforms similarly mask against full-sequence positions. Keep the missing-data contract explicit without relabeling a response-only tensor as `full`, and validate an actual generate-output-to-loss/KL path; the new helper tests never exercise this integration.

vmoens (changes requested):
Leaving unavailable prompt scores unset is reasonable, but using response-only scores as LogProbs.full breaks the existing full-sequence consumers. The five helper tests pass; a direct GRPOLoss consumer probe reproduces the shape failure.

torchrl/modules/llm/policies/vllm_wrapper.py
YeonwooSung: Addressed in 236879931. `LogProbs.prompt` stays unset when the engine omits scores. `LogProbs.full` is assembled to prompt+response length (alignment zeros on the prompt slice) so it matches `tokens.full` and the assistant masks. A response-only tensor is no longer stored as `full`; if prompt tokens are also missing, `full` is left unset.

Added generate-output to consumer tests in `TestVLLMPromptLogprobsConsumers`: the 3+2 token case now goes through `GRPOLoss` and `KLComputation`, and the previous response-only `full` still raises the shape `ValueError`.

torchrl/modules/llm/policies/vllm_wrapper.py
```diff
@@ -2300,6 +2289,133 @@ def _get_generic_dist(self, tensordict: TensorDictBase, **kwargs) -> D.Distribut
...
+        ``None`` when ``full`` cannot be formed without relabeling the
+        response as a full sequence.
+    """
+    if response_lp is None:
+        return None
+    if prompt_lp is not None:
+        return _cat_prompt_response(prompt_lp, response_lp)
+    if prompt_tokens is None:
+        return None
+    return _cat_prompt_response(
```
vmoens, line 2415: [P2] Prevent alignment zeros from being consumed as prompt scores

Leaving prompt unset does not protect consumers of full. With RLHF masking, all_assistant_mask can include earlier assistant turns inside the prompt. Those positions now have an old log-probability of zero, so even an unchanged policy gets a non-unit importance ratio. KLComputation likewise subtracts reference scores from these zeros without a score-validity mask. This restores the unavailable-prompt-score error the PR is intended to remove, despite fixing the previous shape mismatch.

I reproduced this through GRPOLoss._log_weight with a LogProbDistribution using the following values:

```python
# Prompt has three tokens; token 1 is an earlier assistant turn.
old_full = torch.tensor([[0., 0., 0., -0.5, -0.6]])
current_full = torch.tensor([[-0.2, -0.3, -0.4, -0.5, -0.6]])
assistant_mask = torch.tensor([[False, True, False, True, True]])
ratio = (current_full - old_full).exp()[assistant_mask]
# tensor([0.7408, 1.0000, 1.0000])
# The first ratio should also be 1 when the underlying policy is unchanged.
```

Keep full aligned, but enforce that unavailable positions cannot enter these objectives, or obtain actual prompt scores when the requested loss needs them. Add the previous-assistant-in-prompt case and check KL excludes unavailable scores; asserting only shape/finiteness (or full - ref including padding) misses the numerical error.

vmoens (changes requested):
Re-reviewed 2368799. The response-only/full shape mismatch is fixed, and all 12 focused contract/consumer tests pass. However, zero-filled prompt positions in full can still be consumed as real scores by GRPO and KL. The current tests only mask the current response for GRPO and explicitly accept the fabricated prompt contribution for KL, so they do not establish the intended missing-score contract. Details and a numerical reproduction are inline.

I also updated the PR body's stale response-only description and example to match this head. The remaining issue is score validity, not tensor length.

torchrl/modules/llm/policies/vllm_wrapper.py
YeonwooSung: Alignment zeros were being consumed as log P=1. `LogProbs.full` still has prompt+response length, but the missing prompt slice is now NaN rather than 0.0.

`GRPOLoss._log_weight` intersects `dist.mask` with `isfinite(prev_log_prob)` before the importance ratio, so a previous assistant turn inside the prompt cannot enter the objective. With the reproduced inputs (`old_full` padded, `current_full = [-0.2, -0.3, -0.4, -0.5, -0.6]`, `assistant_mask = [F, T, F, T, T]`), `ratio[assistant_mask]` is `[1, 1, 1]` and token 1 is dropped from the loss. `KLComputation` writes 0 at non-finite positions instead of `0 - ref`.

`LogProbs.prompt` stays unset; the engine is not forced to return `prompt_logprobs`. Tests now fail if zeros leak: assemble asserts the prompt slice is non-finite, GRPO asserts the unit ratio plus exclusion, and KL asserts `kl[:3] == 0` against a non-zero reference.

torchrl/objectives/llm/grpo.py
```diff
@@ -943,7 +943,11 @@ def _log_weight(
             )
             raise ValueError(error_msg)
 
-        attention_mask = dist.mask
+        # 0.0 is a valid log-prob (log 1). Alignment pads in generate-path
+        # LogProbs.full are NaN so earlier assistant turns inside the
+        # prompt cannot enter the importance ratio as fabricated scores.
+        attention_mask = dist.mask & torch.isfinite(prev_log_prob)
+        dist.mask = attention_mask
```
vmoens: [P1] Do not assign to LLMMaskedCategorical's read-only mask

The real wrapper distribution exposes mask as a property without a setter. This assignment therefore breaks ordinary GRPO calls, even with completely finite old log-probabilities; it is not confined to the missing-prompt branch. The new LogProbDistribution test passes because that different class stores a mutable mask attribute.

This public loss call returns -1.0 on the preceding head and raises AttributeError on this head:

```python
import math
import torch
from torch import nn
from torchrl.modules.distributions import LLMMaskedCategorical
from torchrl.objectives.llm import GRPOLoss

class Actor(nn.Module):
    def get_dist(self, td, **kwargs):
        return LLMMaskedCategorical(
            logits=torch.zeros(1, 5, 4),
            mask=torch.ones(1, 5, dtype=torch.bool),
        )

td = TensorDict({
    ("tokens", "full"): torch.ones(1, 5, dtype=torch.long),
    ("log_probs", "full"): torch.full((1, 5), -math.log(4)),
    "advantage": torch.ones(1, 5, 1),
}, batch_size=[1])
GRPOLoss(Actor(), entropy_bonus=False)(td)
# AttributeError: property 'mask' of 'LLMMaskedCategorical' object has no setter
```

Carry the effective score-validity mask through the loss without assuming every distribution's mask is assignable, and exercise the distribution class the real wrappers return. Avoid bypassing a distribution's mask/cache invariants by just assigning a private field.

torchrl/envs/llm/transforms/kl.py
```diff
@@ -1492,11 +1492,12 @@ def _step(
...
-        kl = [
-            gen_lp - ref_lp
-            for gen_lp, ref_lp in _zip_strict(gen_log_probs, ref_log_probs)
-        ]
+        # Here gen_log_probs = log p, ref_log_probs = log q.
+        # Non-finite values are alignment pads, not scores; 0.0 is log 1.
+        kl = []
+        for gen_lp, ref_lp in _zip_strict(gen_log_probs, ref_log_probs):
+            diff = gen_lp - ref_lp
+            kl.append(torch.where(torch.isfinite(diff), diff, torch.zeros_like(diff)))
```
vmoens: [P2] Distinguish missing-score NaNs from genuine infinite KL values

Filtering every non-finite difference hides valid infinite log-ratios. A reference policy can assign zero probability to a token, represented by ref_log_prob = -inf. With a finite generation score this must remain an infinite penalty, not zero:

```python
gen_lp = torch.tensor([-0.5, float("nan")])
ref_lp = torch.tensor([float("-inf"), -0.7])
# Expected: [inf, 0.] -- only the second position is unavailable.
# KLComputation now returns: [0., 0.]
```

I reproduced this through KLComputation(add_to_reward=False). Mask the explicit missing-score positions rather than treating all non-finite differences as padding, and preserve the existing behavior for real infinite scores.

vmoens (changes requested):
Re-reviewed 48c8b90. NaN placeholders now distinguish unavailable prompt scores from real zero log-probabilities, and the 12 submitted contract/consumer tests pass. Two regressions remain:

1. GRPOLoss now raises with the LLMMaskedCategorical returned by the real LLM wrappers, including when all old scores are finite. The test uses LogProbDistribution, whose mutable mask hides this failure.
2. KLComputation silently changes genuine infinite log-ratios to zero, although the new missing-score sentinel is NaN.

Both are reproduced below. I corrected the PR body's zero-padding description to match the new NaN sentinel.

torchrl/envs/llm/transforms/kl.py
YeonwooSung: `KLComputation` now zeros only NaN scores (`torch.isnan(gen_lp) | torch.isnan(ref_lp)`). A finite generation score against `ref_log_prob = -inf` stays `+inf`. `test_kl_preserves_infinite_ref_and_zeros_nan_pads` asserts `[inf, 0.]` on the reproduction tensors; `test_generate_output_to_kl_computation` still checks that NaN prompt pads become 0 (`4f1fa38e3`).

torchrl/objectives/llm/grpo.py
YeonwooSung: `_log_weight` now keeps `attention_mask = dist.mask & torch.isfinite(prev_log_prob)` local and returns it as a fourth value. `forward` unpacks that mask for ESS / aggregation; `dist.mask` and `dist._mask` are not written. `test_grpo_loss_with_llm_masked_categorical` runs the public `GRPOLoss` path with `LLMMaskedCategorical` and gets `loss_objective == -1.0` with no exception. NaN prompt pads are still zeroed before the ratio (`4f1fa38e3`).

YeonwooSung:
Addressed the review on `48c8b90` in `4f1fa38e3`.

- GRPOLoss: do not assign `LLMMaskedCategorical.mask`. `_log_weight` returns the effective score-validity mask; `forward` uses that for ESS / aggregation. Public `LLMMaskedCategorical` call returns `loss_objective == -1.0`.
- KLComputation: zero NaN alignment pads only. Finite gen score minus `ref_log_prob = -inf` remains `+inf` (`[inf, 0.]`).

vmoens (commented):
Rechecked 4f1fa38e3. Both substantive findings are resolved: the real `LLMMaskedCategorical` works with full and missing-prompt scores (unchanged-policy loss is -1), and KL preserves `[inf, 0]` for an impossible reference token plus a NaN prompt pad. The 14 wrapper/consumer tests and two existing loss tests pass. Additional GRPO/CISPO probes cover entropy, KL, both mean aggregations, and finite gradients with zero contribution from missing prompt positions.

One mechanical correction remains before clearing the earlier changes request: ufmt rejects the new list comprehension in `vllm_wrapper.py`. Its expected form is:

```python
return [_cat_prompt_response(p, r) for p, r in _zip_strict(prompt, response)]
```

No remaining functional findings from this round. Live vLLM/CUDA execution was not available locally.

YeonwooSung:
a281511e8 puts `_cat_prompt_response` on one line so ufmt is clean, and rebases this branch onto `06d57a05` (#4256).

vmoens (approved):
Rechecked a281511e8. The final formatter correction is fixed, and range-diff confirms the preceding functional changes were preserved by the rebase. All 14 prompt-logprob wrapper/consumer tests pass. The eight GRPO/CISPO forward-and-gradient probes also pass across token/prompt mean aggregation and available/missing prompt scores, with entropy and KL enabled.

All changed Python files pass ufmt. The earlier distribution-mutation, missing-score, and infinite-KL findings are resolved. No remaining actionable findings. Live vLLM/CUDA execution was not available locally.

PR #4256, opened 2026-09-07
[BugFix] Stop sharing Minari mission stacks between observation and next

### LINKED ISSUE #3105: [BUG] Download of categorical data in MinariExperienceReplay is completely broken

YeonwooSung:
## Description

`#3106` already converted Minari categorical fields (missions) from `NonTensorData` to `NonTensorStack`, so later episodes no longer receive the first episode's mission array.

`_preallocate_nontensor_fields` still assigned the **same** `NonTensorStack` object to both `observation` and `("next", "observation")`. Filling the TED current slice then mutated next in place, so `("next", "observation", "mission")` never shifted to the following step.

This PR:

- Allocates two independent stacks for current and next.
- Replaces the remote BabyAI download test (type check of `dataset[0]` only; 429s made it a no-op) with local tests at the production conversion and allocation/update boundaries.
- Asserts per-step scalar missions and independent current/next values with a shifted next mission. Nested key: `("observation", "mission")`.

### Code example

Minimal reproduction at the production allocation boundary: writing the current mission must not overwrite the next mission. No dataset download is needed.

```python
from tensordict import NonTensorData, NonTensorStack, TensorDict
from torchrl.data.datasets.minari_data import _preallocate_nontensor_fields

example = TensorDict({
    "observations": TensorDict({"mission": NonTensorData("mission")}),
})
storage = TensorDict(batch_size=[1])
_preallocate_nontensor_fields(storage, example, 1, {"observations": "observation"})
storage["next", "observation"].update_(
    TensorDict({"mission": NonTensorStack("open the door")}, batch_size=[1])
)
storage["observation"].update_(
    TensorDict({"mission": NonTensorStack("pick up the key")}, batch_size=[1])
)
assert storage[0]["observation", "mission"] == "pick up the key"
assert storage[0]["next", "observation", "mission"] == "open the door"
# Previously, both fields read "pick up the key" because the stacks aliased.
```

## Motivation and Context

close #3105

The issue stayed open after #3106. The remaining shared-stack bug is the part that still made next-step missions wrong. The focused tests check per-step conversion and fail if current and next share a mission stack again.

## Types of changes

## Checklist


test/libs/test_datasets.py
```diff
@@ -521,6 +531,166 @@ def custom_minari_init(custom_envs, num_episodes=5):
...
+                        batch_size=[n_obs],
+                    ),
+                },
+                batch_size=[n_obs],
+            ),
+        }
+    )
+
+
+def _materialize_minari_like_storage(episodes: dict[int, TensorDict]) -> TensorDict:
```
vmoens: [P2] Test the production Minari path instead of copying its implementation

`_materialize_minari_like_storage` reproduces about 95 lines of loader setup, allocation, iteration and copy/update logic. The regression now runs that separate implementation, so a future change to the real download/fill path can break mission handling while this test remains green. The overall test addition is about 173 lines for a 10-line fix. Extend the existing local/fake-download Minari fixture to call MinariExperienceReplay, or use a compact regression at the production preallocation/update boundary; avoid maintaining a second loader in the test.

vmoens (changes requested):
Allocating separate current/next stacks addresses the aliasing bug, and both new tests pass. The replacement regression should exercise the production loader instead of duplicating it.

YeonwooSung:
Addressed the P2 review: the copied download/fill helper is gone.

- Compact regression at `_preallocate_nontensor_fields` + production `update_` (`test_preallocate_nontensor_fields_does_not_alias_next_mission`).
- Production loader: `TestMinari.test_local_categorical_missions_keep_per_episode_values` uses `MinariExperienceReplay(..., load_from_local_minari=True)` on a local Text-mission env.

test/libs/test_datasets.py
```diff
@@ -787,28 +859,91 @@ def fake_download_dataset(dataset_id, *args, **kwargs):
...
+                    break
+        collector.create_dataset(
+            dataset_id=dataset_id,
+            algorithm_name="RandomPolicy",
+            author="torchrl-test",
+        )
         try:
-            exp_replay = MinariExperienceReplay(
-                dataset_id="minigrid/BabyAI-Pickup/optimal-v0",
+            data = MinariExperienceReplay(
```
vmoens: [P2] Make the replacement Minari integration test reach its assertions

This new test fails in MinariExperienceReplay construction with torch 2.11.0, TensorDict 0.14.2 (within the declared supported range), Minari 0.5.3, Gymnasium 1.3.0, and h5py 3.16.0:

```bash
pytest test/libs/test_datasets.py --runslow -q \
  -k test_local_categorical_missions_keep_per_episode_values
```

```text
minari_data.py:385 -> _preallocate_nontensor_fields
minari_data.py:647 -> dst_td.set(...)
RuntimeError: Received a new batch size torch.Size([6, 631])
with an existing batch_size torch.Size([6]) in a lazy TD.
```

The destination observation TensorDict already has batch_size [6, 631] when the helper inserts a [6] NonTensorStack. Thus this integration replacement currently introduces a failing test instead of checking per-episode missions. Please correct/minimize the fixture or retain the compact production-boundary regression; an unrelated loader refactor is not necessary just to add coverage for the shared-stack fix.

vmoens (changes requested):
Re-reviewed f5393bb. The copied loader implementation is gone and the compact production-allocation regression passes, including distinct current/next missions. The other focused conversion test passes as well. The new real Minari integration test fails before reaching its mission assertions; reproduction inline.

There are still 157 added test lines for 10 implementation lines (about 16:1). The compact allocation/update test directly protects this shared-stack fix. Please keep the coverage proportionate: either retain that focused regression and remove unnecessary integration scaffolding, or use a small working fixture that adds distinct coverage without reimplementing the loader.

If retaining the DataCollector test, isolate MINARI_DATASETS_PATH under tmp_path with monkeypatch before creating the dataset. root=tmp_path only isolates TorchRL's output; the fixed mission-env/test-local-v1 dataset is currently created/deleted in the user's default Minari cache, unlike the adjacent isolated-cache test.

test/libs/test_datasets.py
YeonwooSung: The compact production-boundary test `test_preallocate_nontensor_fields_does_not_alias_next_mission` is retained. It already exercises `_preallocate_nontensor_fields` plus the TED `update_` fill and asserts that current/next missions are distinct objects and shift correctly.

`test_local_categorical_missions_keep_per_episode_values` is removed. That DataCollector fixture failed in `MinariExperienceReplay` construction (`NonTensorStack` batch `[6]` into a `[6, 631]` observation TD) before any mission assertion, and it did not add coverage beyond the compact regression. No loader refactor was made to keep it.

vmoens (commented):
Re-reviewed c77e772. Removing the failing DataCollector integration also removes the default-cache collision risk. Both remaining production-boundary regressions pass, including the shifted current/next mission values. The reduced fixture and two tests cover distinct conversion and mutation behavior; I have no further test-design requests.

Only formatting remains before approval. The configured ufmt check fails at test/libs/test_datasets.py:576; it wants the allocation call on one line:

```python
_preallocate_nontensor_fields(dest, episode, total_steps=2, name_map=_NAME_MATCH)
```

Run the repository formatter on that file. I also corrected the PR body's stale description of the removed two-episode integration.

YeonwooSung:
Put the `_preallocate_nontensor_fields(...)` call on one line so ufmt accepts `test/libs/test_datasets.py`. a98c598d0

vmoens (approved):
Rechecked a98c598d0. The remaining formatting correction is fixed. Both local Minari regression cases pass, and ufmt accepts both changed Python files. The separate current/next stacks preserve shifted mission values without aliasing. No remaining actionable findings.

PR #4258, opened 2026-09-07
[BugFix] Cache AsyncEnvPool batch sizes during setup

vmoens:
## Summary

- derive and cache each worker environment batch size from specs already collected during multiprocessing pool setup
- remove the obsolete lazy worker queue lookup and its unused worker command
- verify that reading the public `env_batch_sizes` property performs no worker queue writes

## Test plan

- `python -m pytest test/envs/test_special.py::TestAsyncEnvPool -q` (27 passed)
- `black --check torchrl/envs/async_envs.py test/envs/test_special.py`
- `usort check torchrl/envs/async_envs.py test/envs/test_special.py`
- `flake8 torchrl/envs/async_envs.py test/envs/test_special.py`

### DIFF
--- a/test/envs/test_special.py
+++ b/test/envs/test_special.py
@@ -872,6 +872,9 @@ class TestAsyncEnvPool:
         env = self.make_env(makers=make_envs, backend=backend)
         assert env.batch_size == (4,)
         try:
+            if backend == "multiprocessing":
+                assert env._env_batch_sizes == [torch.Size([])] * 4
+            assert env.env_batch_sizes == [torch.Size([])] * 4
             r = env.reset()
             assert r.shape == env.shape
             s = env.rand_step(r)
--- a/test/envs/test_special.py
+++ b/test/envs/test_special.py
@@ -9,6 +9,7 @@ import pickle
 import threading
 import warnings
 from functools import partial
+from unittest.mock import patch
 
 import numpy as np
 import pytest
@@ -873,8 +874,12 @@ class TestAsyncEnvPool:
         assert env.batch_size == (4,)
         try:
             if backend == "multiprocessing":
-                assert env._env_batch_sizes == [torch.Size([])] * 4
-            assert env.env_batch_sizes == [torch.Size([])] * 4
+                with patch.object(
+                    type(env.input_queue[0]), "put", side_effect=AssertionError
+                ):
+                    assert env.env_batch_sizes == [torch.Size([])] * 4
+            else:
+                assert env.env_batch_sizes == [torch.Size([])] * 4
             r = env.reset()
             assert r.shape == env.shape
             s = env.rand_step(r)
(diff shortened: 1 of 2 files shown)


torchrl/envs/async_envs.py
```diff
@@ -704,6 +704,9 @@ def _setup(self) -> None:
         self._child_specs = []
         for i in range(num_threads):
             self._child_specs.append(self.output_queue[i].get())
+        # Batch sizes are already available from the worker specs. Caching them
+        # here avoids a later queue round trip that may block behind env work.
+        self._env_batch_sizes = [torch.Size(spec.shape) for spec in self._child_specs]
```
vmoens, line 710: With this, the lazy branch below in `env_batch_sizes` (l. 742-749) can only run for a pool with zero envs. Drop it or mark it as a defensive fallback in a comment; and it is worth saying here why the round trip hurt: the message shares `input_queue` with steps, so a first access mid-collection blocked behind the in-flight step.

test/envs/test_special.py
```diff
@@ -872,6 +872,9 @@ def test_specs(self, backend, make_envs):
         env = self.make_env(makers=make_envs, backend=backend)
         assert env.batch_size == (4,)
         try:
+            if backend == "multiprocessing":
+                assert env._env_batch_sizes == [torch.Size([])] * 4
```
vmoens: Nit: this pins the private attribute. The property is what users touch, so the stronger regression check is that `env.env_batch_sizes` sends nothing to the workers, e.g. monkeypatch `env.input_queue[i].put` to raise while accessing it. Fine to keep as is.

vmoens (commented):
Reviewed the diff and the surrounding pool code.

The fix is right: `_setup` already holds every worker's specs, so deriving the batch sizes there removes the lazy `("batch_size", None)` round trip. That round trip was not only slow, it was ordering-sensitive: the message shares `input_queue` with steps, so a first access during collection blocked until the worker finished the step in flight. Good to have it gone.

Two nits, neither blocking; mergeable as is.

- With the eager cache in place, the lazy branch in `env_batch_sizes` (torchrl/envs/async_envs.py:742-749) is unreachable for any pool with at least one env, since `_setup` runs from `__init__`. Either drop it or comment that it is a defensive fallback, otherwise it reads as a live code path.
- The test asserts the private `_env_batch_sizes`; the stronger guard is that accessing `env.env_batch_sizes` sends nothing to the workers (monkeypatch `env.input_queue[i].put` to raise for the duration of the access).

Note for sequencing: #4271 rewrites the spec collection in `_setup` for multi-env workers, so whichever of the two lands second needs to carry this caching over.

torchrl/envs/async_envs.py
vmoens: Agreed. Removed the now-unreachable lazy lookup and its obsolete worker command, and expanded the setup comment to explain that the former round trip shared the step input queues. Addressed in 89736c016.

test/envs/test_special.py
vmoens: Agreed. The test now patches the multiprocessing queue put method to raise while reading the public env_batch_sizes property, so it directly guards against worker round trips without asserting the private cache. Addressed in 89736c016.

PR #4260, opened 2026-09-07
[BugFix] Improve runtime diagnostics and CUDA timing

vmoens:
## Summary

- warn only once per replay buffer when a sample-time batch size conflicts with the constructor value
- include the fixed shared-memory key schema when AsyncEnvPool rejects unsupported input
- add synchronized CUDA timing through the timeit sync option
- emit the nested TransformedEnv default-change warning at most once per process

## Tests

- focused regression tests for all four behaviors
- test/envs/test_special.py: 146 passed, 129 skipped
- test/transforms/test_compose_and_env.py: 173 passed, 16 skipped
- legacy timeit test: 1 passed
- ufmt, flake8, pydocstyle, pyupgrade, codespell, autoflake, docstring-argument checks, and git diff --check

The full replay-buffer and utility modules have unrelated prioritized-buffer failures in this local checkout because the TorchRL C++ segment-tree extension is unavailable. The pinned pre-commit LibCST 1.0.1 source build also requires an unavailable Rust compiler, so ufmt was checked with the same pinned Black/usort versions and a compatible LibCST wheel.

### DIFF
--- a/torchrl/_utils.py
+++ b/torchrl/_utils.py
@@ -241,7 +241,8 @@ class timeit:
         name (str): The name of the timer.
         sync (bool, optional): If ``True``, synchronize CUDA before taking each
             start and elapsed timestamp. This measures completed CUDA work rather
-            than asynchronous launch time. Defaults to ``False``.
+            than asynchronous launch time. This is a no-op when CUDA is unavailable.
+            Defaults to ``False``.
 
     Examples:
         >>> from torchrl import timeit
@@ -286,7 +287,7 @@ class timeit:
         return decorated_fn
 
     def __enter__(self) -> timeit:
-        if self._sync:
+        if self._sync and torch.cuda.is_available():
             torch.cuda.synchronize()
         self.t0 = time.time()
         return self
@@ -307,7 +308,7 @@ class timeit:
             ...     if i % 10 == 0:
             ...         print(f"Elapsed: {timer.elapsed():.3f}s")
         """
-        if self._sync:
+        if self._sync and torch.cuda.is_available():
             torch.cuda.synchronize()
         self.t0 = time.time()
         return self
@@ -326,7 +327,7 @@ class timeit:
             ...     print(f"Elapsed so far: {timer.elapsed():.3f}s")
             ...     # do more work
         """
-        if self._sync:
+        if self._sync and torch.cuda.is_available():
             torch.cuda.synchronize()
         return time.time() - self.t0
 
@@ -1308,8 +1309,9 @@ class set_auto_unwrap_transformed_env(_DecoratorContextManager):
             instances. If ``False``, :class:`~torchrl.envs.TransformedEnv` will not unwrap nested instances.
             Defaults to ``True``.
 
-    .. note:: Until v0.9, this will raise a warning if :class:`~torchrl.envs.TransformedEnv` are nested
-        and the value is not set explicitly (`auto_unwrap=True` default behavior).
+    .. note:: If this value is not set explicitly, nesting
+        :class:`~torchrl.envs.TransformedEnv` instances emits an informational
+        warning and uses ``auto_unwrap=True``.
         You can set the value of :func:`~torchrl.envs.auto_unwrap_transformed_env`
         through:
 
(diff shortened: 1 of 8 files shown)


torchrl/_utils.py
```diff
@@ -282,6 +286,8 @@ def decorated_fn(*args, **kwargs):
         return decorated_fn
 
     def __enter__(self) -> timeit:
+        if self._sync:
+            torch.cuda.synchronize()
```
vmoens, line 291: This raises on a CPU-only build (`AssertionError: Torch not compiled with CUDA enabled`) and on a CUDA build without a device (`RuntimeError: Found no NVIDIA driver`). Guard with `torch.cuda.is_available()` here and in `start`/`elapsed`, or accept a device and synchronise only when it is CUDA. The test mocks `torch.cuda.synchronize`, so it does not exercise this path; a CPU test with `sync=True` would.

torchrl/envs/transforms/_base.py
```diff
@@ -1052,15 +1055,17 @@ def __init__(
...
-                        "To prepare for this change, use set_auto_unwrap_transformed_env(val: bool) "
-                        "as a decorator or context manager, or set the environment variable "
-                        "AUTO_UNWRAP_TRANSFORMED_ENV to 'False'.",
-                        FutureWarning,
-                        stacklevel=2,
-                    )
+                    if not _AUTO_UNWRAP_WARNING_EMITTED:
+                        _AUTO_UNWRAP_WARNING_EMITTED = True
+                        warnings.warn(
+                            "The default behavior of TransformedEnv will change in version 0.9. "
```
vmoens: Stale target: we are on 0.14 and the default has not changed. Since this is being touched, either name the release where `auto_unwrap` flips or make this a one-time informational warning. Also note the default warnings filter already de-duplicates per call site, so the module flag only helps across call sites; each worker process still warns once.

vmoens (commented):
Reviewed all four fixes and the tests.

Three of the four are good as they are: the once-per-buffer batch-size warning (the `getattr` for buffers unpickled from older versions is appreciated), the schema-aware `KeyError` in the shared-slot exchange, and the process-wide flag for the `TransformedEnv` FutureWarning. One change is needed before merging.

**Needs a fix: `timeit(sync=True)` crashes without CUDA.** `torch.cuda.synchronize()` is called unconditionally in `__enter__`, `start` and `elapsed` (torchrl/_utils.py:290, 311, 330). On a CPU-only build it raises `AssertionError: Torch not compiled with CUDA enabled` (just checked on the macOS wheel), and on a CUDA build without a device it raises `RuntimeError: Found no NVIDIA driver`. The flag is meant to be driven from a config value in scripts that also run on CPU (the DreamerV3 tests do exactly that), so guard it with `torch.cuda.is_available()` or take a device and synchronise only when it is CUDA. The unit test monkeypatches `torch.cuda.synchronize`, so it cannot see this; please add a CPU case that passes `sync=True` and checks the timer still records.

Minor:
- The FutureWarning text still promises a change "in version 0.9" while we are on 0.14 and the default has not moved. Since the line is being touched, either name the release where `auto_unwrap` actually flips (and register it for that release) or downgrade it to a one-time informational warning; per CLAUDE.md section 12 the target version has to be real.
- On what the flag buys: Python's default filter already de-duplicates identical warnings per call site, so the process-wide flag only removes repeats coming from different call sites. N worker processes still print N warnings, which is what the 64-worker log showed. Fine as a papercut fix, just noting the papercut is not fully closed.

Cross-PR: #4268 adds unconditional `torch.cuda.synchronize(device)` calls in the DreamerV3 training loop for the same timing purpose, and #4265 adds its own synchronisation. Once this lands, those should use `timeit(..., sync=True)` behind a flag (or CUDA events) instead of hand-rolled syncs.

torchrl/envs/transforms/_base.py
vmoens: Agreed and fixed in 269cf7110. I removed the stale release promise from both public docstrings and changed the once-per-process message to an informational UserWarning describing the current auto-unwrapping behavior.

torchrl/_utils.py
vmoens: Agreed and fixed in 269cf7110. Synchronization now runs only when CUDA is available, and the existing test is parametrized to verify both the CUDA and CPU/no-CUDA paths while still checking the recorded duration.

PR #4264, opened 2026-09-07
[Performance] Batch AsyncBatchedCollector shared-memory coordination

vmoens:
AsyncBatchedCollector now coordinates shared-memory environments from one thread, batches ready requests, and drains inference slots fairly. The collector exposes the environment exchange, preserves transition ownership, and returns an exact final frame batch even when total_frames is not divisible by frames_per_batch. Completed-trajectory mode retains its complete-trajectory behavior.

Integration with current main preserves worker/driver CPU affinity, including feeder-thread inheritance, and compile-safe pause. The existing collector benchmark includes shared-memory selection and a reproducible large-policy preset.

Validation: 32 focused collector and slot-transport tests passed; one Linux affinity test was skipped on macOS. The new non-divisible-budget cases reproduced overshoot before the fix and pass afterward. Pinned formatting, flake8, and diff checks pass. GPU comparisons and Linux affinity validation are pending.

Mixed reset/post-step inference regression: the threaded collector clears TensorDict device metadata before batching requests, while preserving tensor placement. A deterministic delayed-reset test reproduces the previous heterogeneous-stack device failure and verifies continued collection and policy features. The collector class passes 26 CPU cases (one Linux-specific skip).

### DIFF
--- a/torchrl/collectors/_async_batched.py
+++ b/torchrl/collectors/_async_batched.py
@@ -115,10 +115,97 @@ def _env_loop(
             result_queue.put(exc)
 
 
+def _env_ids(tensordict: TensorDictBase) -> list[int]:
+    """Read one environment index for each item in the leading batch dimension."""
+    values = tensordict.get(_ENV_IDX_KEY).tolist()
+    if not isinstance(values, list):
+        values = [values]
+    result = []
+    for value in values:
+        while isinstance(value, list):
+            value = value[0]
+        result.append(int(value))
+    return result
+
+
+def _env_batch_loop(
+    pool: AsyncEnvPool,
+    clients: list[PolicyClientModule],
+    result_queue: queue.Queue,
+    shutdown_event: threading.Event,
+    env_device: torch.device | None,
+    storing_device: torch.device | None,
+):
+    """Coordinate a shared-memory env pool from one thread in ready batches."""
+    exchange_keys = tuple(pool._slot_exchange._input_keys)
+    pending_actions = {}
+    policy_outputs = {}
+    stepping = 0
+    poll_interval = 0.001
+
+    try:
+        env_ids = list(range(pool.num_envs))
+        pool.async_reset_send(env_index=env_ids)
+        observations = pool.async_reset_recv(min_get=pool.num_envs)
+        for env_id, observation in zip(_env_ids(observations), observations.unbind(0)):
+            pending_actions[env_id] = clients[env_id].submit(observation)
+
+        while not shutdown_event.is_set():
+            ready_ids = [
+                env_id for env_id, future in pending_actions.items() if future.done()
+            ]
+            if ready_ids:
+                env_inputs = []
+                for env_id in ready_ids:
+                    action = pending_actions.pop(env_id).result()
+                    if env_device is not None:
+                        action = action.to(env_device)
+                    policy_outputs[env_id] = action
+                    env_inputs.append(action.select(*exchange_keys, strict=False))
+                pool.async_step_and_maybe_reset_send(
+                    lazy_stack(env_inputs), env_index=ready_ids
+                )
+                stepping += len(ready_ids)
+
+            if shutdown_event.is_set():
+                break
+            if not stepping:
+                shutdown_event.wait(poll_interval)
+                continue
+
+            try:
+                transitions, next_observations = pool.async_step_and_maybe_reset_recv(
+                    min_get=1,
+                    max_get=pool.num_envs,
+                    timeout=poll_interval if pending_actions else None,
+                )
+            except TimeoutError:
+                continue
+
+            completed_ids = _env_ids(next_observations)
+            stepping -= len(completed_ids)
+
+            # Submit the next observations before copying the transitions so
+            # policy inference overlaps the batched shared-memory copy.
+            for env_id, observation in zip(completed_ids, next_observations.unbind(0)):
+                pending_actions[env_id] = clients[env_id].submit(observation)
+
+            outputs = lazy_stack(
+                [policy_outputs.pop(env_id) for env_id in completed_ids]
+            )
+            transitions = transitions.clone().update(outputs)
+            if storing_device is not None:
+                transitions = transitions.to(storing_device)
+            result_queue.put(transitions)
+    except Exception as exc:
+        if not shutdown_event.is_set():
+            result_queue.put(exc)
+
+
 class AsyncBatchedCollector(BaseCollector):
     """Asynchronous collector with env slots and a policy server.
 
-    The collector pairs per-env coordinator threads with an
+    The collector pairs environment coordinators with an
     :class:`~torchrl.envs.AsyncEnvPool` and an
     :class:`~torchrl.modules.InferenceServer`.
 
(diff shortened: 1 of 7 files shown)


torchrl/collectors/_async_batched.py
```diff
@@ -132,10 +132,112 @@ def _env_loop(
...
+            for env_id, observation in zip(completed_ids, next_observations.unbind(0)):
+                if pause_request[0] is None:
+                    pending_actions[env_id] = clients[env_id].submit(observation)
+                else:
+                    ready_observations[env_id] = observation
+
+            outputs = lazy_stack(
+                [policy_outputs.pop(env_id) for env_id in completed_ids]
+            )
+            transitions = transitions.clone().update(outputs)
```
vmoens: Double copy: the pool's default `torch.stack` already materialised a private batch inside `recv` (the collector never passes `stack=`), so this `.clone()` copies every transition a second time and the comment above describes a copy that already happened. `update(outputs)` then rewrites every root key with identical values. Clone only for a `LazyStackedTensorDict`, merge `outputs.exclude(*transitions.keys(True, True))`, and re-measure.

torchrl/collectors/_async_batched.py
```diff
@@ -670,7 +810,10 @@ def _rollout_frames(self) -> TensorDictBase:
         while collected < self.frames_per_batch:
             # Block for at least one transition
             td = self._next_result()
-            transitions.append(td)
+            if self._uses_batched_coordinator:
+                transitions.extend(td.unbind(0))
```
vmoens: This extends by whole puts of up to `num_envs` transitions and `collected += td.numel()` below, so batches overshoot `frames_per_batch` by up to `num_envs - 1` and `total_frames` follows (reproduced: 4 envs, `frames_per_batch=10`, `total_frames=60` gives `[12, 12, 12, 11, 12, 11]`). Keep a carry-over and add a test asserting exact counts on the shm path.

vmoens (commented):
Reviewed 4f658ccd2 in depth (ran the full inference-server suite, 119 passed / 18 skipped, and the AsyncEnvPool tests, 27 passed, plus probes against the batched coordinator) and re-checked the two findings below against the merge commit 4bf53120c you pushed since. The round-robin sweep is right and `test_drain_round_robin` genuinely fails without it, the slot futures are race-free under the GIL, the shm read ordering is sound (server copies at collate before resolving, worker overwrites slots only after the action arrives), env exceptions and shutdown still propagate, and 4bf53120c composes correctly with `pause()` from #4263 (held observations are re-submitted on resume, a back-to-back pause re-parks). Two things still need to change; line numbers are for 4bf53120c.

**1. `frames_per_batch` and `total_frames` are not honoured on the shm path (reproduced).** `_rollout_frames` (torchrl/collectors/_async_batched.py:806-838) extends the batch with whole coordinator puts and counts `collected += td.numel()`. One put carries up to `num_envs` transitions, so a batch overshoots by up to `num_envs - 1` frames and `iterator()` carries the overshoot into `total_frames`. With `CountingEnv`, 4 envs, `frames_per_batch=10`, `total_frames=60`: the shm path yields batches of `[12, 12, 12, 11, 12, 11]` = 70 frames, the queue/threading path yields `[10] * 6` = 60. The docstring promises exact counts, and with 64 envs and `frames_per_batch=256` the error is up to 25 percent per batch. Fix: keep a carry-over (`td[:k]` now, remainder consumed first on the next call) and add a test asserting `batch.numel() == frames_per_batch` and the total on the shm path.

**2. The per-step copy this PR set out to remove is still done, twice.** Line 228: `transitions = transitions.clone().update(outputs)`. The collector never passes `stack=` to the pool, so the pool's default `torch.stack` (torchrl/envs/async_envs.py:262) already materialises a fresh, process-private batch inside `recv` (checked: a `torch.stack` of `share_memory_()` tensors is not shared). The `.clone()` is a second full copy of every transition, and the comment above it describes a copy that already happened inside `recv`. `update(outputs)` then rewrites every root key with identical values; only the keys missing from the transition (`policy_version`, log-probs, recurrent state) need merging. This is the "18 percent cloning" item from the diagnosis, moved rather than removed. Fix: clone only when the pool hands back a `LazyStackedTensorDict`, merge `outputs.exclude(*transitions.keys(True, True))`, state in a comment that correctness relies on the dense stack copying observations out of shm before the action is sent back, and re-measure.

Smaller points:
- The batch loop polls: `future.done()` for every pending env plus `shutdown_event.wait(0.001)` / `recv(timeout=0.001)`, and `_receive_batch` keeps waiting to its deadline after the first result. That is fine at 50 ms env steps but adds up to 1 ms per hand-off and 1,000 wakeups/s, and with `service_backend="process"` each `done()` is an `mp.Queue.get(timeout=0)`. The docs recommend shm without saying it targets slow envs; document the regime (or make the interval a parameter), and longer term wake the coordinator from `SlotTransport.resolve`.
- `_default_collate` switching from `lazy_stack` to `maybe_dense_stack` (torchrl/modules/inference_server/_server.py) changes the documented default for every server backend: results become views of a dense batch instead of the caller's tensordicts, and shape mismatches silently fall back to lazy. Not a bug, but it needs a note in the docstring and the release notes (CLAUDE.md section 12), or the collector should pass its own collate.
- `env_exchange="auto"` is rejected unless the backend is multiprocessing, while `AsyncEnvPool` accepts auto with threading and warns that auto becomes its default in v0.15. When the collector follows that default, `env_backend="threading"` will raise out of the box. Reject only `"shm"` there.
- `exchange_keys = tuple(pool._slot_exchange._input_keys)` reaches into a private attribute; expose `AsyncEnvPool.exchange_keys`.
- Tests: nothing drives the batch loop through a failing policy or a mid-step shutdown; `test_shared_memory_uses_batched_coordinator` checks env-id coverage and key presence only (a `next.observation == observation + 1` check is cheap and catches misattribution), and `assert len(collector._workers) == 1` pins internals. `test_default_collate_prefers_dense_batches` restates `maybe_dense_stack`'s contract.
- `_slot.py` still says "``lazy_stack`` still calls ``torch.stack``" next to the new sweep; `_env_ids` duplicates the unwrapping in `_record_trajectory_transition` (and #4271 adds a third copy).

Interaction with the sibling PRs: #4270 and #4272 are stacked on this file and add coordinator modes of their own, so the carry-over and clone fixes have to be re-applied there; #4271 is a semantic duplicate (a second batched coordinator per worker that pins `exchange="queue"`), so the two designs should be reconciled before either merges; #4269 only forwards `static_batch_size` and composes.

torchrl/collectors/_async_batched.py
vmoens: Fixed in 2860da7a3. Dense transition batches now skip the redundant clone, lazy batches retain the safety clone, and only policy keys missing from the transition are merged. Dense collation is scoped to AsyncBatchedCollector so the generic server default remains unchanged. The same 64-env MPS preset measured 876.7 records/s and exactly 1,000 requested frames, versus 806 records/s before this fix.

torchrl/collectors/_async_batched.py
vmoens: Fixed in 2860da7a3. Batched coordinator results are split into individual transitions and excess items are carried into the next rollout. The regression case now yields six batches of exactly 10 frames (60 total) and also checks environment coverage plus next-observation integrity.

PR #4265, opened 2026-09-07
[Performance] Use native replay collection in DreamerV3

vmoens:
DreamerV3 now uses native TorchRL replay throughout: one replay member per environment, synchronous routing by batch dimension and asynchronous routing by integer stream ID. Collectors insert canonical transitions in the parent; sampling returns T+1 records, training consumes T transitions and generation-checked context updates target the following rows. Reset records affect reporting and update scheduling without becoming artificial replay rows.

Depends on #4275 for parent-process collector replay insertion. The inference fixes and native ordered/streaming replay are merged into main. Whole-step compilation and eager warm-up before collection are adapted in #4268.

The previous script-owned replay worker, record builder, shifted writer and sampler are removed. Native replay owns ordered operations, future lifetime, errors, synchronization and shutdown. The full pinned batch is transferred before slicing, graph-backed outputs are cloned before asynchronous updates, and cleanup uses nested finally blocks.

Validation on the refreshed combined stack:
- Full CPU DreamerV3 objective suite: 82 passed, 3 GPU cases deselected.
- Native replay selection: 8 passed, including fresh/uniform sampling, reset accounting, capacity checks, CPU benchmark execution and real sync/async training smokes.
- Formatting, diff and complete outgoing code/metadata isolation audits pass.
- GPU captured learner, matched timing and integration validation are running in the dependent compile stack; performance readiness remains pending those results.

Final integration validation: 904 combined CPU inference/objective/replay tests passed. Eight real collection smoke cases pass on CPU and CUDA in the downstream generic SOTA integration. The seeded async policy update path and native replay schema are included. Six separate-process checkpoint/resume CUDA cases pass with the downstream checkpoint PR.

Oldest-dependency CI found a hard-coded Pendulum-v1 test assumption. Tests now use the existing version-aware Pendulum helper. All three native replay collection smoke cases pass against Gym 0.13; the downstream generic SOTA suite passes eight cases against the same backend. Production defaults are unchanged.

The base is #4275 (`codex/async-collector-replay`). The public optimizer extraction #4282 is merged, so a separate optimizer integration base is no longer needed. The example imports its optimizer from TorchRL core.

Validation after refreshing the stack onto main with #4297 merged (2026-09-08): the feature additions and removals are unchanged. The combined CPU DreamerV3, component, native-replay and config selection passed 394 tests, including separate-process checkpoint/resume. CUDA results and benchmark measurements above precede this restack; new CPU/GPU/dependency CI must validate the refreshed heads.

### DIFF
--- a/benchmarks/ad_hoc/bench_dreamer_v3_learner.py
+++ b/benchmarks/ad_hoc/bench_dreamer_v3_learner.py
@@ -6,8 +6,8 @@
 
 The workload uses the maintained DMC Walker configuration and includes all
 model, actor, value, and replay-value losses, backward, the optimizer step,
-and the slow-value-target update. Replay sampling and environment collection
-are intentionally outside the learner timing.
+and the slow-value-target update. The optional replay workload uses the same
+native replay construction, prefetch, and conditional writeback as training.
 
 Example::
 
@@ -17,6 +17,7 @@ Example::
 from __future__ import annotations
 
 import argparse
+import functools as ft
 import json
 import runpy
 import statistics
@@ -71,6 +72,9 @@ def _make_data(
                 "terminated": torch.zeros(
                     batch, steps, 1, dtype=torch.bool, device=device
                 ),
+                "truncated": torch.zeros(
+                    batch, steps, 1, dtype=torch.bool, device=device
+                ),
             },
         },
         [batch, steps],
@@ -78,24 +82,80 @@ def _make_data(
     )
 
 
-def _measure(
-    learner_update,
-    data: TensorDict,
-    *,
-    device: torch.device,
-    warmup: int,
-    iterations: int,
-) -> list[float]:
+class _ReplayLearnerStep:
+    def __init__(
+        self,
+        example: dict,
+        cfg,
+        learner_update,
+        *,
+        device: torch.device,
+        replay_device: torch.device,
+        obs_dim: int,
+        action_dim: int,
+    ):
+        self.cfg = cfg
+        self.device = device
+        self.learner_update = learner_update
+        self.replay_context_update = example["replay_context_update"]
+        sequence_records = cfg.replay_buffer.seq_len + 1
+        num_streams = 4
+        records_per_stream = sequence_records * cfg.replay_buffer.batch_size
+        cfg.replay_buffer.buffer_size = records_per_stream * num_streams
+        cfg.replay_buffer.online = False
+        self.replay_buffer = example["_build_replay"](
+            cfg, num_streams, replay_device, device
+        )
+        replay_data = _make_data(
+            cfg,
+            replay_device,
+            batch=num_streams,
+            steps=records_per_stream,
+            obs_dim=obs_dim,
+            action_dim=action_dim,
+        )
+        replay_data["is_init"][:, 0] = True
+        self.replay_buffer.extend(replay_data)
+
+    def __call__(self) -> None:
+        sequence_records = self.cfg.replay_buffer.seq_len + 1
+        replay_sample = self.replay_buffer.sample().reshape(
+            self.cfg.replay_buffer.batch_size, sequence_records
+        )
+        sample_info = replay_sample.select("index", "index_generation")
+        sample = replay_sample.exclude("index", "index_generation")
+        sample = sample.to(self.device, non_blocking=True)[:, :-1]
+        if self.learner_update.replay_must_be_idle:
+            self.replay_buffer.synchronize()
+        _, state, belief = self.learner_update(sample)
+        if self.cfg.optimization.cudagraph_train_step:
+            state = state.clone()
+            belief = belief.clone()
+        index, generation, patch = self.replay_context_update(
+            sample_info, state, belief
+        )
+        self.replay_buffer.submit_update_if_present(
+            index=index, generation=generation, patch=patch
+        )
+
+    def synchronize(self) -> None:
+        self.replay_buffer.synchronize()
+        if self.device.type == "cuda":
+            torch.cuda.synchronize(self.device)
+
+    def close(self) -> None:
+        self.replay_buffer.shutdown()
+
+
+def _measure(step, synchronize, *, warmup: int, iterations: int) -> list[float]:
     for _ in range(warmup):
-        learner_update(data)
-    torch.cuda.synchronize(device)
-    samples = []
+        step()
+    synchronize()
+    started = time.perf_counter()
     for _ in range(iterations):
-        started = time.perf_counter()
-        learner_update(data)
-        torch.cuda.synchronize(device)
-        samples.append((time.perf_counter() - started) * 1000)
-    return samples
+        step()
+    synchronize()
+    return [(time.perf_counter() - started) * 1000 / iterations]
 
 
 def main() -> None:
@@ -105,6 +165,11 @@ def main() -> None:
     parser.add_argument("--unroll", type=int, default=8)
     parser.add_argument("--warmup", type=int, default=10)
     parser.add_argument("--iterations", type=int, default=50)
+    parser.add_argument(
+        "--replay-device",
+        choices=("cpu", "cuda"),
+        help="Include native replay sampling and writeback on this device.",
+    )
     parser.add_argument(
         "--variants",
         nargs="+",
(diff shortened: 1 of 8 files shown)


benchmarks/ad_hoc/bench_dreamer_v3_learner.py
```diff
@@ -78,22 +79,87 @@ def _make_data(
...
+        self.pipeline.prefetch(self.replay_buffer)
+
+    def __call__(self) -> None:
+        replay_sample, sample_info = self.pipeline.take(self.replay_buffer)
+        replay_sample = replay_sample.reshape(
+            self.cfg.replay_buffer.batch_size,
+            self.cfg.replay_buffer.seq_len + 1,
+        )
+        sample = replay_sample[:, :-1].to(self.device, non_blocking=True)
+        self.pipeline.apply_pending_context(self.replay_buffer)
```
vmoens: This raises on the second step: `_make_data` has no `("collector", "context_valid")` while `_refresh_replay_context` patches that key unconditionally, so `synchronize()` in `_measure` surfaces `KeyError: "Key ('collector', 'context_valid') does not exist in the storage"` during warmup (reproduced on CPU with a stub learner). Add the key to `_make_data` or build the records with `DreamerV3ReplayRecordBuilder`, and cover `_ReplayLearnerStep` with a CPU test.

sota-implementations/dreamer_v3/train.py
```diff
@@ -694,7 +700,7 @@ def main(cfg: DictConfig):
                     cfg.replay_buffer.batch_size,
                     cfg.replay_buffer.seq_len + 1,
                 )
-                sample = replay_sample[:, :-1].to(device)
+                sample = replay_sample[:, :-1].to(device, non_blocking=True)
```
vmoens: `[:, :-1]` is a non-contiguous view, so this copy goes through a contiguous temporary that is not pinned (`view.contiguous().is_pinned()` is False in the PR venv): the transfer is pageable and host-blocking, as before, plus one extra copy into pinned memory. Copy the contiguous `[B, T+1]` batch and slice on device: `replay_sample.to(device, non_blocking=True)[:, :-1]`.

sota-implementations/dreamer_v3/dreamer_v3_replay.py
```diff
@@ -150,28 +156,37 @@ def has_pending_context(self) -> bool:
...
-        """Return the prefetched batch, and sample the next one."""
+        """Return the prefetched batch, and start sampling the next one."""
         self.prefetch(replay_buffer)
-        current = self._prefetched
-        self._prefetched = replay_buffer.sample(return_info=True)
+        prefetched = self._prefetched
+        current = prefetched.result()
+        self._prefetched = self._executor.submit(replay_buffer.sample, return_info=True)
+        self._inflight.append(self._prefetched)
         return current
```
vmoens: Consumed sample futures stay in `_inflight` until `synchronize()`, so every sampled batch of a collector batch is kept alive (10 of 10 alive after 10 updates in a probe with a real `ReplayBuffer`), pinned when `pin_memory=True`. Remove the consumed future here once `result()` has returned, or track only refresh futures plus the single outstanding prefetch.

sota-implementations/dreamer_v3/train.py
```diff
@@ -654,6 +658,8 @@ def main(cfg: DictConfig):
             cfg, metrics_jsonl_path, completed_episodes, batch_start_action_step
         )
         replay_data = replay_record_builder(data)
+        with timeit("dreamer_v3/replay_wait"):
+            replay_pipeline.synchronize()
```
vmoens: This drain is the contract that makes the design safe (the sampler's online queue and stream lengths are mutated outside `_replay_lock`), but nothing tests it. Also consider draining right before the learner call while `CudaGraphModule` is still warming up: the capture uses `capture_error_mode="global"`, which errors on CUDA calls from other threads, and the worker is pinning/copying at exactly that moment.

vmoens (commented):
Reviewed the diff in depth and ran it: the full non-GPU `test_dreamer_v3.py` (75 passed), plus the sota script on CPU on base and head with identical overrides over 80 updates and a ring buffer that wraps three times. The loss trajectories are bit-identical, which confirms the sampling distribution and the write-back order are unchanged. The concurrency design is sound for the training loop: the buffer's `_replay_lock` makes storage reads and writes atomic, `synchronize()` before every `extend` means the worker never runs while the main thread mutates the buffer or the sampler's online queue, `update_if_present`'s generation check drops write-backs to recycled slots, and worker exceptions surface at `take()`/`synchronize()`/`close()`. Four things need to change before merging.

**1. The new `--replay-device` benchmark path crashes on its second step (reproduced on CPU).** `_make_data` builds records without `("collector", "context_valid")`, but `_refresh_replay_context` patches that key unconditionally (dreamer_v3_replay.py:128-130), so the second `_ReplayLearnerStep.__call__` raises `KeyError: "Key ('collector', 'context_valid') does not exist in the storage"` out of `synchronize()` during warmup. Add the key to `_make_data` (or build the records through `DreamerV3ReplayRecordBuilder`), and add a CPU test that drives `_ReplayLearnerStep` with a stub learner; the class has no CUDA dependency.

**2. The pinned, non-blocking transfer is inert and only adds a copy.** train.py:703 does `replay_sample[:, :-1].to(device, non_blocking=True)`. The slice is a non-contiguous view, and a non-contiguous host-to-device copy goes through a contiguous temporary that is not pinned (checked in the PR venv: `view.is_pinned()` is True, `view.contiguous().is_pinned()` is False), so the actual transfer is a pageable, host-blocking copy, exactly what the code did before, with an extra copy into pinned memory paid by the worker. Transfer the contiguous `[B, T+1]` batch and slice on device: `replay_sample.to(device, non_blocking=True)[:, :-1]` (same in the benchmark). The README sentence about non-blocking copies is not true until then.

**3. Every sampled batch of a collector batch stays alive until the next `synchronize()`.** `take()` appends each new sample future to `_inflight` (dreamer_v3_replay.py:175) and only `synchronize()` clears the list, so the consumed futures keep their full sampled TensorDicts alive for the whole collector batch (checked with a real `ReplayBuffer`: 10 of 10 batches alive after 10 updates, 0 after `synchronize()`). With `pin_memory=True` that is pinned host memory the caching host allocator cannot recycle until the drain; for image observations that is hundreds of MB to GBs per collector batch at `train_ratio` update counts. Remove the consumed future from `_inflight` in `take()` once `result()` returned, or track only refresh futures plus the single outstanding prefetch.

**4. With `cudagraph_train_step=true`, the worker runs CUDA calls during the capture.** tensordict's `CudaGraphModule` captures on the 5th call with `torch.cuda.graph`'s default `capture_error_mode="global"`, which errors on potentially unsafe CUDA API calls from any thread (`cudaMalloc`, synchronous `cudaMemcpy`, `cudaHostAlloc`). In each iteration the main thread submits the next sample and the pending refresh before calling the learner, so during the capturing call the worker may be pinning a batch, doing the device-to-host latent copy inside `update_if_present`, or allocating on the replay device. Expected symptom: `CUDA error: operation not permitted when stream is capturing` surfacing from `take()`/`synchronize()`, timing dependent, on exactly the configurations this targets (`reproduce_dmc_walker.sh --fast`, the benchmark's `cuda_graph` variant with `--replay-device`, the GB200 runs). I could not run this without CUDA, so please test it; the simple fix is to drain the pipeline before the learner call while the graph is still warming up (overlap resumes after capture), and longer term let `CudaGraphModule` accept `capture_error_mode="thread_local"`.

Smaller points:
- Timer semantics changed silently: `dreamer_v3/replay_update` now brackets an `executor.submit` (microseconds) and two async copies, so the write-back cost is no longer measured anywhere, while `dreamer_v3/replay_sample` now includes the wait for the worker plus the host-to-device copy (which, per item 2, currently absorbs the wait for the in-flight GPU update). Time the worker functions themselves and document the new meaning of the keys in the README.
- Tests: the new test checks FIFO order and error propagation with a fake replay, but not the load-bearing contract that `synchronize()` precedes `extend`/`observe_extend` (the sampler's `_online_queue` and `_stream_lengths` are mutated outside `_replay_lock`; dropping the drain would silently lose or duplicate online blocks and no test would notice). A `ReplayBuffer` subclass whose `extend` asserts the pipeline is drained, driven through the same sequence as `train.py`, would cover it. There is also no GPU test although the PR adds CUDA-only behaviour (`pin_memory`, `non_blocking`, worker-side D2H); a `@pytest.mark.gpu` test running six updates with `cudagraph_train_step=True` through the pipeline would have caught items 2 and 4.
- The `stage_context` float copy is redundant: `CudaGraphModule` already clones its outputs on every replay and eager outputs are fresh tensors. Drop it or reword the comment.
- `replay_pipeline.close()` is not in a `finally`, and a failed background sample is sticky (`_prefetched` keeps the failed future, so every later `take()` and `close()` re-raise it, and `close()` re-raising can mask the original error).
- README: "The timing excludes replay sampling" is now only true for the default invocation.

Interaction with the sibling PRs: #4268 edits the same update-loop hunk (it inserts `torch.cuda.synchronize(device)` right around the learner call, which is what this PR is relaxing) and rewrites the benchmark around `_measure`, whose signature this PR changed, so expect textual and semantic conflicts. #4270 has a collector-owned writer that calls `replay_buffer.extend()` off the main thread; if the example adopts it, the drain-before-extend discipline no longer covers writes and the unprotected sampler state above becomes a live race, so the two need a shared hand-off before they can be combined.

benchmarks/ad_hoc/bench_dreamer_v3_learner.py
vmoens: Agreed and fixed in 701878132. The synthetic records now include collector/context_valid, and a CPU regression drives _ReplayLearnerStep through two calls, synchronizes the refresh, and verifies the context-valid writeback.

sota-implementations/dreamer_v3/train.py
vmoens: Agreed and fixed in 701878132 in both the training loop and benchmark. The contiguous [B, T+1] batch now transfers first with non_blocking=True, and the sequence slice is taken on the learner device.

sota-implementations/dreamer_v3/dreamer_v3_replay.py
vmoens: Agreed and fixed in 701878132. _inflight now contains refresh futures only; the pipeline tracks just one prefetch separately. A consumed or failed prefetch is cleared immediately, and the regression checks that samples no longer accumulate in _inflight.

sota-implementations/dreamer_v3/train.py
vmoens: Agreed and fixed in 701878132. Replay extension now goes through a helper that drains first, with an ordering regression. The learner exposes the CUDA-graph warmup/capture window and both the training loop and benchmark drain immediately before those calls. A GPU-marked regression drives six replay-backed updates through capture; CI will run it because this host has no CUDA.

PR #4266, opened 2026-09-07
[Feature] Add CPU affinity controls to AsyncEnvPool

vmoens:
## Summary

- add per-worker Linux CPU affinity masks to multiprocessing AsyncEnvPool and apply them before environment construction
- add a driver affinity mask for AsyncBatchedCollector coordinator and inference-server execution
- document cpuset, CFS quota, and Kubernetes CPU Manager interactions, and add a many-core step-latency jitter benchmark

## Tests

- 56 passed, 3 skipped across TestAsyncEnvPool, TestAsyncBatchedCollector, and TestEnvConfigs on macOS
- Linux-only worker and driver affinity assertions collected and skipped locally for CI
- pre-commit hooks pass on all changed files
- new latency sampler smoke test passes

### DIFF
--- a/docs/source/reference/envs_vectorized.rst
+++ b/docs/source/reference/envs_vectorized.rst
@@ -356,6 +356,65 @@ per-environment receives own their tensor storage. Aggregate batches returned
 with ``stack="lazy"`` are views over the shared slots and remain valid only until
 actions are sent back to the corresponding environments.
 
+.. _async_env_pool_cpu_affinity:
+
+CPU affinity (Linux)
+~~~~~~~~~~~~~~~~~~~~
+
+The multiprocessing backend accepts ``worker_affinity`` to restrict each
+environment worker to selected CPUs. Pass one CPU mask per environment or a
+callable that returns a mask for an environment index. The mask is applied
+before the environment factory runs, so subprocesses created by the factory
+inherit the worker's affinity.
+
+This option is intended for deployments where Python workers, CPU-heavy
+simulators, and driver services share a constrained CPU set. Without explicit
+placement, the scheduler may run them on the same CPUs, increasing contention
+and environment step-time jitter. Most applications should leave the option
+unset and use the operating system's scheduler.
+
+TorchRL can discover the CPUs available to the current process, but that set
+does not reveal which CPUs the application reserves for its driver or other
+services, nor how many threads each environment and its subprocesses require.
+An automatic partition would therefore impose an arbitrary workload policy and
+can reduce throughput. Affinity is explicit for that reason.
+
+.. code-block:: python
+
+    import os
+
+    available_cpus = sorted(os.sched_getaffinity(0))
+    worker_cpus = available_cpus[:8]
+    worker_masks = [
+        tuple(worker_cpus[start : start + 2])
+        for start in range(0, len(worker_cpus), 2)
+    ]
+    env = AsyncEnvPool(
+        [make_env] * len(worker_masks),
+        backend="multiprocessing",
+        exchange="queue",
+        worker_affinity=worker_masks,
+    )
+
+Every subprocess created by an environment factory inherits that worker's
+mask. A one-CPU mask therefore confines both the Python worker and all of its
+environment subprocesses to that single CPU; use a wider window when the
+environment launches CPU-parallel work.
+
+Affinity masks do not replace container resource configuration:
+
+- A process can only run on CPUs granted by its container or cgroup cpuset.
+  TorchRL validates each mask against ``os.sched_getaffinity(0)`` before
+  workers start; build masks from those currently available indices.
+- A CFS CPU quota limits CPU time, not CPU placement. Affinity neither bypasses
+  that quota nor reserves the selected CPUs, and pinning to fewer CPUs than the
+  quota can reduce usable parallelism.
+- In Kubernetes, exclusive container CPUs require the kubelet CPU Manager's
+  ``static`` policy, a ``Guaranteed`` pod, and integer CPU requests (with the
+  matching limits required for ``Guaranteed`` QoS). The resulting cpuset bounds
+  the masks accepted here; the application must still divide those assigned
+  CPUs between its workers and driver threads.
+
 .. note:: This class and its subclasses should work when nested in with :class:`~torchrl.envs.transforms.TransformedEnv` and
     batched environments, but users won't currently be able to use the async features of the base environment when
     it's nested in these classes. One should prefer nested transformed envs within an `AsyncEnvPool` instead.
--- a/torchrl/collectors/_async_batched.py
+++ b/torchrl/collectors/_async_batched.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import contextlib
+import os
 import queue
 import threading
 import time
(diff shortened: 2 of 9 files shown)


vmoens (commented):
Reviewed the diff, the merge against current main, and ran the touched suites locally (macOS, so the two Linux-only affinity tests skipped; 56 passed, 3 skipped, matching the description).

The worker side is right: the mask is applied inside `_env_exec` before the factory runs, acknowledged through a startup handshake, and independent of the start method. Two things need to change before merging.

**1. The conflict with main is semantic, not textual.** #4263 (a6c3e5ec0) made `_env_loop` take a required `pause_request` argument and `_ensure_started` now passes `"pause_request": self._pause_request` in the thread kwargs. This PR rewrites the same block inside its try/finally (torchrl/collectors/_async_batched.py:497-513) without that key. Resolving the conflict by taking this side makes every coordinator thread die at call time with `TypeError: _env_loop() missing 1 required positional argument: 'pause_request'`, raised before `_env_loop`'s own try/except, so it goes to `threading.excepthook` and the collector never yields a frame. Please rebase, keep the try/finally, add the kwarg, and keep both new imports (`contextlib` from main, `os` from here). The pause tests from #4263 will catch a wrong resolution.

**2. `driver_affinity` does not cover the threads that do the per-step serialization.** The pool is built at torchrl/collectors/_async_batched.py:459-468, and the pinned window only opens afterwards at 483-485. `ProcessorAsyncEnvPool._setup` does the first `put` on every `input_queue[i]` during construction (torchrl/envs/async_envs.py:780), which starts one `multiprocessing.Queue` feeder thread per env in the parent (checked on the PR head: two `QueueFeederThread`s exist right after `AsyncEnvPool(...)` returns). Those threads inherit the unpinned mask, and since the collector hard-codes `exchange="queue"` (line 465) every action sent to a worker is pickled and written to the pipe by exactly those threads. On the motivating pod that leaves the per-step serialization free to run on the reserved JVM windows, which is the contention the option exists to remove. Fix: open the pinned window before constructing the pool (workers re-pin themselves in `_env_exec`, so they are unaffected), and extend `test_cpu_affinity` to assert `os.sched_getaffinity(q._thread.native_id) == set(driver_affinity)` for the pool's input queues. Pinning explicitly inside `_env_loop` and the server thread instead of relying on inheritance would be more robust to threads created later.

Smaller points:
- Masks are validated for type and sign only (`_validate_cpu_affinity`, torchrl/envs/async_envs.py:35-53). A CPU outside the container cpuset fails late (`RuntimeError: worker 3 failed to set its CPU affinity: OSError(22, 'Invalid argument')` after all processes were spawned, or a bare `OSError` from the driver path), and Linux silently intersects a partially valid mask, so `{2, 3, 200}` becomes `{2, 3}` without a word. Compare each mask against `os.sched_getaffinity(0)` at validation time and name the offending CPUs. Also accept `numbers.Integral` (masks built with `numpy.array_split` fail the `isinstance(cpu, int)` check).
- The collector validates `driver_affinity` eagerly but `worker_affinity` only at first iteration inside `AsyncEnvPool.__init__`; `len(create_env_fn)` is known in `__init__`, so validate both there.
- The docstring and the docs say a process-backed inference server inherits the driver mask. True for spawn and fork; with `service_backend_options={"mp_context": "forkserver"}` the child inherits the forkserver daemon's mask. Qualify it or apply the mask inside the server process.
- The examples (docs/source/reference/envs_vectorized.rst:380, collectors_single.rst:142) pin each worker to a single CPU, right after the text says subprocesses started by the factory inherit the worker's mask. For the motivating case (a JVM launched by the factory) that confines the JVM to one core. Show a window per worker and say explicitly that a one-CPU worker mask confines every subprocess the env launches.
- Tests: the failure path (`affinity_error` in the worker, terminate/join in the parent) is untested; a Linux-only test with `worker_affinity=[(max(os.sched_getaffinity(0)) + 1,)] * n` always yields EINVAL and would protect that teardown. `test/test_configs.py` is untouched; add the default and one nested-list merge for `BatchedEnvConfig.worker_affinity` (I checked OmegaConf accepts `[[0, 1], [2, 3]]` and rejects `[["a"]]`).
- Benchmark: `benchmark.pedantic` returns the last round, so the `step_latency_*` extra_info describes one round out of five while the compared metric is wall time of 2048 transitions, and only the workers are pinned in the `pinned` variant. Say so in the docstring.

Interaction with #4271: that PR groups several envs per process, while this one indexes `worker_affinity` per environment and hands mask `i` to process `i`, which is only well defined while processes and envs are 1:1. Whichever lands second has to define the per-process mask (union of the group's masks, or raise when they differ).

torchrl/collectors/_async_batched.py
```diff
@@ -450,33 +478,45 @@ def _ensure_started(self) -> None:
...
+            if not self._server.is_alive:
+                self._server.start()
+
+            # Start per-env coordinator threads
+            self._result_queue = queue.Queue()
+            self._shutdown_event = threading.Event()
+
+            self._workers = []
+            for i in range(self._num_envs):
+                t = threading.Thread(
```
vmoens, line 547: Semantic conflict with main: after #4263 `_env_loop` requires `pause_request`, and `_ensure_started` on main passes `"pause_request": self._pause_request` here. Taking this side as is makes every coordinator thread die with a `TypeError` before `_env_loop`'s try/except runs.

torchrl/collectors/_async_batched.py
```diff
@@ -450,33 +478,45 @@ def _ensure_started(self) -> None:
...
-                daemon=True,
-                name=f"AsyncBatchedCollector-env-{i}",
-            )
-            self._workers.append(t)
-            t.start()
+        original_affinity = None
+        try:
+            if self._driver_affinity is not None:
+                original_affinity = os.sched_getaffinity(0)
+                os.sched_setaffinity(0, self._driver_affinity)
```
vmoens, line 507: The window opens after `AsyncEnvPool(...)` was built at l. 459-468. The pool's `_setup` already did the first `put` on each `input_queue[i]` (async_envs.py:780), which started the parent-side `QueueFeederThread`s that pickle and write every action per step (the collector hard-codes `exchange="queue"`). Those threads keep the unpinned mask. Move the window before the pool construction, and assert the feeder threads' masks in `test_cpu_affinity`.

torchrl/envs/async_envs.py
```diff
@@ -31,6 +32,27 @@
...
+            f"{option_name} requires os.sched_setaffinity, which is only "
+            "available on Linux."
+        )
+    try:
+        cpu_affinity = tuple(cpu_affinity)
+    except TypeError as err:
+        raise TypeError(f"{option_name} must be a sequence of CPU indices.") from err
+    if not cpu_affinity:
+        raise ValueError(f"{option_name} must contain at least one CPU index.")
+    if any(isinstance(cpu, bool) or not isinstance(cpu, int) for cpu in cpu_affinity):
```
vmoens: Validate against `os.sched_getaffinity(0)` here and name the offending CPUs: an out-of-cpuset index otherwise fails after all workers were spawned with an opaque `OSError(22)`, and a partially valid mask is silently intersected by the kernel. Also accept `numbers.Integral` so numpy-built masks pass.

docs/source/reference/envs_vectorized.rst
```diff
@@ -356,6 +356,45 @@ per-environment receives own their tensor storage. Aggregate batches returned
...
+
+    import os
+
+    available_cpus = sorted(os.sched_getaffinity(0))
+    worker_cpus = available_cpus[:8]
+    env = AsyncEnvPool(
+        [make_env] * len(worker_cpus),
+        backend="multiprocessing",
+        exchange="queue",
+        worker_affinity=[(cpu,) for cpu in worker_cpus],
```
vmoens: This pins each worker to one CPU right after saying that subprocesses created by the factory inherit the mask; for a JVM-backed env that confines the JVM to a single core. Show a window per worker and state the consequence for subprocesses.

Issue #4262, opened 2026-09-07
AsyncBatchedCollector needs a quiescent pause boundary


vmoens:
## Problem

Running the first lazy `torch.compile` call after `AsyncBatchedCollector` has started can overlap compiler-runtime initialization with the collector's coordinator and inference threads. On highly threaded collectors, this can stall compilation and collection together; stacks commonly show the main thread waiting for compiler-worker results while the collector threads are idle.

The unrelated compile does not acquire the inference server's model lock, so that lock is not the source of the cycle. The practical interaction is process-level compiler initialization and Python thread contention. Compiling and warming up before starting collection is the most robust pattern, but users also need a supported quiescent boundary when lazy compilation cannot be moved earlier.

## Proposed behavior

- Add an `AsyncBatchedCollector.pause()` context manager that finishes in-flight environment and policy work, parks all coordinator threads, and leaves the inference server without requests.
- Resume collection automatically when the context exits, including exceptional exits.
- Document compile-before-collect as the preferred pattern and the pause context as the fallback.
- Cover a threading-backend collector compiling an unrelated function while paused, and verify inference resumes afterward.

PR #4268, opened 2026-09-07
[Performance] Compile the complete DreamerV3 learner step

### LINKED ISSUE #4267: [Performance] Compile the complete DreamerV3 learner step
### Problem

The DreamerV3 SOTA training path can compile the RSSM recurrence and replay the learner update through a CUDA graph, but the surrounding encoder, decoder, distribution heads, KL terms, imagination rollout, return normalization, and replay value loss remain eager. CUDA graph replay preserves that eager kernel topology, leaving the update exposed to launch overhead.

### Proposed change

Add an opt-in optimization.compile_train_step setting that compiles the complete learner forward and backward with TorchInductor before optional CUDA graph capture. Keep optimizer and target-network updates outside capture, keep the default eager behavior unchanged, and provide profiler-backed benchmark variants for kernel count, GPU time, and CPU launch time.

The compiled path should preserve input TensorDict keys and remain covered in the existing DreamerV3 objective and CUDA-graph tests.

vmoens:
Whole-step compilation now wraps the native DreamerV3 learner forward/backward, including model, actor, value and replay-value losses. It owns shared RSSM modules and disables nested compilation. Compile and graph warm-up finish before collection starts; the existing training defaults remain unchanged.

Depends on #4265. Closes #4267.

The learner selects the same input schema for synthetic warm-up and replay batches. Model loss reconstructs nested containers before renaming reward keys, preventing compiled calls from mutating the caller's replay sample. Gradient buffers remain allocated after capture, and an optimizer step with no parameter gradients raises a clear error. Optimizer schedules and target updates remain outside capture.

Validation:
- Eight focused CPU learner/native-replay tests passed, including a real compiled update and input-key preservation.
- Both CUDA graph regressions passed on NVIDIA GB200: capture with and without whole-step compilation, production warm-up, subsequent parameter and target updates, and native replay updates.
- Repository-pinned formatting and Flake8 pass; complete outgoing diff and commit metadata audited.
- Three matched LOW-priority GPU repetitions completed, followed by three longer matched repetitions to investigate timing variation. Full results and limitations are below. The final combined CPU inference/objective/replay suite passed 904 tests; downstream real SOTA and separate-process resume tests also pass on CUDA.

Whole-step compilation is opt-in. The maintained fast preset continues to use its established compiled-scan plus CUDA-graph configuration.

Matched benchmark: NVIDIA GB200, Python 3.12, PyTorch 2.15.0.dev20260906+cu130, TensorDict public main faee822, BF16, batch 2, sequence length 8, unroll 2, CPU native replay, one CPU thread. Each comparison uses identical hardware, dependencies, seed and warm-up. Timing includes sampling, transfer, learner losses/backward, optimizer/target updates and ordered replay writeback, with final synchronization. Host percentiles exclude that final synchronization. Profiler figures come from one additional completed update.

Three repetitions, 10 warm-up and 30 measured updates; ranges across runs:

| Variant | Mean update ms | Transitions/s | p50 host ms | p95 host ms | Peak CUDA MiB | Kernels/update | Profile GPU ms |
|---|---:|---:|---:|---:|---:|---:|---:|
| eager | 134.80–141.49 | 113.08–118.69 | 134.89–141.79 | 136.74–143.85 | 62.83–62.83 | 5428.00–5428.00 | 16.15–16.28 |
| compiled_scan | 134.45–219.49 | 72.90–119.01 | 134.27–218.08 | 138.05–231.43 | 62.21–62.21 | 3849.00–3849.00 | 13.38–16.48 |
| cuda_graph | 10.75–12.33 | 1297.41–1487.94 | 11.19–12.28 | 12.40–15.77 | 22.88–22.88 | 3893.00–3893.00 | 8.09–8.18 |
| compiled_train_step | 43.55–65.53 | 244.16–367.35 | 43.79–65.35 | 44.34–69.03 | 47.44–47.48 | 1598.00–1598.00 | 9.90–13.64 |
| compiled_train_step_cuda_graph | 12.47–15.68 | 1020.23–1283.33 | 12.17–16.07 | 14.44–19.26 | 22.86–22.86 | 1607.00–1607.00 | 7.20–7.93 |

Follow-up: three repetitions with 30 warm-up and 100 measured updates, keeping all other settings identical:

| Variant | Mean update ms | Transitions/s | p50 host ms | p95 host ms | Peak CUDA MiB | Kernels/update | Profile GPU ms |
|---|---:|---:|---:|---:|---:|---:|---:|
| compiled_scan | 110.00–146.80 | 108.99–145.45 | 109.99–146.55 | 112.76–152.48 | 62.21–62.21 | 3857.00–3857.00 | 13.26–14.83 |
| cuda_graph | 10.87–11.45 | 1396.77–1472.19 | 11.41–11.82 | 12.46–17.25 | 22.88–22.88 | 3901.00–3901.00 | 7.06–8.24 |
| compiled_train_step_cuda_graph | 11.28–12.42 | 1288.41–1418.57 | 11.63–12.75 | 13.17–14.55 | 22.86–22.90 | 1615.00–1615.00 | 6.08–7.19 |

Whole-step compilation improves this uncaptured workload over eager execution, but the captured whole-step variant is 3.8–8.4% slower than compiled scan plus CUDA graphs in the paired longer runs. It uses 1,615 kernels versus 3,901 and less profiled GPU time; the end-to-end difference therefore lies outside the reduced GPU kernel time, with host/replay overhead and run variation still relevant. Longer warm-up removes the largest short-run outlier but does not establish a captured throughput gain. This is an opt-in implementation for review, not a replacement for the existing fast preset. No training default changes.

Reproduce with `OMP_NUM_THREADS=1 TORCHINDUCTOR_COMPILE_THREADS=8 python benchmarks/ad_hoc/bench_dreamer_v3_learner.py --batch 2 --steps 8 --unroll 2 --warmup 10 --iterations 30 --replay-device cpu --variants eager compiled_scan cuda_graph compiled_train_step compiled_train_step_cuda_graph`; use `--warmup 30 --iterations 100 --variants compiled_scan cuda_graph compiled_train_step_cuda_graph` for the follow-up.

Oldest-dependency CI found a hard-coded Pendulum-v1 test assumption. Tests now use the existing version-aware Pendulum helper. All three native replay collection smoke cases pass against Gym 0.13; the downstream generic SOTA suite passes eight cases against the same backend. Production defaults are unchanged.

Validation after refreshing the stack onto main with #4297 merged (2026-09-08): the feature additions and removals are unchanged. The combined CPU DreamerV3, component, native-replay and config selection passed 394 tests, including separate-process checkpoint/resume. CUDA results and benchmark measurements above precede this restack; new CPU/GPU/dependency CI must validate the refreshed heads.

### DIFF
--- a/sota-implementations/dreamer_v3/train.py
+++ b/sota-implementations/dreamer_v3/train.py
@@ -104,6 +104,16 @@ class _LearnerUpdate:
         cudagraph_warmup: int = 5,
     ):
         self.cfg = cfg
+        self._sample_keys = [
+            "action",
+            "is_init",
+            "state",
+            "belief",
+            ("next", "observation"),
+            ("next", "reward"),
+            ("next", "done"),
+            ("next", "terminated"),
+        ]
         self.device = device
         self.model_loss = learner.model_loss
         self.actor_loss = learner.actor_loss
@@ -112,10 +122,24 @@ class _LearnerUpdate:
         self.value_target_updater = learner.value_target_updater
         self.state_dim = latent_state_dim(cfg)
         self.use_bfloat16 = cfg.optimization.mixed_precision and device.type == "cuda"
+        self._warmup_steps = (
+            cudagraph_warmup
+            if cfg.optimization.cudagraph_train_step
+            else int(
+                bool(
+                    cfg.optimization.compile_train_step or cfg.optimization.compile_rssm
+                )
+            )
+        )
         self._cudagraph_warmup_remaining = (
             cudagraph_warmup if cfg.optimization.cudagraph_train_step else 0
         )
         train_step = self._forward_backward
+        if cfg.optimization.compile_train_step:
+            train_step = torch.compile(
+                train_step,
+                mode=cfg.optimization.compile_train_step_mode,
+            )
         if cfg.optimization.cudagraph_train_step:
             if device.type != "cuda":
                 raise RuntimeError(
@@ -193,7 +217,7 @@ class _LearnerUpdate:
                 + cfg.optimization.replay_value_loss_weight * replay_loss
             )
 
-        self.optimizer.zero_grad(set_to_none=True)
+        self.optimizer.zero_grad(set_to_none=False)
         total_loss.backward()
         metrics = torch.stack(
             (
@@ -215,12 +239,69 @@ class _LearnerUpdate:
         self,
         sample: TensorDictBase,
     ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        sample = sample.select(*self._sample_keys)
+        # Replay sampling may squeeze boolean feature axes. Match capture inputs.
+        for key in ("is_init", ("next", "done"), ("next", "terminated")):
+            sample.set(key, sample.get(key).reshape(*sample.batch_size, 1))
         result = self.train_step(sample)
         if self._cudagraph_warmup_remaining:
             self._cudagraph_warmup_remaining -= 1
         self.optimizer.step()
         self.value_target_updater.step()
         return result
+
+
+def _fake_learner_sample(
+    cfg: DictConfig,
+    device: torch.device,
+    obs_dim: int,
+    action_dim: int,
+) -> TensorDict:
+    batch_size = (cfg.replay_buffer.batch_size, cfg.replay_buffer.seq_len)
+    return TensorDict(
+        {
+            "state": torch.zeros(*batch_size, latent_state_dim(cfg)),
+            "belief": torch.zeros(*batch_size, cfg.networks.rnn_hidden_dim),
+            "action": torch.zeros(*batch_size, action_dim),
+            "is_init": torch.zeros(*batch_size, 1, dtype=torch.bool),
+            "next": {
+                "observation": torch.zeros(*batch_size, obs_dim),
+                "reward": torch.zeros(*batch_size, 1),
+                "done": torch.zeros(*batch_size, 1, dtype=torch.bool),
+                "terminated": torch.zeros(*batch_size, 1, dtype=torch.bool),
+            },
+        },
+        batch_size,
+        device=device,
+    )
+
+
+def _warm_up_learner(
+    cfg: DictConfig,
+    device: torch.device,
+    learner_update: _LearnerUpdate,
+    obs_dim: int,
+    action_dim: int,
+) -> None:
+    if not learner_update._warmup_steps:
+        return
+
+    cpu_rng_state = torch.get_rng_state()
+    cuda_rng_state = torch.cuda.get_rng_state(device) if device.type == "cuda" else None
+    return_low = learner_update.actor_loss.retnorm.low.detach().clone()
+    return_high = learner_update.actor_loss.retnorm.high.detach().clone()
+    sample = _fake_learner_sample(cfg, device, obs_dim, action_dim)
+    try:
+        for _ in range(learner_update._warmup_steps):
+            learner_update.train_step(sample.select(*learner_update._sample_keys))
+        learner_update._cudagraph_warmup_remaining = 0
+        learner_update.optimizer.zero_grad(set_to_none=False)
+    finally:
+        learner_update.actor_loss.retnorm.low.copy_(return_low)
+        learner_update.actor_loss.retnorm.high.copy_(return_high)
+        torch.set_rng_state(cpu_rng_state)
+        if cuda_rng_state is not None:
+            torch.cuda.set_rng_state(cuda_rng_state, device)
 
 
 def _validated_action_budget(cfg: DictConfig) -> int:
@@ -264,14 +345,22 @@ def _build_learner(
         reward_net,
         reward_decoder,
         continuation_net,
-    ) = build_world_model(cfg=cfg, obs_dim=obs_dim, action_dim=action_dim)
+    ) = build_world_model(
+        cfg=cfg,
+        obs_dim=obs_dim,
+        action_dim=action_dim,
+        compile_rollout=not cfg.optimization.compile_train_step,
+    )
     world_model = world_model.to(device)
     imagination_model = build_imagination_model(
         prior_net=prior_net,
         reward_net=reward_net,
         reward_decoder=reward_decoder,
-        # Compiling changes the categorical draws; "step" must match eager.
-        compile_prior=cfg.optimization.compile_rssm == "scan",
+        # The whole-step compile owns shared modules and subsumes RSSM compile.
+        compile_prior=(
+            cfg.optimization.compile_rssm == "scan"
+            and not cfg.optimization.compile_train_step
+        ),
     ).to(device)
     continuation_model = build_continuation_model(continuation_net=continuation_net).to(
         device
@@ -641,7 +730,8 @@ def main(cfg: DictConfig):
     use_bfloat16 = cfg.optimization.mixed_precision and device.type == "cuda"
     torchrl_logger.info(
         "DreamerV3 execution: device=%s, replay_device=%s, rssm_backend=%s, "
-        "rssm_scan_unroll=%s, mixed_precision=%s, cudagraph_train_step=%s",
+        "rssm_scan_unroll=%s, mixed_precision=%s, compile_train_step=%s, "
+        "cudagraph_train_step=%s",
         device,
         replay_device,
         cfg.optimization.compile_rssm or "eager",
@@ -651,6 +741,7 @@ def main(cfg: DictConfig):
             else "n/a"
         ),
         use_bfloat16,
+        cfg.optimization.compile_train_step,
         cfg.optimization.cudagraph_train_step,
     )
     num_envs = cfg.collector.num_envs
@@ -672,6 +763,7 @@ def main(cfg: DictConfig):
 
     learner = _build_learner(cfg, device, obs_dim, action_dim)
     learner_update = _LearnerUpdate(cfg, device, learner)
+    _warm_up_learner(cfg, device, learner_update, obs_dim, action_dim)
     rb = _build_replay(cfg, num_envs, replay_device, device)
     action_step = 0
     reset_records = num_envs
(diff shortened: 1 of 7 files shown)


sota-implementations/dreamer_v3/train.py
```diff
@@ -107,6 +107,11 @@ def __init__(
         self.state_dim = latent_state_dim(cfg)
         self.use_bfloat16 = cfg.optimization.mixed_precision and device.type == "cuda"
         train_step = self._forward_backward
+        if cfg.optimization.compile_train_step:
+            train_step = torch.compile(
+                train_step,
+                mode="max-autotune-no-cudagraphs",
```
vmoens: Root of the CI failure: `build_imagination_model(..., compile_prior=cfg.optimization.compile_rssm == "scan")` (l. 258) wraps the shared `RSSMPriorV3` in an `OptimizedModule`, and this outer compile then sees the same module bare (model loss) and as `_orig_mod` (actor loss) and asserts. With `compile_train_step` on, pass `compile_prior=False` and skip the scan compile; the outer graph covers both.

Two more points on this block: hard-coding `max-autotune-no-cudagraphs` autotunes every matmul and conv of the model at the first update (many minutes at DreamerV3 sizes) without a measured gain over `default`, so make the mode a config value; and this compile runs at the first update with the collector already live, which is the #4262 hang, so compile before the collector starts or use `collector.pause()` from #4263.

sota-implementations/dreamer_v3/train.py
```diff
@@ -698,6 +705,8 @@ def main(cfg: DictConfig):
             with timeit("dreamer_v3/replay_update"):
                 # Apply the older refresh, thus replay stays one sample ahead.
                 replay_pipeline.apply_pending_context(rb)
+            if device.type == "cuda":
+                torch.cuda.synchronize(device)
```
vmoens: Unconditional host sync on the hot path (also at l. 719-720). It removes any CPU/GPU overlap the loop has and conflicts with #4265. Use CUDA events with `elapsed_time`, or `timeit(..., sync=True)` from #4260 behind a flag, and keep the default loop asynchronous.

vmoens (commented):
Went through the diff and the CI logs. Three things need to change before this can land; the rest is minor.

**1. The two-flag combination this PR enables crashes in Dynamo (blocker, reproduced in CI).** `tests-gpu (3.12, 13.0, 3)` fails on the new `test_dreamer_v3_full_learner_cuda_graph_matches_uncaptured[True-device0]` with

```
AssertionError: UnspecializedNNModuleVariable(RSSMPriorV3) is already tracked for mutation.
Source of new object: ...actor_loss...model_based_env...world_model...module._orig_mod
Source of previously tracked object: ...model_loss...world_model...rssm_prior.module
```

Root cause: `build_imagination_model(..., compile_prior=cfg.optimization.compile_rssm == "scan")` (train.py:258, dreamer_v3_agent.py:611) wraps the shared `RSSMPriorV3` in `torch.compile`, so the outer trace of `_forward_backward` meets the same module object twice, once bare through the model loss and once as the `_orig_mod` of an `OptimizedModule` through the actor loss's imagination env, and Dynamo refuses. Every entry point this PR touches turns on exactly `compile_rssm=scan` plus `compile_train_step=true`: `reproduce_dmc_walker.sh --fast`, the benchmark's default `compiled_train_step_cuda_graph` variant, `test_dreamer_v3_dmc_reproduction_modes` and the GPU test. Fix: when `compile_train_step` is set, do not nest compiles: build the learner with `compile_prior=False` and skip the RSSM scan compile (the outer Inductor graph subsumes both), or unwrap `_orig_mod` before handing the modules to the losses. Please also add a CPU regression that compiles `_forward_backward` with `backend="eager"` under both flags: the assertion fires at trace time, so it needs no CUDA, and the current CPU smoke did not catch it.

**2. Compiling at the first update while the async collector is live deadlocks (major).** With this PR the Inductor compile happens on the first `train_step` call inside the training loop, after the collector threads, the inference-server thread and the env workers are running. That is the configuration that hung in #4262 (main thread in `futex_wait` inside Dynamo's `run_node`, 139 idle threads). #4263 landed today with `AsyncBatchedCollector.pause()` for this purpose. Either compile and warm up on a spec-shaped fake batch before the collector starts, or wrap the first compiled calls in `with collector.pause():`. At minimum the README paragraph for the new option has to say so; today `collector.backend=async` users will hit the hang.

**3. Unconditional `torch.cuda.synchronize(device)` in the training loop (major).** train.py:708-709 and 719-720 add a host sync before and after every update, not behind a timing flag. This serialises the CPU with the GPU on the hot path: whatever the loop overlaps today (with a GPU-resident replay the latent write-back and the next sample are asynchronous) now waits for the graph replay to finish, and it directly contradicts #4265, whose purpose is that overlap. For an accurate `train/time_dreamer_v3_train_update`, record `torch.cuda.Event(enable_timing=True)` pairs on the stream and read `elapsed_time` at log time, or use `timeit(..., sync=True)` from #4260 behind an explicit `optimization.sync_timers` flag. Keep the loop asynchronous by default.

Minor:
- `mode="max-autotune-no-cudagraphs"` is hard-coded (train.py:113). Autotuning every matmul and conv of a 171M-parameter model takes many minutes and the gain over `default` is not measured here; make the mode configurable (or default to `default`) and let users opt into autotune.
- `DreamerV3ModelLoss.forward` (torchrl/objectives/dreamer_v3.py:414-417): the `select(*leaves, as_tensordict=True)` idiom works (available since tensordict 0.12) but reads as odd without context. If `TensorDict.copy()` aliases its input under Dynamo, that is a tensordict bug: please open an issue there, link it in a comment above this line, and keep the workaround minimal. The eager-backend regression test is a good guard.
- Flipping the maintained fast reproduction mode (reproduce_dmc_walker.sh:86) to an option whose CUDA behaviour has not been run, per the PR description, changes wall-clock and possibly numerics for everyone using the script. Keep the reproduction mode on the measured path until the benchmark shows a gain, then flip it in a follow-up with the numbers.
- The profiler summary in the benchmark is useful. Once this runs on a CUDA host, please post kernel count / GPU time / launch time for `cuda_graph` vs `compiled_train_step_cuda_graph` in the PR so the README claim has a number behind it.

Also note #4265 edits the same `_LearnerUpdate` and training-loop region and the same benchmark file; the two will conflict textually, and semantically on the synchronisation point (item 3), so it is worth deciding their order now.

sota-implementations/dreamer_v3/reproduce_dmc_walker.sh
```diff
@@ -83,6 +83,7 @@ elif [ "$fast" -eq 1 ]; then
   benchmark_command+=(
     optimization.compile_rssm=scan
     optimization.rssm_scan_unroll=8
+    optimization.compile_train_step=true
```
vmoens: With the current code this combination (`compile_rssm=scan` + `compile_train_step=true`) is the one that crashes in Dynamo (see the review body). Independently of the fix, I would keep the maintained fast mode on the measured path until the benchmark shows a gain on a CUDA host.

torchrl/objectives/dreamer_v3.py
```diff
@@ -411,7 +411,10 @@ def _forward_value_estimator_keys(self, **kwargs) -> None:
 
     @_maybe_record_function_decorator("dreamer_v3/world_model_loss")
     def forward(self, tensordict: TensorDict) -> tuple[TensorDict, TensorDict]:
-        tensordict = tensordict.copy()
+        tensordict = tensordict.select(
```
vmoens: If `copy()` aliases the input under Dynamo, that is a tensordict bug worth an issue; link it in a comment here so the idiom is not simplified back to `copy()` later.

vmoens:
Addressed the review in 96dec82db.

- Whole-step compilation now owns the shared RSSM modules and suppresses nested rollout/prior compilation. The new CPU regression explicitly requests both compile_rssm=scan and compile_train_step=true, and a real Inductor smoke passes.
- Compilation and CUDA-graph warmup run on a fixed synthetic batch before collector construction.
- CUDA synchronization is opt-in through the merged timeit sync support, so the default path preserves CPU/GPU overlap.
- The compile mode is configurable and defaults to default.
- The maintained fast reproduction preset remains on its previously measured scan plus CUDA-graph path.
- The TensorDict aliasing assessment was correct: issue https://github.com/pytorch/tensordict/issues/1783 tracks it, and the independent fix is https://github.com/pytorch/tensordict/pull/1784.

I also merged current main so this uses the collector-pause and synchronized-timer infrastructure from #4263 and #4260. The profiler benchmark remains available, but CUDA measurements still need a CUDA host.

sota-implementations/dreamer_v3/reproduce_dmc_walker.sh
vmoens: Agreed. I restored the maintained fast reproduction preset to its previously measured scan plus CUDA-graph configuration. Whole-step compilation remains an explicit opt-in until CUDA measurements support changing the preset.

sota-implementations/dreamer_v3/train.py
vmoens: Agreed. The unconditional synchronizations are removed. Train-update timing now passes the configurable sync flag to timeit, with synchronization disabled by default so CPU and GPU work can overlap.

sota-implementations/dreamer_v3/train.py
vmoens: Agreed and fixed in 96dec82db. Whole-step compilation now suppresses nested rollout and prior compilation, and the CPU regression explicitly enables both scan RSSM compilation and whole-step compilation. A real CPU Inductor smoke also passes.

torchrl/objectives/dreamer_v3.py
vmoens: Confirmed as a TensorDict bug on current main. I opened https://github.com/pytorch/tensordict/issues/1783 and the independent fix at https://github.com/pytorch/tensordict/pull/1784. This PR retains the leaf-selection workaround and links the upstream issue.

vmoens:
TensorDict #1784 is merged. Cleanup commit 3e9d63d1a removes the temporary leaf-selection workaround and restores the normal shallow copy path. The full non-GPU DreamerV3 suite and a real CPU Inductor smoke both pass against the exact TensorDict merge commit faee822.

PR #4269, opened 2026-09-07
[Feature] CUDA-graph static inference batches

vmoens:
Static inference batches now capture after initial weight synchronization, retain returned values across later calls, and enforce in-place parameter/buffer updates. Replacing captured storage raises clearly and falls back to eager inference. Partial batches repeat the last request and discard padded outputs.

The collector prepares capture before starting environment coordination; process servers prepare it inside their owning process.

GPU validation found and fixed a capture failure from lazy collation. Static inputs are now materialized before capture, and padding uses batched tensor operations. This also removes a reproducible batch-1 regression caused by cloning requests individually.

Validation:
- Full CPU inference suite after conflict resolution and the input-padding follow-up: 129 passed, 14 skipped, 13 GPU cases deselected. The resolution preserves the previously validated inference library code and capture-before-start ordering.
- Earlier full GPU inference suite: 12 passed, 1 Ray test skipped because Ray is unavailable. Coverage includes retained outputs, stochastic sampling, initial synchronization, in-place updates, storage replacement, and process-server capture.
- `ci/olddeps` enabled before push. Formatting, lint and diff checks pass.

Earlier inference-only benchmark, before the input-padding follow-up: three runs on one NVIDIA GB200, public PyTorch 2.15.0.dev20260906+cu130 and TensorDict 0.14 main (faee822), 170,839,912-parameter policy, identical seeds, 10 warm-up batches and 30 measured batches per case. CUDA graph size is 64 in every graph case. Capture is excluded; round trips include synchronized CPU results. This measures inference serving, not an end-to-end learner.

| Batch | Mode | Requests/s range | Median round-trip ms range | p95 round-trip ms range | Peak CUDA allocation MiB |
|---|---|---:|---:|---:|---:|
| 1 | eager | 171.58–183.98 | 5.36–5.75 | 6.00–6.43 | 921.50–921.50 |
| 1 | cuda_graph | 538.48–577.29 | 1.72–1.79 | 1.88–1.96 | 938.58–938.58 |
| 16 | eager | 1248.55–1303.08 | 11.44–12.03 | 18.52–19.02 | 925.76–925.76 |
| 16 | cuda_graph | 3465.92–3587.45 | 4.32–4.49 | 4.66–6.51 | 942.62–942.62 |
| 64 | eager | 1647.64–1692.65 | 32.35–34.10 | 35.93–36.56 | 939.42–939.42 |
| 64 | cuda_graph | 5291.55–5431.54 | 11.25–11.58 | 13.46–14.27 | 942.62–946.66 |

## Continuous collection benchmark

The main baseline from merged #4287 is now [published](https://docs.pytorch.org/rl/dev/bench/async/) at bedcc8332. Its [matched baseline](https://github.com/pytorch/rl/actions/runs/34157720247) and [isolated comparison including this PR](https://github.com/pytorch/rl/actions/runs/34157722217) completed three CPU/GPU measurement repeats successfully. These runs share the same source base, pinned dependencies, system packages, workload and warm-up settings. GPU instances use the same NVIDIA A10G / AMD EPYC 7R32 runner class and driver 580.82.07.

This end-to-end pixel workload uses 32 CPU environments, 3x84x84 observations, a CNN plus two 1024-wide layers, 1 ms environment steps, 256 frames per output batch and 1,024 frames per measured round. Each fresh-process repeat has two warm-up rounds and five measured rounds. Static batches pad to 32. One regime resets uniformly; the other stalls one environment's reset for 200 ms every 16 steps.

| Integrated collection series | Baseline eager frames/s [run range] | Graph frames/s [run range] | Median change |
| --- | ---: | ---: | ---: |
| Uniform | 828.9 [828.0, 831.0] | 720.6 [700.0, 738.5] | -13.1% |
| Slow reset | 821.9 [817.0, 829.8] | 718.4 [714.6, 719.1] | -12.6% |

The fixed eager controls overlap across revisions. Within the comparison revision, graph serving improves median forward time from about 8.3 ms to 6.6 ms, but complete collection slows and uniform batch p95 increases from about 410 ms to 477 ms. The forward statistic excludes collation/padding and input-device transfer. Batches average about 14 real requests, so preparing fixed batches of 32 is a profiling target; these data do not establish a single cause. The inference-only gains above do not establish an end-to-end gain for this workload.

The first graph case peaks at 44.68 MiB of CUDA allocation versus 32.12 MiB for eager. Later graph cases accumulate about 8.13 MiB each in the same process, consistently across repeats, despite GC before each case. Their absolute peaks include retained allocation and must not be read as isolated collector costs. The increment matches the [pinned cuBLAS workspace size](https://github.com/pytorch/pytorch/blob/v2.14.0/aten/src/ATen/cuda/CublasHandlePool.cpp), cached per handle/stream; new warm-up streams make workspace caching a likely explanation, pending allocation tracing. Raw timings, latency, memory and server statistics are attached to the runs.

### Input-padding follow-up

The core server now transfers real requests to the policy device before padding to graph capacity. Returned values still retain their own storage. The [focused GPU validation](https://github.com/pytorch/rl/actions/runs/34160556698) passed all eight selected graph regressions and three benchmark repetitions with matching dependency/system-package lists and GPU model/driver.

| Fixed static series | Before frames/s [run range] | CUDA padding frames/s [run range] |
| --- | ---: | ---: |
| Uniform | 711.9 [710.7, 716.2] | 744.3 [736.1, 744.8] |
| Slow reset | 723.5 [723.0, 724.0] | 739.4 [735.1, 764.5] |

The focused eager controls are 814.3 [803.9, 833.6] and 821.0 [811.1, 830.2] frames/s. This recovers part of the graph cost (+4.6% uniform, +2.2% slow-reset), while graph collection remains about 9–10% slower than eager in the same focused run. Batch p95 is 446.8/467.2 ms for graph versus 414.3/407.9 ms for eager. Median server forward times are 6.38/6.39 ms versus 8.45/8.01 ms. This workload still does not establish an end-to-end graph win.

Within the focused run, eager uniform peaks at 24.00 MiB of CUDA allocation and the first static case at 37.83 MiB. Final process-tree RSS is 23,720/23,745 MiB respectively, with shared pages counted per process. The focused run selects fewer cases, so its process-wide CUDA peaks have a different preceding workspace history and are not a like-for-like memory comparison with the full run. Keep the full-run memory caveat above.

### DIFF
--- a/test/test_inference_server.py
+++ b/test/test_inference_server.py
@@ -2341,6 +2341,42 @@ class TestAsyncBatchedCollector:
         collector.shutdown()
         assert collected >= 50
 
+    @pytest.mark.parametrize(
+        ("env_backend", "env_exchange"),
+        [("threading", "queue"), ("multiprocessing", "shm")],
+    )
+    def test_pause_for_torch_compile(self, env_backend, env_exchange):
+        """Compilation can run while a started collector is quiescent."""
+        collector = AsyncBatchedCollector(
+            create_env_fn=[_counting_env_factory] * 4,
+            policy=_make_counting_policy(),
+            frames_per_batch=10,
+            total_frames=-1,
+            env_backend=env_backend,
+            env_exchange=env_exchange,
+        )
+        iterator = iter(collector)
+        next(iterator)
+
+        def unrelated_module(value):
+            return value.sin().cos()
+
+        try:
+            with collector.pause():
+                paused_requests = collector.server_stats()["requests"]
+                compiled_module = torch.compile(unrelated_module, backend="eager")
+                value = torch.randn(8)
+                torch.testing.assert_close(
+                    compiled_module(value), unrelated_module(value)
+                )
+                time.sleep(0.05)
+                assert collector.server_stats()["requests"] == paused_requests
+
+            next(iterator)
+            assert collector.server_stats()["requests"] > paused_requests
+        finally:
+            collector.shutdown()
+
     def test_num_envs(self):
         """The collector knows the number of environments."""
         policy = _make_counting_policy()
--- a/test/test_inference_server.py
+++ b/test/test_inference_server.py
@@ -5,8 +5,10 @@
 from __future__ import annotations
 
 import concurrent.futures
+import functools as ft
 import importlib.util
 import multiprocessing as mp
+import os
 import pickle
 import queue
 import threading
@@ -2443,11 +2445,11 @@ class TestProcessInferenceServer:
 class TestAsyncBatchedCollector:
     """Tests for :class:`AsyncBatchedCollector`."""
 
-    def test_basic_collection(self):
+    @pytest.mark.parametrize("total_frames", [60, 61])
+    def test_basic_collection(self, total_frames):
         """Collector yields at least frames_per_batch frames."""
         num_envs = 3
         frames_per_batch = 20
-        total_frames = 60
         policy = _make_counting_policy()
 
         collector = AsyncBatchedCollector(
@@ -2464,7 +2466,7 @@ class TestAsyncBatchedCollector:
             total_collected += batch.numel()
         stats = collector.server_stats()
         collector.shutdown()
-        assert total_collected >= total_frames
+        assert total_collected == total_frames
         assert stats["requests"] > 0
 
     def test_policy_factory(self):
@@ -2503,20 +2505,25 @@ class TestAsyncBatchedCollector:
                 frames_per_batch=10,
             )
 
-    def test_yield_completed_trajectories(self):
+    @pytest.mark.parametrize(
+        ("env_backend", "env_exchange"),
+        [("threading", "queue"), ("multiprocessing", "shm")],
+    )
+    def test_yield_completed_trajectories(self, env_backend, env_exchange):
         """With yield_completed_trajectories, collector yields done trajectories."""
         num_envs = 3
         max_steps = 5
         policy = _make_counting_policy()
 
         collector = AsyncBatchedCollector(
-            create_env_fn=[lambda: CountingEnv(max_steps=max_steps)] * num_envs,
+            create_env_fn=[ft.partial(CountingEnv, max_steps=max_steps)] * num_envs,
             policy=policy,
             frames_per_batch=1,
             total_frames=30,
             yield_completed_trajectories=True,
             max_batch_size=num_envs,
-            env_backend="threading",
+            env_backend=env_backend,
+            env_exchange=env_exchange,
         )
         count = 0
         for batch in collector:
@@ -2526,7 +2533,11 @@ class TestAsyncBatchedCollector:
         collector.shutdown()
         assert count >= 30
 
-    def test_shutdown_idempotent(self):
+    @pytest.mark.parametrize(
+        ("env_backend", "env_exchange"),
+        [("threading", "queue"), ("multiprocessing", "shm")],
+    )
+    def test_shutdown_idempotent(self, env_backend, env_exchange):
         """Calling shutdown twice should not raise."""
         policy = _make_counting_policy()
         collector = AsyncBatchedCollector(
@@ -2534,7 +2545,8 @@ class TestAsyncBatchedCollector:
             policy=policy,
             frames_per_batch=10,
             total_frames=10,
-            env_backend="threading",
+            env_backend=env_backend,
+            env_exchange=env_exchange,
         )
         # Consume one batch to start
         for _batch in collector:
@@ -2559,6 +2571,42 @@ class TestAsyncBatchedCollector:
                 break
         collector.shutdown()
         assert collected >= 50
+
+    @pytest.mark.parametrize(
+        ("env_backend", "env_exchange"),
+        [("threading", "queue"), ("multiprocessing", "shm")],
+    )
+    def test_pause_for_torch_compile(self, env_backend, env_exchange):
+        """Compilation can run while a started collector is quiescent."""
+        collector = AsyncBatchedCollector(
+            create_env_fn=[_counting_env_factory] * 4,
+            policy=_make_counting_policy(),
+            frames_per_batch=10,
+            total_frames=-1,
+            env_backend=env_backend,
+            env_exchange=env_exchange,
+        )
+        iterator = iter(collector)
+        next(iterator)
+
+        def unrelated_module(value):
+            return value.sin().cos()
+
+        try:
+            with collector.pause():
+                paused_requests = collector.server_stats()["requests"]
+                compiled_module = torch.compile(unrelated_module, backend="eager")
+                value = torch.randn(8)
+                torch.testing.assert_close(
+                    compiled_module(value), unrelated_module(value)
+                )
+                time.sleep(0.05)
+                assert collector.server_stats()["requests"] == paused_requests
+
+            next(iterator)
+            assert collector.server_stats()["requests"] > paused_requests
+        finally:
+            collector.shutdown()
 
     def test_num_envs(self):
         """The collector knows the number of environments."""
@@ -2594,9 +2642,16 @@ class TestAsyncBatchedCollector:
         collector.shutdown()
         assert called["count"] >= 1
 
-    @pytest.mark.parametrize("env_backend", ["threading", "multiprocessing"])
-    def test_env_backend_smoke(self, env_backend):
-        """Thread and multiprocessing env backends collect data."""
+    @pytest.mark.parametrize(
+        ("env_backend", "env_exchange"),
+        [
+            ("threading", "queue"),
+            ("threading", "auto"),
+            ("multiprocessing", "queue"),
+        ],
+    )
+    def test_env_backend_smoke(self, env_backend, env_exchange):
+        """Supported environment backend and exchange pairs collect data."""
         collector = AsyncBatchedCollector(
             create_env_fn=[_counting_env_factory] * 2,
             policy=_make_counting_policy(),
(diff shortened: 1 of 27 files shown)


torchrl/modules/inference_server/_server.py
```diff
@@ -624,6 +659,13 @@ def start(self) -> InferenceServer:
         """
         if self._worker is not None and self._worker.is_alive():
             raise RuntimeError("Server is already running.")
+        if self.static_batch_size is not None and self._cudagraph_model is None:
+            if self._cudagraph_request_spec is None:
+                raise RuntimeError(
+                    "static_batch_size requires request_spec so the CUDA graph "
+                    "can be captured before the serve loop starts."
+                )
+            self._prepare_cudagraph(self._cudagraph_request_spec)
```
vmoens: Capture happens here, but `_init_weight_sync()` only runs in `_run` (l. 870). The MP/shared-memory/distributed schemes apply the initial weights on connect with `apply_weights(..., inplace=False)`, which swaps parameter storage, and nothing recaptures afterwards (`_poll_weight_update` returns early for shared memory; `update_policy_weights_` only bumps the version), so the graph keeps reading the pre-swap storage for the whole run. Connect before capturing, or recapture from `update_policy_weights_`, or refuse `static_batch_size` with storage-swapping schemes.

torchrl/modules/inference_server/_server.py
```diff
@@ -798,16 +914,32 @@ def _run(self) -> None:
...
+                                "type used during capture."
+                            )
                         with interaction_context:
-                            result_batch = self.model(batch)
+                            if self._cudagraph_model is None:
+                                result_batch = self.model(batch)
+                            else:
+                                result_batch = self._cudagraph_model(batch)
+                        if self.static_batch_size is not None:
+                            result_batch = result_batch[:real_batch_size]
```
vmoens, line 1041: `CudaGraphModule` returns the incoming tensordict updated with its static output tensors (no clone), and `.to(self.output_device)` below is a no-op when it resolves to the policy device, so `unbind(0)` hands views of the static buffers to the futures and the next replay overwrites them while coordinator threads may still be reading. Clone the sliced batch on the graphed path.

torchrl/modules/inference_server/_server.py
```diff
@@ -798,16 +914,32 @@ def _run(self) -> None:
...
-                        )
+                        (
+                            interaction_context,
+                            batch,
+                            interaction_code,
+                        ) = self._interaction_type_context(batch)
+                        if (
+                            self._cudagraph_model is not None
+                            and interaction_code != self._cudagraph_interaction_code
+                        ):
```
vmoens, line 1018: `interaction_type()` is process-global in tensordict, and capture ran with no interaction key, so any request issued under `set_exploration_type(...)` hits this raise even though the capture itself ran under the same ambient type (reproduced on CPU with `PolicyClientModule` and `static_batch_size=4`). Compare effective types resolved at capture and replay.

test/test_inference_server.py
```diff
@@ -2444,7 +2554,7 @@ def test_device_config_and_server_config(self):
             policy=_make_counting_policy(),
             frames_per_batch=10,
             total_frames=20,
-            server_config=InferenceServerConfig(max_batch_size=2),
+            server_config=InferenceServerConfig(max_batch_size=2, static_batch_size=2),
```
vmoens: With a CPU policy this passes only on CUDA-less hosts: on a CUDA host tensordict builds `torch.cuda.Stream(device)` with `device=cpu` and raises `ValueError: Expected a cuda device`. Keep the eager version here and add `gpu`-marked static variants; also reject a non-CUDA `policy_device` with `static_batch_size` in `__init__`.

vmoens (commented):
Reviewed 3ac9efd45 in depth (full non-GPU inference-server suite locally: 115 passed, 13 skipped; the pinned lint tools; CPU probes of the padding, signature and aliasing logic; the CI lint log) and re-checked everything below against d8c8a2850, which fixes the three flake8-bugbear B023 errors that failed `python-source-and-configs`. Wiring, config parity, docs and the eager default path are in order: `static_batch_size=None` short-circuits the padding, the model is called directly and nothing is sliced. The padding itself is right (pad rows clone the last real request, the slice plus the `len(results) != len(callbacks)` guard keep padded rows from ever reaching a slot). The model-update story and the output ownership have real holes, so this needs changes before it lands. Line numbers are for d8c8a2850.

**1. The graph is captured before weight sync connects, and connect-time storage swaps are never recaptured.** `start()` captures at torchrl/modules/inference_server/_server.py:668, before the worker thread exists, but `_init_weight_sync()` only runs inside `_run` (line 870). The receiver-side `connect()` of the MP, shared-memory and distributed schemes applies the initial weights with `apply_weights(..., inplace=False)`, which goes through `to_module` and replaces parameter storage (checked on CPU: `data_ptr` changes). Nothing recaptures afterwards: `_poll_weight_update` only checks after a non-`None` `receive()` and returns early for shared memory, and `update_policy_weights_` only bumps the version. Scenario: `AsyncBatchedCollector(policy_factory=..., server_config=InferenceServerConfig(static_batch_size=N), weight_sync=SharedMemWeightSyncScheme(...))` captures on the factory's random init, `connect()` swaps the parameters to the shared buffers, the trainer keeps writing into those buffers, the module is correct, and the captured kernels read the pre-swap storage for the whole run. The Ray actor gets this right (it recaptures after `scheme.connect`); the plain server does not. Fix: run the weight-sync connect (or at least the signature check plus recapture) before `_prepare_cudagraph`, recapture from `update_policy_weights_` as well, or refuse `static_batch_size` with schemes that swap storage. The same applies to the documented `TensorDict.from_module(learner).to_module(behavior)` idiom, which swaps storage by default; state in the docs that only in-place copies are valid for a captured server.

**2. The storage signature goes blind after the first swap.** `_model_storage_signature` (lines 717-734) iterates `named_parameters()`/`named_buffers()`. The Ray path applies `weights.to_module(preserve_module_state=False)`, after which the tensors leave `_parameters` and the signature is the empty tuple forever (probe: first push detected with `len(signature) == 0`, second push undetected while the eager output moves from 2.0 to 3.0). Every Ray push after the first updates the module without recapturing. Fix: apply in place on the Ray path (`to_module(inplace=True)`, which also removes a recapture per push), and build the signature from tensor references taken at capture time rather than from `named_parameters()`.

**3. The interaction-type guard rejects every request issued under `set_exploration_type`.** Capture uses `request_spec` with no interaction key, so `_cudagraph_interaction_code` is `None`; `PolicyClientModule` attaches the caller's ambient type, and tensordict's `interaction_type()` is process-global, not thread-local. Reproduced on CPU: a `PolicyClientModule` call under `set_exploration_type(ExplorationType.RANDOM)` succeeds with `static_batch_size=None` and raises `CUDA-graphed inference requires the interaction type used during capture` with `static_batch_size=4` (lines 926-935). The tutorial idiom `with set_exploration_type(RANDOM): for batch in collector:` fails on every batch, and since the capture ran in the same process under the same ambient type, the rejection is a false positive. Compare effective types (resolve the sentinel at capture and replay) and document that the exploration type is frozen at capture.

**4. Resolved results alias the graph's static output buffers.** `CudaGraphModule` returns the incoming tensordict updated with its static output tensors, no clone. The server slices (line 942) and copies only when `output_device` is set; it defaults to `None` and falls back to `env_device`, so with CUDA envs the `.to` is a no-op and `result_batch.unbind(0)` hands views of the static buffers to the futures (checked on CPU for both collates: the unbound row follows an in-place overwrite). With `ThreadingTransport` the serve loop drains the next batch and replays the graph while coordinator threads may still be reading their action. Eager mode allocates fresh outputs per batch, so this is new. Clone the sliced batch on the graphed path, and add a test that holds the first result across a second request.

**5. Recapture runs mid-collection.** `_recapture_cudagraph_if_needed()` is called from `update_model` and from the serve loop with coordinator and env threads live. tensordict captures with the default `capture_error_mode="global"`, which errors on CUDA calls from other threads during capture; this is the situation #4262 tells users to avoid, and the rst advertises it as a feature. Prefer to keep only in-place updates valid for a captured server (raise, or warn and fall back to eager when the signature changes) instead of silently recapturing.

**6. Tests.** (a) The modified collector tests (test/test_inference_server.py:2490 and :2557) pass `static_batch_size=2` with a CPU policy. On a CUDA-less host `CudaGraphModule` is a pass-through, so they pass; on any CUDA host tensordict builds `torch.cuda.Stream(device)` with `device=cpu` and raises `ValueError: Expected a cuda device`, and the process smoke test fails at capture with "The input tensordict device must be of the 'cuda' type". They are correctly not `gpu`-marked, so CI never sees it, but every GPU dev box running the CPU suite will. Keep the eager versions, add `gpu`-marked static variants, and reject a non-CUDA `policy_device` with `static_batch_size` up front. (b) `test_static_batch_pads_and_slices_results` picks `cuda:0` when available but has no `gpu` marker, so `-m gpu` deselects it on GPU runners and it only ever runs on CPU: the real replay path is never executed in CI. Parametrize the device with `pytest.param("cuda", marks=pytest.mark.gpu)`. (c) Nothing covers items 1 to 4.

Smaller points:
- Replay silently tolerates a request/spec key mismatch: `update_` ignores keys absent from the static batch and leaves keys absent from the request at their capture-time values, so a hand-written `request_spec` that omits a policy input makes the policy read zeros with no error. A one-time key-set check on the first batch would make this loud.
- Docs overclaim on randomness: the default CUDA generator is registered automatically, a custom `torch.Generator` is not (`CudaGraphModule` never calls `register_generator_state`) and would repeat samples; the DreamerV3 served policy's per-call `fork_rng` plus `manual_seed` cannot be captured at all (`manual_seed` asserts not-capturing). One sentence in the docs, and a note that `separate_policy_rng` and `static_batch_size` are mutually exclusive.
- The benchmark submits raw requests with no interaction key, so it measures the policy's default interaction type rather than the sampling path this is about; run it through `PolicyClientModule` under `set_interaction_type(RANDOM)`.
- `_factory.py`'s pairing check changed from `(request_spec is None) != (response_spec is None)` to accepting `request_spec` alone; fine, but say so in the description (and #4272 edits the same lines).
- Validation of `static_batch_size` is duplicated in `InferenceServerConfig.__post_init__` and `InferenceServer.__init__` while `ProcessInferenceServer` has none; the collector reaches into `self._server._prepare_cudagraph` on three server classes with three different meanings, a small public hook would be cleaner.

Interaction with the sibling PRs: #4264 edits the import block and `_default_collate` next to this PR's changes and rewrites the `AsyncEnvPool(...)` call right above the `_prepare_cudagraph` insertion in `_ensure_started`, so expect adjacent-hunk conflicts, semantically compatible. #4272 conflicts textually in `_factory.py` and in the `request_spec` docstring, and semantically in `_ensure_started`: its process-env branch starts the server before any env pool exists, so the spec derivation from `self._env_pool.fake_tensordict()[0]` is never reached and `static_batch_size` with process env workers fails at `start()` with "static_batch_size requires request_spec".

torchrl/modules/inference_server/_server.py
vmoens: Fixed in 5e6b53b8d. CUDA-graph preparation now initializes weight synchronization before capture, so a storage-swapping initial connect cannot invalidate the graph immediately. Later storage replacement disables the graph and raises instead of recapturing while serving; a GPU regression covers the initial-sync ordering.

torchrl/modules/inference_server/_server.py
vmoens: Fixed in 5e6b53b8d. Graphed results are sliced and placed first, then cloned before fulfilling the future, so callers own their result storage. The GPU regression retains the first result across a second replay and verifies that it does not change.

torchrl/modules/inference_server/_server.py
vmoens: Fixed in 5e6b53b8d. Missing or sentinel request metadata now resolves to the effective ambient interaction type before graph validation, rather than being compared as None. A GPU test exercises PolicyClientModule inside set_exploration_type contexts.

test/test_inference_server.py
vmoens: Fixed in 5e6b53b8d. A static batch now requires an explicit CUDA policy_device. The existing CPU collector tests remain eager, and separate GPU-marked thread and process collector tests cover static capture. All CUDA-only tests carry both the gpu marker and the CUDA availability guard.

vmoens:
I also addressed the review-wide follow-ups in 5e6b53b8d: request_spec is accepted independently of static batching, static-size validation is centralized, the pre-start capture hook is public, the benchmark submits through PolicyClientModule under RANDOM interaction, and the docs spell out default/custom generator behavior plus the DreamerV3 separate_policy_rng incompatibility. I agree with the sibling-PR conflict warning as a coordination concern, but it does not require a code change until either overlapping PR lands.

vmoens:
Follow-up CI fix: both GPU shard-2 jobs found the same test-fixture issue in test_static_batch_pads_slices_and_owns_results. _BatchSizeModule used value.new_tensor(value.shape[0]), which attempts an unpinned host-to-device copy during graph capture. Commit a559878f5 uses direct scalar addition instead. All other new CUDA-graph regressions passed on both current and stable PyTorch; replacement CI is running.

PR #4270, opened 2026-09-07
[Feature] Support background replay collection in AsyncBatchedCollector

vmoens:
AsyncBatchedCollector can now collect directly into replay while the caller trains independently. `start()` runs the existing iterator on a dedicated writer thread; ordinary iteration retains synchronous replay writes and yields `None`.

The writer consumes owned transitions through the existing bounded collector queues. Queue and shared-memory exchange, grouped environments, process-slot workers, replay routing, post-processing, and post-collect hooks retain their existing collection behavior. Frame batches obey the exact remaining frame budget, including a partial final batch. Background hooks run on the writer thread.

`pause()` also waits for replay writes to finish. `async_shutdown()` interrupts blocked rollout reads, joins the writer, and reports background errors. A user hook or storage operation that outlives the shutdown timeout is reported and its thread remains tracked for a subsequent join. Foreground iteration and background collection cannot consume the same collector concurrently.

The benchmark adds `--replay-mode iterator` and `--replay-mode background` for `async-env-mp`. Both use identical replay storage, frame budgets, and warm-up counts; background mode does not report consumer-side batch latency.

Validation: the CPU AsyncBatchedCollector suite passed (74 passed, 3 Linux-only CPU-affinity cases skipped on macOS), using current TensorDict main. Coverage includes both replay modes across thread/process workers and grouped environments, exact frame counts, ListStorage ownership after exchange-slot reuse, pause, worker failure, blocked-hook shutdown, and background error propagation. Formatting, lint, and docstring-argument checks passed. Iterator and background benchmark smoke runs each stored exactly 32 measured frames after one 16-frame warm-up batch; this validates the benchmark workload, not a performance claim.

Rebased onto current origin/main. GPU and Linux-only validation remain with CI.

### DIFF
--- a/benchmarks/bench_collectors.py
+++ b/benchmarks/bench_collectors.py
@@ -8,6 +8,8 @@ CUDA hosts, ``--mujoco-gl egl``.
 Examples:
     python benchmarks/bench_collectors.py --total-frames 2000
     python benchmarks/bench_collectors.py --num-envs 1,2,4,8 --policy-delay-ms 20
+    python benchmarks/bench_collectors.py --backends async-env-mp --replay-mode iterator
+    python benchmarks/bench_collectors.py --backends async-env-mp --replay-mode background
     python benchmarks/bench_collectors.py --num-envs 64 \
         --backends async-env-mp --env-exchange shm \
         --env-step-latency-ms 50 --policy-hidden-features 9216 \
@@ -35,7 +37,14 @@ from tensordict import TensorDict
 from tensordict.nn import TensorDictModule
 
 from torchrl.collectors import AsyncBatchedCollector, Collector
-from torchrl.data import Bounded, Categorical, Composite, Unbounded
+from torchrl.data import (
+    Bounded,
+    Categorical,
+    Composite,
+    LazyTensorStorage,
+    ReplayBuffer,
+    Unbounded,
+)
 from torchrl.envs import (
     EnvBase,
     GymEnv,
@@ -73,8 +82,8 @@ class BenchmarkResult:
     status: str
     envs_per_worker: int = 1
     warmup_batches: int = 0
-    p50_batch_ms: float = 0.0
-    p95_batch_ms: float = 0.0
+    p50_batch_ms: float | None = 0.0
+    p95_batch_ms: float | None = 0.0
     host_rss_mib: float = 0.0
     cuda_used_mib: float = 0.0
     frames: int = 0
@@ -343,6 +352,7 @@ def bench(
     policy_delay_ms: float,
     warmup_batches: int,
     envs_per_worker: int = 1,
+    replay_mode: str = "none",
 ) -> BenchmarkResult:
     collector = None
     total = 0
@@ -350,21 +360,42 @@ def bench(
     policy_stats: dict[str, float | int] = {}
     try:
         collector = factory()
+        if replay_mode != "none":
+            # Both modes store the same number of transitions in the same storage.
+            collector.total_frames = total_frames + warmup_batches * frames_per_batch
+            collector.replay_buffer = ReplayBuffer(
+                storage=LazyTensorStorage(collector.total_frames)
+            )
         iterator = iter(collector)
         for _ in range(warmup_batches):
             next(iterator)
         if hasattr(collector, "server_stats"):
             collector.server_stats(reset=True)
         latencies = []
+        initial_frames = collector._frames
         t0 = previous = time_module.perf_counter()
-        for batch in iterator:
-            now = time_module.perf_counter()
-            latencies.append((now - previous) * 1000)
-            previous = now
-            n = batch.numel()
-            total += n
-            if total >= total_frames:
-                break
+        if replay_mode == "background":
+            iterator.close()
+            collector.start()
+            collector._replay_thread.join(timeout=300)
+            if collector._replay_thread.is_alive():
+                raise TimeoutError("Background replay benchmark exceeded 300 seconds.")
+            total = collector.replay_buffer.write_count - initial_frames
+            if collector._replay_error is not None:
+                raise RuntimeError(
+                    "Background replay benchmark failed."
+                ) from collector._replay_error
+        else:
+            for batch in iterator:
+                now = time_module.perf_counter()
+                latencies.append((now - previous) * 1000)
+                previous = now
+                if batch is None:
+                    total = collector.replay_buffer.write_count - initial_frames
+                else:
+                    total += batch.numel()
+                if total >= total_frames:
+                    break
         elapsed = time_module.perf_counter() - t0
         if hasattr(collector, "server_stats"):
             policy_stats = collector.server_stats()
@@ -379,7 +410,12 @@ def bench(
         if torch.device(policy_device).type == "cuda":
             free, capacity = torch.cuda.mem_get_info(policy_device)
             cuda_used = capacity - free
-        percentiles = torch.tensor(latencies).quantile(torch.tensor([0.5, 0.95]))
+        # Background collection has no consumer-side batch latency to report.
+        percentiles = (
+            torch.tensor(latencies).quantile(torch.tensor([0.5, 0.95]))
+            if latencies
+            else None
+        )
         fps = total / elapsed if elapsed > 0 else 0.0
         return BenchmarkResult(
             collector=name,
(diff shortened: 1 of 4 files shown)


torchrl/collectors/_async_batched.py
```diff
@@ -190,18 +229,85 @@ def _env_batch_loop(
...
+
+        transitions = []
+        try:
+            collected = 0
+            for descriptor in descriptors:
+                env_id = descriptor[0]
+                policy_output = replay_outputs[env_id].get()
+                if failed:
+                    continue
+                transition = transition_ring.read(descriptor).clone(recurse=False)
```
vmoens: `read` is a `select` on the slot and `clone(recurse=False)` keeps the leaves as views, so the chunk handed to `extend` aliases ring memory that the `finally` below frees right after. Safe only for storages that copy on write: with the default `ListStorage` 187 of 280 rows per env came back with a wrong `next.observation` in my probe. Materialise the chunk before freeing the slots (or restrict to `TensorStorage` and say so). Also, `NonTensorData(env_id)` two lines down makes `LazyTensorStorage.extend` cost grow with `max_size` (189 ms per 4-row chunk at 100k); write a long tensor.

torchrl/collectors/_async_batched.py
```diff
@@ -784,12 +974,42 @@ def iterator(self) -> Iterator[TensorDictBase]:
         total = self.total_frames
         while total < 0 or self._frames < total:
             self._iter += 1
+            if self.replay_buffer is not None:
```
vmoens: Nothing stops the workers and the writer when this loop exits, so the buffer keeps growing past `total_frames` until `shutdown()` (97 frames at loop end, 1174 written 1.5 s later in my run). Stop stepping when the iterator is exhausted so the overshoot is bounded by the in-flight slots, and assert an upper bound in `test_replay_writer`.

benchmarks/bench_collectors.py
```diff
@@ -322,10 +330,16 @@ def bench(
...
+            replay_buffer.write_count if replay_buffer is not None else 0
+        )
         t0 = time_module.perf_counter()
         for batch in iterator:
-            n = batch.numel()
-            total += n
+            if batch is None:
+                total = replay_buffer.write_count - initial_write_count
+            else:
+                total += batch.numel()
```
vmoens: The baseline never inserts into a replay buffer, so the comparison is "store nothing" vs "store everything". Add `replay_buffer.extend(batch)` here so both modes store, and report that pair.

vmoens (commented):
Reviewed the diff in depth and ran it: the AsyncBatchedCollector tests (21 passed, including the three new ones, three times without flakiness), the AsyncEnvPool tests (27 passed), and a set of consistency probes on the writer path with different storages.

Design summary as I read it: the decoupled part is a thread, not a process. Env workers write the complete transition into a per-env ring of shared-memory slots, publish a small descriptor on one queue, and take a free slot before stepping (that is the back-pressure); coordinators stop cloning and only park the policy output; the writer pairs descriptor and policy output FIFO per env, builds a lazy chunk, calls `extend`, and frees the slots. Per-env ordering is preserved (verified), the ring back-pressure test is deterministic, exceptions in `extend` propagate to the iterator, and with `replay_buffer=None` nothing changes. The outline is sound. Four things need to change before merging.

**1. Replay rows alias ring memory, and the default `ReplayBuffer()` silently stores corrupted transitions (reproduced).** torchrl/collectors/_async_batched.py:285-292: `transition_ring.read(descriptor)` is a `select` on the slot (views), `.clone(recurse=False)` does not copy leaves, and `lazy_stack(transitions).reshape(-1)` keeps pointing at the slots, which are released in the `finally` at 306-308 right after `extend` returns. That is only safe when the storage copies on write. `ListStorage` (the default of a bare `ReplayBuffer()`), `LazyStackStorage`, a caching `postproc` or a retaining transform keep the views. With 3 `CountingEnv`s, `ReplayBuffer(storage=ListStorage(100_000))`, `env_exchange="shm"` and ring depth 2: all 839 stored rows are backed by ring memory and 187 of 280 rows per env violate `next.observation == observation + 1` (`next.*` is whatever the slot holds by the time you look, while `observation`/`action`/`policy_version`, which come from the policy output, are right). `LazyTensorStorage` shows 0 violations. Fix: own the memory before freeing the slot (`torch.stack` / `.contiguous()` on the chunk, then free immediately, which also shortens slot occupancy), or validate that the storage is a `TensorStorage` and document the restriction. Add a `ListStorage` test with the consistency assertion.

**2. `env_index` is written as `NonTensorData`, which makes `LazyTensorStorage.extend` scale with `max_size`.** Line 287. Micro-benchmark with 4-row chunks: `LazyTensorStorage(1_000)` 2.15 ms per extend with `NonTensorData` vs 0.11 ms with a tensor; `LazyTensorStorage(100_000)` 189 ms vs 0.16 ms. End to end the writer stored 750 rows in 1.5 s into a 1k storage but 126 rows into a 100k storage under identical settings; at a realistic 1M-frame buffer each chunk takes seconds, the rings fill and every env blocks on `transition_free_queue.get()`. Write a long tensor (`torch.full(transition.batch_size, env_id)`), as `Collector` does for trajectory ids; it also gives users something to `argsort` to split streams.

**3. Writes do not stop when the iterator ends, so `total_frames` does not bound the buffer.** Lines 977-980 yield and continue; the `while` exits at `total_frames` but coordinators, env workers and the writer keep running until `shutdown()`. Reproduced: `total_frames=96`, `frames_per_batch=16`, 3 envs, `collector._frames == 97` at loop end while `write_count` goes 97 to 445 after 0.5 s to 1174 after 1.5 s. The same holds between yields (`frames_per_batch` only paces the `None` yields). In the non-writer path the surplus stays in the result queue; here it lands in the user's buffer, so budgets derived from `total_frames` or `write_count` are off by an unbounded amount, and `test_replay_writer` only asserts `write_count >= 20`. `Collector` with `replay_buffer` stops at `total_frames`. Stop stepping when the iterator is exhausted so the overshoot is bounded by the in-flight slots, document that bound, and assert an upper bound in the test.

**4. The benchmark cannot show the claimed improvement.** benchmarks/bench_collectors.py:339-342 counts `write_count` for the writer path while the baseline branch never inserts into a replay buffer, so the PR's numbers compare "store nothing" with "store everything"; the path being replaced (coordinator clone plus main-thread `rb.extend(batch)`) is not measured, and `write_count` includes frames written after the last yield (item 3). Add the insertion to the baseline branch and report that pair, ideally at the 64-env preset the docstring shows.

Smaller points:
- Rebase: the base #4264 moved to 4bf53120c (merge of main with the #4263 pause and the #4261 mmap bound); the writer branches in `_env_loop`/`_env_batch_loop` have to be re-threaded into the pause-aware loops, keeping the `put` then `send` order the FIFO pairing relies on. `pause()`'s docstring should say the writer keeps flushing during a pause (relevant when pausing to checkpoint the buffer), and `_async_step_and_maybe_reset_replay_recv`'s queue branch should `.clone()` like main now does.
- Tests verify wiring, not data: no check that every frame is written once, that per-env consistency holds, that other storages work, that the writer error path works (I verified it manually), or shutdown while the writer is blocked in `extend`.
- `replay_outputs` is bounded at `replay_writer_buffer_size` with a `put` that has no timeout or shutdown check; by construction the depth never exceeds ring depth + 1, and when it binds on the batched path it blocks the single coordinator for all envs and `shutdown()` cannot interrupt it. Unbounded (or depth + 1) plus a shutdown-aware put.
- Shutdown ordering: the pool is closed before the writer is stopped; a writer stuck in a slow `extend` makes the join time out silently and the daemon thread keeps writing after `shutdown()` returned; `_drain_result_queues` does not drain the transition queue; `_ensure_started` re-entry creates a second writer. Stop and join the writer first, warn on timeout, and make `_next_result` return early when `_stop` is set (today the iterator thread dies with a misleading "inference server died" traceback on `async_shutdown()`).
- `postproc` now runs in the writer thread on variable-size chunks while the docstring still says "each collected batch before yielding"; chunks interleave envs in arrival order, so per-stream consumers (the DreamerV3 ingestion) need an adapter, and nothing requires a real `ReplayBuffer` beyond `.extend`, so type it as such or document the pattern.
- The next observation is copied twice per step (ring plus exchange slot), cheap for counters, not for pixels; `iterator()` is still typed as yielding tensordicts; `AsyncEnvPool.__init__` gains an underscore kwarg in a public signature, which also sidesteps config parity; the collector reaches into four private pool attributes.

Interaction with the sibling PRs: #4271 restructures the same `_setup` into worker groups (`worker_slots(start, stop)`, one input queue per worker with fan-out), so the per-env free queues, the `init_transition_ring` message and `worker_slots(env_index)` conflict and must go through that fan-out. #4272 adds a third transition channel (`transition.clone()` through a `multiprocessing.Queue` from the env process) that bypasses `AsyncEnvPool` entirely and rewrites the same `_ensure_started`/`shutdown`/`_next_result` regions. The stack ends up with three ways to move transitions, each with its own back-pressure story; converging on this ring as the single channel (it fits #4272's workers, which already own the transition in the env process) and landing #4271's grouping before or with it would remove the duplication.

PR #4271, opened 2026-09-07
[Performance] Let AsyncEnvPool workers host multiple environments

vmoens:
`envs_per_worker` lets multiprocessing `AsyncEnvPool` host several environments concurrently in each process, reducing process overhead while preserving global environment indices and per-environment receives. CPU affinity applies before environment threads and factories start; the Hydra configuration exposes the same default.

Targets main; #4269 is merged. Shared-memory and grouped queue exchanges use one collector coordinator to own aggregate receives. It consumes ready resets and steps independently, so a slow initial or later reset cannot stall unrelated streams. Worker exceptions propagate to callers, and shutdown remains bounded when a dead worker leaves a full command queue.

Validation:
- Combined AsyncEnvPool and AsyncBatchedCollector suites: 76 passed, 7 Linux affinity cases skipped on macOS.
- Configuration coverage: 2 passed.
- Behavioral regressions cover slow initial/later resets with both exchanges, grouped worker exceptions, exact frame budgets, pause, and shutdown with a dead worker and full command queue.
- ufmt, flake8, and diff checks pass.
- Linux affinity passed in the combined collector/pool selection. Matched GPU collector runs are recorded in #4272.

The existing benchmark covers grouped workers. Three matched fake-pixel runs with 16 environments on GB200 measured 1073.9–1092.9 frames/s and 14.0–14.5 GiB summed host RSS with one environment per process, versus 963.5–1085.0 frames/s and 5.3–6.3 GiB with four per process. Grouping reduces process memory; these short runs do not establish a throughput gain. Full hardware, workload, timing and memory definitions are in #4272.

### DIFF
--- a/torchrl/collectors/_async_batched.py
+++ b/torchrl/collectors/_async_batched.py
@@ -180,17 +180,31 @@ def _env_batch_loop(
     try:
         env_ids = list(range(pool.num_envs))
         pool.async_reset_send(env_index=env_ids)
-        observations = pool.async_reset_recv(min_get=pool.num_envs)
-        for env_id, observation in zip(_env_ids(observations), observations.unbind(0)):
-            ready_observations[env_id] = observation
+        resetting = pool.num_envs
 
         while not shutdown_event.is_set():
+            if resetting:
+                try:
+                    observations = pool.async_reset_recv(
+                        min_get=1, max_get=resetting, timeout=poll_interval
+                    )
+                except TimeoutError:
+                    pass
+                else:
+                    reset_ids = _env_ids(observations)
+                    resetting -= len(reset_ids)
+                    ready_observations.update(zip(reset_ids, observations.unbind(0)))
             if pause_request[0] is None and ready_observations:
                 for env_id, observation in ready_observations.items():
                     pending_actions[env_id] = clients[env_id].submit(observation)
                 ready_observations.clear()
 
-            if pause_request[0] is not None and not pending_actions and not stepping:
+            if (
+                pause_request[0] is not None
+                and not resetting
+                and not pending_actions
+                and not stepping
+            ):
                 _wait_while_paused(pause_request)
                 continue
 
@@ -204,7 +218,11 @@ def _env_batch_loop(
                     if env_device is not None:
                         action = action.to(env_device)
                     policy_outputs[env_id] = action
-                    env_inputs.append(action.select(*exchange_keys, strict=False))
+                    env_inputs.append(
+                        action.select(*exchange_keys, strict=False)
+                        if exchange_keys
+                        else action
+                    )
                 pool.async_step_and_maybe_reset_send(
                     lazy_stack(env_inputs), env_index=ready_ids
                 )
@@ -220,7 +238,7 @@ def _env_batch_loop(
                 transitions, next_observations = pool.async_step_and_maybe_reset_recv(
                     min_get=1,
                     max_get=pool.num_envs,
-                    timeout=poll_interval if pending_actions else None,
+                    timeout=poll_interval,
                 )
             except TimeoutError:
                 continue
@@ -267,7 +285,7 @@ class AsyncBatchedCollector(BaseCollector):
     * An :class:`~torchrl.envs.AsyncEnvPool` runs *N* environments using
       whatever backend the user chooses (``"threading"``,
       ``"multiprocessing"``).
-    * With a shared-memory environment exchange, one coordinator thread drains
+    * With a shared-memory exchange or grouped workers, one coordinator drains
       whichever environments are ready and submits their observations without
       blocking. Other exchanges use one lightweight coordinator thread per
       environment.
@@ -345,6 +363,10 @@ class AsyncBatchedCollector(BaseCollector):
             :class:`~torchrl.envs.AsyncEnvPool`, one of ``"queue"``, ``"shm"``
             or ``"auto"``. The shared-memory exchange also enables batched
             coordination from one thread. Defaults to ``"queue"``.
+        envs_per_worker (int, optional): Number of environments hosted by each
+            multiprocessing worker. Grouped workers share one coordinator that
+            drains ready environments without waiting for a complete group.
+            Defaults to ``1``.
         policy_backend (str, optional): backend for the inference transport
             used to communicate with the
             :class:`~torchrl.modules.InferenceServer`.  One of
@@ -380,12 +402,12 @@ class AsyncBatchedCollector(BaseCollector):
             step-time jitter; most users should leave it unset. TorchRL knows
             which CPUs are available, but not the application's intended CPU
             partition or each environment's thread requirements, so it cannot
-            choose these masks automatically. Provide one mask per environment,
-            or a callable mapping each environment index to its mask. Defaults
+            choose these masks automatically. Provide one mask per worker process,
+            or a callable mapping each worker index to its mask. Defaults
             to ``None``. See :ref:`async_batched_collector_cpu_affinity` for a
             complete collector example.
         driver_affinity (Sequence[int], optional): Linux CPU affinity mask for
-            the inference-server and per-environment coordinator threads, plus
+            the inference-server and coordinator threads, plus
             parent-side multiprocessing queue feeder threads. Dedicated
             process-backed inference servers are not covered. The thread
             constructing the collector is restored to its original affinity
@@ -430,6 +452,7 @@ class AsyncBatchedCollector(BaseCollector):
         ] = "threading",
         env_backend: Literal["threading", "multiprocessing"] | None = None,
         env_exchange: Literal["queue", "shm", "auto"] = "queue",
+        envs_per_worker: int = 1,
         policy_backend: (
             Literal["threading", "multiprocessing", "ray", "monarch"] | None
         ) = None,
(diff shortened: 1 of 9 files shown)


torchrl/collectors/_async_batched.py
```diff
@@ -115,6 +115,57 @@ def _env_loop(
...
+            action_futures = [
+                clients[env_id].submit(observation)
+                for env_id, observation in zip(
+                    env_ids, observations.unbind(0), strict=True
+                )
+            ]
+            actions = [future.result() for future in action_futures]
+            if env_device is not None:
+                actions = [action.to(env_device) for action in actions]
+            pool.async_step_and_maybe_reset_send(lazy_stack(actions), env_index=env_ids)
```
vmoens: List `env_index` means the aggregate path: `_current_step_reset` is read-modify-written without a lock (`self._lock` in the pool is never used), and with several coordinator threads the interleaving is reachable (reproduced under `sys.setswitchinterval(1e-6)`: all coordinators die with `Cannot await 2 step_and_maybe_reset when only 0 are being stepped`). Also, after the first round `env_ids` is whatever arrived first from any worker, so one message per env is sent anyway and a coordinator holding `k-1` results waits for the k-th (149 vs 93 frames/s with jittered latency). Use per-env send/recv per coordinator, or one harvesting coordinator with `min_get=1`; and this loop needs `pause_request`/`_wait_while_paused` from #4263 or `pause()` times out for grouped workers.

torchrl/collectors/_async_batched.py
```diff
@@ -115,6 +115,57 @@ def _env_loop(
...
+            transitions, observations = pool.async_step_and_maybe_reset_recv(
+                min_get=batch_size, max_get=batch_size
+            )
+            for transition in transitions.unbind(0):
```
vmoens: `_ENV_IDX_KEY` is not set here, so the worker-side `NonTensorData` survives and `batch["env_index"]` is a Python list at `envs_per_worker>1` but an int tensor at `envs_per_worker=1`. Set it per transition and assert the dtype in the test.

torchrl/envs/async_envs.py
```diff
@@ -1040,6 +1157,84 @@ def shutdown(self):
...
+            per_env_step_reset_queue,
+        ) in zip(
+            env_indices,
+            env_makers,
+            create_env_kwargs,
+            per_env_step_queues,
+            per_env_reset_queues,
+            per_env_step_reset_queues,
+        ):
+            env_thread = threading.Thread(
```
vmoens, line 1431: No try/except around `_env_exec` and non-daemon threads: an exception in one env leaves the process alive and the driver blocked forever (3/4 results, both processes alive in my repro), whereas on main the process dies, which is the only liveness signal the collector has. Publish the exception on the result queues and exit non-zero; make the threads daemon and forward shutdown in a `finally`.

vmoens (commented):
Reviewed the diff in depth and ran it (AsyncEnvPool tests: 34 passed; AsyncBatchedCollector tests: 17 passed; env configs: 11 passed, 1 skipped; the same suites on a local merge with current main: 35 and 18 passed), plus a set of experiments with sleeping and raising envs.

The pool side is sound and does what the motivation asked for: contiguous env indices are grouped per process, one thread per hosted env fans out from a dispatcher, the shm exchange hands every worker one contiguous slice of the per-env slot buffers so the slot layout and lease/release semantics are unchanged, and the default `envs_per_worker=1` keeps one process per env. Measured overlap: four envs sleeping 50 ms each in one process complete a round in 55-58 ms, not 200 ms, in both exchanges, and per-env recv, subset reset inside a worker and across workers all return the right env's data. The problems are concentrated in the collector integration (`_env_worker_loop`), so this needs changes before merging.

**1. `_env_worker_loop` drives the pool's non-thread-safe aggregate API from several threads (reproduced).** torchrl/collectors/_async_batched.py:156-159 sends with a list `env_index` and receives with `min_get=batch_size`, which is the shared-counter path in torchrl/envs/async_envs.py:978 / :1004-1007 / :1015: unguarded read-modify-writes on `_current_step_reset` (`self._lock` is created at :298 and never used). The docstring only promises thread safety for the per-env path, which the old `_env_loop` respected. With 16 envs, 2 per worker and 8 coordinators running exactly this call sequence: 0 errors in 16,000 iterations at the default switch interval, but with `sys.setswitchinterval(1e-6)` all 8 coordinators die with `RuntimeError: Cannot await 2 step_and_maybe_reset when only 0 are being stepped` and `_busy` keeps every env forever. On a busy driver this is a rare collector death. Fix: keep one coordinator per env on the per-env send/recv (thread-safe by construction, and it works unchanged on grouped envs), or use a single harvesting coordinator (`min_get=1, max_get=num_envs`, as #4264 does), or lock every aggregate send/recv in `ProcessorAsyncEnvPool`.

**2. Coordinators are not bound to workers, so the batching claim does not hold and head-of-line blocking costs throughput (measured).** After the first round a coordinator's `env_ids` is whatever `batch_size` results arrived first from any worker, and `_send_worker_batches` then emits one message per worker touched: 14,927 of 16,000 coordinator batches spanned two workers with 2 envs per worker, i.e. one pipe message per env, no reduction versus `envs_per_worker=1`. Worse, a coordinator holding `k-1` results waits for the k-th before any of them gets its next action. With 8 envs sleeping U(10, 90) ms the collector yields 149 frames/s at `envs_per_worker=1` and 93 frames/s at `envs_per_worker=4`. The JVM envs this targets have exactly that kind of jitter, so as written the collector path makes the motivating workload slower even though the pool overlaps fine. Same fix as item 1; the single harvesting coordinator is the design that also delivers "fewer threads".

**3. After merging with main, `pause()` from #4263 times out for grouped workers (reproduced on a synthetic merge).** `_env_worker_loop` has no `pause_request` and never parks, `git merge-tree` reports no textual conflict, so the merge is clean and wrong: `envs_per_worker=1` pauses in 0.02 s, `envs_per_worker=2` raises `TimeoutError: Timed out while waiting for AsyncBatchedCollector coordinator threads to pause`. Thread `pause_request` through the loop and extend main's pause test with `envs_per_worker=2`.

**4. `env_index` changes type between layouts.** The old path sets `_ENV_IDX_KEY` to an int tensor; the new loop (lines 160-163) never sets it, so the worker-side `NonTensorData(i)` survives: with 6 envs, `batch.get("env_index")` is a `Tensor[60]` at `envs_per_worker=1` and a `NonTensorStack` at `envs_per_worker=2`, and `batch["env_index"]` becomes a Python list. Masking, comparisons and `.to(device)` on it break when a user flips the layout; the test hides it by iterating with `int(env_id)`. Set the key per unbound transition and assert the dtype in the test.

**5. An exception in one env of a grouped worker is invisible and the process stays alive.** `_env_exec` runs as a plain thread target (torchrl/envs/async_envs.py:1192-1209) with no try/except and no error publication, and the dispatcher has no notion of a dead env thread. With env 1 raising in `_step`: `async_step_recv(min_get=4, timeout=5)` gives `TimeoutError ... 3/4 results`, both processes alive, exit codes `None`; without a timeout the driver blocks forever. On main the same crash kills the process, which is the only liveness signal today; grouped workers remove it. The env threads are also non-daemon and the dispatcher has no `finally` forwarding shutdown, so a dispatcher exception leaves the process alive until the 60 s shutdown timeout. Publish env-thread exceptions on the result queues (or a shared failure flag) and exit the process non-zero; make the threads daemon; add a test with a raising env.

Smaller points:
- Test gaps: the per-env path of the multi-env worker is never exercised (`test_multiprocessing_envs_per_worker` resets env 4, which is alone in its worker), no subset reset within a worker, no exception test, no grouped-collector consistency check beyond the set of env ids; the four `test_invalid_envs_per_worker` cases and `test_envs_per_worker_rejects_threading` restate validation wiring.
- Envs are now constructed and seeded inside worker threads for k>1; libraries needing the main thread (signal handlers, GL/GUI contexts) fail only when grouped, and co-hosted envs share the process-global RNG state. Worth a sentence in the docs paragraph.
- `num_workers` exists only on `ProcessorAsyncEnvPool` while tests and the benchmark treat it as API, and the collector reads the private `_worker_env_indices`; expose both as documented properties on the base class.
- The benchmark drives the pool's single-threaded aggregate path with uniform 50 ms envs, so it cannot show item 2, and the 64 x 1 baseline has not been run anywhere yet (64 spawned CUDA-torch processes on the 16 vCPU benchmark runner is also heavy). Suggest 32 envs with jittered latency and a collector-level counterpart.

Interaction with the sibling PRs: this PR already contains the identical `_env_batch_sizes` caching line as #4258, so one of the two is redundant. #4266 conflicts textually and its per-env `affinity_ready` handshake assumes per-env queues; it composes if the mask is applied inside the env thread. #4270 indexes `input_queue[_env_idx]` and sends `init_transition_ring` per env, both of which must become per-worker. #4270 and #4272 are stacked on #4264, which introduces a competing single-coordinator design for the same problem as `_env_worker_loop`; one of the two should be dropped, and #4264's shape is the one that avoids items 1 and 2.

Pre-existing and not caused by this PR, worth its own issue: after a worker death, `shutdown()` hangs in `input_queue.put` because the dead worker's `Queue(maxsize=1)` still holds the first shutdown message (reproduced on main as well).

vmoens:
The grouped collector needs a scheduling fix before this PR is ready to merge: a slow reset can stall otherwise ready environments.

Grouping processes is reasonable for I/O-bound environments. The pool runs each hosted environment in its own thread, so a reset spent waiting on a simulator or socket need not prevent the other threads from progressing. CPU-heavy Python work can still contend for the GIL.

The problem is in [`_env_worker_loop`](https://github.com/pytorch/rl/blob/ea228d9a8508b471b21eb53c6bd8c4775098e054/torchrl/collectors/_async_batched.py#L160-L187): each coordinator waits for a full worker-sized batch of results before submitting more actions. With two environments in one worker, if A resets for 20 seconds and B finishes immediately, B receives no next action until A finishes. With multiple workers, coordinators consume a shared result queue, so the stalled streams need not belong to the same process.

The coordinator should consume ready completions and promptly resubmit those streams, batching results that are already available instead of requiring a full group. This must preserve per-environment ordering and the collector's pause boundary.

The current benchmark does not protect this behavior: it uses the pool directly with `min_get=1` and zero reset latency. Before marking this PR ready again, add a regression test proving that a fast environment continues stepping while another remains in a long reset, and validate the collector under mixed step/reset latencies.

PR #4272, opened 2026-09-07
[Performance] Run inference from environment processes

vmoens:
`ProcessSlotTransport` gives each spawned environment a fixed shared-memory inference slot. Workers run reset, inference and stepping directly against the process-owned server; the parent receives owned completed transitions. This PR depends on #4271 and supports static CUDA-graph inference from #4269.

Workers inherit configured CPU affinity before constructing environments. Pause waits for in-flight work, worker exceptions and exits reach the iterator even while other streams are active, and repeated shutdown remains bounded. Direct process slots explicitly reject grouped workers; grouping remains available through AsyncEnvPool.

Validation found and fixed an unreadable-result error after worker death. The Linux affinity test now waits for startup acknowledgement before inspecting worker masks. Completed CPU transitions are consolidated into one owned storage before transfer: profiling showed per-leaf file-descriptor negotiation consumed about 90% of parent collection time.

Validation:
- Full CPU inference suite before final transport consolidation: 144 passed, 15 skipped.
- Linux/GPU selection: 28 initially passed; both failures were fixed and the five affinity/error cases passed on rerun. Static capture, including direct process slots, passed.
- After consolidation, Linux process ownership, repeated pause, worker exceptions and worker death: 4 passed.
- Formatting, lint, docstring arguments, Sphinx underline and diff checks pass. Dependency suites are enabled before push.

Matched fake-pixel benchmark: NVIDIA GB200, PyTorch 2.15.0.dev20260906+cu130, TensorDict 0.14 main (faee822), 16 environments, 5 ms environment latency, two 1024-unit policy layers, 256 frames/batch, two warm-up batches and 2,048 measured frames. Three runs per case; all use identical seeds, dependencies and hardware. Startup, server-stat RPCs and memory inspection are excluded from measured collection time. Host RSS sums parent and children at the end; CUDA memory is device-wide used memory on the dedicated GPU, not peak tensor allocation.

| Mode | Frames/s range | Median batch latency ms | p95 batch latency ms | Host RSS MiB | CUDA used MiB |
|---|---:|---:|---:|---:|---:|
| Shared-memory coordinator, one env/worker | 1073.9–1092.9 | 234.30–238.46 | 239.33–246.90 | 14350.5–14865.9 | 892.3 |
| Shared-memory coordinator, four envs/worker | 963.5–1085.0 | 236.04–244.59 | 241.80–319.33 | 5397.6–6452.3 | 892.3 |
| Direct process slots, original per-leaf transfer | 363.8–376.0 | 682.69–699.56 | 697.72–723.99 | 16903.1–18857.1 | 1780.1–1782.1 |
| Direct process slots, consolidated transfer | 1636.9–1960.7 | 122.17–132.40 | 174.89–259.79 | 14753.6–15274.6 | 1584.1 |

The consolidation change removes the reproducible direct-process regression in this workload. Grouping reduces process memory, but these short runs do not establish a throughput gain. Parent replay backpressure, pause and shutdown validation passed in the integration stack; these figures describe this collector workload only.

Final bounded integration stack, three further repetitions with the same workload, dependencies and warm-up:

| Mode | Frames/s | p50 batch ms | p95 batch ms | Host RSS MiB | CUDA used MiB |
|---|---:|---:|---:|---:|---:|
| Shared-memory coordinator, one env/worker | 1009.81–1061.53 | 240.45–242.73 | 245.13–291.56 | 15312.44–15375.44 | 892.31–892.31 |
| Shared-memory coordinator, four envs/worker | 1072.05–1077.66 | 237.55–238.14 | 243.39–244.98 | 5458.88–5968.50 | 892.31–892.31 |
| Direct slots, consolidated transfer | 1940.39–2024.66 | 115.61–122.17 | 180.96–184.33 | 14763.44–15781.44 | 1584.12–1584.12 |

The final integration CPU run passed 904 inference/objective/replay tests. The selected Linux/GPU collector tests passed all 19 cases. The reset-error regression now measures propagation after the worker acknowledges entering reset, rather than including process startup in its failure deadline.

## Continuous pixel-collection comparison

The [isolated #4269/#4271/#4272 stack](https://github.com/pytorch/rl/actions/runs/34163040035) includes the CUDA input-padding follow-up and the process-slot continuous series from #4289. Three GPU repetitions passed with identical dependency/system-package lists and GPU model/driver to the earlier continuous controls: NVIDIA A10G, 32 CPU environments, 1 ms steps, CNN plus two 1024-wide layers, 256-frame output batches, two warm-up rounds and five measured 1,024-frame rounds. One slow-reset stream stalls for 200 ms every 16 steps. This differs from the GB200/16-environment/5 ms workload above.

| Mode | Uniform frames/s [run range] | Slow-reset frames/s [run range] | Uniform batch p95 ms | Uniform summed RSS MiB |
| Grouped coordinator, eager | 849.8 [847.2, 864.2] | 854.8 [835.9, 870.7] | 342.7 | 7338.6 |
| Grouped coordinator, graph | 764.4 [751.7, 778.0] | 759.8 [739.7, 774.4] | 443.0 | 7282.9 |
| Direct slots, eager | 684.5 [681.3, 688.2] | 681.1 [677.2, 688.8] | 524.8 | 26015.8 |
| Direct slots, graph | 660.2 [660.0, 662.3] | 660.8 [658.7, 662.6] | 545.3 | 26208.6 |

These compare supported configurations with the same total environment count: grouped coordination uses four environments per worker; direct slots require one. They do not isolate worker count from transport. Direct server forward medians improve from 1.62 ms eager to 0.87 ms graph, while total collection slows. Profile the remaining transport/collection costs before claiming a gain for this workload. The earlier GB200 win does not generalize to this A10G workload. RSS counts shared pages per process; the benchmark omits the child server's CUDA allocator peak rather than reporting the parent's unrelated value.

### DIFF
--- a/torchrl/collectors/_async_batched.py
+++ b/torchrl/collectors/_async_batched.py
@@ -5,12 +5,13 @@
 from __future__ import annotations
 
 import contextlib
+import multiprocessing as mp
 import os
 import queue
 import threading
 import time
 from collections import deque, OrderedDict
-from collections.abc import Callable, Iterator, Sequence
+from collections.abc import Callable, Iterator, Mapping, Sequence
 from typing import Literal
 
 import torch
@@ -21,11 +22,15 @@ from tensordict import (
     NestedKey,
     TensorDictBase,
 )
-
 from torchrl._comm import MailboxTransportError
-from torchrl._utils import _maybe_record_function_decorator, logger as torchrl_logger
+from torchrl._utils import (
+    _maybe_record_function_decorator,
+    logger as torchrl_logger,
+    timeit,
+)
 from torchrl.collectors._base import BaseCollector
-from torchrl.envs import AsyncEnvPool, EnvBase
+from torchrl.data.utils import CloudpickleWrapper
+from torchrl.envs import AsyncEnvPool, EnvBase, EnvCreator
 from torchrl.envs.async_envs import _validate_cpu_affinity
 from torchrl.modules.inference_server import (
     InferenceDeviceConfig,
@@ -33,6 +38,7 @@ from torchrl.modules.inference_server import (
     InferenceServerConfig,
     PolicyClientModule,
     ProcessInferenceServer,
+    ProcessSlotTransport,
     ThreadingTransport,
 )
 from torchrl.modules.inference_server._config import _resolve_device_config
@@ -141,6 +147,65 @@ def _env_loop(
     except Exception as exc:
         if not shutdown_event.is_set():
             result_queue.put(exc)
+
+
+def _process_env_loop(
+    env_factory: Callable[[], EnvBase],
+    create_env_kwargs: dict,
+    env_id: int,
+    client: PolicyClientModule,
+    result_queue,
+    shutdown_event,
+    pause_event,
+    paused_event,
+    error_queue,
+    worker_affinity: Sequence[int] | None,
+    env_device: torch.device | None,
+    storing_device: torch.device | None,
+):
+    """Run environment stepping and remote inference in one worker process."""
+    env = None
+    try:
+        if worker_affinity is not None:
+            os.sched_setaffinity(0, worker_affinity)
+        torch.set_num_threads(1)
+        env = env_factory(**create_env_kwargs)
+        observation = env.reset()
+        while not shutdown_event.is_set():
+            if pause_event.is_set():
+                paused_event.set()
+                while pause_event.is_set() and not shutdown_event.is_set():
+                    shutdown_event.wait(0.01)
+                paused_event.clear()
+                continue
+            policy_output = client(observation)
+            action_td = observation.update(policy_output)
+            if env_device is not None:
+                action_td = action_td.to(env_device)
+            transition, observation = env.step_and_maybe_reset(action_td)
+            transition.set(_ENV_IDX_KEY, env_id)
+            # multiprocessing.Queue serializes on a feeder thread after put()
+            # returns. Own the transition storage before the environment or
+            # inference response slot can be reused.
+            transition = transition.clone()
+            if storing_device is not None:
+                transition = transition.to(storing_device)
+            if all(
+                value.device.type == "cpu"
+                for value in transition.values(True, True)
+                if isinstance(value, torch.Tensor)
+            ):
+                # Transfer one shared storage instead of negotiating a file
+                # descriptor for every tensor leaf of every transition.
+                transition = transition.consolidate()
+            result_queue.put(transition)
+    except Exception as exc:
+        if not shutdown_event.is_set():
+            error_queue.put(RuntimeError(f"Environment {env_id} failed: {exc!r}"))
+            shutdown_event.wait()
+    finally:
+        if env is not None:
+            env.close()
 
 
 def _env_ids(tensordict: TensorDictBase) -> list[int]:
@@ -289,6 +354,9 @@ class AsyncBatchedCollector(BaseCollector):
       whichever environments are ready and submits their observations without
       blocking. Other exchanges use one lightweight coordinator thread per
       environment.
+    * With a :class:`~torchrl.modules.inference_server.ProcessSlotTransport`,
+      each multiprocessing environment worker talks directly to the dedicated
+      inference process; the driver receives completed transitions only.
     * The :class:`~torchrl.modules.InferenceServer` running in a background
       thread continuously drains observation submissions, batches them, runs
       a single forward pass, and fans actions back out.
@@ -327,9 +395,12 @@ class AsyncBatchedCollector(BaseCollector):
         server_timeout (float, optional): seconds the server waits for work
             before dispatching a partial batch.  Defaults to ``0.01``.
         transport (InferenceTransport, optional): a pre-built transport
-            object.  When provided, it takes precedence over
-            ``policy_backend``.  When ``None`` (default) a transport is
-            created automatically from the resolved ``policy_backend``.
+            object. When provided, it takes precedence over ``policy_backend``.
+            A :class:`~torchrl.modules.inference_server.ProcessSlotTransport`
+            together with multiprocessing environment and server backends runs
+            the complete acting loop in environment worker processes. When
+            ``None`` (default), a transport is created from the resolved
+            ``policy_backend``.
         device (torch.device or str, optional): device for policy inference
             (shorthand for ``InferenceDeviceConfig(policy_device=...)``).
             Defaults to ``None``.
@@ -604,6 +675,38 @@ class AsyncBatchedCollector(BaseCollector):
                 effective_policy_backend, num_slots=self._num_envs
             )
         self._transport = transport
+        self._uses_process_env_workers = isinstance(transport, ProcessSlotTransport)
+        if self._uses_process_env_workers:
+            if envs_per_worker != 1:
+                raise ValueError("ProcessSlotTransport requires envs_per_worker=1.")
+            if (
+                server_backend != "process"
+                or effective_env_backend != "multiprocessing"
+            ):
+                raise ValueError(
+                    "ProcessSlotTransport requires both a process inference server "
+                    "and env_backend='multiprocessing'."
+                )
+            if env_exchange != "queue":
+                raise ValueError(
+                    "env_exchange does not apply when ProcessSlotTransport owns "
+                    "the environment-worker exchange; leave it as 'queue'."
+                )
+            if transport._num_slots < self._num_envs:
+                raise ValueError(
+                    f"ProcessSlotTransport needs at least one slot per environment "
+                    f"({self._num_envs}), but has {transport._num_slots}."
+                )
+            for name, target_device in (
+                ("env_device", self._env_device),
+                ("storing_device", self._storing_device),
+            ):
+                if target_device is not None and target_device.type != "cpu":
+                    raise ValueError(
+                        f"ProcessSlotTransport requires a CPU {name}; got "
+                        f"{target_device}. Keep CUDA policy execution in the "
+                        "inference server process."
+                    )
 
         # ---- build inference server -------------------------------------------
         if server_backend == "process":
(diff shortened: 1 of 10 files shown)


torchrl/collectors/_async_batched.py
```diff
@@ -563,6 +641,69 @@ def _ensure_started(self) -> None:
...
+            self._shutdown_event = ctx.Event()
+            self._workers = []
+            self._uses_batched_coordinator = False
+            try:
+                for env_id, (env_factory, env_kwargs, client) in enumerate(
+                    zip(self._create_env_fn, create_env_kwargs, self._clients)
+                ):
+                    if not isinstance(env_factory, (EnvCreator, CloudpickleWrapper)):
+                        env_factory = CloudpickleWrapper(env_factory)
+                    process = ctx.Process(
```
vmoens: No `daemon=True` here (the server process and the coordinator threads are daemonic), and `_process_env_loop` never observes the driver: after a SIGKILL of the driver both env processes kept stepping and the server kept serving 6 s later, with memory growing in the unbounded result queue. Make the processes daemonic, watch `mp.parent_process()` in the loop, and bound the result queue.

torchrl/collectors/_async_batched.py
```diff
@@ -132,6 +135,44 @@ def _env_loop(
...
+    env = None
+    try:
+        torch.set_num_threads(1)
+        env = env_factory(**create_env_kwargs)
+        observation = env.reset()
+        while not shutdown_event.is_set():
+            policy_output = client(observation)
+            action_td = observation.update(policy_output)
+            if env_device is not None:
+                action_td = action_td.to(env_device)
```
vmoens, line 184: One `clone()` plus one `mp.Queue.put` per env step re-introduces the per-step pickling (a fresh shared-memory segment per leaf per put) that #4264 removed with the shm exchange; this mode forces `env_exchange="queue"`. Publishing through #4270's per-env rings instead would keep the driver read-only. Also note `pause()` from #4263 cannot park these processes: it waits on a `threading.Barrier` they never reach and times out after 30 s; raise `NotImplementedError` for this mode or add an `mp.Event` boundary here.

vmoens (commented):
Reviewed e3a99f842 in depth (full non-GPU inference-server suite, the new docstring example, error-path and killed-server probes, a server-side cost measurement, a driver-death experiment, the benchmark) and re-checked the findings below against the force-pushed head ef8807454 on top of the rebased #4264. Line numbers are for ef8807454.

The transport itself is small and does what it says: request/response attribution, exception propagation, dead-server detection (clients raise `MailboxPeerClosedError` within 0.1 s), round-robin fairness and cross-process visibility all behaved in my runs, and the server-side Python cost is about 2 ms per 64-slot batch of 3x84x84 uint8 observations, so it is not the bottleneck. The new head also fixes the `TestWorkerCrashPropagation` regression the first push had (the message is back to "A collector worker thread raised an exception." for threads). Three things still need to change before merging.

**1. The environment worker processes are neither daemonic nor parent-aware; a dead driver leaves a self-driving pipeline running.** torchrl/collectors/_async_batched.py:680-696 creates `ctx.Process(...)` without `daemon=True` (the server process and the coordinator threads are daemonic), `_process_env_loop` (lines 135-165) only exits on the `shutdown_event` and never reads anything from the driver, and the result queue is an unbounded `ctx.Queue()` (line 670), so `put()` never blocks. Measured with 2 `CountingEnv`s: SIGKILL the driver after the first batch and 6 s later both env processes are still stepping at 9% CPU each and the server is serving at 15% CPU, with memory growing as every pickled transition accumulates in the feeder buffer; the baseline pool workers survive too, but idle at 0% on `input_queue.get()`. Interpreter exit without `shutdown()` hangs in `_exit_function` joining the non-daemon processes (pre-existing for the pool's workers as well, but this adds a second family that keeps burning CPU/GPU). Fix: `daemon=True`, watch the parent in the loop (a thread on `mp.parent_process().sentinel`, or `mp.parent_process().is_alive()` once per step), give the result queue a `maxsize` of a few transitions per env so a stalled driver applies back-pressure, and add a test for the exit path.

**2. `pause()` does not work in this mode.** The rebase brought in #4263's `pause()` (lines ~1000-1050), which builds `threading.Barrier(len(workers) + 1)` and waits for every worker to park in `_wait_while_paused`. In process-env mode `self._workers` are `mp.Process` objects that never consult `_pause_request` (a list of `threading` objects living in the driver), so `pause(timeout=30)` raises `TimeoutError` after 30 s and `pause(timeout=None)` hangs. Either raise `NotImplementedError` at the top of `pause()` when `_uses_process_env_workers`, or add an `mp.Event`-based boundary in `_process_env_loop` (checked between `client()` and `env.step_and_maybe_reset`) plus an `mp.Barrier`, and test it either way.

**3. The coordinator hop is removed, but transitions are pickled per step again and no win is demonstrated.** Lines 155-162: `transition.clone()` then one `mp.Queue.put` per env step; torch's pickler moves every leaf into a fresh shared-memory segment per put, which is the cost #4264 removed with `env_exchange="shm"` (this mode forces `env_exchange="queue"`, and `_rollout_frames` receives one unbatched transition per step). The only numbers in the PR (138.9 vs 138.7 steps/s against an environment ceiling of 160) are environment-bound and cannot show the latency win that motivates the change; a 4-env run here gave 215 fps for the process-slot mode vs 243 fps for the MPTransport process server (single run, macOS, where `sem_timedwait` is emulated by polling, so Linux will differ). The 64-env CUDA measurement with a DreamerV3-sized policy is the acceptance test for this PR and it is missing. Consider publishing transitions through #4270's per-env shared-memory rings instead of a queue, so the env process writes both its request slot and its transition ring and the driver only reads rings.

Smaller points:
- Memory ordering: the readiness byte (`_process_slot.py:121`) is a lock-free `Array("b", lock=False)`; tensor stores are ordered before the flag store only for a reader that synchronised through the semaphore, and `InferenceServer._run` also drains after a timed-out `wait_for_work`, so on ARM (the GB200's CPU) a flag can in principle be observed before its payload. Practically unobservable at Python latencies, but undocumented: trust flags only up to the number of acquired signals, or say explicitly that correctness relies on the semaphore barrier plus the timed drain. The `_acquired_signals` hint logic deserves the same comment.
- Per-slot `Event` plus `SimpleQueue` (`_process_slot.py:283-284`) is 4 semaphores and a pipe per env, inherited by the server and the driver: about 130 fds and 400 named semaphores at 64 envs, above the default `ulimit -n` at 1024 envs. One shared exception queue carrying `(slot, exc)` removes the per-slot pipes.
- Tests cover wiring; `resolve_exception` to client raise, dead server, `result(timeout=...)` keeping the slot, `copy_result=False`, client overflow, the spec/device rejections and the collector's new `ValueError`s are all unpinned (`SharedMemoryTransport` has the analogous tests in the same file). `_shm_actor_fn` sends `torch.full((4,), i + 1)` from every worker, so a response delivered to the wrong slot still satisfies `obs * 2`; offset the payload by `client_id` so attribution is actually asserted.
- `_make_slots`, the CPU-leaf check, `resolve` and the round-robin sweep are verbatim copies of `_shared_memory.py` and `_slot.py` (the sweep is the one #4264 just fixed for starvation in one place); a shared slot-bank helper and one sweep would keep the three transports from drifting.
- Nits: `getattr(transport, "_clients_require_registration", True)` should be a declared attribute on `InferenceTransport`; `create_env_kwargs` shape validation runs at first iteration rather than in `__init__`; re-entering the iterator after one env process died respawns a full new set against a new queue while survivors keep running (heavier than the thread case); `drain_with_timing`'s return annotation and the untyped shared primitives in `_ProcessSlotClient.__init__`; the `MPTransport`/`MonarchTransport` reordering in `__init__.py` is unrelated churn. Docs, `Literal`, the rst entries and the runnable docstring example (9 examples, 0 failures) are in order.

Interaction with the sibling PRs: #4269 edits the same `_factory.py` block, the same rst insertion point and `ProcessInferenceServer.__init__` where this PR changes `_service_client` (textual conflicts; natural synergy, since with this transport the request spec is known and can feed `static_batch_size`). #4270 and #4271 both assume coordinator threads plus an `AsyncEnvPool`, which do not exist in this mode, so `replay_buffer` and `envs_per_worker` would silently not apply unless this PR rejects them or grows equivalents. The stack now has three round-robin slot sweeps, two identical shm slot banks and, with #4270, a third shm ring; if all of these are wanted, the order that minimises rework is #4264, #4269, this PR (items 1 to 3 fixed), #4271, then #4270 absorbing this PR's worker loop so transitions flow env process to ring to writer, which also removes the per-step pickling.

torchrl/modules/inference_server/_process_slot.py
```diff
@@ -0,0 +1,400 @@
...
+        interaction_code = td.get(_REMOTE_INTERACTION_TYPE_KEY, default=None)
+        interaction_slot = self._request_slots[slot].get(_REMOTE_INTERACTION_TYPE_KEY)
+        if interaction_code is None:
+            interaction_slot.fill_(_NO_INTERACTION_TYPE_CODE)
+        else:
+            interaction_slot.copy_(interaction_code)
+        self._response_event.clear()
+        self._response_status[slot] = 0
+        self._submitted_at[slot] = time.monotonic()
+        self._request_ready[slot] = 1
```
vmoens, line 121: Lock-free flag: the tensor stores above are ordered before this store only for a reader that synchronised through the semaphore, and `_run` also drains after a timed-out wait, so on ARM the flag can in principle be seen before the payload. Trust flags only up to the number of acquired signals, or document that correctness relies on the semaphore barrier plus the timed drain.

torchrl/modules/inference_server/_process_slot.py
```diff
@@ -0,0 +1,400 @@
...
+        self._response_slots = self._make_slots(response_spec, "response_spec")
+        self._response_keys = list(
+            response_spec.keys(include_nested=True, leaves_only=True)
+        )
+
+        self._request_ready = self._ctx.Array("b", num_slots, lock=False)
+        self._submitted_at = self._ctx.Array("d", num_slots, lock=False)
+        self._response_status = self._ctx.Array("b", num_slots, lock=False)
+        self._response_events = [self._ctx.Event() for _ in range(num_slots)]
+        self._exception_queues = [self._ctx.SimpleQueue() for _ in range(num_slots)]
```
vmoens, line 284: A `SimpleQueue` per slot is a pipe (2 fds) plus 2 locks inherited by the server and the driver, on top of 4 semaphores per `Event`: about 130 fds at 64 envs, above the default `ulimit -n` at 1024. One shared exception queue carrying `(slot, exc)` is enough.

PR #4304, opened 2026-09-08
[Feature] Serve the DreamerV3 acting policy from environment processes

vmoens:
## Description

Adds `collector.inference_backend` to the DreamerV3 SOTA example so the acting policy can be served from a dedicated inference process instead of a thread of the training process.

- `thread` (default, unchanged): the `InferenceServer` runs in a thread of the training process and coordinator threads relay observations and actions between the environment workers and the server.
- `process`: `AsyncBatchedCollector` is built around a `ProcessSlotTransport`. A dedicated inference process reads each environment worker's shared-memory request slot and writes the action back into it, so no observation or action crosses the training process on the action path; the driver only receives completed transitions. Requires `collector.async_env_backend=multiprocessing` and `collector.envs_per_worker=1`; `collector.env_exchange` is owned by the transport.

Implementation notes:

- `dreamer_v3_agent.build_serving_policy` is a picklable factory (used through `functools.partial` over the resolved config container) that rebuilds the real-world actor inside the server process, including the `DreamerV3SeededPolicy` wrapper when `optimization.separate_policy_rng` is set, so the weight layout matches `collector.update_policy_weights_`.
- `_inference_slot_specs` derives the fixed request/response slot layout from one primed environment's `fake_tensordict()` and a single policy pass.
- The process server only starts with the first collector iteration, so the learner's weights are pushed right after the first collected batch, then after every training batch as before. The periodic sync is factored into `_behavior_policy_weights`.
- README and config document the new key.

## Motivation and Context

Profiling the async DreamerV3 stack on a 64-environment pixel task showed the driver process on the critical path of every environment step (about 2.5 ms of Python per transition across six hand-offs) while the inference server itself was idle most of the time. Serving the policy from the environment side removes the driver from the action path. The mode is opt-in; the default behavior is unchanged.

## Benchmark

DreamerV3 (196M) on a 64-environment pixel task, one 4-GPU node, `train_ratio=16`, otherwise identical configuration (compiled learner step, CUDA-graph static policy batches of 64). Throughput over the same elapsed window (1000 s to 1950 s after training start):

| acting policy served from | steps/s | learner updates/s |
|---|---|---|
| thread server, 4 envs per worker, `shm` exchange (compiled step) | 322 | 0.63 |
| thread server, 4 envs per worker, `shm` exchange (scan, eager step) | 301 | 0.59 |
| process server (`inference_backend=process`), 1 env per worker | 371 | 0.73 |

The learner cadence follows collection through the train ratio in all three runs, so the gain is on the collection path. Note: all three runs used CUDA-graph static policy batches, which turned out to break learning for this policy regardless of the serving backend; see the last section.

## Types of changes

## Checklist

## Tests

- New `test_dreamer_v3_process_inference_collection_smoke[False|True]` runs the example end to end on CPU with `inference_backend=process` (2 environments, multiprocessing backend, one env per worker), for both `separate_policy_rng` settings, and checks the validation error for `envs_per_worker != 1`.
- Existing `test_dreamer_v3_native_replay_collection_smoke` sync/async variants pass unchanged.

## Learning regression (resolved 09 Sep: not the process backend)

On a 64-environment pixel task the example collapsed after ~2M environment steps with this backend and with the thread server alike. The cause was `inference_static_batch_size` (a CUDA graph around the acting policy, which the process server forwards to its child): the DreamerV3 acting policy rewrites its `state` and `belief` input keys, which rebinds the graph's static inputs during capture, so the served policy ignored its recurrent inputs and acted from the reset state at every step. PR #4310 makes `InferenceServer` refuse that combination. With `inference_static_batch_size=null` the example learns normally (eager thread server: episode score 4.0-5.7 at 2-5.5M steps); this process backend without static batches (chunked worker-owned replay, 1,150-1,230 environment steps/s) learns the same curve: episode score 3.9 / 4.2 / 4.4 at 3-4.5M steps.

### DIFF
--- a/sota-implementations/dreamer_v3/README.md
+++ b/sota-implementations/dreamer_v3/README.md
@@ -69,6 +69,21 @@ synchronous `Collector`. `collector.async_env_backend` selects `threading` or
 post-processing, episode reporting, device normalization, and replay writes
 happen through the collector's standard replay integration.
 
+`collector.inference_backend` selects where the acting policy is served.
+`thread` (the default) runs the inference server in a thread of the training
+process, and coordinator threads relay observations and actions between the
+environment workers and the server. `process` starts a dedicated inference
+process that reads each environment worker's shared-memory request slot and
+writes the action back into it, so no observation or action crosses the
+training process; the driver only receives completed transitions. It requires
+`collector.async_env_backend=multiprocessing` and `collector.envs_per_worker=1`,
+uses the worker exchange owned by the transport (`collector.env_exchange` is
+ignored), and takes `collector.inference_static_batch_size` like the thread
+server. The policy is rebuilt from the configuration inside the server process
+and receives the learner's weights through the collector's policy-update path:
+the server starts with the first collected batch, receives the learner's
+weights right after it, and again after every training batch.
+
 For a three-seed median and interquartile reproduction run:
 
 ```bash
--- a/sota-implementations/dreamer_v3/config.yaml
+++ b/sota-implementations/dreamer_v3/config.yaml
@@ -24,6 +24,12 @@ collector:
   inference_min_batch_size: 1
   inference_timeout: 0.01
   inference_static_batch_size: null
+  # Where the acting policy is served. "thread": the inference server runs in a
+  # thread of this process and coordinator threads relay observations and
+  # actions. "process": a dedicated inference process reads each environment
+  # worker's shared-memory slot directly, so nothing crosses the driver on the
+  # action path; needs async_env_backend=multiprocessing and envs_per_worker=1.
+  inference_backend: thread
   num_envs: 1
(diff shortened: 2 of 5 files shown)


vmoens:
Superseded by #4313, which moves the process-slot wiring into the collector (`transport='auto'`) and makes it the example's default where it applies. The review requests on this PR landed there: weights pushed before the server starts are applied before the first request, so batch 0 uses the learner's (restored) weights; a layout test pins `_behavior_policy_weights` against the rebuilt policy; the served policy is created directly on its device and the smoke test has a CUDA variant for the GPU jobs; the dead `env_exchange='queue'` requirement is gone. The note that CUDA-graph static batches broke learning is kept in the new PR's body. I suggest closing this one once #4313 merges.

PR #4305, opened 2026-09-08
[Performance] Chunk process-slot worker results in AsyncBatchedCollector

vmoens:
Stack: #4305 → #4306 → #4307 → #4308 (base).

Batch worker transitions to reduce queue, unpickling, and replay-write overhead.

```python
collector = AsyncBatchedCollector(
    ...,
    transition_chunk_size=64,  # ProcessSlotTransport only; default: 1
)
```

Chunks preserve stream order and frame budgets. Larger chunks increase latency and in-flight memory; batches become dense with tensor `env_index`.

Coverage: ordering/budgets, replay routing, trajectories, pause/shutdown. Reported CPU benchmark: ~6× less driver CPU/frame.

### DIFF
--- a/benchmarks/ASYNC_BENCHMARKS.md
+++ b/benchmarks/ASYNC_BENCHMARKS.md
@@ -62,6 +62,16 @@ matches `async-process-slots`. Each point records its configuration in the
 changed. Use the fixed `async-process-slots*` series to separate the individual
 changes.
 
+`async-process-slots-chunked` is the same direct-transport workload with
+`transition_chunk_size=64`: each environment process sends 64 consecutive
+transitions per message instead of one, so the driver concatenates whole chunks
+into each 256-transition batch rather than unpickling and stacking transitions
+one at a time. Compare it with `async-process-slots` to see what the driver's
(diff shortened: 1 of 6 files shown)


vmoens:
```text
Measured locally:     0.226 -> 0.037 ms/frame (~6x less driver CPU)
Projected collection: ~1.5x; ceiling ~370 -> ~1,000 steps/s
```

The collection estimates are unmeasured GPU projections. Reaching the ceiling requires sufficient learner throughput (#4307).

vmoens:

|---|---|---|
| speed-up | 3.46x | 3.46x |

vmoens:
Learning check (09 Sep): on the 64-environment task the stack keeps its throughput without CUDA-graph inference batches (1,234 env steps/s and 2.41 updates/s with `inference_static_batch_size=null`, against 1,249 with them) and the learning curve then matches the previous example (episode score 3.9 / 4.2 / 4.4 at 3-4.5M environment steps). The collapse seen with static batches is explained and guarded in #4310 (the acting policy rewrites its `state`/`belief` input keys, which a captured CUDA graph cannot follow).

PR #4306, opened 2026-09-08
[Performance] Serve process-slot inference passes from pinned staging batches

vmoens:
Stack: #4305 → #4306 → #4307 → #4308; depends on #4305.

Batch process-slot inference through reusable pinned buffers, non-blocking transfers, and one CUDA event per pass. Gather/scatter replaces per-request copies; only response keys return to CPU.

```python
server = InferenceServer(
    policy,
    transport=transport,  # ProcessSlotTransport; batched I/O is automatic
    policy_device="cuda",
)
```

Custom collate functions retain their existing path.

Validation: CPU inference suite and metadata regression tests pass. CUDA validation and throughput measurements pending.

### DIFF
--- a/benchmarks/bench_inference_server.py
+++ b/benchmarks/bench_inference_server.py
@@ -1,0 +1,191 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""Throughput benchmark for the process-hosted inference server.
+
+Measures requests per second served by a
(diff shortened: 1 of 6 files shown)


vmoens:
```text
Projected server pass cost: -40–50%
Projected server ceiling:   ~1,000 -> ~1,800–2,000 steps/s
```

These CUDA estimates remain unmeasured. No direct end-to-end gain is expected until #4305 removes the driver bottleneck.

vmoens:

|---|---|---|
| speed-up | 3.46x | 3.46x |

PR #4307, opened 2026-09-08
[Performance] Keep the DreamerV3 replay write-back off the learner stream

vmoens:
Stack: #4305 → #4306 → #4307 → #4308; depends on #4306.

Enqueue CUDA→CPU replay copies at submission; the update thread waits on CUDA events, avoiding copies queued behind the next learner step. Also deduplicate coordinates and group ensemble writes with stable sorts.

```python
index, generation, patch = replay_context_update(sample, state, belief)
future = replay_buffer.submit_update_if_present(
    index=index, generation=generation, patch=patch
)
```

Validation: 28 focused CPU replay/Dreamer tests pass. CUDA validation pending.

### DIFF
--- a/benchmarks/ad_hoc/bench_dreamer_v3_learner.py
+++ b/benchmarks/ad_hoc/bench_dreamer_v3_learner.py
@@ -12,6 +12,7 @@ native replay construction, prefetch, and conditional writeback as training.
 Example::
 
     python benchmarks/ad_hoc/bench_dreamer_v3_learner.py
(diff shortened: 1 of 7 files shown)


vmoens:
```text
Projected GPU: ~1.4 -> 0.7–0.9 s/update (~1.5–2x faster)
Measured CPU:  overall neutral; dedup 6% faster, write-back 17% faster
```

The GPU estimates remain unmeasured on the profiled node. Faster learner updates should raise collection throughput at a fixed train ratio.

vmoens:

|---|---|---|
| speed-up | 3.46x | 3.46x |

PR #4308, opened 2026-09-08
[Performance] DreamerV3: freeze the setup heap before the training loop

vmoens:
Stack: #4305 → #4306 → #4307 → #4308; depends on #4307.

Freeze the DreamerV3 setup heap so cyclic GC stops rescanning it during training.

```python
# After learner, collector, and replay setup:
gc.freeze()
try:
    for _ in collector:
        ...
finally:
    gc.unfreeze()
```

Reported validation: CPU SOTA smoke and lint pass; local reproduction reduces GC time ~8×. Profiled GPU configuration still needs validation.

### DIFF
--- a/sota-implementations/dreamer_v3/train.py
+++ b/sota-implementations/dreamer_v3/train.py
@@ -1202,6 +1202,8 @@ def main(cfg: DictConfig):
             },
         )
     finally:
+        # Let setup cycles be collected between Hydra multirun jobs.
+        gc.unfreeze()
         for signum, handler in previous_handlers.items():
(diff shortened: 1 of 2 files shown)


vmoens:
```text
Measured local GC time:      0.208 -> 0.025 s (~8x less)
Measured local receive loop: 165 -> ~108 us/transition
```

Projected end-to-end gain is at most a few percent and remains unmeasured on the GPU node. The benefit should shrink after #4305 reduces unpickling.

vmoens:

|---|---|---|
| speed-up | 3.46x | 3.46x |

PR #4310, opened 2026-09-08
[BugFix] Async DreamerV3: explicit exploration, stored episode flags, cross-episode replay sequences

vmoens:
Three fixes for asynchronous DreamerV3 training with `AsyncBatchedCollector` and the inference server, plus a regression test for CUDA-graph inference batches with recurrent policies.

## 1. Async requests were stamped with the process-wide interaction type

```python
# learner thread: every LossModule.forward sets the process-wide context (objectives/common.py)
with set_interaction_type(InteractionType.DETERMINISTIC):
    loss = loss_module(batch)

# collector threads, concurrently (PolicyClientModule.submit, before this PR)
request["interaction_type"] = interaction_type()  # DETERMINISTIC or RANDOM depending on timing

server.serve([request_a, request_b])
# RuntimeError: InferenceServer received a mixed interaction-type batch
# (with a captured CUDA graph: "requires the interaction type used during capture")
```

Fix: the collector owns the mode and stamps every request with it; CUDA capture uses the same explicit mode. Standalone ambient-mode behavior is preserved.

```python
collector = AsyncBatchedCollector(
    env_fns, policy=policy, frames_per_batch=256,
    exploration_type=ExplorationType.RANDOM,  # default
)
```

## 2. Slice samplers marked every sequence start as an episode start

```python
# SliceSampler._maybe_emit_init_marker (meant for recurrent modules that restart from a stored state)
info["is_init"] = slice_start | stored_is_init
sample.update(info)  # ReplayBufferEnsemble merges it into the sample

# RSSMRolloutV3 reads is_init as an episode reset
state = torch.where(is_init, 0, state)
belief = torch.where(is_init, 0, belief)
action = torch.where(is_init, 0, action)
# every training sequence started from a zero state with a zeroed first action;
# the stored warm start and its write-back never applied
```

Fix: `init_key` on `SliceSampler`, `StreamingSliceSampler`, `PrioritizedSliceSampler` and `SliceSamplerWithoutReplacement` (mirrored in the Hydra configs). The default `"is_init"` keeps the marker; `None` returns the stored flags untouched, which the DreamerV3 example uses.

```python
StreamingSliceSampler(slice_len=seq_len + 1, end_key=("next", "done"), init_key=None)
```

## 3. Terminal transitions never reached the learner

```python
sampler = StreamingSliceSampler(slice_len=seq_len + 1, end_key=("next", "done"))  # strict_length=True
sample = rb.sample().reshape(B, seq_len + 1)[:, :-1]  # the extra record only addresses write-back rows
# a slice never crosses a done flag, so a terminal record can only be the dropped last slot;
# episodes shorter than seq_len + 1 are never sampled at all
```

Fix: with async collection each stream holds one environment, so the stream is the trajectory: sequences may span episode ends and the rollout resets on the stored `is_init`.

```python
StreamingSliceSampler(
    slice_len=seq_len + 1, end_key=("next", "done"), traj_key="env_index", init_key=None
)
```

`ReplayBufferEnsemble.end_streams` (checkpoint resume) used to refuse every trajectory-keyed sampler because restarted producers may reuse ids. It now accepts a member whose records all carry one trajectory id, since reuse is then by construction and the tail is closed as for end-key samplers; members holding distinct ids still raise.

## 4. CUDA-graph inference batches ignored the inputs of a policy that rewrites its input keys

`tensordict.nn.CudaGraphModule` used to capture on a shallow copy of the input tensordict. A module that sets an output under one of its input keys replaced that entry of the static input during capture; replays then copied the new inputs into the replaced tensor (immediately overwritten by the graph) while the captured kernels kept reading the original buffer.

```python
class Recurrent(nn.Module):
    def forward(self, x, h):
        h = torch.tanh(h + x)
        return h.sum(-1, keepdim=True), h

td = lambda h: TensorDict({"x": torch.ones(4, 3), "h": torch.full((4, 3), h)}, [4]).to("cuda")
policy = TensorDictModule(Recurrent(), in_keys=["x", "h"], out_keys=["y", "h"])  # rewrites "h"
graphed = CudaGraphModule(policy, warmup=2, device="cuda:0")
for _ in range(3):
    graphed(td(0.0))
graphed(td(1.0))["y"] == graphed(td(-1.0))["y"] == policy(td(0.0))["y"]  # h was ignored
```

`RSSMStateEstimatorV3` writes `state` and `belief` under its input keys, so with `static_batch_size` the served DreamerV3 policy ignored `state`, `belief` and `previous_action` and acted from the reset state at every step (its belief equalled the direct call with `is_init=True`). Training degraded after about 2M environment steps, while the eager server learned normally.

The capture is fixed in tensordict (replays copy new inputs into the leaves the graph reads), so this PR adds no workaround: `test_static_batch_serves_policy_rewriting_its_inputs` serves such a policy through `InferenceServer(static_batch_size=...)` and checks every request against the eager module. The test needs a tensordict build with that fix; the next release pins one.

## Tests

- `test/test_inference_server.py`, `test/collectors`: explicit exploration stamping and capture; `test_static_batch_serves_policy_rewriting_its_inputs` (GPU).
- `test/rb/test_samplers.py`: `init_key=None` keeps the stored flags; a custom `init_key` receives the markers.
- `test/rb/test_ensemble.py`: `test_end_streams_with_per_stream_trajectory_key` (single-id members close, mixed-id members raise).
- `test/objectives/test_dreamer_v3.py`: `test_dreamer_v3_replay_samples_keep_episode_starts`, `test_dreamer_v3_async_replay_sequences_cross_episode_ends`; the checkpoint-resume test expects resets to follow closed records in async streams.

### DIFF
--- a/docs/source/reference/modules_inference_server.rst
+++ b/docs/source/reference/modules_inference_server.rst
@@ -135,9 +135,17 @@ In-place parameter copies preserve the captured graph; for example, use
 storage-replacing update raises an error and disables the graph, leaving the
 server on the safe eager path until it is stopped and prepared again. This
 avoids capturing while collector or environment threads are live. The
-interaction type is frozen to its effective value at capture time, including an ambient
-``set_interaction_type`` context; later requests using another type are
-rejected.
+interaction type is fixed at capture time: pass it to
+:meth:`~torchrl.modules.inference_server.InferenceServer.prepare_cudagraph`
+(:class:`~torchrl.collectors.AsyncBatchedCollector` passes its
+``exploration_type``), otherwise the graph is captured under the policy's
+default interaction type. The ambient ``set_interaction_type`` context is never
+consulted, neither at capture nor when serving: it is process-wide state that
+another thread, typically the learner's loss forward, may change at any time.
+Requests must carry the captured mode, which
+:class:`~torchrl.modules.inference_server.PolicyClientModule` stamps from its
+``interaction_type`` argument or, when none is given, from the caller's active
+context; requests stamped with another mode are rejected.
 
 CUDA operations using PyTorch's default generator advance its graph-safe state
 across replays. Custom generators must manage CUDA graph state explicitly, and
--- a/docs/source/reference/modules_inference_server.rst
+++ b/docs/source/reference/modules_inference_server.rst
@@ -138,10 +138,11 @@ avoids capturing while collector or environment threads are live. The
 interaction type is fixed at capture time: pass it to
 :meth:`~torchrl.modules.inference_server.InferenceServer.prepare_cudagraph`
 (:class:`~torchrl.collectors.AsyncBatchedCollector` passes its
-``exploration_type``), otherwise the graph is captured under the policy's
-default interaction type. The ambient ``set_interaction_type`` context is never
-consulted, neither at capture nor when serving: it is process-wide state that
-another thread, typically the learner's loss forward, may change at any time.
+``exploration_type``). Without an explicit mode, capture uses the mode stamped on
+the request specification, otherwise the ambient ``set_interaction_type``
+context, or the policy default when no context is active. Raw, unstamped requests
+also retain their ambient-context behavior. Pass an explicit mode when other
+threads may change this process-wide context.
 Requests must carry the captured mode, which
 :class:`~torchrl.modules.inference_server.PolicyClientModule` stamps from its
 ``interaction_type`` argument or, when none is given, from the caller's active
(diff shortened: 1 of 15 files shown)


vmoens:
CI on 289d579da: 89 checks pass; the only failure is tests-cpu (3.13, bulk), where `test/modules/test_dreamer_components.py::TestDreamerV3Components::test_block_gru_torch_compile` fails. The same job fails on main's latest run of that workflow (https://github.com/pytorch/rl/actions/runs/34226083772), so it is unrelated to this PR.

PR #4312, opened 2026-09-09
[Test] Baseline benchmarks for async collectors and DreamerV3

vmoens:
Establish continuous baselines before #4304, #4305, #4306, #4307 and #4308 merge.

- Add an integrated process-slot collection series that adopts static inference and chunked results as those options become available.
- Benchmark the DreamerV3 training loop on a fixed pixel workload, with low/high training ratios and thread/process inference. Process inference skips until the example supports it.
- Pin the training configuration independently of example defaults, trigger benchmarks on DreamerV3 changes, add the required Hydra dependencies, and retry snapshot package installation failures.

The training series uses an eager CUDA learner and inference batches of one to avoid the current mixed exploration-context failure. Batched inference remains covered by the collector series. Startup, warm-up and shutdown are excluded from timing. Workload and measurement details are in [`benchmarks/ASYNC_BENCHMARKS.md`](https://github.com/pytorch/rl/blob/async-collector-benchmark/benchmarks/ASYNC_BENCHMARKS.md).

Validation: 10 reporting tests, formatting/lint, environment checks, and a CPU `thread-ratio2` harness run with five measured rounds, 20 learner updates and graceful shutdown. [CUDA validation is running](https://github.com/pytorch/rl/actions/runs/34331912977).

Merge this first, then wait for a successful main-branch baseline before merging the follow-ups.

### DIFF
--- a/.github/async-benchmark-requirements.txt
+++ b/.github/async-benchmark-requirements.txt
@@ -1,6 +1,7 @@
 # Pinned async trend v1 dependencies: CPython 3.10, Linux x86-64, CUDA 12.6.
 # Use stable wheels so a cache eviction cannot depend on expired nightly wheels.
 # TensorDict source is pinned in benchmarks.yml; bump the trend version with this lock.
+antlr4-python3-runtime==4.9.3
 cloudpickle==3.1.2
 cmake==4.4.0
 cuda-bindings==12.9.7
@@ -10,6 +11,7 @@ exceptiongroup==1.3.1
 filelock==3.32.5
 fsspec==2026.7.0
 hoptorch==0.1.4
+hydra-core==1.3.6
 importlib-metadata==9.0.0
 iniconfig==2.3.0
 jinja2==3.1.6
@@ -33,6 +35,7 @@ nvidia-nccl-cu12==2.29.3
 nvidia-nvjitlink-cu12==12.9.86
 nvidia-nvshmem-cu12==3.4.5
 nvidia-nvtx-cu12==12.6.77
+omegaconf==2.3.1
 orjson==3.11.9
 packaging==26.2
 pluggy==1.6.0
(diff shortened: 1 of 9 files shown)


vmoens:
Branch validation run with the current head (skip-upload): https://github.com/pytorch/rl/actions/runs/34331912977. Reporting validation, CPU and GPU benchmark jobs and the summary step all passed; 31 series per repeat on GPU, three fresh-process repeats.

New series on GPU (median of the three run medians, run range, execution recorded in the JSON):

| Series | frames/s | Run range | Execution |
| --- | ---: | ---: | --- |
| `test_async_collection_pixels[async-process-slots-integrated-uniform]` | 991.7 | 989.5-998.5 | graph; 1 envs/worker; process acting |
| `test_async_collection_pixels[async-process-slots-integrated-slow-reset]` | 995.2 | 991.9-996.1 | graph; 1 envs/worker; process acting |
| `test_dreamer_v3_async_training[thread-ratio2]` | 85.2 | 84.8-85.6 | eager learner on cuda:0; thread inference; train ratio 2; 8 envs; inference batch <= 1; 0.33 learner updates/s |
| `test_dreamer_v3_async_training[thread-ratio16]` | 53.1 | 53.1-53.7 | same with train ratio 16; 1.66 learner updates/s |

`async-process-slots-uniform` (eager) measured 995.9 frames/s in the same run, so the integrated series matches the fixed one on this revision as intended. Run-to-run spread of the new series is about 1 percent. The `process` DreamerV3 series skipped as designed (no `collector.inference_backend` on this revision).

Cost: each pytest repeat took 17.3 minutes on the GPU runner; the two DreamerV3 series account for 105 s (`ratio2`) and 156 s (`ratio16`) of that, about 13 minutes per job over the three repeats. Startup and shutdown of the example are about 20 s per test; the rest is the seven 1,024-step rounds. With inference limited to one request per batch, the `ratio2` series moves about 85 frames/s across 8 environments, that is roughly 90 ms per environment step, so its level is set by per-request policy latency rather than by the batched pipeline.

PR #4315, opened 2026-09-09
[Performance] Send at least 64 transitions per worker message by default

vmoens:
## Description

`transition_chunk_size="auto"` now resolves to at least 64 consecutive transitions per environment-worker message with process slots, `frames_per_batch // num_envs` when that is larger, and never more than `frames_per_batch`. It resolved to `frames_per_batch // num_envs` before, chosen so that each batch held one contiguous run per environment and a transition waited at most one batch.

The driver's work per message (unpickling, concatenation, the routed replay write) competes with training for the interpreter, so the message count matters more than that layout. Two otherwise identical runs of the DreamerV3 example on the 64-environment pixel task (1,024-frame batches, process slots, compiled learner, train ratio 16), started within a minute of each other on identical hardware:

| `transition_chunk_size` | Steps/s (last window, 38 min in) | Learner updates/s |
| --- | ---: | ---: |
| `auto` = 16 (previous rule) | 1,012 | 1.98 |
| 64 | 1,227 | 2.40 |

The 64-transition run matches the throughput measured for the performance stack before it merged (1,252 steps/s, 2.45 updates/s). The gap was the same at the 30-minute status.

What changes for users:
- Delivery latency: a transition reaches the driver about 64 environment steps after it was produced instead of one batch's worth. With a replay buffer, which is how this path is used, that is irrelevant; consumers that iterate over batches and need step-level freshness pass `transition_chunk_size=1`.
- Batch composition: a batch may now contain runs from a subset of the environments rather than one run per environment. Replay routing is per environment stream, so replay users see no difference.
- Small batches are unaffected: the message never exceeds `frames_per_batch`.

Docstring, the collector documentation, the DreamerV3 README and config comment state the new rule and the trade-off. The two tests that asserted the resolved value are updated (they exercise the cap), and `test_auto_transition_chunk_size_floor_and_cap` checks the floor (8 environments x 256 frames: previous rule 32, resolved 64), a per-environment run above the floor (2 x 256: 128) and the cap (2 x 12: 12).

The class docstring gains a second example showing the tuned configuration end to end: a picklable, device-aware policy factory, `transport="auto"` with multiprocessing workers, `transition_chunk_size=64`, explicit server and device configurations, the resolved `server_backend`, a weight push before the first iteration, and the same layouts built by hand with `ProcessSlotTransport`.

## Motivation and Context

Follow-up to #4313. The default should be the fast configuration; the benchmark above shows the previous rule left about 17% of collection throughput unused on the workload the process-slot path was built for.

## Types of changes

- [x] Performance (default value change of an unreleased option)

## Checklist

### DIFF
--- a/test/test_inference_server.py
+++ b/test/test_inference_server.py
@@ -4224,7 +4224,8 @@ class TestAsyncBatchedCollector:
         try:
             assert collector._uses_process_env_workers
             assert collector.server_backend == "process"
-            assert collector._transition_chunk_size == 12 // num_envs
+            # At least 64 per message, capped at one batch.
+            assert collector._transition_chunk_size == 12
             frames = 0
             for batch in collector:
                 frames += batch.numel()
@@ -4360,7 +4361,7 @@ class TestAsyncBatchedCollector:
         try:
             assert collector.server_backend == "process"
             assert collector._env_backend == "multiprocessing"
-            assert collector._transition_chunk_size == 2
+            assert collector._transition_chunk_size == 4
             assert sum(batch.numel() for batch in collector) == 8
         finally:
             collector.shutdown()
--- a/test/test_inference_server.py
+++ b/test/test_inference_server.py
@@ -4349,6 +4349,33 @@ class TestAsyncBatchedCollector:
         finally:
             collector.shutdown()
 
+    @pytest.mark.parametrize(
+        "num_envs, frames_per_batch, expected",
+        [
+            # frames_per_batch // num_envs is 32: the floor of 64 applies.
+            (8, 256, 64),
+            # One run per environment and batch is longer than the floor.
+            (2, 256, 128),
+            # A batch shorter than the floor caps the message at one batch.
+            (2, 12, 12),
+        ],
+    )
+    def test_auto_transition_chunk_size_floor_and_cap(
+        self, num_envs, frames_per_batch, expected
+    ):
+        collector = AsyncBatchedCollector(
+            create_env_fn=[_counting_env_factory] * num_envs,
+            policy_factory=_make_counting_policy,
+            transport="auto",
+            frames_per_batch=frames_per_batch,
+            env_backend="multiprocessing",
+        )
+        try:
+            assert collector._uses_process_env_workers
+            assert collector._transition_chunk_size == expected
+        finally:
+            collector.shutdown()
+
     def test_process_slot_transport_implies_process_server_and_workers(self):
         """An explicit ProcessSlotTransport needs no server or env backend settings."""
         collector = AsyncBatchedCollector(
--- a/torchrl/collectors/_async_batched.py
+++ b/torchrl/collectors/_async_batched.py
@@ -760,6 +760,87 @@ class AsyncBatchedCollector(BaseCollector):
         ...     print(batch.shape)
         ...     break
         >>> collector.shutdown()
+
+        The fast configuration for many process-backed environments serves the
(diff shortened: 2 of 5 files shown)


torchrl/collectors/_async_batched.py
```diff
@@ -753,6 +760,87 @@ class AsyncBatchedCollector(BaseCollector):
...
+        ... )
+        >>> collector.server_backend
+        'process'
+        >>> learner_policy = make_policy("cuda:0")  # the copy being trained
+        >>> collector.update_policy_weights_(learner_policy)
+        >>> for batch in collector:
+        ...     print(batch.shape)  # dense: whole 64-step chunks, env_index is a tensor
+        ...     break
+        torch.Size([1024])
+        >>> collector.shutdown()
```
vmoens, line 813: @theap06 this seems to be working pretty well for dreamer v3 -- getting some really good throughput in the collector with async envs now

vmoens:
Final numbers from the two paired runs (both ended after 9.1 hours of training when their node was reclaimed; both exited cleanly and learned normally throughout):

| `transition_chunk_size` | Steps/s, whole run | Updates/s, whole run | Steps/s, steady windows | Environment steps in 9.1 h |
| `auto` = 16 (previous rule) | 1,018 | 1.94 | 1,000 to 1,033 | 33.3M |
| 64 (this PR's `auto`) | 1,219 | 2.33 | 1,198 to 1,245 | 39.9M |

The gap stayed between 17% and 20% in every half-hourly window over the nine hours. The 64-transition run matches the throughput of the performance stack measured before it merged (1,252 steps/s, 2.45 updates/s).

test/test_inference_server.py
```diff
@@ -4224,7 +4224,8 @@ def test_auto_transport_serves_from_environment_processes(
         try:
             assert collector._uses_process_env_workers
             assert collector.server_backend == "process"
-            assert collector._transition_chunk_size == 12 // num_envs
+            # At least 64 per message, capped at one batch.
+            assert collector._transition_chunk_size == 12
```
theap06, line 4228: This assert hits the cap branch that this pr introduces.

test/test_inference_server.py
```diff
@@ -4360,7 +4361,7 @@ def test_process_slot_transport_implies_process_server_and_workers(self):
         try:
             assert collector.server_backend == "process"
             assert collector._env_backend == "multiprocessing"
-            assert collector._transition_chunk_size == 2
+            assert collector._transition_chunk_size == 4
```
theap06, line 4391: ditto

theap06 (changes requested):
The results are awesome. left a couple of small comments regarding the tests with the new cap rule.

test/test_inference_server.py
theap06: Both updated asserts have frames_per_batch below 64, so the cap makes the result equal the batch and the floor is never exercised. One case like 8 envs x 256 (old rule 32, resolved 64) would catch this issue

theap06 (approved):

test/test_inference_server.py
vmoens: Right, both of those hit the cap. Added `test_auto_transition_chunk_size_floor_and_cap` in 8d409de43 with your 8 environments x 256 frames case (previous rule 32, resolved 64), plus 2 x 256 (128, the per-environment run above the floor) and 2 x 12 (12, the cap). The two collection tests keep their asserts as the cap case.

PR #4317, opened 2026-09-09
[BugFix] Clamp negative MultiCategorical projections

### LINKED ISSUE #4316: [BUG] MultiCategorical.project leaves negative values outside the spec
## Describe the bug

`MultiCategorical.project()` clips values above each category's upper bound, but leaves negative values unchanged. The result can remain outside the spec's domain, including when projection is used by a `SafeModule` with `safe=True`.

## To Reproduce

```python
import torch
from torchrl.data import MultiCategorical

spec = MultiCategorical([3, 2])
projected = spec.project(torch.tensor([-1, 4]))
print(projected)              # tensor([-1, 1])
print(spec.is_in(projected))  # False
```

## Expected behavior

Projection should return a value in the spec's domain, as documented by `TensorSpec.project()`. Here, clipping to both category bounds should produce `tensor([0, 1])`, consistent with unmasked `Categorical.project()`.

## System info

- TorchRL source checkout at `a2706e86d2ad7a81ad7df21e3a0587dddc65daa2`
- Windows, Python 3.12
- PyTorch `2.13.0+cpu`, TensorDict `0.14.1`

## Reason and Possible fixes

`MultiCategorical._project()` only applies the upper bound. Clamping negative values to zero before converting to the spec's dtype also prevents negative inputs from wrapping for unsigned output dtypes.

## Checklist

- [x] I have checked that there is no similar issue in the repo.
- [x] I have read the documentation.
- [x] I have provided a minimal working example to reproduce the bug.

yurekami:
## Description

Clamp negative values to zero in `MultiCategorical.project()` before converting to the output dtype. For example, projecting `[-1, 4]` with cardinalities `[3, 2]` now produces `[0, 1]`, which belongs to the spec, instead of `[-1, 1]`.

## Motivation and Context

The missing lower bound violates the documented projection contract and allows invalid negative actions through `SafeModule(safe=True)`. Applying the clamp before the dtype conversion also preserves boolean outputs and prevents unsigned wraparound.

Closes #4316.

- [x] I have raised an issue to propose this change.

## Types of changes

## Checklist

- [x] I have read the contribution guide.

## Validation

- `python -m pytest test/test_specs.py -k multi_discrete -q --tb=short` — 83 passed.
- The regression checks exact projected values and domain membership for scalar inputs, batched samples, heterogeneous cardinalities, and signed, floating, unsigned, and boolean output dtypes.
- Manually checked direct projection and a `SafeModule` using a nested action key.
- All applicable pre-commit hooks passed for the two changed files; the pinned formatter hooks ran under Python 3.11 for wheel compatibility.

### DIFF
--- a/test/test_specs.py
+++ b/test/test_specs.py
@@ -321,6 +321,23 @@ class TestRanges:
             projection[..., 0] = -1
         assert not ts.is_in(projection)
 
+    @pytest.mark.parametrize(
+        "dtype", [torch.float32, torch.int64, torch.uint8, torch.bool]
+    )
+    @pytest.mark.parametrize(
+        "nvec, value, expected",
+        [
+            (3, -2, 0),
+            ([3, 2, 4], [[-1, 1, 6], [2, -3, -4]], [[0, 1, 3], [2, 0, 0]]),
+            ([[2, 4], [3, 2]], [[-1, 8], [5, -2]], [[0, 3], [2, 0]]),
+        ],
+    )
+    def test_multi_discrete_project_out_of_bounds(self, dtype, nvec, value, expected):
+        spec = MultiCategorical(nvec, dtype=dtype)
+        projected = spec.project(torch.tensor(value))
+        torch.testing.assert_close(projected, torch.tensor(expected, dtype=dtype))
+        assert spec.is_in(projected)
+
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM thanks

PR #4323, opened 2026-09-10
[CI] Pin PyTorch matrix for the release workflow

vmoens:
## Summary

- default the manually dispatched release workflow to pytorch/test-infra release/2.14
- pass that release ref to every platform build invoked by the release run
- preserve the explicit pytorch_release override for future releases

## Scope

This changes only .github/workflows/release.yml. Ordinary pull-request, main, nightly, and release-branch push wheel workflows continue using their existing test-infra defaults.

## Motivation

The PyTorch 2.14 release matrix supports CUDA 12.6, 13.0, and 13.2. Pinning the release run to that matrix prevents it from inheriting CUDA 13.4 from test-infra main before matching PyTorch wheels exist.

## Validation

- parsed the edited workflow as YAML
- generated the release/2.14 Linux release matrix and confirmed CPU, cu126, cu130, and cu132 only
- ran git diff --check

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,10 +34,10 @@ on:
         required: true
         type: string
       pytorch_release:
-        description: 'PyTorch release branch (e.g., release/2.8)'
+        description: 'PyTorch release branch that defines the supported build matrix'
         required: false
         type: string
-        default: 'main'
+        default: 'release/2.14'
       skip_sanity_checks:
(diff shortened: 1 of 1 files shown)


vmoens:
Closing this approach: the repository default must remain test-infra main. Release-specific compatibility should be selected for the release branch or individual release invocation, not pinned globally in main.

PR #4329, opened 2026-09-11
[Feature] Deploy low-level policies for high-level RL

vmoens:
## Summary

Adds a generic controller deployment path for running a pretrained low-level policy behind a high-level TorchRL environment.

- `LowLevelController` adapts high-level decisions to low-level policy inputs and keeps recurrent state under the public controller namespace.
- `ClosedLoopMultiAction` executes several physical steps per high-level decision, refreshes feedback on every step, handles partial termination, and supports sum, mean, last, or stacked reward aggregation.
- Hydra configs, reference docs, a tutorial, and benchmarks cover the public surface.
- Collector coverage verifies that current and next controller state survive collection.

## What was missing and what is fixed

`MultiAction` repeats a precomputed action and cannot run a feedback policy or carry its recurrent state between physical steps:

```python
# Before: one low-level action is computed once, then replayed open-loop.
env = TransformedEnv(env, MultiAction(dim=steps))

# After: the controller observes every physical next state and publishes
# its recurrent state under ("controller", ...).
controller = LowLevelController(low_level_policy, decision_spec)
env = ClosedLoopMultiAction.from_env(env, controller, steps=steps)
```

The `TransformedEnv` inverse-next propagation and runtime `InitTracker` shapes in this diff are required controller-state plumbing: without them, recurrent next state is discarded or initialized with the wrong expanded batch shape. MicroDuck-specific integration, trainer construction, replay sampler checkpointing, and categorical logging remain split into focused pull requests.

## Tests

- `TestClosedLoopMultiAction`: 22 passed.
- `TestModuleConfigs::test_controller_deployment_from_config`: passed.
- Collector coverage verifies both current and next recurrent controller state.


torchrl/modules/tensordict_module/controllers.py
```diff
@@ -0,0 +1,203 @@
...
+    _owns_tensordict_primers = True
+
+    def __init__(
+        self,
+        policy: TensorDictModuleBase,
+        decision_spec: Composite,
+        *,
+        adapter: TensorDictModuleBase | Transform | None = None,
+        group_key: NestedKey | None = None,
+        state_key: NestedKey = "_controller",
```
theap06: heads up, `Collector._postproc` drops any key with a leading underscore so `_controller` never makes it into a collector batch (checked with and without compact_obs). the replay buffer part of the description isnt true as written, either make the default non-private or document that collectors drop it.

torchrl/envs/custom/mujoco/microduck_controller.py
```diff
@@ -0,0 +1,324 @@
...
+        observation = group["observation"]
+        group["observation"] = Unbounded(
+            (*observation.shape[:-1], observation.shape[-1] + self.num_skills),
+            dtype=observation.dtype,
+            device=observation.device,
+        )
+        return spec
+
+
+def microduck_skill_env(
```
theap06: theres no `reset_key` here so the end to end snippet in the description raises a TypeError, needs forwarding to `MicroDuckController`.

torchrl/envs/custom/mujoco/microduck_controller.py
```diff
@@ -0,0 +1,324 @@
...
+        )
+        primers[self.skill_path] = Categorical(
+            num_skills, shape=spec.shape, device=spec.device
+        )
+        super().__init__(primers, default_value=0, expand_specs=False)
+
+    def _step(self, td, next_td):
+        group = next_td if self.group_key is None else next_td[self.group_key]
+        source = td if self.group_key is None else td[self.group_key]
+        next_td[self.fallen_path] = td[self.fallen_path] | group["fallen"]
```
theap06: this reads `group["fallen"]` unconditionally but `MicroDuckEnv` doesnt expose `fallen` on this branch (only in #4330), so the preset cant wrap the real env and dies with a NoneType error on the first step. tests only use the mock env so this never got caught, would like a test against the real env here.

theap06 (changes requested):
core looks solid, the stepping and reward aggregation matched my own references and CI is green. two blockers: the `_controller` default doesnt survive the collector, and `microduck_skill_env` cant wrap a real MicroDuckEnv on this branch.

vmoens:
Addressed the existing review by narrowing this PR to the generic controller. The default recurrent-state namespace is now controller, and the collector integration test verifies that both current and next state survive collection. The MicroDuck-specific wrapper was extracted to #4392; it forwards reset_key and MicroDuckEnv now exposes its real fallen signal, with behavioral and native-backend tests.

vmoens:
@theap06 Both requested points are now covered:

- #4329 has collector coverage showing that current and next recurrent controller state survive collection.
- The MicroDuck-specific wrapper lives in the focused #4392, which now includes an end-to-end regression wrapping the fixture-backed `MicroDuckEnv` rather than only a synthetic environment.

I also updated the PR body to distinguish required generic controller-state plumbing from the MicroDuck and trainer work that was split out. Could you take another look?

PR #4338, opened 2026-09-11
[BugFix] Only forward the applicable SACLossConfig fields to each SAC variant

### LINKED ISSUE #4337: [BUG] SACLossConfig cannot be instantiated: fields of the other SAC variant are forwarded to the loss
## Describe the bug

`_make_sac_loss` passes every `SACLossConfig` field to the loss constructor, but `SACLoss` and `DiscreteSACLoss` accept different kwargs. Since #3693 added `action_space`, `num_actions` and `target_entropy_weight` to the config, instantiating it fails for both variants:

* `discrete=False` (the default): `SACLoss` receives `action_space`, `num_actions` and `target_entropy_weight`
* `discrete=True`: `DiscreteSACLoss` receives `value_network`, `action_spec`, `delay_actor` and `delay_value`

This also breaks `sota-implementations/sac_trainer`, whose config uses `loss@loss: sac`. Calling `instantiate(cfg.loss)` on the composed config raises the same error.

## To Reproduce

```python
from hydra.utils import instantiate
from torchrl.trainers.algorithms.configs.modules import (
    MLPConfig,
    TanhNormalModelConfig,
    TensorDictModuleConfig,
)
from torchrl.trainers.algorithms.configs.objectives import SACLossConfig

cfg = SACLossConfig(
    actor_network=TanhNormalModelConfig(
        network=MLPConfig(in_features=10, out_features=4, depth=2, num_cells=32),
        in_keys=["observation"],
        out_keys=["action"],
    ),
    qvalue_network=TensorDictModuleConfig(
        module=MLPConfig(in_features=12, out_features=1, depth=2, num_cells=32),
        in_keys=["observation", "action"],
        out_keys=["state_action_value"],
    ),
)
instantiate(cfg)
```

```bash
hydra.errors.InstantiationException: Error in call to target 'torchrl.trainers.algorithms.configs.objectives._make_sac_loss':
TypeError("SACLoss.__init__() got an unexpected keyword argument 'action_space'. Did you mean 'action_spec'?")
```

With `discrete=True` added to the same config:

```bash
hydra.errors.InstantiationException: Error in call to target 'torchrl.trainers.algorithms.configs.objectives._make_sac_loss':
TypeError("DiscreteSACLoss.__init__() got an unexpected keyword argument 'value_network'. Did you mean 'qvalue_network'?")
```

## Expected behavior

The config builds a `SACLoss` or a `DiscreteSACLoss`, and each constructor only receives the fields it accepts.

## System info

* TorchRL source checkout at `ed803e488dbfd0632453e2fbc304831d233540d8`
* Linux, Python 3.13
* PyTorch `2.11.0`, TensorDict `0.14.2`, hydra-core `1.3.6`

## Reason and Possible fixes

`_make_iql_loss` already handles the same situation: it pops `action_space` before building `IQLLoss` and `deactivate_vmap` before building `DiscreteIQLLoss`. `_make_sac_loss` needs the same per-variant filtering. I checked the other loss factories against their targets' `__init__` signatures and none of them have this problem. No test instantiates `SACLossConfig`, which is probably why it went unnoticed. I have a fix with a regression test and will open a PR.

## Checklist

yupengtang:
## Description

`_make_sac_loss` now drops the config fields the selected SAC variant does not accept, the same way `_make_iql_loss` does for IQL:

* `SACLoss`: `action_space`, `num_actions`, `target_entropy_weight`
* `DiscreteSACLoss`: `value_network`, `action_spec`, `delay_actor`, `delay_value`

The `SACLossConfig` docstring already said which fields only apply to the discrete variant; it now also names the ones that only apply to the continuous variant.

## Motivation and Context

Closes #4337.

`SACLossConfig` could not be instantiated for either variant, so the `sac` loss group failed with `SACLoss.__init__() got an unexpected keyword argument 'action_space'`, including in `sota-implementations/sac_trainer`.

Testing:

* `TestLossConfigs::test_sac_loss_config[False]` and `[True]` build both variants from the config, check that `num_qvalue_nets` reaches the loss and, for the discrete one, that `target_entropy` comes out as `0.5 * log(num_actions)` from `num_actions` and `target_entropy_weight`. Both fail on `main` with the errors from the issue and pass with this change.
* `pytest test/test_configs.py`: 258 passed, 67 skipped, 60 xfailed on this branch, against 256 passed on `main`. The same 3 prioritized sampler tests fail on both because my local build has no C++ extension (`SumSegmentTreeFp32 is not available`).
* `instantiate(cfg.loss)` on the composed `sota-implementations/sac_trainer/config/config.yaml` now returns a `SACLoss`.
* Pinned `ufmt`, `flake8`, `pydocstyle` and `codespell` on the changed files.

#4100 also edits `_make_sac_loss` (value transforms), in a different part of the function.

## Types of changes

## Checklist

### DIFF
--- a/test/test_configs.py
+++ b/test/test_configs.py
@@ -10,6 +10,7 @@ import dataclasses
 import enum
 import importlib.util
 import inspect
+import math
 import os
 import pkgutil
 import subprocess
@@ -2015,6 +2016,72 @@ class TestLossConfigs:
         assert loss.delay_actor
         assert "target_actor_network_params" in dict(loss.named_children())
         assert loss.max_importance_ratio == 100.0
+
+    @pytest.mark.parametrize("discrete", [False, True])
+    def test_sac_loss_config(self, discrete):
+        # SACLoss and DiscreteSACLoss accept different kwargs, so each variant
+        # must only receive the config fields that apply to it
+        from hydra.utils import instantiate
+        from torchrl.modules import ProbabilisticActor
+        from torchrl.modules.distributions import OneHotCategorical
+        from torchrl.objectives.sac import DiscreteSACLoss, SACLoss
+        from torchrl.trainers.algorithms.configs.modules import (
+            MLPConfig,
+            TanhNormalModelConfig,
+            TensorDictModuleConfig,
+        )
+        from torchrl.trainers.algorithms.configs.objectives import SACLossConfig
+
+        num_actions = 4
+        if discrete:
+            actor_network = ProbabilisticActor(
+                TensorDictModule(
+                    torch.nn.Linear(10, num_actions),
+                    in_keys=["observation"],
+                    out_keys=["logits"],
+                ),
+                in_keys=["logits"],
+                distribution_class=OneHotCategorical,
+            )
+            qvalue_network = TensorDictModuleConfig(
+                module=MLPConfig(
+                    in_features=10, out_features=num_actions, depth=2, num_cells=32
+                ),
+                in_keys=["observation"],
+                out_keys=["action_value"],
+            )
+        else:
+            actor_network = TanhNormalModelConfig(
+                network=MLPConfig(
+                    in_features=10, out_features=4, depth=2, num_cells=32
+                ),
+                in_keys=["observation"],
+                out_keys=["action"],
+            )
+            qvalue_network = TensorDictModuleConfig(
+                module=MLPConfig(in_features=12, out_features=1, depth=2, num_cells=32),
+                in_keys=["observation", "action"],
+                out_keys=["state_action_value"],
+            )
+        cfg = SACLossConfig(
+            actor_network=actor_network,
+            qvalue_network=qvalue_network,
+            discrete=discrete,
+            action_space="one-hot",
+            num_actions=num_actions,
+            target_entropy_weight=0.5,
+            num_qvalue_nets=3,
+        )
+
+        loss = instantiate(cfg)
+        assert loss.num_qvalue_nets == 3
+        if discrete:
+            assert type(loss) is DiscreteSACLoss
+            # target_entropy="auto" is derived from num_actions and target_entropy_weight
+            expected_entropy = 0.5 * math.log(num_actions)
+            assert loss.target_entropy.item() == pytest.approx(expected_entropy)
+        else:
+            assert type(loss) is SACLoss
 
     @pytest.mark.skipif(not _has_gymnasium, reason="Gymnasium is not installed")
     def test_dqn_loss_config(self):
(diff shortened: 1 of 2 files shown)


test/test_configs.py
```diff
@@ -2016,6 +2017,72 @@ def test_ppo_loss_config_delay_actor(self, loss_type):
         assert "target_actor_network_params" in dict(loss.named_children())
         assert loss.max_importance_ratio == 100.0
 
+    @pytest.mark.parametrize("discrete", [False, True])
+    def test_sac_loss_config(self, discrete):
+        # SACLoss and DiscreteSACLoss accept different kwargs, so each variant
+        # must only receive the config fields that apply to it
+        from hydra.utils import instantiate
+        from torchrl.modules import ProbabilisticActor
+        from torchrl.modules.distributions import OneHotCategorical
```
vmoens, line 2026: Non-blocking: move these unconditional TorchRL imports to module scope, per the repository import rules. The Hydra-dependent config imports can remain guarded/lazy, and the test can reuse the existing module-level `instantiate_config` alias.

test/test_configs.py
```diff
@@ -2016,6 +2017,72 @@ def test_ppo_loss_config_delay_actor(self, loss_type):
...
+                    in_keys=["observation"],
+                    out_keys=["logits"],
+                ),
+                in_keys=["logits"],
+                distribution_class=OneHotCategorical,
+            )
+            qvalue_network = TensorDictModuleConfig(
+                module=MLPConfig(
+                    in_features=10, out_features=num_actions, depth=2, num_cells=32
+                ),
```
vmoens, line 2049: Non-blocking: the regression coverage is useful, but the setup is disproportionate to the change (67 added test lines versus 6 added factory lines, roughly 11:1). Both branches repeat the Q-network construction; build that config once after selecting its input/output dimensions and keys. The stored-value-only `num_qvalue_nets` assertion can also be omitted: constructing both loss variants and checking the discrete entropy calculation provide the relevant regression signal. Keep both variants and the entropy assertion.

vmoens (approved):
The per-variant filtering matches the constructor signatures exactly and fixes both configuration paths without changing the supported SAC options. No blocking correctness findings.

Validation on 9230158:

- `python -m pytest test/test_configs.py::TestLossConfigs -q --tb=short`: 16 passed, including both new SAC cases.
- Re-ran the two new cases with only `_make_sac_loss` replaced in memory by its parent-commit implementation: both failed with the expected unexpected-keyword errors (`action_space` / `value_network`).
- Composed `sota-implementations/sac_trainer/config/config.yaml` and successfully instantiated `cfg.loss` as `SACLoss`.
- `git diff --check HEAD^ HEAD` passed.

Two non-blocking test-maintenance comments below. Full CI is still in progress; the local checks above used Python 3.12, PyTorch 2.11 on CPU.

Issue #4227, opened 2026-09-05
[BUG] GRPOLoss masking strategies produce incompatible shapes


YeonwooSung:
## Describe the bug

`GRPOLoss` is documented to support `masking_strategy` in `{"sft", "rlhf", "generic"}`, but the test that actually compares those strategies is skipped because of a shape mismatch. CI therefore will not catch a regression in the public loss when the mask strategy changes.

This issue asks for the intended output shape per strategy and for the test to be unskipped once the loss matches that contract. It is not a request to land a speculative fix.

## To Reproduce

The skip sits on `TestGRPOLoss.test_grpo_loss_with_real_models`:

```1398:1408:test/llm/test_llm_objectives.py
    @pytest.mark.skipif(not _has_vllm, reason="vllm not available")
    @pytest.mark.parametrize("masking_strategy", ["sft", "rlhf"])
    @pytest.mark.skip(
        reason="GRPOLoss shape mismatch between masking strategies - needs investigation"
    )
    def test_grpo_loss_with_real_models(
        self,
        vllm_instance,
        transformers_instance,
        masking_strategy,
    ):
```

The test builds tokens-mode input for `"sft"` and history-mode input for `"rlhf"` (lines 1413-1442), generates with `vLLMWrapper(..., return_log_probs=True)`, then constructs

```1469:1469:test/llm/test_llm_objectives.py
        loss_fn = GRPOLoss(actor_network=wrapper, masking_strategy=masking_strategy)
```

and expects a finite `result.loss_objective`.

On the loss side, `masking_strategy` selects different distribution helpers:

```609:620:torchrl/objectives/llm/grpo.py
            if self.masking_strategy == "sft" and hasattr(
                self.actor_network, "_get_sft_dist"
            ):
                dist = self.actor_network._get_sft_dist(tensordict)
            elif self.masking_strategy == "rlhf" and hasattr(
                self.actor_network, "_get_rlhf_dist"
            ):
                dist = self.actor_network._get_rlhf_dist(tensordict)
            elif self.masking_strategy == "generic" and hasattr(
                self.actor_network, "_get_generic_dist"
            ):
                dist = self.actor_network._get_generic_dist(tensordict)
```

The class docstring already warns that a mismatch with the advantage mask produces shape errors (`grpo.py:377-383`):

> The masking strategy must match the strategy used for advantage computation to avoid shape mismatches.

Because the only cross-strategy test is skipped, that warning is not enforced.

## Expected behavior

[...]

PR #4340, opened 2026-09-12
[Doc] Document the GRPOLoss masking strategy contract and add deterministic tests

n-dlms:
## Description

Documents the masking strategy contract of `GRPOLoss` in its docstring and adds a deterministic CPU test class that pins it down.

Context: issue #4227 reports a suspected shape mismatch between masking strategies. I reproduced the full matrix on CPU with `trl-internal-testing/tiny-Qwen2ForCausalLM-2.5` and the `Qwen/Qwen2.5-0.5B` tokenizer, fabricating the response tokens so that no generation engine is required. Where a supported combination runs, masks are boolean tensors of shape `(batch, T)` over the full padded sequence, the distribution log probability has the same shape, and a per trajectory advantage of shape `(batch, 1, 1)` broadcasts cleanly. I could not reproduce a shape bug in the loss reduction itself.

What the docstring note now states, and what the tests assert:

* `tokens` input: `sft` masks prompt positions using `("tokens", "prompt")`, `generic` masks padding using the attention mask, and `rlhf` uses a caller supplied `("masks", "all_assistant_mask")` when the tensordict carries one and raises a `ValueError` only when it does not. The tests compare the distribution mask against those expected tensors, so a drift in any selection fails in CI.
* `history` input: the assistant mask is computed from the chat template. `rlhf` uses it directly and `sft` falls back to it whenever `("tokens", "prompt")` is unavailable, which makes `sft` and `rlhf` select the same tokens in this mode.
* a batch 2 history input with heterogeneous prompts and fabricated assistant responses runs through `GRPOLoss.forward` with a finite loss for all three strategies, and the unreduced loss keeps the `(B, T, 1)` token layout the advantage broadcasts against.

No behavior is changed: the docstring documents current behavior, and the two open contract questions from my comment on #4227 (whether `sft` in history mode should warn about the fallback, and what error `rlhf` with `tokens` input should suggest) remain for the maintainers to decide.

## Correction to my earlier comment on #4227

In that comment I claimed the transformers wrapper history path fails at batch sizes greater than one. That was wrong, and the probe script I referenced was the cause: it built the `History` incorrectly, stacking prompt and response as separate conversation slots instead of concatenated messages, which produced a malformed `(batch, 2, T)` token tensor. With a correctly built `History` (per row `History.from_chats`, extended along the message dimension with `dim=-1`, then lazy stacked) the wrapper handles batch 2 with heterogeneous prompts fine, including the assistant masks, for all three strategies. The wrapper batch path is therefore not implicated in the skipped integration test as I suggested.

## Test results

```
pytest test/llm/test_llm_objectives.py -k "TestGRPOLossMaskingContract or TestLosses"
13 passed in 42.73s
```

## Motivation and Context

Fixes the documentation side of #4227 and pins the contract with deterministic CPU tests. The skipped vLLM integration test is left untouched: its path involves actual generated output, which the fabricated response cases here do not exercise, so validating that integration remains a follow-up before its skip could be lifted.

## Types of changes

## Checklist

### DIFF
--- a/test/llm/test_llm_objectives.py
+++ b/test/llm/test_llm_objectives.py
@@ -1539,7 +1539,7 @@ def _grpo_masking_history_td(model, tokenizer):
         generate=False,
         return_log_probs=True,
         pad_output=True,
-        tokenizer_kwargs={"chat_template_name": "qwen"},
+        chat_template_name="qwen",
     )
     td = td.update(wrapper(td))
     log_probs = td.get(("log_probs", "full"), as_padded_tensor=True)
@@ -1551,10 +1551,12 @@ def _grpo_masking_history_td(model, tokenizer):
 class TestGRPOLossMaskingContract:
     """Deterministic CPU checks of the masking strategy contract.
 
-    Every supported combination of wrapper input mode and masking strategy must
-    run through GRPOLoss.forward with a finite loss, and ``rlhf`` with ``tokens``
-    input must fail with the assistant mask error. Response tokens are fabricated
-    so no generation engine is required (issue 4227).
+    Each supported combination of wrapper input mode and masking strategy must
+    select the tokens the contract promises, run through GRPOLoss.forward with
+    a finite loss, and keep the unreduced loss in the (B, T, 1) token layout.
+    Response tokens are fabricated so no generation engine is required
+    (issue 4227). The vLLM integration path of the skipped upstream test is not
+    covered by these cases.
     """
 
     @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


torchrl/objectives/llm/grpo.py
```diff
@@ -388,6 +388,23 @@ class GRPOLoss(LossModule):
...
         The masking strategy must match the strategy used for advantage computation to avoid shape mismatches.
 
+    .. note::
+        The mask actually selected depends on both the masking strategy and the input mode of the
+        actor network. For a transformers based actor the following holds:
+
+        * ``tokens`` input: ``"sft"`` masks the prompt positions using ``("tokens", "prompt")``,
+          ``"generic"`` masks padding positions using the attention mask, and ``"rlhf"`` requires
+          an assistant mask which is only computed in history mode, so ``"rlhf"`` with ``tokens``
+          input raises a ``ValueError``.
```
vmoens: [P2] Qualify the tokens-mode RLHF restriction

The tokens log-prob path preserves a caller-supplied `("masks", "all_assistant_mask")`, and the RLHF distribution uses it. I verified that supplying a boolean response mask makes `GRPOLoss(masking_strategy="rlhf")` return a finite loss with exactly that mask. Please say this raises only when no assistant mask is available, and cover both missing and supplied masks in the tests. Tokens input itself is supported.

test/llm/test_llm_objectives.py
```diff
@@ -1444,6 +1444,174 @@ def test_distillation_to_follows_module_device(self, data):
...
+        model, tokenizer = tiny_qwen_and_tokenizer
+        wrapper, td = _grpo_masking_tokens_td(model, tokenizer)
+
+        sft = GRPOLoss(actor_network=wrapper, masking_strategy="sft")(td)
+        generic = GRPOLoss(actor_network=wrapper, masking_strategy="generic")(td)
+        assert torch.isfinite(sft.loss_objective)
+        assert torch.isfinite(generic.loss_objective)
+        # sft excludes the prompt positions, generic includes them, so with the
+        # same data and advantage the two losses must differ
+        assert sft.loss_objective.item() != generic.loss_objective.item()
```
vmoens: [P2] Assert the expected token selection

Unequal scalar losses do not establish that either strategy selects the correct tokens. I swapped SFT and generic for tokens input and made generic use assistant masking for history input; all six new cases still passed. With the balanced per-trajectory advantages, the history equality check can also compare zero losses despite different token selections. Please assert the expected distribution masks: attention with prompt positions cleared for tokens-mode SFT, attention for generic, and the assistant mask for history-mode SFT/RLHF when prompt tokens are absent. Retain the forward checks and verify the unreduced `(B, T, 1)` loss shape to protect the shape contract.

torchrl/objectives/llm/grpo.py
```diff
@@ -388,6 +388,23 @@ class GRPOLoss(LossModule):
...
+        * ``history`` input: the assistant mask is computed from the chat template. ``"rlhf"`` uses
+          it directly, and ``"sft"`` falls back to it whenever ``("tokens", "prompt")`` is not
+          available, which makes ``"sft"`` and ``"rlhf"`` select the same tokens in this mode.
+        * ``"generic"`` always uses the attention mask regardless of the input mode.
+
+        In every supported combination the masks are boolean tensors of shape ``(batch, T)`` over
+        the full padded sequence and the distribution log probability has the same shape, so a per
+        trajectory advantage of shape ``(batch, 1, 1)`` broadcasts against them.
```
vmoens: [P3] Describe the log-weight dimension used for broadcasting

The advantage broadcasts against the `(B, T, 1)` log-weight produced by `_log_weight`, rather than directly against the `(B, T)` log probabilities or masks. Directly combining `(B, 1, 1)` with `(B, T)` would broadcast to `(B, B, T)`. Please state the intermediate singleton dimension and the supported `(B, 1, 1)` per-trajectory / `(B, T, 1)` per-token advantage shapes explicitly; `forward` requires advantage and log-weight to have the same rank.

vmoens (changes requested):
The new CPU cases exercise the loss with supported input shapes, but the documentation rules out a supported tokens-mode RLHF input, and the scalar comparisons do not establish which tokens each strategy selects. Please address the inline comments before merging.

Suggested next steps:

1. Correct the assistant-mask requirement and describe the intermediate shapes explicitly: masks and log probabilities are `(B, T)`, log-weights are `(B, T, 1)`, and advantages can be `(B, 1, 1)` per trajectory or `(B, T, 1)` per token for these batched inputs.
2. Strengthen the existing cases with expected distribution masks and unreduced loss shapes. Cover tokens-mode RLHF with both an absent and a caller-supplied assistant mask. These checks can fit into the existing cases without adding another layer of redundant tests.
3. In the history helper, remove `tokenizer_kwargs={"chat_template_name": "qwen"}` or use the top-level `chat_template_name="qwen"`; the history path constructs its own tokenizer kwargs, so the nested setting is ignored.

Further follow-ups: make the missing-assistant-mask error suggest supplying the mask or using history input, and investigate the original skipped vLLM integration with actual generated output. The fabricated-response tests do not establish that this integration is fixed, so validate that path before removing its skip and clarify the PR's replacement claim accordingly.

Validation on `e0f3d7b`: `pytest test/llm/test_llm_objectives.py -k 'TestGRPOLossMaskingContract or TestLosses'` passed all 13 selected tests on CPU with PyTorch 2.11.0 and Transformers 4.56.2. A separate probe confirmed caller-supplied assistant masks work in tokens mode. It also confirmed all six new cases still pass with SFT/generic swapped for tokens input and generic incorrectly using assistant masking for history input. GPU/vLLM integration was not run. I found no evidence here that warrants changing the loss math.

torchrl/objectives/llm/grpo.py
n-dlms: Done in 9dc005c. The docstring now states that rlhf with tokens input uses a caller supplied assistant mask and raises only when it is absent, and the tokens rlhf case covers both sides: without a mask it raises, with a supplied boolean mask the distribution mask equals exactly that mask and the loss is finite.

test/llm/test_llm_objectives.py
n-dlms: Done in 9dc005c. Each case now compares the distribution mask against the expected tensor: attention with prompt positions cleared for tokens sft, attention for generic, assistant mask for history sft and rlhf. The scalar loss equality case is removed in favor of those comparisons, and the parametrized history case also checks the unreduced (B, T, 1) shape with aggregation='none'. One note: the wrapper output masks arrive as int64 in the tokens path, so the comparisons cast both sides to bool; the history masks are already boolean.

torchrl/objectives/llm/grpo.py
n-dlms: Done in 9dc005c. The note now states the (B, T, 1) log weight that the advantage broadcasts against, the supported (B, 1, 1) per trajectory and (B, T, 1) per token advantage shapes, and the same rank requirement enforced by forward. Also fixed the history helper to pass chat_template_name to the wrapper constructor instead of the ignored nested tokenizer_kwargs. Both follow ups (the missing mask error message and validating the skipped vLLM integration with generated output) are acknowledged as separate work; the PR description no longer claims to replace that test.

PR #4351, opened 2026-09-12
[BugFix] Build IncrementalTokenizer tokens.full from NestedKey

### LINKED ISSUE #4344: [BUG] IncrementalTokenizer builds the wrong tokens.full NestedKey on step
## Describe the bug

On `_step`, `IncrementalTokenizer` reuses `tokens.full` as the next prompt. The lookup is:

```python
tokens_full_key = (
    (self.tokens_key[0], "full")
    if isinstance(self.tokens_key, tuple)
    else "tokens_full"
)
```

- `tokens_key="tokens"` looks for `"tokens_full"`, not `("tokens", "full")` (what the wrappers write).
- `tokens_key=("obs", "tokens", "prompt")` looks for `("obs", "full")`, not `("obs", "tokens", "full")`.

Existing `test_custom_keys` only covers reset, not this reuse path.

## Expected behavior

`tokens_full_key` should replace the last component of `tokens_key` with `"full"`, for both string and NestedKey inputs. `_step` should copy that tensor onto `tokens_key` without re-tokenizing when it is present.

## Checklist

YeonwooSung:
## Description

On `_step`, `IncrementalTokenizer` reused `tokens.full` with `(tokens_key[0], "full")` or `"tokens_full"`. A NestedKey such as `("obs", "tok", "prompt")` looked up `("obs", "full")`. A string `"tokens"` looked up `"tokens_full"` instead of `("tokens", "full")`.

The lookup now replaces the last NestedKey component with `"full"`.

### Code example

Minimal step-hook regression: reuse the full token sequence from the same nested namespace. This path does not invoke the tokenizer.

```python
from types import SimpleNamespace
import torch
from torchrl.envs.llm.transforms import IncrementalTokenizer

transform = IncrementalTokenizer(
    SimpleNamespace(vocab_size=32), tokens_key=("obs", "tok", "prompt"),
)
td = TensorDict({
    ("obs", "tok", "full"): torch.tensor([1, 2, 3]),
    ("obs", "full"): torch.tensor([9, 9, 9]),  # Old lookup selected this decoy.
}, batch_size=[])
next_td = transform._step(td, TensorDict(batch_size=[]))
torch.testing.assert_close(next_td["obs", "tok", "prompt"], torch.tensor([1, 2, 3]))
```

## Motivation and Context

close #4344

## Types of changes

## Checklist

### DIFF
--- a/test/llm/test_llm_transforms.py
+++ b/test/llm/test_llm_transforms.py
@@ -724,60 +724,47 @@ class TestIncrementalTokenizer:
         tokens = result.get(("tokens", "prompt"), as_list=True)
         assert tokens[0].numel() > 0
 
-    def test_step_nested_tokens_full_key(self):
-        """Nested tokens_key looks up (*prefix, 'full'), not (first, 'full')."""
+    @pytest.mark.parametrize(
+        "tokens_key, expected_full_key, decoy_key",
+        [
+            pytest.param(
+                ("obs", "tok", "prompt"),
+                ("obs", "tok", "full"),
+                ("obs", "full"),
+                id="nested",
+            ),
+            pytest.param(
+                "tokens",
+                ("tokens", "full"),
+                "tokens_full",
+                id="string",
+            ),
+            pytest.param(
+                ("tokens",),
+                ("tokens", "full"),
+                "full",
+                id="one_tuple",
+            ),
+        ],
+    )
+    def test_step_tokens_full_key(self, tokens_key, expected_full_key, decoy_key):
+        """Reuse tokens.full under the last NestedKey component; decoys fail the old lookups."""
         tokens_full = torch.tensor([1, 2, 3])
         decoy = torch.tensor([9, 9, 9])
         transform = IncrementalTokenizer(
             self.DummyTokenizer(),
-            tokens_key=("obs", "tok", "prompt"),
+            tokens_key=tokens_key,
         )
         td = TensorDict(
             {
-                ("obs", "tok", "full"): tokens_full,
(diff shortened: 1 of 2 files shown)


test/llm/test_llm_transforms.py
```diff
@@ -720,6 +724,61 @@ def test_empty_history_handling(self, tokenizer):
         tokens = result.get(("tokens", "prompt"), as_list=True)
         assert tokens[0].numel() > 0
 
+    def test_step_nested_tokens_full_key(self):
```
vmoens: [P3] Parametrize the three copies of the same reuse test

The nested, string, and one-tuple tests repeat the same setup and assertion across roughly 55 lines. They protect one contract with three key representations. Use one parameterized case table for tokens_key, expected full key, and decoy key; retain the decoys so incorrect lookup still fails. This keeps the meaningful coverage while following the repository test-design guidance.

vmoens (commented):
The NestedKey lookup correction is sound and all three new cases pass. One test-maintainability cleanup remains, so I am leaving this as a comment review rather than approving yet.

test/llm/test_llm_transforms.py
YeonwooSung: Folded the three cases into one parameterized table (`nested` / `string` / `one_tuple`). Each row still plants a decoy at the old lookup so a wrong key still fails. 8de055e08

vmoens (approved):
Re-reviewed 8de055e. The duplicated setup is now one parametrized regression covering flat, nested, and custom tokens.full keys, including a decoy key. All 3 cases pass. No remaining actionables.

Issue #4224, opened 2026-09-05
[BUG] SFTLoss and DistillationLoss cache self.device and break after .to()


YeonwooSung:
## Describe the bug

`SFTLoss` and `DistillationLoss` both assign `self.device = device` on a `LossModule` / `nn.Module`. `module.to(...)` moves parameters and buffers only, so the cached Python attribute goes stale. After `loss.to("cuda")` (or `.cpu()`), mask recovery and the actor forward still use the constructor device (often `None` or `"cpu"`).

This violates the TorchRL module-device contract (Agents.md §6a): do not cache a module device; derive placement from the involved tensor.

## To Reproduce

Both classes store the constructor argument as module state:

```259:298:torchrl/objectives/llm/sft.py
    def __init__(
        self,
        ...
        device: torch.device | None = None,
    ):
        super().__init__()
        ...
        self.device = device
```

```307:345:torchrl/objectives/llm/distillation.py
    def __init__(
        self,
        ...
        device: torch.device | None = None,
    ):
        super().__init__()
        ...
        self.device = device
```

Forward then wraps tokenizer / actor work in `torch.device(self.device)`:

```345:376:torchrl/objectives/llm/sft.py
        if assistant_masks is None:
            with torch.device(
                self.device
            ) if self.device is not None else contextlib.nullcontext():
                token_struct = history.apply_chat_template(
                    tokenizer=self.tokenizer, **self.tokenizer_kwargs
                )
        ...
        with torch.device(
            self.device
        ) if self.device is not None else contextlib.nullcontext():
            output_loss = self.actor_network(input_loss)
```

The same pattern is in `DistillationLoss._get_masks` (`distillation.py:365-370`) and `DistillationLoss.forward` (`distillation.py:439-442`).

Sketch:

```python
loss = SFTLoss(actor_network=actor, tokenizer=tokenizer, device="cpu")
loss.to("cuda")          # moves parameters/buffers; self.device stays "cpu"
out = loss(td.cuda())    # mask/token work still runs under torch.device("cpu")
```

The same applies to `DistillationLoss`.

## Expected behavior

After `loss.to("cuda")` or `loss.to("cpu")`, mask recovery and the actor forward should use the device of the input tensordict or of a parameter / buffer, not a cached constructor string. Dropping `self.device` and deriving the device from the involved tensor is the contract used elsewhere in TorchRL.

## Screenshots

N/A

## System info

Observed on a source checkout of pytorch/rl at `1d3de3db`.

[...]

PR #4352, opened 2026-09-12
[BugFix] Stop caching Tokenizer parent.device

### LINKED ISSUE #4345: [BUG] Tokenizer caches parent.device and goes stale after env.to
## Describe the bug

Both LLM `Tokenizer` and `torchrl.envs.transforms` `Tokenizer` cache `parent.device` on first access:

```python
@property
def device(self):
    if "_device" in self.__dict__:
        return self._device
    ...
    self._device = parent.device
    return device
```

`env.to(...)` / `transform.to(...)` moves parameters and the parent device, but not this Python attribute (Agents.md 6a / same class as #4224). Later tokenize calls still `.to()` onto the stale device.

## Expected behavior

Do not cache a single module device. Read the parent (or the input tensor) on each call. After `env.to("meta")`, new `input_ids` must be on `meta`.

## Checklist

YeonwooSung:
## Description

Both LLM `Tokenizer` and `torchrl.envs.transforms.Tokenizer` cached `parent.device` on first access. `env.to(...)` does not move that Python attribute, so later tokenize calls stayed on the stale device.

`out_device` now reads `parent.device` on each call and describes where token tensors and attention masks are written. After `env.to("meta")`, both outputs are on `meta`. When `out_device` is `None`, outputs keep the device chosen by the tokenizer.

Both Tokenizer implementations retain `device` as a deprecated alias, with removal scheduled for TorchRL v0.17. Reading the alias emits a `DeprecationWarning`; normal tokenization uses `out_device` and does not warn.

### Code example

Use `out_device` to inspect tokenization output placement. Tokenization follows an environment device move even after the first call. The small tokenizer stub avoids a model download; the same API is available on `torchrl.envs.transforms.Tokenizer`.

```python
import torch
from torchrl.envs import TransformedEnv
from torchrl.envs.llm.transforms import Tokenizer
from torchrl.testing.mocking_classes import CountingEnv

class FixedTokenizer:
    def __call__(self, value, **kwargs):
        return {
            "input_ids": torch.tensor([[1, 2]]),
            "attention_mask": torch.tensor([[1, 1]]),
        }

transform = Tokenizer(
    in_keys=["text"], out_keys=["input_ids"], tokenizer=FixedTokenizer(),
)
env = TransformedEnv(CountingEnv(), transform)
assert transform(TensorDict(text="hello"))["input_ids"].device.type == "cpu"
env.to("meta")
assert transform.out_device == torch.device("meta")
assert transform(TensorDict(text="hello"))["input_ids"].device.type == "meta"
# The second result used to stay on CPU because parent.device was cached.
env.close()
```

## Motivation and Context

close #4345

Same class of bug as #4224.

## Types of changes

## Checklist

### DIFF
--- a/docs/source/reference/envs_transforms.rst
+++ b/docs/source/reference/envs_transforms.rst
@@ -29,6 +29,11 @@ By default, the transformed environment will inherit the device of the
 It is now apparent that this can bring a significant speedup depending on the kind of
 operations that is to be computed.
 
+:attr:`Tokenizer.out_device` describes the destination of token tensors and
+attention masks. It follows the current parent environment device; ``None``
+leaves outputs on the tokenizer's chosen device. ``Tokenizer.device`` is a
+deprecated alias and will be removed in TorchRL v0.17.
+
 A great advantage of environment wrappers is that one can consult the environment up to that wrapper.
 The same can be achieved with TorchRL transformed environments: the ``parent`` attribute will
 return a new :class:`TransformedEnv` with all the transforms up to the transform of interest.
--- a/docs/source/reference/llms_transforms.rst
+++ b/docs/source/reference/llms_transforms.rst
@@ -7,6 +7,11 @@ LLM Transforms
 
 Transforms for LLM environments, including tools and utilities.
 
+:attr:`Tokenizer.out_device` describes the destination of token tensors and
+attention masks. It follows the current parent environment device; ``None``
+leaves outputs on the tokenizer's chosen device. ``Tokenizer.device`` is a
+deprecated alias and will be removed in TorchRL v0.17.
+
 .. autosummary::
     :toctree: generated/
     :template: rl_template.rst
--- a/test/llm/test_llm_transforms.py
+++ b/test/llm/test_llm_transforms.py
@@ -25,7 +25,7 @@ from torchrl.envs.llm.transforms import (
     ToolRegistry,
     XMLBlockParser,
 )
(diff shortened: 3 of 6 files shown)


vmoens (approved):
The property now follows the live parent device in both Tokenizer implementations. Both device-move regressions passed locally. No actionable findings. GitHub CI has not run yet; the workflows are awaiting approval.

vmoens (commented):
The live parent lookup fixes the stale output placement correctly. Naming recommendation: expose this as `out_device`, since it describes where tokenization results are written, rather than claiming all module state lives on one device.

Keep the existing `device` property as a deprecated alias until TorchRL v0.17, and use `out_device` internally so normal tokenization does not emit deprecation warnings. Apply the same naming to both Tokenizer implementations. The replacement property block would be:

```python
# Module-level import:
import warnings

# Inside Tokenizer:
@property
def out_device(self) -> torch.device | None:
    parent = self.parent
    return None if parent is None else parent.device

@property
def device(self) -> torch.device | None:
    warnings.warn(
        "Tokenizer.device is deprecated and will be removed in TorchRL v0.17. "
        "Use out_device instead.",
        DeprecationWarning,
        stacklevel=2,
    )
    return self.out_device
```

`None` should continue to mean no enforced destination. This is a naming/deprecation recommendation; the existing cache fix remains valid. It does not require adding a new constructor override or changing where outputs are placed.

PR #4353, opened 2026-09-12
[BugFix] Stop caching constructor device on ModuleTransform and KL transforms

### LINKED ISSUE #4346: [BUG] ModuleTransform, KLRewardTransform and RetrieveLogProb keep constructor self.device
## Describe the bug

These transforms are `nn.Module`s and store the constructor `device` as Python state:

- `ModuleTransform._call` does `tensordict.to(self.device)`
- `KLRewardTransform` / `RetrieveLogProb` use `torch.device(self.device)` the same way

`.to("meta")` moves parameters; `self.device` stays at the constructor value. A CPU-constructed `ModuleTransform` after `.to("meta")` casts the tensordict back to CPU and then runs a meta module.

Related to #4224 and the Tokenizer parent-device cache.

## Expected behavior

Do not keep a single-device attribute on the module. Place work from the input tensordict or a parameter/buffer. After `.to("meta")`, `_call` must not recast to the constructor device.

## Checklist

YeonwooSung:
## Description

`ModuleTransform`, `KLRewardTransform`, and `RetrieveLogProb` stored the constructor `device` and recast tensordicts to it at call time. `.to("meta")` moved parameters; `self.device` stayed CPU.

`device=` places the wrapped module at construction and remains the explicit I/O placement policy until v0.17. A later `.to(device)` updates that policy to the same destination so inputs are not recast to the stale constructor device. A dtype-only `.to()` leaves the policy unchanged. With no explicit device, inputs retain their placement; no device is inferred from the first parameter or buffer. `KLRewardTransform.ref_model` is a registered submodule so `.to` moves it.

The public `device` attributes are scheduled for removal in v0.17 and emit `DeprecationWarning` on get/set.

### Code example

```python
import torch
from torchrl.envs.transforms import ModuleTransform

module = TensorDictModule(
    torch.nn.Linear(3, 2), in_keys=["observation"], out_keys=["embedding"],
)
transform = ModuleTransform(module=module, device="cpu")
transform.to("meta")
td = TensorDict(
    {"observation": torch.ones(2, 3, device="meta")}, batch_size=[2], device="meta",
)
out = transform(td)
assert out["embedding"].device.type == "meta"
```

## Motivation and Context

close #4346

## Types of changes

## Checklist


torchrl/envs/llm/transforms/kl.py
```diff
@@ -487,8 +488,6 @@ def _step(
             next_tensordict.set(self.out_keys[0], reward)
         next_tensordict.set(self.out_keys[1], kl)
         next_tensordict.set(self.out_keys[2], ref_log_prob)
-        if original_device is not None:
-            next_tensordict = next_tensordict.to(original_device)
         return next_tensordict
```
vmoens, line 518: [P2] Restore KL results to the environment/input device

The removed `original_device` conversion was required when the reference model intentionally lived on a different device from the env. For CPU input and a reference model on MPS, `_step` now returns an entirely MPS TensorDict, while the transform still advertises CPU specs. The same problem applies to CUDA model/CPU env configurations. Resolve the model device dynamically for computation, then restore the output to the original runtime device; fixing the cached constructor device does not require changing this contract.

torchrl/envs/transforms/module.py
```diff
@@ -256,24 +263,20 @@ def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
...
+        if device is not None:
+            tensordict = tensordict.to(device)
         with torch.no_grad() if self.no_grad else nullcontext():
-            with (
-                tensordict.to(self.device)
-                if self.device is not None
-                else nullcontext(tensordict)
-            ) as td:
-                return self.module(td)
+            return self.module(tensordict)
```
vmoens: [P2] Preserve ModuleTransform copy-back behavior across devices

The previous TensorDict `.to(...)` context manager copied results back into the original TensorDict on exit. Replacing it with a plain conversion drops that behavior. With CPU input and an MPS module, the original input now remains without the module output key; running the same probe on the base implementation writes that key back onto CPU. This affects callers relying on in-place updates. Keep the dynamic module-device lookup while preserving the conversion context/copy-back behavior, in both `_call` and `_inv_call`; cover distinct input and model devices with real tensors.

vmoens (changes requested):
The live-device lookup fixes the stale cache, but dropping the input-device restoration changes the transforms' existing behavior for models intentionally placed on another device. Four new tests pass; CPU-input/MPS-model probes reproduce the output-device change.

torchrl/envs/llm/transforms/kl.py
YeonwooSung: `_step` still resolves the compute device from `_module_device(self.ref_model)` (falling back to `coef`) rather than a cached constructor device. After the model runs, `next_tensordict` is moved back to the original tensordict/env device, matching the previous contract for CPU env + CUDA/MPS reference model. Covered by `test_kl_reward_transform_restores_output_to_input_device` (5938fb314).

torchrl/envs/transforms/module.py
YeonwooSung: `_call` and `_inv_call` still look up the module device per call via `_module_device`, but they now use TensorDict `.to(device)` as a context manager again so new keys are written back onto the original tensordict. `test_copies_module_output_back_to_input` checks CPU input against a CUDA/MPS module for both paths (5938fb314).

vmoens (changes requested):
The latest commit restores the previously flagged copy-back/return-device paths. Two device-contract issues remain:

1. **[P2] Do not infer a module-wide placement from its first parameter.** A module can span several devices, and registration order does not specify where each input belongs. `_module_device` followed by a whole-TensorDict `.to(device)` still imposes a single-device assumption. Preserve correctly placed inputs and let the wrapped module own its routing; allocate intermediate tensors from the relevant input/output tensor. Any explicit input/output placement policy must be named and documented as such.

   The split CPU/MPS example below succeeds both directly and through the base implementation of `ModuleTransform(device=None)`. At this PR head, the transform moves both inputs to CPU and fails in the MPS branch. Tested with real CPU/MPS tensors. Run on a host with CUDA or MPS:

```python
import torch
from torch import nn
from torchrl.envs.transforms import ModuleTransform


class SplitModule(nn.Module):
    in_keys = ["cpu_input", "accelerator_input"]
    out_keys = ["result"]

    def __init__(self, accelerator):
        super().__init__()
        self.cpu_branch = nn.Linear(2, 2, device="cpu")
        self.accelerator_branch = nn.Linear(2, 2, device=accelerator)

    def forward(self, td):
        td["result"] = self.cpu_branch(td["cpu_input"]) + self.accelerator_branch(
            td["accelerator_input"]
        ).to("cpu")
        return td


accelerator = "cuda" if torch.cuda.is_available() else "mps"
module = SplitModule(accelerator)
td = TensorDict({
    "cpu_input": torch.ones(1, 2, device="cpu"),
    "accelerator_input": torch.ones(1, 2, device=accelerator),
}, batch_size=[1])
assert td.device is None
assert torch.isfinite(module(td.clone())["result"]).all()

ModuleTransform(module=module)(td.clone())  # Raises: input is on CPU but expected on MPS.
```

2. **[P2] Soft-deprecate the existing public `device` attributes, with removal in TorchRL v0.17.** `ModuleTransform`, `KLRewardTransform`, and `RetrieveLogProb` currently expose mutable attributes, but this diff removes them immediately. Retain legacy access during the transition, emit `DeprecationWarning` explicitly naming v0.17, and migrate internal callers off the deprecated attribute. Preserve existing supported read/write behavior during that period; the compatibility value must not be presented as the location of every parameter/buffer or synthesized from the first parameter. A constructor `device=` argument is a separate API: retaining it does not require a permanent module-wide `device` attribute, and changing its existing semantics also needs compatibility consideration.

Please cover the actual mixed-device regression and the legacy warning/removal contract with compact tests.

YeonwooSung:
3f9724754 addresses both remaining P2s.

**P2: do not infer a module-wide placement from the first parameter.** `_module_device` and the default whole-TensorDict `.to(device)` path are gone. Constructor `device=` still places the wrapped module at init and is then dropped. With no explicit I/O policy the tensordict is left in place and the module owns routing. Intermediate tensors follow the relevant input/output tensor (`ref_log_prob.to(curr_log_prob.device)`). The split CPU/MPS `SplitModule` repro now succeeds directly and through `ModuleTransform(device=None)` (`test_does_not_move_split_module_inputs`).

**P2: soft-deprecate public `device`, removal in v0.17.** `ModuleTransform.device`, `KLRewardTransform.device`, and `RetrieveLogProb.device` are restored as a named, documented, deprecated explicit input/output placement policy. Get/set emit `DeprecationWarning` naming v0.17. The stored value is whatever the caller set; it is not inferred from the first parameter and is not presented as the location of every parameter/buffer. Internal callers use `_in_out_device`. Setting the attribute applies the legacy `.to()` / copy-back policy (`test_copies_module_output_back_to_input`, `test_device_attr_is_deprecated_io_policy`).

torchrl/envs/transforms/module.py
```diff
@@ -191,9 +204,11 @@ def __init__(
                 "Only one of `module` or `module_factory` must be provided."
             )
         self.module = module if module is not None else module_factory()
+        if device is not None:
```
vmoens, line 213: [P2] Preserve explicit constructor-device behavior during the deprecation

The old implementation stored the explicitly supplied device and used it to route inputs. The new descriptor starts at None even when device= was supplied. This changes both the legacy read and the actual call behavior before v0.17; merely warning on a later attribute read does not warn callers about the constructor behavior change.

This succeeds on the PR base (46c5daee9) and fails on this head, tested with real MPS tensors:

```python
import torch
from torch import nn
from torchrl.envs.transforms import ModuleTransform

accelerator = "cuda" if torch.cuda.is_available() else "mps"
module = TensorDictModule(
    nn.Linear(2, 2, device=accelerator), in_keys=["x"], out_keys=["y"]
)
transform = ModuleTransform(module=module, device=accelerator)
td = TensorDict({"x": torch.ones(1, 2)}, batch_size=[1], device="cpu")
transform(td)  # Now: input is on cpu but expected on mps.
```

The module is already on the requested device in both versions, so this isolates the lost explicit I/O policy. Previously transform.device also returned the constructor value; now it returns None. Preserve this supported behavior through the deprecation window, without reintroducing first-parameter inference, and audit the equivalent constructor transitions in RetrieveLogProb/KLRewardTransform. Requiring existing callers to add transform.device = accelerator is an immediate migration, not a soft deprecation.

vmoens (changes requested):
Re-reviewed 3f97247. The actual split CPU/MPS regression now passes, and the public attributes emit v0.17 deprecation warnings. Eleven focused tests pass, with four CUDA cases skipped locally.

The compatibility transition is still incomplete: the old constructor device value and its explicit input-placement behavior disappear immediately. A previously working, already-placed Linear module now gets CPU input and raises. Details inline. Restoring only the ability to assign the deprecated attribute does not preserve existing constructor calls.

Please keep the regression focused on the actual public call. The diff is now about 302 added test lines versus 127 implementation/doc lines (2.4:1); the self-moving mock and getter-only assertions miss this compatibility regression. Also inline _in_out_device's single getattr expression, and correct the getter warning's stacklevel: a direct top-level read currently reports sys:1 instead of the caller.

torchrl/envs/transforms/module.py
YeonwooSung: Constructor `device=` is stored on `_io_device` (not via the descriptor, so construction does not warn) and is the explicit I/O tensordict placement policy until v0.17. Incoming tensordicts are moved with TensorDict `.to()` copy-back; the wrapped module is still placed at init. `transform.device` returns that constructor value. No first-parameter inference and no `self.device =` assignment.

`_in_out_device` is inlined as `getattr(obj, "_io_device", None)`. Getter/setter stacklevel is 3, so a top-level `.device` read points at the caller file, not `sys:1`.

Public-call regression matches the CPU-input + already-placed Linear + constructor `device=accelerator` snippet (MPS if no CUDA). Same constructor transition on `RetrieveLogProb` / `KLRewardTransform`. Removed the self-moving mock, meta-recast, and getter-only tests.

59f2644e5

YeonwooSung:
59f2644e5 keeps constructor `device=` as the explicit I/O tensordict placement policy until v0.17. `transform.device` returns that value; incoming tensordicts are moved with TensorDict `.to()` copy-back. Public-call regression matches the CPU-input + already-placed Linear snippet.

torchrl/envs/transforms/module.py
```diff
@@ -191,9 +207,12 @@ def __init__(
...
+            # Place the module and keep device as the explicit I/O policy until v0.17.
+            self.module = self.module.to(device)
+            self._io_device = torch.device(device)
```
vmoens, line 216: [P2] Handle the retained I/O policy when the transform moves

`_io_device` stays at the constructor value after `.to(...)` moves the registered module. The next call recasts correctly placed input back to the old device. This reintroduces the original bug, and the previous move regressions were removed in this revision:

```python
module = TensorDictModule(
    torch.nn.Linear(2, 2), in_keys=["x"], out_keys=["y"]
)
transform = ModuleTransform(module=module, device="cpu").to("mps")
td = TensorDict({"x": torch.ones(1, 2, device="mps")}, [1], device="mps")
transform(td)
# RuntimeError: input is on cpu but expected on mps
```

The same CPU-to-MPS move fails for `RetrieveLogProb` and `KLRewardTransform`; the existing meta example raises `Cannot copy out of meta tensor`. Handle explicit device moves without inferring one device from a module's first parameter, and retain a real call regression for both construction and subsequent movement. Keep the legacy get/set behavior and v0.17 warning window.

vmoens (changes requested):
Rechecked 59f2644e5. Constructor I/O compatibility and the deprecated getter now work, including the real CPU-input / MPS-Linear call. The split-device case also passes. However, retaining `_io_device` without handling subsequent module moves restores the original stale-constructor-device bug; the PR body's meta reproduction fails again, and numerical MPS calls fail for all three transforms. Please preserve both constructor compatibility through v0.17 and correct behavior after a device move.

Validation: 5 focused tests passed, 4 CUDA cases skipped; additional public-call probes reproduced the move failure. Please also run ufmt on `test/transforms/test_module_transforms.py`, `test/llm/test_llm_transforms.py`, and `torchrl/envs/llm/transforms/kl.py` (three formatting differences remain).

torchrl/envs/transforms/module.py
YeonwooSung: `.to(device)` now updates the retained constructor I/O policy. `_update_io_device_from_to` parses the destination with `torch._C._nn._parse_to`; if `device is not None` and `_io_device` is already set, it is written with `object.__setattr__` so the deprecated descriptor does not warn. A dtype-only `.to()` leaves the policy unchanged. Same `to()` handling on `ModuleTransform`, `RetrieveLogProb`, and `KLRewardTransform`. No first-parameter inference and no `_apply` device guess.

Public-call regressions cover constructor `device=` (CPU input + already-placed Linear) and subsequent `.to(accelerator)` / `.to("meta")`. ufmt on the three flagged files. Rebased onto `06d57a05` (#4256).

b77c5ba32

YeonwooSung:
b77c5ba32 updates the retained constructor I/O policy when `.to(device)` moves the transform (`torch._C._nn._parse_to`, `object.__setattr__` on `_io_device`). Same handling on `ModuleTransform`, `RetrieveLogProb`, and `KLRewardTransform`. Public-call regressions cover constructor `device=` and later `.to(accelerator)` / `.to("meta")`. ufmt on the three flagged files. Rebased onto `06d57a05` (#4256); no merge.

vmoens (approved):
Rechecked b77c5ba32, including the changes after the rebase. The remaining device-move regression is fixed: constructor I/O compatibility is preserved through the v0.17 deprecation window, subsequent explicit `.to(device)` calls update the retained policy, and the mixed-device case does not infer a device from the first parameter. The previous meta and numerical MPS reproductions now pass for ModuleTransform, RetrieveLogProb, and KLRewardTransform.

Validation: 13 focused tests passed, 7 CUDA cases skipped; all changed Python files pass ufmt. The PR body's meta example now succeeds. No remaining actionable findings in the reviewed change.

PR #4355, opened 2026-09-12
[BugFix] Accept History on ChatEnv.reset

### LINKED ISSUE #4348: [BUG] ChatEnv._reset always wraps data_key as a raw user utterance
## Describe the bug

`ChatEnv._reset` always does `History(role=user_role, content=content)` with a FIXME: \"Assume the text is not formatted and this is just content\".

If `query` / `data_key` is already a `History` (or a list of chat messages), it is wrapped again as a single user utterance and roles/content are corrupted.

## Expected behavior

- A string / unformatted content still becomes a user `History` (current behavior).
- An existing `History` (or message list) is used as-is, optionally prepended with `system_prompt` if that is set.

## Checklist

YeonwooSung:
## Description

`ChatEnv._reset` always wrapped `data_key` as a raw user utterance (`History(role=user, content=content)`). An already-built `History` or message list was wrapped again and roles were corrupted.

A `History` or list of `{role, content}` messages is now used as-is. A string is still wrapped as a user turn. `system_prompt` is still prepended when set.

### Code example

Pass a conversation with an explicit environment batch dimension. Reset preserves the existing user and assistant turns.

```python
from torchrl.data.llm import History
from torchrl.envs.llm import ChatEnv

query = History.from_chats([[
    {"role": "user", "content": "Hello"},
    {"role": "assistant", "content": "Hi"},
]])
env = ChatEnv(batch_size=(1,), input_mode="history", device="cpu")
td = env.reset(TensorDict(query=query, batch_size=[1], device="cpu"))
prompt = td["history"][0].prompt
assert list(prompt.role) == ["user", "assistant"]
assert list(prompt.content) == ["Hello", "Hi"]
# Previously, the whole History was wrapped as a new user utterance.
env.close()
```

## Motivation and Context

close #4348

## Types of changes

## Checklist

### DIFF
--- a/torchrl/envs/llm/chat.py
+++ b/torchrl/envs/llm/chat.py
@@ -9,7 +9,7 @@ from collections.abc import Callable
 from typing import Any, Literal, TYPE_CHECKING
 
 import torch
-from tensordict import lazy_stack, TensorDictBase
+from tensordict import lazy_stack, NonTensorData, NonTensorStack, TensorDictBase
 from tensordict.utils import _zip_strict
 from torch.utils.data import DataLoader
 from torchrl.data import Composite, NonTensor
@@ -75,15 +75,23 @@ def _is_chat_payload(obj: Any) -> bool:
 def _reset_query_payload(content: Any) -> Any:
     if isinstance(content, History):
         return content
-    # NonTensorStack.data is the first element; prefer tolist() for stacks.
-    data = getattr(content, "data", None)
-    if isinstance(data, History):
-        return data
+    # NonTensorData.data is the payload. NonTensorStack.data is only the first
+    # env -- unwrap stacks with tolist() / unbind so every History is kept.
+    if isinstance(content, NonTensorData):
+        data = content.data
+        return data if isinstance(data, History) else _reset_query_payload(data)
+    if isinstance(content, NonTensorStack):
+        try:
+            items = content.tolist()
+        except Exception:
+            items = content.unbind(0)
+        return [_reset_query_payload(item) for item in items]
     if hasattr(content, "tolist") and not isinstance(content, (str, bytes)):
         try:
             return content.tolist()
         except Exception:
             pass
+    data = getattr(content, "data", None)
     if data is not None:
         return data
     return content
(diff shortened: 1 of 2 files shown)


torchrl/envs/llm/chat.py
```diff
@@ -450,36 +506,69 @@ def _step_tokens(self, tensordict):
         empty_td.set("tokens", new_history)
         return empty_td
 
+    def _align_history_batch(self, history: History) -> History:
+        env_bs = torch.Size(self.batch_size)
+        if history.batch_size[: len(env_bs)] == env_bs:
```
vmoens: [P2] Distinguish the conversation axis from the environment batch

The prefix check accepts a time-only History of shape [2] as an already batched history when env.batch_size is [2]. A two-message user/assistant History wrapped in NonTensorData and supplied to a two-env reset is consequently split into one user message for env 0 and one assistant message for env 1. The same input with three envs correctly broadcasts the entire conversation. Check the presence of the conversation dimension before accepting a batch-prefix match, so broadcasting does not depend on conversation length, and add the matching-length case.

vmoens (changes requested):
All 11 new reset-format cases pass. A separate broadcast case still corrupts the conversation when its length equals the env batch size. Please also inline the single-expression _is_chat_message helper rather than introducing that extra indirection.

torchrl/envs/llm/chat.py
YeonwooSung: `_align_history_batch` now requires a trailing conversation axis (`history.ndim > len(env.batch_size)`) before accepting a batch-prefix match. A time-only History of shape `[T]` is broadcast to every env even when `T` equals the env batch size, so a two-message user/assistant query is copied in full rather than split across envs. Already-batched History of shape `[*env_batch, time]` is unchanged. Regression coverage is `test_chat_env_reset_broadcasts_time_only_history` for `n_envs` in `{2, 3}`.

`_is_chat_message` is inlined at the `_is_chat_payload` / `_query_to_history` call sites.

vmoens (approved):
Re-reviewed afd8810. Time-only History is now broadcast even when its length happens to equal the environment batch size; explicitly batched History remains distinct. The reset tests check the messages themselves, including the two- and three-environment cases, and the unnecessary message-type helper is gone.

Validation: all 13 focused ChatEnv reset tests passed. The inspected SGLang CI job fails during dependency installation, before tests, because its TensorDict version does not satisfy the required minimum. No remaining code review actionables.

PR #4356, opened 2026-09-12
[Feature] Promote DoneTransform for group-level done expansion

### LINKED ISSUE #2536: [Feature Request] Include group-level done fields in the VMAS environment
## Motivation

In the current implementation of the VMAS environment, the done fields are only available on the root of the tensordict. However, for training, it is useful to have them in the group-level.

This is also an issue for the TorchRL tutorial scripts `multiagent_ppo.py` and `multiagent_competitive_ddpg.py`, which have to manually add the done and terminated at group-level with the right shape

https://github.com/pytorch/rl/blob/main/tutorials/sphinx-tutorials/multiagent_ppo.py#L655
https://github.com/pytorch/rl/blob/main/tutorials/sphinx-tutorials/multiagent_competitive_ddpg.py#L732

The PettingZoo env implementation does include the group-level done fields.

So, can we also include these fields in the VMAS code? If so, I'd be willing to work on this feature.

## Checklist

YeonwooSung:
## Description

VMAS still writes only root done. This PR does not change that. It promotes the sota `DoneTransform` into `torchrl.envs.transforms` with NestedKey in/out keys, uses it from the multi-agent PPO tutorial instead of the hand expand, and re-exports it from the sota helper.

Group dones are written as observations so env termination stays global.

### Code example

Expand shared termination flags to the agent reward shape before computing per-agent value targets.

```python
import torch
from torchrl.envs.transforms import DoneTransform

td = TensorDict({
    "done": torch.tensor([[False], [True]]),
    "terminated": torch.tensor([[False], [True]]),
    "agents": {"reward": torch.zeros(2, 3, 1)},
}, batch_size=[2])

out = DoneTransform(reward_key=("agents", "reward"))(td)
assert out["agents", "done"].shape == (2, 3, 1)
assert not out["agents", "done"][0].any()
assert out["agents", "done"][1].all()
# Each of the three agents receives its environment's shared done flag.
```

## Motivation and Context

close #2536

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist


torchrl/envs/transforms/_clip.py
```diff
@@ -298,3 +300,211 @@ def transform_output_spec(self, output_spec: Composite) -> Composite:
...
+        if reward is not None:
+            return reward
+        parent = self.parent
+        if parent is None:
+            return None
+        try:
+            _, ref_spec = ExpandAs._find_key_spec(parent.output_spec, self.reward_key)
+        except KeyError:
+            return None
+        return value.new_empty(ref_spec.shape)
```
vmoens: [P2] Include runtime batch dimensions in the reset reference

At reset the reward is absent, so the fallback uses the static parent spec shape. That drops extra runtime batch dimensions supported by batch-unlocked envs. For example, `TransformedEnv(TicTacToeEnv(), DoneTransform(reward_key=("player0", "reward"))).reset(TensorDict(batch_size=[2]))` fails in `expand_as_right`: the done is [2, 1] but this reference is [1]. Construct the reference shape using the incoming batch plus the reward event/group dimensions, and cover an unlocked batched reset.

vmoens (changes requested):
The six new tests pass, including GAE and the observation-spec behavior, but reset fails for batch-unlocked environments. Please also omit the new explicit __init__ return annotation and add DoneTransformConfig to the config reference documentation.

torchrl/envs/transforms/_clip.py
YeonwooSung: Reset now builds the reference as incoming batch + reward spec event/group dims (`parent.batch_size` stripped from the static spec shape). `TransformedEnv(TicTacToeEnv(), DoneTransform(reward_key=("player0", "reward"))).reset(TensorDict(batch_size=[2]))` writes `("player0", "done")` / `("player0", "terminated")` with shape `[2, 1]`.

Also dropped the explicit `DoneTransform.__init__` `-> None` annotation and added `DoneTransformConfig` to `docs/source/reference/config.rst`.

vmoens (commented):
Re-reviewed 60681c7. The unlocked-batch reset regression is fixed, the Config cross-reference is present, and the newly added __init__ return annotation is removed. All 7 DoneTransform tests pass.

Only the PR's own lint cleanup remains: end-of-file-fixer reports an extra blank line after the final return in torchrl/envs/transforms/_clip.py. Remove it, or run:

```bash
pre-commit run end-of-file-fixer --files torchrl/envs/transforms/_clip.py
```

The inspected documentation failure concerns other class imports, not DoneTransform. I am leaving approval pending the small formatting fix.

torchrl/trainers/algorithms/configs/transforms.py
```diff
@@ -1003,6 +1003,22 @@ def __post_init__(self) -> None:
         super().__post_init__()
 
 
+@dataclass
+class DoneTransformConfig(TransformConfig):
+    """Hydra configuration for :class:`~torchrl.envs.transforms.DoneTransform`."""
+
+    in_keys: list[str] | None = None
+    out_keys: list[str] | None = None
+    reward_key: list[str] | str | None = None
```
vmoens: [P1] Use an OmegaConf-compatible reward_key schema

OmegaConf does not support this union of a container and a scalar. DoneTransformConfig is registered eagerly by _register_configs(), so the new field breaks importing the entire trainer Config package, including users of unrelated trainers. This is the source of the ConfigValueError visible in the Dreamer CI failures.

```python
import torchrl.trainers.algorithms.configs
# omegaconf.errors.ConfigValueError: Unions of containers are not supported:
# reward_key: Union[list[str], str]
```

Use the repository's Hydra-compatible representation for flat/nested keys, normalize it in the factory, and add one actual Config registration/instantiation regression. Testing only DoneTransform itself misses this import-time failure.

vmoens (changes requested):
Correction to my previous follow-up: this PR has a functional Config registration failure in addition to the EOF formatting issue. Tracing the CPU/GPU CI failures reaches the newly added DoneTransformConfig, and a direct local import reproduces it. The seven transform-only tests did not exercise the Config registry.

Please fix the OmegaConf-compatible field schema and cover registration/instantiation through the public Hydra path. The trailing blank line in _clip.py also still needs the formatter fix noted previously.

torchrl/trainers/algorithms/configs/transforms.py
YeonwooSung: `reward_key` is now `Any = None`, matching the other Hydra NestedKey fields. `import torchrl.trainers.algorithms.configs` no longer hits OmegaConf's union-of-containers error; `_make_done_transform` still runs `_normalize_hydra_key`. `TestTransformConfigs.test_done_transform_config` instantiates both `"reward"` and `["player0", "reward"]` through `hydra.utils.instantiate` and checks the resulting `DoneTransform.reward_key`. Extra trailing blank line in `_clip.py` is also gone.

YeonwooSung:
Addressed in 467a499fa84ae0503947e741726c61ced658241b: OmegaConf-compatible `DoneTransformConfig.reward_key: Any = None` plus a Hydra instantiate regression in `TestTransformConfigs`; extra trailing blank line removed from `_clip.py`.

vmoens (approved):
Rechecked 467a499fa. The OmegaConf registration failure is fixed: the trainer config package imports successfully, and Hydra instantiates DoneTransform with both flat and nested reward keys. The seven transform cases plus the new config regression pass (8 total); ufmt accepts the changed Python files. No remaining actionable findings.

vmoens:
This needs rebasing

YeonwooSung:
Rebased onto 06d57a05 (#4256). No functional change. HEAD is b513b6ab8667d2d02720a4496dd4dce41b31fc0f.

vmoens (approved):
Rechecked the rebased head b513b6ab8. Range-diff confirms the approved behavior is preserved. The seven DoneTransform cases plus flat/nested Hydra instantiation regression pass again (8 total), and ufmt accepts the changed Python files. Approval reaffirmed on the current head; no remaining actionable findings.

PR #4358, opened 2026-09-12
[BugFix] Import ale_py for classic Atari environment ids

### LINKED ISSUE #2996: [BUG] Getting Environment `PongNoFrameskip` doesn't exist. error while running a2c_atari implementation.
## Describe the bug

Getting ```Environment `PongNoFrameskip` doesn't exist.``` error while running the a2c_atari implementation

## To Reproduce

1. Install the dependencies (for atari)
2. Run ```sh sota-check/run_a2c_atari.sh```

## Expected behavior

Gymnasium should recognise the environment. Training should start

## Screenshots

Image

## System info

 * Installed from local using ```python setup.py develop```
 * Python version: 3.10
 * Versions of any other relevant libraries: Gymnasium: 1.1.1, Gymnasium: 1.1.0  - All giving same error

## Reason and Possible fixes
Adding    ```import ale_py``` in the script a2c_atari.py fixes the error. 
References: https://ale.farama.org/gymnasium-interface/ , https://stackoverflow.com/questions/76150278/getting-an-error-trying-to-import-environment-openai-gym

It seems with newer versions of gymnasium, we need to import ale_py separately for atari games.

## Checklist

YeonwooSung:
## Description

Gymnasium 1.x does not register ALE ids until `ale_py` is imported. `GymEnv` already did that for `ALE/` names. Classic ids such as `BreakoutNoFrameskip-v4` still failed.

GymEnv now imports `ale_py` for `ALE/`, `ale_py:`, and `*NoFrameskip*` names. A2C/PPO Atari helpers default to `ALE/Pong-v5` to match their Hydra configs.

### Code example

Run with Gymnasium 1.x, ale-py, and the Atari ROMs installed. GymEnv now imports the registration dependency for the classic ID too.

```python
from torchrl.envs import GymEnv, set_gym_backend

# No explicit `import ale_py` is needed in user code.
with set_gym_backend("gymnasium"):
    env = GymEnv("BreakoutNoFrameskip-v4")
    td = env.reset()  # Previously: environment ID was not registered.
    transition = env.step(env.rand_action(td))
    assert ("next", "reward") in transition.keys(True, True)
    env.close()
```

## Motivation and Context

close #2996

## Types of changes

## Checklist

### DIFF
--- a/sota-implementations/ppo/utils_atari.py
+++ b/sota-implementations/ppo/utils_atari.py
@@ -45,7 +45,7 @@ from torchrl.record import VideoRecorder
 
 
 def make_base_env(
-    env_name="ALE/Pong-v5",
+    env_name="BreakoutNoFrameskip-v4",
     frame_skip=4,
     gym_backend="gymnasium",
     is_test=False,
--- a/sota-implementations/ppo/utils_atari.py
+++ b/sota-implementations/ppo/utils_atari.py
@@ -45,7 +45,7 @@ from torchrl.record import VideoRecorder
 
 
 def make_base_env(
-    env_name="BreakoutNoFrameskip-v4",
+    env_name="ALE/Pong-v5",
     frame_skip=4,
     gym_backend="gymnasium",
     is_test=False,
--- a/test/libs/test_gym.py
+++ b/test/libs/test_gym.py
@@ -35,6 +35,7 @@ from torchrl.envs.libs.gym import (
     _gym_to_torchrl_spec_transform,
     _has_gym,
     _is_from_pixels,
+    _looks_like_ale_env,
     _torchrl_to_gym_spec_transform,
     gym_backend,
     GymEnv,
@@ -1955,6 +1956,42 @@ class TestGym:
         check_env_specs(env)
         td = env.rand_step(env.reset())
         assert td["next", "reward"].shape == expected_reward_shape
+
+    @pytest.mark.parametrize(
+        "env_name,expected",
+        [
+            ("ALE/Pong-v5", True),
+            ("ale_py:ALE/Pong-v5", True),
+            ("PongNoFrameskip-v4", True),
+            ("BreakoutNoFrameskip-v4", True),
+            ("CartPole-v1", False),
+            ("HalfCheetah-v4", False),
+        ],
+    )
+    def test_looks_like_ale_env(self, env_name, expected):
+        assert _looks_like_ale_env(env_name) is expected
+
+    @pytest.mark.skipif(
+        not _has_ale,
+        reason="ALE not available (missing ale_py); skipping Atari gym test.",
+    )
+    @pytest.mark.parametrize("env_name", ["ALE/Pong-v5", "PongNoFrameskip-v4"])
+    def test_gymenv_ale_constructor(self, env_name):
+        if not _has_atari_for_gym():
+            pytest.skip(
+                "Atari not available for current gym version; skipping Atari gym test."
+            )
+        try:
+            env = GymEnv(env_name)
+        except Exception as err:
+            # NameNotFound is the regression: GymEnv must import ale_py first.
+            if isinstance(err, AttributeError) and "ale_py" in str(err):
+                pytest.skip(f"ALE/gym version incompatibility: {err}")
+            raise
+        try:
+            assert env.env_name == env_name
+        finally:
+            env.close()
 
 
 @pytest.mark.skipif(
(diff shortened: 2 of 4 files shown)


test/libs/test_gym.py
```diff
@@ -1956,6 +1957,46 @@ def _test_gymnasium_num_envs(self, num_envs, request):
...
+        not _has_ale,
+        reason="ALE not available (missing ale_py); skipping Atari gym test.",
+    )
+    @pytest.mark.parametrize("env_name", ["ALE/Pong-v5", "PongNoFrameskip-v4"])
+    def test_gymenv_ale_constructor(self, env_name):
+        # Constructor ids are gymnasium-only; gym 0.13/0.19 jobs ship ale_py
+        # without registering ALE/Pong-v5.
+        return
+
+    @implement_for("gymnasium")
```
vmoens, line 1997: [P2] Preserve the optional-dependency skip through version dispatch

The new `skipif(not _has_ale)` is applied inside `implement_for`; in this test setup the version-dispatched method loses the effective skip marker. Running `pytest test/libs/test_gym.py -k gymenv_ale_constructor` with Gymnasium installed and no ale_py executes both cases and fails with NamespaceNotFound/NameNotFound instead of skipping. Put the pytest-facing marks on an ordinary test wrapper and dispatch a helper, or perform the optional-dependency skip inside the selected implementation.

sota-implementations/ppo/utils_atari.py
```diff
@@ -45,7 +45,7 @@
 
 
 def make_base_env(
-    env_name="BreakoutNoFrameskip-v4",
+    env_name="ALE/Pong-v5",
```
vmoens: [P2] Keep the existing default task in this registration fix

Changing the utility default from BreakoutNoFrameskip-v4 to ALE/Pong-v5 changes the game (and ALE defaults) for callers that omit env_name. Importing ale_py already repairs the intended classic-ID path, and the Hydra callers supply their configured name. Please revert this and the matching A2C default change; changing the default experiment should be separate work.

torchrl/envs/libs/gym.py
```diff
@@ -106,6 +106,30 @@ def _patch_legacy_ale_py_gym_env(env_name: str) -> None:  # noqa: F811
         return
 
 
+def _looks_like_ale_env(env_name: str) -> bool:
```
vmoens: [P3] Inline the ID predicate and test the import behavior

`_looks_like_ale_env` only wraps one self-explanatory boolean expression, and its six tests restate that expression. Keep the condition directly in the import path and exercise that path (including a classic ID before ale_py is imported). This follows the repository guidance against one-expression wrappers and trivial implementation-mirroring tests.

vmoens (changes requested):
The import condition covers the intended IDs, but the new constructor tests fail when the optional dependency is absent. The A2C/PPO default-game changes also exceed the registration fix. Validation: six predicate cases passed; both constructor cases failed on an installation without ale_py.

torchrl/envs/libs/gym.py
YeonwooSung: Removed `_looks_like_ale_env`. The `ALE/` / `ale_py:` / `NoFrameskip` check is inline in `_import_ale_py_if_needed`. The six predicate cases are replaced by `test_import_ale_py_if_needed`, which hooks `builtins.__import__` so a classic `PongNoFrameskip-v4` id is observed to execute `import ale_py` even if the module is already loaded, and `CartPole-v1` is not.

test/libs/test_gym.py
YeonwooSung: skipif and parametrize now live on `test_gymenv_ale_constructor`; `implement_for` only dispatches `_test_gymenv_ale_constructor`. The gym helper is still a no-op so ALE/Pong-v5 is not constructed on gym 0.13/0.19. With gymnasium and no ale_py the two cases skip; with ale_py they construct.

sota-implementations/ppo/utils_atari.py
YeonwooSung: Reverted `make_base_env` defaults in both `sota-implementations/ppo/utils_atari.py` and `sota-implementations/a2c/utils_atari.py` to `BreakoutNoFrameskip-v4`. Hydra configs are unchanged; callers that omit `env_name` get the previous game again.

vmoens (approved):
Re-reviewed 0093999. The classic Atari defaults are restored, the version-dispatched constructor test now skips correctly when ALE is absent, and the import test exercises the actual import boundary without the extra predicate helper.

Validation: 2 focused import tests passed; 2 ALE constructor cases skipped because ale_py is not installed locally. I inspected the reported CPU CI failure: it is in the replay-buffer stale-slot test, outside this change. No remaining code review actionables.

PR #4359, opened 2026-09-12
[BugFix] Mask post-done rewards in CEM and MPPI planners

### LINKED ISSUE #1660: [BUG] Planning (`MPCPlannerBase`) should consider `done`
## Describe the bug

In the current implementation, all subclasses of `MPCPlannerBase` do not consider `done` thrown by env during the planning process, which means that MPC is invalid in a large class of environments. For example, in [CEM](https://github.com/pytorch/rl/blob/10bb5917fff1808e746d1487fd55509a2c2b0db9/torchrl/modules/planners/cem.py#L194C1-L194C1):

```python
            sum_rewards = optim_tensordict.get(self.reward_key).sum(
                dim=TIME_DIM, keepdim=True
            )
```

Specifically, one type of environment indicates that the agent has entered a dangerous state by throwing `done` (usually the reward is positive in non-dangerous states), including many environments of gym-mujoco, such as InvertedPendulum and Hopper. The MPC algorithm needs to identify `done` thrown by the environment and find the action sequence that maximizes the cumulative reward before `done`.

## To Reproduce

Just try CEM on InvertedPendulum.

## Reason and Possible fixes

For CEM, a simple fix chould be:
```python
            dones = optim_tensordict.get(("next", "done"))
            rewards = optim_tensordict.get(self.reward_key)
            assert rewards.shape == dones.shape
            for candidate in range(self.num_candidates):
                if (~dones[candidate]).all():  # no done
                    continue

                if len(dones[candidate].shape) == 1:  # both done and reward are float
                    idx = torch.nonzero(dones[candidate]).min()
                    rewards[candidate, idx:] = 0
                elif len(dones[candidate].shape) == 2:  # both done and reward are 1-dim tensor
                    idx = torch.nonzero(dones[candidate, :, 0]).min()
                    rewards[candidate, idx:, 0] = 0
                else:
                    raise ValueError("Unsupported shape for done and reward")

            sum_rewards = rewards.sum(
                dim=TIME_DIM, keepdim=True
            )
```

I'm more than happy to submit my changes, but they may require further style uniformity and standardization. At the same time, it is likely that there is a more efficient way.

## Checklist

YeonwooSung:
## Description

CEM summed the whole imagined horizon. MPPI rolled out the same way. After `done`, later rewards (including auto-reset) leaked into the plan score.

Both planners now roll out with `break_when_any_done=False` and zero rewards after the first effective environment `done`, keeping the reward at that step. Done keys come from the environment: an ancestor done group takes precedence over its descendants at every nesting level, while independent sibling groups are combined with a logical OR. Scoring uses `done`, not `terminated`, so truncation also stops subsequent rewards from contributing.

### Code example

Minimal scoring regression using the internal helper shared by CEM and MPPI: keep the terminal reward, then ignore the remaining imagined rewards.

```python
import torch
from torchrl.modules.planners.common import _mask_post_done_reward

rollout = TensorDict(
    {"next": {
        "reward": torch.tensor([[1.0], [2.0], [1000.0]]),
        "done": torch.tensor([[False], [True], [False]]),
    }},
    batch_size=[3], names=["time"],
)
# Previously, summing the full imagined horizon gave 1003.
assert rollout["next", "reward"].sum() == 1003
scored_rewards = _mask_post_done_reward(rollout)
torch.testing.assert_close(scored_rewards, torch.tensor([[1.0], [2.0], [0.0]]))
assert scored_rewards.sum() == 3
```

## Motivation and Context

close #1660

## Types of changes

## Checklist

### DIFF
--- a/test/modules/test_planners.py
+++ b/test/modules/test_planners.py
@@ -8,14 +8,182 @@ import argparse
 
 import pytest
 import torch
-from tensordict import TensorDict
+from tensordict import TensorDict, TensorDictBase
 from torch import nn
+from torchrl.data import Composite, Unbounded
+from torchrl.envs import EnvBase
 from torchrl.modules import CEMPlanner, ValueOperator
+from torchrl.modules.planners.common import _mask_post_done_reward
 from torchrl.modules.planners.mppi import MPPIPlanner
 from torchrl.objectives.value import TDLambdaEstimator
 
 from torchrl.testing import get_default_devices
 from torchrl.testing.mocking_classes import MockBatchedUnLockedEnv
+
+FIRST_REWARD = 1.25
+LATER_REWARD = 50.0
+PLANNING_HORIZON = 4
+
+
+class _DoneAfterOneStepEnv(EnvBase):
+    """Batch-unlocked env that is done after one step.
+
+    The first transition yields ``FIRST_REWARD``. After that, a non-breaking
+    rollout auto-resets and later transitions yield ``LATER_REWARD``.
+    """
+
+    @classmethod
+    def __new__(cls, *args, **kwargs):
+        return super().__new__(cls, *args, _batch_locked=False, **kwargs)
+
+    def __init__(self, device="cpu"):
+        super().__init__(device=device)
+        self.observation_spec = Composite(
+            observation=Unbounded((1,), device=device),
+            device=device,
+        )
+        self.action_spec = Unbounded((1,), device=device)
+        self.reward_spec = Unbounded((1,), device=device)
+        self._started = False
+
+    def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
+        if tensordict is None:
+            batch_size = self.batch_size
+            device = self.device
+        else:
+            batch_size = tensordict.batch_size
+            device = tensordict.device if tensordict.device is not None else self.device
+        if tensordict is None or "_reset" not in tensordict.keys():
+            self._started = False
+        done = torch.zeros(*batch_size, 1, dtype=torch.bool, device=device)
+        return TensorDict(
+            {
+                "observation": torch.zeros(
+                    *batch_size, 1, device=device, dtype=torch.get_default_dtype()
+                ),
+                "done": done,
+                "terminated": done.clone(),
+            },
+            batch_size=batch_size,
+            device=device,
+        )
+
+    def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
+        batch_size = tensordict.batch_size
+        device = tensordict.device if tensordict.device is not None else self.device
+        reward_value = LATER_REWARD if self._started else FIRST_REWARD
+        self._started = True
+        done = torch.ones(*batch_size, 1, dtype=torch.bool, device=device)
+        return TensorDict(
+            {
+                "observation": tensordict.get("observation") + 1,
+                "reward": torch.full(
+                    (*batch_size, 1),
+                    reward_value,
+                    device=device,
+                    dtype=torch.get_default_dtype(),
+                ),
+                "done": done,
+                "terminated": done.clone(),
+            },
+            batch_size=batch_size,
+            device=device,
+        )
+
+    def _set_seed(self, seed: int | None) -> None:
+        pass
+
+
+class _SurviveIfPositiveEnv(EnvBase):
+    """Reward is 1 each step. The episode ends when ``action[..., :1] < 0``."""
+
+    @classmethod
+    def __new__(cls, *args, **kwargs):
+        return super().__new__(cls, *args, _batch_locked=False, **kwargs)
+
+    def __init__(self, device="cpu"):
+        super().__init__(device=device)
+        self.observation_spec = Composite(
+            observation=Unbounded((1,), device=device),
+            device=device,
+        )
+        self.action_spec = Unbounded((1,), device=device)
+        self.reward_spec = Unbounded((1,), device=device)
+
+    def _reset(self, tensordict: TensorDictBase, **kwargs) -> TensorDictBase:
+        if tensordict is None:
+            batch_size = self.batch_size
+            device = self.device
+        else:
+            batch_size = tensordict.batch_size
+            device = tensordict.device if tensordict.device is not None else self.device
+        done = torch.zeros(*batch_size, 1, dtype=torch.bool, device=device)
+        return TensorDict(
+            {
+                "observation": torch.zeros(
+                    *batch_size, 1, device=device, dtype=torch.get_default_dtype()
+                ),
+                "done": done,
+                "terminated": done.clone(),
+            },
+            batch_size=batch_size,
+            device=device,
+        )
+
+    def _step(self, tensordict: TensorDictBase) -> TensorDictBase:
+        action = tensordict.get("action")
+        device = action.device
+        done = action[..., :1] < 0
+        return TensorDict(
+            {
+                "observation": tensordict.get("observation") + 1,
+                "reward": torch.ones(
+                    *tensordict.batch_size,
+                    1,
+                    device=device,
+                    dtype=torch.get_default_dtype(),
+                ),
+                "done": done,
+                "terminated": done.clone(),
+            },
+            batch_size=tensordict.batch_size,
+            device=device,
+        )
+
+    def _set_seed(self, seed: int | None) -> None:
+        pass
+
+
+class _SumRewardAdvantage(nn.Module):
+    """Advantage equal to the trajectory return, written at every time step."""
+
+    def forward(self, tensordict: TensorDictBase) -> TensorDictBase:
+        reward = tensordict.get(("next", "reward"))
+        ret = reward.sum(dim=-2, keepdim=True)
+        tensordict.set("advantage", ret.expand_as(reward))
+        return tensordict
+
+
+def _make_planner(name, env, *, planning_horizon=PLANNING_HORIZON):
+    if name == "cem":
+        return CEMPlanner(
+            env,
+            planning_horizon=planning_horizon,
+            optim_steps=1,
+            num_candidates=8,
+            top_k=2,
+            reward_key=("next", "reward"),
+        )
+    return MPPIPlanner(
+        env,
+        _SumRewardAdvantage(),
+        temperature=1.0,
+        planning_horizon=planning_horizon,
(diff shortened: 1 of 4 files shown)


torchrl/modules/planners/common.py
```diff
@@ -8,29 +8,61 @@
...
+    time_dim: int | None = None,
+) -> torch.Tensor:
+    """Zero rewards that follow the first ``done`` along the time dimension.
+
+    The reward at the first ``done`` step is kept. ``done`` is used (not
+    ``terminated``) so a truncated episode also stops contributing to the
+    planning score.
+    """
+    reward = tensordict.get(reward_key)
+    done = tensordict.get(done_key)
```
vmoens: [P2] Resolve the environment done key when calling the mask

Both planners invoke this helper without `done_key`, so they always read `("next", "done")`. EnvBase supports done flags only under a nested group. With an otherwise identical batch-unlocked planning env whose flags are `("agent", "done")` and `("agent", "terminated")`, both CEM and MPPI reach this code with `done=None` and fail on `done.shape`. Derive the applicable done key(s) from the environment and pass them through. The nested helper test does not exercise either public planner, so extend that coverage through a planner call.

test/modules/test_planners.py
```diff
@@ -5,18 +5,343 @@
...
+FIRST_REWARD = 1.25
+LATER_REWARD = 50.0
+PLANNING_HORIZON = 4
+DIE_REWARD = 1.0
+LIVE_REWARD = 5.0
+JACKPOT_REWARD = 50.0
+TWO_SEQ_HORIZON = 3
+
+
+class _DoneAfterOneStepEnv(EnvBase):
```
vmoens: [P2] Consolidate the overlapping regression scaffolding

This adds about 520 test lines for roughly 70 implementation/docstring additions (over 7:1). Three new env classes, a separate rollout/reference implementation, private-mask tests, and multiple planner tests largely cover the same post-done reward behavior. Keep the deterministic two-candidate end-to-end regression parametrized over CEM/MPPI and termination/truncation, reuse existing env scaffolding where possible, and retain extra low-level coverage only for a distinct invariant. This would preserve the regression signal without maintaining several overlapping implementations.

vmoens (changes requested):
The scalar root-done cases work and all 16 planner tests pass, but both planners now fail for a valid environment with nested done keys. The test addition also needs consolidation.

torchrl/modules/planners/common.py
YeonwooSung: CEM and MPPI now resolve done keys from `env.done_keys`, keep only entries that end in `"done"` (so truncation still masks), prefix them with `"next"`, and pass that list into `_mask_post_done_reward`. A root `("next", "done")` is not synthesized when the env only exposes `("agent", "done")` / `("agent", "terminated")`. `test_planner_nested_done_keys` runs both planners on that env and checks that the live sequence is still selected.

test/modules/test_planners.py
YeonwooSung: Dropped the extra env classes and the overlapping one-step / survive / private-mask value tests. The remaining regression is the deterministic two-candidate case, parametrized over CEM/MPPI and terminate/truncate. Nested done keys are covered by a planner call on the same env (`done_group="agent"`); the helper test is kept only for nested-key lookup.

torchrl/modules/planners/common.py
```diff
@@ -5,32 +5,99 @@
...
 
+def _planning_done_keys(env: EnvBase) -> list[NestedKey]:
+    """Return rollout ``done`` keys used to mask post-done planning rewards.
+
+    Only keys that end with ``"done"`` are used, so a truncated episode
+    (``done`` and not ``terminated``) still stops contributing. Keys are
+    taken from the environment and prefixed with ``"next"``; a root
+    ``"done"`` is not invented when the env only exposes a nested group.
+    """
+    return [
```
vmoens: [P2] Preserve the environment's root/nested termination semantics

Collecting every done key and OR-ing them makes an agent-level done override a false root done. EnvBase's _terminated_or_truncated instead lets a root done group take precedence over deeper groups. For an environment with a root reward, one agent can finish while the environment continues, and those later rewards must still count.

The mask currently turns the valid return 101 into 1:

```python
import torch
from torchrl.modules.planners.common import _mask_post_done_reward

rollout = TensorDict({"next": {
    "reward": torch.tensor([[[1.], [50.], [50.]]]),
    "done": torch.zeros(1, 3, 1, dtype=torch.bool),
    "agent": {"done": torch.ones(1, 3, 1, dtype=torch.bool)},
}}, batch_size=[1, 3])
score = _mask_post_done_reward(
    rollout, done_key=[("next", "done"), ("next", "agent", "done")]
).sum()
# score is 1; root done is false throughout, so the env return is 101.
```

Using the PR's deterministic two-sequence fixture with both root and nested specs, CEM and MPPI both choose the competing return-15 sequence. Please derive the effective planning termination keys/mask consistently with EnvBase, preserving nested-only support, and cover coexisting root and nested groups in the public planner regression.

vmoens (changes requested):
Re-reviewed e41c33b. Nested-only done specs now work and all 11 planner tests pass. There is still a correctness issue when root and nested done keys coexist: a nested agent finishing currently cuts off an environment trajectory whose root done remains false. I reproduced the wrong selected action in both CEM and MPPI; details inline.

The test diff is smaller, but still about 305 added test lines versus 118 implementation/doc lines (2.6:1). Please consolidate the overlapping root/nested cases into the public planner parameter table and remove the duplicated rollout/reference-mask checks where the public action-selection regression covers the same behavior. Also run the formatter on common.py: that file is the reported ufmt CI failure.

torchrl/modules/planners/common.py
YeonwooSung: Root done now takes precedence over nested groups in both `_planning_done_keys` and `_mask_post_done_reward`, matching EnvBase. Nested-only environments still use the nested `done` key(s). With the two-sequence fixture exposing both specs (root `done` false, nested `agent.done` true), CEM and MPPI keep the jackpot return 101 instead of selecting the competing return-15 sequence.

The public planner table is parametrized over CEM/MPPI, terminate/truncate, and `done_layout` in {root, nested, both}. The private helper cases and duplicated rollout/reference-mask checks were dropped. `common.py` is ufmt-clean.

YeonwooSung:
Addressed the root/nested done review on 3895f8f32.

A root-level `done` now takes precedence over nested groups in `_planning_done_keys` and `_mask_post_done_reward` (EnvBase semantics). Nested-only environments still use the nested `done` key(s). The public CEM/MPPI table covers root, nested-only, and coexisting groups; overlapping helper and rollout/reference-mask checks were removed. `common.py` is ufmt-clean.

torchrl/modules/planners/common.py
```diff
@@ -5,32 +5,121 @@
...
+        return (unravel_key(done_key),)
+    return tuple(unravel_key(key) for key in done_key)
+
+
+def _effective_planning_done_keys(
+    done_key: NestedKey | Sequence[NestedKey],
+) -> tuple[NestedKey, ...]:
+    keys = _normalize_done_keys(done_key)
+    # EnvBase lets a root done group take precedence over nested groups.
+    root_keys = tuple(key for key in keys if _is_root_planning_done_key(key))
```
vmoens: [P2] Apply done-group precedence at every nesting level

When the keys are `("team", "done")` and `("team", "agent", "done")`, the team flag must prevail over its child just as root done prevails over agent done. `_terminated_or_truncated` applies this rule recursively. Here neither key is recognized as root, so both are OR-ed and an early child completion truncates the still-running parent's return:

```python
rollout = TensorDict({"next": {
    "reward": torch.tensor([[1.0], [50.0], [50.0]]),
    "team": {
        "done": torch.zeros(3, 1, dtype=torch.bool),
        "agent": {"done": torch.tensor([[True], [False], [False]])},
    },
}}, batch_size=[3], names=["time"])
keys = [("next", "team", "done"), ("next", "team", "agent", "done")]
_mask_post_done_reward(rollout, done_key=keys).sum()
# Returns 1; the valid parent trajectory returns 101.
```

Exclude descendant done groups when an ancestor group declares done, while keeping independent sibling groups. The same fixture used in the public planner table reproduces the wrong CEM/MPPI selection when its current `both` done layout is nested under `team`.

vmoens (changes requested):
Rechecked 3895f8f32. The root-plus-nested case is fixed, the planner tests are consolidated, and all 16 planner tests plus ufmt pass. One termination hierarchy case remains: parent precedence is recursive in EnvBase, but the new filter recognizes only a top-level root.

I verified a nested-parent environment with `check_env_specs()`. EnvBase reports it still running, while both CEM and MPPI select action 5 (return 15) instead of action -1 (return 101) because the deeper agent's done still masks the parent trajectory. Add this layout to the existing public planner table rather than adding another parallel fixture/test layer.

torchrl/modules/planners/common.py
YeonwooSung: `_effective_planning_done_keys` now drops any done key whose group prefix has an ancestor in the provided list, matching EnvBase recursive precedence. `("next", "team", "done")` prevails over `("next", "team", "agent", "done")`; independent sibling groups stay OR-ed; a nested-only key is still used. The helper snippet now returns 101.

The existing public planner table adds `done_layout="nested_parent"` (the previous `both` layout nested under `team`). CEM/MPPI select action -1 (return 101) instead of 5 (return 15). Root / nested / both coverage is unchanged. a769f2e01.

YeonwooSung:
a769f2e01 applies done-group precedence at every nesting level in `_effective_planning_done_keys`. An ancestor group (including a nested parent such as `team`) excludes descendant `done` keys; siblings are still OR-ed. The public planner table now includes `done_layout="nested_parent"` alongside root / nested / both. Helper snippet sum is 101; `pytest test/modules/test_planners.py` is 20 passed.

vmoens (approved):
Rechecked a769f2e01. Recursive ancestor precedence fixes the last planner finding while preserving independent sibling groups. The previously failing nested-parent environment passes spec checks, and both CEM and MPPI now choose the trajectory returning 101 instead of the one returning 15. The expanded public planner table covers root, nested-only, coexisting root/child, and nested parent/child layouts.

All 20 planner tests pass, and ufmt accepts all changed Python files. No remaining actionable findings.

PR #4360, opened 2026-09-12
[Feature] Add LastAction transform

### LINKED ISSUE #1676: [Feature Request] Add a Transform that remembers the previous action taken

YeonwooSung:
## Description

Adds `LastAction`, a CatFrames-style transform that copies the previous action into the next observation.

- NestedKey `in_keys` / `out_keys` (default last component `last_action`)
- Reset default: zeros, nan, or a provided value
- Action spec cloned into the observation spec
- Hydra `LastActionConfig` with constructor parity

### Code example

With Gymnasium installed, expose the previous action to the policy on the next step. Reset starts with zeros.

```python
import torch
from torchrl.envs import GymEnv, TransformedEnv
from torchrl.envs.transforms import LastAction

env = TransformedEnv(GymEnv("Pendulum-v1"), LastAction())
td = env.reset()
assert torch.count_nonzero(td["last_action"]) == 0

td = env.rand_action(td)
transition = env.step(td)
torch.testing.assert_close(transition["next", "last_action"], td["action"])
env.close()
```

## Motivation and Context

close #1676

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist


torchrl/envs/transforms/_action.py
```diff
@@ -2340,3 +2343,245 @@ def transform_output_spec(self, output_spec: Composite) -> Composite:
...
+
+    def _call(self, next_tensordict: TensorDictBase) -> TensorDictBase:
+        return next_tensordict
+
+    def _reset(
+        self, tensordict: TensorDictBase, tensordict_reset: TensorDictBase
+    ) -> TensorDictBase:
+        _reset = _get_reset(self.reset_key, tensordict)
+        for in_key, out_key in _zip_strict(self.in_keys, self.out_keys):
+            fill = self._make_default(in_key)
```
vmoens: [P2] Size reset defaults from the actual reset batch

`_make_default` uses only the parent action spec shape. A batch-unlocked parent can receive a larger TensorDict batch at reset: `TransformedEnv(TicTacToeEnv(), LastAction()).reset(TensorDict(batch_size=[2]))` currently raises `RuntimeError: batch dimension mismatch` because the default is scalar. Expand the action spec/default by the runtime batch dimensions, preserving the action feature shape, and cover this public reset path.

torchrl/envs/transforms/_action.py
```diff
@@ -2340,3 +2343,245 @@ def transform_output_spec(self, output_spec: Composite) -> Composite:
...
+    ) -> TensorDictBase:
+        for in_key, out_key in _zip_strict(self.in_keys, self.out_keys):
+            action = tensordict.get(in_key, default=None)
+            if action is None:
+                if not self.missing_tolerance:
+                    raise KeyError(
+                        f"{self}: '{in_key}' not found in tensordict {tensordict}"
+                    )
+                continue
+            next_tensordict.set(out_key, action)
```
vmoens: [P2] Keep the remembered action independent of the input buffer

This stores the same tensor object, so reusing or modifying the input action changes the already-produced `next.last_action`. I reproduced this by stepping a ContinuousActionVecMockEnv and then calling `td["action"].fill_(123)`; the returned previous-action observation also becomes 123. Policies that reuse output buffers must not rewrite the remembered action. Copy/clone the action when recording it and exercise buffer reuse in the regression coverage.

vmoens (changes requested):
The ordinary and nested-key paths pass, but two reset/storage issues need fixing before approval. Validation: 15 focused transform tests and the Hydra test passed (one inverse test skipped); separate probes reproduced both findings. Please also replace the new argument-binding lambdas with partials, avoid the config test assertions that only restate constructor arguments, and add the benchmark required for a new stepping transform.

torchrl/envs/transforms/_action.py
YeonwooSung: `_make_default` now takes the reset tensordict batch and, on a batch-unlocked parent, calls `spec.zero(runtime_batch)` so the fill is `(*runtime_batch, *action_feature)`. `TransformedEnv(TicTacToeEnv(), LastAction()).reset(TensorDict(batch_size=[2]))` is covered by `test_reset_unlocked_runtime_batch`.

Also in this push: SerialEnv/ParallelEnv factories use `functools.partial` instead of lambdas; the Hydra test instantiates `LastActionConfig(default=-1.0)` and asserts reset/step behavior rather than stored kwargs; `benchmarks/test_envs_benchmark.py` adds `test_last_action_reset` / `test_last_action_step`.

torchrl/envs/transforms/_action.py
YeonwooSung: `_step` now stores `action.clone()`, so an in-place write to the policy action buffer cannot rewrite `next.last_action`. `test_step_does_not_alias_action` steps `ContinuousActionVecMockEnv` and `fill_(123)`s the action; the remembered observation stays at the original values.

test/test_configs.py
```diff
@@ -4635,6 +4635,20 @@ def test_init_tracker_config(self):
         assert cfg.init_key == "is_test_init"
         instantiate(cfg)
 
+    @pytest.mark.skipif(not _has_hydra, reason="Hydra is not installed")
+    def test_last_action_config(self):
+        from hydra.utils import instantiate
+        from torchrl.envs import TransformedEnv
```
vmoens: [P3] Move the non-optional TorchRL imports to module scope

Please move TransformedEnv and ContinuousActionVecMockEnv to the module-top imports. Neither is an optional dependency or a circular-import case. The Hydra-dependent imports can retain their existing optional-dependency handling. The behavioral test itself passes.

vmoens (commented):
Re-reviewed c5c866f. Both functional findings are resolved: reset uses the runtime batch for an unlocked environment, and next.last_action owns a copy of the input action. The Hydra test now exercises the transform, and reset/step benchmarks have been added.

Validation: 20 focused tests passed, 1 inverse test skipped, and both new benchmark functions passed a smoke run with benchmarking disabled. Only the import placement cleanup noted inline remains before approval.

test/test_configs.py
YeonwooSung: Moved `TransformedEnv` and `ContinuousActionVecMockEnv` to the module-top imports. `hydra.utils.instantiate` and `LastActionConfig` stay local to the Hydra-gated test. a6b8e2179

vmoens (approved):
Re-reviewed a6b8e21. The remaining non-optional imports are now at module scope. The runtime-batch reset and action-aliasing fixes remain intact; 20 focused LastAction/Config tests pass, with the inverse test skipped as expected. No remaining actionables.

PR #4363, opened 2026-09-12
[BugFix] Fix GRPOLoss padding config and DAPO instantiation

coder-jayp:
## Description

This PR addresses two structural problems in the `GRPOLoss` family and its KL-divergence handling logic:

1. **GRPOLoss Padding Refactor**: Added `ref_log_prob_padding_side` and `ref_log_prob_padding_value` as kwargs to `GRPOLoss.__init__`. Previously, `forward()` would hardcode left padding with `0.0` and fetch the reference log probabilities directly, bypassing `_kl_to_ref`'s key fetching logic and causing `set_keys(ref_log_probs=...)` to be silently ignored for the KL penalty path.
2. **DAPO Instantiation Bug**: Removed the broken `DAPO.__init__` override. It was incorrectly copied from the body of a loss function and expected a `tensordict` as its first positional argument, which resulted in a `TypeError` when users tried to instantiate it with a standard `actor_network`. `DAPO` now correctly inherits from `GRPOLoss`.

Added robust behavioral regression tests for both fixes to ensure they do not regress (testing the actual behavior rather than internal implementation details).

## Motivation and Context
These changes are required to ensure that `GRPOLoss` and its subclasses strictly adhere to TorchRL's `TensorDict` contracts. Specifically:
- Users need to be able to use `set_keys()` to configure where `ref_log_probs` are pulled from.
- Users need to be able to instantiate `DAPO` with an `actor_network` just like any other policy loss module.
- It solves the inline `TODO`/`FIXME` comments left in `grpo.py` regarding parameterizing the padding.

- [ ] I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes) *(Note: If you haven't opened an issue yet, you can leave this unchecked or open a quick issue to link here!)*
## Types of changes
## Checklist

### DIFF
--- a/test/llm/test_llm_objectives.py
+++ b/test/llm/test_llm_objectives.py
@@ -28,6 +28,7 @@ from torchrl.objectives.llm.distillation import (
 )
 from torchrl.objectives.llm.grpo import (
     CISPOLoss,
+    DAPO,
     GRPOLoss,
     GRPOLossOutput,
     MCAdvantage,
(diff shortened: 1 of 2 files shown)


coder-jayp:
@torchrlbot reviewer @vmoens

vmoens (changes requested):
Thanks for tackling this — the direction is right and the DAPO fix is genuinely valuable. I verified the behavior empirically (CPU env, torch 2.14 / tensordict 0.14.2) by running the PR's new tests against both `main` and the PR head. A few findings below; the main one is a blocker.

## ✅ The DAPO half is correct and valuable

Verified: on `main`, `DAPO(actor_network, clip_epsilon=0.2)` raises `TypeError: DAPO.__init__() got an unexpected keyword argument 'clip_epsilon'` — DAPO was un-instantiable. On the PR head, `DAPO(policy, clip_epsilon=(0.2, 0.28))` constructs cleanly via the inherited `GRPOLoss.__init__` and keeps `output_type = DAPOLossOutput`. The removal is total (no other code references that signature). The accompanying test passes on the head and fails on `main`. Good regression test.

## 🟡 The `set_keys` bug is real — but misdiagnosed in the description and tests

The PR says the old prefetch "bypassed `_kl_to_ref`'s key fetching logic and caused `set_keys(ref_log_probs=...)` to be silently ignored". That's not quite what happens: the old prefetch already used `self.tensor_keys.ref_log_probs`, i.e., the configured key. I ran the PR's own `test_set_keys_ref_log_probs_is_respected_by_kl_to_ref` against **unpatched `main` — it passes** (`kl_to_ref = 0`).

The real bug is in the *missing-key* path: on `main`, when the configured key is absent from the tensordict, the prefetch returns `None` and `_kl_to_ref` silently falls back to the hardcoded `("next", "ref_log_probs")` key. I reproduced the loss computing `kl_to_ref = 0.1931` from data stored under a key the user never configured. Your fix (routing `key=self.tensor_keys.ref_log_probs` through) genuinely addresses this — but a test that distinguishes old from new must cover the *missing* configured key: the new code should raise `KeyError` naming the configured key, where the old code silently consumed another key's values. That case is currently untested.

## 🔴 The refactor introduces a crash for `T == 1` — the PR's own new tests catch it (2 of 3 fail)

I ran the PR's actual `TestGRPOLossRefactorBehavior` class against the PR head code:

```
FAILED test_set_keys_ref_log_probs_is_respected_by_kl_to_ref
FAILED test_kl_to_ref_hand_calculated_value
1 passed (dapo), 2 failed
RuntimeError: expand_as_right requires the destination tensor to have less dimensions
  than the input tensor, got tensor.ndimension()=2 and dest.ndimension()=1
```

Root cause: previously the forward prefetch passed `ref_log_prob` **unsqueezed**; now `_kl_to_ref` always applies `.squeeze(-1)` to fetched refs, while `_cur_log_prob` and `mask` keep their trailing dim. Any single-token (`T == 1`) sequence — exactly what both failing tests construct via `_policy_loss_data` — collapses `(B, 1)` → `(B,)` and blows up in `expand_as_right`. A plain `B=2, T=1` example that works fine on `main` also crashes on the head.

So the PR ships tests that fail against its own code (CI hasn't run them yet — workflows are pending maintainer approval for external contributors, which is presumably why nobody noticed), and the change regresses a case that works on `main`. Suggested fix: squeeze only when the fetched tensor actually has the extra trailing feature dim, e.g. condition on `ref_log_prob.ndim == cur_log_prob.ndim + 1` (or `ref_log_prob.shape[-1] == 1 and ref_log_prob.ndim == cur_log_prob.ndim + 1`), and keep a shape check with padding context in the error message.

(The same `T == 1` squeeze hazard already exists on `main` for the `kl_to_inference` path, which already fetched+squeezed inside `_kl_to_ref` — an existing landmine worth fixing while you're in there, not a blocker for this PR's direction.)

## 🔸 Smaller issues

- **Docstring nit**: the `GRPOLoss` docstring switched to `r"""` and now contains `Defaults to \"sft\"` — in a raw string those backslashes render literally in the Sphinx docs. The escaping is unnecessary (`"sft"` is fine inside `"""`); drop the backslashes (and probably the `r` prefix). Unindenting the `.. note::` about parameters/buffers out of the `masking_strategy` param block is a nice fix unto itself, though. 👍
- **Removed shape check**: removing it per the TODO is fine, but keep *some* guard — mismatches now surface as the cryptic `expand_as_right` error above, and broadcastable shape mismatches (e.g. `(B,T)` vs `(B,T,1)`) would otherwise pass silently through `torch.where` and produce wrong numbers.
- **Out-of-scope hunks**: the reflows in `torchrl/collectors/llm/base.py` and `test/llm/test_llm_collectors.py` are purely cosmetic and unrelated to this bugfix — I'd drop them from this PR.

## Test coverage

- Intent is right — these target real behavior, not trivial wiring.
- `test_kl_to_ref_hand_calculated_value` (k3 estimator, hand-computed) is a good *value* test once the squeeze regression is fixed.
- `test_set_keys_ref_log_probs_is_respected_by_kl_to_ref` doesn't test what its name/docstring says (it passes on the old code) — retarget it to the missing-key fallback described above.
- Consider an explicit `T == 1` case (single-token sequences), since that's what exposed the squeeze bug.

## Verdict

**Requesting changes.** The DAPO removal and the `key=` routing are right, and they fix real bugs. But as submitted, the PR crashes for single-token sequences (2 of its own 3 new tests fail), misdescribes the bug it fixes, and includes unrelated formatting churn. Fix the squeeze handling, retarget the `set_keys` test to the missing-key fallback, drop the drive-by hunks and the `r"""`/escaping noise, and this should be good to go.

coder-jayp:
@vmoens All feedback addressed!

* **T=1 crash:** Fixed. `_kl_to_ref` now fetches `cur_log_prob` first and conditionally squeezes only when `ref_log_prob.ndim == cur_log_prob.ndim + 1`. This also safely resolves the same squeeze hazard on the inference path. Added a shape guard with a descriptive `ValueError`.

* **`set_keys` missing-key fallback:** Fixed. The test was rewritten to explicitly target the fallback bug. It now configures a custom key but stores data only under the default key. As requested, it asserts that the new code raises `KeyError` naming the missing custom key.

* **T=1 test:** Added `test_kl_to_ref_single_token_sequence` to explicitly document and catch the regression.

* **Docstring:** Removed the `r"""` prefix and the `\"` escapes around `"sft"`.

* **Out-of-scope hunks:** Removed. The purely cosmetic changes were dropped via a clean rebase onto `upstream/main`.

vmoens (approved):

PR #4365, opened 2026-09-13
[BugFix] Fix star-import of torchrl.modules.distributions

Anai-Guo:
## Description

`torchrl/modules/distributions/__init__.py` has two problems in `__all__`:

1. **`"distributions"` isn't defined in the module.** The only similar names are `torch_dist` (the `torch.distributions` alias) and `distributions_maps`. Because of this, a star-import fails:

   ```console
   $ python -c "from torchrl.modules.distributions import *"
   AttributeError: module 'torchrl.modules.distributions' has no attribute 'distributions'. Did you mean: 'distributions_maps'?
   ```

   The entry has been there since `__all__` was first added to this file (the pyupgrade lint pass). This PR removes it.

2. **`"LLMMaskedCategorical"` is listed twice.** It was added once in #3041 and again in #3030. This PR removes the second copy.

That's `+0/−2`, and no runtime code changes.

`torchrl.modules` imports `distributions_maps` explicitly, so it isn't affected. If you'd like `distributions_maps` exported from `torchrl.modules.distributions` too, I can replace the stale entry with it instead of deleting it.

## Motivation and Context

Linters don't catch this: ruff and flake8 don't check that `__all__` entries resolve, and nothing in the test suite does a star-import of this module.

Verified against the `torchrl==0.14.0` wheel, whose `distributions/__init__.py` is byte-identical to `main`:
- before: the star-import raises the `AttributeError` above
- after: the star-import succeeds, every `__all__` entry resolves, there are no duplicates, and `import torchrl.modules` still works

## Types of changes

## Checklist

AI-assisted, human reviewed.


coder-jayp (approved):
Thanks LGTM

coder-jayp:
@Anai-Guo Could you also remove the duplicate LLMMaskedCategorical from torchrl/modules/__init__.py

Anai-Guo:
Done — removed the duplicate `LLMMaskedCategorical` from `torchrl/modules/__init__.py` `__all__` in the latest commit. Thanks for the review, @coder-jayp!

Anai-Guo:
Good catch, @coder-jayp — done. I merged current `main` into the branch and the duplicate `"LLMMaskedCategorical"` entry in `torchrl/modules/__init__.py`'s `__all__` is now removed (the diff against `main` shows a single `-"LLMMaskedCategorical",`), so the name is exported exactly once.

torchrl/modules/__init__.py
```diff
@@ -291,4 +290,4 @@
     "symexp",
     "symlog",
     "RandomPolicy",
-]
+]
```
vmoens, line 293: This is blocking, otherwise LGTM

PR #4368, opened 2026-09-13
[BugFix] Build TD3Trainer with multiprocessing collectors

### LINKED ISSUE #4366: [BUG] TD3Trainer cannot be built with AsyncCollector, MultiSyncCollector or MultiAsyncCollector (reads collector.env)
## Describe the bug

`_make_td3_trainer` reads the action spec from `collector.env`:

```python
env = collector.env
action_spec = getattr(env, "action_spec_unbatched", None) or env.action_spec
```

Only the single-process `Collector` has an `env` attribute. `MultiSyncCollector`, `MultiAsyncCollector` and `AsyncCollector` (which subclasses `MultiAsyncCollector`) do not, so building a `TD3Trainer` with any of them raises `AttributeError`. Of the four registered collector configs (`sync`, `async`, `multi_sync`, `multi_async`), only `sync` works. The shipped `sota-implementations/td3_trainer` config uses `sync`, so it is fine as is, but switching it to a multi-process collector (as `sac_trainer` does with `multi_async`) fails at trainer construction.

TD3 is the only trainer factory that touches `collector.env`. `_make_ddpg_trainer` and `_make_sac_trainer` never do, and work with every collector.

## To Reproduce

```python
import functools

import torch
from torchrl.collectors import AsyncCollector, Collector, MultiAsyncCollector, MultiSyncCollector
from torchrl.modules import TanhModule, ValueOperator
from torchrl.objectives import SoftUpdate, TD3Loss
from torchrl.testing.mocking_classes import ContinuousActionVecMockEnv
from torchrl.trainers.algorithms.configs.trainers import _make_td3_trainer


def make_env():
    return ContinuousActionVecMockEnv()


def build_trainer(collector_cls):
    actor = TensorDictSequential(
        TensorDictModule(torch.nn.Linear(7, 7), in_keys=["observation"], out_keys=["param"]),
        TanhModule(in_keys=["param"], out_keys=["action"], low=-1, high=1),
    )
    qvalue = ValueOperator(
        torch.nn.Linear(14, 1), in_keys=["observation", "action"], out_keys=["state_action_value"]
    )
    env_fn = make_env if collector_cls in (Collector, AsyncCollector) else [make_env, make_env]
    collector = collector_cls(env_fn, policy=actor, frames_per_batch=8, total_frames=16)
    try:
        _make_td3_trainer(
            collector=collector,
            total_frames=16,
            loss_module=functools.partial(TD3Loss),
            actor_network=actor,
            qvalue_network=qvalue,
            target_net_updater=functools.partial(SoftUpdate, eps=0.99),
            optimizer=functools.partial(torch.optim.Adam, lr=1e-3),
            logger=None,
            clip_norm=None,
            replay_buffer=None,
            save_trainer_file=None,
            seed=0,
            create_env_fn=None,
        )
        return "ok"
    except AttributeError as exc:
        return f"AttributeError: {exc}"
    finally:
        collector.shutdown()


if __name__ == "__main__":
    for cls in (Collector, AsyncCollector, MultiSyncCollector, MultiAsyncCollector):
        print(f"{cls.__name__:20s} {build_trainer(cls)}")
```

```bash
Collector            ok
AsyncCollector       AttributeError: 'AsyncCollector' object has no attribute 'env'
MultiSyncCollector   AttributeError: 'MultiSyncCollector' object has no attribute 'env'
MultiAsyncCollector  AttributeError: 'MultiAsyncCollector' object has no attribute 'env'
```

All three failures come from `torchrl/trainers/algorithms/configs/trainers.py:1548`.

There is no workaround from the config side either. `TD3LossConfig` has `bounds` and `action_spec` fields, but setting `bounds` on the loss does not help: with a multi-process collector the factory still fails at `collector.env` before reaching the loss, and with the single-process `Collector` the factory injects `action_spec` on top, so `TD3Loss` raises `One of 'bounds' and 'action_spec' must be provided, but not both or none`.

## Expected behavior

`TD3Trainer` can be built with any of the registered collectors, or at least there is a supported way to give it the action bounds when the collector has no local env.

## System info

* TorchRL source checkout at `aa442e4b204adf9ea532b981421b4f4c5ed17573`
* Linux, Python 3.13
* PyTorch `2.11.0`, TensorDict `0.14.2`, hydra-core `1.3.6`

## Reason and Possible fixes

The obvious fix, building one env from `collector.create_env_fn[0]` in the main process to read its spec, goes against what `_multi_base.py` says about `fake_tensordict`: creating an env in the main process "defeats the purpose of a multi-process collector" because some backends (Isaac Lab, mujoco-mjx) can only run in workers. So I did not want to pick a fix without asking.

Two options that keep the env out of the main process:

1. Let the loss config carry the bounds. In `_make_td3_trainer`, only read the env spec when the loss partial does not already provide `bounds` or `action_spec`, and raise a clear error when it has neither and the collector has no `env`. This also makes the existing `TD3LossConfig.bounds` field usable from the trainer.
2. Take the bounds from the actor. The action bounds already live on the actor's `TanhModule` (as `action_low` / `action_high` buffers when built from `low` / `high`), though the `TensorDictSequential` wrapper does not expose a spec, so this would need a small lookup.

Option 1 seems the least surprising to me. Happy to send a PR with a regression test for whichever you prefer.

## Checklist

aswanth-07:
## Description

Allow `TD3Trainer` to use action bounds already configured on its loss partial. The trainer now reads a local collector environment only when the loss does not provide `bounds` or `action_spec`.

This lets `AsyncCollector`, `MultiSyncCollector`, and `MultiAsyncCollector` build a TD3 trainer without creating an environment in the main process. The existing single-process fallback still derives the action spec from `collector.env`.

The regression builds the trainer against collector interfaces that deliberately have no `env` attribute and verifies that the configured bounds reach `TD3Loss` unchanged.

## Motivation and Context

Closes #4366.

## Types of changes

## Validation

- `python -m pytest -q test/test_configs.py -k "td3_trainer_uses_loss_bounds_without_collector_env or td3_trainer_config"` (4 passed)
- `python -m ruff check torchrl/trainers/algorithms/configs/trainers.py test/test_configs.py`
- `git diff --check`

## Checklist

AI assistance disclosure: AI-assisted development tools were used during investigation and implementation. The reported tests were run against the final diff.


test/test_configs.py
```diff
@@ -2803,6 +2805,54 @@ def test_td3_trainer_config(self):
         assert cfg.policy_update_delay == 2
         assert cfg.clip_grad_norm is True
 
+    @pytest.mark.parametrize(
+        "collector_cls", [AsyncCollector, MultiSyncCollector, MultiAsyncCollector]
+    )
+    def test_td3_trainer_uses_loss_bounds_without_collector_env(self, collector_cls):
+        from torchrl.objectives import SoftUpdate, TD3Loss
+        from torchrl.trainers.algorithms.configs.trainers import _make_td3_trainer
```
vmoens, line 2813: Nonblocking test cleanup: the three class-spec mocks exercise the same missing-`env` branch without starting any workers. With roughly 50 added test lines for 18 implementation lines (~2.8:1), keep the useful real actor/loss setup but prefer parametrizing meaningful cases such as explicit `bounds`, explicit `action_spec`, and the local-env fallback over equivalent mock interfaces. This would cover more of the changed behavior without growing the fixture. Also move the newly added imports to module scope, per the repository import rule.

vmoens (approved):
The TD3 fix looks correct and appropriately scoped: explicitly configured `bounds` or `action_spec` take precedence, and the existing local-environment fallback remains available. No blocking correctness findings.

Validation at `0a7a9ed`: the four targeted pytest cases passed; additional construction checks covered explicit bounds/specs, a Composite spec, a Hydra loss partial, local fallback/overrides, and rejection of missing or conflicting action domains. Construction also passed with real `AsyncCollector`, `MultiSyncCollector`, and `MultiAsyncCollector` workers, including shutdown. The broader CI workflows are still awaiting approval.

Follow-ups to keep separate from this fix:

- Apply the same “only query the environment when necessary” principle to [SACTrainer](https://github.com/pytorch/rl/blob/0a7a9edefb2e2e3cc56a6839f84b4bdf2da09901/torchrl/trainers/algorithms/sac.py#L276-L278) and [CQLTrainer](https://github.com/pytorch/rl/blob/0a7a9edefb2e2e3cc56a6839f84b4bdf2da09901/torchrl/trainers/algorithms/cql.py#L193-L199). They already respect an explicit loss action spec, but still call `getattr_env` when it is absent even if numeric `target_entropy` or the actor's spec makes that query unnecessary. This matters for collectors that do not expose that accessor.
- Remove analogous local-env assumptions from other training recipes, notably [GAIL's `make_gail_discriminator(cfg, collector.env, ...)`](https://github.com/pytorch/rl/blob/0a7a9edefb2e2e3cc56a6839f84b4bdf2da09901/sota-implementations/gail/gail.py#L143-L144), whose helper only needs observation/action specs. TD3 could also eventually infer omitted bounds through the existing `collector.getattr_env` API where supported.
- Support async environments and async policy inference consistently across **all recipes**, using `torchrl.envs.AsyncEnvPool` and `torchrl.modules.inference_server.InferenceServer` / `PolicyClientModule`, with [`AsyncBatchedCollector`](https://github.com/pytorch/rl/blob/0a7a9edefb2e2e3cc56a6839f84b4bdf2da09901/torchrl/collectors/_async_batched.py#L528-L564) as the integrated path. Expose batching/device placement through `InferenceServerConfig` / `InferenceDeviceConfig` and shared Hydra configuration; the current registry only has `sync`, `async`, `multi_sync`, and `multi_async`. Include recipe smoke coverage for collection, weight synchronization, and shutdown, while preserving each algorithm's rollout and policy-freshness requirements.

PR #4370, opened 2026-09-13
[BugFix] Restore TensorDictMaxValueWriter heap entries as tuples on loads

### LINKED ISSUE #4369: [BUG] TensorDictMaxValueWriter raises TypeError on the first write after ReplayBuffer.loads
## Describe the bug

A `TensorDictMaxValueWriter` restored with `ReplayBuffer.loads` cannot be written to anymore. The first `add` or `extend` after loading raises `TypeError: '<' not supported between instances of 'tuple' and 'list'`, both when the buffer still has free slots and when it is full. The buffer that was saved keeps working, so a checkpointed top-k buffer cannot be resumed.

## To Reproduce

```python
import tempfile

import torch
from torchrl.data import LazyTensorStorage, TensorDictMaxValueWriter, TensorDictReplayBuffer


def make_rb():
    return TensorDictReplayBuffer(
        storage=LazyTensorStorage(4),
        writer=TensorDictMaxValueWriter(rank_key="reward"),
    )


rb = make_rb()
rb.extend(TensorDict(reward=torch.tensor([1.0, 5.0, 3.0]), obs=torch.arange(3), batch_size=[3]))

with tempfile.TemporaryDirectory() as path:
    rb.dumps(path)
    restored = make_rb()
    restored.loads(path)

new_data = TensorDict(reward=torch.tensor([2.0]), obs=torch.tensor([3]), batch_size=[1])
rb.extend(new_data.clone())
print(sorted(rb[:]["obs"].tolist()))
restored.extend(new_data.clone())
```

```bash
[0, 1, 2, 3]
  File "/tmp/repro/repro.py", line 26, in <module>
    restored.extend(new_data.clone())
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/tmp/repro/torchrl/data/replay_buffers/replay_buffers/base.py", line 106, in wrapper
    return func(self, *args, **kwargs)
  File "/tmp/repro/torchrl/data/replay_buffers/replay_buffers/tensordict.py", line 317, in extend
    index = super()._extend(tensordicts)
  File "/tmp/repro/torchrl/data/replay_buffers/replay_buffers/base.py", line 2194, in _extend
    index = self._writer.extend(data)
  File "/tmp/repro/torchrl/data/replay_buffers/writers/max_value.py", line 264, in extend
    storage_idx = self.get_insert_index(sample)
  File "/tmp/repro/torchrl/data/replay_buffers/writers/max_value.py", line 209, in get_insert_index
    heapq.heappush(self._current_top_values, (rank_data, ret))
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'tuple' and 'list'
```

With a full buffer the error is the same, raised from the `heapq.heappush` a few lines further down in `get_insert_index`.

## Expected behavior

`restored.extend(...)` behaves like `rb.extend(...)`: the item goes into the free slot, and once the buffer is full, a higher-ranked item replaces the lowest-ranked entry.

## System info

* TorchRL source checkout at `aa442e4b204adf9ea532b981421b4f4c5ed17573`
* Linux, Python 3.13.15
* PyTorch `2.11.0`, NumPy `2.5.2`

## Reason and Possible fixes

The writer keeps its current top values as a `heapq` heap of `(rank, storage_index)` tuples (see `get_insert_index` in `torchrl/data/replay_buffers/writers/max_value.py`). `dumps` saves the heap as a tensor, and `loads` turns it back into Python objects with `tolist()`:

```python
self._current_top_values = MemoryMappedTensor.from_filename(
    filename=path / "current_top_values.memmap",
    dtype=_STRDTYPE2DTYPE[dtype],
    shape=shape,
).tolist()
```

`tolist()` returns `[rank, index]` lists with a float index, for example `[[1.0, 0.0], [5.0, 1.0], [3.0, 2.0]]` for the buffer above. The next write pushes a `(rank, index)` tuple onto that heap, and comparing a tuple with a list raises. `test_max_value_writer_serialize` compares `torch.tensor(...)` of both heaps and never writes after `loads`, which is why it passes. The loading code goes back to #1733.

Rebuilding the entries as `(rank, int(index))` tuples in `loads` fixes it without changing the file format, so existing checkpoints load the same way. I have a patch with a regression test and can open a PR.

Something related I noticed while looking at this: `dumps` builds the tensor with `torch.as_tensor(self._current_top_values)`, which gives float32, so storage indices above `2**24` do not come back exactly (`16777217` is read back as `16777216.0`). That only affects very large buffers and I did not include it in the patch, but saving with `dtype=torch.float64` would avoid it.

## Checklist

yupengtang:
## Description

`TensorDictMaxValueWriter.loads` now restores its heap entries as `(rank, index)` tuples.

`loads` rebuilt the heap from the saved tensor with `tolist()`, which returns `[rank, index]` lists with a float index. The heap is maintained with `heapq` and new entries are `(rank, index)` tuples, so the first `add` or `extend` after loading compared a tuple with a list and raised `TypeError`. The entries are now rebuilt as `(rank, int(index))` tuples. The saved files do not change, so existing checkpoints load as before.

## Motivation and Context

close #4369

`test_max_value_writer_loads_then_writes` saves a partly filled buffer with `dumps`, loads it into a new buffer and writes the same data to both: an `extend` into the free slot, an `add` that replaces the lowest ranked entry, and an `extend` where one item is rejected and the other replaces an entry. Both buffers have to end up holding the entries worked out by hand. The test fails on `main` with the `TypeError` and passes with this change.

* `pytest test/rb` on CPU: 3221 passed, 1139 failed and 1015 skipped, against 3220 passed, 1140 failed and 1015 skipped on `main`. Apart from the new test, the same tests fail on both sides, all because the `_torchrl` C++ extension (segment trees for the prioritized samplers) is not built in my environment.

## Types of changes

## Checklist


test/rb/test_writers.py
```diff
@@ -146,6 +146,32 @@ def test_max_value_writer_serialize(
...
+            )
+
+        rb = make_rb()
+        rb.extend(data([1.0, 5.0, 3.0], [0, 1, 2]))
+        rb.dumps(tmpdir)
+        restored = make_rb()
+        restored.loads(tmpdir)
+        # the restored buffer keeps ranking new data against the loaded entries,
+        # first into its free slot, then by replacing the lowest ranked entry
+        for buffer in (rb, restored):
```
vmoens, line 169: Nonblocking test simplification: the `rb` iteration repeats ordinary ranking behavior already covered by `test_max_value_writer`. Since the expected contents are explicitly worked out below, running this sequence only on `restored` would retain the checkpoint regression coverage and make the test more focused.

vmoens (approved):
The fix restores the tuple entries required by heapq and the integer indices required by subsequent storage writes, while preserving the checkpoint format. No blocking findings.

Validated on the PR head: `python -m pytest test/rb/test_writers.py::TestMaxValueWriter -q` passed all 34 cases on CPU. I also ran the new regression test with the original `loads` method substituted in memory; it fails at the first post-load write with the reported tuple/list TypeError.

The full GitHub CI workflows are still awaiting approval; this review is based on source inspection and focused local validation.

PR #4372, opened 2026-09-14
[BugFix] Allow CrossQLoss.set_keys once a value estimator exists

### LINKED ISSUE #4371: [BUG] CrossQLoss.set_keys raises AttributeError once a value estimator exists
## Describe the bug

Once a `CrossQLoss` has a value estimator, every call to `set_keys` raises `AttributeError`. That covers `make_value_estimator(...)` followed by `set_keys(...)`, which is the order `sota-implementations/crossq` uses, and also a loss whose default estimator was created lazily by a first forward pass. Calling `set_keys` before any estimator exists works, which is why the existing `test_crossq_tensordict_keys` passes. `SACLoss` handles the same sequence fine.

## To Reproduce

```python
import torch
from tensordict.nn import NormalParamExtractor, TensorDictModule
from torch import nn
from torchrl.data import Bounded
from torchrl.modules import ProbabilisticActor, TanhNormal, ValueOperator
from torchrl.objectives import CrossQLoss

n_obs, n_act = 3, 4


class QValue(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(n_obs + n_act, 1)

    def forward(self, observation, action):
        return self.linear(torch.cat([observation, action], -1))


actor = ProbabilisticActor(
    TensorDictModule(
        nn.Sequential(nn.Linear(n_obs, 2 * n_act), NormalParamExtractor()),
        in_keys=["observation"],
        out_keys=["loc", "scale"],
    ),
    in_keys=["loc", "scale"],
    spec=Bounded(-torch.ones(n_act), torch.ones(n_act), (n_act,)),
    distribution_class=TanhNormal,
)
qvalue = ValueOperator(QValue(), in_keys=["observation", "action"])

loss = CrossQLoss(actor, qvalue)
loss.make_value_estimator(gamma=0.99)
loss.set_keys(reward="my_reward")
```

```bash
  File "/tmp/repro_rl/torchrl/objectives/common.py", line 400, in set_keys
    self._forward_value_estimator_keys(**kwargs)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/tmp/repro_rl/torchrl/objectives/crossq.py", line 462, in _forward_value_estimator_keys
    value=self.tensor_keys.value,
          ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: '_AcceptedKeys' object has no attribute 'value'

The above exception was the direct cause of the following exception:

  File "/tmp/repro_rl/repro.py", line 34, in <module>
    loss.set_keys(reward="my_reward")
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/tmp/repro_rl/torchrl/objectives/crossq.py", line 452, in set_keys
    out = super().set_keys(**kwargs)
  File "/tmp/repro_rl/torchrl/objectives/common.py", line 402, in set_keys
    raise AttributeError(
    ...<4 lines>...
    ) from err
AttributeError: To utilize `.set_keys(...)` for tensordict key configuration, the subclassed loss module must define an _AcceptedKeys dataclass containing all keys intended for configuration. If the default `._forward_value_estimator_keys()` implementation is insufficient, the subclass must override it to forward modified tensordict keys to the underlying value_estimator.
```

## Expected behavior

`set_keys` updates the keys and forwards `reward`, `done` and `terminated` to the value estimator, the same as for `SACLoss`.

## System info

* TorchRL source checkout at `c65e77636578d7c788f14bd51471a13dacfd798d`
* Linux, Python 3.13.15
* PyTorch `2.11.0`

## Reason and Possible fixes

`CrossQLoss._forward_value_estimator_keys` passes `value=self.tensor_keys.value`, but CrossQ's `_AcceptedKeys` has no `value` field (it has `state_action_value`). The line is only reached when `self._value_estimator` is set, and `LossModule.set_keys` turns the resulting `AttributeError` into the generic message about defining `_AcceptedKeys`, which hides the actual cause.

CrossQ never reads the estimator's `value` key: `qvalue_loss` computes the next state-action value itself and calls `self.value_estimator.value_estimate(tensordict, next_value=...)`, and `make_value_estimator` only passes `reward`, `done` and `terminated`. So dropping `value=` from `_forward_value_estimator_keys` fixes it without changing any loss value. I checked that with the change, the losses after `make_value_estimator` then `set_keys(reward=..., done=..., terminated=...)` on data using the new keys are identical to the losses with the default keys. I can open a PR with a regression test in `test/objectives/test_sac.py`.

A quick check across the other losses found no other `_forward_value_estimator_keys` that reads a key missing from its `_AcceptedKeys`.

## Checklist

yupengtang:
## Description

`CrossQLoss.set_keys` no longer fails once the loss has a value estimator.

`_forward_value_estimator_keys` passed `value=self.tensor_keys.value` to the estimator, but CrossQ's `_AcceptedKeys` has no `value` field. That line only runs when `self._value_estimator` is set, so `set_keys` raised `AttributeError` after `make_value_estimator(...)` (the order `sota-implementations/crossq` uses) or after the default estimator had been created by a forward pass, and `LossModule.set_keys` reported it as a missing `_AcceptedKeys` definition.

CrossQ never reads the estimator's `value` key: `qvalue_loss` computes the next state-action value and passes it to `value_estimate(tensordict, next_value=...)`, and `make_value_estimator` already forwards only `reward`, `done` and `terminated`. The forward now sends the same three keys.

## Motivation and Context

close #4371

`test_crossq_set_keys_after_value_estimator` builds a loss, calls `make_value_estimator` with each supported estimator, then renames `reward`, `done` and `terminated` on a copy of it. The copy runs on data that only has the renamed keys, and its losses must match the original loss on the default keys. It fails on `main` at `set_keys` with the `AttributeError` and passes with this change.

* `pytest test/objectives/test_sac.py -k TestCrossQ` on CPU: 123 passed, against 120 passed and the 3 new cases failing on `main`. `pytest test/objectives/test_loss_module.py`: 204 passed and 1 skipped on both.

## Types of changes

## Checklist


vmoens (approved):
No actionable findings at 2ff9764afca6f5261a76dff67131269fb26c3f75.

Removing the nonexistent `value` key is correct: all three supported estimators use the explicitly supplied `next_value`, and forwarding now matches `make_value_estimator`. The regression test exercises renaming after estimator creation and verifies unchanged loss outputs.

Validated locally on CPU with Python 3.12 and PyTorch 2.11.0:
- CrossQ tests: 123 passed.
- LossModule tests: 204 passed, 1 skipped (CUDA unavailable).
- Substituting the base commit's `_forward_value_estimator_keys` in memory makes all three new regression cases reproduce the original missing-`value` AttributeError.

CI is still running.

PR #4374, opened 2026-09-14
[BugFix] Count CSVLogger text steps with text_counter

### LINKED ISSUE #4373: [BUG] CSVLogger numbers text files with the video counter
## Describe the bug

`CSVExperiment.add_text` numbers its files with `videos_counter` instead of `text_counter`. Texts and videos logged under the same tag therefore share one step counter, so a text logged after two videos is written as `<tag>2.txt` rather than `<tag>0.txt`, and `text_counter` stays empty. Since #3967 the logger checkpoints `text_counter` separately from `videos_counter`, so the saved text counts end up in the video counter and the text counter in the checkpoint is always `{}`.

## To Reproduce

```python
import os
import tempfile

import torch
from torchrl.record.loggers.csv import CSVLogger

with tempfile.TemporaryDirectory() as log_dir:
    logger = CSVLogger("exp", log_dir=log_dir)
    video = torch.zeros(1, 2, 3, 4, 4, dtype=torch.uint8)
    logger.log_video("eval", video)
    logger.log_video("eval", video)
    logger.experiment.add_text("eval", "first note")
    logger.log_hparams({"lr": 0.1})

    print(sorted(os.listdir(os.path.join(log_dir, "exp", "videos"))))
    print(sorted(os.listdir(os.path.join(log_dir, "exp", "texts"))))
    state = logger.state_dict()["local"]
    print("videos_counter:", dict(state["videos_counter"]))
    print("text_counter:", dict(state["text_counter"]))
```

```bash
['eval_0.pt', 'eval_1.pt']
['eval2.txt', 'hparams0.txt']
videos_counter: {'eval': 3, 'hparams': 1}
text_counter: {}
```

## Expected behavior

```
['eval_0.pt', 'eval_1.pt']
['eval0.txt', 'hparams0.txt']
videos_counter: {'eval': 2}
text_counter: {'eval': 1, 'hparams': 1}
```

## System info

* TorchRL source checkout at `c65e77636578d7c788f14bd51471a13dacfd798d`
* Linux, Python 3.13.15
* PyTorch `2.11.0`

## Reason and Possible fixes

In `torchrl/record/loggers/csv.py`, `add_text` was copied from `add_video` and kept its counter:

```python
def add_text(self, tag, text, global_step: int | None = None):
    if global_step is None:
        global_step = self.videos_counter[tag]
        self.videos_counter[tag] += 1
```

`text_counter` is created in `__init__` and saved and restored by `CSVLogger._checkpoint_state` / `_load_checkpoint_state`, but nothing ever increments it. Using `self.text_counter` in `add_text` fixes the numbering and makes the checkpointed counters mean what their names say.

One thing to be aware of: a checkpoint written before such a change stores the text counts under `videos_counter`, so a logger resumed from it would start numbering texts from 0 again, and `add_text` opens files with `"w+"`, which overwrites an existing `hparams0.txt`. I can open a PR with a regression test in `test/test_loggers.py`.

## Checklist

yupengtang:
## Description

`CSVExperiment.add_text` now numbers text files with `text_counter`.

It used `videos_counter`, so texts and videos logged under the same tag shared one step counter (a text logged after two videos was written as `<tag>2.txt`), and `text_counter`, which `CSVLogger` checkpoints since #3967, was never incremented.

A checkpoint written before this change stores its text counts under `videos_counter`, so a logger resumed from such a checkpoint starts numbering texts from 0 again.

## Motivation and Context

close #4373

`test_text_and_video_steps_are_counted_separately` logs two videos and a text under the same tag plus a `log_hparams` call, and checks the file names (`demo_0.pt`, `demo_1.pt`, `demo0.txt`, `hparams0.txt`). It then resumes a second logger from `state_dict()` and checks that the next `log_hparams` writes `hparams1.txt` without touching `hparams0.txt`. It fails on `main` with `['demo2.txt', 'hparams0.txt'] == ['demo0.txt', 'hparams0.txt']` and passes with this change.

* `pytest test/test_loggers.py -k "csv or CSV"` on CPU: 17 passed, against 16 passed and the new test failing on `main`.

## Types of changes

## Checklist


torchrl/record/loggers/csv.py
```diff
@@ -107,8 +107,8 @@ def add_video(self, tag, vid_tensor, global_step: int | None = None, **kwargs):
 
     def add_text(self, tag, text, global_step: int | None = None):
         if global_step is None:
-            global_step = self.videos_counter[tag]
-            self.videos_counter[tag] += 1
+            global_step = self.text_counter[tag]
+            self.text_counter[tag] += 1
```
vmoens, line 111: [P2] Preserve text logs when resuming older checkpoints

A checkpoint saved by the existing implementation after `log_hparams({"lr": 0.1})` contains `videos_counter={"hparams": 1}` and `text_counter={}`. After loading that checkpoint on this version, the next `log_hparams` chooses step 0 again, and the `open(filepath, "w+")` below silently truncates the existing `hparams0.txt`. I reproduced this through `get_logger(..., state_dict=state)`; resuming the same checkpoint on the base version preserves the original file. Please handle the legacy counter state, or otherwise prevent automatic text numbering from overwriting existing files, and cover an old-to-new checkpoint resume. The new test only restores checkpoints created with the fixed implementation, so it misses this data-loss regression.

vmoens (changes requested):
The counter separation works for new runs, but resuming a checkpoint created before this change can silently overwrite existing text logs. Please address the compatibility issue below.

Validation: `pytest test/test_loggers.py -k "csv or CSV" -q` completed with 15 passed and 2 skipped (Ray unavailable). A separate reproduction using the base implementation to create the checkpoint confirmed that base-to-base resume preserves the original hyperparameters file, whereas base-to-PR resume overwrites it.

torchrl/record/loggers/csv.py
yupengtang: Thanks for catching this. Fixed in 0645c279c: when the loaded `text_counter` is empty, which is always the case for checkpoints saved before this PR, `_load_checkpoint_state` seeds it from `videos_counter`. Text numbering then continues after the steps the old shared counter handed out, so nothing gets truncated.

Added `test_resume_checkpoint_saved_without_text_counter`, which builds that old checkpoint shape from `state_dict()` and resumes it through `get_logger(..., state_dict=state)`. It fails on the previous head (`hparams0.txt` ends up as `lr: 0.2`) and passes now; `pytest test/test_loggers.py -k "csv or CSV"` gives 18 passed. I also rebased on main.

One side effect: a new checkpoint that has logged videos but no text yet takes the same path, so a text logged later under the same tag as a video starts numbering after that video count. It never overwrites a file, so I did not add a version marker to the checkpoint format for it, but I can if you prefer.

vmoens (approved):

PR #4386, opened 2026-09-16
[Feature] MenagerieEnv: any MuJoCo Menagerie robot by name

theap06:
## Description

MuJoCo Menagerie ships curated robot models, not tasks. `MujocoEnv` can already load any MJCF with `patch_xml=False`, but using a Menagerie robot meant subclassing to implement `_compute_reward` / `_compute_done` and locating the assets by hand (`CubeBowlEnv` does this for UR5e + Robotiq). This PR adds `MenagerieEnv`, which loads any Menagerie robot by name on the existing backends, and a `MenagerieTask` config.

- Model resolution: `menagerie_path` (a checkout, the robot directory or the XML) > `TORCHRL_MUJOCO_MENAGERIE_PATH` (what CI already uses for the cube-bowl tutorial) > the official `mujoco-menagerie` package cache, which `download=True` fills. The package is an optional dependency; it pins every robot to one Menagerie commit and verifies checksums. A metaclass resolves the XML once before the native backend spawns workers.
- Observation: `qpos`, `qvel`, `sensordata` when the model defines sensors, and `site_positions` for `task.site_names`. Adds a `sensordata` accessor to the three physics backends.
- Reset around the model's `home` keyframe (`qpos` and `qvel`), falling back to `qpos0`; `task.keyframe` picks another one.
- Termination on non-finite state, plus an optional `terminate_below_height` for floating bases.
- Reward is zero by default so a transform can supply the task reward (the docstring shows a reach reward on a UR5e site). `MenagerieTask` weights turn on a Gaussian pose term, a normalized control cost and an alive bonus; `MenagerieEnv.hold_pose_task()` is the named preset.
- Default backend is `"mujoco"`: Menagerie's standard `scene.xml` files use collision pairs mujoco-torch does not implement (plane vs cylinder) and the Panda hits a shape mismatch there. The entries Menagerie maintains for MJX (`scene_mjx`, `mjx_scene`) run on `mujoco-torch` and `mjx`; only the accelerometer sensor differs from MuJoCo. Documented in the class docstring.
- `CubeBowlEnv.MENAGERIE_ENV_VAR` reuses the shared constant.

Verified locally on real assets: UR5e on the C bindings, go2 `scene_mjx` on mujoco-torch (sensors and sites in the observation), `SerialEnv` / `ParallelEnv` batching, a reward transform on the flange site, free-camera pixels with `camera_id=-1`, and the package download path.

Noticed but not changed here: `_TorchBackend._init_model` runs one physics step to derive dtypes, so its `qpos0` / `qvel0` carry one step of gravity and torch-backend envs reset with a small nonzero velocity. `MenagerieEnv` shifts the reset by the keyframe state, so it is exact; the sibling envs still inherit it.

## Tests

- `test/libs/test_mujoco.py`: six tests on a synthetic one-robot checkout (specs, keyframe reset, sensors and sites on every backend; keyframe selection, the hold-pose reward formula and fall termination; checkout resolution and worker dispatch; package resolution without network, skipped when the package is missing), plus a UR5e test that runs where `TORCHRL_MUJOCO_MENAGERIE_PATH` is set.
- Full `test/libs/test_mujoco.py` locally: 110 passed, 1 CUDA skip. Doctests and pre-commit clean.

## Docs

- `docs/source/reference/envs_api.rst`: paragraph in the MuJoCo custom environments section and autosummary entries for `MenagerieEnv` and `MenagerieTask`.


vmoens (commented):
This is a useful addition. The resolution order is clear, keeping network access behind `download=True` is a good default, and the synthetic model exercises meaningful behavior across the three backends.

A couple of things need addressing before merge:

- The new `menagerie-ppo` example currently fails in CI because constructing `Collector` emits the `InitTracker` auto-registration `FutureWarning`, and the examples job treats that warning as an error. Please either pass `auto_register_policy_transforms=True` or install the required transform explicitly.
- `test_menagerie_package_resolution` currently skips when `mujoco-menagerie` is absent, but the dedicated MuJoCo job rejects skipped tests, so that job fails at test setup. The package integration needs to run in a job that installs the optional dependency, or the dependency needs to be included in this job. Since this changes an optional-dependency integration, please also add the `ci/optdeps` label before rerunning the relevant workflow.

Minor style point: please omit the explicit `-> None` annotations on the newly added `__init__` and `__post_init__` methods.

Non-blocking suggestion for a follow-up PR: it would be useful to generalize the model-resolution layer to robot models hosted in arbitrary third-party GitHub repositories, for example [SouthColumn76/universal_robots_ur3e](https://github.com/SouthColumn76/universal_robots_ur3e). A `download=True` argument could remain the explicit permission for network access, but I would make the source include `repo`, a pinned `revision`, and an explicit repository-relative `entry` at first. Repository layouts vary substantially: XML can be at the root or several directories deep, assets may be in sibling directories, and one repository may contain several valid scenes. The downloader should therefore preserve the full repository tree, cache by owner/repository/resolved commit, and avoid silently choosing the first `scene.xml`. An optional manifest and conservative discovery of a unique compile-valid root could come later. This probably belongs in a generic `GitHubModelSource` / `MujocoModelEnv` resolver, with `MenagerieEnv` remaining a convenience wrapper around the curated Menagerie source.

PR #4392, opened 2026-09-16
[Feature] Add the MicroDuck skill environment

vmoens:
## Summary

Adds the MicroDuck-specific environment layer on top of the generic controller in #4329 and the skill artifact in #4394.

- `MicroDuckSkillController` maps a high-level skill id (and optional command argument) to the task-conditioned policy inputs and joint targets.
- `MicroDuckSkillEnv.from_env(...)` explicitly promotes joint-level task dynamics into a skill-level MDP.
- The high-level environment owns the frozen skill policy as part of its transition dynamics, aggregates physical-step rewards, and resets recurrent state independently for ducks that fall.
- The names `skill_policy`, `task_library`, `skill_ids`, and `control_steps_per_decision` make the two policy levels and the temporal boundary explicit.
- Related TorchRL classes now show this API in their examples and link to one another through `.. seealso::` blocks.

## What was wrong and what is fixed

The previous branch exposed a free helper that accepted a neural policy directly. Its short signature concealed that it changed the action space, observations, reset behavior, temporal resolution, and reward aggregation:

```python
# Before: a free function silently converted joint-level dynamics into a
# different MDP, and "walker" suggested one gait rather than many skills.
env = microduck_skill_env(
    base_env, walker, tasks, skills=[0, 2], steps=5
)

# After: the artifact keeps policy metadata together and the environment class
# makes the MDP boundary explicit.
skills = MicroDuckSkills.from_pretrained()
env = MicroDuckSkillEnv.from_env(
    base_env,
    skills,
    skill_ids=[0, 2],
    control_steps_per_decision=5,
)
rollout = env.rollout(10, high_level_policy)
```

The superseded names existed only on this unmerged branch, so no released API is removed.

## Tests

- The real fixture-backed `MicroDuckEnv` deployment regression checks public spec/action plumbing, finite rewards, augmented high-level observations, and recurrent state.
- Synthetic behavior tests cover discrete and parameterized commands and custom per-agent reset signals.
- Generic `ClosedLoopMultiAction` tests continue to pass.
- `MicroDuckSkillController` and `MicroDuckSkillEnv` doctests pass.


vmoens:
CI triage: the two failed GPU jobs on the previous head are unrelated to this PR. Both fail in `test/test_inference_server.py::TestProcessSlotTransport::test_server_batched_pass_on_cuda[True]` with PyTorch/CUDA reporting `cudaStreamCaptureStatusInvalidated` / `cudaErrorStreamCaptureInvalidated`. No MicroDuck test is in either failure path. The PR has since been updated for the separate µfmt failure, which was caused by this stack.

PR #4397, opened 2026-09-16
[Feature] Train MicroDuck skills and navigation

vmoens:
## Summary

- add the focused MicroDuck PPO training launcher
- support task remapping and optional critic restoration when loading checkpoints
- cover checkpoint restoration across navigation, skill, and critic parameters
- exclude the long-running launcher from generic example execution and smoke-test it in the native MuJoCo tutorial environment

This PR is intentionally limited to the training entry point. Its temporary synthetic base collects the focused dependencies from #4329, #4388, #4389, #4390, #4391, #4392, #4393, #4394, #4395, #4396, and #4398. It will be rebased onto main as those land.

## Test plan

- launcher checkpoint/load/remap tests: 3 passed
- full two-stage launcher smoke run: passed
- python -m examples.microduck.train_skills --help: passed

PR #4399, opened 2026-09-16
[Doc] Add the MicroDuck skills tutorial

vmoens:
## Summary

- add a focused tutorial for training a multi-skill MicroDuck policy with PPO
- show how to restore the published walker and compose it into waypoint navigation
- register the tutorial and its Hugging Face dependency in the docs and tutorial CI environments

This is intentionally separate from the environment walkthrough in #4398 and the training launcher in #4397. It is temporarily stacked on the launcher because the executable tutorial exercises that public workflow; it will be rebased as the dependency stack lands.

## Test plan

- MicroDuck skills tutorial executed end to end in fast mode with native macOS MuJoCo: 1 passed


vmoens:
Closing because this tutorial depends on #4397 and documents its obsolete synchronous 10M-frame workflow. It also inherits the launcher failures. A replacement tutorial should follow the new training implementation once that API exists. The branch is retained.

PR #4400, opened 2026-09-16
[Refactor] Consolidate stats() in BaseCollector and add WandB monitoring example

coder-jayp:
Consolidates single-process and multiprocessing collector statistics in `BaseCollector.stats()`, preserving shared in-flight trajectory counters, non-blocking aggregate snapshots, and the existing per-worker control-channel behavior. Workers return data to the coordinator, where logging happens.

The public `stats()` API stays the same. The before/after below shows the regression-test scenario: one healthy Ray worker is held inside its first collection call for longer than the stats timeout.

**Before:** the stats request waits behind collection on the same actor queue, times out, and the healthy worker is omitted from the snapshot.

```python
snapshot = collector.stats(workers="both", timeout=1.0)
snapshot["workers_alive"]  # 0, even though the worker is collecting
"worker_0/frames" in snapshot  # False: no worker snapshot arrived
```

**After:** the stats request runs in its own concurrency group and can report progress while collection is still busy.

```python
snapshot = collector.stats(workers="both", timeout=1.0)
snapshot["workers_alive"]  # 1: the worker answered during collection
snapshot["worker_0/frames"]  # 0: its first batch has not finished yet
```

The new example uses this snapshot API through `LoggerMonitor` to send aggregate and per-worker progress to W&B from the coordinator:

```python
with LoggerMonitor(logger, poll_interval=0.5) as monitor:
    monitor.watch(
        collector,
        name="ray_collector",
        schedule=Every.counter("frames", 100),
        stats_kwargs={"workers": "both"},
    )
    for batch in collector:
        ...  # Training or other batch processing; monitoring runs separately.
```

Collection and state-changing calls retain the default actor queue. Remote polling remains timeout-bounded, and caller-supplied actor options and concurrency groups are preserved.

Validation:

- All 11 `TestCollectorStats` cases pass, including two real-Ray tests with Ray 2.54.0.
- The new regression holds collection behind a gate, verifies stats respond while the gate is closed, then releases it and verifies the collected batch. Restoring the old stats dispatch makes this test fail because `workers_alive` becomes zero.
- The exact example completes with offline W&B. A separate manual monitor poll verifies persisted aggregate and per-worker frame counters from two live Ray workers.
- Flake8, ufmt, and `git diff --check` pass.
- The example is registered as runnable CI coverage with offline W&B and an isolated log directory. The manifest/sharding checks and actual Ray/W&B example pass through the CI runner (12 passed).


examples/ray_wandb_monitor.py
```diff
@@ -0,0 +1,50 @@
...
+        [env_maker, env_maker],
+        policy=RandomPolicy(action_spec=GymEnv("Pendulum-v1").action_spec),
+        frames_per_batch=100,
+        total_frames=1000,
+    )
+
+    # Monitor the collector and report per-worker stats
+    # Setting workers="both" requests both the aggregate and the per-worker metrics
+    with LoggerMonitor(logger, poll_interval=0.5) as monitor:
+        monitor.watch(
```
theap06, line 31: I ran this locally and it raises an error each run. If you set the stat_kwargs to workers: both, it should work correctly.

torchrl/collectors/_base.py
```diff
@@ -442,27 +452,84 @@ def stats(self) -> dict[str, int | float | bool]:
...
+                num_workers = len(remote_collectors)
+            stats["workers"] = int(num_workers)
+
+            try:
+                # If we need worker_frames for aggregate RayCollector, we have to map_fn.
+                # To minimize overhead for MultiSyncCollector (which didn't do this),
+                # we only do it if workers in ("per_worker", "both") OR if we are a RayCollector.
+                worker_results = []
+                if workers in ("per_worker", "both") or hasattr(
+                    self, "remote_collectors"
```
theap06, line 503: the ray actors lose on the time out as the actors are single-threaded. I think creating a new execution queue on the actor is the right approach.

theap06 (changes requested):
left some comments. overall design lgtm.

PR #4401, opened 2026-09-16
[Feature] MujocoModelEnv with GitHub and Menagerie model sources

theap06:
## Description

Follow-up to #4386, stacked on it: the base branch is `menagerie-env` in this repository (a mirror of #4386's head), so the diff is this PR's own delta. Retarget to `main` once #4386 merges. It implements the suggestion from that review: a model-resolution layer that is not tied to Menagerie.

- `MujocoModelEnv` is the bare-simulator env of #4386 (`qpos`, `qvel`, `sensordata` and `site_positions` observations, keyframe reset, `MujocoModelTask` config with the hold-pose terms) taking a model source: the path of an XML, or any object with `resolve(download=...)` (the `ModelSource` protocol). `download=True` remains the explicit permission for network access, and the metaclass resolves the source once before the native backend spawns workers.
- `GitHubModelSource(repo, revision, entry, root=None)` names `owner/name`, a pinned revision (commit SHA, tag or branch) and an explicit repository-relative entry, so nothing is guessed. The whole repository tree at that revision is fetched as a GitHub zip archive and cached under `~/.cache/torchrl/github_models/owner/name/<commit>` with an atomic move, so includes and sibling asset directories resolve as in a checkout. A tag or branch is resolved to its commit through the GitHub API (`GITHUB_TOKEN` is used when set) on the first `download=True`, and the mapping is cached under `refs/`, so later resolutions are offline and stay pinned until the cache entry is removed. A missing entry lists the XML files the tree contains instead of picking one.
- `MenagerieModelSource(robot, entry=None, menagerie_path=None)` is the curated source (checkout, then `TORCHRL_MUJOCO_MENAGERIE_PATH`, then the `mujoco-menagerie` package cache), extracted from `MenagerieEnv.resolve_model`, which now delegates to it.
- `MenagerieEnv` becomes a thin subclass of `MujocoModelEnv` over that source. Its constructor, attributes and behaviour are unchanged; `MenagerieTask` stays as an alias of `MujocoModelTask`.
- The pre-resolution step lives in `_MujocoMeta` itself: after validating the backend and batching arguments and before building any env or worker, it calls the class' `_resolve_before_batching` hook (identity by default). That removes the per-subclass metaclasses (`_MujocoModelMeta`, and the existing `_MicroDuckMeta`), makes argument errors precede any download, and lets MicroDuck's downloader reuse the shared GitHub tree fetcher. The backend alias `ModelSource = str | Path` in `_backends.py` is renamed `XmlSource` so the public protocol keeps the name.

- `examples/menagerie/ppo.py` takes `--repo`, `--revision` and `--entry`, so the same PPO recipe and `rlrender` factories work on any MuJoCo model on GitHub through `MujocoModelEnv` and `GitHubModelSource`.

Left for later, as suggested: a manifest format and conservative discovery of a unique compile-valid root.

Verified on the repository cited in the review: `MujocoModelEnv(GitHubModelSource("SouthColumn76/universal_robots_ur3e", revision="5f042ffca6b5885fd18f5448e17b71ab46274fa3", entry="ur3e.xml"), download=True)` downloads the tree in about two seconds, resolves `main` to that commit and caches it, passes `check_env_specs`, resets to the `home` keyframe and rolls out. A wrong entry reports the tree's XML files.

## Tests

- `test_github_model_source` mocks the archive download and the API call, and asserts the cache layout by owner, repository and commit, that nothing is fetched without `download=True`, pinned offline resolution afterwards, no API call for a full SHA, the entry error, argument validation, and one download for batched native workers.
- `test_mujoco_model_env_from_path`; the existing Menagerie tests pass unchanged apart from one error message.
- Doctests of the three modules.

## Docs

- `envs_api.rst`: paragraph and autosummary entries for `MujocoModelEnv`, `MujocoModelTask`, `GitHubModelSource`, `MenagerieModelSource` and `ModelSource`.


vmoens:
> the diff includes its commits until it merges


Now that you're maintainer you can push to PyTorch (and not your fork) to make "real" gh PR stacks!

PR #4402, opened 2026-09-17
[Quality] Generalize target-updater typing and documentation

vmoens:
## Summary

- type `TargetNetUpdater`, `SoftUpdate`, and `HardUpdate` against the generic `LossModule` contract they already implement
- document target/source discovery for delayed actors, values, and other target parameters
- add a concise PPO-EWMA example and make the missing-delay diagnostic cover delayed actors
- strengthen the PPO-EWMA regression to verify the exact soft-update interpolation of proximal actor parameters

## Testing

- `PYTHONPATH=. uv run --no-project --python /Users/vmoens/repos/rl/.venv/bin/python --with transformers python -m pytest test/objectives/test_ppo.py -k "decoupled_proximal_policy and not kl_pen" -q`
- `PYTHONPATH=. /Users/vmoens/repos/rl/.venv/bin/python -m pytest test/objectives/test_loss_module.py -k updater -q`
- `uvx --from pre-commit pre-commit run --files torchrl/objectives/utils.py test/objectives/test_ppo.py` (all available hooks passed; the local system hook lacked the `autoflake` executable)
- `uvx autoflake --check torchrl/objectives/utils.py test/objectives/test_ppo.py`
## Before and after

```python
loss = ClipPPOLoss(actor, critic, delay_actor=True)

# Before: the updater worked at runtime, but its public annotation admitted only
# a hard-coded list of DQN/DDPG/SAC/REDQ/TD3 losses.
updater = SoftUpdate(loss, eps=0.5)
# type checker: incompatible argument type "ClipPPOLoss"

# After: every LossModule with matching source/target parameter children is
# accepted. Here an optimizer step has already moved the live actor.
source = loss.actor_network_params.clone()
target = loss.target_actor_network_params.clone()
updater = SoftUpdate(loss, eps=0.5)
updater.step()
for key, actual in loss.target_actor_network_params.items(True, True):
    torch.testing.assert_close(
        actual,
        target.get(key).lerp(source.get(key), 0.5),
    )
```


vmoens (commented):
**Review (COMMENT)** — this is my own PR, so GitHub only lets me leave a comment review.

Overall: clean, strictly-better change. The `LossModule`-based typing now matches the runtime contract (`isinstance` check in `__init__` plus duck-typed discovery of `target_*_params` children), and the revised missing-delay diagnostic no longer sends `ClipPPOLoss(delay_actor=True)` users chasing `delay_value`.

Two small nits, non-blocking:

1. `torchrl/objectives/utils.py` — the new `SoftUpdate` example uses `eps=0.889` with no explanation of where that number comes from. Either a rounder illustrative value (e.g. `0.5`, mirroring the strengthened test) or a one-line note on the provenance of 0.889 would read better.

2. Same example: the prose says the updater runs "after each optimizer step", but the snippet itself doesn't show that ordering. One more line (`# after optimizer.step()`) would make the recipe self-contained; fine as-is since it's `+SKIP` anyway.

The strengthened PPO-EWMA regression (exact `lerp(source, 0.5)` check key-by-key, then verifying the loss tracks the moved proximal policy) is exactly the right assertion — it pins both *which* keys move and *by how much*.

LGTM modulo the nits.

PR #4403, opened 2026-09-17
[Feature] Add namespaced logger views

vmoens:
## Summary

- add composable `Logger.with_prefix()` views for scalar batches, videos, histograms, and strings
- preserve hyperparameter keys and delegate lifecycle, checkpoint state, and experiment ownership to the underlying logger
- make views over owning loggers satisfy existing `Logger` validation while keeping service-client views restricted
- carry namespaces through process logger clients and preserve independent W&B step metrics per namespace
- document the public `PrefixLogger` API

## Test plan

- `pytest test/test_loggers.py -k "prefix_logger or prefixed_views or owner_clients_fifo_video_and_errors" -q`
- `pytest test/test_loggers.py -q -k "not test_log_video"` (53 passed, 29 skipped)
- flake8, pydocstyle, ufmt, and diff checks on changed files

Video tests are excluded locally because the optional MoviePy/imageio-ffmpeg dependency stack is unavailable or rejects one-channel GIF frames; they are unrelated to this change.

## Before and after

```python
# Before: every call site had to spell its own namespace. One missed prefix
# silently placed a metric at the root, and W&B had no namespace-specific step.
logger.log_scalar("training/loss", 1.2, step=10)
logger.log_metrics({"reward": 3.0}, step=4)  # logged as root key "reward"
logger.log_video("evaluation/rollout", video, step=4)

# After: a view applies the namespace to every supported payload consistently.
training = logger.with_prefix("training")
evaluation = logger.with_prefix("evaluation")
assert training.log_metrics({"loss": 1.2}, step=10) == {
    "training/loss": 1.2
}
evaluation.log_scalar("reward", 3.0, step=4)
evaluation.log_video("rollout", video, step=4)

# W&B associates these with independent "training/step" and
# "evaluation/step" metrics. Hyperparameter keys remain unchanged.
training.log_hparams({"learning_rate": 3e-4})

# Before the review fix, this failed existing trainer/config validation because
# PrefixLogger was not recognized as a Logger. Owner-backed views now satisfy
# that contract, while process-service clients remain intentionally restricted.
assert isinstance(training, Logger)
assert not isinstance(process_logger.client().with_prefix("workers"), Logger)

# Delegated lifecycle/client methods also preserve an explicit None return
# instead of constructing a broken PrefixLogger(None, ...).
assert optional_logger.with_prefix("training").client() is None
```


vmoens (commented):

Overall: nice feature, well tested. The flattening in `PrefixLogger.__init__` (chained views collapse to a single level and retain `prefix` bookkeeping), the empty-prefix identity, and the delegation of state/lifecycle via `__getattr__` all behave as documented, and the service-client test proves the namespace survives pickling into spawn workers. The W&B namespaced-step behavior comes from the pre-existing `_step_key` derivation, so the views get it for free — good.

1. **`isinstance(logger, Logger)` guards reject views.** `torchrl/trainers/algorithms/configs/trainers.py` has ~10 sites of the form `if not isinstance(logger, Logger) and logger is not None: raise ...`. `PrefixLogger` deliberately isn't a `Logger` subclass, so passing a view as a trainer logger fails validation. If views are only meant for call sites *inside* a trainer/collector while the trainer owns the root logger, that's a fine design choice — but it should be stated in the `with_prefix` docstring, because passing `logger.with_prefix("training")` to a config object is the first thing users will try. Otherwise, those checks need to accept `PrefixLogger`.

2. **`__getattr__` would wrap `None`.** In `preserve_view`, if the wrapped `client`/`start` ever returns `None`, we return `PrefixLogger(None, prefix)` instead of `None`, which then explodes confusingly at first use (`AttributeError: 'NoneType' object has no attribute ...`). Today `start()` returns `Self` everywhere so it's latent, but a `if logger is None: return None` guard would make the delegation total.

3. Minor test gap: the `TypeError` branch for a non-string `prefix` and the `__repr__` are uncovered. One extra `pytest.raises(TypeError)` line in `test_prefix_logger_composes_namespaces_and_delegates_state` would do.

Docs change is clear and the `log_metrics` return contract (prefixed keys in, prefixed keys out) is pinned by tests on both the plain logger and the W&B branch.

vmoens:
CI triage: the five reported failures are unrelated to this PR, so I am not changing the branch for them.

- `TestTrackioLogger::test_log_histogram[steps1]` fails in both multiprocessing matrices and recurs identically on independent PRs #4405, #4415, #4417, and #4420. This change adds prefix views; the failing test uses the unprefixed Trackio histogram path.
- `TestProcessSlotTransport::test_server_batched_pass_on_cuda[True]` fails in both GPU matrices across the same independent PRs. This PR does not change inference-server or slot-transport code.
- `TestDreamerV3Components::test_discrete_actor[cpu-autocast]` is outside the logger implementation and tests changed here; no Dreamer code is in this diff.

The focused logger tests pass locally. These failures share no changed execution path with #4403 and need no PR-specific patch.

PR #4404, opened 2026-09-17
[Feature] Let GAE consume validity masks

vmoens:
## Summary

- add an opt-in `valid` TensorDict key for right-padded GAE batches
- stop return recursion at the valid-to-padding boundary and sanitize invalid reward/value inputs
- exclude invalid rows from advantage normalization and zero their advantage/value-target outputs
- preserve per-agent normalization in `MultiAgentGAE` and expose nested `valid_key` values through `GAEConfig`
- document that the final valid transition bootstraps from its own `("next", value)` entry

This is intentionally narrower than the original proposal. It does not add a separate trajectory-boundary signal: trajectory boundaries continue to use GAE's existing configurable `done` and `terminated` keys, while recurrent starts continue to use `is_init`.

## Why this is separate from `done` / `truncated`

`done` and `terminated` describe how a real transition ends. A padding row is not a transition at all. A boundary flag can stop temporal recursion, but it cannot express that a row must also be excluded from normalization and produce no training target. The `valid` mask only covers that padding contract.

The validity boundary is recursion-only: the final valid transition still bootstraps unless its real `terminated` flag is set. Its `("next", "state_value")` entry must therefore describe the state reached by that transition.

## Before and after

```python
import torch
from torchrl.objectives.value import GAE

valid = torch.tensor(
    [[True, True, True, False, False], [True, True, False, False, False]]
)
td = TensorDict(
    {
        "state_value": torch.zeros(2, 5, 1),
        "collector": {"mask": valid},
        "next": {
            "state_value": torch.ones(2, 5, 1),
            "reward": torch.arange(10, dtype=torch.float).reshape(2, 5, 1),
            "done": torch.zeros(2, 5, 1, dtype=torch.bool),
            "terminated": torch.zeros(2, 5, 1, dtype=torch.bool),
        },
    },
    batch_size=[2, 5],
)

# Padding may contain arbitrary storage values. NaNs make the failure obvious.
invalid = ~valid
td["state_value"][invalid] = torch.nan
td["next", "state_value"][invalid] = torch.nan
td["next", "reward"][invalid] = torch.nan

# Before: GAE cannot identify padding. The padded suffix is treated as part of
# each trajectory, leaks backward through recursion, and enters normalization.
legacy = GAE(
    gamma=0.9,
    lmbda=0.95,
    value_network=None,
    vectorized=False,
    average_gae=True,
)
legacy_result = legacy(td.clone())
assert not torch.isfinite(legacy_result["advantage"][valid]).all()

# After: the validity key isolates the real prefix. Invalid inputs cannot leak,
# statistics use valid transitions only, and padded outputs are exactly zero.
masked = GAE(
    gamma=0.9,
    lmbda=0.95,
    value_network=None,
    vectorized=False,
    average_gae=True,
)
masked.set_keys(valid=("collector", "mask"))
result = masked(td.clone())
assert torch.isfinite(result["advantage"][valid]).all()
assert not result["advantage"][invalid].any()
assert not result["value_target"][invalid].any()
torch.testing.assert_close(result["advantage"][valid].mean(), torch.zeros(()))
```

## Test plan

- `python -m pytest test/objectives/test_values.py -q -k 'gae and not tensor_gamma'` (874 passed)
- focused validity/normalization selection (3 passed)
- `python -m pytest test/objectives/test_mappo.py -q` (37 passed)
- nested `GAEConfig` selection (1 passed)
- `ufmt`, `flake8`, `pydocstyle`, and `git diff --check` passed for the changed library file


vmoens (commented):

I traced the full path and the semantics line up with the docstring claims:

- `done` gates the recursion and `terminated` gates the bootstrap in both `generalized_advantage_estimate` and `vec_generalized_advantage_estimate`, so OR-ing boundary masks into `done` while leaving `terminated` alone is exactly the "stop recursion, keep bootstrap" behavior. The two tests prove this empirically (advantage at the cut = `gamma * V_next`, suffix poisoning invisible).
- Computing `recursion_terminated` **before** boundaries are OR-ed into `done` is load-bearing — a future refactor that reorders these two lines would silently turn every artificial cut into a termination (no bootstrap). Please leave a one-line comment in the code guarding that ordering; the tests would probably catch it, but only in the `auto_reset_env=False` configuration.
- The `set_keys` switch from `self._AcceptedKeys.__dict__` to `__dataclass_fields__` is a necessary fix, not a cosmetic one: now that `GAE._AcceptedKeys` subclasses the base dataclass, `__dict__` would have rejected every inherited key (`value`, `reward`, …) on `GAE.set_keys`. Good catch; the new check is uniformly more correct for all estimators.
- Masked normalization is wired correctly: `loc`/`scale` are computed over valid entries only, and invalid positions are zeroed afterward rather than contributing garbage to the statistics.

Two documentation nits:

1. The indexing convention for `trajectory_end` is only discoverable from the test: the mask marks the **last** step of the segment, not the first step of the new one. Both readings are natural; the `_AcceptedKeys` docstring should pin it down ("the entry at `t` marks the transition after which the trajectory (re)starts").
2. For `valid`, the last valid transition bootstraps from `("next", state_value)` at the boundary row by design (`done` set, `terminated` clear) — with padded batches that row is padding, so whatever the critic predicted on stale/empty input leaks into that one target. That's standard truncation semantics, but worth one sentence in the `valid` docstring so callers know to give the row after their last valid step a meaningful value (e.g. by evaluating the critic there anyway, which is cheap since it's already computed).

`GAEConfig` plumbing and the `test_mappo` override stay signature-compatible; `in_keys` extension is dup-safe. LGTM modulo nits.

vmoens:

This diff is limited to GAE validity handling, its config, and objective tests; it does not touch inference-server, logger, CUDA-capture, or dependency setup code. No PR change is indicated by these failures.

PR #4405, opened 2026-09-17
[BugFix] Create W&B save directories before initialization

vmoens:
## Summary

- accept path-like W&B save directories and expand user paths
- create configured save or log directories before W&B authentication and initialization
- reject invalid filesystem targets before creating a remote run
- preserve the legacy behavior that an empty save_dir or log_dir means “not provided”
- keep the tilde-expansion regression test portable by using USERPROFILE on Windows and HOME on POSIX

## Test plan

- PYTHONPATH=. python -m pytest test/test_loggers.py test/test_configs.py -k wandb -q
- pytest test/test_loggers.py -q -k 'wandb_creates_save_directory_before_init or wandb_rejects_file_save_directory_before_init' (3 passed)
- ufmt, flake8, pydocstyle, and git diff --check on changed files

## Before and after

```python
from pathlib import Path

import pytest
from torchrl.record.loggers import WandbLogger

requested = Path("~/wandb/runs/nested")

# Before: TorchRL forwarded the PathLike value unchanged and called W&B before
# validating the missing destination:
# wandb.init(dir=Path("~/wandb/runs/nested"))

# After: the same call resolves and creates the destination before wandb.init.
logger = WandbLogger(exp_name="demo", save_dir=requested)
resolved = requested.expanduser()
assert Path(logger.save_dir) == resolved
assert resolved.is_dir()

# Invalid targets now fail locally before a remote run can be created.
invalid = resolved.parent / "is-a-file"
invalid.write_text("not a directory")
with pytest.raises(FileExistsError):
    WandbLogger(exp_name="will-not-start", save_dir=invalid)

# Empty strings keep their previous “not provided” meaning.
fallback = resolved.parent / "fallback"
logger = WandbLogger(exp_name="fallback", save_dir="", log_dir=fallback)
assert Path(logger.save_dir) == fallback
```


vmoens (commented):

The fix is right and well targeted: `expanduser` + `os.fspath` normalizes `PathLike`/`~` inputs, `makedirs(..., exist_ok=True)` both creates the destination and turns the "target is an existing file" case into a clean local `FileExistsError` *before* any remote run is created (nicely pinned by `init.assert_not_called()`).

Two things to double-check:

1. **Empty-string behavior change.** The old branch `save_dir if save_dir and not log_dir else log_dir` silently treated `save_dir=""` as "not provided" (falling back to `log_dir`/wandb default); the new `is not None` checks make `save_dir=""` proceed to `os.makedirs("")`, which raises `FileNotFoundError`. Nobody should be passing `""`, so failing is arguably better — but it's a silent semantic change in a "BugFix" PR. Either treat falsy as not-provided for back-compat or call it out in the summary.

2. **Windows portability of the `log_dir` test.** `test_wandb_creates_save_directory_before_init` monkeypatches `HOME` and relies on `expanduser("~")` honoring it. On Windows, `os.path.expanduser` has used `USERPROFILE`/`HOMEDRIVE`+`HOMEPATH` (not `HOME`) since Python 3.8, so if `test_loggers.py` runs anywhere near a Windows CI job this parametrization will fail. Suggest `pytest.mark.skipif(os.name == "nt", ...)` on the `log_dir` case, or monkeypatching `os.path.expanduser`/using the platform-appropriate variable.

Nit: the `save_dir` attribute now stores the expanded absolute-ish string rather than the user's original value — that's the point of the change, just noting it's observable via `logger.save_dir`/`logger.log_dir` (the test pins it, good).

vmoens:

- `TestTrackioLogger::test_log_histogram[steps1]` fails in both multiprocessing matrices and recurs identically on independent PRs #4403, #4415, #4417, and #4420. This diff is confined to W&B directory normalization and its tests; it does not change Trackio.

The focused W&B directory tests pass locally. No PR-specific patch is appropriate for these failures.

PR #4406, opened 2026-09-17
[Refactor] Export PolicyVersion from generic transforms

vmoens:
## Summary

- move the `PolicyVersion` implementation into the generic environment-transform package
- export it from `torchrl.envs.transforms` while preserving the LLM import as an identical compatibility alias
- update collectors, examples, and reference docs to use the generic namespace

## Tests

- `/Users/vmoens/repos/rl/.venv/bin/python -m pytest test/llm/test_llm_transforms.py::TestPolicyVersion -q`
- `/Users/vmoens/repos/rl/.venv/bin/python -m pytest test/test_collectors.py::TestPolicyVersion -q`
## Before and after

```python
# Before: a generic collector feature could only be imported through the
# domain-specific LLM package.
from torchrl.envs.transforms import PolicyVersion
# ImportError: cannot import name 'PolicyVersion'

from torchrl.envs.llm.transforms import PolicyVersion  # misleading dependency

# After: the generic namespace is canonical.
from torchrl.envs.transforms import PolicyVersion

tracker = PolicyVersion(version_type="int")
assert tracker.version == 0
tracker.increment_version()
assert tracker.version == 1

# Existing callers keep working without receiving a duplicate class.
from torchrl.envs.llm.transforms import PolicyVersion as LegacyPolicyVersion

assert LegacyPolicyVersion is PolicyVersion
```


vmoens (commented):

Verified the move is faithful and the compatibility story is solid:

- The new `torchrl/envs/transforms/policy_version.py` is byte-identical in behavior to the old LLM module (docstrings generalized from `LLMCollector` to collectors), and the old module is a re-export shim, so `torchrl.envs.llm.transforms.PolicyVersion is torchrl.envs.transforms.PolicyVersion` — pinned by `test_compatibility_alias`.
- The shim also preserves **pickle compatibility**: existing pickles naming `torchrl.envs.llm.transforms.policy_version.PolicyVersion` will resolve to the new class, no migration needed. Worth stating that in the commit message / PR body — it's the main reason to prefer a re-export shim over a deprecation warning, and it justifies keeping the shim until `llms_transforms.rst`'s announced v0.17 removal lands.
- No stale references: `grep` across `torchrl/`, `docs/`, and `sota-implementations/` shows every runtime import and every `:class:` cross-reference already on the new path; the remaining mentions in `weight_update/_shared.py` are comments and `data/replay_buffers/samplers/staleness.py` already imports from the generic namespace.
- `VersionChange` is re-exported in the shim's `__all__` even though only `PolicyVersion` gets docs — good, since the dataclass appears in `version_history` and users may have imported it from the old path.

Nothing blocking; this is mergeable as-is.

PR #4407, opened 2026-09-17
[BugFix] Preserve policy versions in shared replay schemas

vmoens:
## Summary

- configure multi-collector policy-version tracking before shared replay-buffer initialization
- include the worker-produced `next/policy_version` field in the lazy storage prototype
- validate pre-initialized shared TensorDict storage against the configured integer or UUID version representation
- preserve UUID-backed policy-version fields instead of applying the integer-only dtype check

## Before and after

```python
from functools import partial

from torchrl.collectors import MultiAsyncCollector
from torchrl.envs.llm.transforms import PolicyVersion
from torchrl.modules import RandomPolicy
from torchrl.testing.mocking_classes import CountingEnv

env_fn = partial(CountingEnv, max_steps=4)
probe = env_fn()
policy = RandomPolicy(probe.action_spec)
probe.close()

# Before: the shared schema was frozen before policy-version tracking was
# configured, so worker-produced ("next", "policy_version") values vanished.
replay = ReplayBuffer(storage=LazyTensorStorage(64), batch_size=4, shared=True)
collector = MultiAsyncCollector(
    [env_fn], policy, replay_buffer=replay, frames_per_batch=8,
    total_frames=16, trajs_per_batch=1, track_policy_version=True,
)
try:
    list(collector)
finally:
    collector.shutdown()
assert ("next", "policy_version") in replay.storage[: len(replay)].keys(True, True)

# Before the review fix, this valid pre-initialized UUID schema was rejected by
# an unconditional `dtype == torch.int64` check. After: validation follows the
# tracker representation, so UUID/non-tensor schemas remain valid while an
# integer tracker still requires an int64 field.
tracker = PolicyVersion(version_type="uuid")
template_env = env_fn()
template = template_env.fake_tensordict()
template.set("next", tracker._step(template, template.get("next")))
template_env.close()
uuid_replay = ReplayBuffer(
    storage=LazyTensorStorage(64), batch_size=4, shared=True
)
uuid_replay.extend(template.unsqueeze(0))

# Before: constructing the collector raised
# RuntimeError("... must have dtype torch.int64 ...").
# After: the existing non-tensor UUID field is accepted and remains writable
# by workers; incompatible tensor-backed UUID schemas are still rejected.
uuid_collector = MultiAsyncCollector(
    [env_fn], policy, replay_buffer=uuid_replay, frames_per_batch=4,
    total_frames=4, track_policy_version=tracker,
)
try:
    list(uuid_collector)
finally:
    uuid_collector.shutdown()
```

## Test plan

- `python -m pytest test/test_collectors.py::TestPolicyVersion::test_multi_collector_accepts_preinitialized_shared_uuid_version_schema -q` (1 passed)
- the earlier integer/missing-schema and lazy shared-schema focused tests also pass (4 passed)
- ufmt, flake8, pydocstyle, and `git diff --check` passed


vmoens (commented):

The root cause and fix check out: shared storage froze its schema from `fake_tensordict()` before the tracker was configured, so the worker-stamped `("next", "policy_version")` column never made it into the lazy-storage prototype. Moving `_setup_multi_policy_version_tracking` ahead of `_setup_multi_replay_buffer` and stamping the fake td via `tracker._step(fake_td, fake_td.get("next"))` (read-only, no version bump) reserves the field, and the pre-initialized-storage path now fails loudly instead of silently dropping the column. The in-code comment explaining the ordering constraint is appreciated — that's exactly the sort of thing a later refactor would undo.

One real gap and one nit:

1. **The dtype validation assumes `version_type="int"`.** `track_policy_version` also accepts a pre-built `PolicyVersion(version_type="uuid")` (the `_hasattr("increment_version")` branch), whose workers stamp `NonTensorData` strings, not int64. With a *pre-initialized* storage + uuid tracker, `_validate_policy_version_replay_schema` raises `"must have dtype torch.int64"` even though the schema is perfectly valid for that tracker. Either branch on `self.policy_version_tracker.version_type` (int → int64 check; uuid → accept NonTensor/string entries), or explicitly reject uuid trackers with shared replay buffers with a dedicated message (if that combination is unsupported anyway — note the lazy-init path with a uuid tracker likely can't `share()`/`memmap` a `NonTensorData` column either, so the second option may just be honest).

2. Nit: the rename of `_add_policy_outputs_to_fake_td` → `_add_collector_outputs_to_fake_td` is truer to what it does now, and I confirmed the only caller was updated; just flagging that it's a private-but-plausibly-Subclassed hook, so renaming without an alias could break downstream subclasses. Acceptable for a private name.

The new tests cover both the `False`/`True` schema outcomes and the rejection path; the `match=` on the RuntimeError guards against message regressions.

vmoens:
CI triage for `6ccee12a1`:

- The Python 3.10/3.13 `mp` jobs fail only in the Trackio logger tests because the installed `trackio` imports the removed `huggingface_hub.hf_api.CommitOperationAdd`. The same environment failure appears on #4408, #4418, and #4419.
- Both GPU shard-2 jobs fail in `TestProcessSlotTransport::test_server_batched_pass_on_cuda[True]` with PyTorch's `Invalid stream capture status`; the identical failure appears on #4408, #4418, and #4419.
- Python 3.14 bulk fails in `TestUpdateIfPresent::test_never_written_slot_is_stale`, and stable-GPU shard 3 fails in `TestDreamerV3Components::test_discrete_actor[cuda-autocast]`.

None of these tests or their implementation areas are touched by this PR (`torchrl/collectors/_multi_base.py` plus its collector test). I found no branch-caused CI failure to patch.

PR #4409, opened 2026-09-17
[BugFix] Filter spawn-bootstrap worker warnings

vmoens:
## Summary

- suppress warnings emitted while spawn workers import bootstrap modules when subprocess filtering is enabled
- serialize spawn startup around the temporary `PYTHONWARNINGS` override and restore the parent environment on success or failure
- preserve existing warning behavior for non-spawn workers and when `torchrl.filter_warnings_subprocess` is disabled
- document that the suppression is intentionally widened from worker `run()` to bootstrap imports

## Test plan

- `python -m pytest test/test_utils.py -q -k "process_spawn_filters_bootstrap_warnings or process_spawn_restores_warning_environment_on_start_failure"` (5 passed)
- `ufmt check torchrl/_utils.py`
- `flake8 --config=setup.cfg torchrl/_utils.py`

## Before and after

```python
# bootstrap_warning.py -- imported by Python while starting a spawn worker
import warnings
warnings.warn("warning emitted during spawn bootstrap", UserWarning)
def target():
    pass

# driver.py
import os
import warnings

import torchrl
from torchrl._utils import _ProcessNoWarnSpawn

# Ignore the intentional warning from the parent import; the regression is the
# second import performed by the fresh spawned interpreter.
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    from bootstrap_warning import target

os.environ["PYTHONWARNINGS"] = "always"
torchrl.filter_warnings_subprocess = True
process = _ProcessNoWarnSpawn(target=target)
process.start()
process.join()
process.close()

# Before: the filter was installed in Process.run(), after the fresh Python
# interpreter had already imported bootstrap_warning.py.
# stderr: "UserWarning: warning emitted during spawn bootstrap"  # leaked

# After: warnings are filtered during spawn startup itself. This deliberately
# extends the pre-existing run-time suppression to bootstrap imports.
# stderr: ""
assert os.environ["PYTHONWARNINGS"] == "always"  # parent setting restored
```


vmoens (commented):

Root cause and fix both check out: the old filter was installed in `Process.run()`, i.e. *inside* the already-booted spawned interpreter, so anything emitted while that interpreter re-imported modules to unpickle `target` leaked to stderr. Moving suppression to the parent's `PYTHONWARNINGS` around `start()` is the right lever as `exec` captures the environment synchronously before `start()` returns, so restoring immediately after is race-free. The global start lock correctly acknowledges that `os.environ` is process-global in the parent.

Checks I ran through:

- Both `_ProcessNoWarn` (fork context, early-yield) and `_ProcessNoWarnSpawn` (spawn) are covered, and the fork path deliberately touches nothing.
- Failure-path restoration is tested via a patched `start` raising mid-call — the `try/finally` and the unset-vs-previous `PYTHONWARNINGS` branches both hold.
- End-to-end stderr assertion (`capfd`, exact warning count) is the right kind of test here; the parametrized `filter_warnings=False` leg proves the suppression is conditional on the torchrl flag.

Nits, non-blocking:

1. `PYTHONWARNINGS="ignore"` suppresses *everything* the child imports during bootstrap, including warnings user code may rely on seeing exactly once. That's already the semantic of the `run()`-time filter this extends, so it's consistent — but one sentence in the contextmanager docstring saying "this widens the existing run()-time suppression backward to cover bootstrap imports" would make the intent explicit.
2. Serializing all spawn `start()`s behind one global lock is fine (`start()` is just `popen` creation), but if collectors ever fan out dozens of workers through this path it becomes a visible serialization point; worth a `TODO` note rather than a fix now.

vmoens:
CI triage: the current failures are unrelated to this PR's `_utils.py` / `test_utils.py` diff.

- Stable GPU shard 3 additionally has an isolated `TestDreamerV3Components::test_discrete_actor[cuda-autocast]` numerical assertion. This PR does not touch Dreamer or module code, and the other stable/nightly shard-3 executions passed.

No branch patch is appropriate for these dependency/upstream CUDA/numerical failures.

PR #4410, opened 2026-09-17
[Feature] Expose acknowledged collector policy versions

vmoens:
## Summary

- add an explicit per-worker policy-version query for multiprocessing collectors
- expose the cheap aggregate only while all workers are known to have acknowledged the same synchronous updates
- invalidate the aggregate after partial, asynchronous, or failed synchronization and preserve it through checkpoints
- keep asynchronous collection running after an explicit worker-version query
- document the invariant that the base updater increments once per synchronized model before the aggregate is reconstructed

## Test plan

- `python -m pytest test/test_collectors.py -q -k "test_multi_collector_restores_worker_policy_versions or test_async_weight_update_invalidates_aggregate_policy_version or test_multi_collector_bumps_on_update" --tb=short --timeout=60` (6 passed, 506 deselected)
- `ufmt check torchrl/collectors/_multi_base.py`
- `flake8 --config=setup.cfg torchrl/collectors/_multi_base.py`

## Before and after

```python
collector = MultiSyncCollector(
    [make_env, make_env],
    policy=policy,
    frames_per_batch=20,
    total_frames=200,
    track_policy_version=True,
)
try:
    collector.update_policy_weights_()
    collector.update_policy_weights_()
    assert collector.policy_version == 2

    collector.update_policy_weights_(worker_ids=[0])

    # Before: policy_version was just the parent-side tracker (3 here), even
    # though worker 0 had version 3 and worker 1 still had version 2. Treating
    # that scalar as the version of all collected data was incorrect, and no
    # public API exposed the divergence.

    # After: the cheap aggregate is available only after every worker has
    # synchronously acknowledged the same update.
    assert collector.policy_version is None
    assert collector.worker_policy_versions() == {0: 3, 1: 2}
finally:
    collector.shutdown()
```


vmoens (commented):

The semantics are now honest: a scalar aggregate only exists when it's actually meaningful, and partial/async/failed sync invalidates it rather than lying. The failure-path ordering is correct — `_acknowledged_policy_version = None` is set *before* `super().update_policy_weights_()`, so an exception mid-sync (with some workers already bumped) leaves the aggregate invalidated, exactly as intended. Checkpoint round-trip preserves invalidation, and the worker-query/re-`continue` dance keeps async collection alive after RPCing workers (the `test_multi_collector_bumps_on_update` drain assertions prove collection resumes).

One fragility to keep an eye on:

- **The reset-and-redo couples tightly to `super().update_policy_weights_()`'s increment count.** The override resets `tracker.version` to the pre-update acknowledged version and then re-increments once *per model_id in weights_dict*. That produces the right answer iff the base class's internal increments are (a) fully resettable and (b) equal in number to `model_ids` for the full-sync path. If the base ever changes to "one increment per `update_policy_weights_` call" while `weights_dict` carries two models, the aggregate silently drifts +1 per update. A comment in the override stating the invariant ("base increments once per synced model on the full path") would at least make the coupling greppable when someone touches the base.

- Note: this interplay is *why* the per-frame `("next", "policy_version")` tags remain the source of truth — worth keeping that sentence in the class docstring even though the property is now trustworthy when non-None.

Nits:

- `worker_policy_versions()` sending `"continue_random"`/`"continue"` after the RPC is correct for the async pipe protocol, but the reliability hinges on `map_fn` having fully drained worker replies; a one-line code comment pointing at the protocol requirement would help future maintainers (the docstring's "should not race" warning covers user threads, not the pipe-state dependence).
- `worker_ids` as a single `torch.device` → `all_workers_targeted = False` is conservative and fine; just noting the asymmetry with a single `int` is intentional-looking but undocumented.

The three new test legs (failed sync keeps workers at {0:2, 1:2} with aggregate None; partial update → {0:3, 1:2}; async scheme → aggregate None immediately) pin the exact semantics the PR promises.

vmoens:
CI triage: the current failures do not overlap this collector policy-version diff.

This PR changes only `test_collectors.py` and `_multi_base.py`; it does not touch Trackio/loggers or the inference server. No branch patch is appropriate for these failures.

PR #4411, opened 2026-09-17
[BugFix] Make multiprocess collector shutdown quiet and idempotent

vmoens:
## Summary

- finalize each worker inner collector when the parent pipe closes, breaks, or a process-group interrupt reaches the worker
- make parent shutdown idempotent and tolerant of already-dead workers and closed IPC resources
- apply one global monotonic deadline across graceful join, terminate, and kill fallback phases, while joining every worker that can be reaped
- keep the internal deadline out of TorchRL's global timing/telemetry registry
- preserve ordinary worker exceptions instead of treating all connection-related failures as shutdown

## Test plan

- `python -m pytest test/test_collectors.py -q -k "test_multiprocess_shutdown_is_bounded_and_idempotent or test_multiprocess_shutdown_tolerates_dead_worker or test_worker_disconnect_runs_collector_cleanup" --tb=short --timeout=90` (3 passed, 511 deselected)
- prior focused collector lifecycle and failure-propagation selection: 7 passed
- `ufmt check torchrl/collectors/_multi_base.py`
- `flake8 --config=setup.cfg torchrl/collectors/_multi_base.py`
- broader `TestCollectorGeneric` run reached 75 passed and 8 skipped; four subprocess-only cases failed because this host has no executable named `python` on `PATH`, then the run was stopped

## Before and after

```python
import time
from torchrl.collectors import Collector, MultiSyncCollector

class SlowWorkerCollector(Collector):
    def shutdown(self, timeout=None, close_env=True, raise_on_error=True):
        time.sleep(60)  # simulate a worker stuck during cleanup

collector = MultiSyncCollector(
    [make_env, make_env],
    collector_class=SlowWorkerCollector,
    frames_per_batch=20,
    total_frames=-1,
    cat_results="stack",
)

# Before: shutdown polled workers and then joined them independently; a stuck
# worker could outlive the requested timeout. The first bounded implementation
# also registered an internal deadline in the global TorchRL timing registry.

# After: one plain-monotonic deadline covers graceful close, terminate, and
# kill fallback without adding lifecycle samples to training telemetry.
started = time.monotonic()
collector.shutdown(timeout=0.5)
assert time.monotonic() - started < 1.0
assert all(not process.is_alive() for process in collector.procs)

collector.shutdown(timeout=0.5)  # idempotent no-op, including after dead workers
```


vmoens (commented):

The design is right and the three failure modes are now handled where they occur:

- Child side: only the three *expected* disconnect paths (parent pipe EOF, broken pipe, process-group `KeyboardInterrupt`) are swallowed after cleanup; other `ConnectionError`s propagate (tested). Capturing `inner_collector` in a ref list and clearing it after the explicit shutdown avoids double-shutdown.
- Parent side: claiming `closed` + `_shutdown_complete` up-front makes repeated calls and `__del__`-driven shutdowns no-ops even halfway through teardown; pipe/queue errors during teardown are tolerated rather than fatal.
- The 80/15/5 deadline split bounds the entire sequence, and the kill-fallback `AttributeError` guard acknowledges `Process.kill()` is 3.7+.

Findings, all minor:

1. **`timeit` registers `multi_collector_shutdown_deadline` into the global `timeit` registry.** Every shutdown appends a sample that will surface in `timeit.printreg()` dumps (SOTA scripts log the whole registry periodically), and there is no `timeit.erase(...)` for it. For an internal deadline, plain `time.monotonic()` avoids polluting the training-telemetry namespace with a lifecycle event. (API-wise, `.start()`/`.elapsed()` do exist, so this works — it's about registry hygiene, not correctness.)

2. The graceful phase no longer drains the `"closed"` replies workers send before exiting. That's fine (small message, no buffer-full deadlock), just noting the asymmetry is deliberate-looking rather than accidental; a short comment would help.

3. Test cleanup loop (`if not proc._closed and proc.is_alive(): terminate`) is good hygiene, but `_closed` is CPython-internal; fine in tests, noted for future us.

The `timeout < 0` guard and phase math check out: `timeout=0` still forces terminate→kill→join(0) promptly; `elapsed < 1.0` with in-process 60s-sleeping workers is the right regression target. Bound is honored sequentially because every phase computes `remaining` from the same monotonic deadline.

vmoens:
CI triage: the current failures do not overlap this collector-shutdown diff.

This PR changes collector lifecycle code and its tests; it does not touch Trackio/loggers or the inference server. No branch patch is appropriate for these failures.

PR #4412, opened 2026-09-17
[Feature] Add explicit trajectory replay write mode

vmoens:
## Summary

- add an explicit replay_write_mode selector for rollout or complete-trajectory replay writes
- keep replay_buffer plus trajs_per_batch as a backward-compatible legacy selector while separating yielded trajectory batch sizing from replay insertion
- propagate the mode through direct, multiprocess, Hydra, and Ray collector paths; retain trajs_per_write grouping and validate explicit conflicts before worker startup
- retain RPC support through its remote `collector_kwargs` path and document that routing explicitly
- document the trajectory-boundary contract, update a maintained training example, and report automatic replay sharing at info level because it mutates user-provided storage

## Tests

- `python -m pytest test/test_collectors.py::TestTrajsPerBatch test/test_collectors.py::TestTrajsPerBatchReplayBuffer -q --tb=short --timeout=90` (53 passed)
- focused explicit rollout, trajectory, grouping, incomplete-episode, sync and async multiprocess tests (12 passed)
- collector Hydra config parity tests (7 passed)
- `ufmt` and `flake8` checks on the review follow-up files

## Before and after

```python
import torch
from torchrl.collectors import MultiAsyncCollector
from torchrl.data import LazyTensorStorage, ReplayBuffer, SliceSampler

rb = ReplayBuffer(
    storage=LazyTensorStorage(16_000, ndim=1),
    sampler=SliceSampler(slice_len=32, end_key=("next", "done")),
)

# Before: fixed-frame worker rollouts could be interleaved in 1-D replay.
broken_write = TensorDict(
    {
        "collector": {"traj_ids": torch.tensor([12, 12, 91, 91])},
        "next": {"done": torch.tensor([False, False, False, False])},
    },
    batch_size=[4],
)
ids = broken_write["collector", "traj_ids"]
done = broken_write["next", "done"]
hidden_boundary = ids[:-1].ne(ids[1:]) & ~done[:-1]
assert hidden_boundary.any()  # SliceSampler sees no end between traj 12 and 91.

# The only old opt-in for complete writes was the overloaded trajs_per_batch,
# which also controls the collector's yielded-batch contract.

# After: replay ingestion is selected independently and explicitly.
collector = MultiAsyncCollector(
    [make_env] * 4,
    policy=policy,
    replay_buffer=rb,
    frames_per_batch=256,
    total_frames=-1,
    replay_write_mode="trajectory",
    trajs_per_write=8,  # grouping only; not a yielded-batch setting
)
collector.start()

# Every replay extend now contains flat, complete trajectories, for example:
# collector/traj_ids: [12, 12, 12, 91, 91]
# next/done:          [ F,  F,  T,  F,  T]
# SliceSampler can recover every boundary without crossing episodes.
```


vmoens (commented):

The separation this introduces (yielded-batch contract vs replay-ingestion contract) is the right call, and I verified the plumbing end to end:

- Legacy resolution is faithful: `replay_buffer + trajs_per_batch + mode=None` still resolves to trajectory writes in `_validate_replay_write_mode`, and *conflicting* explicit combinations raise before worker startup rather than silently picking a side. The parametrized validation test pins all five shapes.
- Multi-process propagation is consistent: parent nulls its `trajs_per_batch` (routing `__iter__` to `iterator()`), workers detect trajectory writes either via the explicit kwarg or the legacy `trajs_per_batch` presence, and the inner collector re-validates with the real `replay_buffer` it receives.
- `reset()`-time flushing of partial assembly is unconditional in both single and multi paths, so the new mode can't leak pre-reset steps into post-reset batches (checked — `_flush_trajectory_assembly` is not gated on `trajs_per_batch`).
- Docs are coherent across the four pages touched, and the `trajs_per_write` re-scoping (grouping only, not a yielded-batch setting) is now stated everywhere it matters.

Open points, in descending order of importance:

1. **RPCCollector silently dropped from the supported list.** The BaseCollector docstring previously listed `RPCCollector` among collectors supporting trajectory writes; now it names only MultiSync/MultiAsync/RayCollector, and `collectors_distributed.rst` matches. Mechanically RPCCollector subclasses MultiCollector so it likely still works — so this reads as either an intentional demotion (untested/unsupported) or an accidental documentation removal. Either way, please make it deliberate: if RPC is no longer covered, the old text shouldn't have claimed it; if it is, keep it listed.

2. **Automatic `.share()` demoted from warning to debug.** In `_setup_multi_replay_buffer`, mutating the user's buffer to shared now happens silently. Re-sharing re-allocates lazy storage as memmapped tensors — not a cheap no-op — so a user passing an unshared buffer by mistake now has no signal that their object was modified. I'd keep at least an `info`-level log; `debug` makes the surprise invisible.

3. Nit: `self._worker_replay_write_mode = replay_write_mode` stores the *raw* argument rather than the validated result. Today they're only equal in the one case that reaches workers, but assigning the validated value (`self.replay_write_mode` would over-normalize, so really just `replay_write_mode` after validation has run) costs nothing and removes the implicit coupling.

4. Nit: `trajs_per_write is only supported with replay_write_mode='trajectory'` — the string concatenation `"only supported with " "replay_write_mode=..."` is just ufmt-appeasing noise; a single line under 88 chars works.

The `_flush_trajectory_assembly` docstring update ("when trajectory assembly is in use") matches the code; the `AsyncBatchedCollector` cross-reference in `_single.py` keeps the old terminology consistent.

vmoens:
CI triage: the current failures do not overlap this trajectory replay-ingestion diff.

This PR does not touch Trackio/loggers or the inference server. No branch patch is appropriate for these dependency and upstream CUDA failures.

PR #4413, opened 2026-09-17
[Quality] Add safe recurrent GAE construction and validation

vmoens:
## Summary

- add `GAE.for_recurrent(...)` with correctness-first defaults: `shifted=False`, `deactivate_vmap=True`, and `average_gae=True`
- require time to be the final batch dimension and validate root/next recurrent state plus `is_init` before value evaluation
- infer recurrent state keys from the value-network input/output contract and document that the estimator is bound to the network passed to the factory
- accept nested group and validity-mask keys without duplicating the grouped normalization introduced in #4212
- document the safer constructor in the recurrent state lifecycle guide

## Stacking

This PR targets `value-estimator-boundary-masks` because it wires the `valid` key introduced by #4404. It has been rebased onto the narrowed, valid-only version of that prerequisite and no longer carries the removed `trajectory_end` API.

No `GAE.__init__` keyword was added; the existing `GAEConfig` coverage for `group_key`, `valid_key`, and the constructor overrides remains unchanged.

## Tests

- focused `GAE.for_recurrent` selection (8 passed)
- `ufmt`, `flake8`, and `git diff --check` passed for the changed Python files
- `pydocstyle` passed for the changed library file; the existing broad test module has unrelated pre-existing docstring findings

## Before and after

```python
import torch
from torch import nn
from torchrl.modules import GRUModule
from torchrl.objectives import GAE

B, T, H = 2, 8, 4
gru = GRUModule(
    input_size=3,
    hidden_size=H,
    in_keys=["observation", ("recurrent", "h")],
    out_keys=["features", ("next", "recurrent", "h")],
)
critic = TensorDictSequential(
    gru,
    TensorDictModule(nn.Linear(H, 1), ["features"], ["state_value"]),
)

# A realistic replay sample where the next recurrent state was dropped.
td = TensorDict(
    {
        "observation": torch.randn(B, T, 3),
        "is_init": torch.zeros(B, T, 1, dtype=torch.bool),
        "recurrent": {"h": torch.zeros(B, T, 1, H)},
        "next": {
            "observation": torch.randn(B, T, 3),
            "is_init": torch.zeros(B, T, 1, dtype=torch.bool),
            # ("next", "recurrent", "h") is missing
            "reward": torch.randn(B, T, 1),
            "done": torch.zeros(B, T, 1, dtype=torch.bool),
            "terminated": torch.zeros(B, T, 1, dtype=torch.bool),
        },
    },
    batch_size=[B, T],
    names=[None, "time"],
)

# Before: GAE(...) had no recurrent-data contract. Unsafe shifted/vmap choices
# and missing reset/state tensors were discovered only inside value evaluation.

# After: correctness-first defaults plus validation at the GAE boundary.
gae = GAE.for_recurrent(
    gamma=0.99,
    lmbda=0.95,
    value_network=critic,
)
assert gae.shifted is False
assert gae.deactivate_vmap is True

gae(td)
# KeyError: "GAE.for_recurrent requires root and next reset/state inputs;
# missing keys: [('next', 'recurrent', 'h')] ..."
```


vmoens (commented):
**Review (COMMENT)** — own PR, comment-only. (Reviewing the delta on top of #4404 — the boundary/valid machinery itself was already reviewed there.)

The factory + validation combo is conservative in the right places:

- State-key inference via "input key with a matching `("next", ·)` output" is compatible with `GRUModule`/`LSTMModule`, which auto-append `"is_init"` to `in_keys` — I verified the doctest example in the classmethod actually constructs (GRU case appends it at `rnn.py:2536`).
- Opt-in via classmethod means zero behavior change for the plain `GAE(...)` path; `_recurrent_state_keys` is a plain instance attribute set post-construction, so deepcopy/state_dict round-trips are unaffected.
- The validation is explicitly *structural* (presence, dtype, shape pairing, time-last layout) and cheap relative to the value-net call; the time-last check also cleanly rejects the `value_chunk_dim > 0` corner instead of letting it fail cryptically downstream.

Points:

1. **Docstring accuracy nit:** the summary/classmethod docs say the factory "discovers a named `time` dimension at call time". What the code actually does is resolve `time_dim` and then *require* it to equal `batch_dims - 1`; the name-based discovery is preexisting `_get_time_dim` behavior shared with plain GAE. Consider rewording to "requires time to be the last batch dimension" so the factory doesn't claim a capability the base class owns.

2. **No cross-check that state keys coincide with the value network's `in_keys` at call time.** Inference happens at construction; if the same GAE instance is later pointed at a different value net (e.g. loss rebuild sharing the estimator), `_recurrent_state_keys` silently describes the old network. Cheap guard: recompute in `forward` if `self.value_network` changed identity, or just document that the estimator is bound to the network passed to `for_recurrent`.

3. The double `_get_time_dim` resolution (once in `_validate_recurrent_inputs`, once later in `forward`) is harmless; not worth restructuring.

4. Test coverage targets the right failures (missing root/next state, wrong `is_init` dtype, non-last time) and parametrizes `shifted` for both paths — good. One gap: no test for the `for_recurrent(..., value_network=None)` ValueError branch, though it's a one-liner.

Stacking note acknowledged: diff rebase noise from #4404 makes this look bigger than it is; actual delta is the factory, the validator, docs, and tests.

vmoens:

This diff is limited to recurrent GAE construction/validation, its objective tests, and recurrent-state documentation; it does not touch inference-server, logger, CUDA-capture, or dependency setup code. No PR change is indicated by these failures.

PR #4414, opened 2026-09-17
[Feature] Add structured batch-time output to SliceSampler

vmoens:
## Summary

- add `SliceSampler(output_layout="batch_time")` for stable `[num_slices, time]` batches with a named time dimension
- pad short non-strict rows with repeated safe indices and emit `("collector", "mask")`
- emit a configurable nested `slice_end_key` at the final real transition while leaving stored `done`, `terminated`, and `truncated` values unchanged
- preserve the legacy flat layout by default and warn once per sampler when it returns fewer transitions than requested
- keep `SliceSampler(replacement=False, ...)` forwarding and Hydra configuration in parity
- document structured recurrent/GAE use and benchmark both layouts
- clarify the saved requested-length fallback, add sampler context to invariant errors, and document transform responsibility for renamed dimensions

## Before and after

```python
import torch
from torchrl.data import LazyTensorStorage, SliceSampler, TensorDictReplayBuffer

# One non-terminal trajectory has only two real transitions.
data = TensorDict(
    {
        "traj_id": torch.zeros(2, dtype=torch.long),
        "step": torch.arange(2),
        ("next", "done"): torch.zeros(2, 1, dtype=torch.bool),
    },
    batch_size=[2],
)

# Before: non-strict sampling returned a concatenated short batch. A caller
# requesting two four-step rows received only four items, with no sampled-slice
# boundary distinct from the environment's done flag.
flat = TensorDictReplayBuffer(
    storage=LazyTensorStorage(2),
    sampler=SliceSampler(
        slice_len=4, traj_key="traj_id", strict_length=False
    ),
    batch_size=8,
)
flat.extend(data)
sample = flat.sample()
assert sample.batch_size == torch.Size([4])  # requested 8
sample.reshape(2, 4)
# RuntimeError: the four returned transitions cannot form [2, 4].

# After: opt into a stable recurrent layout. Padding repeats a safe index, while
# the mask excludes it and slice_end marks the final real sampled transition.
structured = TensorDictReplayBuffer(
    storage=LazyTensorStorage(2),
    sampler=SliceSampler(
        slice_len=4,
        traj_key="traj_id",
        strict_length=False,
        output_layout="batch_time",
    ),
    batch_size=8,
)
structured.extend(data)
sample = structured.sample()

assert sample.batch_size == torch.Size([2, 4])
assert sample.names == [None, "time"]
assert sample["step"].tolist() == [[0, 1, 1, 1], [0, 1, 1, 1]]
assert sample["collector", "mask"].tolist() == [
    [True, True, False, False],
    [True, True, False, False],
]
assert sample["collector", "slice_end"].squeeze(-1).tolist() == [
    [False, True, False, False],
    [False, True, False, False],
]
assert not sample["next", "done"].any()  # environment semantics are untouched

# Existing GAE semantics express a sampled cut by stopping recursion through
# done while leaving terminated unchanged, so the cut still bootstraps.
sample["next", "done"] |= sample["collector", "slice_end"]
```

## Stack

This targets `value-estimator-boundary-masks` from #4404 because structured samples use its `valid` mask to exclude padding. It has been rebased onto the narrowed, valid-only prerequisite. Callers that feed sampled cuts to GAE combine `slice_end` into the configured `done` field while leaving `terminated` unchanged.

## Testing

- focused batch-time, short-batch warning, and replacement selection (6 passed)
- the end-to-end recurrent/GAE test still pins the cut-but-bootstrap result at `1 + 0.9 * 1 = 1.9`
- `ufmt`, `flake8`, and `git diff --check` passed for the changed Python files
- `pydocstyle` passed for the changed library file; the existing broad test module has unrelated pre-existing docstring findings


vmoens (commented):
**Review (COMMENT)** — own PR, comment-only. Delta reviewed against #4404's branch (the advantages.py hunks in this diff are the stacked prerequisite).

The design composes cleanly: `batch_time` reuses the existing `pad_output` index-padding machinery for non-strict short rows rather than growing a parallel one, `slice_end` supersedes `truncated_key` marking only in structured mode, and the `is_init` OR-ing logic is shared with the flat path (real resets inside slices are preserved, fabricated ones aren't emitted when the storage has no marker — the "we'd be lying" comment is right).

The end-to-end test (`test_slice_sampler_batch_time_recurrent_and_gae_boundaries`) pins the whole stack: padded rows get zero advantage/target through #4404's `valid` plumbing, and the last *real* step gets `1 + 0.9*1 = 1.9` under `trajectory_end` — recursion cut, bootstrap kept. Also good: strict mode deliberately emits no `("collector", "mask")`, so "no mask" remains distinguishable from "all valid".

Points:

1. **Dead fallback in `_get_index`.** `structured_seq_length` is only `seq_length` when `isinstance(seq_length, int)`, but it's only *used* when `isinstance(seq_length, torch.Tensor)` — the assignment can never produce a non-None value. As written it's harmless (`time_size` falls back to the int `seq_length`), but either delete it or add a comment explaining the case it was meant to cover; as-is it reads like a bug-in-waiting for the layout-shape `RuntimeError` below it.

2. `time_size is None or index.shape[0] != num_slices * time_size` raising "Please report this issue" — fair, since the preceding logic is supposed to preclude it. Consider including `output_layout`, `strict_length`, `pad_output`, and the shapes in the message to make the inevitable bug report actionable on first contact.

3. Small API-surface note: `_set_sample_names` is applied after collate but *before* `self._transform(data)`. Transforms that reshape/rename batch dims (rare for RB transforms) will inherit stale names. Fine to leave, possibly worth a sentence in the docstring.

4. Docs/config/benchmark parity is complete (`pad_output` was even missing from `SliceSamplerConfig` and got added). `SliceSamplerWithoutReplacementConfig` forwards the three new fields — good.

The once-per-sampler short-batch warning for the flat layout is a nice touch: it keeps the legacy behavior audible without being spammy, and the test proves "once" rather than "filtered by warnings machinery".

vmoens:

This diff is limited to SliceSampler/replay layout behavior, its config, docs, tests, and benchmark; it does not touch inference-server, logger, CUDA-capture, or dependency setup code. No PR change is indicated by these failures.

PR #4415, opened 2026-09-17
[Feature] Add a Trainer hook for scheduled asynchronous evaluation

vmoens:
## Summary

- add EvaluatorHook to schedule and poll asynchronous evaluation through Trainer lifecycle hooks
- snapshot policy weights at trigger time, coalesce missed frame intervals, and log results under evaluation/*
- drain pending work and run one latest-policy final evaluation without duplicating a completed final-frame evaluation
- bound shutdown waits to 60 seconds by default, with an explicit timeout override and None opt-in for unbounded waits
- checkpoint scheduling progress without serializing evaluator work
- preserve the existing Evaluator busy policy and LogValidationReward

## Before and after

```python
# Existing objects:
#   trainer   - a Trainer whose collector reports frame counts
#   evaluator - an asynchronous torchrl.collectors.Evaluator

# Before: applications had to implement cadence themselves. This common pattern
# silently lost intervals when collection jumped over an exact multiple and did
# not evaluate the final policy between interval boundaries.
if trainer.collected_frames % 10_000 == 0 and not evaluator.pending:
    evaluator.trigger_eval(current_policy_weights(), step=trainer.collected_frames)

# Example batch-end frame counts: 10k, 25k, 35k.
# Result: only 10k is evaluated; the crossed 20k interval and final 35k policy
# are never measured.

# After: scheduling, coalescing, final evaluation, and bounded draining are a
# Trainer lifecycle concern.
from torchrl.trainers import EvaluatorHook

EvaluatorHook(
    evaluator,
    every_frames=10_000,
    policy="loss_module.actor_network",
    run_at_end=True,
    wait_at_end=True,
    wait_at_end_timeout=60.0,
).register(trainer)
trainer.train()

# Triggered frames are [10_000, 25_000, 35_000]. If an evaluation never
# finishes, shutdown exits through the existing error-and-cleanup path after
# 60 seconds instead of waiting forever. Pass None only to request no limit.
```

## Tests

- python -m pytest test/test_trainer.py -k EvaluatorHook -q (5 passed)
- full test/test_trainer.py: 156 passed, 19 skipped; 9 unrelated prioritized-replay cases require the unavailable SumSegmentTree C++ extension
- ufmt, flake8, pydocstyle, Python compilation, and git diff --check


vmoens (commented):

Verified against the real `Evaluator` API: `trigger_eval` returns a bool, `pending` covers queued work, and `poll`/`wait`/`shutdown` exist with the exact semantics the hook assumes. The coalescing math (`intervals = (step - next_due) // every_frames + 1` evaluated on the *trigger-time* step) is right for both the normal and busy cases, and the tests pin all three schedules (normal cadence, busy coalescing, duplicate-final suppression).

Things I checked that are fine:

- **No double-logging.** Appending to `trainer._log_dict` + setting `_last_log[name]` does not re-log through the trainer's pass (`Trainer._log` only forwards what it's handed as kwargs at call time); the hook's direct `logger.log_metrics` is the single write. Using the same step-key contract (`evaluation/step`) as the W&B namespacing machinery from #4403 is a nice fit.
- **`busy_policy` interplay:** the hook never triggers while `pending`, so `busy_policy="error"` can't fire from this path and `"queue"` degenerates to one outstanding request — preserves the existing busy policy as advertised.
- Checkpoint contents are exactly the durable part (next due frame + last completed) and in-flight work is discarded, as documented; the resume test shows a current-policy re-evaluation happens at setup after restoring a pending-era state, which is the sane interpretation of "discard".
- `_setup` shuts the evaluator down on any exception — prevents a half-registered hook from leaking a service.

Points:

1. **`_drain` can block shutdown forever.** `self.evaluator.wait()` is called with no timeout in `_drain`, which `wait_at_end=True` (the default) runs at `_shutdown`. If the evaluation env hangs, `trainer._shutdown_hook()` never returns and the training process can't exit cleanly. Suggest either a `wait_at_end_timeout` hook parameter or documenting that `wait_at_end=True` has no upper bound.

2. **`_schedule` ignores the `accepted` return.** If `trigger_eval` returns `False` (busy skip), `_last_triggered_frame` is not advanced but `_schedule` will retry next batch anyway because `pending` is the real gate — so behavior is correct; the variable naming just implies the value is used for something. Nit.

3. Naming nit: `register()` on the hook vs trainer-driven registration — every other hook is registered via `trainer.register_op(...)`; this one self-registers ops + module. That's arguably cleaner for a multi-op hook, but the docstring should show it (the `Examples:` does call `.register(trainer)`, good).

4. `wait_error` path: on `wait()` failure, `_shutdown` propagates *and* still calls `evaluator.shutdown()` in `finally` — the combination (error visible + service cleaned up) is exactly right, covered by the test.

vmoens:

- `TestTrackioLogger::test_log_histogram[steps1]` fails in both multiprocessing matrices and recurs identically on independent PRs #4403, #4405, #4417, and #4420. This diff adds the evaluator Trainer hook and does not change Trackio.

The focused `EvaluatorHook` tests pass locally. No PR-specific patch is appropriate for these failures.

PR #4416, opened 2026-09-17
[Feature] Report in-flight trajectory collection progress

vmoens:
## Motivation

Complete-trajectory collection can hold long unfinished episodes outside replay. Replay write counts therefore remain flat even while environments are stepping, which makes healthy asynchronous collectors look stalled and prevents frame-budget accounting from using collector progress.

## Changes

- add cheap collector progress metrics for stepped frames, completed-trajectory frames, pending trajectory frames, replay-written frames, and completed trajectories
- update progress at the environment-step, trajectory-completion, and actual replay-write ownership points without scanning buffered trajectory data
- give each multiprocessing worker and the coordinator distinct parent-owned shared counter rows, so aggregate stats remain non-blocking and require no worker RPC
- preserve the existing `frames` meaning
- aggregate the same metrics from Ray worker snapshots
- checkpoint cumulative counters; reset the pending gauge on reset, shutdown, and resume because partial trajectory payloads and environment state are not checkpointed
- count pending frames when trajectory assembly is selected either by legacy `trajs_per_batch` or by the explicit `replay_write_mode="trajectory"` selector

## Before and after

```python
from torchrl.collectors import Collector
from torchrl.envs import GymEnv, StepCounter, TransformedEnv
from torchrl.envs.utils import RandomPolicy

# Completed trajectories are written to replay only after the second step.
env = TransformedEnv(GymEnv("Pendulum-v1"), StepCounter(max_steps=2))
replay = ReplayBuffer(storage=LazyTensorStorage(32))
collector = Collector(
    env,
    RandomPolicy(env.action_spec),
    frames_per_batch=1,
    total_frames=8,
    replay_buffer=replay,
    trajs_per_batch=1,
)
iterator = iter(collector)

assert next(iterator) is None  # one environment step, trajectory still open

# Before: the existing delivered-frame count could advance while replay stayed
# empty, but it could not say whether the frame was pending, completed, or lost.
# In the explicit replay-write mode, the first implementation also keyed only
# on trajs_per_batch and therefore reported zero pending frames.
assert collector.stats()["frames"] == 1
assert len(replay) == 0

# After: the same cheap snapshot reports where every collected frame is. The
# pending condition also recognizes replay_write_mode="trajectory" when that
# selector is used by the collector.
stats = collector.stats()
assert stats["stepped_frames"] == 1
assert stats["trajectory_pending_frames"] == 1
assert stats["trajectory_completed_frames"] == 0
assert stats["replay_written_frames"] == 0
assert stats["completed_trajectories"] == 0

assert next(iterator) is None  # the two-step trajectory now completes
stats = collector.stats()
assert stats["stepped_frames"] == 2
assert stats["trajectory_pending_frames"] == 0
assert stats["trajectory_completed_frames"] == 2
assert stats["replay_written_frames"] == 2
assert stats["completed_trajectories"] == 1
collector.shutdown()
```

## Tests

- unfinished trajectory progress, completion transfer, and no double-counting
- reset and shutdown behavior
- single-collector checkpoint/resume
- two-worker shared aggregation without worker RPC
- multiprocessing checkpoint/resume
- coordinator-side complete-trajectory assembly through the existing evaluator-process tests
- existing collector reload and complete-trajectory replay selections
- review follow-up: `python -m pytest test/test_collectors.py -q -k "test_complete_trajectory_progress_and_checkpoint or test_multiprocess_trajectory_progress_is_parent_shared" --tb=short --timeout=90` (2 passed, 511 deselected)
- `ufmt` and `flake8` checks on `torchrl/collectors/_base.py`

Focused collector result: 13 passed, 1 skipped (Ray unavailable). After fixing coordinator-owned trajectory assembly, all four evaluator-process regressions and both focused progress tests pass together (6 passed). The existing collector benchmark suite covers this hot path, but its local smoke run could not collect because `psutil` is not installed in the shared development environment.


vmoens (commented):

Verified the accounting invariants by walking each collector configuration:

- **Single collector, `trajs_per_batch` set:** `rollout()` increments pending per stepped frame (`trajectory_pending=True`), completion in `_traj_ingest` subtracts exactly the completed frames, flush zeroes the gauge. The single-process test pins all five counters across steps, reset, checkpoint, and resume.
- **Multi, no replay buffer (assembly on parent):** workers count steps only; the parent's ingest path adds pending via `_collector_progress_pending_on_ingest`. No double counting, and the coordinator row exists precisely for this case.
- **Multi, replay buffer, legacy `trajs_per_batch`:** workers assemble; their stepped calls carry `trajectory_pending=True` since the inner collector's `trajs_per_batch` is non-None. Tested.
- Single-writer-per-row + `snapshot(None)` sum gives the non-blocking aggregate (`stats()` patched-`map_fn` assertion proves no RPC), and `state_dict` round-trips restore cumulative rows on both levels while zeroing the gauge.

**One real gap (forward-compatibility with #4412):** in worker-side trajectory assembly selected via `replay_write_mode="trajectory"` (the *explicit* mode), the inner collector has `trajs_per_batch=None`, so `_record_stepped_frames(..., trajectory_pending=self.trajs_per_batch is not None)` is False and `_collector_progress_pending_on_ingest` is never forwarded to workers. Result: `trajectory_pending_frames` reads 0 *precisely in the configuration the metric exists for* (async collectors holding long unfinished episodes outside replay). Since #4412 is landing in the same chain, plumb the flag through `_run_processes` kwargs to the inner collector (or make `increment_stepped`'s pending condition `trajs_per_batch is not None or replay_write_mode == "trajectory"`) in whichever PR lands second — otherwise users of the new explicit mode get a permanently-zero gauge.

Smaller notes:

- After `total_frames` is exhausted mid-episode, the pending gauge keeps the in-flight remainder until `reset()`/`shutdown()` — that's documented, but worth knowing: a "completed" collector still reports pending frames. Fine as-is; just confirming that's the intended reading.
- `RawArray("q", ...)` single-writer rows: reads can in principle tear an int64 while a worker writes, but for telemetry counters this is an acceptable, documented-as-cheap trade — the delimiters (`num_workers * len(_KEYS)` layout) are consistent across all writers/readers I checked.
- The `stats(workers="both")` namespace (`worker_<idx>/stepped_frames`) matches the existing `worker_frames` convention, and Ray's aggregation skips missing keys gracefully.

The `_traj_ingest` return-value change is backward compatible for ignore-the-result callers, and the test that frames can be stepped without a completion while `len(replay) == 0` nails the motivating observability hole.

vmoens:
CI triage: the current failures do not overlap this collector-progress diff.

This PR changes collector progress accounting and its tests; it does not touch Trackio/loggers or the inference server. No branch patch is appropriate for these failures.

PR #4417, opened 2026-09-17
[Feature] Add standard PPO telemetry

vmoens:
## Summary

- add telemetry="standard" diagnostics to OnPolicyTrainer, with telemetry="minimal" preserving the existing metric set and avoiding the extra collection work
- log new frame, episode, terminal, reward, optimizer, throughput, collector, and replay diagnostics through a training/* logger view while retaining legacy metric names
- segment complete episodes by reset markers as well as trajectory-id changes, so a reused id cannot merge independent episodes
- retain the primary optimizer learning-rate metric and expose every parameter group when multiple groups are configured
- keep Hydra config and factory arguments aligned, document the behavior, and benchmark the standard-mode overhead

## Before and after

```python
import torch
from torchrl.trainers.algorithms import PPOTrainer

# Two complete four-step episodes deliberately reuse trajectory id 0. The
# second is still distinct because is_init marks its first step.
batch = TensorDict(
    {
        "is_init": torch.tensor([[True], [False], [False], [False]] * 2),
        ("collector", "traj_ids"): torch.zeros(8, dtype=torch.long),
        ("next", "reward"): torch.arange(1.0, 9.0).unsqueeze(-1),
        ("next", "done"): torch.tensor(
            [[False], [False], [False], [True]] * 2
        ),
        ("next", "terminated"): torch.tensor(
            [[False], [False], [False], [True]] * 2
        ),
    },
    batch_size=[8],
)

# Before (and still available as the explicit low-overhead mode), PPO emitted
# only the legacy set; repeated ids could collapse episode summaries and only
# optimizer.param_groups[0] was visible.
PPOTrainer(**common, telemetry="minimal").train()
assert not any(name.startswith("training/") for name in logger.records)

# After: standard telemetry is namespaced, honors reset boundaries, and exposes
# every optimizer group while retaining the primary learning-rate metric.
optimizer.add_param_group(
    {"params": [extra_parameter], "lr": 2e-4, "name": "critic"}
)
PPOTrainer(**common, optimizer=optimizer, telemetry="standard").train()
assert logger.records["training/episodes/completed"][-1][1] == 2
assert logger.records["training/episodes/return/mean"][-1][1] == 18.0
assert "training/optimizer/learning_rate" in logger.records
assert "training/optimizer/learning_rate/critic" in logger.records
```

## Stack

This PR is intentionally based on #4403 for Logger.with_prefix. It should be rebased onto main after that narrow prerequisite lands.

## Testing

- pytest test/test_trainer.py test/test_configs.py -k 'OnPolicyTelemetry or ppo_telemetry' -q — 4 passed
- pytest benchmarks/test_objectives_benchmarks.py -k ppo_telemetry_overhead --benchmark-only -q — 2 passed; local means were about 0.036 ms for minimal and 0.204 ms for standard on a 1,280-transition batch
- full test/test_trainer.py — 154 passed, 19 skipped; 9 unrelated prioritized replay cases require the unavailable C++ SumSegmentTreeFp32 extension
- ufmt, flake8, pydocstyle, and git diff --check


vmoens (commented):

The structure is good: legacy metrics untouched, new metrics go through one `training/`-prefixed view (composing with #4403 namespacing), every optional metric degrades to omission rather than zero-fill, and PPO defaults flip the overhead on with an explicit `"minimal"` escape hatch.

Things I verified rather than took on faith:

- Timer phases are correct: `post_steps(N-1) -> pre_steps_log(N)` is exactly collection time (optim for N-1 happens before `post_steps`), and `pre_steps_log(N) -> post_optim_complete_log(N)` is exactly optim time. The update counts use fresh baselines from the prior batch, no cross-batch bleed.
- The four op hooks exist (`setup`, `pre_steps_log` gets the batch, `post_optim_complete_log` gets `(optim_count, average_losses)`, `post_steps`); calling conventions match.
- `_scalar_per_transition`'s squeeze-to-batch-ndims rule handles `[B,T,1]` and gracefully declines anything feature-heavy — omission is the documented behavior.
- The per-key interval gate in `_log_standard` uses its own `training/`-namespaced `_last_log` entries, so it can't collide with same-named legacy keys (the legacy keys keep their un-namespaced history entries).
- Episode completeness rule (both `is_init` and `done` seen inside the same batch, mask-respected) is the only defensible definition without cross-batch state, and "omit if undefined" is honored both for missing `traj_ids`/`is_init` and for batches with zero complete trajectories.

Points, none blocking:

1. **`_complete_episode_metrics` merges episodes that share a `traj_id` within one batch.** If `_TrajectoryPool` ever re-issues an id inside the same delivered batch (collector restart with the pool reset mid-batch), two independent episodes get summed into one return — silently wrong, not noisy. A cheap guard is `unique_consecutive` per sorted-id run instead of `unique`, mirroring `_traj_ingest`'s approach in the collectors. If ids are documented unique-per-batch, a comment citing that invariant is enough.

2. `optimizer/learning_rate` logs `param_groups[0]["lr"]` only. For losses with separate actor/critic groups (the common PPO setup in this repo's SOTA scripts), the *second* group is often the more interesting one. Either log all groups as `optimizer/lr/<group_name>` or note the limitation.

3. `timeit("on_policy/collection")` / `("on_policy/optimization")` add permanent entries to the global `timeit` registry; unlike #4411's shutdown deadline these are per-batch uses, which is at least telemetry-shaped. Fine — just confirming it's deliberate, since they'll show in every `printreg()` dump.

4. Nice touch reusing `done/terminated/truncated` sibling-key resolution from the trainer's configured keys rather than hard-coding `"next"` paths.

5. `replay/size` falls back to `len()` — good; note the fallback's exception catch (`AttributeError, RuntimeError, TypeError`) won't protect against an exotic buffer whose `__len__` itself blocks (remote), but that's a stretch case.

Coverage observation: the telemetry-off path (minimal) is asserted via absence of `training/` keys; the "no collection work happens at all" claim is only pinned by the benchmark addition. If overhead matters enough to ship a mode for, one cheap unit test asserting `_OnPolicyTelemetry` is never instantiated in minimal mode would outlive benchmark rot.

vmoens:

- `TestTrackioLogger::test_log_histogram[steps1]` fails in both multiprocessing matrices and recurs identically on independent PRs #4403, #4405, #4415, and #4420. This diff adds on-policy telemetry and does not change Trackio histogram handling.

The focused telemetry and configuration tests pass locally. No PR-specific patch is appropriate for these failures.

PR #4418, opened 2026-09-17
[Feature] Add a rate-limited replay buffer

vmoens:
## Summary

- add `RateLimitedReplayBuffer`, a `ReplayBuffer` subclass that caps cumulative sampled items relative to cumulative inserts
- reserve sample budget atomically for direct and shared-memory callers, with timeout, cancellation, and shutdown wakeups
- reject prefetching because it can consume budget before the caller reserves it
- checkpoint the ratio target and wait counter together with the existing replay-buffer counters

## Scope

This replaces the standalone `ReplayFlowControl` coordinator. Policy-version tracking and publication gating have been removed: per-entry freshness remains the responsibility of existing APIs such as `StalenessAwareSampler` and `PolicyAgeFilter`, while collectors and trainer hooks continue to own weight publication.

The result adds one specialized replay-buffer subclass rather than a second lifecycle-owning base abstraction.

## Stack

Depends on #4408 for replay readiness conditions and cumulative 64-bit insert/sample counters. This PR contains no producer-admission policy.

## Validation

- `TestRateLimitedReplayBuffer` plus the shared-process regression: 5 passed
- rate-limit benchmark: passed
- ufmt, flake8, and `git diff --check`: passed


vmoens (commented):
**Review (COMMENT)** — own PR, comment-only. (Built on #4408's counters; the two changesets compose as intended.)

The reservation design is the right shape: budget checked and reserved atomically under the buffer's readiness condition, released in `finally` regardless of how `rb.sample` exits, and the underlying buffer's own cumulative counters stay the source of truth for `write_count`/`samples_returned`. I walked the failure interleavings (concurrent reserves, timeout mid-wait, sampler exception after reserve, cancel during `rb.sample`) and the accounting stays exact.

Things I verified:

- **Cumulative semantics are the point:** after circular wrap the ratio still uses lifetime counters (the test with 8 writes / 2×2 samples / budget 0 pins this) — circular capacity is documented as *not* being a backpressure signal, consistent with #4408's docs.
- Nested `condition` re-entry in `stats()` → `can_publish()` works because both `threading.Condition` and `multiprocessing.Condition` default to RLock.
- A sampler returning *fewer* items than requested (flat SliceSampler, strict_length=False) refunds the slack correctly, since the finally releases `batch_size` while the buffer only counts what it returned — the budget owed to the learner remains. Worth a comment in `sample()` saying this is deliberate; it looks like a bug until you think about who owes whom.
- Cross-process coordination via fork inheritance of `lock=False` values is covered by the shared test; spawn inheritance of raw shared-memory ctypes also works, so the "create after share(True)" guidance suffices.

Points:

1. **`record_policy_publication` rejects the *initial* publication before any sample exists.** `can_publish` returns `False` when `policy_lag_count == 0`, so the first-ever real version bump after `initial_policy_version` must wait for a sample to establish the lag baseline. That's the conservative choice (no data → can't prove lag), but it should be stated in the class docstring; someone wiring "publish v1 right after training starts" will trip it.

2. **Gating uses the latest sample's *min* version only.** `policy_version - latest_min <= lag` means a batch spanning versions [3..7] under lag=1 lets v8 publish even though half the batch is at v4 (mean lag 4). Consider whether the intended semantic is min (optimistic) or a stricter one — and if min is the choice, say why (freshness of the newest data?) in the docstring, because I'll bet someone reads "oldest version in the latest sampled batch" and implements the mean.

3. `torch.as_tensor(versions)` on a `NonTensorData` (uuid version) won't reach the friendly "must use an integer dtype" TypeError; it'll raise whatever as_tensor raises for strings. Since uuid trackers are a documented `track_policy_version` mode, a one-line NonTensorData guard would keep the failure mode self-explanatory. Sampled data without the key raises KeyError — good, strict is right here.

4. Checkpoint restores config *and* counters but deliberately not reservations (`_reserved_samples=0`) — matches the "checkpoint rb and control together" note. One thing the note could add: `write_count`/`samples_returned` live in the *buffer's* checkpoint, so restoring control alone gives a misleading budget until the next write.

5. Nit: `sample_wait_count` increments once per waiting *stint*, not per poll iteration — the right choice; just noting the test's `sample_wait_count == 2` encodes it.

The 64-bit migration of the rb counters from #4408 is load-bearing for the ratio math here (a 32-bit `write_count` wrapping would break the budget at ~4B frames); good that it landed first.

vmoens:
CI triage for `e2c0d2a16`:

- The Python 3.10/3.13 `mp` jobs fail only in Trackio logger setup because the installed `trackio` imports the removed `huggingface_hub.hf_api.CommitOperationAdd`.
- Both GPU shard-2 jobs fail in `TestProcessSlotTransport::test_server_batched_pass_on_cuda[True]` with PyTorch's `Invalid stream capture status` during CUDA graph capture.

These exact failures recur on #4407, #4408, and #4419 and do not exercise replay flow control. I found no branch-caused CI failure to patch.

test/rb/test_rb_core.py
```diff
@@ -1929,6 +1930,179 @@ def wait_for_replay():
...
+        control = ReplayFlowControl(
+            rb,
+            samples_per_insert=1.0,
+            max_policy_lag=1,
+            policy_version_key=version_key,
+            initial_policy_version=7,
+        )
+
+        control.sample(timeout=1)
+        assert not control.can_publish(8)
```
theap06: flaky, one run in four samples both 7s. use a sampler without replacement.

torchrl/data/replay_buffers/flow_control.py
```diff
@@ -0,0 +1,571 @@
...
+        """
+        batch_size = self._resolve_batch_size(batch_size)
+        deadline = self._deadline(timeout)
+        self._reserve_sample(batch_size, deadline=deadline, cancel_event=cancel_event)
+        condition = self.replay_buffer._readiness_condition
+        try:
+            remaining = None
+            if deadline is not None:
+                remaining = max(0.0, deadline - time.monotonic())
+            data, info = self.replay_buffer.sample(
```
theap06: prefetched batches skip the reservation and feed the lag gate stale data. reject prefetching buffers in the constructor.

theap06 (changes requested):

theap06 (approved):
LGTM Thanks!

PR #4420, opened 2026-09-17
[Refactor] Adopt logger namespaces in SOTA recipes

vmoens:
## Summary

- adopt Logger.with_prefix across 17 CI-maintained SOTA recipes
- route unprefixed metrics through training, evaluation, and timing views
- keep base loggers responsible for lifecycle, checkpoint state, and existing video wiring
- add CSV-backed PPO and SAC smoke tests that check on-policy and off-policy namespace layouts end to end
- omit SAC's unavailable first-iteration speed instead of sending None to scalar-only backends

This PR is stacked on #4403 and should merge after it. Logged series intentionally move from train/*, eval/* or test/*, and time/* to training/*, evaluation/*, and timing/*. Existing checkpoint metric payloads keep their legacy keys.

For W&B, each namespace owns its step metric. A training iteration can therefore advance training/step, evaluation/step, and timing/step independently instead of sharing one global step axis.

## Before and after

```python
# Before: every recipe manually encoded the namespace in every key.
# A missing prefix created a root-level chart, while combining this payload
# with a logger view would accidentally produce "training/train/loss".
metrics = {
    "train/loss": loss,
    "eval/reward": reward,
    **timeit.todict(prefix="time"),
}
logger.log_metrics(metrics, step)

# After: views own namespaces and payload keys stay local to their domains.
training = logger.with_prefix("training")
evaluation = logger.with_prefix("evaluation")
timing = logger.with_prefix("timing")

training.log_metrics({"loss": loss}, step)
evaluation.log_metrics({"reward": reward}, step)
timing.log_metrics(timeit.todict(), step)

# Scalar-only backends receive speed only once tqdm has computed one.
speed = pbar.format_dict["rate"]
if speed is not None:
    timing.log_metrics({"speed": speed}, step)
```

## Scope

The migration covers A2C, PPO, RND, GAIL, DDPG, TD3, SAC, discrete SAC, CrossQ, online CQL, online IQL, and DQN recipes. Bespoke logging systems that call W&B directly or inspect concrete logger types are left for focused follow-ups.

## Test plan

- pytest test/test_loggers.py -k 'prefix or prefixed' -q
- pytest '.github/unittest/linux_sota/scripts/test_sota.py::test_commands[ppo_mujoco]' -q (1 passed)
- pytest '.github/unittest/linux_sota/scripts/test_sota.py::test_commands[sac]' -q (1 passed)
- representative A2C, CrossQ, and online CQL SOTA smokes
- pinned ufmt, flake8, Python compilation, and git diff --check


vmoens (commented):

The migration is mechanically clean across all 17 recipes. Things I checked:

- **No key drift.** Grep over the added lines shows no remaining manually-prefixed keys (`train/`, `eval/`, `test/`, `time/`) reaching a view, so no accidental `training/train/...` — the exact failure mode the summary calls out — and no prefixed video keys: views are used only for metrics, existing video wiring stays on base loggers.
- **Value handling is unchanged.** The old per-recipe `utils.log_metrics` was a trivial `logger.log_metrics` wrapper, so SAC-style raw multi-element tensors (`episode_rewards`) hit the same `_make_metrics_safe` path before and after (multi-element → `tolist()`). No silent semantic change hiding in the helper removal.
- **Checkpoint payloads keep legacy keys** via explicit reconstruction (`train/`, `eval/`, `time/` mapping in ppo_mujoco), and the no-logger path still excludes the `time/*` block because `timing_metrics` is only populated under `if logger:`. Someone clearly thought about the `latest_metrics` consumers rather than just swapping calls.
- **End-to-end pin on ppo_mujoco** (csv backend + assertions over the actual directory layout, including the "exactly one `scalars` root" check) is the right kind of test — it would catch both double-prefixing and a view that lost its prefix.

Points, non-blocking:

1. **The CSV-layout smoke only covers the on-policy pattern.** The other 16 recipes are by construction similar, but a `timing/` or `evaluation/` typo in any one of them is only caught by eyeballing now. Extending the same 4-line assertion to one off-policy recipe (SAC or TD3) would make the whole class of mistakes CI-visible; alternatively note the analogy coverage explicitly in the PR description.
2. Minor inconsistency: `ppo_mujoco` guards `timing_metrics["speed"]` against `pbar.format_dict["rate"] is None`; the other recipes assign it unconditionally (preexisting behavior). First-iteration `None` flows into `log_metrics` — the stack handles it, but either guard everywhere or drop the guard in ppo for symmetry.
3. W&B side effect worth a sentence in the PR body: with #4403's per-namespace step metrics, each recipe iteration now issues up to three `log()` + `define_metric` cycles (`training/step`, `evaluation/step`, `timing/step`). That's the design, but dashboards keyed on a single global step now have three step axes; worth calling out since dashboards outside the repo may compare against `train/*` series that no longer update.

Checked against #4403 behavior: `with_prefix` returns `self` for `""`, views delegate `state_dict`/lifecycle, and `log_metrics` returns prefixed keys — all consistent with how these recipes use it.

vmoens:

- `TestTrackioLogger::test_log_histogram[steps1]` fails in both multiprocessing matrices and recurs identically on independent PRs #4403, #4405, #4415, and #4417. This diff updates SOTA recipe call sites and smoke coverage; it does not change Trackio.

The CSV-backed PPO and SAC smoke tests pass locally. No PR-specific patch is appropriate for these failures.

PR #4434, opened 2026-09-19
[Algorithm] Add RLHF reward-model training recipe (#3923)

theap06:
Re-lands #3923 on `main`.

#3923 was squash-merged into its stacked base branch `theap06/reward-model-loss` rather than into `main`, so the recipe never reached `main`. Its base #3922 has since merged as 0daf99d438. This branch is the #3923 squash commit cherry-picked onto `main`; the resulting tree is identical to the approved #3923 head 75f46e432e.

See #3923 for the full description, review and CI history.


torchrl/data/llm/reward.py
```diff
@@ -29,6 +30,12 @@ class RewardData:
 class PairwiseDataset:
     """Represents a dataset in a pairwise manner (chosen vs rejected).
 
+    .. deprecated:: 0.14
```
vmoens, line 33: This should have been edited.
@theap06

Issue #4433, opened 2026-09-19
[Feature Request] Add Support for FQL


neuriv:
https://arxiv.org/abs/2502.02538

Add reusable components and a reference implementation for offline and offline-to-online RL.
- Add a flow-matching policy module with configurable networks and a distilled one-step actor.
- Add FQLLoss covering flow-matching behavior cloning, policy distillation, Q-maximization, and critic learning.
- Reuse existing replay buffers, collectors, target updates, and evaluation utilities.
- Add a state-based SOTA recipe with Hydra configuration, documentation, tests, CI coverage, and benchmark comparisons against the reference implementation.

i'm interested in seeing this in TRL & can make a draft pr for it.

PR #4436, opened 2026-09-19
[Feature] Add Flow Policies

neuriv:
## Description

Add configurable flow-matching and one-step policies with optional explicit noise inputs. Use Euler integration for flow sampling and expose bounded actions through TorchRL’s existing actor interface.

Part of #4433.

<details>
<summary>Testing</summary>

Ub22.04, A100 40 GB, Python 3.11.16, PyTorch 2.14.0+cu126 and TensorDict 0.14.2.

```bash
root=/home/ubuntu/fql-validation-20260919
suite="$root/tests"
evidence="$root/evidence"
export PATH="$root/rl/.venv/bin:$PATH"
export PYTHONPATH="$PWD" FQL_TEST_DEVICE=cpu,cuda COMPOSITE_LP_AGGREGATE=0
export CUDA_VISIBLE_DEVICES=0 MUJOCO_GL=egl OMP_NUM_THREADS=1
export HYDRA_FULL_ERROR=1 WANDB_MODE=disabled WANDB_SILENT=true

pre-commit run --from-ref ad969ceebf8a8d13f9e5ccc778682165b6aa5069 --to-ref HEAD
python -m pytest -q "$suite/test_flow.py" test/modules/test_actor.py -k 'not diffusion'

for seed in 7 11 19; do
    python "$suite/check_flow.py" --device cuda --seed "$seed" --output "$evidence/flow/seed-$seed"
    python "$suite/verify_flow_evidence.py" "$evidence/flow/seed-$seed"
done
```

144 tests passed; one optional-dependency test skipped. New policy tests ran on CPU and CUDA. Pre-commit passed.

Each seed ran 800 flow updates, 400 distillation updates and 64 Pendulum steps. Saved weights reproduced the held-out measurements; all rollout rewards were finite and actions stayed within bounds.

| Seed | Flow MSE, before → after | Distillation MSE, before → after |
| --- | --- | --- |
| 7 | 1.214591 → 0.170245 | 0.119854 → 0.000303 |
| 11 | 1.252260 → 0.169834 | 0.163158 → 0.000558 |
| 19 | 1.180907 → 0.165501 | 0.128044 → 0.000355 |

</details>


torchrl/modules/models/flow.py
```diff
@@ -0,0 +1,73 @@
...
+        if num_steps < 1:
+            raise ValueError("num_steps must be positive")
+        self.velocity_network = velocity_network
+        self.action_dim = action_dim
+        self.num_steps = num_steps
+
+    def velocity(self, observation, action, time):
+        return self.velocity_network(torch.cat((observation, action, time), -1))
+
+    def forward(self, observation, noise=None):
```
vmoens: Would be cool to type these inputs

torchrl/modules/models/flow.py
```diff
@@ -0,0 +1,73 @@
...
+        if noise is None:
+            noise = observation.new_empty(
+                (*observation.shape[:-1], self.action_dim)
+            ).normal_()
+        action = noise
+        for step in range(self.num_steps):
```
vmoens, line 70: We could use scan to make it fast when compiled

torchrl/modules/models/flow.py
```diff
@@ -0,0 +1,73 @@
...
+        self.network = network
+        self.action_dim = action_dim
+
+    def forward(self, observation, noise=None, *, clamp=True):
+        if noise is None:
+            noise = observation.new_empty(
+                (*observation.shape[:-1], self.action_dim)
+            ).normal_()
+        action = self.network(torch.cat((observation, noise), -1))
+        return action.clamp(-1, 1) if clamp else action
```
vmoens: We should make clamp low and and high parametrizable no?

torchrl/modules/models/flow.py
```diff
@@ -0,0 +1,73 @@
...
+    def forward(self, observation, noise=None):
+        if noise is None:
+            noise = observation.new_empty(
+                (*observation.shape[:-1], self.action_dim)
+            ).normal_()
+        action = noise
+        for step in range(self.num_steps):
+            time = action.new_full((*action.shape[:-1], 1), step / self.num_steps)
+            action = action + self.velocity(observation, action, time) / self.num_steps
+        return action.clamp(-1, 1)
```
vmoens: Shouldn't it be parametrizable (Low and high)?

torchrl/modules/models/flow.py
neuriv: i think that would be BC breaking

torchrl/modules/models/flow.py
```diff
@@ -0,0 +1,145 @@
...
+
+@implement_for("torch", "2.14")
+def get_flow_scan() -> Callable | None:  # noqa: F811
+    return torch._higher_order_ops.scan
+
+
+FLOW_SCAN = get_flow_scan()
+
+
+class FlowMatchingPolicy(nn.Module):
```
vmoens: [P2] Expose TensorDict-compatible public policies

Both `FlowMatchingPolicy` and `OneStepPolicy` are exported as policies, but are plain `nn.Module`s with no `in_keys`/`out_keys` and accept tensors only. Direct invocation on a TensorDict fails with `AttributeError: 'TensorDict' object has no attribute 'normal_'`. The documented `Actor(...)` wrapper works, but the public policy classes should follow TorchRL's TensorDict policy interface and declare their inputs/outputs. Please expose them through the existing actor/TensorDict module interface, retaining the tensor-only integration logic as a model implementation and wiring explicit noise through declared keys.

torchrl/modules/models/flow.py
```diff
@@ -0,0 +1,145 @@
...
+        self.velocity_network = velocity_network
+        self.action_dim = action_dim
+        self.num_steps = num_steps
+        self.register_buffer("low", torch.as_tensor(low))
+        self.register_buffer("high", torch.as_tensor(high))
+
+    def velocity(self, observation: Tensor, action: Tensor, time: Tensor) -> Tensor:
+        return self.velocity_network(torch.cat((observation, action, time), -1))
+
+    def forward(self, observation: Tensor, noise: Tensor | None = None) -> Tensor:
```
vmoens, line 61: [P2] Commit behavioral coverage for the new public classes

There are no tests for either class in this PR or its head tree. The external `test_flow.py` and validation scripts named in the description are not committed, so repository CI cannot exercise the new eager/compiled scan paths, explicit-noise inputs, or clipping behavior. Please add compact tests in the existing actor/module test area: a known velocity field with independently calculated Euler output, supported eager/compiled agreement, and an Actor integration case covering noise/bounds and student gradients. Any CUDA-only cases also need `pytest.mark.gpu`.

vmoens (changes requested):
Rechecked the latest head `c34f9ead1d5b9f337bd520e4d7b31e166bbd284f`. Requesting changes for the public policy interface and missing repository tests, detailed inline. The documented Actor wrappers and local CPU eager/compiled arithmetic checks pass; the interface comment concerns the exported policy contract, not the documented wrapped usage. CUDA and native Inductor execution were not validated locally.

neuriv:
Thanks for the reviews! Updated - @vmoens

torchrl/modules/tensordict_module/actors.py
```diff
@@ -153,6 +154,113 @@ def get_dist(self, tensordict: TensorDictBase) -> torch.distributions.Distributi
         return Delta(action)
 
 
+class FlowMatchingPolicy(SafeModule):
```
vmoens: SafeModule introduces a bit of overhead. For now I'd use a regular TensorDictModule and assess later if we want it to be Safe

torchrl/modules/tensordict_module/actors.py
```diff
@@ -153,6 +154,113 @@ def get_dist(self, tensordict: TensorDictBase) -> torch.distributions.Distributi
...
+        super().__init__(
+            FlowMatchingModel(
+                velocity_network, action_dim, num_steps, low=low, high=high
+            ),
+            in_keys=["observation", "noise"] if in_keys is None else in_keys,
+            out_keys=["action"] if out_keys is None else out_keys,
+        )
+
+
+class OneStepPolicy(SafeModule):
```
vmoens: Ditto

vmoens (commented):
Thanks for adding this! A few comments, mostly around the scan implementation.

The current scan path looks more complicated/restricted than necessary to me:
```

if (
    FLOW_SCAN is not None
    and is_compiling()
    and not torch.is_grad_enabled()
    and num_steps > 1
):
```

In particular, why do we disable scan when gradients are enabled? Recent PyTorch scan has explicit autograd support (ScanAutogradOp), and its backward is itself implemented in terms of a reverse scan. Since this PR targets/tests PyTorch 2.14, I think we should be able to rely on that rather than falling back to a Python loop for training.

Using scan for training is also useful beyond just avoiding Python overhead. In particular, scan can expose loop-specific compilation options such as unroll/similar controls, allowing several iterations to be unrolled/fused together by the compiler. This seems particularly relevant here since an Euler sampler is exactly the sort of small, fixed recurrence where reducing loop/kernel overhead can matter.

Similarly, I’m not sure we need the is_compiling() restriction. scan is the structured-control-flow representation of this computation, so ideally the model should express the recurrence as a scan whenever the supported PyTorch version provides it, and let eager/compile/autograd deal with that representation appropriately. If there is a specific eager or compile+autograd limitation in 2.14 that requires these restrictions, could we document it and add a regression test for it?

I’d expect the implementation conceptually to be closer to:
```

if FLOW_SCAN is not None:
    # scan implementation
else:
    # compatibility fallback for older torch
```

rather than maintaining two execution paths based on compile and grad state.

The first-step peeling and action.clone() are more understandable: scan requires carry metadata to remain stable, so the first step can establish the actual carry dtype/layout, and scan currently has aliasing restrictions between carry/output that make the clone necessary. It might be worth commenting explicitly that these are scan API requirements/workarounds, though, because otherwise the implementation looks unnecessarily mysterious.

Relatedly, the current gradient test doesn’t actually exercise scan gradients: test_flow_policy_nested_keys_and_gradients enables gradients, which forces FlowMatchingModel onto the Python-loop branch. I think we should have a test that compares outputs and parameter/input gradients between the scan implementation and the explicit Euler loop. That would both validate the intended semantics and prevent us from accidentally carrying an obsolete no-grad restriction.

A few other comments from reading the diff:

* Could we expose/configure the scan unrolling option on FlowMatchingModel? For a policy with e.g. 10 Euler steps this could be quite useful under torch.compile, and is one of the advantages of representing this as scan rather than a Python loop.
* FLOW_SCAN = get_flow_scan() plus the two @implement_for definitions feels like quite a bit of machinery. We could just wrap the forward.
* It would be useful to test numerical equivalence between scan and the reference loop for several num_steps, dtypes/devices, rather than having the choice of implementation implicit in compiled/grad mode.
* Since low and high are registered independently, should construction validate low < high (after broadcasting), or is the convention elsewhere in TorchRL to leave invalid bounds unchecked?
* FlowMatchingPolicy / OneStepPolicy currently accept arbitrary in_keys, but the implementation semantically requires one or two inputs in a specific order. It may be worth validating their length and documenting clearly that the first is observation and second is noise, otherwise malformed in_keys will fail later in a less obvious way.
* More generally, I wonder whether we can keep FlowMatchingModel very small and make scan the canonical implementation, with the explicit loop only being the old-PyTorch compatibility path. Right now having separate eager, compiled-inference, and training behavior adds quite a lot of surface area for what is fundamentally just an Euler recurrence.

So my main suggestion would be: use scan as the normal implementation on versions where it’s supported, including with autograd; retain the Python loop only where we have a demonstrated compatibility requirement; expose/use scan’s unrolling capabilities; and add a scan-vs-loop forward/backward equivalence test. That would make this both simpler and more useful for compiled training.

test/modules/test_actor.py
vmoens, line 1: We should always put tests in a test class.
Names should be shorter.
Compile compatibility without graph breaks needs to be asserted.

vmoens:
Sorry about the unroll suggestion, that introduced a lot of complexity -- I wrongly assumed PyTorch was like Jax and had that kind of kwarg 
Corrected that

vmoens (approved):
Wonderful thanks!

PR #4443, opened 2026-09-20
[BugFix] Build ActionDiscretizer bin positions with arange(nint) / nint

Nicholas022400701:
`ActionDiscretizer.custom_arange` builds the bin positions with `torch.arange(0, 1, step=1 / nint)`. When `1 / nint` does not accumulate to exactly 1.0 that call returns `nint + 1` positions instead of `nint` (nint = 49, 98, 103, 107, ... 140 values below 2000, same in float32 and float64). With an int `num_intervals` the transform then fails in `transform_input_spec` with `The expanded size of the tensor (49) must match the existing size (50)`. With a tensor `num_intervals` the extra position goes unnoticed: `(1 - result).flip(0)` is aligned one element off, so MEDIAN and HIGH sampling silently return the lower edge of every bin (bin k gives `low + k * width` instead of `low + (k + 0.5) * width` or `low + (k + 1) * width`).

This switches to `torch.arange(nint) / nint`, which always returns exactly `nint` positions in `[0, 1)` and matches the previous values for every `nint` that already worked.

Test: `test_num_intervals_arange_rounding` in `TestActionDiscretizer` uses 49 intervals on `ContinuousActionVecMockEnv`, checks that the discrete spec has 49 bins per dimension and that stepping bins 0, 8, ..., 48 lands on `low + (k + offset) * width` for LOW, MEDIAN and HIGH, with int and tensor `num_intervals`. On main 5 of the 6 cases fail (the three int cases with the expand error, the tensor MEDIAN and HIGH cases with wrong values). With the fix all 6 pass and the rest of `TestActionDiscretizer` is unchanged (38 passed, 83 skipped for missing mujoco and gym).


vmoens (approved):
LGTM thanks for contributing

PR #4444, opened 2026-09-20
[BugFix] Honor min_value=0 and max_value=0 in ParameterScheduler

Nicholas022400701:
## Description

`ParameterScheduler.__init__` (torchrl/data/replay_buffers/scheduler.py L45-46) stores the bounds as `min_value or float("-inf")` and `max_value or float("inf")`, so `min_value=0` and `max_value=0` are treated like the `None` default and silently dropped. With `StepScheduler(rb, "alpha", gamma=-0.2, mode="additive", min_value=0)` alpha is not clipped at 0: it reaches -0.1 on the third step and `PrioritizedSampler.alpha` raises `ValueError: alpha must be greater or equal than 0`. `beta` has no setter check, so it just goes negative.

This PR replaces the two `or` fallbacks with explicit `is not None` checks. Nothing else changes.

## Test

`test_parameter_scheduler_zero_bound` in test/rb/test_samplers.py runs a `StepScheduler` with `min_value=0` on alpha (0.5 -> 0.3 -> 0.1 -> 0.0 -> 0.0) and with `max_value=0` on beta, for a python float and a tensor parameter. On main all 4 cases fail (2 with the `ValueError` above, 2 with the wrong value), with this change they pass, and the existing `test_prioritized_parameter_scheduler` (32 cases) still passes.


vmoens (approved):
LGTM thanks for contributing

PR #4445, opened 2026-09-20
[BugFix] Drop the backend module from ParameterScheduler.state_dict

Nicholas022400701:
## Description

`ParameterScheduler.state_dict()` (torchrl/data/replay_buffers/scheduler.py L60-67) returns a copy of `self.__dict__` minus the sampler. `self.__dict__` also holds `self.backend`, which is the `numpy` or `torch` module itself, so `torch.save(scheduler.state_dict(), ...)` and `pickle.dumps(...)` raise `TypeError: cannot pickle 'module' object` for every `LinearScheduler`, `StepScheduler` and `LambdaScheduler`. The scheduler state can therefore not be checkpointed at all.

This PR drops the `backend` entry from `state_dict()` too. It is derived from the type of the initial parameter value, so a freshly constructed scheduler already has the right one and `load_state_dict` (a plain `__dict__.update`) keeps it. The docstring is updated to match.

## Test

`test_parameter_scheduler_state_dict` in test/rb/test_samplers.py steps a `StepScheduler` twice, saves its `state_dict` with `torch.save`, loads it into a second scheduler and checks the step count and the backend, for a python float and for a tensor parameter. Both cases fail on main with the `TypeError` above and pass with this change. The existing `test_prioritized_parameter_scheduler` (32 cases) still passes.


vmoens (approved):
LGTM thanks for contributing

Issue #4448, opened 2026-09-21
[Feature Request] Port TDMPC2 as a trainer


bsprenger:
## Motivation

[TD-MPC2](https://arxiv.org/abs/2310.16828) is one of the current state of the art algorithms for continuous control. It would be a great world model algorithm to include in torchRL's trainer framework, because:
- Unlike DreamerV3, it is decoder free
- It demonstrates state of the art performance on continuous control, outperforming models like DreamerV3 for single tasks
- It demonstrates state of the art performance on multi-task learning as well.

[The official repository](https://github.com/nicklashansen/tdmpc2) is easy to read and already uses pytorch, including some pieces of torchRL such as replay buffers.

I'm particularly interested to try this out on the Microduck environment for multi-task learning!

## Solution

Have a working trainer implementation of TD-MPC2, on a single task and state-based observations to start. Future work can add multi-task learning and pixel observations. The first implementation should just focus on getting the core pieces in place.

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

bsprenger:
I currently have a working branch for this, and will start opening a sequence of PRs to add each piece in isolation, so as to make reviews easier 🙏 stay tuned!

bsprenger:
I now have pushed all the PRs. I couldn't figure out how to diff each one against the previous, since these branches live in my own fork and not in the `upstream`. I'll rebase/retarget as they start getting merged so that it's easier to see what's new in each PR.

There are 8 PRs in total to fully implement the TD-MPC2 algorithm, trainer, and all config dataclasses:

#4449 
#4451 
#4452 
#4453 
#4454 
#4455 
#4456 
#4457 

I really hope that this stack makes it easier to review, but apologies in advance if this is actually more challenging, due to Github preventing diffing against a base that isn't in `upstream` 🙈

PR #4453, opened 2026-09-21
[Feature] Add TD-MPC2 Q-network ensemble

bsprenger:
## Description

> [!NOTE]
> This PR stacks on #4452 . I cannot find a good way to compare the diff against that base, since it is in my own fork.

This PR introduces the TD-MPC2 Q ensemble, which is a key part of the algorithm. The Q-ensemble consists of $N$ parallel Q-functions to compute TD targets during training of the model and to be used during the planning stage of the algorithm.

## Motivation and Context

This is the 4th PR in the chain intended to address #4448 . The Q-ensemble is described in depth in the [original paper](https://arxiv.org/pdf/2310.16828) and associated [implementation](https://github.com/nicklashansen/tdmpc2).

## Types of changes

## Checklist


torchrl/trainers/algorithms/tdmpc2.py
```diff
@@ -0,0 +1,409 @@
...
+    :meth:`forward` returns the logits of every Q-function. :meth:`reduce`
+    follows TD-MPC2's value-estimation rule by selecting two Q-functions at
+    random, decoding their logits, and taking either their minimum or average.
+    Online, detached, and target parameter sources are available for both
+    operations.
+
+    Args:
+        q_networks: Sequence of identically shaped single-network Q-functions.
+        num_bins: Number of categorical bins. ``0`` uses an identity output and
+            ``1`` uses a symmetric exponential output.
```
vmoens: Advertised num_bins modes are rejected.

The Q-ensemble documentation says 0 provides identity regression and 1 provides symlog regression, but the [constructor rejects both](https://github.com/pytorch/rl/pull/4453/files). The config factories and loss also only support categorical mode.

Either implement the documented/reference behavior or explicitly narrow the API and documentation to num_bins > 1.

bsprenger: I updated the documented behaviour. Thanks!

PR #4454, opened 2026-09-21
[Feature] Add TD-MPC2 loss module

bsprenger:
## Description

> [!NOTE]
> This PR stacks on #4453 . I cannot find a good way to compare the diff against that base, since it is in my own fork.
> Once it is merged, I will re-target this one to make the diff easier to read.

This PR introduces the core loss objective for the TD-MPC2 algorithm, porting over the logic into the trainer framework.

## Motivation and Context

5th PR in the chain to address #4448 .  The design of the loss is is described in depth in the [original paper](https://arxiv.org/pdf/2310.16828) and associated [implementation](https://github.com/nicklashansen/tdmpc2).

- [x] I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

What types of changes does your code introduce? Remove all that do not apply:

- [x] New feature (non-breaking change which adds core functionality)

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.
If you are unsure about any of these, don't hesitate to ask. We are here to help!

- [x] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.

PR #4455, opened 2026-09-21
[Feature] Add TD-MPC2 planner

bsprenger:
## Description

> [!NOTE]
> This PR stacks on #4454 . I cannot find a good way to compare the diff against that base, since it is in my own fork.

This PR introduces the central part of the TD-MPC2 algorithm: the planner module. The planner is responsible for rolling out candidate trajectories at inference time and planning which sequence of actions leads to the best outcome.

## Motivation and Context
6th PR in the chain to address #4448 . The planning algorithm is described in depth in the [original paper](https://arxiv.org/pdf/2310.16828) and associated [implementation](https://github.com/nicklashansen/tdmpc2).

## Types of changes

## Checklist


test/modules/test_tdmpc2.py
vmoens, line 1: Let's use test classes whenever we can + short test names

torchrl/trainers/algorithms/tdmpc2.py
vmoens, line 1: not sure modules/functionals really belong here.

Also another finding from Codex:

Evaluation-time exploration cannot be disabled.

The planner [always adds Gaussian noise to the selected action](https://github.com/pytorch/rl/pull/4455/files). The reference TD-MPC2 implementation skips this perturbation in evaluation mode. Add an explicit evaluation mode or respect TorchRL’s interaction type, with a regression test.

test/modules/test_tdmpc2.py
bsprenger: Done, I refactored the whole chain of PRs to use test classes and short names. These changes are folded into the commits on the PRs that introduced each feature.

torchrl/trainers/algorithms/tdmpc2.py
bsprenger: I moved the modules and helpers -- does the new layout look better to you? 

And nice find from Codex. I missed that because in the [official implementation here](https://github.com/nicklashansen/tdmpc2/blob/main/tdmpc2/evaluate.py#L80), the `evaluate.py` script actually does not set this eval mode. So I thought it was deprecated/unused. Looking closer , it is used in a few other places, so I added that here. Thanks!

PR #4457, opened 2026-09-21
[Feature] Add TD-MPC2 Trainer

bsprenger:
## Description

> [!NOTE]
> This PR stacks on #4456 . I cannot find a good way to compare the diff against that base, since it is in my own fork.

This PR adds a trainer-based implementation of TD-MPC2, a scalable model-based reinforcement learning algorithm for continuous control. 

Original paper: [paper](https://arxiv.org/abs/2310.16828)
Original Implementation: [implementation](https://github.com/nicklashansen/tdmpc2)

TD-MPC2 learns an implicit latent world model and performs short-horizon planning at action time. The original work reports strong performance across 104 online tasks with a single hyperparameter configuration.

https://github.com/user-attachments/assets/e3d7e7e1-5c8f-41ef-b772-0715a2e50b58

## Motivation and Context

Final PR towards addressing #4448 .

TD-MPC2 is an important model-based RL baseline for continuous control because it combines learned latent dynamics with decision-time planning, providing a practical balance between data efficiency, robustness, and scalability.

The modular implementation also exposes the individual world-model, policy, value, planning, and optimization components for reuse and further research.

## Types of changes

## Checklist


vmoens (commented):
Let's bring the sota implementation to the CI as well:

There is no sota-check/run_tdmpc2.sh and no TD-MPC2 command in .github/unittest/linux_sota/scripts/test_sota.py.

Consequently, the new runnable trainer is not exercised by SOTA CI, contrary to the repository requirements.

torchrl/trainers/algorithms/configs/trainers.py
```diff
@@ -1668,6 +1674,175 @@ def _make_td3_trainer(*args, **kwargs) -> TD3Trainer:
...
+    if optimization_stepper is not None and (
+        optimizer_model is not None or optimizer_actor is not None
+    ):
+        raise ValueError(
+            "Pass optimizers directly or provide optimization_stepper, not both."
+        )
+
+    if optimization_stepper is None:
+        capturable_setting = (
+            next(loss_module.world_model.parameters()).device.type == "cuda"
```
vmoens: device placement is inferred from the first parameter.

Both the [replay sample device](https://github.com/pytorch/rl/pull/4457/files) and [optimizer capturable setting](https://github.com/pytorch/rl/pull/4457/files) assume the whole model uses its first parameter’s device. 

This directly violates the module-device rule (ie, modern architectures are often using PP so we can't guarantee anything from the first parameter of a model) and breaks mixed-device modules. Derive placement from the relevant input/component or leave it unset when there is no unique device.

bsprenger: Fixed. I made the replay sample use the same device as the replay buffer storage (which is the same pattern other trainers use like SAC/DQN/TD3). For `capturable`, I made it so that it can be configured or will leave it unset.

bsprenger:
@vmoens I added it to the CI scripts, thanks for catching that 👍

.github/unittest/linux_sota/scripts/test_sota.py
```diff
@@ -17,6 +17,20 @@
...
+  collector.init_random_frames=10 \
+  collector.frames_per_batch=16 \
+  trainer.batch_size=4 \
+  trainer.optim_steps_per_batch=1 \
+  trainer.seed_pretrain_steps=1 \
+  models.planner.num_samples=16 \
+  models.planner.num_elites=4 \
+  models.planner.num_pi_trajs=2 \
+  models.planner.iterations=2 \
+  logger.backend= \
```
theap06: this passes logger.backend but the recipe uses logger@logger: csv and CSVLoggerConfig has no backend field. Right now the hydra fails due to that. same for the overrides in sota-check/run_tdmpc2.sh.

sota-implementations/tdmpc2_trainer/config/config.yaml
```diff
@@ -0,0 +1,185 @@
...
+  world_model: ${models.world_model}
+  policy_prior: ${models.policy_prior}
+  q_ensemble: ${models.q_ensemble}
+  planner: ${models.planner}
+  replay_buffer: ${replay_buffer}
+  logger: ${logger}
+  hooks:
+    - ${metric_hooks.training_reward}
+    - ${metric_hooks.training_action}
+  batch_size: 128
```
theap06: I noticed that for our config you are setting it to 128 compared to the 256 in the reference tdmpc2 config file has. Is there a reason for that?

theap06 (changes requested):
Left some feedback. overall looks p good. Also, this pr needs a rebase

PR #4462, opened 2026-09-21
[BugFix] Squeeze a Choice spec through its choices

### LINKED ISSUE #4461: [BUG] Choice.squeeze() raises TypeError because the base implementation rebuilds the spec from shape
## Describe the bug

`Choice.squeeze()` raises `TypeError`. `Choice` overrides `expand`, `unsqueeze`, `clone`, `_reshape` and `to`, but not `squeeze`, so it falls through to `TensorSpec.squeeze`, which rebuilds the spec with `self.__class__(shape=..., device=..., dtype=...)`. `Choice.__init__` takes a list of specs and accepts none of those keywords.

The same call on every other spec type works, so anything that squeezes specs generically (a transform, or simply undoing an `unsqueeze`) breaks as soon as one of them is a `Choice`.

## To Reproduce

```python
from torchrl.data import Bounded, Choice

spec = Choice([Bounded(0, 2.5, (1, 2)), Bounded(10, 12, (1, 2))])
print(spec.unsqueeze(0).shape)  # torch.Size([1, 1, 2]), fine
spec.squeeze(0)
```

```bash
  File "repro.py", line 5, in <module>
    spec.squeeze(0)
  File "torchrl/data/tensor_specs.py", line 980, in squeeze
    return self.__class__(shape=shape, device=self.device, dtype=self.dtype)
TypeError: Choice.__init__() got an unexpected keyword argument 'shape'
```

`spec.squeeze()` with no argument fails the same way. `spec.squeeze(1)` happens to work, because dimension 1 is not of size 1 and the base method returns `self` before rebuilding.

## Expected behavior

`squeeze` returns a `Choice` whose choices are squeezed, as its sibling shape operations already do:

```python
def expand(self, *shape):
    return self.__class__([choice.expand(*shape) for choice in self._choices])

def unsqueeze(self, dim: int):
    return self.__class__([choice.unsqueeze(dim) for choice in self._choices])
```

## System info

Installed from source at 3ae78133a.

```python
import torchrl, numpy, sys
```

Run from a source checkout at 3ae78133a, so `torchrl.__version__` prints `None`. numpy 2.5.2, torch 2.11.0+cu130, Python 3.13.15, linux.

## Reason and Possible fixes

`torchrl/data/tensor_specs.py`, base implementation at line 980:

```python
shape = _squeezed_shape(self.shape, dim)
if shape is None:
    return self
return self.__class__(shape=shape, device=self.device, dtype=self.dtype)
```

Adding the missing override to `Choice` next to `unsqueeze` fixes it:

```python
def squeeze(self, dim: int | None = None):
    return self.__class__([choice.squeeze(dim) for choice in self._choices])
```

I found this while sweeping the spec classes for invariants (`is_in(rand())`, `clone() == spec`, `unsqueeze` then `squeeze`), and `Choice` was the only class where a shape operation raised. I am happy to send the PR with a regression test in `TestChoiceSpec`.

## Checklist

yupengtang:
## Description

`Choice` overrides `expand`, `unsqueeze`, `clone`, `_reshape` and `to`, but not `squeeze`. So `squeeze` fell through to `TensorSpec.squeeze`, which rebuilds the spec with `self.__class__(shape=..., device=..., dtype=...)`, and `Choice.__init__` takes a list of specs and accepts none of those keywords:

```python
from torchrl.data import Bounded, Choice

spec = Choice([Bounded(0, 2.5, (1, 2)), Bounded(10, 12, (1, 2))])
spec.squeeze(0)
# TypeError: Choice.__init__() got an unexpected keyword argument 'shape'
```

The fix delegates to the choices, the way the neighbouring overrides already do.

`spec.squeeze(1)` never raised, because dimension 1 is not of size 1 and the base method returns `self` before rebuilding. The regression test covers that case too, so the no-op path stays a no-op.

## Motivation and Context

Anything that squeezes specs generically breaks when one of them is a `Choice`, including simply undoing an `unsqueeze`. I found it while sweeping the spec classes for invariants (`is_in(rand())`, `clone() == spec`, `unsqueeze` followed by `squeeze`); `Choice` was the only class where a shape operation raised.

close #4461

## Types of changes

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) document.
- [x] My change requires a change to the documentation. (no, the behaviour now matches what the other shape ops document)

## Tests

`test_squeeze` in `TestChoiceSpec`, parametrized over `dim` in `None, 0, 1`, asserting through the public API only (shape, `num_choices`, `rand()`, `is_in`).

- With the fix: `pytest test/test_specs.py::TestChoiceSpec` gives 7 passed.
- Without the `tensor_specs.py` change: `test_squeeze[None]` and `test_squeeze[0]` fail with the `TypeError` above, `test_squeeze[1]` still passes.
- Whole file: `pytest test/test_specs.py` gives 1798 passed, 9 skipped.

AI assistance (Claude Code) was used here, mostly for the invariant sweep that surfaced the bug and for drafting this description. I reviewed the change and ran the tests above.


vmoens (approved):
LGTM thanks!

PR #4475, opened 2026-09-22
[BugFix] Drop VLAWrapperBase device cache

### LINKED ISSUE #4467: [BUG] VLAWrapperBase keeps constructor self.device
## Describe the bug

`VLAWrapperBase` is an `nn.Module` and stores the constructor `device` as Python state:

```python
self.device = None if device is None else torch.device(device)
```

The kwarg is not documented. Runtime placement already uses `out.device` / `chunk.device`, so `device="cuda"` is a silent no-op for the base class: it neither moves parameters nor is consulted after init.

`.to("meta")` / `.cpu()` / `.cuda()` move parameters and buffers, not that Python attribute. A CPU-constructed wrapper still reports `self.device == torch.device("cpu")` after `.to("meta")`.

Related to #4224, #4345, and #4346.

## To Reproduce

```python
import torch
from tensordict import NonTensorStack, TensorDict
from torchrl.modules.vla import TinyVLA, VLAWrapperBase

base = VLAWrapperBase(action_dim=2, chunk_size=2, device="cpu")
assert base.device == torch.device("cpu")
base.to("meta")
assert base.device == torch.device("cpu")  # stale: .to() did not update it

policy = TinyVLA(action_dim=2, chunk_size=2, device="cpu")
td = TensorDict(
    {
        "observation": {
            "image": torch.zeros(1, 3, 16, 16, dtype=torch.uint8),
            "state": torch.zeros(1, 5),
        },
        "language_instruction": NonTensorStack("pick"),
    },
    batch_size=[1],
)
policy(td)  # materialize lazy modules
policy.to("meta")
assert next(policy.parameters()).device.type == "meta"
assert policy.device == torch.device("cpu")  # stale constructor cache
```

## Expected behavior

Do not keep a single-device attribute on the module. The constructor may still accept `device=` so subclasses can place parameters at init (`TinyVLA` already calls `self.to(device)`). It must not retain that value as module state. After `.to("meta")`, there must be no Python `self.device` left at the constructor value. Runtime placement should keep using the involved tensor (`chunk.device` / `out.device`).

## Checklist

YeonwooSung:
## Description

`VLAWrapperBase` assigned `self.device` from the constructor `device` argument. `.to()` / `.cpu()` / `.cuda()` move parameters and buffers, not that Python attribute, so the cache went stale. The kwarg was also undocumented; the base class never used it after init (`device="cuda"` was a silent no-op), and runtime placement already uses `out.device` / `chunk.device`.

The constructor still accepts `device=` for API stability and for subclasses that place parameters at init (`TinyVLA` already calls `self.to(device)`). The value is not stored as module state.

None of the `VLAWrapperBase` subclasses (`TinyVLA`, `LeRobotPolicyWrapper`, `OpenVLAOFTWrapper`, `OpenVLAOFTL1Wrapper`) read `self.device`. `TinyVLA` already derives runtime placement from `image.device`.

### Code example

```python
import torch
from tensordict import NonTensorStack, TensorDict
from torchrl.modules.vla import TinyVLA, VLAWrapperBase

base = VLAWrapperBase(action_dim=2, chunk_size=2, device="cpu")
assert "device" not in vars(base)
base.to("meta")
assert "device" not in vars(base)

policy = TinyVLA(action_dim=2, chunk_size=2, device="cpu")
td = TensorDict(
    {
        "observation": {
            "image": torch.zeros(1, 3, 16, 16, dtype=torch.uint8),
            "state": torch.zeros(1, 5),
        },
        "language_instruction": NonTensorStack("pick"),
    },
    batch_size=[1],
)
policy(td)
policy.to("meta")
assert "device" not in vars(policy)
assert {p.device.type for p in policy.parameters()} == {"meta"}
```

## Motivation and Context

A cached `self.device` violates TorchRL's module-device contract (Agents.md 6a) and breaks after `.to("meta")`.

close #4467

Same class of bug as #4224, #4345, and #4346.

## Types of changes

## Checklist


torchrl/modules/vla/common.py
```diff
@@ -219,7 +223,7 @@ def __init__(
         self.default_interaction_type = default_interaction_type
         self.log_probs_mode = log_probs_mode
         self.inplace = True if inplace is None else inplace
-        self.device = None if device is None else torch.device(device)
+        # `device` is accepted for subclass init placement and is not stored.
```
vmoens: This looks bc breaking to me, we should raise a warning when device is accessed and announce the removal in 0.16

YeonwooSung: `VLAWrapperBase.device` is a deprecated descriptor until v0.16: get/set emit `DeprecationWarning` and return the constructor value. It is not stored as `self.device` on the module, so `.to()` still only moves parameters. Runtime placement is unchanged (`chunk.device` / `out.device`).

